PR #2751, opened 2022-05-12
Installed Apps page revamp

leog:
## What does this PR do?

Revamping "Installed Apps" page to just show necessary things.

Fixes #2715
Fixes #2571

Loom Video: https://www.loom.com/share/0870dda91c484e87995a0be26c7f646f
Loom Video follow-up: https://www.loom.com/share/8531837b3eb44506bcd42b9eccdc357d

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [X] Other

## How should this be tested?

- Go to "Installed Apps" page and see only installed apps shown
- New empty state would be visible if no installed app is present
- Webhooks were moved to "Developer" tab in Settings along with API Keys (moved from "Security" tab)
- Connect additional calendars section in Installed Apps page was moved to a new dropdown next to default calendar dropdown

PR #2945, opened 2022-06-02
Fix uninstalled status of giphy and hubspot

hariombalhara:
## What does this PR do?

- Fixes uninstalled status of Hubspot and Giphy apps
- Hubspot needs to check environment variables, Giphy is always installed.

## Checklist


hariombalhara:
Just checked that it's already fixed with revamped installed apps page as it only shows those apps which are actually connected.

zomars:
Tagging #2751 for visibility

PR #2956, opened 2022-06-03
update google api callback url

chasebolt:
## What does this PR do?

callback URL has changed and is no longer valid in the README. updating to the correct URL required.

## Type of change


zomars:
These docs are not for google login. Are for adding a Google calendar, in which case the old endpoint is still the correct one. Feel free to add both or close this PR.

chasebolt:
ah i missed that, thanks!

PR #2969, opened 2022-06-06
Test: Embed-Preview - Ensure embed.js loads.

hariombalhara:
Tests that could have avoided [embed-preview regression ](https://github.com/calcom/cal.com/pull/2962)on production. These tests ensure that preview.html actually is able to load embed.js(which wasn't being loaded in preview in production)


PeerRich (approved):
good test

PR #3001, opened 2022-06-07
Disable dark mode on the success page only for the organizer 

CarinaWolli:
## What does this PR do?

Right now, dark mode is disabled on the success page for all logged-in users. This PR disables it only when a user is redirected to the success page after clicking on a booking to see the booking details.

## Type of change

## How should this be tested?

- Switch to dark mode
- Book a meeting --> success page should be in dark mode now 
- Click on a booking to see booking details --> success page should be in light mode

## Checklist


apps/web/pages/success.tsx
```diff
@@ -257,7 +257,7 @@ export default function Success(props: SuccessProps) {
     return t("emailed_you_and_attendees" + titleSuffix);
   }
   const userIsOwner = !!(session?.user?.id && eventType.users.find((user) => (user.id = session.user.id)));
-  const { isReady, Theme } = useTheme(userIsOwner ? "light" : props.profile.theme);
+  const { isReady, Theme } = useTheme(listingStatus ? "light" : props.profile.theme);
```
CarinaWolli, line 260: disable dark mode when we click on a booking to see the booking details

apps/web/pages/bookings/[status].tsx
```diff
@@ -60,61 +62,64 @@ export default function Bookings() {
...
-                  headline={t("no_status_bookings_yet", { status: t(status) })}
-                  description={t("no_status_bookings_yet_description", {
-                    status: t(status),
-                    description: descriptionByStatus[status],
-                  })}
-                />
-              )}
+    isReady && (
+      <Shell heading={t("bookings")} subtitle={t("bookings_description")} customLoader={<SkeletonLoader />}>
+        <Theme />
```
Comment (zomars) [inline apps/web/pages/bookings/[status].tsx:67]: Shouldn't we handle this in the `Shell` itself? 🤔

apps/web/pages/success.tsx
alannnc: What is the original purpose of listingStatus? Probably we can use a new parameter that can tell us that we navigated from a booking list click and change the theme from there.

alannnc (commented):
Looks and works fine, just a comment about using a specific parameter to trigger light/dark theme.

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -166,6 +166,7 @@ const BookingPage = ({
           location,
           eventName: profile.eventName || "",
           bookingId: id,
+          isSuccessBookingPage: true,
```
CarinaWolli, line 169: new parameter that defines if user is redirected to that page after successfully booking an event

apps/web/pages/success.tsx
```diff
@@ -622,10 +622,15 @@ function RecurringBookings({
 }: RecurringBookingsProps) {
   const [moreEventsVisible, setMoreEventsVisible] = useState(false);
   const { t } = useLocale();
-  return !isReschedule && recurringBookings && listingStatus === "upcoming" ? (
+
+  const recurringBookingsSorted = recurringBookings
```
CarinaWolli, line 626: dates for recurring events currently aren't sorted

apps/web/pages/success.tsx
CarinaWolli: `listingStatus` is used for recurring events, to know if we should show all dates (when coming from the "upcoming" section) or only the specific date of the booking (when coming from "recurring"). I introduced a new parameter `isSucessBookingPage` that defines if user is redirected to the page after successfully booking an event (https://github.com/calcom/cal.com/pull/3001/files#r891948713)

apps/web/components/booking/pages/BookingPage.tsx
emrysal: When does this variable turn false?

CarinaWolli: always true when it comes from the booking page, but we also redirect to /success from bookingListItem to see the booking details. So it is here to check if we we redirected from the booking page or come from a bookingListItem

pumfleet (approved):

PR #3007, opened 2022-06-07
Prevents past booking confirmation emails to be sent

### LINKED ISSUE #2907: Past booking confirmation emails shouldn't be sent
### Issue Summary

It happens constantly that some booking may expire requiring confirmation. But the email reminders are being sent anyways. If you click the link you get sent to the upcoming bookings page but there's nothing be done anymore.

zomars:
## What does this PR do?

Fixes #2907

**Environment**: Staging

## Type of change

## How should this be tested?

- [ ] Make an opt-in booking that ends very soon
- [ ] You should not receive a reminder email


apps/web/pages/api/cron/bookingReminder.ts
```diff
@@ -29,6 +29,8 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         createdAt: {
           lte: dayjs().add(-interval, "minutes").toDate(),
         },
+        // Only send reminders if the event hasn't finished
+        endTime: { gte: new Date() },
```
alannnc, line 33: Wouldn’t be better if we changed this to startTime? And add little interval time? What I understand its that this reminder its for accept pending bookings, right?
I think we don't wanna send email if the booking has started and not finish

PeerRich (approved):
this has been annoying me quite frequently, glad you fixed it.🙏

PR #3035, opened 2022-06-09
Feature/booking page refactor

emrysal:
Goal of this PR: Get rid of getServerSideProps altogether, use getStaticProps & tRPC to make slot lookahead possible.

Current state: Able to take into account workingHours, the DatePicker is entirely revamped (but looks the same). Much will happen serverside going forward.


apps/web/server/routers/viewer/slots.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+        return {
+          workingHours,
+          busy: busy.map((a) => ({
+            start: dayjs(a.start).subtract(currentUser.bufferTime, "minute").toDate(),
+            end: dayjs(a.end).add(currentUser.bufferTime, "minute").toDate(),
+          })),
+        };
+      })
+    );
```
zomars, line 208: Made some adjustments here to use `getUserAvailability` as much as possible.

zomars:
Instead of making code suggestions made another PR https://github.com/calcom/cal.com/pull/3067

apps/web/pages/[user].tsx
```diff
@@ -46,71 +46,53 @@ interface EvtsToVerify {
...
-    ? props.users.map((user) => {
-        return user.username || "";
-      })
-    : [];
-  const eventTypes = isDynamicGroup
-    ? defaultEvents.map((event) => {
-        event.description = getDynamicEventDescription(dynamicUsernames, event.slug);
-        return event;
-      })
-    : props.eventTypes;
```
Comment (zomars) [inline apps/web/pages/[user].tsx:71]: We did all this already on the server. Let's just pass it as props

apps/web/pages/[user]/[type].tsx
```diff
@@ -144,187 +106,185 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
...
-            hideBranding: true,
-            plan: true,
-            timeZone: true,
-          },
-        },
-      },
-    });
+  const eventType = user.eventTypes.find((et, i) =>
+    user.plan === UserPlan.FREE ? i === 0 && et.slug === slug : et.slug === slug
+  );
```
Comment (zomars) [inline apps/web/pages/[user]/[type].tsx:117]: Fixes E2E tests that allowed free users to use more than 1 event type

apps/web/server/routers/viewer/slots.tsx
```diff
@@ -0,0 +1,256 @@
...
+    endTime: stringToDayjs,
+    // Event type ID
+    eventTypeId: z.number().optional(),
+    // or list of users (for dynamic events)
+    usernameList: z.array(z.string()).optional(),
+  })
+  .refine(
+    (data) => !!data.eventTypeId || !!data.usernameList,
+    "Either usernameList or eventTypeId should be filled in."
+  );
```
zomars, line 31: So we have one or another

packages/core/getUserAvailability.ts
```diff
@@ -82,6 +106,12 @@ export async function getUserAvailability(
   let eventType: EventType | null = initialData?.eventType || null;
   if (!eventType && eventTypeId) eventType = await getEventType(eventTypeId);
 
+  /* Current logic is if a booking is in a time slot mark it as busy, but seats can have more than one attendee so grab
+  current bookings with a seats event type and display them on the calendar, even if they are full */
+  let currentSeats: CurrentSeats | null = initialData?.currentSeats || null;
+  if (!currentSeats && eventType?.seatsPerTimeSlot)
+    currentSeats = await getCurrentSeats(eventType.id, dateFrom, dateTo);
```
zomars, line 113: So we don't overfetch if we already have this data

packages/lib/hooks/useSlots.ts
```diff
@@ -0,0 +1,8 @@
+type UseSlotsProps = {
+  boundaries: {
+    startTime: Date;
+    endTime: Date;
+  }[];
+};
+
+export const useSlots = ({ boundaries }: UseSlotsProps) => {};
```
zomars, line 8: Is this empty on purpose @emrysal ?

zomars (approved):
Looking Gucci

packages/lib/hooks/useSlots.ts
emrysal: I think this is like, very early designing when I was still thinking "boundaries" are the way to go. will delete (it's unused)

apps/web/server/routers/viewer/slots.tsx
```diff
@@ -169,76 +163,39 @@ export const slotsRouter = createRouter().query("getSchedule", {
...
-            startTime: true,
-            _count: {
-              select: {
-                attendees: true,
-              },
-            },
-          },
-        })
-      : [];
+    const workingHours = userSchedules.flatMap((s) => s.workingHours);
```
emrysal, line 196: This works, but caused an issue I'm working around in the simplest method I could think of: #3082

PR #3037, opened 2022-06-10
Manage Link showing for organisers

sean-brydon:
## What does this PR do?

This PR allows the organisers/attendee[0] to be given the option to manage the event - reschedule/cancel 

Also hides info field if its an empty string. We were getting headings showing with no content (additional notes)

Fixes # (issue)

## Type of change

## Checklist


packages/emails/src/components/Info.tsx
```diff
@@ -7,7 +7,7 @@ export const Info = (props: {
   withSpacer?: boolean;
   lineThrough?: boolean;
 }) => {
-  if (!props.description) return null;
+  if (!props.description || props.description=="") return null;
```
sean-brydon, line 10: Some cases like additionlNotes are set to "" and was causing a random header to appear in the email field

packages/emails/src/components/ManageLink.tsx
```diff
@@ -4,27 +4,29 @@ import type { CalendarEvent, Person } from "@calcom/types/Calendar";
...
-        <>{t("need_to_reschedule_or_cancel")}</>
-      </p>
-      <p style={{ fontWeight: 400, lineHeight: "24px" }}>
-        <a href={getCancelLink(props.calEvent)} style={{ color: "#3e3e3e" }}>
-          <>{t("manage_this_event")}</>
-        </a>
-      </p>
-    </div>
-  );
+  if(props.attendee.email === props.calEvent.attendees[0].email || props.calEvent.organizer.email  === props.attendee.email ){
```
sean-brydon, line 8: I switch this if statement cause I couldn't for the life of me to get it to work and I'm not sure why :S

zomars: Let's figure out why and take the learning instead of just patching

pumfleet (approved):
Looks good to me

zomars:
It wasn't ready to merge we found out some quirks on emails @baileypumfleet

PR #3038, opened 2022-06-10
CalDav Setup page

leog:
## What does this PR do?
Moving CalDav integration setup into its own setup page as we have for Apple Calendar and Zapier.
Mouse_Highlight_Overlay
Also added a "Cancel" button to both Apple and CalDav to go back to the previous page
Included localization to the Apple Calendar setup page as it was missing

## Type of change

## How should this be tested?

Add a new integration with CalDav and notice the new setup page


packages/app-store/applecalendar/pages/setup/index.tsx
```diff
@@ -70,23 +70,26 @@ export default function AppleCalendarSetup() {
...
                 {errorMessage && <Alert severity="error" title={errorMessage} className="my-4" />}
-                <div className="mt-5 sm:mt-4 sm:flex sm:flex-row-reverse">
+                <div className="mt-5 space-x-2 space-x-reverse sm:mt-4 sm:flex sm:flex-row-reverse">
+                  <Button type="button" onClick={() => router.back()}>
+                    {t("cancel")}
+                  </Button>
                   <Button type="submit" loading={form.formState.isSubmitting}>
-                    Save
+                    {t("save")}
                   </Button>
```
PeerRich, line 93: where does this flex-row-reverse come from all the time? lets get rid of this. likely copied from an old component.

Cancel should always be left, with `color="secondary"`  never two primary buttons

PeerRich: just in case cc @Jaibles

PeerRich (changes requested):

packages/app-store/applecalendar/pages/setup/index.tsx
```diff
@@ -70,23 +70,26 @@ export default function AppleCalendarSetup() {
...
-                    label="Password"
+                    label={t("password")}
                     placeholder="•••••••••••••"
                     autoComplete="password"
                   />
                 </fieldset>
 
                 {errorMessage && <Alert severity="error" title={errorMessage} className="my-4" />}
-                <div className="mt-5 sm:mt-4 sm:flex sm:flex-row-reverse">
+                <div className="mt-5 space-x-2 space-x-reverse sm:mt-4 sm:flex sm:flex-row-reverse">
```
PeerRich, line 87:
```suggestion
                <div className="mt-5 space-x-2 sm:mt-4 flex">
```

packages/app-store/applecalendar/pages/setup/index.tsx
```diff
@@ -70,23 +70,26 @@ export default function AppleCalendarSetup() {
...
+                  <Button type="button" onClick={() => router.back()}>
```
PeerRich, line 88:
```suggestion
                  <Button color="secondary" type="button" onClick={() => router.back()}>
```

packages/app-store/applecalendar/pages/setup/index.tsx
PeerRich: wait this is not correct.

PeerRich:
@leog it should be like this: 
image

PeerRich (changes requested):
@leog it should be like this

image

can you double check if the other set up pages have the same problem?

Cancel is always color="secondary"

leog:
Done! Good catch @PeerRich. Here they are now:
caldav_server
apple_server

PeerRich (approved):

PR #3045, opened 2022-06-13
Fix/app multiple installs

alannnc:
## What does this PR do?

Fixed some conditionals on app page, to display that they can only be installed once and for calendars they can be installed multiple times

Fixes # (issue)

 Loom Video: https://www.loom.com/share/2a35144a84f942418f75d916b7e668cd

## Type of change

## How should this be tested?

- [ ] Install Wipe My Cal app, return to app page and it should say Installed
- [ ] Install any calendar app, return to app page and it should display 1 installed connection with the add another button.

## Checklist


apps/web/components/App.tsx
```diff
@@ -36,6 +36,7 @@ export default function App({
   email,
   tos,
   privacy,
+  multipleInstall = false,
```
alannnc, line 39: New flag to know if an app can be installed multiple times

apps/web/components/App.tsx
```diff
@@ -114,36 +116,59 @@ export default function App({
               </div>
 
               <div className="mt-4 sm:mt-0 sm:text-right">
-                {!isLoading ? (
-                  isGlobal || installedAppCount > 0 ? (
-                    <div className="space-x-3">
+                {isLoading && <SkeletonButton width="24" height="10" />}
```
alannnc, line 119: Changed ternary operation for explicit conditional

hariombalhara (approved):

apps/web/components/App.tsx
```diff
@@ -114,36 +116,59 @@ export default function App({
...
+                          </Button>
+                        )}
+                      />
+                    )}
+                  </>
+                )}
+
+                {!isLoading && !multipleInstall && (
+                  <>
+                    {installedAppCount > 0 && (
```
leog, line 154: Google Meet and Cal Video should be showing the "Installed" treatment and as they are global, with `installedAppCount === 0` they are not showing. This will get them that treatment, tested locally. 
```suggestion
                    {(isGlobal || installedAppCount > 0) && (
```

apps/web/components/App.tsx
```diff
@@ -53,6 +54,7 @@ export default function App({
   email: string; // required
   tos?: string;
   privacy?: string;
+  multipleInstall?: AppType["multipleInstall"];
```
leog, line 57: This should be just `boolean`, right? 😄

leog (changes requested):
Great job @alannnc. Take a look at some comments I left, and also, I don't think we should be committing the `yarn.lock` file 🤔

zomars (changes requested):
I'd rather not add any more things to metadata as is going to be replace by the App model in the DB. Why not user the calendar category for now as it was before? Or there are any non-calendar apps that can be installed multiple times?

apps/web/components/App.tsx
alannnc: Yes, but with this, in the future if the value change you only have to change it on the origin.

alannnc:
> I'd rather not add any more things to metadata as is going to be replace by the App model in the DB. Why not user the calendar category for now as it was before? Or there are any non-calendar apps that can be installed multiple times?

Yeah, will switch to calendar instead then. Didn't about upcoming AppModel in DB

zomars (approved):

PR #3052, opened 2022-06-13
Cutting down calls for installed apps + other UI tweaks

leog:
## What does this PR do?

 - Cutting down installed apps page calls to just rely on a single call to installed apps of all variants.
 - Also, made sure the "Additional calendar" and "Destination calendar" dropdowns are showing correctly when the content gets too long. Installed_Apps___Cal_com
 - Correcting CalDav Setup page component name

## Type of change

## How should this be tested?

Go to Installed App page and notice the time to load to have improved. Also, for dropdown fix, install a very long named calendar (radicale used in the above screenshot) to see it cropping the name. Couldn't make "..." ellipsis show within `react-select` without pseudo-class hacky ways, once I found a way I will create a follow-up PR tackling it.


apps/web/components/AdditionalCalendarSelector.tsx
```diff
@@ -65,6 +65,10 @@ const AdditionalCalendarSelector = ({ isLoading }: AdditionalCalendarSelectorPro
                 return {
                   ...defaultStyles,
                   borderRadius: "2px",
+                  "@media only screen and (min-width: 640px)": {
```
leog, line 68: I miss Tailwind when things like this have to be done 😓

apps/web/components/integrations/CalendarListContainer.tsx
```diff
@@ -207,7 +207,11 @@ function ConnectedCalendarsList(props: Props) {
   );
 }
 
-export function CalendarListContainer(props: { heading?: boolean; fromOnboarding?: boolean }) {
+export function CalendarListContainer(props: {
+  heading?: boolean;
+  items: inferQueryOutput<"viewer.integrations">["items"];
```
leog, line 212: Passing down already fetched items to just show them, instead of having to fetch the passed down variant

packages/app-store/caldavcalendar/pages/setup/index.tsx
```diff
@@ -8,7 +8,7 @@ import { Alert } from "@calcom/ui/Alert";
 import Button from "@calcom/ui/Button";
 import { Form, TextField } from "@calcom/ui/form/fields";
 
-export default function AppleCalendarSetup() {
+export default function CalDavCalendarSetup() {
```
leog, line 11: Was not causing any problem but it was confusing

apps/web/components/integrations/CalendarListContainer.tsx
```diff
@@ -217,10 +221,12 @@ export function CalendarListContainer(props: { heading?: boolean; fromOnboarding
...
   const query = trpc.useQuery(["viewer.connectedCalendars"]);
-  const installedCalendars = trpc.useQuery([
-    "viewer.integrations",
-    { variant: "calendar", onlyInstalled: true },
-  ]);
+  let installedCalendars;
+  if (props.items) {
+    installedCalendars = props.items.filter((item) => item.variant == "conferencing");
+  } else {
+    installedCalendars = trpc.useQuery(["viewer.integrations", { variant: "calendar", onlyInstalled: true }]);
```
zomars, line 228: [Hooks should not be used conditionally](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

apps/web/pages/apps/installed.tsx
```diff
@@ -233,10 +230,10 @@ export default function IntegrationsPage() {
...
             return data.items.length > 0 ? (
               <>
-                <IntegrationsContainer variant="conferencing" />
-                <CalendarListContainer />
-                <IntegrationsContainer variant="payment" className="mt-8" />
-                <IntegrationsContainer variant="other" className="mt-8" />
+                <IntegrationsContainer items={data.items} variant="conferencing" />
+                <CalendarListContainer items={data.items} />
+                <IntegrationsContainer items={data.items} variant="payment" className="mt-8" />
+                <IntegrationsContainer items={data.items} variant="other" className="mt-8" />
```
zomars, line 236: We're discarding the benefits of using tRPC here, we should lookup how to pass initial data to queries instead.

zomars: Also by passing these props we're losing the portability of these "containers"

apps/web/pages/apps/installed.tsx
```diff
@@ -94,52 +95,48 @@ function ConnectOrDisconnectIntegrationButton(props: {
...
 
-const IntegrationsContainer = ({ variant, className = "" }: IntegrationsContainerProps): JSX.Element => {
+const IntegrationsContainer = ({
+  items,
+  variant,
+  className = "",
+}: IntegrationsContainerProps): JSX.Element => {
   const { t } = useLocale();
-  const query = trpc.useQuery(["viewer.integrations", { variant, onlyInstalled: true }], { suspense: true });
-
```
zomars, line 104: This should not be a problem, from React-query: [Deduping multiple requests for the same data into a single request](https://react-query.tanstack.com/overview#motivation)

If the query is not being deduped then something else is wrong. Let's focus on that.

zomars (dismissed):
Nice work so far @leog my two cents would be focusing on why tRPC and react-query is not behaving as expected instead of patching around it. 

My guess is that the added inputs  `{ variant, onlyInstalled: true }` are the culprits. Let's find a way to optimize this call maybe revert to calling all apps in the parent and find a way to filter on client or vice versa, removing the parent query and let each component fetch it's own variant.

apps/web/components/AdditionalCalendarSelector.tsx
PeerRich: can you not do that with tailwind? with `sm:max-w-xs` ?

PeerRich: or does the Select not take classNames?

leog: exactly, it doesn't take classes from a specific part of the component, I could add a class to the whole thing and add tailwind classes with `@apply` from our custom css files, but that somehow didn't feel more maintainable than this 😅

PeerRich: good point!

apps/web/pages/apps/installed.tsx
leog: Reverted

leog:
@zomars thank you very much for your feedback! I went through `react-query` and found a way to avoid invalidating a whole query, that I think it was messing with the cache for other calls.

apps/web/components/integrations/CalendarListContainer.tsx
```diff
@@ -207,13 +207,19 @@ function ConnectedCalendarsList(props: Props) {
...
+  fromOnboarding?: boolean;
+}) {
   const { t } = useLocale();
   const { heading = true, fromOnboarding } = props;
   const utils = trpc.useContext();
   const onChanged = () =>
     Promise.allSettled([
-      utils.invalidateQueries(["viewer.integrations"]),
+      utils.invalidateQueries(["viewer.integrations", { variant: "calendar", onlyInstalled: true }], {
+        exact: true,
```
leog, line 221: This is what I think is going to improve the cache management for `viewer.integrations` query.

leog:
Please @zomars @alannnc @emrysal, this is taking quite a bit to merge 😅

zomars (approved):

PR #3055, opened 2022-06-13
feat: Adds new customPrisma for hosted API

agustif:
## What does this PR do?
adds a new customPrisma() in our client, that will be used for the new Hosted API funcitonality, basically allowing us to let self-hosted enterprise edition users to use the API  through api.cal.com but point to their backend/database.

This will be done passing a custom datasource db connection string in API.
`customPrisma({ datasources: { db: { url: credentials } } }),
`


zomars (approved):

packages/lib/constants.ts
```diff
@@ -18,3 +18,4 @@ export const TRIAL_LIMIT_DAYS = 14;
 export const HOSTED_CAL_FEATURES = process.env.HOSTED_CAL_FEATURES || BASE_URL === "https://app.cal.com";
 /** @deprecated use `WEBAPP_URL` */
 export const NEXT_PUBLIC_BASE_URL = process.env.NEXT_PUBLIC_WEBAPP_URL || `https://${process.env.VERCEL_URL}`;
+export const PRISMA_CLIENT_CACHING_TIME = 1000 * 60 * 60 * 24;
```
zomars, line 21: Let's keep this in the API

zomars (changes requested):
Nitpicks

packages/lib/constants.ts
agustif: Done, will move to @calcom/api/lib/constants.ts

zomars (approved):

PR #3060, opened 2022-06-14
Update CalendarService.ts

### LINKED ISSUE #2864: Recurring events from caldav server not blocked

### Issue Summary

I am running the docker version of calcom and I have two caldav backends configured for my user. There are many recurring events in my calendar and none of these is blocked. Normal (one-time) events are blocked as they should.

### Steps to Reproduce

1. start docker version
2. run a caldav server, e.g. radicale and create recurring events
3. import calendar into calcom
4. go to booking
5. realize, that blocked dates are still available

I think this is a bug, as recurring events should block the corresponding dates

m4tze:
Fix timezone handling for recurring icalendar events

## What does this PR do?

It fixes the default utc interpretation of recurring events, as the comment reg. caldav specs is not correct, because we are dealing with icalender, which allows timezone information.

Fixes #2864 

-->

both! 

## Type of change

## How should this be tested?

It should be tested with existing tests regarding recurring events (maybe the tests have to be fixed to correct the utc misunderstanding)

## Checklist

### DIFF
--- a/packages/lib/CalendarService.ts
+++ b/packages/lib/CalendarService.ts
@@ -284,8 +284,14 @@ export default abstract class BaseCalendarService implements Calendar {
             }
           }
           if (!currentEvent) return;
-          // as pointed out in https://datatracker.ietf.org/doc/html/rfc4791#section-9.6.5
-          // recurring events are always in utc
+          // do not mix up caldav and icalendar! For the recurring events here, the timezone
+          // provided is relevant, not as pointed out in https://datatracker.ietf.org/doc/html/rfc4791#section-9.6.5
+          // where recurring events are always in utc (in caldav!). Thus, apply the time zone here.
+          if (vtimezone) {
+            const zone = new ICAL.Timezone(vtimezone);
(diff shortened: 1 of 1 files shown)


zomars:
Thank you for you contribution. Just to be clear. CalendarService is being used in both caldav and Apple calendar. Would this cause unexpected behavior in Apple calendars?

m4tze:
@zomars I cannot test this, as I do not have an apple calendar account. I guess not, as my fix does not convert the timestamps to an illegal timezone, it just applies the timezone that is stored in the event to the timestamps.

Besides this, I think the current implementation also calculates wrong dates for apple calendar (especially if the calendar timezone is not UTC and the server's timezone is not the same timezone as the calendar timezone). But this should better be confirmed by another user testing that setup.

zomars (approved):

m4tze:
@zomars I cannot access the Vercel services. Thus, I cannot help you here right now.

pumfleet (approved):

PR #2829, opened 2022-05-20
Putting "Additional Input Required" doesn't bar you from creating a booking #2660

gitstart:
## What does this PR do?

Fixes #2660

Loom Video: https://www.loom.com/share/baba361ae36d4c1ba6ee6e97369c09ce

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #3062, opened 2022-06-15
Putting "Additional Input Required" doesn't bar you from creating a booking

### LINKED ISSUE #2660: Putting "Additional Input Required" doesn't bar you from creating a booking

### Issue Summary

when you click 'is required' for a new custom input field, you can still make a booking without filling in the custom input field

### Steps to Reproduce

1. (for example) Went to event types
2. Clicked on an event
3. Set an addtl input to be 'required'
4. Went and tried to book the link on incognito mode without entering the required information
5. Was still able to book with the 'required' info not put in

gitstart:
## What does this PR do?

Fixes #2660

Demo with the frontend validation off for testing: 
Loom Video: https://www.loom.com/share/b342b1d22301462a9adf5a6ad847bba9

Demo of the input validation:
https://www.loom.com/share/2fbeb4a4219c4d3ea4c92206604d9c85

## Type of change


apps/web/pages/api/book/event.ts
```diff
@@ -377,11 +377,33 @@ async function handler(req: NextApiRequest) {
...
   }
+  const customInputEmpty: Array<string> = [];
+  customInputsPrisma?.customInputs.forEach(({ label }) => {
+    if (!customInputs[label]) {
+      customInputEmpty.push(label);
+    }
+  });
+
+  if (customInputEmpty.length > 0) {
+    throw new HttpError({ statusCode: 400, message: `Missing input ${customInputEmpty.toString()}` });
```
CarinaWolli, line 405: I don't that's needed. With your changes in the BookingPage we already make sure that the form can't be submitted when required fields are empty

yarn.lock
```diff
@@ -17630,4 +17630,4 @@ zwitch@^1.0.0:
 zwitch@^2.0.0:
   version "2.0.2"
   resolved "https://registry.yarnpkg.com/zwitch/-/zwitch-2.0.2.tgz#91f8d0e901ffa3d66599756dde7f57b17c95dce1"
-  integrity sha512-JZxotl7SxAJH0j7dN4pxsTV6ZLXoLdGME+PsjkL/DaBrVryK9kTGq06GfKrwcSOqypP+fdXGoCHE36b99fWVoA==
+  integrity sha512-JZxotl7SxAJH0j7dN4pxsTV6ZLXoLdGME+PsjkL/DaBrVryK9kTGq06GfKrwcSOqypP+fdXGoCHE36b99fWVoA==
```
CarinaWolli, line 17633: We shouldn't commit the changes from yarn.lock

CarinaWolli (changes requested):
Good job, works great for me! I left two comments that should be addressed first before I can approve it

apps/web/pages/api/book/event.ts
gitstart: Hi @CarinaWolli, thanks for the review! 

This was a suggestion from @zomars when we first pushed the validation
https://github.com/calcom/cal.com/pull/2829#pullrequestreview-985351190

what should we do next?

yarn.lock
gitstart: @CarinaWolli believe this was a bug as the lines are identical! will manage to fix it

apps/web/pages/api/book/event.ts
zomars: You're both right @CarinaWolli and @gitstart the BookingPage prevents required inputs to be missing in the frontend. But in order for this to be a complete solution we need to validate in the backend as well so someone cannot bypass the call by modifying the request. 

This solution may work but I would rather try to use a dynamic zod schema validation for these use cases. This validation should happen as early as possible in the request.

gitstart:
@zomars @CarinaWolli closing this as we will be working on the suggestion of validation from @zomars. We will push another one to avoid noise on your repo, thank you

PR #3086, opened 2022-06-16
Save currency to db

joeauyeung:
## What does this PR do?

This PR fixes the bug where currency was not saved to the DB so all Stripe transactions were happening in USD

Fixes # (issue)

## Type of change

## How should this be tested?

- Connect a Stripe account
	- For this test it should be in a currency other than USD
- Create a new event type and add a price
	- In the DB you should see the currency is saved as your default one from Stripe
- Edit an event type and add a price
- Create a booking, you should see your currency rendered on the availability and booking page
- Remove the Stripe app
- You should still be able to create and edit event types


PeerRich (approved):

zomars (changes requested):
This PR doesn't attack the root problem neither. Left my comments in #3096

zomars (approved):

PR #3087, opened 2022-06-16
Feature/parallel booking availability

zomars:
## What does this PR do?

- Enables caching for public endpoint
- ~~Fetches, caches and revalidates each day availability in parallel instead of batching~~
- Adds a 30 second caching layer for fetching external calendars
- Fixes out of bounds availability 

## Type of change

## How should this be tested?

- [ ] https://cal-git-feature-parallel-booking-availability-cal.vercel.app/pro/yoga-class
- [ ] See if it's faster or slower


apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -165,17 +166,26 @@ const SlotPicker = ({
...
+    /*  const { slots, isLoading } = useSlots({
+      eventTypeId: eventType.id,
+      startTime: dayjs(props.date).startOf("day").toDate(),
+      endTime: dayjs(props.date).endOf("day").toDate(),
+    });
+
+    console.log("slots", slots); */
+
+    return <Day {...props} />;
+  };
```
zomars, line 188: This is kinda what we we're talking about @emrysal

apps/web/pages/_app.tsx
```diff
@@ -70,8 +79,17 @@ export default withTRPC<AppRouter>({
...
+        splitLink({
+          // check for context property `skipBatch`
+          condition: (op) => op.context.skipBatch === true,
+          // when condition is true, use normal request
+          true: httpLink({ url }),
+          // when condition is false, use batching
+          false: httpBatchLink({
+            url,
+            /** @link https://github.com/trpc/trpc/issues/2008 */
+            // maxBatchSize: 7
```
zomars, line 91: Disabled until solved

KATT: FYI `maxBatchSize` will disappear in the new major in favor of a property called `maxURILength`

KATT: So I think you just don't want to use it.

zomars: And what would be de difference in functionality?

agustif: https://github.com/trpc/trpc/issues/1801#issue-1208401673
Add a new config option maxBatchURILength(?) to the httpBatchLink that will split the batch into smaller chunks before sending requests if the initial batch URI exceeds the limit.

Hey @zomars I don't know how it works now, but I'm thinking that since the trpc call it's just a long HTTP URI, it will limit the max size of a batch call, by the number of characters in the string but maybe @KATT or @jlalmes have more details on this

agustif (approved):
Looks great to me!

Cache is hitting after first test, but comparing to cal.dev/pro/yoga-class I'd say it's like 10x faster It does 2secs for the new code and 20secs in the old one! Great work as always @zomars

emrysal (changes requested):
Found an issue that if you go the the current month, then navigate to the next month, then navigate back - the slots disappear - will attempt at fixing 👍

emrysal:
Not seeing the same perf improvements as @agustif (in my case, testing with 5m slots), but that's OK - the fact this introduces caching per month is already another great step forward. Made a small fix to the disabled state on hover as the border was still showing (base>hover>disabled>dark-hover>dark-disabled) so had to make sure to "untoggle" the hover state border colour with follow up disable: variant.

emrysal (approved):
Fixed the issue navigating back to the current month :), works great @zomars

apps/web/pages/_app.tsx
jlalmes: ^ Yes - that's exactly it @agustif!

PR #3093, opened 2022-06-17
As Google Meet is always installed, this fixes the doubling

emrysal:
## What does this PR do?

* Fixes double google meet when you have google calendar installed.
* Fixes (see below) - also makes Google <LinkIcon> clickable entirely
* Fixes an updateEvent return value - this broke things for sure but not sure what 😓 
* Fixes types of VideoClient/CalendarClient
* Fixes Exchange 2013/2016 return values

image


PeerRich (approved):

alannnc (approved):
Tested and it's working

packages/emails/src/components/LocationInfo.tsx
```diff
@@ -30,19 +30,18 @@ export function LocationInfo(props: { calEvent: CalendarEvent; t: TFunction }) {
...
+            <a
+              href={meetingUrl}
+              target="_blank"
+              title={t("meeting_url")}
+              style={{ color: "#3E3E3E" }}
+              rel="noreferrer">
+              {providerName} <LinkIcon />
+            </a>
+          ) : (
+            <>{t("something_went_wrong")}</>
```
emrysal, line 43: @PeerRich What would you like to do here?

Basically internal error, so something went wrong and there is no meetingUrl :(

packages/emails/src/components/LocationInfo.tsx
```diff
@@ -84,26 +83,14 @@ export function LocationInfo(props: { calEvent: CalendarEvent; t: TFunction }) {
...
-              <>{t("meeting_url_provided_after_confirmed")}</>
-            </p>
-          ) : null
+          <a
+            href={hangoutLink}
+            target="_blank"
+            title={t("meeting_url")}
+            style={{ color: "#3E3E3E" }}
+            rel="noreferrer">
+            {providerName} <LinkIcon />
```
emrysal, line 92:
```suggestion
            Google <LinkIcon />
           
```

zomars (approved):

PR #3096, opened 2022-06-17
Grab currency from payment credentials

joeauyeung:
## What does this PR do?

This PR grabs a user's currency from their payment app credentials and renders it properly on the availability and booking page.

Fixes # (issue)

## Type of change

## How should this be tested?

- Add Stripe to your account
- Edit an event type to include a price
- When going to the availability and booking page of that event type the currency should render properly

## Checklist


apps/web/lib/getCurrency.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (paymentCredential === null) return "usd";
+
+  if (paymentCredential.type === "stripe_payment") {
+    const stripeKeySchema = z.object({
+      default_currency: z.string(),
+    });
+    const { default_currency } = stripeKeySchema.parse(paymentCredential.key);
+    return default_currency;
+  }
```
joeauyeung, line 27: Created this to future proof if we have more than one payment app

apps/web/lib/getCurrency.ts
```diff
@@ -0,0 +1,30 @@
+import { z } from "zod";
+
+import prisma from "@lib/prisma";
+
+const getCurrency = async (userId: number) => {
+  const paymentCredential = await prisma.credential.findFirst({
```
joeauyeung, line 6: Assuming a user can only have one payment app at a time

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -63,7 +63,9 @@ import type { AvailabilityTeamPageProps } from "../../../pages/team/[slug]/[type
 dayjs.extend(utc);
 dayjs.extend(customParseFormat);
 
-type Props = AvailabilityTeamPageProps | AvailabilityPageProps | DynamicAvailabilityPageProps;
+type Props = (AvailabilityTeamPageProps | AvailabilityPageProps | DynamicAvailabilityPageProps) & {
+  currency?: string;
+};
```
zomars, line 68: Let's add currency to each one to keep inference 

```suggestion
type Props = (AvailabilityTeamPageProps | AvailabilityPageProps | DynamicAvailabilityPageProps)
```

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -465,7 +467,7 @@ const AvailabilityPage = ({ profile, eventType }: Props) => {
                             <FormattedNumber
                               value={eventType.price / 100.0}
                               style="currency"
-                              currency={eventType.currency.toUpperCase()}
+                              currency={currency?.toUpperCase() || eventType.currency?.toUpperCase()}
```
zomars, line 470: Why it was override from eventTypes?

apps/web/lib/getCurrency.ts
zomars: I think it’s overkill, evenTypes should have the currency already. There’s no need for an extra query just to get the currency

zomars (changes requested):

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -515,7 +517,7 @@ const BookingPage = ({
                       <FormattedNumber
                         value={eventType.price / 100.0}
                         style="currency"
-                        currency={eventType.currency.toUpperCase()}
+                        currency={currency?.toUpperCase() || eventType.currency.toUpperCase()}
```
zomars, line 520: Instead of passing the currency manually, let's focus on figure out why the eventType doesn't have the correct currency and:

1. See if it can be resolved at runtime, or..
2. If the DB is wrong see if we can resolve it with a single migration

joeauyeung:
Got clarification of issue, moving fixes to #3086

joeauyeung:
Got clarification of issue, moving fixes to https://github.com/calcom/cal.com/pull/3086

PR #3099, opened 2022-06-17
Adding Close.com library for console

leog:
## What does this PR do?

This PR adds a library used for now in the Console to handle Close.com API (see https://github.com/calcom/console/pull/14), will be used for the Close.com App in the App Store as well if that comes to fruition.

## Type of change

- [X] Chore (non-breaking change which adds needed logic)

## How should this be tested?

No UI or functionality affected


agustif (approved):
LGTM ✅

PR #3100, opened 2022-06-17
fix adds patch for console new dep prisma-encrypt types to call prisma-admin/client

agustif:
Required for https://github.com/calcom/console/pull/15


agustif:
Found a better way https://github.com/47ng/prisma-field-encryption/issues/18#issuecomment-1122850798 closing this!

PR #3103, opened 2022-06-18
fixed month buttons on booking page for darkmode

PeerRich:
before:

image

after:

default

image

hover:

image

disabled:

image


emrysal:
Excellent! this was unspotted for a long time 🤯

PR #3112, opened 2022-06-20
includedDates would always disable excludedDates due to empty array

emrysal:
Fixes future usecases using excludedDates


sean-brydon (approved):
LGTM - Made this change locally to fix and working as expected

PR #3116, opened 2022-06-20
Hotfix/fix legacy user eventtypes

emrysal:
Adds all pre _user_eventtype userIds into the users

@zomars Please take as much caution as you need :)


zomars (approved):
approved GIF

PR #3118, opened 2022-06-20
AvailabilityPage improvements

emrysal:
@zomars EDIT:

- Now you can browse different dates without losing the currently selected date
- Enables local slots cache so they don't get lost when switching between months
- Fixes date not switching to New Years after current
- Allows permalinking a pre-selected date and a browsing date. They can be different dates, event different months.

@emrysal EDIT:

- Browsing date changed to "month=" and allows identical features as mentioned before about browsing date.
- AvailableTimes skeleton added - visible during initial page load.
- DatePicker refactored to Dayjs for tz awareness, making sure the dayjs instance does not lose track of the selected tz.
- Day shifting between timezones now functional
- Performance increases due to decreased logic within checkForAvailability - less logic to be ran per slot means speed increase.


zomars:
Made some improvements @emrysal please let me know if there's something missing to merge

zomars (approved):

agustif (dismissed):
Hey, so over the surface this looks good (and faster!)
I just hit this bug when trying too far into the future (2023) where it won't show dates, thoughts @zomars @emrysal sharing loom
Basically when it tries to change month to jan 2023 it goes back to 2022 june
https://www.loom.com/share/1aed432af2be4b3095746fcfada60cbd

zomars:
That's a nice catch @agustif it will definitely bite us later on. We should check on that.

zomars:
Made a quick demo of what I've made here:

https://www.loom.com/share/3a4a5b99334d4f13ba134ee867239052

emrysal:
Quick demo of the updates within this PR:
https://www.loom.com/share/f56dea8d444d4a538f0e1d0e2bb81153

zomars (dismissed):
Looking gucci 🙌

zomars:
https://user-images.githubusercontent.com/3504472/175793690-f9690e13-f1d6-45a5-8a6f-4cc4e4a8e14b.MOV

It seems like sometimes it just keeps loading. Testing on mobile Safari. Will check tomorrow

agustif:
Looking awesome now!

leog (approved):
Looking really good! Can't wait to have this deployed for our users.
The skeleton animation on dark mode is very subtle, maybe we can make it more flashy? 
I wonder if we should have some kind of polyfill to be covered with common issues we have with Safari Mobile.

Jaibles:
@emrysal For the skeleton can we do the effect on where the text would be rather than where the border would be? Let me know if you need designs for this.

emrysal:
@Jaibles @sean-brydon Designs are super helpful here, sounds like it may be a general change to the way skeletons operate so maybe a change to the design system.

packages/ui/booker/DatePicker.tsx
```diff
@@ -39,11 +38,11 @@ export const Day = ({
   date,
   active,
   ...props
-}: JSX.IntrinsicElements["button"] & { active: boolean; date: Date }) => {
+}: JSX.IntrinsicElements["button"] & { active: boolean; date: Dayjs }) => {
   return (
     <button
       className={classNames(
-        "hover:border-brand disabled:text-bookinglighter absolute top-0 left-0 right-0 bottom-0 mx-auto rounded-sm border border-transparent text-center font-medium disabled:cursor-default disabled:border-transparent disabled:font-light dark:hover:border-white disabled:dark:border-transparent",
+        "hover:border-brand disabled:text-bookinglighter absolute top-0 left-0 right-0 bottom-0 mx-auto w-full rounded-sm border border-transparent text-center font-medium disabled:cursor-default disabled:border-transparent disabled:font-light dark:hover:border-white disabled:dark:border-transparent",
```
alannnc, line 45: @emrysal This is what fixes the shrinking on main

PR #3120, opened 2022-06-20
removed hardcoded url for non-cal.com

PeerRich:
on cal.com:
image

not on cal.com:
image

### DIFF
--- a/apps/web/components/Shell.tsx
+++ b/apps/web/components/Shell.tsx
@@ -497,10 +497,14 @@ function UserDropdown({ small }: { small?: boolean }) {
             <span className="flex flex-grow items-center truncate">
               <span className="flex-grow truncate text-sm">
                 <span className="block truncate font-medium text-gray-900">
-                  {user?.username || "Nameless User"}
+                  {user?.name || "Nameless User"}
                 </span>
                 <span className="block truncate font-normal text-neutral-500">
-                  {user?.username ? `cal.com/${user.username}` : "No public page"}
+                  {user?.username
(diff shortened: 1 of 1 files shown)


apps/web/components/Shell.tsx
```diff
@@ -497,10 +497,14 @@ function UserDropdown({ small }: { small?: boolean }) {
             <span className="flex flex-grow items-center truncate">
               <span className="flex-grow truncate text-sm">
                 <span className="block truncate font-medium text-gray-900">
-                  {user?.username || "Nameless User"}
+                  {user?.name || "Nameless User"}
                 </span>
                 <span className="block truncate font-normal text-neutral-500">
-                  {user?.username ? `cal.com/${user.username}` : "No public page"}
+                  {user?.username
+                    ? process.env.NEXT_PUBLIC_WEBSITE_URL === "https://cal.com"
```
PeerRich, line 504: we should probably turn this into a `isCalcom` helper function

PeerRich: since it's being used in a couple places

zomars: We do have one

zomars: I just don't remember where

agustif: We've this on the 404 page, not being exported though
`const isCalcom = process.env.NEXT_PUBLIC_WEBAPP_URL === "https://app.cal.com";` could be moved to a shared package or lib or so later

zomars (approved):

PR #3121, opened 2022-06-21
HotFix - Background not being transparent for [user]/[type] in embed

hariombalhara:
## What does this PR do?

Fixes multiple issues related to booking flow

### Embed background on [user]/[type] page isn't transparent. Reported by @PeerRich on Twist DM for embed at https://w7th.com/work-with/ 
After:
Screenshot 2022-06-21 at 1 01 51 PM
Before:
Screenshot 2022-06-21 at 1 02 10 PM
_Note: - The issue isn't replicable on local because in dev mode, getStaticProps is always executed(and thus behave like SSR)_

### Border Fix
After: https://www.loom.com/share/3ecc903ec0e34bf9ba2fa02570719c4e
Before: https://www.loom.com/share/51291a10b1e044bca8342480f2b0f5e0

### Auto opening of Availability with ?date= parameter. Also shows that hydration error logs are gone now. These logs were visible in console in production but in UI in development mode.
After: https://www.loom.com/share/242f7c5f68fd4a69b852645128f53071
Before: https://www.loom.com/share/82ddc2a99de54c56bdcaf874b7ee38d9

**Environment**:  Production

## Type of change

## How should this be tested?

- See the looms

## Checklist

### DIFF
--- a/apps/web/pages/[user]/[type].tsx
+++ b/apps/web/pages/[user]/[type].tsx
@@ -1,6 +1,8 @@
 import { UserPlan } from "@prisma/client";
 import dayjs from "dayjs";
 import { GetStaticPropsContext } from "next";
+import { useRouter } from "next/router";
+import { useEffect } from "react";
 import { JSONObject } from "superjson/dist/types";
 import { z } from "zod";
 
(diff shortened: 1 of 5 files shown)


hariombalhara:
A web performant way to handle embed(as styling is different for embed/non-embed) for getStaticProps(with fallback) generated pages might be to introduce a route param `embed` to [user]/[type] page i.e. make it [embed]/[user]/[type] where embed="embed" if it is an embed case otherwise
This would allow the embed value to be available at server and thus embed classed can be applied to Served HTML itself.
It would also cause embed to be a caching parameter and thus with embed and without embed you would get different cached responses.

apps/web/pages/[user]/[type].tsx
```diff
@@ -19,7 +22,14 @@ export type AvailabilityPageProps = inferSSRProps<typeof getStaticProps>;
 
 export default function Type(props: AvailabilityPageProps) {
   const { t } = useLocale();
-
+  const isEmbed = useIsEmbed();
+  useEffect(() => {
+    // Embed background is handled in _document.tsx but this particular page(/[user][/type] is statically rendered and thus doesn't have `embed` param at that time)
+    // So, for static pages, handle the embed background here. Make sure to always keep it consistent with _document.tsx
+    if (isEmbed) {
+      document.body.style.background = "transparent";
```
Comment (PeerRich) [inline apps/web/pages/[user]/[type].tsx:30]: NIT: we could also append a bg-transparent class to stay in tailwind world, but this works too

PeerRich (approved):

PR #3124, opened 2022-06-21
Update Prisma to v3.15.2 to support data proxy

pumfleet:


hariombalhara:
Cal deployment failed due to prisma version update probably.

zomars:
It should run post-install yes

zomars (approved):

PR #3129, opened 2022-06-21
Fixes redirect to upcoming bookings after canceling booking

CarinaWolli:
## What does this PR do?

Fixes bug, that it always redirect to bookings/recurring after canceling a booking.

## Type of change

## How should this be tested?

- Cancel a not recurring event and see that you will be redirected to bookings/upcoming after clicking 'Back to bookings'
- Cancel a recurring event and see that you will be redirected to bookings/recurring after clicking 'Back to bookings'

## Checklist


leog (approved):
Good catch!

PR #3131, opened 2022-06-21
The time format does not work as expected

### LINKED ISSUE #2557: The time format does not work as expected.

The time format does not work as expected. In my settings I configured 12 hours and when I send the link of my availability the hours appear in 24-hour format.

I trying to find in the setting of the event but there is not anyting to configure 12 hours.

### Technical details

Firefox

gitstart:
## What does this PR do?

Fixes [#2557](https://github.com/calcom/cal.com/issues/2557)

 Loom Video: [https://www.loom.com/share/e2b4ad12406d4cce9dcc52981ae6a517](https://www.loom.com/share/e2b4ad12406d4cce9dcc52981ae6a517)

## Type of change

## How should this be tested?

- [x] Change time format preferences in settings
- [x] Schedule a new event

### DIFF
--- a/apps/web/pages/settings/profile.tsx
+++ b/apps/web/pages/settings/profile.tsx
@@ -21,6 +21,7 @@ import { Icon } from "@calcom/ui/Icon";
 import { withQuery } from "@lib/QueryCell";
 import { asStringOrNull, asStringOrUndefined } from "@lib/asStringOrNull";
 import { getSession } from "@lib/auth";
+import { is24h, timeZone } from "@lib/clock";
 import { nameOfDay } from "@lib/core/i18n/weekday";
 import { isBrandingHidden } from "@lib/isBrandingHidden";
 import { inferSSRProps } from "@lib/types/inferSSRProps";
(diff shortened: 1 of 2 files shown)


apps/web/pages/settings/profile.tsx
```diff
@@ -156,6 +158,11 @@ function SettingsView(props: ComponentProps<typeof Settings> & { localeProp: str
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  function setTimePreferences(selectedTimeFormat: number) {
+    is24h(selectedTimeFormat === 24);
+    timeZone(dayjs.tz.guess());
```
emrysal, line 163: So this is setting the localStorage of the organizer, I don't think this is a bad idea but the profile page has a timeZone setting; if you can take that into account that'd be excellent.

emrysal (changes requested):

apps/web/pages/settings/profile.tsx
gitstart: Hi @emrysal, can you kindly take a look again?

joeauyeung:
Fixed in #3441 with user profile settings taken into account.

PR #3132, opened 2022-06-21
When no location is set in an event, where would the meeting occur?

### LINKED ISSUE #2900: When no location is set in an event, where would the meeting occur ?

### Issue Summary
When no location is selected in an event, the email received after booking has a `Where` heading but no text. Maybe we should make it very clear that no location was selected by the organizer and the booker should contact the organizer for the meeting location.

Also, maybe we should not allow empty locations at all. If the organizer chooses to set no location, he should be asked to drop a comment about where the meeting is supposed to happen. 

This is to solve the confusion for a booker when such an event is booked.

Screenshot 2022-05-27 at 11 10 55 AM

### Steps to Reproduce

1. Create an event. By default it has Cal Video as location.
2. Go and remove that location.
3. Book a meeting
4. Check the email recieved.

### Technical details

- Node.js version

gitstart:
## What does this PR do?
- Adds CAL-video as the default location when no location is set for an event type

Fixes #2900

Loom video: https://www.loom.com/share/7ffd8948682541c4a2075087fded4982

## Screenshots
Screenshot before making changes:
Screenshot from 2022-06-20 13-26-36

Screenshot after making changes:
Screenshot from 2022-06-20 13-21-41

## Type of change


pumfleet (approved):
Looks simple enough to me

zomars (changes requested):

PR #3133, opened 2022-06-21
Negative days in trial license period

### LINKED ISSUE #2955: Negative days in trial license period

### Issue Summary

On event-types page:
Screenshot from 2022-06-03 08-40-01

### Steps to Reproduce
1. When you 14 days off free license expires, in the event-types page, the message shows a negative day

gitstart:
## What does this PR do?

Fixes #2955

 Bug fixed: Screenshot from 2022-06-03 10-23-07

## Type of change

## How should this be tested?

This case will happen if the cron job for moving the user to the free plan didn't happen yet.
But we can virtually reproduce it by going to constants and change TRIAL_LIMIT_DAYS so that the days are "negative".


PeerRich:
we need to rethink what to show there for self-hosters. on cal.com it wouldn't show negative days, its only on self hosters

apps/web/public/static/locales/en/common.json
```diff
@@ -902,6 +902,7 @@
   "meeting_url_in_conformation_email": "Meeting url is in the confirmation email",
   "url_start_with_https": "URL needs to start with http:// or https://",
   "number_provided": "Phone number will be provided",
+  "trial_expired": "Your license has expired",
```
zomars, line 905:
```suggestion
  "trial_expired": "Your trial has expired",
```

zomars (approved):

zomars:
Good enough for now

PR #3144, opened 2022-06-23
Fixes over Turborepo improvements

hariombalhara:
## What does this PR do?

Fixes following issues
- Fixes the issue where *.generated files weren't being created
- Fixes the issue where `yarn dx` was crashing because of use of `prisma migrate dev` as it's dependency which is an interactive command. Changed that back to prisma migrate deploy
- Fixes embed tests by ensuring seeds run before tests.
- Fixes type checks for embed-react and @calcom/web by ensuring their dependencies which are typescript are built first.

## Type of change

## How should this be tested?

- yarn dx from root
- yarn dev and yarn dx from apps/web


zomars (changes requested):
Thanks for this @hariombalhara, I want to take advantage of this PR to make some observations:
- Scripts functions should be as pure as possible (`yarn dev` generating schemas and App Store is not pure, `dev` should only be concerned on running `dev` and nothing else)
- For scripts that depend on other we should come up with reasonable pipelines
- Let's remove app-store-cli generation from every project
- Let's commit the generated app-store files (they won't change that much so it doesn't make sense to generate them EVERY time)

If we need to update the generated files we can just trigger it manually and commit again.

hariombalhara:
Thanks @zomars Will check and update accordingly. 

I am mainly concerned about app-store Dev Experience. Auto generated files makes you fall into pit of success. Committing then after manual generation doesn't sound good DX. It might also create a scenario where generated file is outdated in git. 

But your point is valid that it isn't that frequently regenerated, so an approach might be to commit the files but ensure that whenever committing, git hooks would ensure that the files are regenerated and if something changed it's committed.

zomars:
> I am mainly concerned about app-store Dev Experience. Auto generated files makes you fall into pit of success. Committing then after manual generation doesn't sound good DX. It might also create a scenario where generated file is outdated in git. 

Agreed but I'm this case the burden for the benefit is not that great. Also we had some DX complaints that the generated files were sometimes missing. We rather have outdated files that missing files.

hariombalhara:
@zomars Removes app-store files generation from the overall picture.. Generated files are now committed and ensured to be up to date using pre-commit hook.

zomars (approved):
Awesome work @hariombalhara one more thing. We should
- add the generated files to prettierignore or...
- make the generator script run prettier on it after creation

Whichever is fine

hariombalhara:
@zomars Added to ignore.

PR #3145, opened 2022-06-23
Saves app id to zoom credential

CarinaWolli:
## What does this PR do?

AppId wasn't saved to zoom credential. This fixes the issue that when we delete zoom it doesn't change locations to Cal Video

## Type of change

## Checklist

### DIFF
--- a/packages/app-store/zoomvideo/api/callback.ts
+++ b/packages/app-store/zoomvideo/api/callback.ts
@@ -35,6 +35,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         create: {
           type: "zoom_video",
           key: responseBody,
+          appId: "zoom",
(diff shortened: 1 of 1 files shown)


packages/app-store/zoomvideo/api/callback.ts
```diff
@@ -35,6 +35,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         create: {
           type: "zoom_video",
           key: responseBody,
+          appId: "zoom",
```
hariombalhara, line 38: Looks correct. But it would only fix credentials created going forward. We should run query to update existing credentials as well.

hariombalhara (approved):

zomars (approved):

PR #3147, opened 2022-06-23
scroll to times after clicking a day, two column layout mobile

### LINKED ISSUE #451: Scroll to Time After Selecting Date on Mobile

Users are confused on mobile when tapping a date because it is not obvious when doing so that the times appear under the calendar on mobile.

Scroll to the top of the time area when date is clicked, maybe print the selected date above the available times.

Make the times appear in a new screen all together that replaces the calendar screen.

PeerRich:
## What does this PR do?

- scroll to time on mobile after selecting a day
- add a two grid column for mobile

closes #451 

https://www.loom.com/share/20afe28d82214dc6968df72f2ff47577

## Type of change

## How should this be tested?

- [ ] go to mobile /pro
- [ ] test on a phone

### DIFF
--- a/apps/web/components/booking/AvailableTimes.tsx
+++ b/apps/web/components/booking/AvailableTimes.tsx
@@ -57,7 +57,7 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
           </span>
         </span>
       </div>
-      <div className="flex-grow overflow-y-auto md:h-[364px]">
+      <div className="grid flex-grow grid-cols-2 gap-x-2 overflow-y-auto sm:block md:h-[364px]">
         {slots?.length > 0 &&
           slots.map((slot) => {
             type BookingURL = {
--- a/packages/ui/booker/DatePicker.tsx
+++ b/packages/ui/booker/DatePicker.tsx
@@ -53,7 +53,9 @@ export const Day = ({
       data-disabled={props.disabled}
       {...props}>
       {date.getDate()}
-      {dayjs(date).isToday() && <span className=" absolute left-0 bottom-1 mx-auto w-full text-4xl">.</span>}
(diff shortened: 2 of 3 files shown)


apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -57,7 +57,7 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
           </span>
         </span>
       </div>
-      <div className="flex-grow overflow-y-auto md:h-[364px]">
+      <div className="grid flex-grow grid-cols-2 gap-x-2 overflow-y-auto sm:block md:h-[364px]">
```
PeerRich, line 60: two grid column for mobile

packages/ui/booker/DatePicker.tsx
```diff
@@ -94,7 +96,13 @@ const Days = ({
           ) : (
             <DayComponent
               date={day}
-              onClick={() => props.onChange(day)}
+              onClick={() => {
+                props.onChange(day);
+                window.scrollTo({
+                  top: 360,
+                  behavior: "smooth",
```
PeerRich, line 103: we could do something like "isMobile" to only scroll on mobile, however given the desktop view has more than 360px height it doesn't scroll.

it still runs "scrollTo" as an invisible operation but it's a very light JS API so I doubt we need to check if it's mobile or not

PeerRich: @emrysal NIT: worth extracting into a named function?

emrysal: For a single use this is fine, keep in mind window isn't always available - but it looks OK in this situation

emrysal (approved):
LGTM

PR #3149, opened 2022-06-23
feat(web): case insensitive slug search

heuels:
Not sure if this is something you want, but after upgrading we have lots of differently-cased event types that we still need to support 😞

### DIFF
--- a/apps/web/pages/[user]/[type].tsx
+++ b/apps/web/pages/[user]/[type].tsx
@@ -105,7 +105,7 @@ async function getUserPageProps(context: GetStaticPropsContext) {
   const eventTypeIds = user.eventTypes.map((e) => e.id);
   const eventTypes = await prisma.eventType.findMany({
     where: {
-      slug,
+      slug: { equals: slug, mode: "insensitive" },
       /* Free users can only display their first eventType */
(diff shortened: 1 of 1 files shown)


agustif:
I'd go as far to say it might be good to normalize always slugs to be lowercase and use only dash as separator?
I know we recently had to cleanup some chars in event-slugs too...
Thoughts @zomars ?

heuels:
@zomars yes, that's right!

zomars:
@heuels @agustif I see this like something desirable yes. But we should make sure it is being used consistently in the codebase. Or find a way to enable this globally:

image

agustif:
@zomars we could add maybe a prisma middleware for this, that extends event-type calls adding the `      slug: { equals: slug, mode: "insensitive" },` where prop?

hariombalhara:
@heuels Were you done with your changes?

heuels:
@hariombalhara I have not sadly. We deprioritized using Cal.com for now and this has fallen off my radar.

emrysal:
This needs to be fixed in the middleware to catch the many, many different places we query things by slug. - further improvement but this will definitely stay on our radar for implementation. Thanks @heuels for the PR and ensuing discussions too.

PR #3153, opened 2022-06-24
fix/stripe-customer-sync

alannnc:
## What does this PR do?

Takes care of keeping in sync user data with stripe customer data

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Change email or username on settings page then look up on stripe for customer's metadata and they should be the same as our DB.

## Checklist


apps/web/server/routers/viewer.tsx
```diff
@@ -707,12 +709,30 @@ const loggedInViewerRouter = createProtectedRouter()
...
+        select: {
+          id: true,
+          username: true,
+          email: true,
+          metadata: true,
+        },
       });
+
+      // Notify stripe about the change
+      if (updatedUser && updatedUser.metadata && hasKeyInMetadata(updatedUser, "stripeCustomerId")) {
```
alannnc, line 726: @zomars Should I add the same flag to know if we are on self hosted? And don't execute any stripe code.

zomars: Some self-hosted users will use Stripe Connect, If this is only for premium usernames then we can skip it

zomars (approved):

apps/web/server/routers/viewer.tsx
alannnc: Let me get this right, connect can be for self hosted users, but this it's only involves our user plan handling so I should add flag here for `IS_SELF_HOSTED`.

zomars: Nevermind seems like is good to go 👍🏽

PR #3168, opened 2022-06-27
Fix saving price and currency

alannnc:
## What does this PR do?

Fixes price not being saved on event type.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] With stripe credentials, try to save price and currency to DB.

## Checklist

### DIFF
--- a/apps/web/server/routers/viewer/eventTypes.tsx
+++ b/apps/web/server/routers/viewer/eventTypes.tsx
@@ -266,7 +266,6 @@ export const eventTypesRouter = createProtectedRouter()
         users,
         id,
         hashedLink,
-        price,
         ...rest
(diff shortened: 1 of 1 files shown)


apps/web/server/routers/viewer/eventTypes.tsx
```diff
@@ -266,7 +266,6 @@ export const eventTypesRouter = createProtectedRouter()
         users,
         id,
         hashedLink,
-        price,
```
alannnc, line 269: This was added previously, so it was being omitted on the data object to update.

leog (approved):
Nicely done

zomars (approved):

PR #3182, opened 2022-06-28
Hotfix: Recurring config missing from event booking

leog:
## What does this PR do?

Recurring config was missing when booking an event

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

PR #3183, opened 2022-06-28
Recurring event config missing

leog:
See #3182.


hariombalhara:
It has 0 changes now. I guess the changes are already merged in main in other ways. We can close it.

PR #3195, opened 2022-06-29
Fix hubspot callback

hariombalhara:
## What does this PR do?

Fixes # (issue)
Hubspot installation fails at last step in the redirection.

**Environment**: Production

## Type of change

## How should this be tested?

- Simply install hubspot app. It should be successfully installed.

## Checklist


leog (approved):
Works perfect, thanks @hariombalhara 
image

PR #3197, opened 2022-06-29
Attach credential id to destinationCalendar records

joeauyeung:
## What does this PR do?

This PR address a bug where if a user had multiple Google Cal accounts connect to their account bookings would be created on all accounts.

Before when creating calendar events it would grab all the user's calendar credentials and only create calendar events on credentials that would match the destination calendar. This is why the bug only appeared with two Google Cals and not one Google Cal and an Outlook credential.

This PR adds the credential ID to the destination calendar record so when creating bookings, rather than looping through all credentials from the user. This should also increase booking performance.

Fixes # (issue)

## Type of change

## How should this be tested?

- Connect two Google Calendar accounts
- Create a booking
- When looking at the calendars of both accounts, the event should have only been created on the destination calendar

## Checklist


packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -177,6 +177,7 @@ export default class Office365CalendarService implements Calendar {
               integration: this.integrationName,
               name: cal.name ?? "No calendar name",
               primary: cal.isDefaultCalendar ?? false,
+              readOnly: !cal.canEdit && true,
```
joeauyeung, line 180: Outlook only has one field for if the user can write events to calendar

packages/core/EventManager.ts
```diff
@@ -282,6 +282,23 @@ export default class EventManager {
...
+        const credential = await prisma.credential.findFirst({
+          where: {
+            id: event.destinationCalendar.credentialId,
+          },
+        });
+
+        if (credential) {
+          return [await createEvent(credential, event)];
+        }
+      }
```
joeauyeung, line 300: When create events check if there is a credential id first. If not then continue with old logic

sean-brydon (approved):
LGTM - tested locally for google calendar and works perfectly. I don't have an outlook account but the code looks good for that too :) 

Nice job

PR #3201, opened 2022-06-29
Fixes the slow performance due to the Dayjs timezone plugin

emrysal:
## What does this PR do?

The `dayjs.tz()` function returns an instance with abysmal performance on the `isBetween` and `isBefore` (and `isAfter`) calls. We do however need this elsewhere in the code to properly run `startOf` - this just modifies the code that has the performance hit (`isBetween` takes ~4ms to execute, x the amount of slots, x the busy times). By running this within utc mode the speed of `isBetween` (& friends) goes to 0.00Xms.

Performance can be improved elsewhere too - but this is the primary offender.

### DIFF
--- a/apps/web/server/routers/viewer/slots.tsx
+++ b/apps/web/server/routers/viewer/slots.tsx
@@ -62,9 +62,10 @@ const checkForAvailability = ({
   const slotStartTimeWithBeforeBuffer = time.subtract(beforeBufferTime, "minutes").utc();
   const slotEndTimeWithAfterBuffer = time.add(eventLength + afterBufferTime, "minutes").utc();
 
-  return busy.every((busyTime): boolean => {
-    const startTime = dayjs(busyTime.start);
-    const endTime = dayjs(busyTime.end);
+  return busy.every((busyTime) => {
+    const startTime = dayjs.utc(busyTime.start);
+    const endTime = dayjs.utc(busyTime.end);
+
     // Check if start times are the same
     if (time.utc().isBetween(startTime, endTime, null, "[)")) {
       return false;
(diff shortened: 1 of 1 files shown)


apps/web/server/routers/viewer/slots.tsx
```diff
@@ -59,15 +58,16 @@ const checkForAvailability = ({
...
   }
 
-  const slotEndTime = time.add(eventLength, "minutes");
-  const slotStartTimeWithBeforeBuffer = time.subtract(beforeBufferTime, "minutes");
-  const slotEndTimeWithAfterBuffer = time.add(eventLength + afterBufferTime, "minutes");
+  const slotEndTime = time.add(eventLength, "minutes").utc();
+  const slotStartTimeWithBeforeBuffer = time.subtract(beforeBufferTime, "minutes").utc();
+  const slotEndTimeWithAfterBuffer = time.add(eventLength + afterBufferTime, "minutes").utc();
+
+  return busy.every((busyTime) => {
```
emrysal, line 65: Typescript knows.

apps/web/server/routers/viewer/slots.tsx
```diff
@@ -59,15 +58,16 @@ const checkForAvailability = ({
...
+    const startTime = dayjs.utc(busyTime.start);
+    const endTime = dayjs.utc(busyTime.end);
```
emrysal, line 67: For servers w/ non-UTC timezones, if you call as dayjs() the busyTime.start and busyTime.end are translated to local time. Parsing as UTC parses it as UTC which should work better on those servers.

zomars (approved):
Fingers crossed 🤞

Issue #3177, opened 2022-06-28
Test Required: Add an App installation testcase for HubSpot App


hariombalhara:
Hubspot being a special case with lying in both `calendar` and `other` category it must have an installation testcase and maybe others.
Refer to https://github.com/calcom/cal.com/pull/3176 for a regression that occurred due to it

Issue #2255, opened 2022-03-23
RFC: Expand Booking E2E tests


zomars:
Since there's a ton of possible event type configurations we should expand our testing with some of the most common use cases, these are the ones that come to my mind:

- Can book an event with Zoom Video as location
- Can book an event with Google Meet as location
- Can book an event with Daily.co Video as location
- Can book an event with Jitsi Meet as location
- Can book an event with Huddle01 Video as location
- Can book an event with Link Meeting as location
- Can book an event with In Person as location
- Can book an event with Phone as location
- Can book an event with Additional Fields
- Can book an event with Web3 enabled (This is a tricky one with playwright)

Expanding testing for these will allow us to make changes with peace of mind that none of the critical paths are breaking. I can tell from experience with App Store. 

Knowing this, there are more questions on how to achieve this.

- How can we save/reuse/mock testing credentials for each of the needed integrations?
- Can we mock it enough to have confidence in the booking not breaking?

Will expand as soon as I have more information....

zomars:
@hariombalhara @denik1981 I would love to hear your input on this 🙏🏽

PeerRich:
great RFC

hariombalhara:
I have added zoom tests in PR #2202 

It right now just tests that user is able to install zoom app. We can very easily expand its to Event creation. 

The philosophy I am using here is that third party code shouldn't be tested in core. We should just mock the API as per their documentation and use those mocks.(Core Tests that run on every PR) 
Benefits:
- You can create any type of third party API scenario. With actual credentials it would almost be impossible. 
- You can test with any version of API(Alpha, Beta, Deprecated, Stable) 
- Tests would be faster to execute. 
- Flaky Test won't be there. 
- You understand the documentation well and are able to quickly add cose for all scenarios. 

But we must test atleast one happy path with each third party integration. (Extended Tests that run once or at scheduled time every day on main)

zomars:
Yeah, we already use Checkly to run playwright on staging/production. We could setup those tests in there and use the example account with all integrations enabled.

denik1981:
Thanks for taking me into consideration. 
Sure, I will take a look and I will be happy to contribute to this.

> Expanding testing for these will allow us to make changes with peace of mind that none of the critical paths are breaking. I can tell from experience with App Store.

[...]

PR #3210, opened 2022-07-01
Zoom/Hubspot Tests with MSW mocking of requests initiated from Next.js server

hariombalhara:
Fixes 
https://github.com/calcom/cal.com/issues/3177 
https://github.com/calcom/cal.com/issues/2255
## Tests that verify
- Zoom 
	- Installation/Uninstallation. 
	- Add zoom to an event-type and ensures that Zoom meeting is created
	- 2 Emails are sent with zoom links
- Hubspot
	- 	Installation/Uninstallation

### Approach
Tests with MSW require that playwright workers launch their own Next.js instance, so that within the tests we can mock requests based on the testcase

### Known Limitations -
- More than 1 workers don't currently work because our codebase doesn't support building once and then running with different ports. We need to use the port of the app itself when it was launched.
- Due to the above limitation I am running these tests in separate workflow.yml so that they don't slow down other tests.


.env.example
```diff
@@ -77,6 +77,14 @@ NEXT_PUBLIC_HELPSCOUT_KEY=
...
 SEND_FEEDBACK_EMAIL=
 
+# ZOOM
+ZOOM_CLIENT_ID="ZOOM_CLIENT_ID"
+ZOOM_CLIENT_SECRET="ZOOM_CLIENT_SECRET"
+
+# HUBSPOT
+HUBSPOT_CLIENT_ID="HUBSPOT_CLIENT_ID"
+HUBSPOT_CLIENT_SECRET="HUBSPOT_CLIENT_SECRET"
+
```
zomars, line 87: AFAIK these should be on `.env.appStore`

zomars (commented):
Code looking Gucci. Testing locally

zomars:
Solved conflitcts and updated this PR @hariombalhara do you think this is mergable?

hariombalhara:
@zomars yeah. It had forgotten to remove "Self review Pending" from description

PeerRich:
are these only tests? should be safe to merge then. just noticed some merge conflicts @hariombalhara

hariombalhara:
Yeah these are just tests. Will resolve the conflicts.

hariombalhara:
Conflicts resolved.

hariombalhara:
@zomars @zlwaterfield Please review this. Should be quick one as it affects only tests and they are passing.

zomars (approved):

PR #3218, opened 2022-07-02
Fixes "Time slot initially unavailable becomes available by browsing?"

emrysal:
* Ensures that a startTime is given in UTC and then properly translated to invitee timezone time regardless of server time:
e.g. `2022-07-06T04:00:00.000Z 2022-07-06T04:00:00-04:00`

* Cache key got hit on calls for different times due missing `dateTo` and `dateFrom` - these are both strings at this point so `format()` is not needed.


zomars (approved):
Awesome 🙏

PR #3221, opened 2022-07-04
Booking Module Improvements

hariombalhara:
## What does this PR do?

Fixes # (issue)
- Fix momentary render of dates before skeleton loader.([Before Loom](https://www.loom.com/share/ee4d4e3c38c24934bf9930d094620bb6) , [After Loom](https://www.loom.com/share/a79a6d62af9f4a04ad80dac076277091))
- Fix the Hydration issue - We shouldn't guess timezone on server as it would be different from client causing hydration error. Errors are visible in this loom which customer shared in an unrelated issue https://www.loom.com/share/626b5be3253244f8aa0b2bdd7acf808b
- Fixes the wrong trpcQuery result when clicking on a particular date. Also, fixes last day of the month being never available.
[Before](https://www.loom.com/share/b92d6d90216d4f47946f39e6c9ca7c44), [After](https://www.loom.com/share/dccf4176e1a54b7da2f5244805883c3f)

## Type of change

## Checklist

### DIFF
--- a/apps/web/server/routers/viewer/slots.tsx
+++ b/apps/web/server/routers/viewer/slots.tsx
@@ -151,16 +151,17 @@ export const slotsRouter = createRouter().query("getSchedule", {
     const startTime =
       input.timeZone === "Etc/GMT"
         ? dayjs.utc(input.startTime)
-        : dayjs.utc(input.startTime).tz(input.timeZone, true);
+        : // The time is already sent in UTC, so mark it UTC by passing true flag and without changing the time. After that convert into the timezone
+          dayjs(input.startTime).utc(true).tz(input.timeZone);
     const endTime =
       input.timeZone === "Etc/GMT"
         ? dayjs.utc(input.endTime)
-        : dayjs.utc(input.endTime).tz(input.timeZone, true);
+        : // The time is already sent in UTC, so mark it UTC by passing true flag and without changing the time. After that convert into the timezone
+          dayjs(input.endTime).utc(true).tz(input.timeZone);
 
     if (!startTime.isValid() || !endTime.isValid()) {
       throw new TRPCError({ message: "Invalid time range given.", code: "BAD_REQUEST" });
     }
-
     let currentSeats: CurrentSeats | undefined = undefined;
 
     const userSchedules = await Promise.all(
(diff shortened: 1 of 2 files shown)


apps/web/server/routers/viewer/slots.tsx
```diff
@@ -151,16 +151,17 @@ export const slotsRouter = createRouter().query("getSchedule", {
...
         ? dayjs.utc(input.startTime)
-        : dayjs.utc(input.startTime).tz(input.timeZone, true);
+        : // The time is already sent in UTC, so mark it UTC by passing true flag and without changing the time. After that convert into the timezone
+          dayjs(input.startTime).utc(true).tz(input.timeZone);
     const endTime =
       input.timeZone === "Etc/GMT"
         ? dayjs.utc(input.endTime)
-        : dayjs.utc(input.endTime).tz(input.timeZone, true);
+        : // The time is already sent in UTC, so mark it UTC by passing true flag and without changing the time. After that convert into the timezone
+          dayjs(input.endTime).utc(true).tz(input.timeZone);
```
emrysal, line 160: The issue with this line as far as I'm aware is that `dayjs(input.endTime)` will immediately cast to local time, so if you give a UTC date it will auto-convert to system time. utc(true) will then correctly go into UTC mode but the time conversion has already happened at this point. Not an issue if the system is running in +00:00 but it will be within other timezones.

zomars (commented):
I'll defer this review to @emrysal since he's the timezone wizard here

apps/web/server/routers/viewer/slots.tsx
emrysal, line 160:
```suggestion
          dayjs(input.endTime).utc().tz(input.timeZone);
```

Because it's hard to explain why that's needed; see the following, the problem occurs when system time != UTC:
image

emrysal (changes requested):

emrysal (approved):
Good to go 🚀 , nice work

apps/web/server/routers/viewer/slots.tsx
zomars: @emrysal should we open this issue in the days repo?

PR #3233, opened 2022-07-05
Add getSchedule tests

hariombalhara:
Adds integration tests for getSchedule.
Define scenarios directly in DB, External Calendar and expect slots
Screenshot 2022-07-20 at 7 01 42 PM
.


packages/lib/availability.ts
```diff
@@ -65,6 +65,7 @@ export function getWorkingHours(
   },
   availability: { days: number[]; startTime: ConfigType; endTime: ConfigType }[]
 ) {
+  // console.log("Availability in getWorkingHours", availability);
```
zomars, line 68:
```suggestion
```

packages/lib/availability.ts
```diff
@@ -87,10 +88,11 @@ export function getWorkingHours(
       utcOffset;
     const endTime =
       dayjs.utc(schedule.endTime).get("hour") * 60 + dayjs.utc(schedule.endTime).get("minute") - utcOffset;
-
     // add to working hours, keeping startTime and endTimes between bounds (0-1439)
     const sameDayStartTime = Math.max(MINUTES_DAY_START, Math.min(MINUTES_DAY_END, startTime));
     const sameDayEndTime = Math.max(MINUTES_DAY_START, Math.min(MINUTES_DAY_END, endTime));
+    // console.log("SCHEDULE TEST", schedule, startTime, endTime, sameDayStartTime, sameDayEndTime);
+
```
zomars, line 95:
```suggestion
```

zomars (commented):
Looking good so far. Testing...

zomars (changes requested):
- Should we include a `.env.test.example`?
- If not, we should add to the README how to run the test locally.

hariombalhara:
Added README and .env.test.example.

package.json
```diff
@@ -42,7 +42,7 @@
     "test-playwright": "yarn playwright test --config=tests/config/playwright.config.ts",
     "embed-tests-quick": "turbo run embed-tests-quick",
     "embed-tests": "turbo run embed-tests",
-    "test-e2e": "turbo run test-e2e --scope=\"@calcom/web\" --concurrency=1",
+    "test-e2e": "turbo run test --scope=\"@calcom/web\" && yarn turbo run test-e2e --scope=\"@calcom/web\" --concurrency=1",
```
zomars, line 45: Do we want to run this together? Can we run them on parallel on CI instead?

zomars (approved):

PR #3238, opened 2022-07-05
Contains multiple fixes centered around Zoom

emrysal:
Fixes:

* parseRecurringDates parsed based on browser time, not organizer time or attendee time. 99% of the time correct, not always.
* Zoom start and end times MUST be given in UTC (<date>+01:00 is parsed as UTC) - we don't care; I fixed the issue by casting the overall event start and end time to UTC.
* Added organizer timeZone to user.me query to use later in parseRecurringDates on the booking page (note, not having that there before likely means that there is more depending on browser time than I fixed in this PR 🤷)

Features:
* Zoom supports recurring events; which I had a look at and mapped with.

### DIFF
--- a/packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
+++ b/packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
@@ -1,8 +1,10 @@
 import { Credential } from "@prisma/client";
 import { z } from "zod";
 
+import dayjs from "@calcom/dayjs";
 import { handleErrorsJson } from "@calcom/lib/errors";
 import prisma from "@calcom/prisma";
+import { Frequency } from "@calcom/prisma/zod-utils";
 import type { CalendarEvent } from "@calcom/types/Calendar";
 import type { PartialReference } from "@calcom/types/EventManager";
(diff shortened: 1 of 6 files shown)


packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -202,8 +204,69 @@ const zoomAuth = (credential: Credential) => {
...
+      switch (recurringEvent.freq) {
+        case Frequency.DAILY:
+          recurrence = {
+            type: 1,
+          };
+          break;
+        case Frequency.WEEKLY:
+          recurrence = {
+            type: 2,
+            weekly_days: dayjs(startTime).tz(attendees[0].timeZone).day() + 1,
```
emrysal, line 238: dayjs is 0-6 and zoom is 1-7 so we have to add one, luckily their start date is Sunday as is Dayjs so we're good there.

alannnc (approved):
image
Tested it and it worked on my zoom account

zomars (approved):
Looking gucci

PR #3243, opened 2022-07-06
Hotfix: Potential fix for Zoom token expiry

hariombalhara:
## What does this PR do?

Fixes # (issue)
While testing I was able to reach a state where I had 3 zoom connections and recently added one(which had the latest refreshToken) wasn't being used. The reason was that due to some reason credentials were returned in some weird order. This change ensures that latest credential would be returned for videos. After this change my issue was fixed.

**Environment**: Production

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


zomars (approved):
Thank you @hariombalhara Looking good!

alannnc:
Nice work, this should fix the problem partially. I'm gonna open a new issue, the real problem here it's that we shouldn’t save duplicated credentials on any video app.

PR #3252, opened 2022-07-06
Upgrading nextjs and collect

leog:
## What does this PR do?

Upgrades `next-collect` which in turn requires upgrading `next` to v12.2.0.


apps/web/pages/api/auth/saml/userinfo.ts
```diff
@@ -24,7 +24,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     // check for query param
     if (!token) {
       let arr: string[] = [];
-      arr = arr.concat(req.query.access_token);
+      arr = arr.concat(req.query.access_token!);
```
zomars, line 27: Let's avoid null assertion, instead validate `req.query`:

Whenever we are parsing unknown json where we know the shape the object should have but it is not typed from source, it is worth the extra effort to use zod so we get both type and runtime safety.

📄 [From Cal.com's Gitbook](https://app.gitbook.com/o/6snd8PyPYMhg0wUw6CeQ/s/VXRprBTuMlihk37NQgUU/engineering/best-practices/validate-using-zod-instead-of-type-casting)

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

apps/web/pages/api/user/[id].ts
```diff
@@ -12,7 +12,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   }
 
   const userIdQuery = req.query?.id ?? null;
-  const userId = Array.isArray(userIdQuery) ? parseInt(userIdQuery.pop() || "") : parseInt(userIdQuery);
+  const userId = Array.isArray(userIdQuery) ? parseInt(userIdQuery.pop() || "") : parseInt(userIdQuery!);
```
Comment (zomars) [inline apps/web/pages/api/user/[id].ts:15]: Avoid null assertions.

Usually this means we're not validating enough or a validation is incorrect

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix&nbsp; | &nbsp;**Tags:** Brittle

zomars (changes requested):
You beat me into doing this, thanks! Some points tho:
- Let's remove all null assertions
- Let's update all private packages as well to prevent mismatches


__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

apps/web/pages/api/app-store/[...static].ts
```diff
@@ -15,6 +15,10 @@ import path from "path";
  */
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
   const [appName, fileName] = Array.isArray(req.query.static) ? req.query.static : [req.query.static];
+
+  if (!fileName) throw Error("No file name provided");
+  if (!appName) throw Error("No app name provided");
```
Comment (zomars) [inline apps/web/pages/api/app-store/[...static].ts:20]: These will throw 500 errors, clogging our data dog alerts. Let's return a proper 400 "Bad Request" error.

apps/web/pages/api/user/[id].ts
```diff
@@ -12,7 +12,15 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   }
 
   const userIdQuery = req.query?.id ?? null;
-  const userId = Array.isArray(userIdQuery) ? parseInt(userIdQuery.pop() || "") : parseInt(userIdQuery);
+  const userId = Array.isArray(userIdQuery)
+    ? parseInt(userIdQuery.pop() || "")
+    : typeof userIdQuery === "string"
+    ? parseInt(userIdQuery)
+    : userIdQuery;
+
```
Comment (zomars) [inline apps/web/pages/api/user/[id].ts:20]: Let's use a zod [transform or refine](https://zod.dev/?id=transform) here instead.

apps/web/pages/api/user/[id].ts
```diff
@@ -11,12 +12,12 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
 
+  const stringToNumber = z.string().transform((val) => parseInt(val));
+
   const userIdQuery = req.query?.id ?? null;
   const userId = Array.isArray(userIdQuery)
     ? parseInt(userIdQuery.pop() || "")
-    : typeof userIdQuery === "string"
-    ? parseInt(userIdQuery)
-    : userIdQuery;
+    : stringToNumber.parse(req.query?.id);
```
Comment (zomars) [inline apps/web/pages/api/user/[id].ts:20]: Since we're not using defaultHandler this error won't be matched so we can:
1.  Refactor this endpoint to use `defaultHandler` and `defaultResponder` or...
2. use `safeParse` and check for success.

apps/web/pages/api/user/[id].ts
```diff
@@ -11,12 +12,12 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
 
   if (userId === null) {
```
Comment (zomars) [inline apps/web/pages/api/user/[id].ts:22]: This won't pass if userId is an empty string

apps/web/pages/api/user/[id].ts
```diff
@@ -11,8 +12,16 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     return res.status(401).json({ message: "Not authenticated" });
   }
 
+  const stringToNumber = z.string().transform((val) => parseInt(val));
+
   const userIdQuery = req.query?.id ?? null;
```
Comment (zomars) [inline apps/web/pages/api/user/[id].ts:17]: Let's just parse the whole req.query shall we?

zomars (changes requested):

apps/web/pages/api/user/[id].ts
Comment (zomars) [reply apps/web/pages/api/user/[id].ts:17]: Why userIdQuery can be an array?

Comment (zomars) [reply apps/web/pages/api/user/[id].ts:17]: If don't use an array for anything other that make it "not an array" then we shouldn't support an array.

`?userId=1&userId=2` will produce an array. What should happen here? Should we take the first one? The last one? Should we error?

My personal take is to error rather than allowing to introduce unexpected behavior

Comment (leog) [reply apps/web/pages/api/user/[id].ts]: Totally agree.

zomars (approved):

PR #3257, opened 2022-07-07
Update calendar when new attendees are added to bookings w/ seats

joeauyeung:
## What does this PR do?

This PR updates the organizer's calendar when new attendees are added to booking's with seats.

While working through this I noticed there are some improvements I can make to the seats feature so I'll be working on those.

Fixes # (issue)

## Type of change

## How should this be tested?

- Create an event type with seats
- Book a slot
- Now book that same slot with a different attendee
- Your connected calendar should show both attendees

## Checklist


apps/web/pages/api/book/event.ts
```diff
@@ -362,8 +363,19 @@ async function handler(req: NextApiRequest) {
...
+        references: {
+          select: {
+            type: true,
+            uid: true,
+            meetingId: true,
+            meetingPassword: true,
+            meetingUrl: true,
+            externalCalendarId: true,
+          },
+        },
```
joeauyeung, line 378: These are needed for the update calendar method

apps/web/pages/api/book/event.ts
```diff
@@ -401,6 +413,11 @@ async function handler(req: NextApiRequest) {
 
     await sendScheduledSeatsEmails(evt, invitee[0], newSeat);
 
+    user = users[0];
+    const credentials = await refreshCredentials(user.credentials);
+    const eventManager = new EventManager({ ...user, credentials });
+    await eventManager.addAttendee(evt, booking);
```
joeauyeung, line 419: Create an EventManager inside the if loop for seats

packages/core/EventManager.ts
```diff
@@ -156,7 +156,7 @@ export default class EventManager {
    *
    * @param event
    */
-  public async update(
+  public async reschedule(
```
joeauyeung, line 159: Since this method in the EventManager is only for rescheduling bookings, I decided to rename it to reflect it's purpose

packages/core/EventManager.ts
```diff
@@ -266,6 +266,10 @@ export default class EventManager {
     };
   }
 
+  public async addAttendee(event: CalendarEvent, booking: PartialBooking) {
+    await this.updateAllCalendarEvents(event, booking);
+  }
```
joeauyeung, line 271: Create a new method that only calls on updating the calendar, without the added logic of rescheduling

packages/core/EventManager.ts
zomars: This would need to be updated in every `eventManager.update` call or we're going to have trouble.

emrysal: Agree with @zomars - best for now is to @deprecate the update( function) or update all calls.

packages/core/EventManager.ts
```diff
@@ -266,6 +266,10 @@ export default class EventManager {
     };
   }
 
+  public async updateCalendarAttendees(event: CalendarEvent, booking: PartialBooking) {
```
emrysal, line 269: I wonder if you can make booking: more specific 💭

joeauyeung: Do you mean by adding all the fields that should be in booking? Would we just use the Booking type from Prisma client?

zomars: I think he means that we can be more explicit on which booking properties are needed for this function to work properly instead of just `PartialBooking`.

joeauyeung: @emrysal what would be best practice then? Would each function have it's own interface? So for this one `booking` would have `updateCalendarAttendeesBooking`?

emrysal (approved):
Having had a look; EventManager is the interface in front of our other integrations so it's a higher level - .reschedule( should be fine and PartialBooking is a minor issue that is non-blocking. Approving and LGTM

PR #3272, opened 2022-07-09
RFC: Custom eslint plugin

zomars:
## What does this PR do?

Since we're expecting many code changes and we may want to start enforcing best practices and avoiding known pitfalls. It occurred to me that a good way to achieve this is by start creating some custom eslint rules specific to our code base. This PR add a very basic one that prevent importing `dayjs` from other place that `@calcom/dayjs` to prevent a common pitfall of plugin conflicts.

Some other ideas I have for custom rules are:

- Using `defaultHandler` `defaultResponder` for API endpoints.
- Prevent importing server code on client code.
- Start displaying warnings for new bare Next.js API endpoints (and nudging into creating a tRPC router instead)
- Prefer early returns
- Prefer type inference
- etc, etc.

Please test this out and reply if this something we want and also more ideas for custom rules are welcome

https://www.loom.com/share/5902789bcb5d433b95fe72047828d7af

## Type of change

## How should this be tested?

- [ ] Eslint errors on your IDE
- [ ] `yarn lint`


PeerRich:
this is really cool

PeerRich (approved):

packages/eslint-plugin/src/rules/my-first-rule.ts
```diff
@@ -0,0 +1,23 @@
+import type { Rule } from "eslint";
```
baileypumfleet, line 1: Do we still need this my first rule?

pumfleet (dismissed):
Will block this until we get confirmation on if the example rule should still be included in here

packages/eslint-plugin/src/rules/my-first-rule.ts
zomars: I've kept it for reference/template but we can remove it. It's not enabled anyways.

PR #3315, opened 2022-07-11
When rescheduling only change calendar associated with booking

joeauyeung:
## What does this PR do?

Previously when rescheduling a booking, the event manager would try and update all connected calendars. This triggers the alarm system since the booking is only created on one calendar. This PR aims to fix that by only finding the calendar that the booking was created on initially.

Fixes # (issue)

## Type of change

## How should this be tested?

- Connect two calendars of the same type (I connected two google calendars)
- Create a booking
- Reschedule that booking now
- Go into the the DB and finding the calendar booking reference. Delete the credential ID
- Reschedule that booking again. It should be rescheduled on your calendar

## Checklist


apps/web/server/routers/viewer.tsx
```diff
@@ -498,12 +498,13 @@ const loggedInViewerRouter = createProtectedRouter()
...
         So create a default destination calendar with the first primary connected calendar
         */
-        const { integration = "", externalId = "" } = connectedCalendars[0].primary ?? {};
+        const { integration = "", externalId = "", credentialId } = connectedCalendars[0].primary ?? {};
         user.destinationCalendar = await ctx.prisma.destinationCalendar.create({
           data: {
             userId: user.id,
             integration,
             externalId,
+            credentialId,
```
joeauyeung, line 507: Add the credential ID to the destination calendar

packages/core/EventManager.ts
```diff
@@ -185,9 +187,9 @@ export default class EventManager {
             meetingPassword: true,
             meetingUrl: true,
             externalCalendarId: true,
+            credentialId: true,
           },
         },
-        destinationCalendar: true,
```
joeauyeung, line 190: Since we have the credential ID from the booking reference we don't need to grab the destination calendar

packages/core/EventManager.ts
```diff
@@ -377,36 +379,49 @@ export default class EventManager {
     event: CalendarEvent,
     booking: PartialBooking
   ): Promise<Array<EventResult<NewCalendarEventType>>> {
-    return async.mapLimit(this.calendarCredentials, 5, async (credential: Credential) => {
```
joeauyeung, line 380: Previous behaviour was to loop through every calendar credential. This was causing the errors and was adding unnecessary work. Should increase performance

packages/core/EventManager.ts
```diff
@@ -377,36 +379,49 @@ export default class EventManager {
...
-        let message = `Tried to 'updateAllCalendarEvents' but there was no '{thing}' for '${credential.type}', userId: '${credential.userId}', bookingId: '${booking.id}'`;
-        if (error instanceof Error) message = message.replace("{thing}", error.message);
-        console.error(message);
-        return Promise.resolve({
-          type: credential.type,
+    try {
+      // Bookings should only have one calendar reference
+      const calendarReference = booking.references.filter((reference) =>
+        reference.type.includes("_calendar")
+      )[0];
```
joeauyeung, line 386: Bookings are only created on one calendar so we just grab the first result from filter

packages/core/EventManager.ts
```diff
@@ -377,36 +379,50 @@ export default class EventManager {
...
+      const { uid: bookingRefUid, externalCalendarId: bookingExternalCalendarId } = calendarReference;
+
+      if (!bookingExternalCalendarId) throw new Error("externalCalendarId");
+
+      const result = [];
+      if (calendarReference.credentialId) {
+        const credential = this.calendarCredentials.filter(
+          (credential) => credential.id === calendarReference.credentialId
+        )[0];
+        result.push(updateEvent(credential, event, bookingRefUid, bookingExternalCalendarId));
```
joeauyeung, line 399: If the credential id is present then we can grab just the credential and update the event

packages/core/EventManager.ts
```diff
@@ -377,36 +379,50 @@ export default class EventManager {
...
+      } else {
+        const credentials = this.calendarCredentials.filter(
+          (credential) => credential.type === calendarReference.type
+        );
+        for (const credential of credentials) {
+          result.push(updateEvent(credential, event, bookingRefUid, bookingExternalCalendarId));
+        }
```
joeauyeung, line 406: If not then this is the previous behaviour of looping through all calendar credentials

packages/prisma/schema.prisma
```diff
@@ -110,8 +110,7 @@ model DestinationCalendar {
   externalId   String
   user         User?       @relation(fields: [userId], references: [id])
   userId       Int?        @unique
-  booking      Booking?    @relation(fields: [bookingId], references: [id])
-  bookingId    Int?        @unique
+  booking      Booking[]
```
joeauyeung, line 113: Changed this in destination calendar. Should have been a 1 to many relationship with bookings

apps/web/server/routers/viewer.tsx
```diff
@@ -560,7 +561,6 @@ const loggedInViewerRouter = createProtectedRouter()
       let where;
 
       if (eventTypeId) where = { eventTypeId };
-      else if (bookingId) where = { bookingId };
```
zomars, line 563: This is causing a lint warning on line 548

packages/core/EventManager.ts
zomars: Why not? is the destination calendar related to the credential? What happen to team event types where you can override the destinationCalendar per event-type?

packages/core/EventManager.ts
```diff
@@ -210,7 +217,7 @@ export default class EventManager {
     const results: Array<EventResult<Event>> = [];
     // If and only if event type is a dedicated meeting, update the dedicated video meeting.
     if (isDedicated) {
-      const result = await this.updateVideoEvent(evt, booking);
+      const result = await this.updateVideoEvent(evt, booking as unknown as PartialBooking);
```
zomars, line 220:
```suggestion
      const result = await this.updateVideoEvent(evt, booking);
```

I'm sorry @joeauyeung I can't allow you to do that.

packages/core/EventManager.ts
```diff
@@ -220,7 +227,7 @@ export default class EventManager {
     }
 
     // Update all calendar events.
-    results.push(...(await this.updateAllCalendarEvents(evt, booking)));
+    results.push(...(await this.updateAllCalendarEvents(evt, booking as unknown as PartialBooking)));
```
zomars, line 230:
```suggestion
    results.push(...(await this.updateAllCalendarEvents(evt, booking)));
```

packages/prisma/schema.prisma
```diff
@@ -234,6 +233,7 @@ model BookingReference {
   bookingId          Int?
   externalCalendarId String?
   deleted            Boolean?
+  credentialId Int?
```
zomars, line 236:
```suggestion
  credentialId     Int?
```

packages/prisma/schema.prisma
```diff
@@ -282,7 +282,8 @@ model Booking {
   status              BookingStatus        @default(ACCEPTED)
   paid                Boolean              @default(false)
   payment             Payment[]
-  destinationCalendar DestinationCalendar?
+  destinationCalendar DestinationCalendar? @relation(fields: [destinationCalendarId], references: [id])
+  destinationCalendarId Int?
```
zomars, line 286: run `yarn prisma format` to format these. Also VS Code should do this with prettier.

zomars (changes requested):
This is a nice upgrade @joeauyeung Nice work! There are some TS crimes that I cannot let you pass under the rug unfortunately. Let me know if you want to sync about those.

packages/core/EventManager.ts
joeauyeung: Since we're rescheduling a booking that means that the booking references are already created which means it should contain the specific credential id for the destination calendar of that event type.

PR #3322, opened 2022-07-11
Don't show readonly action buttons

Geczy:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


CarinaWolli:
When we remove the edit, duplicate and delete buttons, then we should probably also remove the embed button (as far as I tested it, it doesn't work with read-only). Also, the user can still click on the event type to get redirected to the editing page. So either we disable that too (but that would mean that the user also can't see all the detailed settings of the event type anymore) or we could at least disable the update button or maybe even disable all input fields on the editing page. Maybe @Jaibles could give some input here, what do you think would make the most sense?

Geczy:
i think setting all input fields to readonly makes sense

zlwaterfield:
@Geczy are you able to finish up this PR?

zomars:
Closing due stale changes

PR #3323, opened 2022-07-11
Only make web3 call if credential exists

### LINKED ISSUE #3207: Calls exodia.io/api/trending?page=1

### Issue Summary

Cal.com calls exodia.io and should not. OR it should be possible to disable it

In other words : we expect cal.com to be a schedule app not a NFT / crypto thing.

### Steps to Reproduce

1. Open network inspector
2. Edit an event type
3. 🍿

### Technical details

- Node.js version

joeauyeung:
## What does this PR do?

This PR adds the `hasweb3credential` prop to the event type page so it avoids making unnecessary calls.

Fixes #3207 

## Type of change

## How should this be tested?

- Go to an event type page without meta mask connected
- Check the console, token list should not be listed
- Go to the app store and connect meta mask
- Go to an event type page, token list should be logged


apps/web/pages/event-types/[type].tsx
```diff
@@ -346,40 +346,43 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
...
-          }
-        });
+          setTokensList(unifiedList);
+        } catch (err) {
+          showToast("Failed to load ERC20s & NFTs list. Please enter an address manually.", "error");
+        }
+      };
 
-        const unifiedList: Array<Token> = [...erc20sList, ...nftsList];
+      console.log(tokensList); // Just here to make sure it passes the gc hook. Can remove once actual use is made of tokensList.
```
Comment (agustif) [inline apps/web/pages/event-types/[type].tsx:379]: Can we get rid of the console log?

Comment (PeerRich) [inline apps/web/pages/event-types/[type].tsx:379]:
```suggestion
  
```

agustif:
LGTM, wondering if would be preferable to check for event-type.smartContractAdress presence instead and only load the token list in case of event-type actually being web3 type?

It should be way less calls than being tied to user settings, lots of users might have the metamask app installed, and this will still be called when they're looking at regular event-types. 

Just a thought, is a good fix too as it is maybe an improvement for a next PR later to bring external calls down a bit more. @zomars any thoughts on this?

zomars (changes requested):

apps/web/pages/event-types/[type].tsx
```diff
@@ -346,40 +346,41 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
...
-        showToast("Failed to load ERC20s & NFTs list. Please enter an address manually.", "error");
-      }
-    };
+          setTokensList(unifiedList);
+        } catch (err) {
+          showToast("Failed to load ERC20s & NFTs list. Please enter an address manually.", "error");
+        }
+      };
 
-    console.log(tokensList); // Just here to make sure it passes the gc hook. Can remove once actual use is made of tokensList.
```
Comment (joeauyeung) [inline apps/web/pages/event-types/[type].tsx:378]: Removed console.log

apps/web/pages/event-types/[type].tsx
```diff
@@ -346,40 +346,41 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
...
-    const fetchTokens = async () => {
-      // Get a list of most popular ERC20s and ERC777s, combine them into a single list, set as tokensList
-      try {
-        const erc20sList: Array<Token> =
-          //   await axios.get(`https://api.bloxy.info/token/list?key=${process.env.BLOXY_API_KEY}`)
-          // ).data
-          bloxyApi.slice(0, 100).map((erc20: Token) => {
-            const { name, address, symbol } = erc20;
-            return { name, address, symbol };
+    if (eventType.isWeb3Active) {
```
Comment (joeauyeung) [inline apps/web/pages/event-types/[type].tsx:349]: Uses isWeb3Active

apps/web/pages/event-types/[type].tsx
```diff
@@ -346,40 +346,43 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
   };
 
   useEffect(() => {
```
Comment (emrysal) [inline apps/web/pages/event-types/[type].tsx:348]: You can extract this logic into a separate component, then enable like:

```javascript
{props.hasweb3Integration && <Web3TokenList />
```

emrysal (commented):
None of suggestions are blocking - but have a look at the comments by Agusti and Zomars (which are)

apps/web/pages/event-types/[type].tsx
```diff
@@ -346,40 +346,43 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
...
-          //   await axios.get(`https://api.bloxy.info/token/list?key=${process.env.BLOXY_API_KEY}`)
-          // ).data
-          bloxyApi.slice(0, 100).map((erc20: Token) => {
-            const { name, address, symbol } = erc20;
-            return { name, address, symbol };
+    if (props.hasweb3Integration) {
+      const fetchTokens = async () => {
+        // Get a list of most popular ERC20s and ERC777s, combine them into a single list, set as tokensList
+        try {
+          const erc20sList: Array<Token> =
```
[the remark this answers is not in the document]
Comment (joeauyeung) [reply apps/web/pages/event-types/[type].tsx:353]: Why would we not want to type here? I always pictured the best case scenario that everything would be typed?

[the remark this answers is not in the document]
Comment (emrysal) [reply apps/web/pages/event-types/[type].tsx:353]: Github threw off the ordering of the comments. But it's a happy dance between typing explicitly and type inference, and you want to leave the inference happen as much as possible. Two ways of doing that; either typing both `erc20sList` and `nftsList` or typing the resulting array. As the resulting array is the one you use, I think typing the latter is the most important, leaving type inference to handle the rest.

apps/web/pages/event-types/[type].tsx
Comment (joeauyeung) [reply apps/web/pages/event-types/[type].tsx:348]: I just realized `tokensList` isn't even being used anywhere in the event type page. Do we even need to keep all those calls being made?

apps/web/pages/event-types/[type].tsx
```diff
@@ -346,40 +346,43 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
...
 
-        const nftsList: Array<Token> = exodiaList.map((nft: NFT) => {
-          const { name, contracts } = nft;
-          if (nft.contracts[0]) {
-            const { address, symbol } = contracts[0];
-            return { name, address, symbol };
-          }
-        });
+          setTokensList(unifiedList);
+        } catch (err) {
```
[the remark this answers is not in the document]
Comment (joeauyeung) [reply apps/web/pages/event-types/[type].tsx:374]: I'm not sure what you mean here. Could you give an example of how you would make it more specific?

zomars:
> > why do we even make this request in the first place? what does it do? it seems like its getting a list of token names but we don't seem to use them?
> 
> I know right, the calls change `tokenList` which isn't even being used anywhere. Should we just comment it out for now or delete it entirely?

I would rather remove it. If we need this code again, we can always check git file history.

emrysal:
If this list truly isn't used anywhere, why was is there, what's it's intended to do? Could be a historical merge gone wrong we just never noticed. Worth investigating. If it does turn out to be useless, ditch it; dead code.

zomars (approved):
easy GIF

PR #3326, opened 2022-07-11
Fix Caldav not valid destination calendar and only display calendars that user can edit

joeauyeung:
## What does this PR do?

This PR fixes the bug where users were not able to set caldav calendars as the destination calendar. This PR also only displays possible destination calendars that users are able to create events on.

Fixes # (issue)

## Type of change

## How should this be tested?

- Connect a caldav calendar
- Go to an event type and choose the caldav calendar as the destination calendar
- Create a booking
	- The event should show on the caldav calendar
	
- Add an other calendar account that has a read only calendar
- On the installed apps and event type page when choosing a destination calendar, the read only calendar should not be an option

## Checklist


apps/web/components/DestinationCalendarSelector.tsx
```diff
@@ -61,10 +61,12 @@ const DestinationCalendarSelector = ({
     query.data.connectedCalendars.map((selectedCalendar) => ({
       key: selectedCalendar.credentialId,
       label: `${selectedCalendar.integration.title} (${selectedCalendar.primary?.name})`,
-      options: (selectedCalendar.calendars ?? []).map((cal) => ({
-        label: cal.name || "",
-        value: `${cal.integration}:${cal.externalId}`,
-      })),
+      options: (selectedCalendar.calendars ?? [])
+        .filter((cal) => cal.readOnly === false)
```
joeauyeung, line 65: Only show calendars that are not readOnly

packages/core/CalendarManager.ts
```diff
@@ -44,6 +44,7 @@ export const getConnectedCalendars = async (
         const calendars = _(cals)
           .map((cal) => ({
             ...cal,
+            readOnly: cal.readOnly || false,
```
joeauyeung, line 47: When getting list of calendars set default readOnly to false unless stated

zomars (approved):
Noice Thats Nice GIF

PR #3345, opened 2022-07-12
Fix possible 404 when obtaining calendar busy lists

alannnc:
## What does this PR do?

When a calendar it's disconnected it leaves a row on selectedCalendars. When obtaining user availability this disconnected calendar gets called and return 404 throwing error and stopping slot checkin.
This fixes any error while obtaining user busy times and when disconnecting calendar we now delete selectedCalendar if required.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Connect several calendars from different accounts, mark them as selected.
- [ ] Then try disconnecting them and they should not longer remain on selected calendars.

## Checklist


apps/web/components/integrations/CalendarListContainer.tsx
```diff
@@ -153,6 +153,7 @@ function ConnectedCalendarsList(props: Props) {
                     actions={
                       <DisconnectIntegration
                         id={item.credentialId}
+                        externalId={item.primary?.externalId}
```
alannnc, line 156: In case there exist an externalId we pass it down to mutation.

apps/web/server/routers/viewer.tsx
```diff
@@ -1040,14 +1041,41 @@ const loggedInViewerRouter = createProtectedRouter()
...
+          }
+
+          if (externalId) {
+            const existingSelectedCalendar = await prisma.selectedCalendar.findFirst({
+              where: {
+                externalId: externalId,
+              },
+            });
+            // @TODO: SelectedCalendar doesn't have unique ID so we should only delete one item
+            if (existingSelectedCalendar) {
```
alannnc, line 1068: This was necessary because after disconnecting a calendar it will remain on selectedCalendars.

packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -140,16 +140,18 @@ export default class Office365CalendarService implements Calendar {
...
             acc.concat(
-              subResponse.body.value
-                .filter((evt) => evt.showAs !== "free" && evt.showAs !== "workingElsewhere")
-                .map((evt) => {
-                  return {
-                    start: evt.start.dateTime + "Z",
-                    end: evt.end.dateTime + "Z",
-                  };
-                })
+              subResponse.body?.value
```
alannnc, line 145: When a calendar was disconnected and getting here, subResponse.body was { error: Object } and not value thus throwing error and stopping slot available validation.

zomars (approved):
approved GIF

pumfleet (approved):
Looks good to me, haven't tested

zomars:
@alannnc can we check why tests are failing?

PR #3361, opened 2022-07-14
kbar follow up

PeerRich:
https://www.loom.com/share/2ed0ad9af45045969ddd442633da1713

- added more routes to kbar
- moved icon next to logo
- added keyboard hints
- added gray back drop
- increased width to 640px
- added `react-feather` since we are moving from HeroIcons to React-Feather anyway so may as well start
- **new** introduce `side` prop to `<ToolTip />` with "left, top, right, bottom and undefined" values


packages/ui/Tooltip.tsx
```diff
@@ -12,6 +15,7 @@ export function Tooltip({
   children: React.ReactNode;
   content: React.ReactNode;
   open?: boolean;
+  side?: "top" | "right" | "bottom" | "left" | undefined;
```
alishaz-polymath, line 18:
```suggestion
  position?: "top" | "right" | "bottom" | "left" | undefined;
```

alishaz-polymath (commented):
Looks good but wouldn't `position` be better name than `side` for the tooltip position prop? 💭

apps/web/components/Kbar.tsx
```diff
@@ -185,33 +195,40 @@ export const KBarContent = () => {
...
+          <div className="flex items-center justify-center border-b">
+            <SearchIcon className="mx-3 h-4 w-4 text-gray-500" />
+            <KBarSearch className="w-full rounded-sm py-2.5 focus-visible:outline-none" />
+          </div>
           <RenderResults />
+          <div className="flex items-center space-x-1 border-t px-2 py-1.5 text-gray-500">
+            <SwitchVerticalIcon className="h-4 w-4" /> <span className="pr-2 text-xs">Navigate</span>
+            <CornerDownLeft className="h-4 w-4" />
+            <span className="pr-2 text-xs">Open</span>
+            <Command className="h-3 w-3" />
```
alishaz-polymath, line 208: Also, don't we want it to be CTRL for Windows?

packages/ui/Tooltip.tsx
emrysal: I also thought this, then Googled and in Bootstrap for example it's called `placement` - position there implies the orientation. Naming in Tooltip world is quite messy.

apps/web/components/Kbar.tsx
```diff
@@ -185,33 +195,40 @@ export const KBarContent = () => {
...
+            <span className="pr-1 text-xs">+ K </span>
+            <span className="pr-2 text-xs">Close</span>
```
alishaz-polymath, line 210: Command wouldn't be the best for windows users, also we can close the bar using ESC key (I believe on both mac and windows) so maybe we can think about that?

packages/ui/Tooltip.tsx
alishaz-polymath: Oh well. Placement still sounds better than side to me though.

PeerRich: `side` is actually the native prop that comes from radix, which is why I used it

image

apps/web/components/Kbar.tsx
PeerRich: fixed

alishaz-polymath (approved):
Amazing stuff. 🙏 🚀

PR #3375, opened 2022-07-14
When rescheduling update specific calendar

joeauyeung:
## What does this PR do?

*Repeat of #3315*

Previously when rescheduling a booking, the event manager would try and update all connected calendars. This triggers the alarm system since the booking is only created on one calendar. This PR aims to fix that by only finding the calendar that the booking was created on initially.

Fixes # (issue)

## Type of change

## How should this be tested?

- Connect two calendars of the same type (I connected two google calendars)
- Create a booking
- Reschedule that booking now
- Go into the the DB and finding the calendar booking reference. Delete the credential ID
- Reschedule that booking again. It should be rescheduled on your calendar

## Checklist


joeauyeung (commented):
Addressed lint error, prisma format, and no cheating typescript.

zomars (approved):

PR #3386, opened 2022-07-15
Fix the path for "schema.prisma" file in README.md

taha-crafter:
## What does this PR do?

This is a tiny PR to fix path to the schema.prisma file in the README.md, I noticed that the previous path `apps/web/prisma/schema.prisma` doesn't exist anymore.


zomars (changes requested):
Nice catch, and thank you. Just one thing, can we make this PR without the `yarn.lock` file changes?

taha-programmer:
> Nice catch, and thank you. Just one thing, can we make this PR without the `yarn.lock` file changes?

Thanks Omar, I removed the yarn.lock file.

zomars (approved):

PR #3401, opened 2022-07-16
Remove shallow routing

sean-brydon:
Fixes /booking/[status] not updating


hariombalhara (approved):
This seems to be a bug with Next.js Middleware and Shallow Routing together in latest versions
Even though it says that it has been fixed, I can still replicate the issue when middleware file exists. After removing middleware file, it goes away

https://github.com/vercel/next.js/pull/37716

But for now, as a hotfix this is good to go 👍

PR #3411, opened 2022-07-18
Ensure / is used for paths in import

hariombalhara:
Ensure apps.browser.generated.tsx imports have / in them even for windows.


alishaz-polymath (approved):
Works perfectly on Windows 🙏 🚀

PR #3427, opened 2022-07-19
Hotfix: Unrelated bookings marked as cancelled due to recurring event

hariombalhara:
Reported internally, so more details in this internal twist thread https://twist.com/a/195135/ch/554218/t/3633091/

### DIFF
--- a/apps/web/pages/api/cancel.ts
+++ b/apps/web/pages/api/cancel.ts
@@ -171,20 +171,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   // action always succeeds even if subsequent integrations fail cancellation.
   if (bookingToDelete.eventType?.recurringEvent) {
     const recurringEventId = bookingToDelete.recurringEventId;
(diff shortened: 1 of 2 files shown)


apps/web/pages/api/cancel.ts
```diff
@@ -170,11 +170,24 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+    if (recurringEventId === null) {
+      where = {
+        uid,
+      };
+    } else {
+      where = {
+        recurringEventId,
+      };
+    }
+
```
alishaz-polymath, line 187:
```suggestion
    const where = recurringEventId === null ? {uid} : {recurringEventId};
```
This should work too, right?

alishaz-polymath (approved):
Looks good to me as a solid starting point for a hotfix

leog (commented):
Looks good, testing locally now

zomars (approved):

PR #3448, opened 2022-07-19
Hotfix: Cancelling recurring vs single event

leog:
## What does this PR do?

When cancelling a single event in a recurring event series, it was cancelling all events, not a single instance.
This also covers a fix to show correctly the remaining events in a series taking into account cancelled/rejected events, as well as just cancelling remaining events in the future, not all events in a series, to reflect the correct state to past events that were accepted that probably did happen.

## Type of change

## How should this be tested?

Confirm a recurring event upon creation or use the mocked ones. Try to cancel one instance from "Upcoming" tab, and see that you only get an email to just one instance, as well as the correct cancel page with the correct information for just that instance. Also, in "Recurring" tab for bookings, once cancelled, the number of remaining events should be correct as well as show upon hovering the dotted line, the correct remaining event instances dates.

### DIFF
--- a/apps/web/components/booking/BookingListItem.tsx
+++ b/apps/web/components/booking/BookingListItem.tsx
@@ -14,7 +14,6 @@ import { useState } from "react";
 import { useMutation } from "react-query";
 
 import dayjs from "@calcom/dayjs";
-import { parseRecurringEvent } from "@calcom/lib";
 import classNames from "@calcom/lib/classNames";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import showToast from "@calcom/lib/notification";
@@ -28,7 +27,7 @@ import { HttpError } from "@lib/core/http/error";
 import useMeQuery from "@lib/hooks/useMeQuery";
 import { linkValueToString } from "@lib/linkValueToString";
 import { LocationType } from "@lib/location";
-import { parseRecurringDates } from "@lib/parseDate";
+import { extractRecurringDates } from "@lib/parseDate";
 import { inferQueryInput, inferQueryOutput, trpc } from "@lib/trpc";
 
 import { EditLocationDialog } from "@components/dialog/EditLocationDialog";
@@ -41,7 +40,7 @@ type BookingItem = inferQueryOutput<"viewer.bookings">["bookings"][number];
 
 type BookingItemProps = BookingItem & {
   listingStatus: BookingListingStatus;
-  recurringCount?: number;
+  recurringBookings?: BookingItem[];
 };
 
 function BookingListItem(booking: BookingItemProps) {
@@ -127,7 +126,11 @@ function BookingListItem(booking: BookingItemProps) {
         booking.listingStatus === "recurring" && booking.recurringEventId !== null
           ? t("cancel_all_remaining")
           : t("cancel"),
-      href: `/cancel/${booking.uid}`,
+      href: `/cancel/${booking.uid}${
+        booking.listingStatus === "recurring" && booking.recurringEventId !== null
+          ? "?allRemainingBookings=true"
+          : ""
+      }`,
       icon: XIcon,
     },
     {
@@ -202,14 +205,10 @@ function BookingListItem(booking: BookingItemProps) {
   let recurringStrings: string[] = [];
   let recurringDates: Date[] = [];
   const today = new Date();
-  if (booking.recurringCount && booking.eventType.recurringEvent?.freq !== undefined) {
-    [recurringStrings, recurringDates] = parseRecurringDates(
-      {
-        startDate: booking.startTime,
-        timeZone: user?.timeZone,
-        recurringEvent: parseRecurringEvent(booking.eventType.recurringEvent),
-        recurringCount: booking.recurringCount,
-      },
+  if (booking.recurringBookings && booking.eventType.recurringEvent?.freq !== undefined) {
+    [recurringStrings, recurringDates] = extractRecurringDates(
+      booking.recurringBookings,
+      user?.timeZone,
       i18n
     );
     if (booking.status === BookingStatus.PENDING) {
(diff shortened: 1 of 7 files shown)


apps/web/components/booking/BookingListItem.tsx
```diff
@@ -127,7 +126,11 @@ function BookingListItem(booking: BookingItemProps) {
         booking.listingStatus === "recurring" && booking.recurringEventId !== null
           ? t("cancel_all_remaining")
           : t("cancel"),
-      href: `/cancel/${booking.uid}`,
+      href: `/cancel/${booking.uid}${
+        booking.listingStatus === "recurring" && booking.recurringEventId !== null
+          ? "?allRemainingBookings=true"
```
leog, line 131: When cancelling we need to let the UI and the API know if the intention is to cancel all remaining bookings or just that booking instance.

apps/web/components/booking/BookingListItem.tsx
```diff
@@ -202,14 +205,10 @@ function BookingListItem(booking: BookingItemProps) {
...
-  if (booking.recurringCount && booking.eventType.recurringEvent?.freq !== undefined) {
-    [recurringStrings, recurringDates] = parseRecurringDates(
-      {
-        startDate: booking.startTime,
-        timeZone: user?.timeZone,
-        recurringEvent: parseRecurringEvent(booking.eventType.recurringEvent),
-        recurringCount: booking.recurringCount,
-      },
+  if (booking.recurringBookings && booking.eventType.recurringEvent?.freq !== undefined) {
+    [recurringStrings, recurringDates] = extractRecurringDates(
```
leog, line 209: We now have the list of bookings to extract these dates as we should have. Before we always generated the list of dates from the recurring configuration, always showing all booked instances whether cancelled or not, not the real remaining events in the series.

apps/web/lib/parseDate.ts
```diff
@@ -46,3 +47,15 @@ export const parseRecurringDates = (
   });
   return [dateStrings, rule.all()];
 };
+
+type BookingItem = inferQueryOutput<"viewer.bookings">["bookings"][number];
+
+export const extractRecurringDates = (
```
leog, line 53: From a list of bookings, return the list of start date of all of them as dates and as strings.

apps/web/pages/api/cancel.ts
```diff
@@ -168,12 +173,17 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
     const recurringEventId = bookingToDelete.recurringEventId;
     const where = recurringEventId === null ? { uid } : { recurringEventId };
-    // Proceed to mark as cancelled all recurring event instances
+    // Proceed to mark as cancelled all remaining recurring events instances (greater than or equal to right now)
     await prisma.booking.updateMany({
-      where,
+      where: {
+        ...where,
+        startTime: {
+          gte: new Date(),
```
leog, line 184: We just want to cancel the remaining recurring events, not past instances that were probably accepted may have happened.

apps/web/pages/cancel/[uid].tsx
```diff
@@ -81,9 +81,13 @@ export default function Type(props: inferSSRProps<typeof getServerSideProps>) {
...
-                            {props.cancellationAllowed && !props.booking?.eventType.recurringEvent
-                              ? t("reschedule_instead")
-                              : t("event_is_in_the_past")}
+                            {!props.booking?.eventType.recurringEvent
+                              ? props.cancellationAllowed
+                                ? t("reschedule_instead")
+                                : t("event_is_in_the_past")
+                              : allRemainingBookings === "true"
+                              ? t("cancelling_all_recurring")
+                              : t("cancelling_event_recurring")}
```
Comment (leog) [inline apps/web/pages/cancel/[uid].tsx:90]: We are now relying on knowing if the cancellation is just for an event in a series or the entire series to show correct messaging when the user wants to cancel a booking.

apps/web/pages/cancel/[uid].tsx
```diff
@@ -272,10 +278,11 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
   });
 
   let recurringInstances = null;
-  if (booking.eventType?.recurringEvent) {
+  if (booking.eventType?.recurringEvent && allRemainingBookings === "true") {
     recurringInstances = await prisma.booking.findMany({
       where: {
         recurringEventId: booking.recurringEventId,
+        NOT: [{ status: "CANCELLED" }, { status: "REJECTED" }],
```
Comment (leog) [inline apps/web/pages/cancel/[uid].tsx:285]: Now we show the correct recurring instances remaining when trying to cancel the entire series.

apps/web/server/routers/viewer.tsx
```diff
@@ -428,11 +428,6 @@ const loggedInViewerRouter = createProtectedRouter()
         skip,
       });
 
-      const groupedRecurringBookings = await prisma.booking.groupBy({
-        by: [Prisma.BookingScalarFieldEnum.recurringEventId],
-        _count: true,
-      });
-
```
leog, line 435: As a bonus, we removed this extra query from the bookings query as we now rely on the already returned `bookings` list.

apps/web/pages/cancel/[uid].tsx
```diff
@@ -224,6 +229,7 @@ export default function Type(props: inferSSRProps<typeof getServerSideProps>) {
 export const getServerSideProps = async (context: GetServerSidePropsContext) => {
   const ssr = await ssrInit(context);
   const session = await getSession(context);
+  const allRemainingBookings = asStringOrNull(context.query.allRemainingBookings) || "";
```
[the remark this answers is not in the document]
Comment (zomars) [reply apps/web/pages/cancel/[uid].tsx:232]: We can even cast to an actual boolean using `zod` instead of checking for the `true` string.

apps/web/server/routers/viewer.tsx
zomars: Does this is still needed on `apps/web/pages/api/book/confirm.ts` tho? at line 200?

https://github.com/calcom/cal.com/blob/805419d58ba6918a305f66a874a3e2fdddbdad4b/apps/web/pages/api/book/confirm.ts#L198

zomars (approved):
Overall is looking good. Impressive work @leog left some minor nitpicks and testing right now.

zomars:
Maybe unrelated but found out this:

Clicking on "Back to bookings" after cancelling a single recurring event.

image

It navigates to `/bookings` and it doesn't redirect you to `/bookings/upcoming` so the skeleton is just stays there, loading.

image

zomars:
@leog more feedback, I've forgot about the auto merge label:

https://clips.twitch.tv/SullenHappyChimpanzeePeteZarollTie-xs9DXjpUm7nN50hp

apps/web/server/routers/viewer.tsx
leog: Yes, is needed, as that covers the API side when confirming events to be able to show correctly the number of actual events booked in the email, as a recurring event may allow 10 events and a user can book less than that.

PR #3453, opened 2022-07-19
fix/error-handling-with-proper-status

alannnc:
## What does this PR do?

When receiving 204 on handleError we try to do response.json() but since it's 204 there is no content and we flag it as Error. Now we validate response statusCode and if different from 2XX we throw error.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Schedule a meeting with zoom video as location
- [ ] Cancel it, now it should be cancelled and don't throw Error

## Checklist


packages/lib/errors.ts
```diff
@@ -11,15 +11,21 @@ export function getErrorFromUnknown(cause: unknown): Error & { statusCode?: numb
 }
 
 export function handleErrorsJson(response: Response) {
-  if (!response.ok) {
+  if (response.status === 204) {
+    return {};
+  }
```
alannnc, line 16: Early return empty object in case its 204

packages/lib/errors.ts
```diff
@@ -11,15 +11,21 @@ export function getErrorFromUnknown(cause: unknown): Error & { statusCode?: numb
...
+  if (!response.ok && response.status < 200 && response.status >= 300) {
```
alannnc, line 17: Handle all non successful statusCodes.

zomars (approved):

zomars (approved):

zomars (approved):

PR #3465, opened 2022-07-20
Fix - Avatar on New Event Type

### LINKED ISSUE #3336: Avatar on New Event Type

### Issue Summary

When selecting a new event type, if you have an individual plan as well as a team plan. The Team Plan's Avatar shows, but the individual users avatar does not show. For consistency, the users avatar should show as well.
CleanShot 2022-07-11 at 17 26 38@2x

### Steps to Reproduce

1. (for example) Went to ...
2. Clicked on...
3. ...

### Technical details

- Node.js version

noobyogi0010:
## What does this PR do?

When selecting a new event type, if you have an individual plan as well as a team plan. The Team Plan's Avatar shows, but the individual user's avatar does not show. So, this PR fixes this issue. Now if there is no avatar then the default avatars will be shown.

Fixes #3336 

 Loom Video: https://www.loom.com/share/3d21119e4bec498da7135e79c1306c22

## Type of change

## How should this be tested?

- Login with a user who has a team associated with him.
- Make sure that the user doesn't have an avatar and the team has an avatar,
- Navigate to `Event Type` and click on the `New Event Type` dropdown. You will be able to see the avatar of the team and the default avatar for the user.


sean-brydon (approved):
Tested Locally - works as expected. 

Thank you @noobyogi0010

noobyogi0010:
Thank you for validating it! @sean-brydon

zomars (approved):

PR #3478, opened 2022-07-21
Disallow search engine indexing of hidden event types

### LINKED ISSUE #3311: Disallow search engine indexing of hidden event types
### Discussed in https://github.com/calcom/cal.com/discussions/3285

<sup>Originally posted by **GlitchWitch** July  5, 2022</sup>

When creating a hidden event type, there is the chance that if the link is shared that it will be crawled and indexed by search engines.

When marking an event type as hidden, we should add `<meta name="robots" content="noindex">` to the page html header to [prevent indexing](https://developers.google.com/search/docs/advanced/crawling/block-indexing).

Another possible solution is to add a checkbox to allow the user to decide if the event type can be indexed or not, however I think just linking this to the hidden parameter is good enough.

### Additional context

One of my hidden event types currently shows up in search engines as I share it on a public post with a specific CTA and purpose. However I have deliberately chosen to make this event type hidden and do not want it indexed. </div>

gitstart:
## What does this PR do?

When creating a hidden event type, there is the chance that if the link is shared that it will be crawled and indexed by search engines.

Fixes #3311

 Loom Video: https://www.loom.com/share/55bcf150df944877ae1cf08fd552b757

## Type of change


emrysal (approved):

emrysal:
Excellent KISS implementation

PR #3481, opened 2022-07-21
No days available to schedule a meeting

### LINKED ISSUE #3340: No days available to schedule a meeting

### Issue Summary

When I click "preview" to schedule a meeting time, there is no available day

### Steps to Reproduce

1. Create a new event
2. Click on preview

All options are off:
Screenshot from 2022-07-12 09-31-26

gitstart:
## What does this PR do?

When I click on "preview" to schedule a meeting there were no days available, this problem was being caused by import problems

the problem was that the "performance" is not defined in some components, so the pages where this problem existed had problems.
But if you change the node version to >=16 the problem is solved too

Fixes #3340 

Screenshot from 2022-07-12 09-59-55

## Type of change


hariombalhara:
@PeerRich Should we close this PR or may be add some label to filter it out  as it has no activity for almost a month and it isn't clear what it's fixing.

PeerRich:
@hariombalhara can you confirm this helped?

zomars:
Thank you for your contribution 🙏 this was properly fixed in #3823

PR #3484, opened 2022-07-21
Migrates all tRPC code to a monorepo package

### LINKED ISSUE #3392: Extract bookings router to shared @calcom/trpc package to be re-used in Public API
In order to provide same actions to Public API calls than webapp uses, we need to extract out our TRPC routers (querys/mutations) so they can be called from both, and we mantain a single source of truth for it.

Create a new @calcom/trpc shared package that hosts this new shared trpc routers, and start with bookings router.
Make both webapp and public API re-use this new router.

### Additional context
- trpc v10
- trpc zART and other repos that show how to share trpc api server code between clients

zomars:
Fixes #3392 

needed for #3490 

## What does this PR do?

- Migrates all tRPC code to a monorepo package
- Allows reusing backend code in App Store
- Allows reusing backend logic in open API
- Allows easier migration to tRPC v10 once it's released

## Type of change

## How should this be tested?

- [ ] `yarn build`
- [ ] `yarn test-e2e`


zomars:
Would love to hear you input as well @KATT

KATT (commented):
Makes sense to me to have a pure trpc or "api" package :)

pumfleet (commented):
Read the code, didn't test. Looks good to me

package.json
```diff
@@ -11,7 +11,7 @@
   ],
   "scripts": {
     "build": "turbo run build --scope=\"@calcom/web\" --include-dependencies",
-    "clean": "turbo run clean && rm -rf node_modules",
+    "clean": "find . -name node_modules -o -name .next -o -name .turbo -o -name dist -type d -prune | xargs rm -rf",
```
hariombalhara, line 14: Nice. Didn't know you can do OR like this 👍

hariombalhara: Doesn't handle special cases like this. Not a blocker but why are we moving away from the earlier approach.
Maybe we can use both, for cases like this.
Screenshot 2022-07-22 at 7 36 07 PM
)

hariombalhara:
Codewise looks good. Testing now

hariombalhara:
Screenshot 2022-07-22 at 7 52 28 PM
Getting this error when running website server. Can't seem to get away with it. 

I am not sure if this is related to it, or something related to my local setup only

hariombalhara (approved):
Looks good. Did basic testing of most of the pages, app store, google calendar, Routing Forms, Availability, Booking, settings

hariombalhara:
Tests are failing though

apps/web/pages/team/[slug].tsx
```diff
@@ -8,14 +8,14 @@ import React, { useEffect } from "react";
 
 import { useIsEmbed } from "@calcom/embed-core/embed-iframe";
 import { CAL_URL } from "@calcom/lib/constants";
+import { getTeamWithMembers } from "@calcom/trpc/server/queries/teams";
```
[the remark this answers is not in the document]
Comment (zomars) [reply apps/web/pages/team/[slug].tsx:11]: Should we move them to `@calcom/prisma` or `@calcom/lib`instead? @hariombalhara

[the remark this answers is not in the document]
Comment (hariombalhara) [reply apps/web/pages/team/[slug].tsx:11]: I would prefer, @calcom/lib. There can be any way to fetch the data. Details don't matter to user.

package.json
zomars: Will follow up @hariombalhara

leog:
Was going to ask if there is already in place an ESLint warning for this as you created for something else to improve DX when using ´trpc´ @zomars

zomars:
> Was going to ask if there is already in place an ESLint warning for this as you created for something else to improve DX when using ´trpc´ @zomars

It should be trivial to do, taking the `dayjs` one as example @leog

PR #3393, opened 2022-07-15
Self-hosted onboarding first admin wizard

leog:
## What does this PR do?

Implements a wizard to create the first user (admin) as an on-boarding step to the app. The on-boarding will continue to expand in the near future which is why it was implemented this way.

UI:
<kbd>image</kbd>

Can't be used again once at least one user exists in the DB:
<kbd>image</kbd>

Implements #3384 

**Environment**: Staging(main branch) / Production

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How should this be tested?

Remove all users from the DB and go to the login page (not having a session active). It will show the new onboarding wizard. You can also test going to `/auth/setup` having users in the DB to see the second screen above, indicating the step is done (not needed). The API that supports this new functionality also checks if there is no user in the DB to be able to enable creating an admin user through the new onboarding wizard.

Issue #3384, opened 2022-07-15
Self-hosted onboarding first admin wizard


leog:
<div type='discussions-op-text'>

<sup>Originally posted by **@zomars** July 14, 2022 in [Twist thread](https://twist.com/a/195135/ch/555266/t/3624478/)</sup> 

We should strive to improve self-hosted experience by including an onboarding for new installs. This is very common for PHP Symfony projects (cc @emrysal who reminded me of this) to have a “[Create your first user](https://docs.boltcms.io/5.0/manual/first-user)” screen when you’re on a new install:

<img src="https://user-images.githubusercontent.com/467258/179259019-0a273503-002b-45bd-9203-b928989cf8fa.png" width="400"/>

We already have the new onboarding design on console so it should be trivial to make a new screen for it.

</div>

<div type='discussions-op-text'>
> @PeerRich: this should live in the deployment wozard

@zomars: Agreed but also should live on webapp on a new install without users
</div>

zomars:
- On `/auth/login` check the `prisma.users.count` being `0`
- Redirect to `/auth/setup` (new page) where the wizard will be displayed to create the first admin user.
- Once created, redirect back to `/auth/login`

zomars:
We're planning to add more customization to the wizard later on. So we might just do it like this from start. Also we delegate this decision to self-hostess. Nothing is blocking them to make their own seeder for their own needs.

leog:
Implemented by #3486.

PR #3051, opened 2022-06-13
Feat/design system 

sean-brydon:
## What does this PR 

Check [Cal DS v2.0 Figma file](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%F0%9F%94%98-Cal-DS-v2.0?node-id=309%3A39382) to see progress of what has been created in storybook.

https://62bb0d967cb8b3b4e3e3f43a-bhhilfxoly.chromatic.com/ for the current build in chromatic

**Environment**: Staging(main branch) / Production

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

- I haven't added tests that prove my fix is effective or that my feature works

PR #3486, opened 2022-07-21
Feat/onboarding admin

zomars:
Re-opening #3393 

## What does this PR do?

Implements a wizard to create the first user (admin) as an on-boarding step to the app. The on-boarding will continue to expand in the near future which is why it was implemented this way.

UI:
<kbd>image</kbd>

Can't be used again once at least one user exists in the DB:
<kbd>image</kbd>

Implements #3384 

## Type of change

## How should this be tested?

Remove all users from the DB and go to the login page (not having a session active). It will show the new onboarding wizard. You can also test going to `/auth/setup` having users in the DB to see the second screen above, indicating the step is done (not needed). The API that supports this new functionality also checks if there is no user in the DB to be able to enable creating an admin user through the new onboarding wizard.


apps/web/pages/auth/setup.tsx
```diff
@@ -0,0 +1,293 @@
...
+import { zodResolver } from "@hookform/resolvers/zod";
+import { useRouter } from "next/router";
+import { Controller, useForm } from "react-hook-form";
+import * as z from "zod";
+
+import { isPasswordValid } from "@calcom/lib/auth";
+import classNames from "@calcom/lib/classNames";
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+import { inferSSRProps } from "@calcom/types/inferSSRProps";
+import WizardForm from "@calcom/ui/WizardForm";
```
zomars, line 11: @leog cc @sean-brydon maybe we want to put this in `/v2/` for now. Also, is this the same that we have on console?

packages/lib/auth.ts
```diff
@@ -9,3 +9,19 @@ export async function verifyPassword(password: string, hashedPassword: string) {
   const isValid = await compare(password, hashedPassword);
   return isValid;
 }
+
+export function isPasswordValid(password: string) {
```
zomars, line 13: Let's put this in `packages/lib` instead

packages/ui/Stepper.tsx
```diff
@@ -0,0 +1,50 @@
+import Link from "next/link";
+
+type DefaultStep = {
+  title: string;
+};
+
+function Stepper<T extends DefaultStep>(props: { href: string; step: number; steps: T[] }) {
```
zomars, line 7: Put it in `/v2/` for now? Also let's try to hurry with #3051

packages/ui/WizardForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+import Stepper from "./Stepper";
+
+type DefaultStep = {
+  title: string;
+  description: string;
+  content: JSX.Element;
+  enabled?: boolean;
+};
+
+function WizardForm<T extends DefaultStep>(props: { href: string; steps: T[] }) {
```
zomars, line 12: Same v2 comment

zomars (commented):
Got some Qs. Looking good so far, nice work @leog 🙏🏽

apps/web/pages/auth/setup.tsx
leog: Yes, is the same, moved it from `console` to here and added with a few tweaks. Sure, I can move it to `v2` along the other components.

packages/ui/WizardForm.tsx
sean-brydon: V2 Should be good to merge into main now - agreed these should live in V2

leog:
Will be reworked once #3051 is merged, converting to draft in the meantime.

packages/lib/auth.ts
leog: @zomars you mean in the ´index.ts´ file instead of ´auth.ts´?

apps/web/pages/auth/setup.tsx
leog: @zomars after chatting with @sean-brydon it seems like we can't move these components to `v2` as there is a color scheme mismatch between that and web where these components are being used. So, for now we should stay put.

packages/ui/Stepper.tsx
leog: We can't move, see comment above.

packages/ui/WizardForm.tsx
leog: We can't move, see comment above.

apps/web/pages/auth/setup.tsx
zomars: Let's update these components to match the new color scheme. This shouldn't be a blocker

leog:
Here is a Loom with the latest changes: https://www.loom.com/share/ebf85cfb4c4145bf97de5c88b8ccbb6d

zomars (commented):
Awesome job @leog I cannot approve my own PR but it is approved in spirit 🙏🏽

PR #3494, opened 2022-07-22
Hotfix: Fix Cancellation 500 error if one of the new video apps are installed

hariombalhara:
## What does this PR do?

Fixes 500 errors on cancellation, causing emails to be not sent but cancellation to happen.
_The error occurred when either of Riverside, Whereby, or Around apps was installed. The root cause being the Apps don't define VideoApiAdapter._

**Environment**: Production

## Type of change

## How should this be tested?

- [ ] Add Riverside App and see that whenever cancellating a booking, you would see 500 error. After the fix it would be gone
## Checklist


emrysal (approved):
Looks good to me

PR #3517, opened 2022-07-25
When no location, defaults to cal video

### LINKED ISSUE #2900: When no location is set in an event, where would the meeting occur ?

om-ray:
## What does this PR do?

Currently, if an event type does not have a location set for it, the email that is sent will have an empty `Where` heading with no text that tells you what platform the meeting will be on. This PR solves that issue by defaulting to Cal video instead of just having no location set.

Fixes #2900 

## Type of change


PeerRich:
we'll review today!

om-ray:
@PeerRich Thank you!

packages/app-store/exchange2013calendar/lib/CalendarService.ts
```diff
@@ -33,6 +33,8 @@ import {
   NewCalendarEventType,
 } from "@calcom/types/Calendar";
 
+import { LocationType } from "../../locations";
```
zomars, line 36: This type is not being used

packages/app-store/exchange2016calendar/lib/CalendarService.ts
```diff
@@ -33,6 +33,8 @@ import {
   NewCalendarEventType,
 } from "@calcom/types/Calendar";
 
+import { LocationType } from "../../locations";
```
zomars, line 36:
```suggestion
```

packages/app-store/exchange2013calendar/lib/CalendarService.ts
zomars, line 36:
```suggestion
```

zomars (changes requested):
Thank you for your contribution 🙏 Let's leave the `yarn.lock` out of the PR

zomars:
@om-ray The lock file is still present

om-ray:
@zomars when i remove the lockfile the checks fail, but ill delete it if thats fine

zomars:
@om-ray I don't mean to delete the file. I mean to undo the changes and leave it the same as the `main` branch.

om-ray:
@zomars ahhh that makes more sense, ill do that right now

om-ray:
@zomars had a little trouble with it, but its gone now!

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -68,7 +69,8 @@ export const bookingsRouter = createProtectedRouter()
       newLocation: z.string(),
```
zomars, line 69:
```suggestion
      newLocation: z.string().transform((val) => val || LocationType.Daily),
```

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -68,7 +69,8 @@ export const bookingsRouter = createProtectedRouter()
...
     }),
     async resolve({ ctx, input }) {
-      const { bookingId, newLocation: location } = input;
+      const { bookingId, newLocation: location } =
+      input.newLocation === "" ? { bookingId: input.bookingId, newLocation: LocationType.Daily } : input;
```
zomars, line 73:
```suggestion
      const { bookingId, newLocation: location } = input;
```

zomars (changes requested):
BTW we can achieve the same behavior by using a zod transform

zomars (approved):

PR #3520, opened 2022-07-25
Fix V1 Colors

sean-brydon:
Restores V1 colours into main . As V2 colors got merged into main as per the DS PR. 

Ooops :)


emrysal (approved):
Not only safer but my guess is we don't want to deploy inter yet either.

PR #3586, opened 2022-07-29
Adds env runtime checks

zomars:
## What does this PR do?

TLDR: If you are missing any critical env variables, the app won't build/start.

Idea stolen from @KATT feedback [here](https://github.com/t3-oss/create-t3-app/issues/140)

## Approach
- Define a schema for all env variables and divide them into 2 schema categories - `Client` and `Server`. 
- `Client` schema would always contain NEXT_PUBLIC prefixed env variables which are supposed to be shared with the browser.
- All env variables available from .env and .env.appStore are loaded in process.env and are separately tracked in `allEnv`. `allEnv` contains environment variables defined in those environment files and process.env contain everything else as well. 
- User would run `yarn dev` which would then execute `next.config.mjs` which in turn would validate the Client and Server schemas against `allEnv`.
- Any variable that exists in the environment files but isn't defined in the schema would be logged
- By default, if a variable failing schema validation is encountered, it would be logged with the error message but Next.js build or Next.js dev server will continue. This behaviour can be changed by setting `CRASH_ON_INVALID_VARS` env variable to 1 which would crash the build or dev server.
- Codebase would still access process.env so that it has access to env variables that are not defined in the schema. This is important so that any unknown variable used in Self Hosting is still accessible(with a warning now). 
- Using process.env has an added benefit in that process.env usages are still there for every env variable so that Next.js won't remove that env variable.

TODO:
- [ ] Some recently added variables are not yet defined in schema. Add them as well
Screenshot 2023-01-02 at 6 32 12 PM


PeerRich:
@zomars still relevant?

zomars:
> @zomars still relevant?

Will get back to this eventually...

hariombalhara:
@zomars Pushed some changes and updated the description with the approach that is being used right now.
Would appreciate it, if you can share more details on what else is needed to get this ready.

PeerRich:
dumb question: would this make docker easier or is this unrelated? i keep reading .env issues with docker

hariombalhara:
@PeerRich It might make it easier but I haven't been using docker so can't confirm. 
Docker would certainly require it's own efforts

PeerRich:
@zomars 😆

zomars:
Closing to reopen in another PR

PR #3617, opened 2022-08-01
Adds middleware to get V2 early access

zomars:
## What does this PR do?

Allows admins to opt-in to V2 early access via cookie
## Type of change

## How should this be tested?

As an admin user:

- [ ] Go to `/settings/admin`, you should see V1 page
- [ ] Go to `/api/v2-opt-in` to set the cookie
- [ ] Go to `/settings/admin` again, you should now see the V2 WIP page


apps/web/lib/app-providers.tsx
```diff
@@ -22,6 +22,7 @@ export type AppProps = Omit<NextAppProps, "Component"> & {
   Component: NextAppProps["Component"] & {
     requiresLicense?: boolean;
     isThemeSupported?: boolean | ((arg: { router: NextRouter }) => boolean);
+    getLayout?: (page: React.ReactElement) => ReactNode;
```
zomars, line 25: This will allow us to set reusable layouts at page level

apps/web/middleware.ts
```diff
@@ -1,8 +1,28 @@
...
+  const token = await getToken({ req });
+
+  /**
+   *  TODO: Add logic to only redirect certain pages if `v2` cookie is present.
+   * As of now, only admins can view the new v2 pages
+   **/
+  if (token?.role === "ADMIN" && url.pathname.startsWith(`/settings/admin`)) {
+    // rewrite to the current subdomain under the pages/sites folder
+    url.pathname = `/v2${url.pathname}`;
+  }
```
zomars, line 19: Only admins and admin routes will be served from v2 for now

apps/web/pages/_app.tsx
```diff
@@ -29,6 +29,9 @@ function MyApp(props: AppProps) {
   } else if (router.pathname === "/500") {
     pageStatus = "500";
   }
+  // Use the layout defined at the page level, if available
+  const getLayout = Component.getLayout ?? ((page) => page);
+
```
zomars, line 34: This will allow us to set reusable layouts at page level

apps/web/pages/v2/settings/admin/apps.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+function AdminAppsView() {
+  return (
+    <>
+      <h1>App listing</h1>
+    </>
+  );
+}
+
+AdminAppsView.getLayout = getLayout;
```
zomars, line 11: No more `<Shell>` everywhere

packages/ui/v2/navigation/tabs/VerticalTabItem.tsx
```diff
@@ -71,13 +74,16 @@ const VerticalTabItem: FC<VerticalTabItemProps> = ({ name, href, tabName, info,
...
             <div className="ml-auto self-center">
               <ChevronRight width={20} height={20} className="h-auto w-[20px] stroke-[1.5px]" />
             </div>
-          )}
+          )} */}
         </a>
       </Link>
+      {props.children?.map((child) => (
+        <VerticalTabItem key={child.name} {...child} isChild />
+      ))}
```
zomars, line 86: Recursive nested children support

zomars (commented):
Still WIP but testable

sean-brydon:
I was just testing this out and its super cool great job @zomars  - would love to get this merged so I can somehow move all my work in `feat/V2.0` over to this approach. 

As I've currently done all my work in `apps/web/pages` I'm guessing I'd need to take a copy of pages from main and move all my current work into `apps/web/pages/v2` when it comes round to it?

apps/web/middleware.ts
sean-brydon: This would be super easy to detect if the user has a cookie present right?

packages/ui/v2/navigation/tabs/VerticalTabItem.tsx
sean-brydon: This is super neat - Never thought about this

apps/web/middleware.ts
zomars: yup @sean-brydon that's my next step

apps/web/pages/_app.tsx
PeerRich: super cool, this is new, right?

zomars: New in Next? No. New for our codebase? Yes

apps/web/middleware.ts
```diff
@@ -1,8 +1,25 @@
 import { collectEvents } from "next-collect/server";
+// eslint-disable-next-line @next/next/no-server-import-in-page
+import { NextMiddleware, NextResponse } from "next/server";
 
 import { extendEventData, nextCollectBasicSettings } from "@calcom/lib/telemetry";
 
+const V2_WHITELIST = ["/settings/admin"];
```
sean-brydon, line 7: I will possibly explore a cleaner way to do this - wonder if we can somehow glob the V2 directory ?

sean-brydon (approved):
Awesome job - will convert my PR into this V2 structure today :)

apps/web/middleware.ts
zomars: Don't hive it much of a thought. It will all be gone after V2 launch anyways @sean-brydon

Issue #3410, opened 2022-07-18
Weird bugs in App with Next.js 12.2.x with middleware


hariombalhara:
- https://github.com/vercel/next.js/pull/37716 - This bug seems to be the reason for shallow routing not working anywhere. It caused /booking navigations to stop working. [+Client side bookings navtabs broken](https://twist.com/a/195135/ch/554218/t/3628515/)
- Embed Modal Box doesn’t close
- Deleting a routing Form doesn’t update UI.
- Add a new routing form, doesn’t show the new Form in UI(which is the reason the tests are failing which @Peer R reported.) - 
_Happens only in production mode and not in dev mode._

All the bugs are demoed in here https://www.loom.com/share/dcf41ca221024bdc9de1fe8cd7c41449

hariombalhara:
It also seems to be related dynamic catch-all routes.  I have observed that, when we go to Routing Forms(`/apps/routing_forms/forms`) through client-side navigation,` /apps/routing_forms/forms.json` request is sent after the page is rendered(which is the reason on next visit to the page, it loads correct number of forms)

zomars:
Is this related to our production shenanigans that we patched in 1.8.3?

zomars:
I'm setting up a small repro here: https://github.com/zomars/next-middleware-dynamic-route-static-page-repro

PeerRich:
still open?

hariombalhara:
I think we can close it. We have worked around all such issues AFAIK

PR #3620, opened 2022-08-01
Hotfix: Fix Dialog not closing.

### LINKED ISSUE #3581: Can't Leave Embed Page

### Issue Summary

Could not reproduce, but I was on the embed page, went away from my laptop, re-visited the page and I couldn't leave the embed page unless I refreshed. https://app.birdeatsbug.com/sessions/5b5dB1kfarq9hhv3JPzK5sVx_tTjdF_ApM8xK0b67m6G

### Steps to Reproduce

1. (for example) Went to ...
2. Clicked on...
3. ...

### Technical details

- Node.js version

hariombalhara:
Fixes https://github.com/calcom/cal.com/issues/3581

Root cause is here https://github.com/calcom/cal.com/issues/3410. Shallow routing is not necessarily required and disabling it fixes the issue.


alishaz-polymath:
Great catch @hariombalhara 🙌
When and why did we change it to true?

hariombalhara:
https://github.com/vercel/next.js/pull/37716 This bug seems to be responsible. It had broken after next.js update

PeerRich (approved):

PR #3636, opened 2022-08-01
Fix/google calendar false alert

alannnc:
## What does this PR do?

There is a time when we can't match selectedCalendar.externalId with a specific google credential, so google returns 404 in that case and we were receiving it as an error thus sending broken email. If any other error occurs we should still notify the user. 
E.g. my event has destinationCalendar with my alan@email.com externalId. but it's **not** linked to any credential, we try to create the event on any google_calendar credential from the user for this scenario, but we could receive another credential from other added integration thus throwing 404 on this case.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Link 2 different email google calendars.
- [ ] Event destinationCalendar should be null.
- [ ] Schedule a time and you should no longer receive broken integration email (calendar)

## Checklist


packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -120,36 +120,36 @@ export default class GoogleCalendarService implements Calendar {
       const calendar = google.calendar({
         version: "v3",
       });
+      const selectedCalendar = calEventRaw.destinationCalendar?.externalId
+        ? calEventRaw.destinationCalendar.externalId
+        : "primary";
```
alannnc, line 125: Improving this conditional with a const since it was being calculated twice.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -120,36 +120,36 @@ export default class GoogleCalendarService implements Calendar {
...
-            eventId: event.data.id || "",
-            requestBody: {
-              description: getRichDescription({
-                ...calEventRaw,
-                additionalInformation: { hangoutLink: event.data.hangoutLink || "" },
-              }),
-            },
-          });
-
+          if (event && event.data.id && event.data.hangoutLink) {
```
alannnc, line 139: If we don't receive any event don't trigger any update event api.

packages/core/CalendarManager.ts
```diff
@@ -160,10 +160,17 @@ export const createEvent = async (
...
-    ? await calendar.createEvent(calEvent).catch(async (e) => {
-        await sendBrokenIntegrationEmail(calEvent, "calendar");
-        log.error("createEvent failed", e, calEvent);
+    ? await calendar.createEvent(calEvent).catch(async (error) => {
         success = false;
+        /**
+         * There is a time when selectedCalendar externalId doesn't match witch certain credential
+         * so google returns 404.
+         * */
+        if (error?.code === 404) {
```
alannnc, line 169: Early return for a 404 calendar event api

packages/core/EventManager.ts
```diff
@@ -141,8 +141,8 @@ export default class EventManager {
     const referencesToCreate = results.map((result) => {
       return {
         type: result.type,
-        uid: result.createdEvent?.id.toString() ?? "",
-        meetingId: result.createdEvent?.id.toString(),
+        uid: result.createdEvent?.id?.toString() ?? "",
```
alannnc, line 144: I was receiving error while doing toString on undefined prop.

zomars (approved):
Impressive detective work @alannnc

PR #3640, opened 2022-08-02
DX: run unit & integration tests on pull request

shapel:
## What does this PR do?

I found out that unit & integrations tests were not triggered by pull requests. So by this PR adds one more GitHub action to run tests (unit & integrations) on a pull request.

## How should this be tested?

- [ ] Let's see what checks will be applied to this PR

### DIFF
--- a/.github/workflows/required-checks.yml
+++ b/.github/workflows/required-checks.yml
@@ -26,6 +26,10 @@ jobs:
               paths:
                 - /**/*.tsx
                 - /**/*.ts
+            - job: test
+              paths:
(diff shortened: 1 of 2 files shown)


hariombalhara:
@shapel They do run when e2e tests run

hariombalhara:
@shapel But it might be a good idea to rename E2E tests to reflect this in github workflow.

shapel:
Yeah, the approach that `yarn test-e2e` runs all kinds of tests (like unit & integration & e2e) seems slightly confusing to me.

PR #3661, opened 2022-08-02
Fixes security issue when updating workflows

CarinaWolli:
## What does this PR do?

- Check that user belongs to the event types that are set as active when updating a workflow
- Only delete reminders when deleting a workflow was successful

## Type of change

## Checklist


packages/features/ee/workflows/components/NewWorkflowButton.tsx
```diff
@@ -231,8 +231,8 @@ export function NewWorkflowButton() {
               </div>
             )}
           </>
-          <div className="mt-8 flex flex-row-reverse gap-x-2">
-            <Button type="submit">{t("continue")}</Button>
+          <div className="flex flex-row-reverse mt-8 gap-x-2">
+            <Button type="submit" disabled={createMutation.isLoading}>{t("continue")}</Button>
```
CarinaWolli, line 235: disable button when mutation is loading

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -172,6 +149,32 @@ export const workflowsRouter = createProtectedRouter()
         },
       });
 
+      if (deleted.count === 1) {
```
CarinaWolli, line 152: Delete the scheduled reminders only if the workflow was successfully deleted

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -224,6 +226,40 @@ export const workflowsRouter = createProtectedRouter()
...
+            team: {
+              include: {
+                members: true,
+              }
+            }
+          }
+        })
+
+        if (newEventType && newEventType.userId !== user.id && newEventType?.team?.members.filter(membership => { membership.userId === user.id }).length) {
+          throw new TRPCError({ code: "UNAUTHORIZED" });
```
CarinaWolli, line 258: Throw an error when a user tries to set an event type as active that does not belong to the user or a team where the user is member of

packages/trpc/server/routers/viewer/workflows.tsx
zomars: Do we have onCascade deletion on prisma schema? In theory if we have this, we only need to delete the workflow and all reminders will be deleted without any extra logic

baileypumfleet: I'm fairly certain we did add this for user account deletion

CarinaWolli: @zomars True, we already have onCascade deletion in the Prisma schema defined. Didn't think about that, but we still need to make sure to delete the scheduled reminders on SendGrid and Twilio with deleteScheduledEmailReminder and deleteScheduledSMSReminder. I'll made some changes here: https://github.com/calcom/cal.com/pull/3661/commits/8d4cc753675efd4675af5b527b69e83354a14304

zomars (approved):

PR #3665, opened 2022-08-03
Fix early return when expanding RRULE in CalDav/iCal

### LINKED ISSUE #3589: Availability fetching is bugged for CalDAV
### Issue Summary

So, I found a weird one while trying to debug CalDAV (but I’m not sure if this one is limited to CalDAV or spreads across other integrations as well)

When we book a slot in one of the event types, only that slot within that specific event type is blocked, but other event types still have the slots during that booked time available. Note that the booking was made at the CalDAV calendar as well, yet I think the freebusy data fetched from it was empty.

### Steps to Reproduce

[Loom Video](https://www.loom.com/share/88df05c8a4224172a7b520bef4f4d2f4)

### Technical details

Integration in use: CalDAV

willgittoes:
## What does this PR do?

Recurring events in CalDav and iCal were only using the first instance of the event for availability checking. This fixes that issue. 

(might fix) #3589  (I'm not sure if this is the same issue or not, it looks to be a bit subtler)

No visual changes

**Environment**: Fixed against main.

## Type of change

## How should this be tested?

I looked for unit/integration tests but couldn't find any, so I might be missing some. To repro I created a new google calendar, accessed via CalDav. I created a single weekly recurring event in the calendar, and checked availability for each day in the month view for an event type in Cal.com. I could see that after my fix, the time slot with a clash was correctly not showing up in all 4 weeks, rather than just the first week.

## Checklist

I did all those things, I really think this is a pretty simple fix. Feel free to disagree though! 😊


packages/lib/CalendarService.ts
```diff
@@ -293,7 +293,7 @@ export default abstract class BaseCalendarService implements Calendar {
           currentStart = dayjs(currentEvent.startDate.toJSDate());
 
           if (currentStart.isBetween(start, end) === true) {
-            return events.push({
+            events.push({
```
willgittoes, line 296: The `return` here stopped the loop what was supposed to accumulate multiple instances of this recurring event, instead limiting it to a single instance (the first in-range). I think it was just a typo.

The loop still terminates correctly, since whe while condition checks the date range and has a safety limit of maxIterations.

zomars (approved):
Nice catch @willgittoes Thank you for your contribution 🙏

willgittoes:
Thanks so much for accepting this! It was really easy to get the dev environment set up and working ☺

PR #3667, opened 2022-08-03
refactor: use BookingReference instead of DailyEventReference

### LINKED ISSUE #3389: Cleanup: Migrate DailyEventReference to BookingReference
In our attempts to keep our DB as generic as possible, we should try to clean it up a little bit. A quick win would be updating the Daily video specific model into using `BookingReference`.

shapel:
## What does this PR do?

Fixes #3389

## Type of change

- [ ] Refactoring with migrating and dropping DailyEventReference table

## How should this be tested?

- [ ] Test that exists daily meeting work as expected
- [ ] Test that new daily meeting work as expected

### DIFF
--- a/packages/prisma/migrations/20220803090845_migrate_daily_event_reference_to_booking_reference/migration.sql
+++ b/packages/prisma/migrations/20220803090845_migrate_daily_event_reference_to_booking_reference/migration.sql
@@ -1,0 +1,4 @@
+UPDATE "BookingReference"
+SET "meetingUrl" = "dailyurl", "meetingPassword" = "dailytoken"
+FROM "DailyEventReference"
+WHERE "DailyEventReference"."bookingId" = "BookingReference"."bookingId" AND "BookingReference"."type" = 'daily_video'
--- a/apps/web/pages/video/[uid].tsx
+++ b/apps/web/pages/video/[uid].tsx
@@ -3,7 +3,6 @@ import { NextPageContext } from "next";
 import { getSession } from "next-auth/react";
 import Head from "next/head";
 import Link from "next/link";
-import { useRouter } from "next/router";
 import { useEffect } from "react";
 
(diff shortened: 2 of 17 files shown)


packages/prisma/migrations/20220803090845_migrate_daily_event_reference_to_booking_reference/migration.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE "BookingReference"
+SET "meetingUrl" = "dailyurl", "meetingPassword" = "dailytoken"
+FROM "DailyEventReference"
+WHERE "DailyEventReference"."bookingId" = "BookingReference"."bookingId" AND "BookingReference"."type" = 'daily_video'
```
zomars, line 4: Will this migration move everything from `DailyEventReference` to `BookingReference` I'm having a hard time understanding what this query does exactly.

shapel: From my understanding of how the system deals with booking and observation from DB, the system has two records for each daily video call booking, one inside a "BookingReference" with "type" = 'daily_video' and "meetingUl" eq `/video/{uid}` and one in "DailyEventReference" with real daily call url and token.

So by this migration, we will update all "BookingReference" records with "type" = 'daily_video' and related to the same booking Id with info from "DailyEventReference".

If, for historical reasons, the system could have "DailyEventReference" records without identical "BookingReference" then we need to write one more SQL query.

zomars:
cc @agustif let's check if this PR could break the api repo

shapel:
Hey @zomars, @agustif  
I've rebased my branch on the last main changes. 
Please feel free to provide feedback on this PR.

zomars (approved):
Made some improvements to daily video adapter, but overall very impressive work. Thank you for your contribution 🙏 

Computer Reaction GIF by GitHub

PR #3669, opened 2022-08-03
Embed Test: Verifies that app remains hidden till link is ready

### LINKED ISSUE #3278: Playwright Test Required: Ensure that embedded body remains display=none till __iframeReady event is fired.
#3277 
https://www.loom.com/share/7fe36fcc525742dca6b07f783185b54d
This is to ensure that embedded cal is not visible till it's ready to be visible. Possible issues if it fails,
- A flicker of theme might occur
- Content might load with lesser height initially and then load with full height.

hariombalhara:
Add test to verify that webapp remains hidden till link is considered ready
Closes #3278


pumfleet (approved):
Code looks good to me

PR #3677, opened 2022-08-03
Setup Wizard Not Pulling Timezone

### LINKED ISSUE #3654: Setup Wizard Not Pulling Timezone

### Issue Summary

When a user joins Cal.com, they go through a set-up wizard, part of that set-up wizard is Step 1 selecting your timezone. Currently, it doesn't auto-pull my timezone based on my location (no VPN running) 
CleanShot 2022-08-02 at 09 41 02@2x

### Steps to Reproduce

1. Create a new account
2. Navigate to the first step and check to see if your timezone pulls through or not

Solution: Pull through the correct timezone (HST)

### Technical details

- Brave browser, Macbook pro MacOS Monterey

gitstart:
## What does this PR do?

Fixes #3654 

## Demo
https://www.loom.com/share/f9ae6d74a63e4a30bca2e4f44b0ae4ad

## Type of change

## How should this be tested?

- Create a new account
- Navigate to the first step and check to see if your timezone pulls through or not

## Checklist


apps/web/pages/getting-started.tsx
```diff
@@ -377,12 +377,12 @@ export default function Onboarding(props: inferSSRProps<typeof getServerSideProp
                   </label>
                   <p className="text-sm leading-tight text-gray-500 dark:text-white">
                     {t("current_time")}:&nbsp;
-                    <span className="text-black">{dayjs().tz(selectedTimeZone).format("LT")}</span>
+                    <span className="text-black">{dayjs().tz(dayjs.tz.guess()).format("LT")}</span>
```
baileypumfleet, line 380: The whole point of having this here is for the user to confirm that their selected timezone matches up with the time on their computer, so this defeats the whole purpose.

apps/web/pages/getting-started.tsx
```diff
@@ -377,12 +377,12 @@ export default function Onboarding(props: inferSSRProps<typeof getServerSideProp
...
                   </p>
                 </section>
                 <TimezoneSelect
                   id="timeZone"
-                  value={selectedTimeZone}
+                  value={dayjs.tz.guess()}
```
baileypumfleet, line 385: Also this doesn't make sense as this change would lock the timezone select to only show the guessed timezone instead of what's been selected by the user

pumfleet (changes requested):
Really not sure of the meaning behind these changes, could you please confirm this on each of my comments?

apps/web/pages/getting-started.tsx
gitstart: Sorry about this @baileypumfleet. PR has been updated now, kindly look into it again

pumfleet (dismissed):
I still don't understand the purpose with this new change either; the old one already guesses the user's timezone if a default is not set, so that should work just fine

gitstart:
> I still don't understand the purpose with this new change either; the old one already guesses the user's timezone if a default is not set, so that should work just fine

@baileypumfleet
Yeah, that should have worked if the default was not set. But we have [this](#3654 ) issue because of default value being set to `Europe/London`. Since there's always a default value, it does not move on to `dayjs.tz.guess()`.

apps/web/pages/getting-started.tsx
```diff
@@ -150,7 +150,7 @@ export default function Onboarding(props: inferSSRProps<typeof getServerSideProp
   const bioRef = useRef<HTMLInputElement>(null);
   /** End Name */
   /** TimeZone */
-  const [selectedTimeZone, setSelectedTimeZone] = useState(user.timeZone ?? dayjs.tz.guess());
+  const [selectedTimeZone, setSelectedTimeZone] = useState(dayjs.tz.guess());
```
hariombalhara, line 153: The actual problem is this https://github.com/calcom/cal.com/blob/main/packages/prisma/schema.prisma#L138
There is a default timezone set for all users. So, `user.timeZone` is always set even if it is not set by user.

`user.timeZone` check is actually required because user might abandon onboarding in between after selecting his timezone. This allows him to not redo this step of setting the timezone.

Because we need to first fix the default in schema.prisma, it looks like the right fix to me till then(plus for users who have their account created already and are on onboarding step). Because it is a bigger problem to have wrong timezone shown by default to new users.

cc @zomars I don't think there should be a problem in removing the default timezone in schema.prisma.

hariombalhara (approved):

PR #3688, opened 2022-08-04
Event description now supports markdown formatting

### LINKED ISSUE #3348: Formatting of description missing (bold, italic, new line)

right now I cannot provide structured information to register a calendar entry. Everything goes into one long line with arbitrary line breaks.

on event types, there is a description field.
Here it is missing to be able to use common formatting:

- bold
- italic
- underlined

and the most painfull issue

- new line is missing and cannot be entered

om-ray:
## What does this PR do?

This PR adds the ability to add markdown to event descriptions.

### Dependencies:
 - DOMPurify (sanitizes inputs to prevent XSS attacks)
 - Markdown-it (parses the text for markdown)

Fixes #3348 

## Type of change


package.json
```diff
@@ -60,6 +60,10 @@
     "prettier": "^2.5.1"
   },
   "dependencies": {
+    "@types/dompurify": "^2.3.3",
+    "@types/markdown-it": "^12.2.3",
+    "dompurify": "^2.3.10",
+    "markdown-it": "^13.0.1",
```
PeerRich, line 66: i dont believe we wanna have the dependencies in this package json -- instead we should abstract this into a a markdown component, thoughts @zomars ?

PeerRich: we also need to test those packages for health

zomars: Agreed @PeerRich we should check for weight, health, maintainability and overall complexity.

zomars (changes requested):
The more I think about this, the more I'm leaning towards having this sanitization [Prisma middleware level](https://www.prisma.io/docs/concepts/components/prisma-client/middleware/session-data-middleware). That way we can "set and forget" it and make sure the `description` fields is always properly parsed and sanitized.

apps/web/pages/team/[slug].tsx
```diff
@@ -35,6 +36,8 @@ function TeamPage({ team }: TeamPageProps) {
   const telemetry = useTelemetry();
   const router = useRouter();
 
+  const md = new MarkdownIt();
```
Comment (zomars) [inline apps/web/pages/team/[slug].tsx:39]: This will create a new Markdown instance on each render. This is a no-no for performance.

apps/web/pages/team/[slug].tsx
```diff
@@ -51,6 +54,7 @@ function TeamPage({ team }: TeamPageProps) {
             "hover:border-brand group relative rounded-sm border border-neutral-200   dark:border-neutral-700 dark:bg-neutral-800 dark:hover:border-neutral-600",
             isEmbed ? "" : "bg-white hover:bg-gray-50"
           )}>
+          {(type.description = md.render(`${type.description}`))}
```
Comment (zomars) [inline apps/web/pages/team/[slug].tsx:57]: Making an assignment inside a render function is not considered a optimal practice

apps/web/pages/event-types/[type].tsx
```diff
@@ -1469,7 +1469,7 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
                                           />
                                           <select
                                             id=""
-                                            className="block w-full rounded-sm border-gray-300 py-2 pl-3 pr-10 text-base text-sm focus:outline-none"
+                                            className="block w-full rounded-sm border-gray-300 py-2 pl-3 pr-10 text-sm focus:outline-none"
```
Comment (zomars) [inline apps/web/pages/event-types/[type].tsx:1472]: Why is this change needed?

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -1,6 +1,7 @@
 // Get router variables
 import { EventType } from "@prisma/client";
 import * as Collapsible from "@radix-ui/react-collapsible";
+import DOMPurify from "dompurify";
```
zomars, line 4: On further inspection, `dompurify` isn't needed if we use `markdown-it` in strict mode.

zomars:
Made some adjustments:
- Sanitizes and parses markdown at prisma middleware
- Instead of replacing the eventType description, it adds a new `descriptionAsSafeHTML` field.
- Adds a `EventTypeDescriptionSafeHTML` to render that specific untyped field
- Only renders markdown on public facing pages, leaves the internal description untouched for now
- Removes irrelevant dompurify dependency
- Moved `markdown-it` dependency to `apps/web` where it belongs.

package.json
zomars: @PeerRich removed `dompurify`, and we're now using `markdown-it` only on the server so client bundle size isn't affected.

apps/web/pages/event-types/[type].tsx
Comment (om-ray) [reply apps/web/pages/event-types/[type].tsx:1472]: @zomars text-base and text-sm both have the same rules and it was giving me a warning and I didn't see any harm in removing it

apps/web/pages/event-types/index.tsx
```diff
@@ -68,6 +69,11 @@ const Item = ({
   connectedCalendars: ConnectedCalendars[] | undefined;
 }) => {
   const { t } = useLocale();
+  const md = new MarkdownIt();
+
+  if (type.description !== null) {
+    type.description = md.render(`${type.description}`);
+  }
```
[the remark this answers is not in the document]
om-ray: @zomars ahhhh, that makes this whole thing a lot easier than going to every place `eventTypes.description` is used!

zomars (approved):
Although it got an almost total rewrite. Thank you for your contribution 🙏 sometimes a push is needed to get things done 👍🏽

Issue #3680, opened 2022-08-03
[CAL-1424] Disable standard confirmation email in even type settings


CarinaWolli:
### Is your proposal related to a problem?

No

### Describe the solution you'd like

Possibility to disable the standard confirmation email in event type settings.

### Additional context

Requested by a user. He wants to set up his own confirmation email with workflows. 

<sub>[CAL-1424](https://linear.app/calcom/issue/CAL-1424/disable-standard-confirmation-email-in-even-type-settings)</sub>

PeerRich:
<img width="858" alt="CleanShot 2022-08-04 at 12 32 53@2x" src="https://user-images.githubusercontent.com/8019099/182826519-dc0635fd-014a-4d5b-beee-c32e43798e86.png">
could be added as a checkbox in the new advanced settings:

dezerb:
@CarinaWolli @PeerRich 
I prepared some proposal based on current design. There is additional flag saved od eventType object and according to them mail is or not sent to the attendee.  By default it's true.
https://github.com/calcom/cal.com/pull/3695/files

![image](https://user-images.githubusercontent.com/14241061/182870005-8beaf391-8205-444b-8fd9-46d6f35fdecb.png)

PeerRich:
> @CarinaWolli @PeerRich 
> 
> I prepared some proposal based on current design. There is additional flag saved od eventType object and according to them mail is or not sent to the attendee.  By default it's true.
> 
> https://github.com/calcom/cal.com/pull/3695/files
> 
> 
> 
> ![image](https://user-images.githubusercontent.com/14241061/182870005-8beaf391-8205-444b-8fd9-46d6f35fdecb.png)
> 
> 

i think this should be negated:

- [ ] "Don't send email confirmation"

with an explanation text

"Disable email confirmation in case you want to send your own custom emails"

PeerRich:
we'll need to think harder on this, its not trivial and can cause a lot of issues for people who mistakenly check that box. no checkbox should ever turn of core parts of the product. that'll be a debugging nightmare for customer support. 

we will likely add this checkbox in workflows whenever you create an email alternative but haven't found a good way for the UI yet. 

we will revisit in october

CarinaWolli:
The feature to disable our standard confirmation email got requested again (the user wants to send a customized confirmation email with workflows)

@Jaibles Should we add that? If yes, how? as @PeerRich mentioned, we want to avoid that users turn off emails by mistake

[...]

PR #3695, opened 2022-08-04
add checkbox for send confirmation after event booking

dezerb:
## What does this PR do?
Add possibility to disable standard mail notification after event booking.

Fixes # (issue)
https://github.com/calcom/cal.com/issues/3680

By default it is unchecked
Screenshot 2022-08-05 at 2 15 05 PM

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B


PeerRich:
as I mentioned before, I think it would be better if it is negated and default false

"Don't send confirmation"

PeerRich:
it's better to have no check box checked initially, if that makes sense?

dezerb:
@PeerRich it's make perfect sense, I will make adjustment ;)

dezerb:
updated
https://www.loom.com/share/93b27bbfa00d45888e4de46675d823c0

hariombalhara:
@dezerb Thanks for your contribution 🙏

This is the behaviour right now @PeerRich.

Once I select the option(which is by default unchecked)
- On a booking, no email is sent
- On cancellation, emails are sent as usual
- On rescheduling, no email is sent.
- On Request Reschedule, emails are sent as usual.

I am a little confused as to the usage of this option. What if the user wants to set up his own cancellation email? Do we give a separate checkbox then?

Also, because confirmation emails are important to be sent, it should somehow be made more clear to user(what if he checks this option and forgets to setup his own emails). Maybe we can integrate it with Workflows more seamlessly, automatically detecting that confirmation/cancellation emails are setup so that it is foolproof.

PeerRich:
> @dezerb Thanks for your contribution 🙏
> 
> This is the behaviour right now @PeerRich.
> 
> Once I select the option(which is by default unchecked)
> 
> * On a booking, no email is sent
> * On cancellation, emails are sent as usual
> * On rescheduling, no email is sent.
> * On Request Reschedule, emails are sent as usual.
> 
> I am a little confused as to the usage of this option. What if the user wants to set up his own cancellation email? Do we give a separate checkbox then?
> 
> Also, because confirmation emails are important to be sent, it should somehow be made more clear to user(what if he checks this option and forgets to setup his own emails). Maybe we can integrate it with Workflows more seamlessly, automatically detecting that custom confirmation/cancellation emails are setup and then this option can be automatically checked or allowed to be checked.

yeah, now that i think about it, i don't think you should be able to disable such core functionality.

i think a better alternative would be to customise the email which then overwrites the old confirmation mail

this feature has a lot of risks of unintended misbehavior

CarinaWolli:
> yeah, now that i think about it, i don't think you should be able to disable such core functionality.
> 
> i think a better alternative would be to customise the email which then overwrites the old confirmation mail
> 
> this feature has a lot of risks of unintended misbehavior

@PeerRich  What about putting it into workflows? So once a user creates a workflow with the trigger 'when event is created' and the action 'Send email to ...' we show a checkbox that asks if the user wants to disable the standard confirmation email?

dezerb:
@hariombalhara @PeerRich  @CarinaWolli 
in last iteration I introduced some new approach. 
While confirming the booking the backend will check if there is a suitable workflow and if so, it will skip the default notification. Thus, no new checkbox is required, assuming if someone has created a workflow, he does not need a standard confirmation.
Please take a look and comment.

hariombalhara:
I would leave it to @CarinaWolli to decide on the approach as she owns workflows.

PeerRich:
> I would leave it to @CarinaWolli to decide on the approach as she owns workflows.

I think its a good idea to have the checkbox in workflows to replace the default mail. that's where you would expect it

dezerb:
@PeerRich ok I will take care about that and introduce the new checkbox. Should it be b default checked?

PeerRich:
> @PeerRich ok I will take care about that and introduce the new checkbox. Should it be b default checked?

default: off

dezerb:
@PeerRich  updated
https://www.loom.com/share/75f63bf46b85438b8dccfd0a1b312a22

CarinaWolli:
@Jaibles Can you take a look at this? https://www.loom.com/share/75f63bf46b85438b8dccfd0a1b312a22
I think the should adjust the text and design a little bit

zomars (commented):
Code LGTM. Delegating my approval to @CarinaWolli since she owns workflows and @Jaibles since it's a core product change

CarinaWolli (commented):
Code looks good to me and it works as expected 👍  But I think we need to rethink the design. Currently, we have an issue when we have the same action more than once:

Screenshot 2022-08-12 at 20 22 35
 
It will also be hard to keep track for what event types the confirmation email is disabled (especially if you have several active workflows). So thinking of the v2 design, it would probably be best to somehow add this to the event-type settings in the workflow tab. @Jaibles What do you think?

PeerRich:
closing for now, moving to 2.1

PR #3718, opened 2022-08-05
Cleans up cli generator and enables prettier formatting

zomars:
## What does this PR do?

Even if this file is auto-generated, it is a file that devs would need to keep referencing to; So good legibility and formatting is still important IMO.

## Type of change

## How should this be tested?

- [ ] `yarn app-store:build`


hariombalhara (approved):
Totally agree with the idea behind the change. Thanks @zomars for this.

PR #3726, opened 2022-08-07
Adds rescheduling dropdown to past unconfirmed bookings

### LINKED ISSUE #3641: cannot reschedule unconfirmed events in the past
### Issue Summary

If a booking is made which needs confirmation, but is not confirmed before the time of the event passes, it cannot be rescheduled.

### Steps to Reproduce

1. Configure an event type which requires confirmation prior to booking
2. Someone requests a booking using this event type
3. Ignore request until time of booking passes
4. Observe that booking appears in the `Past` tab, but is permanently stuck there and cannot be rescheduled:

image
 
### Expected behaviour

I would expect to have options which are similar to the ones available for bookings in the `Upcoming` tab:

image

The motivation is that if someone has put in the effort to try and book my time, and I didn't respond to their request quick enough, I want to make it up to them by working with them to find a new time.  This is much preferable to requiring them to start the booking process over from scratch, which would have disadvantages:

1. It would require me to message them somehow outside of cal.com to explain what happened and ask them to rebook.
2. The consequent rebooking would create a second set of unrelated emails in both parties' inboxes.

dallenpyrah:
## What does this PR do?

This PR adds a new feature to any bookings that have not been confirmed and have exceeded the meeting date resulting in them moving to the "past" tab. This feature allows users the ability to reschedule their past bookings that were unconfirmed which in turn no longer requires someone to message the other party somehow to explain what happened and ask them to rebook. 
It also stops the booking from creating a second set of unrelated emails in both parties' inboxes.

Fixes #3641

https://www.loom.com/share/de23b8512736497aa85f582975d0eebe

## Type of change

## How should this be tested?

- [X] Create a booking event type that requires confirmation. Schedule a booking with your local user. Change the booking time to a past date in the Prisma Studio booking table. Notice that it is unconfirmed but it now has the cancel and edit rebooking button/dropdown. 
- [X] Reschedule the booking in the past tab, once you confirm it the booking should now reappear in the upcoming tab. It will still be unconfirmed and should allow you the ability to confirm the booking again. 

## Checklist


apps/web/components/booking/BookingListItem.tsx
```diff
@@ -392,6 +392,9 @@ function BookingListItem(booking: BookingItemProps) {
               {isRejected && <div className="text-sm text-gray-500">{t("rejected")}</div>}
             </>
           ) : null}
+          {isPending ? (
```
hariombalhara, line 395: This can potentially cause duplicate Actions in UI.
**Theoretically:** For a booking that is in future and not accepted.
- `isUpcoming` is true
-` isCancelled` if false
- `isPending` is true

hariombalhara (changes requested):

apps/web/components/booking/BookingListItem.tsx
dallenpyrah: Good point, I'll take a look and see what I can do to fix that.

dallenpyrah: Added a "isPast" conditional to go along with "isPending". This should stop duplicate actions from occurring.

dallenpyrah: Did some manual tests to make sure that the theoretical bug @hariombalhara brought up does not occur, this action only shows up on unconfirmed bookings in the past.

apps/web/components/booking/BookingListItem.tsx
```diff
@@ -396,6 +397,7 @@ function BookingListItem(booking: BookingItemProps) {
               {isRejected && <div className="text-sm text-gray-500">{t("rejected")}</div>}
             </>
           ) : null}
+          {isPast && isPending ? <>{!isConfirmed && <TableActions actions={bookedActions} />}</> : null}
```
zomars, line 400:
```suggestion
          {isPast && isPending && !isConfirmed ? <TableActions actions={bookedActions} /> : null}
```

Wouldn't this work as well?

zlwaterfield (approved):
Looks good!

@hariombalhara do you want to take another quick look to make sure the logic makes sense to you?

hariombalhara (approved):

PR #3728, opened 2022-08-07
added ping.gg app

PeerRich:


hariombalhara (approved):
Fixed app not showing issue. Variant for video apps is supposed to be `conferencing` now. Fixed it in cli as well.

There is a lot of glue code to get a video app working. I would work on extracting that out and simplifying that in a followup PR
Tracking Issue is created here https://github.com/calcom/cal.com/issues/3737

PR #3742, opened 2022-08-08
Same-day schedules created invalid workingHours

### LINKED ISSUE #3575: Team Availability not Pulling Through

### Issue Summary

For the collective links, I’ve noticed some issues around availability not being pulled through. Most recently with this link: https://cal.com/team/cal/sales (I checked the link as well and my availability is not being pulled)

913d-4270-4df3-816219199b29 (1)

Someone was able to book a time outside working hour my availability:
913d-4270-4df3-816219199b29

### Steps to Reproduce

1. (for example) Went to ...
2. Clicked on...
3. ...

### Technical details

- Node.js version

emrysal:
## What does this PR do?

Fixes #3575 

https://www.loom.com/share/80de68e0ac684c428f52017241d85063

## Type of change

### DIFF
--- a/packages/trpc/server/routers/viewer/slots.tsx
+++ b/packages/trpc/server/routers/viewer/slots.tsx
@@ -209,19 +209,24 @@ export async function getSchedule(
     (currentValue: ValuesType<typeof userSchedules>["workingHours"], s) => {
       // Collective needs to be exclusive of overlap throughout - others inclusive.
       if (eventType.schedulingType === SchedulingType.COLLECTIVE) {
-        s.workingHours.forEach((workingHour) => {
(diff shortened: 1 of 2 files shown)


leog (changes requested):
Code looks good. Just pointed out some scenario I couldn't make it work, you are working on it already. Just wanted to reflect that here 👍

zomars (approved):
This is a great improvement but there's also a new flaw that you may be aware of. If there's a collective event where the attendees don't have any intersection in availability It should display as "Unavailable" but right now it just displays the availability of the first attendee.

Is still way better than we currently have. We may follow up this in another issue.

Great work as always @emrysal 🙏🏽

PR #3743, opened 2022-08-08
Prevent saving error responses from Zoom to user credentials

joeauyeung:
## What does this PR do?

The purpose of this PR is to prevent error responses from Zoom to be saved under a user's credentials. This would wipe out the refresh token which prevents a user from ever getting a valid token again.

In the future, I want to do an audit of our all our apps to make sure this isn't happening.

Fixes # (issue)

## Type of change

## How should this be tested?

- Add Zoom to the account
- In the DB, find the zoom credential and delete a few characters from the refresh token and change the expiration date
- Try and create an event with Zoom as the location, you should receive an error but the credential's should still exist


packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -45,18 +45,19 @@ export const zoomMeetingsSchema = z.object({
...
-  expiry_date: z.number(),
+  scope: z.string().regex(new RegExp("meeting:write")).optional(),
+  expiry_date: z.number().or(z.null()),
   expires_in: z.number().optional(), // deprecated, purely for backwards compatibility; superseeded by expiry_date.
-  token_type: z.literal("bearer"),
-  access_token: z.string(),
-  refresh_token: z.string(),
+  token_type: z.literal("bearer").optional(),
+  access_token: z.string().optional(),
+  refresh_token: z.string().optional(),
```
joeauyeung, line 53: Schema changes are due to the error response from Zoom not containing these fields

leog (approved):
Testing as per the given steps works, code looks good 👌

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -77,21 +77,25 @@ const zoomAuth = (credential: Credential) => {
...
-            id: credential.id,
-          },
-          data: {
-            key: responseBody,
-          },
-        });
-        credentialKey.expiry_date = responseBody.expiry_date;
-        credentialKey.access_token = responseBody.access_token;
-        return credentialKey.access_token;
+        if (responseBody.refresh_token) {
```
zomars, line 80:
```suggestion
        if (!responseBody.refresh_token)  {
       // Reject promise here
```

https://app.gitbook.com/o/6snd8PyPYMhg0wUw6CeQ/s/VXRprBTuMlihk37NQgUU/engineering/best-practices/prefer-early-returns

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
zomars: Let's make a `z.union` so we can receive two schemas. The one where the zoom response is successful and the schema where we don't have `scope` and other fields instead of optional everything

zomars (changes requested):
Code quality nitpicks 🙏🏽

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -44,14 +44,23 @@ export const zoomMeetingsSchema = z.object({
...
+    access_token: z.string(),
+    refresh_token: z.string(),
+  }),
+  // Error API response
+  z.object({
+    error: z.string(),
+    reason: z.string(),
+    expire_date: z.union([z.number(), z.null()]),
+  }),
+]);
```
joeauyeung, line 63: @zomars you're right. Union is much cleaner

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -77,6 +86,7 @@ const zoomAuth = (credential: Credential) => {
     })
       .then(handleErrorsJson)
       .then(async (responseBody) => {
+        if (!responseBody.refresh_token) Promise.reject(new Error("Invalid credentials"));
```
joeauyeung, line 89: Early return

joeauyeung (commented):
Addressed feedback

zomars (approved):

zlwaterfield:
@joeauyeung @zomars should we get this merged?

zomars:
@zlwaterfield @joeauyeung there's a type error to be fixed first:
image

zlwaterfield:
@joeauyeung can you fix that type error?

PR #1292, opened 2021-12-10
Adds preact for production builds

zomars:
### Before 

image
image

### After

image
image

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code and corrected any misspellings
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #3754, opened 2022-08-09
Replace React in client production build with Preact

zomars:
## What does this PR do?

take 2 of #1292

## Type of change

## How should this be tested?

- [ ] `yarn build` and `yarn start`
- [ ] `yarn test-e2e`

## Checklist


Udit-takkar:
@zomars are there any more benefits of preact other than reduced bundle size ?

zomars:
> @zomars are there any more benefits of preact other than reduced bundle size ?

Other than reducing client side bundle size I'm not pretty sure. But every KB we can save without sacrificing functionality and performance is a win if you ask me.

zlwaterfield:
@zomars do you still want to implement this change? Preact is a pretty strong alternative (https://preactjs.com/guide/v8/differences-to-react/) but my main concern is versioning. As Cal keeps the Next.js version up to date, they may bump the React version under the hood and I'm thinking this will be a cause of concern around keeping the Preact version inline. Do you think the space saving is worth it? IMO it is not.

zomars:
I would worry too much about versioning as this change is easily reversible. Even [Next.js maintainers use this technique all the time](https://github.com/leerob/leerob.io/commit/74a3ebb14b81f785ea96f9c73fbe4f6c2463679c).

alannnc:
@zomars PR has been updated in case anyone wanna try it out. It worked for me

alannnc (approved):

PeerRich:
how much do we save? anyone has benchmarks?

zomars:
Closing since the gain/effort to debug tests is not worth right now.

PR #3760, opened 2022-08-10
Fixes issues with video location links from apps

CarinaWolli:
## What does this PR do?

Fixes the following issues with the video links of Riverside, Around, Whereby and Ping.gg. 

- Updating the location of a booking did not work 
- Around link was not showing on availability page
- No line breaks when a link was too long on the availability page

## Type of change

## How should this be tested?

- Install Riverside, Around, Whereby and Ping.gg
- Try updating the location of a booking with all four location types
- Check if current location changed to the the specified link

## Checklist


apps/web/components/booking/BookingListItem.tsx
```diff
@@ -187,7 +187,11 @@ function BookingListItem(booking: BookingItemProps) {
     if (
       newLocationType === LocationType.InPerson ||
       newLocationType === LocationType.Link ||
-      newLocationType === LocationType.UserPhone
+      newLocationType === LocationType.UserPhone ||
+      newLocationType === LocationType.Riverside ||
+      newLocationType === LocationType.Around ||
+      newLocationType === LocationType.Whereby ||
+      newLocationType === LocationType.Ping
```
CarinaWolli, line 194: To save the video links and not the location type

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -54,6 +54,7 @@ export const locationKeyToString = (location: LocationObject, t: TFunction) => {
     case LocationType.Link:
     case LocationType.Ping:
     case LocationType.Riverside:
+    case LocationType.Around:
```
CarinaWolli, line 57: That's why around link was not showing on availability page

packages/core/EventManager.ts
```diff
@@ -172,7 +172,10 @@ export default class EventManager {
     }
 
     // Update the calendar event with the proper video call data
-    results.push(...(await this.updateAllCalendarEvents(evt, booking)));
+    const calendarReference = booking.references.filter((reference) => reference.type.includes("_calendar"))[0];
```
zomars, line 175: If we're using the first one we can just use `find` then. It will return the first item found.
```suggestion
    const calendarReference = booking.references.find((reference) => reference.type.includes("_calendar"));
```

zomars (approved):

PR #3805, opened 2022-08-11
fix: stories not appearing in storybook

hussamkhatib:
## What does this PR do?

Many of the stories are not visible in the storybook. 
This PR fixes almost all of them.
This was because
- Import/Exports 
Coinsider the exports 
```
// ui/v2/index.tsx
export * from "./Card";
export { default as Badge } from "./Badge";
```
And imports in their respective stories
```
import {  Card } from "@calcom/ui/v2/Card"; // works
import { Badge } from "@calcom/ui/v2";  // does not work 
import { Badge } from "@calcom/ui/v2/Badge";  //works
```
- Component named with lower case 
Example from `banner.tsx` to `Banner.tsx` 
--- 
Other changes 
- `Select.stories.tsx`  was having a type error and UI dint behave properly as well. This [video](https://www.loom.com/share/f75d09e71f094e28bef5cbd85adecc00) shows how it is now.
- type error fix in `Date.stories.tsx` 

- There is no change in `yarn.lock`, yet this is included in the PR. I have made sure there is no whitespace difference too. 

**Environment**: Developement


zomars (changes requested):
Let's fix the index files instead on modifying all stories. The desired way to consume it is the one that isn't working right now. Also tagging @sean-brydon who has more context on this.

zlwaterfield:
@sean-brydon thoughts on how to move this PR forward? Is it worth fixing? Or should we close it?

PR #3820, opened 2022-08-12
fix: rate limit auth

### LINKED ISSUE #3819: Rate limiting
### Issue Summary
Rate limit needed  #3820 will fix this

agustif:
## What does this PR do?
Adds rate limiting. (10 tries per minute) to our login./auth process, this applies both to email/password and 2fa totpCode auth. It uses user.email as key for the cache to check if more than the limit has been requested

Fixes #3819

## Type of change

## How should this be tested?
- [ ] Try to login, input a bad password 10 times,
- [ ] or try to login with a 2fa enabled account, try to input a random bad 2fa code more than ten times
image


packages/lib/rateLimit.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import LRU from "lru-cache";
```
zomars, line 2: What's the benefit of using this over `memory-cache` ?

zomars (commented):
LGTM. Left a Q there 🙏🏽

packages/lib/rateLimit.ts
agustif: I just ported the work from https://github.com/calcom/website/pull/70 

I can replace it for memory-cache =) good catch!

we can look into rate limiting requests later with lru cache or similar for other api endpoints, maybe even a rateLimiting middlware/handler

agustif: Looking at google because I had no idea but this is the difference over both:
`**Least-recently-used (LRU) evicts the key-value used-the-least when the cache is full and it needs to add a value. Whereas a MemoryCache evicts the oldest key-values, or those past their 'use-by-date' if they happen to have one.Sep 19, 2015**`

zomars: I'm just trying to keep dependencies as low as possible. Feel free to migrate either of those.

pumfleet (commented):
Code looks good to me, needs testing though

yarn.lock
```diff
@@ -17351,7 +17438,7 @@ next-api-middleware@^1.0.1:
   dependencies:
     debug "^4.3.2"
 
-next-auth@^4.10.3:
+next-auth@^4.10.3, next-auth@^4.3.3:
```
zomars, line 17441: We have a version mismatch here @agustif let's sync all submodules before committing the `yarn.lock` file. Or since we're not adding more dependencies, discard it altogether

agustif: yes, removed all yarn.lock changes since we're not adding a new dep `lru-cache` and its types anymore, good catch thanks zomars

packages/lib/rateLimit.ts
agustif: Fixed and using memory-cache now!

zomars (approved):

zomars (changes requested):
Onboarding tests are failing @agustif let's check those first 🙏🏽

zomars (changes requested):
It seems like this PR has a major flaw, I could be wrong tho. It seems like this is limiting the endpoint calls globally instead of per users/requester.

agustif:
> It seems like this PR has a major flaw, I could be wrong tho. It seems like this is limiting the endpoint calls globally instead of per users/requester.

I will test it out more, I'm passing user.email as the -token- to be tracked, so I dont know something could be wrong with the rateLimit util logic tho, thanks for spotting @zomars

zomars:
> > It seems like this PR has a major flaw, I could be wrong tho. It seems like this is limiting the endpoint calls globally instead of per users/requester.
> 
> I will test it out more, I'm passing user.email as the -token- to be tracked, so I dont know something could be wrong with the rateLimit util logic tho, thanks for spotting @zomars

Also I think this is why E2E is failing since logins occur way to fast and they hit the logins per hour limit

agustif:
> > > It seems like this PR has a major flaw, I could be wrong tho. It seems like this is limiting the endpoint calls globally instead of per users/requester.
> > 
> > 
> > I will test it out more, I'm passing user.email as the -token- to be tracked, so I dont know something could be wrong with the rateLimit util logic tho, thanks for spotting Omar López
> 
> Also I think this is why E2E is failing since logins occur way to fast and they hit the logins per hour limit

Should we do more than 10/hour btw? I can sometimes be stubborn forget a password and see myself trying 20 or 30 times... until blocked

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -100,6 +101,16 @@ const providers: Provider[] = [
         }
       }
 
+      const limiter = rateLimit({
+        interval: 60 * 1000, // 1 minute
+        uniqueTokenPerInterval: 100, // Max 10 login retries per minute
```
Comment (zomars) [inline apps/web/pages/api/auth/[...nextauth].tsx:106]: Why do we need `uniqueTokenPerInterval` here? That's a LRU specific thing for setting max cache size. This doesn't apply to `memory-cache`.

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -100,6 +101,16 @@ const providers: Provider[] = [
...
+      });
+      const { isRateLimited } = await limiter.check(100, user.username); // 10 requests per minute
```
Comment (zomars) [inline apps/web/pages/api/auth/[...nextauth].tsx:108]: - Can the make this function throw instead of returning a `isRateLimited`?
- Why it says `10 request per minute` and we're passing a `100` there?

apps/web/pages/api/auth/[...nextauth].tsx
Comment (agustif) [reply apps/web/pages/api/auth/[...nextauth].tsx:106]: Will fix, thanks for spotting it @zomars

apps/web/pages/api/auth/[...nextauth].tsx
Comment (agustif) [reply apps/web/pages/api/auth/[...nextauth].tsx:108]: Sure, will change and update, I was testing more requests to see if that fixed the onboarding tests, but what should be the limit we allow? Should it be process.ENV based too in case we want to change this without updating the codebase? @zomars

Comment (zomars) [reply apps/web/pages/api/auth/[...nextauth].tsx:108]: > Should it be process.ENV based too in case we want to change this without updating the codebase?

Not now

agustif:
> It seems like this PR has a major flaw, I could be wrong tho. It seems like this is limiting the endpoint calls globally instead of per users/requester.

Hey @zomars I think the issue was the uniqueTokenPerInterval, seems to be working now and all tests/build is passing too!

Leaving tests to be implemented by @alannnc in a new PR in th future!

packages/lib/rateLimit.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import cache from "memory-cache";
+
+import { ErrorCode } from "./auth";
+
+const rateLimit = (options: { interval: number }) => {
+  return {
+    check: (limit: number, token: unknown) => {
```
zomars, line 8:
```suggestion
    check: (limit: number, token: string) => {
```

Correct me if I'm wrong @agustif

packages/lib/rateLimit.ts
```diff
@@ -0,0 +1,27 @@
...
+      const tokenCount: any = cache.get(token) || [0];
```
zomars, line 9:
```suggestion
      const tokenCount = cache.get(token) || [0];
```

No unnecessary  `any`s please

packages/lib/rateLimit.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import cache from "memory-cache";
+
+import { ErrorCode } from "./auth";
+
+const rateLimit = (options: { interval: number }) => {
```
zomars, line 6:
```suggestion
const rateLimit = (options: { intervalInMs: number }) => {
```

zomars (approved):
Thanks again @agustif 🙏🏽

PR #3840, opened 2022-08-14
fix: underfetching of upcoming booking in case of recurring events

hussamkhatib:
## What does this PR do?

Fixes 
>    // Remove duplicate recurring bookings for upcoming status.
      // Couldn't use distinct in query because the distinct column would be different for recurring and non recurring event.
      // We might be actually sending less then the limit, due to this filter
      // TODO: Figure out a way to fix it.

- performs two queries for status upcoming 
    - 1st Distinct recurring events 
    - Non recurring events 
- Concatenates it and sort it by startDate.

- This will fetch upto 22 records,
   - can be reduced to say 14 records by changes `take: take + 1` to  `take: take - 4` on both queries. (will change is this will be better)

- Other booking Status excluding **upcoming** stays unaffected.

## Type of change

## How should this be tested?

A good amount to upcoming booking with both recurring and non-recurring type should be enough to see the new change.

### DIFF
--- a/packages/trpc/server/routers/viewer.tsx
+++ b/packages/trpc/server/routers/viewer.tsx
@@ -370,56 +370,115 @@ const loggedInViewerRouter = createProtectedRouter()
       };
       const passedBookingsFilter = bookingListingFilters[bookingListingByStatus];
       const orderBy = bookingListingOrderby[bookingListingByStatus];
-      const bookingsQuery = await prisma.booking.findMany({
-        where: {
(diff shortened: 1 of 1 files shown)


hussamkhatib:
I just realized this will not always fetch correct results.
I tried to explore all possibilties I could solve this using prisma, the only solution I think is left is to perform rawQuery

PeerRich:
closed due to inactivity

PR #3850, opened 2022-08-14
Allow scrolling when the dialog exceeds screen height

emrysal:
## What does this PR do?

Currently when we have a screen height that is not enough to for the dialog to fully display, there is no scroll behaviour. This introduces that, it has a padding of `py-4` to maintain a little offset from the top.

Main difference - overflow now wraps content, rather than them being separated.

### DIFF
--- a/apps/web/components/eventtype/CreateEventType.tsx
+++ b/apps/web/components/eventtype/CreateEventType.tsx
@@ -152,7 +152,7 @@ export default function CreateEventTypeButton(props: CreateEventTypeBtnProps) {
         options={props.options}
       />
 
-      <DialogContent className="overflow-y-auto">
+      <DialogContent>
         <div className="mb-4">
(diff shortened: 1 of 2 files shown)


apps/web/components/eventtype/CreateEventType.tsx
```diff
@@ -152,7 +152,7 @@ export default function CreateEventTypeButton(props: CreateEventTypeBtnProps) {
         options={props.options}
       />
 
-      <DialogContent className="overflow-y-auto">
+      <DialogContent>
```
emrysal, line 155: This didn't work before; but it did break the new scroll behaviour so for this reason I've edited this.

sean-brydon (approved):
LGTM :)

pumfleet (approved):
Code looks good to me

PR #3851, opened 2022-08-15
Hotfix: Embed - Fix issue in accessing sessionStorage in certain scenarios

### LINKED ISSUE #2618: Iframe not working in incognito mode => fails to access localstorage
### Issue Summary

I have embedded our calendar as an Iframe on our website. Unfortunately, it sometimes does not show up or just shows a white or black screen. I have noticed that this problem only occurs when in incognito mode or with restricted cookie settings. A colleague reported the same problem in Safari - but I can not confirm.

The page with the Iframe: https://clipflip.video/get-started.html

In console there is the following error message:
`framework-45ce3d09104a2cae.js:1 DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

### Steps to Reproduce

1. Embed Iframe on page
2. open incognito mode or strict cookie settings
3. check whether it works

### Technical details

- We are using the hosted version from Cal.com
- The cal.com URL is: https://cal.com/clipflip/demo

Thank you for your help.

hariombalhara:
Chrome in incognito mode blocks access to sessionStorage in cross-domain scenario. Even reading sessionStorage throws error.

Fixes https://github.com/calcom/cal.com/issues/2618

## What does this PR do?

Fixes # (issue)

**Environment**: Production

## Type of change

## How should this be tested?

- [x] Run embed-core dev server and then open http://yoogle.com:3100/embed/?only=ns:default#cal-booking-place-default-iframe(. yoogle.com always refer to localhost. This is a way to similar cross-domain behaviour on local). At this point embed won't work and there would be errors in console. This change should fix the issue.

## Checklist


emrysal:
Is there a specific reason for using try {} catch {} instead of
```typeof window.localStorage !== "undefined"``` or ```typeof window.localStorage?.getItem() === "function"``` (both untested pseudocode, but you get the gist)

hariombalhara:
Yeah, Chrome is very strict. You are not even allowed to read localStorage/sessionStorage properties in cross-domain incognito mode.

Screenshot 2022-08-15 at 6 27 28 PM

zomars (approved):
Loving the eslint rule approach. Nice work @hariombalhara

zomars:
@hariombalhara should we fallback to using cookies or is that restricted as well?

hariombalhara:
@zomars Nope, cookies are also not available. Though it fails silently, like there is no problem.

Screenshot 2022-08-16 at 9 41 46 AM

PR #3877, opened 2022-08-16
Attempt at redirect to 404

emrysal:
404 should be able to handle the traffic no problem - better ideas welcome. Just a temp fix I imagine; we're not going to want to keep this nor can we expect the User-Agent to continue identifying the traffic source.


emrysal:
@zomars this may work.

zomars (approved):

PR #3892, opened 2022-08-17
Improves Zapier response and payload

CarinaWolli:
## What does this PR do?

Improves the response of Zapier Test Trigger (listBookings endpoint) and the payload when calling the subscriber Url. 

New response: 
title 30min between Pro Example and Test User
name Test User

Before: 
Screenshot 2022-08-17 at 19 38 16

## Checklist


apps/web/lib/linkValueToString.ts
```diff
@@ -39,3 +39,30 @@ export const linkValueToString = (
...
+    case "integrations:whereby_video":
+      return "Whereby";
+    case "integrations:around_video":
+      return "Around";
+    case "integrations:riverside_video":
+      return "Riverside";
+    default:
+      return location;
+  }
+};
```
PeerRich, line 68: @zomars is this something we wanna do? maybe there is a better way to return the name of the integration. I feel like this will throw errors once we add another Location-based App and forget to add it here.

I really don't want to have app-specific code spread out into the core code base

apps/web/lib/webhooks/sendPayload.tsx
```diff
@@ -4,8 +4,51 @@ import { compile } from "handlebars";
...
+function getZapierPayload(data: CalendarEvent & EventTypeInfo & { status?: string }): string {
+  const attendees = data.attendees.map((attendee) => {
+    return {
+      name: attendee.name,
+      email: attendee.email,
+      timeZone: attendee.timeZone,
+    };
+  });
+
+  const location = integrationLocationToString(data.location || "");
```
PeerRich, line 29: maybe instead we add `data.locationName`

PeerRich: .. I believe there is a way to return the name without the switch

PeerRich (commented):
left some comments around not having app-specific code in core code @zomars

apps/web/lib/linkValueToString.ts
CarinaWolli: > I feel like this will throw errors once we add another Location-based App and forget to add it here.

It shouldn't throw an error, it should just use the default location which then would be the unformatted location string

apps/web/lib/webhooks/sendPayload.tsx
CarinaWolli: updated it to use the new `getHumanReadableLocationValue` function

zomars (changes requested):
It seems like E2E is failing since the updated web hook response. Let's update the web hook snapshot as well.

PR #3898, opened 2022-08-18
Fixes collective availability for teams with overlapping day timezones

zomars:
## What does this PR do?

The current collective availability algorithm was merging the team mates availability BEFORE offsetting them to UTC which was causing cases where Timezones that have little overlap don't show any slots.

## Type of change

## How should this be tested?

- [ ] yarn test


CodingRockz:
This does not yet solve the team collective schedule slots issue. Testing with the code changes still show no open slots for the team, infact it is now worse - no slots even if the team only has 1 user.

zomars:
> This does not yet solve the team collective schedule slots issue. Testing with the code changes still show no open slots for the team, infact it is now worse - no slots even if the team only has 1 user.

Sir. This is a WIP draft PR.

packages/core/getUserAvailability.ts
```diff
@@ -78,14 +77,14 @@ export const getCurrentSeats = (eventTypeId: number, dateFrom: Dayjs, dateTo: Da
 
 export type CurrentSeats = Awaited<ReturnType<typeof getCurrentSeats>>;
 
+/** This should be called getUsersWorkingHoursAndBusySlots (...and remaining seats, and final timezone) */
```
emrysal, line 80: Yes it does too much at the moment, but I have ideas 😎

packages/lib/availability.ts
```diff
@@ -65,21 +65,15 @@ export function getWorkingHours(
...
   if (!availability.length) {
-    return [
-      {
-        days: [0, 1, 2, 3, 4, 5, 6],
-        // shorthand for: dayjs().startOf("day").tz(timeZone).diff(dayjs.utc().startOf("day"), "minutes")
-        startTime: MINUTES_DAY_START,
-        endTime: MINUTES_DAY_END,
-      },
-    ];
+    return [];
```
emrysal, line 69: BTW, this fixes the unrelated bug that we default to ALL availability when no availability is set in a schedule.

packages/lib/slots.ts
```diff
@@ -52,14 +56,14 @@ const getSlots = ({ inviteeDate, frequency, minimumBookingNotice, workingHours,
...
-    { utcOffset: -inviteeDate.utcOffset() },
-    workingHours.map((schedule) => ({
-      days: schedule.days,
-      startTime: startOfDay.add(schedule.startTime, "minute"),
-      endTime: startOfDay.add(schedule.endTime, "minute"),
-    }))
-  ).filter((hours) => hours.days.includes(inviteeDate.day()));
+  const workingHoursUTC = workingHours.map((schedule) => ({
+    days: schedule.days,
+    startTime: /* Why? */ startOfDayUTC.add(schedule.startTime, "minute"),
```
emrysal, line 61: The getWorkingHours function works with ConfigType instances that can't make sense of startTime given in minutes since the start of the day (in UTC)

emrysal (commented):
Still reviewing 👍

emrysal (approved):

packages/lib/availability.ts
zomars: Had to update users fixtures since test users we're being created without availability.

emrysal: Oh! Good that we caught that as this is an impossible situation due to the onboarding flow these days 💯 - nice catch!

packages/lib/slots.ts
zomars: Would it make sense to assume inside `getWorkingHours` that if the `startTime/endTime` is type of number then we skip all conversions altogether?

PR #3932, opened 2022-08-23
Attempt to fix telemetry loop

emrysal:
* Fixes deprecation notice
* Attempt at NextResponse.next() instead of rewrite to stop triggering middleware too much (potentially, it's a guess)


zomars (approved):
ty @emrysal

PR #3933, opened 2022-08-23
Refactors google calendar service

zomars:
## What does this PR do?

- Uses zod to actually validate API keys and user Credentials
- Optimizes auth legibility
- Uses fetcher helper to simplify API calls

## Type of change

- [x] Chore

## How should this be tested?

- [ ] Test Google Calendar integration


packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -326,22 +303,20 @@ export default class GoogleCalendarService implements Calendar {
               },
             },
             (err, apires) => {
-              if (err) {
+              if (err || !apires?.data.calendars) {
```
zlwaterfield, line 306: `!apires?.data.calendars` doesn't mean `err` will be defined, would it be better to have another check?

```
if (!apires?.data.calendars) {
	reject("No calendar data found...")
	return;
}

zlwaterfield (commented):
@zomars what is the best way to test this locally? I don't see Google Calendar as an app locally.

packages/app-store/googlecalendar/lib/CalendarService.ts
zomars: That check passes if any of those is true

zomars: Here are the instructions to [setup Google](https://github.com/calcom/cal.com#obtaining-the-google-api-credentials)

zlwaterfield: Yes but if `err` is `null` and it hits on `!apires?.data.calendars` then you'll be rejecting on null (`reject(null)`) and there will be no context in why the error occurred.

zomars: Agreed. will update

hariombalhara (approved):

PR #3934, opened 2022-08-23
fix: add required property on input

### LINKED ISSUE #3918: Required additional inputs are not actually required
### Issue Summary

When completing the form to book en event I can sumbit the form even if some custom additional inputs that are required are empty. 

### Steps to Reproduce

1. Went to the admin dashboard, add a custom addition field to an event 
2. Checked the box "Is required"
3. When testing the event, even if the new custom field is empty I can book the event
4. Only the first two default fields are actually required (Name & Email)

I use the live version of Cal, not self-hosted.

Udit-takkar:
Fixes: #3918 
Also added checks in /api/book if required customInputs are missing.

Loom Video:- https://www.loom.com/share/01ff8da8e2764a1691f23572a37344a6


PeerRich:
awesome work! super fast.👏 @hariombalhara can you review and approve this?

apps/web/pages/api/book/event.ts
```diff
@@ -249,8 +255,24 @@ async function handler(req: NextApiRequest) {
 
   const eventType =
     !eventTypeId && !!eventTypeSlug ? getDefaultEvent(eventTypeSlug) : await getEventTypesFromDB(eventTypeId);
+
   if (!eventType) throw new HttpError({ statusCode: 404, message: "eventType.notFound" });
 
+  // Check if required custom inputs exist
+  if (eventType.customInputs) {
+    eventType.customInputs.forEach((event) => {
```
emrysal, line 263:
```suggestion
    eventType.customInputs.forEach((customInput) => {
```

apps/web/pages/api/book/event.ts
```diff
@@ -249,8 +255,24 @@ async function handler(req: NextApiRequest) {
...
+      if (
+        event.required === true &&
+        (reqBody.customInputs.length === 0 ||
+          reqBody.customInputs.filter(
```
emrysal, line 267: Can reqBody.customInputs ever be empty? If so `.length` could crash

emrysal (commented):
Also have a look at using zod in the book/event endpoint; could be helpful

apps/web/pages/api/book/event.ts
Udit-takkar: i don't it would crash as this is also being used on line 509
```
  if (reqBody.customInputs.length > 0) {
    reqBody.customInputs.forEach(({ label, value }) => {
      customInputs[label] = value;
    });
  }
  ```

**Comment by Udit-takkar** [comment]
@emrysal  where exactly can zod be helpful in 'book/event endpoint'. i think it is already being used at the start 
```
 extendedBookingCreateBody.parse(req.body);
 ```

apps/web/pages/api/book/event.ts
```diff
@@ -231,6 +232,11 @@ async function ensureAvailableUsers(
   }
   return availableUsers;
 }
+const checkIfValueIsPresent = (val: string | boolean) => {
+  if (typeof val == "boolean" && val === false) return false;
```
hariombalhara, line 236: A boolean can have false as a valid value, I don't think false should be considered empty. `undefined` or `null` is empty. Due to this, a checkbox if marked required isn;t allowing any other value value except true.

hariombalhara (approved):
_Ignore this_

hariombalhara (dismissed):
I left a comment that needs to be addressed before the merge.

Also, there is a cleaner way to do it using zod, which allows schema validation to work for frontend as well as backend with a single definition. See zodResolver usage in BookingPage.tsx. I wouldn't consider this a blocker for merge but it is good to have.

Thanks again for your contribution @Udit-takkar

apps/web/pages/api/book/event.ts
Udit-takkar: i thought if a checkbox is marked required that it should be true (like i agree terms and conditions ). there is only one checkbox so how do we differentiate whether user wanted to mark false or just skip it and mark empty ?

emrysal: Alternative approach:
```javascript
const checkIfValueIsPresent = (val: string | boolean) => {
  if (typeof val == "string" && val.trim() === "") return false;
  return !!val;
};

Comment (Udit-takkar) [reply apps/web/pages/api/book/event.ts:238]: @emrysal ok we can use this but i am trying out zod here like

`export const checkStringOrBooleanExist = z.union([z.string().min(1), z.boolean()]);`

```
  if (eventType.customInputs) {
    eventType.customInputs.forEach((customInput) => {
      if (
        customInput.required === true &&
        (reqBody.customInputs.length === 0 ||
          reqBody.customInputs.filter(
            ({ label, value }) => label === customInput.label && checkStringOrBooleanExist.parse(value)
          ).length != 1)
      ) {
        throw new Error("Missing required input");
      }
    });
  }
```
which one should i use ?

and  how would required custom input checkbox would work ? do we allow not marking it even if it is a required input . well in this case both required and not required would work the same (you can mark it or leave it).

Comment (emrysal) [reply apps/web/pages/api/book/event.ts:238]: What do you think about the version below?

```typescript
const requiredCustomInputSchema = z.union([
  // string must be given & nonempty
  z.string().trim().min(1),
  // boolean must be true if set.
  z.boolean().refine((v) => v === true),
]);

if (eventType.customInputs) {
    eventType.customInputs.forEach((customInput) => {
      if (customInput.required === true) {
        requiredCustomInputSchema.parse(reqBody.customInputs.find(
          userInput => userInput.label === customInput.label)?.value
        ));
      }
    });
  }
```

`.parse` throws an error of its own, so you can use that instead of throwing a custom error. Alternatively use `.safeParse` and read the .success return value.

Udit-takkar: This one is much better 👍

Udit-takkar: done

emrysal (approved):
Excellent PR - thanks for the contribution and follow ups! 👍

PR #3945, opened 2022-08-24
Fixes 400 being thrown for NaN on disabling seats

### LINKED ISSUE #3944: Error deactivating seats in event type
### Issue Summary

I have an event type that has the "offer seats" option activated. When I edit the event type, deselect the option and hit "save!", I get an error in the console:

cal com bug

### Steps to Reproduce

1. Create an event type that has the "offer seats" option activated.
2. Edit it.
3. Deselect "offer seats".
4. Click "save"

### Technical details

I'm using the latest Chrome on macOS.

sean-brydon:
## What does this PR do?

Fixes #3944

## Type of change


zomars (changes requested):
Good solution @sean-brydon but I have some Qs.  Why is seatPerTimeSlot is returning as NaN in the first place? My guess is that this is coming upstream in the flow. We can either, prevent the NaN from happening earlier instead of patching down here. My suspicion is a faulty `parseInt` or something like that.

PeerRich:
> > Good solution @sean-brydon but I have some Qs. Why is seatPerTimeSlot is returning as NaN in the first place? My guess is that this is coming upstream in the flow. We can either, prevent the NaN from happening earlier instead of patching down here. My suspicion is a faulty parseInt or something like that.
> 
> @zomars Just had a look into this and it appears that we are sending null over the form but for some reason its getting translated to a NaN - looking into this more but I can't seem to figure out why this is happening

whats the step forward? merge or investigate more=

sean-brydon:
@zomars after digging into this this morning I got nowhere - we send over null via rhf and it ends up in the form state as NaN. Not 100% sure as the type is set as `number | null`

manuelmeurer:
Would it be possible to deploy this, so the bug is fixed, and keep looking for a better solution?
If not, is there some workaround to disable seats for an event type in the meantime?

zomars (approved):
Merging for now but this general form needs a major refactor

PR #3951, opened 2022-08-24
feat: additional hidden property added in event types

### LINKED ISSUE #3930: Hidden additional fields

I'm using the hosted version and I need to pre-fill a custom input field on the Cal react component.
I then want to capture that value from the payload in the "BOOKING_CREATED" webhook that fires after the event is scheduled. 

I could add an additional input field to the event type and pre-fill it, but this is not a field that I want to expose to the customer in the booking form.

It would be great to have an additional field of type: "hidden" available in the advanced event type settings.
Alternatively, having some input options on the component would be good.

Peer suggested adding a hidden checkbox in the additional field UI.

IMG_7070

Exposing the field to the user. 
Currently I'm getting the value on the server in the webhook listener but I'd rather just pass it along to the Cal component and retrieve it from the payload.

### Additional context

Hidden fields are very commonly used across a wide range of verticals, for marketing and dev purposes.
Would be a great addition to the "consumer" side of the product!

Udit-takkar:
Fixes: #3930 
Loom video Demo:- https://www.loom.com/share/7f4da463816d4b16bccedf128513660c

cc: @zomars


packages/prisma/schema.prisma
```diff
@@ -342,6 +342,7 @@ model EventTypeCustomInput {
   label       String
   type        EventTypeCustomInputType
   required    Boolean
+  hidden      Boolean                  @default(false)
```
zomars, line 345: Why adding a new property instead of adding a new `EventTypeCustomInputType` ? AFAIK inputs of type "hidden" are strings anyways so I don't see much point of having "bool" hidden for example.

zomars (changes requested):
Thanks @Udit-takkar but this solution is not quite there yet. Also what kind of error are we displaying if the input is of type hidden and is a required one? cc @Jaibles @PeerRich  to provide design feedback

packages/prisma/schema.prisma
Udit-takkar: well EventTypeCustomInputType is for the type of input ( text, number ). what peer mentioned here https://calendso.slack.com/archives/C01NXDSRY9W/p1661256038898709?thread_ts=1661217673.551779&cid=C01NXDSRY9W is to have an additional checkbox to mark the additional custom as input very similar to required checkbox. 

i don't understand how would adding new EventTypeCustomInputType help   ?

Udit-takkar:
@zomars is there any example of prefilling hidden custom input ? any app that uses it ? I'll make all the necessary changes on this branch

packages/prisma/schema.prisma
zomars: Hidden should not be a property of a field but a field type on it's own and render them as:

`<input type="hidden" />`

Instead of 

`<input type="text/number/checkbox/etc..." className="hidden" />`

PeerRich:
i dont think hidden can be required. since hidden fields will be used programmatically its not smart to show an error to the end user

packages/prisma/schema.prisma
Udit-takkar: i am rendering it as a field type `<input  type={input.hidden ? "hidden" : "number"} />`.  hidden is stored in db so that we know which custom input type(number, text) was marked hidden by checkbox .  

so what i should do now is create new EventTypeCustomInputType of type hidden and not create any checkbox for marking hidden?

zomars:
After a brief discussion with the team, we'll closing this since this functionality is already achievable using metadata and the pre-fill feature. Thanks for you patience and understanding @Udit-takkar 🙏🏽

PR #3955, opened 2022-08-24
Fixing migration issue with postgres version 11

CarinaWolli:
## What does this PR do?

Fixes issue with Migration 20220811234822_add_after_meeting_ends_trigger with Postgres version 11

Fixes  #3904

## Type of change


packages/prisma/migrations/20220811234822_add_after_meeting_ends_trigger/migration.sql
```diff
@@ -1,5 +1,3 @@
 -- AlterEnum
 ALTER TYPE "WebhookTriggerEvents" ADD VALUE 'MEETING_ENDED';
```
simobasso, line 2: Hey 👋 , @CarinaWolli nice to meet you!

I've just tried this pr on my local env, and it seems that it doesn't fix the problem 😢 . 

Prisma wraps every migration inside an explicit transaction block, and in PostgreSQL 11, as said in the main issue, an alter type to add a new value on an enum is not supported.

If you run this migration with PostgreSQL 11, this error pop up:

```
ERROR: ALTER TYPE ... ADD cannot run inside a transaction block
```

the [PostgreSQL v11 doc](https://www.postgresql.org/docs/11/sql-altertype.html) stated that:

> ALTER TYPE ... ADD VALUE (the form that adds a new value to an enum type) cannot be executed inside a transaction block.
>
> Comparisons involving an added enum value will sometimes be slower than comparisons involving only original members of the enum type. This will usually only occur if BEFORE or AFTER is used to set the new value's sort position somewhere other than at the end of the list. However, sometimes it will happen even though the new value is added at the end (this occurs if the OID counter "wrapped around" since the original creation of the enum type). The slowdown is usually insignificant; but if it matters, optimal performance can be regained by dropping and recreating the enum type, or by dumping and restoring the database.

So we have a few options:

1. run the migration outside a transaction block: as @AlessandroLorenzi suggests in #3904, you have to close with a `COMMIT` the transaction block at the start of the migration like:

```
-- close the transaction to support the new enum value on pg11
COMMIT;
-- AlterEnum
ALTER TYPE "WebhookTriggerEvents" ADD VALUE 'MEETING_ENDED';
```

2. As the pg documentation suggests: you can drop and recreate the enum with the new value.
3. Ask the final user to run this migration manually as suggested by https://github.com/prisma/prisma/issues/9154, but every automatic migration run will break cal.com on pg 11 until someone runs the migration by hand.

Thanks for your amazing job; Feel free to ask if there is something unclear!

Cheers!

packages/prisma/migrations/20220803091114_drop_daily_event_reference/migration.sql
```diff
@@ -9,3 +9,6 @@ ALTER TABLE "DailyEventReference" DROP CONSTRAINT "DailyEventReference_bookingId
 
 -- DropTable
 DROP TABLE "DailyEventReference";
+
```
simobasso, line 12: Please be aware that editing an already applied migration can be [potentially dangerous](https://www.prisma.io/docs/concepts/components/prisma-migrate#do-not-edit-or-delete-migrations-that-have-been-applied) 🌻

agustif: Yep @simobasso I think we should probably create a new migration that fixes the old one right?

zomars: > Please be aware that editing an already applied migration can be [potentially dangerous](https://www.prisma.io/docs/concepts/components/prisma-migrate#do-not-edit-or-delete-migrations-that-have-been-applied) 🌻

- If the migration was already applied (for postgres >12) Changing the migration won't affect them since it will be already marked as applied
- If the migration failed (<11), changing the migration won't affect them since they can try apply the migration again.

@CarinaWolli just followed my suggestion of updating the migrations based on my points above.

packages/prisma/migrations/20220811234822_add_after_meeting_ends_trigger/migration.sql
zomars, line 2:
```suggestion
-- close the transaction to support the new enum value on pg11
COMMIT;
-- AlterEnum
ALTER TYPE "WebhookTriggerEvents" ADD VALUE 'MEETING_ENDED';
```

Do you think this suggestion would fix this? @CarinaWolli @simobasso

CarinaWolli: I am running the migrations locally using a postgres 11 heroku database and for me, it only works with the changes I made. Even once adding the COMMIT I get the error message 'ERROR: ALTER TYPE ... ADD cannot run inside a transaction block'. @simobasso does it work for you when adding the COMMIT?

PeerRich:
is this still relevant @CarinaWolli ?

PeerRich:
closed due to inactivity. happy to reopen

PR #3891, opened 2022-08-17
fix/auto-connect-calendar-3582

alannnc:
## What does this PR do?

- Removes label that says no connected calendar to an event type
- Shows new info Alert on event-types page when no destinationCalendar (default calendar) is found
- Adds onCascade: true to destinationCalendar model
- Now we display destinationCalendar (default) in event-type/$id page in select when no specific calendar is selected.

Fixes #3582 

https://www.loom.com/share/dbf7372d65ad4104baea2bcc46b78980

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [ ] Remove any calendar, go to event-types pages an alert should be showing up
- [ ] Enter event-type/$id page and when no destinationCalendar it's selected for that event type we should display user.destinationCalendar(default) as placeholder in Select component.

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works

PR #3961, opened 2022-08-25
V1->V2 connect calendar

sean-brydon:
## What does this PR do?

https://github.com/calcom/cal.com/pull/3891

Puts these changes from V1 into V2 components


apps/web/pages/v2/event-types/index.tsx
```diff
@@ -609,6 +614,22 @@ const EventTypesPage = () => {
...
+                      You can connect your calendar from
+                      <a href="/apps/categories/calendar" className="underline">
+                        here
+                      </a>
+                      .
+                    </Trans>
+                  }
+                />
+              )}
               {data.eventTypeGroups.map((group, index) => (
```
PeerRich, line 633: this didnt seem to work in the old PR and was reverted, are we sure we wanna keep this?

sean-brydon: @alannnc  - Is it worth removing this?

alannnc: Just take whats on main and you should be ready to go. I think on main its called as <NoCalendarConnectedAlert/>

apps/web/pages/v2/event-types/index.tsx
```diff
@@ -634,4 +655,26 @@ const EventTypesPage = () => {
...
+    });
+    defaultCalendarConnected = !!defaultCalendar;
+  }
+  return {
+    props: {
+      defaultCalendarConnected,
+    },
+  };
+}
+
```
alannnc, line 679: With new component this is not required now.

alannnc (approved):

PR #3966, opened 2022-08-25
AppsTab UX + fixes warptext on horizontal tabs

sean-brydon:
## What does this PR do?
Fixes no apps being installed + fixed the horizontal tab from warping on mobile	

CleanShot 2022-08-25 at 14 03 19


emrysal (changes requested):
Should be all fine after the eslint errors are fixed 👍

apps/web/components/v2/eventtype/EventAppsTab.tsx
```diff
@@ -70,8 +71,20 @@ export const EventAppsTab = ({
...
 
+  const installedApps = [hasPaymentIntegration, hasGiphyIntegration].filter(Boolean).length;
+
+  if (installedApps === 0) {
+    return (
+      <EmptyScreen
+        Icon={Icon.FiGrid}
+        headline={t("empty_installed_apps_headline")}
+        description={t("empty_installed_apps_description")}
+        buttonRaw={<Button href="/apps">{t("empty_installed_apps_button")} </Button>}
```
PeerRich, line 82: it would be nice if we save event type settings here before routing

sean-brydon: Will pick this up in a separate PR cause I agree :) - modal popop warning the user too

zomars (approved):
Nice work @sean-brydon 🙏🏽

PR #3968, opened 2022-08-25
Forbid reserved keywords as usernames

afzalsayed96:
## What does this PR do?

Prevents registering key username such as `jobs`

Screenshot 2022-08-25 at 8 19 53 PM

image


packages/features/ee/common/lib/checkPremiumUsername.ts
```diff
@@ -10,6 +10,14 @@ interface ResponseUsernameApi {
 
 export async function checkPremiumUsername(_username: string): Promise<ResponseUsernameApi> {
   const username = slugify(_username);
+  if (username in FORBIDDEN_USERNAMES) {
```
afzalsayed96, line 13: Assuming `key in object` will be faster than `array.includes(string)`

packages/features/ee/common/lib/checkPremiumUsername.ts
```diff
@@ -10,6 +10,14 @@ interface ResponseUsernameApi {
...
+    return {
+      available: false as const,
+      premium: false,
+      message: "A user exists with that username",
```
afzalsayed96, line 17: This message can be changed

packages/lib/constants.ts
```diff
@@ -36,3 +36,14 @@ export const DEVELOPER_DOCS = "https://developer.cal.com";
...
+  about: "about",
+  blog: "blog",
+  download: "download",
+  jobs: "jobs",
+  open: "open",
+  privacy: "privacy",
+  roadmap: "roadmap",
+  security: "security",
+  terms: "terms",
+};
```
afzalsayed96, line 49: Not exhaustive but covers most clashes with current website pages

PeerRich:
cool idea! luckily those accounts dont affect website behaviour but its probably a good idea

PeerRich:
alternatively we can just create accouts for each page that we create. not clean but also gets the job done

zomars (changes requested):

- I would rather have this at a per-database level since self-hosted shouldn't have to worry about this.
- We already have some guards on our private website repo
- I'd rather have as less hardcoding as possible

PR #3977, opened 2022-08-25
Adds deprecation notice to QueryCell

zomars:
## What does this PR do?

- Adds deprecation notice for `QueryCell`
- It is becoming hard to debug/understand/customize
- We're making efforts to adopt a custom skeleton approach per component instead of generic `Loader`
- The original maintainer is no longer an active contributor

## Type of change


sean-brydon (approved):
Approved: As we discussed on around the other day - query cell is a massive PINTA to work with.

PR #3986, opened 2022-08-26
Static images through webpack copy

emrysal:
Fixes an issue with `[...static].jsx` that seems to bundle all images up into one giant serverless function within Vercel.

Next 12.2 allows turning off next/image optimisation which I’ve done; optimisation is probably broken across the board on `<Image />` for public/ assets but we’ve historically replaced with  tags where this became a problem (see logo’s, app icons). We don’t really care about that feature anyway, not that kind of website.

Also rewrites don’t seem to work on assets, redirect worked no problem but a rewrite just didn’t.


emrysal:
@alishaz-polymath could you verify whether this implementation supports Windows? 😅

zomars (approved):

PR #3994, opened 2022-08-26
fixing cursor for Default badge svg, from text type to default

edgarallanglez:
## What does this PR do?

The Default badge in the Schedule edit screen inside the availability component was changing the cursor to text type on hover, and now is showing the default cursor style since is not editable or clickable.

image


PeerRich:
thank you for the minor change!

PR #3997, opened 2022-08-27
feat: Adds new v2 design for app detail page. Fixes #3779

### LINKED ISSUE #3779: 2.0 Apps / App Store / Singe App {View}
[View in Figma ](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=13319%3A80442)

Image

JeroenReumkens:
## What does this PR do?

Implements new design for app detail page. 

Screenshots of actual work: 
image
Screenshot 2022-08-27 at 14 44 14

Fixes #3779
 Loom Video: https://www.loom.com/share/13288fbc55b04b9287094deae378bda6

## Type of change

## How should this be tested?

Visit app detail pages and ensure layout is correct. Also verify v1 page is NOT impacted by this change. 

## Open questions

1. @Jaibles  Let me know what you think about the mobile version :) And of course any other design feedback is welcome! 
2. During the Loom video I also resize my browser to a very small height. Currently I'm using the default CSS `sticky`. Its default behaviour however is that you only see the out of view portion once you've reached the bottom of the page. I'd rather have it immediately scroll down, and then stay sticky at the bottom. However, in order to do this, we need to either write quite some JS, or implement a fixed height and scrollbar in the sticky section on the right. Both don't really have my preference, since the use case for this is pretty small (so adding JS might not make sense), and adding a fixed height + scroll bar makes the UX worse. What do you guys think?


apps/web/components/v2/apps/App.tsx
```diff
@@ -0,0 +1,337 @@
+import classNames from "classnames";
```
JeroenReumkens, line 1: Any logic in this file (for eg pricing, showing of buy button etc) is coming directly from v1. I only touched styling in this file. 

Only new thing is that `<Slider>` isn't shown directly in the body anymore. We now extract the images from the MDX's meta data, and render it in its own column.

apps/web/components/v2/apps/App.tsx
```diff
@@ -0,0 +1,337 @@
...
+                            mutation.mutate({ type });
+                          },
+                          loading: mutation.isLoading,
+                        };
+                      }
+                      return (
+                        <Button
+                          StartIcon={Icon.FiPlus}
+                          {...props}
+                          // @TODO: Overriding color and size prevent us from
```
JeroenReumkens, line 156: Let me know if this is the best to do for now @sean-brydon. Currently the types of `InstallAppButton` use the old types of button for `props`. Because of this the button component gives a TS error. By manually overriding color + size, everything is fine. I was unsure whether it would make sense to only for changes the types would make sense to duplicate the `InstallAppButton`, since this also brings overhead in maintenance etc.

apps/web/pages/v2/apps/[slug]/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+      <a {...otherProps} />
+    </Link>
+  ),
+  img: ({ src = "", alt = "", placeholder, ...rest }: JSX.IntrinsicElements["img"]) => (
+    <Image src={src} alt={alt} {...rest} />
+  ),
+  // @TODO: In v2 the slider isn't shown anymore. However, to ensure the v1 pages keep
+  // working, this component is still rendered in the MDX content. To skip them in the v2
+  // content we have to render null here. In v2 the gallery is shown by directly
+  // using the `items` property from the MDX's meta data.
```
Comment (JeroenReumkens) [inline apps/web/pages/v2/apps/[slug]/index.tsx:29]: Because in v1 the `<Slider>` is part of the content (there can be text above and underneath the content), we can not remove it yet. Therefore we render an empty fragment here (null isn't allowed in MDXRemote's type 🤷‍♂️). As soon as v1 is gone, we can remove the `Slider` component from the MDX. In v2 we directly use the items from the MDX's meta data. (see line 50).

PeerRich:
really cool PR! 👏 we'll review tomorrow

Jaibles:
@JeroenReumkens this looks fantastic. Great work 👏

apps/web/middleware.ts
```diff
@@ -11,6 +11,9 @@ const V2_WHITELIST = [
   "/availability",
   "/bookings",
   "/event-types",
+  // Apps contains trailing slash to prevent app overview from being rendered as v2,
+  // since it doesn't exist yet.
+  "/apps/",
```
leog, line 16: This actually works fine for app pages but it gives an error when navigating to app categories, such as http://localhost:3000/apps/categories/video. We may want to change the logic and rely on regex, WDYT @zomars?

apps/web/components/v2/apps/App.tsx
```diff
@@ -0,0 +1,337 @@
...
+    getInstalledApp(type);
+  }, [type]);
+  const allowedMultipleInstalls = categories.indexOf("calendar") > -1;
+
+  return (
+    <>
+      <div className="px-4 pb-3 pt-3 md:px-8 lg:px-0 lg:pt-0">
+        <Link href="/apps">
+          <a className="text-md hover:text-brand-400 inline-flex rounded-sm py-2 font-semibold  text-gray-900">
+            <Icon.FiArrowLeft className="mr-2 h-6 w-6" /> {t("app_store")}
```
leog, line 88: Not sure your thoughts @Jaibles but having the "<- App Store" back button scroll with the screenshots looks weird to me, maybe is just a NIT. The designs doesn't show the desired behavior here. And when the text side is long enough, scrolling phases come up, see: 

https://user-images.githubusercontent.com/467258/187283383-b75e0e28-e5c0-486a-95d2-fa07893e55c9.mov

apps/web/components/v2/apps/App.tsx
```diff
@@ -0,0 +1,337 @@
...
+                  currency: "USD",
+                  useGrouping: false,
+                }).format(price)}
+                {feeType === "monthly" && "/" + t("month")}
+              </>
+            )}
+          </span>
+
+          <h4 className="mt-8 mb-2 font-semibold text-gray-900 ">{t("learn_more")}</h4>
+          <ul className="prose-sm -ml-1 -mr-1 leading-5">
```
leog, line 227: I think we are missing the "App Store / Single App View v2 (when we have more content)" part, which I'm guessing it will take care of the use case when a lot of information is included in the text side of the page. Would you be willing to expand your already impressive work to accommodate for that use case @JeroenReumkens? Check it out: https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=1413%3A76791

apps/web/middleware.ts
leog: Something like:

```javascript
const V2_WHITELIST = [
  "/settings/admin",
  "/settings/my-account/[a-z](-?[a-z])*",
  "/availability",
  "/bookings/[a-z](-?[a-z])*",
  "/event-types",
  "/apps/[a-z](-?[a-z])*",
];

// ...

if (
    req.cookies.has("calcom-v2-early-access") &&
    V2_WHITELIST.some((p) => new RegExp(`^${p.replace("/", "\\/")}$`).test(url.pathname))
  ) {
    // rewrite to the current subdomain under the pages/sites folder
    url.pathname = `/v2${url.pathname}`;
    return NextResponse.rewrite(url);
  }
```

leog: Oh my, I just noticed "Routing Forms" app has the wrong slug, it uses underscore "routing_forms", we will need to standardize that or adjust the regex 😅

apps/web/pages/v2/apps/[slug]/index.tsx
Comment (leog) [reply apps/web/pages/v2/apps/[slug]/index.tsx:29]: I think we should be able to work out all the changes we need for the release and not hold back on these changes introducing workarounds, but to make it work the changes should not be picked up for production, or we simply expose App Pages v2 to the public earlier. Not sure. @zomars?

leog (dismissed):
Impressive work @JeroenReumkens, thank you very much for your contribution. As we chatted through Slack, let's work things out to have your work merged for our v2 release 💪

apps/web/middleware.ts
JeroenReumkens: I like the regex solution as well. Think this is a lot safer!

apps/web/components/v2/apps/App.tsx
JeroenReumkens: Oh interesting! I actually assumed this was more of a design experiment rather than a variant that would really be used, since it wasn't mentioned anywhere in the issue 😄  How do we decide based on the MDX what variant to show? Maybe it would make even more sense to abstract a few things into a separate component, rather than trying to combine this into one template — that would complicate the template by a lot I think. Could definitely give this a go once I know how we want to use this.

JeroenReumkens:
@leog I also see aaaalll the linting errors now, mostly coming from the original template 😅 Do we want to improve this with the v2 as well? ♥️ Could be quite the work (and introduce other dependencies that need to be fixed as well), but would make it better of course!

apps/web/middleware.ts
zomars: This logic will be gone post v2 launch so I'm okay on adding individual routes and not over engineer this logic.

apps/web/pages/v2/apps/[slug]/index.tsx
Comment (zomars) [reply apps/web/pages/v2/apps/[slug]/index.tsx:29]: not worth the effort for now

zomars (approved):
Impressive work @JeroenReumkens 🙏🏽 I would say this is mergeable as it is now since it doesn't affect v1 users. We can either 
1. Merge as it is and followup improvements in another PR
2. Wait for the improvements in this PR

PeerRich:
> Impressive work @JeroenReumkens 🙏🏽 I would say this is mergeable as it is now since it doesn't affect v1 users. We can either
> 
> 1. Merge as it is and followup improvements in another PR
> 2. Wait for the improvements in this PR

lets do 1 and merge

PeerRich:
thank you so much @JeroenReumkens 👏

PeerRich:
i had to "Merge without waiting for requirements to be met" because tests are failing for external contributors @zomars

PR #4001, opened 2022-08-28
fix: use json.parse to get id

### LINKED ISSUE #3849: Canceled appointment doesnt get removed in outlook calendar
### Issue Summary

An appointment which got booked and canceled afterwards doesnt get removed from the connected outlook calendar.

### Steps to Reproduce

1. Book a meeting
2. Cancel the meeting
3. Check outlook calendar

Appointment should get removed from the calendar, else it is very unreliable.

Udit-takkar:
Fixes:  #3849 

so i tried debugging outlook as this was annoying for some users  and found that uid in BookingReferences table is empty string due to which when we try delete event at calendar service it fails 
```
 const response = await this.fetcher(`/me/calendar/events/${uid}`, {
        method: "DELETE",
      });
```

i found out this is due to the fact that response after created event  ( Eventmanager.ts) has to parsed first to access it.
       
cc: @joeauyeung @PeerRich 

working demo:-

https://user-images.githubusercontent.com/53316345/187066182-0f7dca63-ac19-42d3-adbc-3ca410806e73.mov


packages/core/EventManager.ts
```diff
@@ -102,12 +102,18 @@ export default class EventManager {
...
     const referencesToCreate = results.map((result) => {
+      const createdEventObj = JSON.parse(result?.createdEvent as string);
       return {
         type: result.type,
-        uid: result.createdEvent?.id?.toString() ?? "",
-        meetingId: result.createdEvent?.id?.toString(),
-        meetingPassword: result.createdEvent?.password,
-        meetingUrl: result.createdEvent?.url,
+        uid:
+          result.type === "office365_calendar"
```
dallenpyrah, line 109: Could we create a variable for "office365_calendar" so if needed we only have to change in one place?

Udit-takkar: @dallenpyrah  what exactly do you mean by creating a new variable ? do you mean like storing "office365_calendar" in one variable and then use it to compare?

zomars: We should already have a constant somewhere

Udit-takkar: This code is outdated

PeerRich:
@emrysal you were looking into this

packages/core/EventManager.ts
```diff
@@ -102,12 +107,16 @@ export default class EventManager {
     results.push(...(await this.createAllCalendarEvents(evt)));
 
     const referencesToCreate = results.map((result) => {
+      let createdEventObj: NewCalendarEventType | VideoCallData | null = null;
+      if (typeof result?.createdEvent === "string") {
+        createdEventObj = JSON.parse(result?.createdEvent);
```
zomars, line 112: @Udit-takkar can use `zod` to parse the actual result and keep the type safety on runtime?

Udit-takkar: ok

Udit-takkar:
@zomars  can you check if zod has been used correctly.

zomars (approved):
It is @Udit-takkar Thank you for your contribution 🙏

PR #4014, opened 2022-08-29
Get redirect URLs working again, await signIn() never finished

emrysal:
Idea is to remove onboarding from clientside logic, simplifying our system and instead handle it through NextAuth.

* Fixes redirect URLs being broken post v4, they must be absolute or await signIn() is rejected (and never gets to the follow-up logic)


PeerRich:
close due to inactivity. happy to reopen

PR #4015, opened 2022-08-29
fix/3928-reschedule-unexpected-error

### LINKED ISSUE #3928: reschedule unexpected error
imagepotentially after already rescheduled?

alannnc:
## What does this PR do?

Fixes api/book/event error handling. Now if a syntax error is received in defaultResponder message should be custom and not retrieved to user. Also improved default error message.

Fixes #3928

## Type of change

## How should this be tested?

- [ ] On api/book/event you can use JSON.parse("Error parsing") to be thrown on any part of the code and it should be handled properly.

## Checklist

### DIFF
--- a/packages/lib/server/getServerErrorFromUnknown.ts
+++ b/packages/lib/server/getServerErrorFromUnknown.ts
@@ -33,6 +33,12 @@ export function getServerErrorFromUnknown(cause: unknown): HttpError {
       cause,
     });
   }
+  if (cause instanceof SyntaxError) {
+    return new HttpError({
+      statusCode: 500,
+      message: "Unexpected error, please reach out for our customer support.",
+    });
+  }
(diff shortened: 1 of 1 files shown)


packages/lib/server/getServerErrorFromUnknown.ts
```diff
@@ -33,6 +33,12 @@ export function getServerErrorFromUnknown(cause: unknown): HttpError {
       cause,
     });
   }
+  if (cause instanceof SyntaxError) {
+    return new HttpError({
+      statusCode: 500,
+      message: "Unexpected error, please reach out for our customer support.",
+    });
+  }
```
alannnc, line 41: SyntaxError were being handled and response to user was misleading.

zomars (approved):
Nice catch @alannnc 🙏🏽

PR #4018, opened 2022-08-29
V2 Settings - Security View

joeauyeung:
## What does this PR do?

This PR adds two screens to the security section of the settings for V2.
- Change password
- Enable 2FA

Note that the settings shell is still a WIP so the pages will be missing headers.

Fixes #3784 

**Environment**: Staging(main branch) / Production

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- Visit http://localhost:3000/v2/settings/security/password
	 - Enter an invalid password, an error should appear
 	- Enter a valid password, it should be changed

- Visit http://localhost:3000/v2/settings/security/two-factor-auth
	- Enable 2FA
	- Disable 2FA

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #4021, opened 2022-08-30
Adds Meta component to handle layout headings/metadata

zomars:
To be merged after #4018 

## What does this PR do?

- Adds a `Meta` component that allows to update metadata from any page children
- Exposes metadata via a `useMeta` hook, this can be used to sync page metadata and display it in UI (Like Headings, subtitles, etc.)
- Adds `Meta` to pages in `/v2/settings`
- Adds example usage in `SettingsLayout` as `ShellHeader`

## Type of change

## How should this be tested?

- [ ] Go to `/v2/settings/my-account/profile`
- [ ] Navigate between settings pages and see how headings change

### DIFF
--- a/packages/ui/v2/core/Meta.tsx
+++ b/packages/ui/v2/core/Meta.tsx
@@ -1,0 +1,56 @@
+import Head from "next/head";
+import React, { createContext, useContext, useState } from "react";
+
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+
+type MetaType = {
+  title: string;
+  description: string;
+};
+
+const initialMeta = {
+  title: "",
+  description: "",
+};
+
+const MetaContext = createContext({
+  meta: initialMeta,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  setMeta: (newMeta: Partial<MetaType>) => {},
+});
+
+export function useMeta() {
+  return useContext(MetaContext);
+}
+
+export function MetaProvider({ children }: { children: React.ReactNode }) {
+  const [value, setValue] = useState(initialMeta);
+  const setMeta = (newMeta: Partial<MetaType>) => {
+    setValue((v) => ({ ...v, ...newMeta }));
+  };
+
+  return <MetaContext.Provider value={{ meta: value, setMeta }}>{children}</MetaContext.Provider>;
+}
+
+/**
+ * The purpose of this component is to simplify title and description handling.
+ * Similarly to `next/head`'s `Head` component this allow us to update the metadata for a page
+ * from any children, also exposes the metadata via the `useMeta` hook in case we need them
+ * elsewhere (ie. on a Heading, Title, Subtitle, etc.)
+ * @example <Meta title="Password" description="Manage settings for your account passwords" />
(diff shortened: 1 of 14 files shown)


packages/ui/v2/core/Meta.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { setMeta, meta } = useMeta();
+  /* @TODO: maybe find a way to have this data on first render to prevent flicker */
+  if (meta.title !== title || meta.description !== description) {
+    setMeta({ title, description });
+  }
+
+  return (
+    <Head>
+      <title>{t(title)} | Cal.com</title>
+      <meta name="description" content={t(description)} />
```
PeerRich, line 53: i've noticed that translated descriptions dont work in social previews

we should just use english here, to i18n function @zomars

packages/ui/v2/core/Meta.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { t } = useLocale();
+  const { setMeta, meta } = useMeta();
+  /* @TODO: maybe find a way to have this data on first render to prevent flicker */
+  if (meta.title !== title || meta.description !== description) {
+    setMeta({ title, description });
+  }
+
+  return (
+    <Head>
+      <title>{t(title)} | Cal.com</title>
```
PeerRich, line 52: same for title

PeerRich (changes requested):

packages/ui/v2/core/Meta.tsx
zomars: We don't need social previews here. These are being used in the authed screens.

zlwaterfield: Is there a way to make sure these aren't used on public pages?

packages/ui/v2/core/Meta.tsx
```diff
@@ -0,0 +1,56 @@
+import Head from "next/head";
+import React, { createContext, useContext, useState } from "react";
+
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+
+type MetaType = {
```
zlwaterfield, line 6: My worry with this PR is that the `MetaType` attributes are not typed so if an unsupported title/description is passed in it will use that string directly. And since the strings are snakecase it won't look good. 

For example for the conferencing page if `conferencing2` was passed in by accident, it would pass typing but not match supported within the i18n lookups. 
Screen Shot 2022-08-30 at 12 53 06 PM
Screen Shot 2022-08-30 at 12 53 01 PM

This might be out of scope for this PR but we should look into typing i18n lookups.

We could also add a "Cal.com" backup for the title so it doesn't show up weird if the lookup fails.

packages/ui/v2/core/Meta.tsx
zomars: @zlwaterfield not worth the effort to enforce via linting if you ask me. We can add some comments and make sure it doesn't slip on reviews.

zomars:
Will merge as it is to unblock @sean-brydon and @joeauyeung who we're asking for this

PR #4073, opened 2022-09-01
Bugfix/typescript performance

emrysal:
Typescript investigation today after it became too crashy to develop

> A file named next-env.d.ts will be created at the root of your project. This file ensures Next.js types are picked up by the TypeScript compiler. You should not remove it or edit it as it can change at any time. This file should not be committed and should be ignored by version control (e.g. inside your .gitignore file).

source: https://nextjs.org/docs/basic-features/typescript

It's way too easy to create circular barrel behaviour by internally importing the barrel, this PR removes these imports within @calcom/ui (v2)


sean-brydon (approved):
LGTM - nice catch

PR #4080, opened 2022-09-01
fix: add extra word-break for chromium

### LINKED ISSUE #4076: If event links are too long, user isn’t able to edit them
chrome-capture-2022-7-31

I think this input bugs out once the link is at a certain length.

lgastler:
## What does this PR do?

Adds the `break-all` class (`word-break: break-all;`) to break long text values so the text field does not overflow. Different browsers break the text different with `break-all` applied the text field stops overflowing in all cases. 

Fixes #4076

CleanShot 2022-09-01 at 13 07 35

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)


PeerRich:
thank you, great fix!

Issue #4009, opened 2022-08-29
2.0 Feedback Collection


PeerRich:
- [ ] i18n is flickering 

![CleanShot 2022-08-29 at 13 28 38](https://user-images.githubusercontent.com/8019099/187191367-14c8e968-cf0e-4e7e-a11e-c167362cac5c.gif)

- [ ] missing padding for tablet view

<img width="384" alt="CleanShot 2022-08-29 at 13 29 23@2x" src="https://user-images.githubusercontent.com/8019099/187191449-19bcc18d-9b31-440f-9016-487cbc67f13d.png">

- [ ] SidebarCard is re-rendering when navigating
![CleanShot 2022-08-29 at 13 30 04](https://user-images.githubusercontent.com/8019099/187191632-0a4dba45-545c-4338-b1f2-547a62209613.gif)
 
- [ ] /booking: missing icons and wrong padding  (see [figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=1705%3A81198))

<table>
<tr>
<td>
<img width="789" alt="CleanShot 2022-08-29 at 13 31 00@2x" src="https://user-images.githubusercontent.com/8019099/187191683-6321a1cd-a586-4aec-bee8-aa1e945e6cd2.png">

</td>
<td>
<img width="844" alt="CleanShot 2022-08-29 at 13 31 31@2x" src="https://user-images.githubusercontent.com/8019099/187191762-66aafb2b-7ac0-4e5e-8071-e874f1e3545c.png">

</td>
</tr>
</table>

sean-brydon:
Ill take this - I believe the select font size should be fixed when @alishaz-polymath has finished with the select component.

alishaz-polymath:
Yes, I think so @sean-brydon . This week 🤞

hariombalhara:
I18n flickering and padding fix for tablet is already in Routing Forms v2. 0 PR

PeerRich:
awesome! lets get that merged soon

PR #4086, opened 2022-09-02
V2 for app setup pages

CarinaWolli:
## What does this PR do?

Adds V2 pages with v2 components for all app setup pages. 

Fixes 'App setup pages are still in old design and redirect not working in V2' of #4009


apps/web/pages/v2/apps/[slug]/setup.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const router = useRouter();
+  const slug = router.query.slug as string;
+  const { status } = useSession();
+
+  if (status === "loading") {
+    return (
+      <div className="absolute z-50 flex h-screen w-full items-center bg-gray-200">
+        <Loader />
+      </div>
+    );
```
Comment (PeerRich) [inline apps/web/pages/v2/apps/[slug]/setup.tsx:19]: I believe we wanna use skeletons in the future for everything, no? @zomars

PeerRich (commented):
hey @zomars we wanna get rid of Loaders in favour of Skeletons, right?

apps/web/pages/v2/apps/[slug]/setup.tsx
Comment (zomars) [reply apps/web/pages/v2/apps/[slug]/setup.tsx:19]: Agreed @PeerRich but in this context I don't see the benefit of using skeletons

It's just a blank setup page

image

zomars (approved):

PR #4138, opened 2022-09-03
V2.0 Shell - Progressive Rendering with Skeleton

hariombalhara:
## What does this PR do?

Fixes # (issue)
[Demo Loom](https://www.loom.com/share/143c8669b07f4d9fb5e835c6537e0919)
Also Fixes
- Routing Forms Navigation is not shown as current for Form Builder and Forms list pages
- Fixes move up and down buttons as per new UI
Screenshot 2022-09-03 at 5 15 11 PM

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


PeerRich:
this doesnt look good, just remove that skeleton here
CleanShot 2022-09-05 at 10 41 49@2x

hariombalhara:
@PeerRich Done. Though this will introduce CLS.

PeerRich:
> @PeerRich Done. Though this will introduce CLS.

I see. maybe instead we wanna have the skeleton as the same size as the content? have one skeleton left for headline and one small one for the button on the right

hariombalhara:
Yeah, that's some extra work just for the skeleton. By default, if no content is there in the element, it just collapses. We need a way to not have the element collapse. 
Just thought maybe we can have opacity 0 text in the skeleton element, that would not let it collapse and it won't be visible. This would allow us to just keep using SkeletonText without explicitly hardcoding min-height

PeerRich:
> Yeah, that's some extra work just for the skeleton. By default, if no content is there in the element, it just collapses. We need a way to not have the element collapse. Just thought maybe we can have opacity 0 text in the skeleton element, that would not let it collapse and it won't be visible. This would allow us to just keep using SkeletonText without explicitly hardcoding min-height

that's a good approach

apps/web/components/eventtype/CreateEventType.tsx
```diff
@@ -51,7 +52,7 @@ interface CreateEventTypeBtnProps {
 export default function CreateEventTypeButton(props: CreateEventTypeBtnProps) {
   const { t } = useLocale();
```
zomars, line 53: @hariombalhara why not use this instead?

```suggestion
  const { t, isLocaleReady } = useLocale();
```

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -14,6 +14,11 @@ export function useViewerI18n() {
   });
 }
 
+export function useIsI18nLoading() {
+  const i18n = useViewerI18n();
+  return i18n.status === "loading";
+}
+
```
zomars, line 21: [See my comment](https://github.com/calcom/cal.com/pull/4138/files#r963085478)

zomars (approved):
Removed  `isI18nLoading` in favor of  `isLocaleReady`, tested and it seems good to go. Thanks @hariombalhara please DM me if you feel this was a bad move.

PR #4213, opened 2022-09-06
fixed height of recurring event select

### LINKED ISSUE #4174: /event-type-detail - spacing not being applied between text
I think there's a `pb-2` followed by a `-mt-2` resulting in no space here.
Image

There should be space

Image

Double check all spacing in here while we're here :)

Image

### LINKED ISSUE #4178: /event-types-detail /recurring - I spy an input with 38px height 🧐
Image

PeerRich:
- also increased height between title and descriptioin

before
CleanShot 2022-09-06 at 20 07 36@2x

after
CleanShot 2022-09-06 at 20 06 47@2x


zomars (approved):
LGTM

PR #4217, opened 2022-09-06
changed all "new" buttons, added i18n for routing forms

### LINKED ISSUE #4189: Mobile - Page header
Will add designs for these but creating issue so I don't forget. We need to clean up all page headers on mobile. 

- [x] Buttons should just say "New"
- [x] Probably needs to go below text or we need to hide description 

image

PeerRich:
CleanShot 2022-09-06 at 21 07 53@2x


zomars (approved):
LGTM

PR #4243, opened 2022-09-07
Impersonate UI

### LINKED ISSUE #4241: 2.0 Settings / Admin {View}
https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=18408%3A171954

### LINKED ISSUE #4270: /settings - admin panel is visible for non admins
I tried impersonating my test account I made but it says I don’t have permission which is good but it should be hidden

sean-brydon:
Adds impersonate UI to settings

Also adds a new permission container 

CleanShot 2022-09-07 at 15 50 51


packages/ui/v2/core/PermissionContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+}) => {
+  const session = useSession();
+  if (skip) {
+    return <>{children}</>;
+  }
+  // Admin can do everything
+  if (session.data?.user.role !== roleRequired || !UserPermissionRole.ADMIN) return null;
+  const Component = as ?? Fragment;
+  return <Component {...rest}>{children}</Component>;
+};
```
PeerRich, line 30: this is cool. bascially hides content for non-admins?

PeerRich (approved):
great stuff. love that new admin check component

sean-brydon:
@zomars could you sanity check this admin component before we merge?

packages/ui/v2/core/PermissionContainer.tsx
```diff
@@ -0,0 +1,30 @@
+import { useSession } from "next-auth/react";
+import { FC, Fragment } from "react";
+
+import { UserPermissionRole } from ".prisma/client";
+
+type AdminRequiredProps = {
+  as?: keyof JSX.IntrinsicElements;
+  /** Use this if you want to skip the admin check - used when wrapping reusable components that might require an admin check  */
+  skip?: boolean;
```
zomars, line 9: I'm not 100% convinced with the skip approach but it's ok for now

packages/ui/v2/core/PermissionContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children?: React.ReactNode;
+  /**Not needed right now but will be useful if we ever expand our permission roles */
+  roleRequired?: UserPermissionRole;
+};
+
+export const PermissionContainer: FC<AdminRequiredProps> = ({
```
zomars, line 15: Let move this file to `@calcom/features/permissions`

packages/ui/v2/core/navigation/tabs/HorizontalTabItem.tsx
```diff
@@ -6,9 +6,12 @@ import { MouseEventHandler } from "react";
 import classNames from "@calcom/lib/classNames";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 
+import { PermissionContainer } from "../../PermissionContainer";
```
zomars, line 9: HorizontalTabs should not contain outside logic. We can either create a container with logic or figure another way to achieve this

packages/ui/v2/core/navigation/tabs/HorizontalTabItem.tsx
```diff
@@ -6,9 +6,12 @@ import { MouseEventHandler } from "react";
...
+
 export type HorizontalTabItemProps = {
   name: string;
   disabled?: boolean;
+  adminRequired?: boolean;
```
zomars, line 14: If rather filter out in a parent container than adding logic to a dumb component

packages/ui/v2/core/navigation/tabs/VerticalTabItem.tsx
```diff
@@ -17,6 +19,7 @@ export type VerticalTabItemProps = {
   children?: VerticalTabItemProps[];
   textClassNames?: string;
   className?: string;
+  adminRequired?: boolean;
```
zomars, line 22: Same here

packages/ui/v2/core/navigation/tabs/HorizontalTabItem.tsx
PeerRich: true. i didnt see this

packages/ui/v2/core/PermissionContainer.tsx
zomars: From Twist:

> Also. Instead of the adminRequired prop. You can have an `const adminRequiredKeys = [“impersonation”]` and check against that array

zomars (changes requested):
Blocking until feedback is addressed

PeerRich:
related: https://github.com/calcom/cal.com/issues/4270

it seems like all admin settings are visible to non-admins (backend still checks for permission luckily)

sean-brydon:
> related: #4270
> 
> it seems like all admin settings are visible to non-admins (backend still checks for permission luckily)

Should be fixed in update - will wait for @zomars blessing :)

sean-brydon:
@zomars this creates a content flash if you are not an admin (if you manually navigate to an admin page) - not 100% sure the best way to prevent that do you have any ideas?

zomars (approved):

zomars:
Not ideal but let's follow up instead

PR #4247, opened 2022-09-07
Feature/goal reduce flaky tests

emrysal:
## What does this PR do?

* Attempts to speed up playwright a bit and making it less flaky - by reducing flaky tests that actually prevents additional thread usage due to trying to re-run tests.


apps/web/components/security/ChangePasswordSection.tsx
```diff
@@ -1,17 +1,20 @@
...
-
 const ChangePasswordSection = () => {
   const [oldPassword, setOldPassword] = useState("");
   const [newPassword, setNewPassword] = useState("");
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
   const [isSubmitting, setIsSubmitting] = useState(false);
-  const { t } = useLocale();
+  const { t, isLocaleReady } = useLocale();
+  // hold display until the locale is loaded
+  if (!isLocaleReady) {
```
emrysal, line 15: It works, and didn't want to do much more given it's a v1 test.

zomars: An improvement but we can go deeper an use Skeletons at the field level

zomars: Fair enough

zomars (approved):

PR #4253, opened 2022-09-07
fixes toggle

### LINKED ISSUE #4245: /event-type-detail /recurring - toggle goes down a row when active
Image

agustif:
## What does this PR do?

Fixes #4245 

## Type of change
image

## How should this be tested?
Go to a recurring event, checkbox should stay on top and not move down now


apps/web/components/eventtype/RecurringEventController.tsx
```diff
@@ -47,8 +47,8 @@ export default function RecurringEventController({
           <Alert severity="warning" title={t("warning_payment_recurring_event")} />
         ) : (
           <>
-            <div className="relative flex items-start">
-              <div className="flex h-5 items-center">
+            <div className="absolute flex items-start">
```
zomars, line 50: Is the absolute needed? Shouldn't flex-start suffice?

PeerRich: agree. i dont think absolute should be used here. it would be better to stay flex but in the same container so it doesn't move down

PeerRich (changes requested):
dont use absolute, instead stay in flex but move the switch into the headline container

apps/web/components/eventtype/RecurringEventController.tsx
agustif: oops yes! removing wasn't meant to be on the commit and from previous testing

agustif:
yeah removed that absolute as it wasn't needed, forgot to remove it from previous testing, fixed it now!

PeerRich (approved):

PeerRich:
great stuff, thanks!

PR #4263, opened 2022-09-07
Removes dynamic dimensions for V2 Skeletons in favor of classnames

zomars:
## What does this PR do?

- Adds deprecation notices to V1 Skeletons
- Removes `width` and `height` props from V2 Skeletons
- Migrates V2 skeletons to use classnames

For more context on this decision: https://tailwindcss.com/docs/content-configuration#dynamic-class-names 

## Type of change


zomars (commented):
Ready for review 

Motion Capture Dance GIF by Ankita

sean-brydon (approved):
LGTM - whoever wrote them dynamic classNames should be shot 😅 (It was me)

packages/ui/skeleton/index.tsx
```diff
@@ -1,15 +1,19 @@
...
   height?: string;
   className?: string;
 };
 
 interface AvatarProps extends SkeletonBaseProps {
-  // Limit this cause we don't use avatars bigger than thi
-  width: "2" | "3" | "4" | "5" | "6" | "8" | "12";
-  height: "2" | "3" | "4" | "5" | "6" | "8" | "12";
+  // Limit this cause we don't use avatars bigger than this
+  /** @deprecated @see https://tailwindcss.com/docs/content-configuration#dynamic-class-names */
```
PeerRich, line 13: why show deprecated when we just replace it everywhere?

PeerRich (approved):
does it change anything visually? or just a refactor cleanup?

packages/ui/skeleton/index.tsx
zomars: Not worth the effort now as we will remove it post v2

PR #4276, opened 2022-09-08
fix: use flex

Udit-takkar:
Fixes #4274 

Screenshot 2022-09-08 at 3 55 10 PM


packages/ui/v2/core/Dialog.tsx
```diff
@@ -115,7 +115,7 @@ export const DialogContent = React.forwardRef<HTMLDivElement, DialogContentProps
         )}
         {!props.useOwnActionButtons && (
           <DialogFooter>
-            <div className="mt-2">
+            <div className="mt-2 flex">
```
sean-brydon, line 118:
```suggestion
            <div className="mt-2 flex space-x-2">
```

sean-brydon (dismissed):
LGTM nice simple fix - can we have spacing between the buttons to match the button group we have in Figma :)

sean-brydon (approved):
LGTM

PR #4297, opened 2022-09-08
removed <hr> from event type apps

### LINKED ISSUE #4246: /event-type-single /apps - Shouldn't be a divider when the app is inactive
Current

Image

Should be

Image

[View in Figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=1413%3A66900)

PeerRich:
app settings need to be extracted into /packages/app-store before re-adding


zomars (approved):
LGTM

PR #4319, opened 2022-09-08
Fix #4318 fix getBusyTimes to factor in all users from collective bookings

### LINKED ISSUE #4318: getBusyTimes doesn't consider user busy unless they are the booking owner.
### Issue Summary

With a collective booking, when a meeting is accepted. All people who are attached to that event type should have their availability blocked off for that booking. 

Currently, it will only consider the event owner blocked off. 

### Steps to Reproduce

1. Create a collective booking with more than one member.
2. Book the "collective event"
3. try and book a personal meeting with one of the members that didn't initially create that collective booking during the same time slot as the collective event.

When you got to step 3, every person in the collective event should be considered busy during that time slot.

isaiahdahl:
## What does this PR do?

This PR changes the query for bookings in `getBusyTimes.ts` to look for all bookings where the user is attached to the event, instead of just bookings where the user is the organizer.

Fixes #4318

## Type of change

## How should this be tested?

It's a pretty simple change, I verified it by deploying it to my self-hosted instance and checking that the time slot that wasn't being blocked off for a user is now being blocked off.

The issue that I initially had is now resolved for me.


emrysal (approved):
Excellent find! yes userId is historical (still used to indicate individual ownership) and it looks like this check never got updated to the "new" method of users. eventType.users should always be set and we have done db migrations to ensure this is actually the case; so LGTM

emrysal (changes requested):
Actually this won't work as-is, if you apply the same logic to round-robin you would not be able to book any slots that have been booked with a single person, even though other users may still be available. 🤔

isaiahdahl:
Ahh yes of course. Ok I'll adjust.

isaiahdahl:
@emrysal I pushed an update, pulled code from `packages/trpc/server/routers/viewer/booking.tsx`

Are you able to confirm that when it's a round robin event type, the person who gets booked or selected will be the userId on the booking? Not just the user who created the round robin event?

I still think this code would be right but if it's not updating the userId on the booking when it's round robin, maybe it should be.

PeerRich:
hey @isaiahdahl can you update your branch? it seems like we can't:
CleanShot 2022-09-10 at 17 53 14@2x

isaiahdahl:
@PeerRich yup, done.

isaiahdahl:
@PeerRich or @emrysal Is there anything else I can do to get this merged? I keep pressing the update branch button but it doesn't take long before it's out of date again.

alannnc:
Is this PR still relevant? Or should we close it? @isaiahdahl

isaiahdahl:
@alannnc Yes, it's still relevant. I tried keeping the branch up to date for a while but then no one was taking a look at it so I stopped.

PeerRich:
@emrysal will investigate, sorry for the long wait @isaiahdahl

emrysal:
Closing as we're switching out the underlying logic soon, I've looked at this PR a few times but it requires some work to make it work with all the types of event types - will keep you up to date when we fix this @isaiahdahl

PR #4330, opened 2022-09-09
Add rounded-none to combined buttons by default

### LINKED ISSUE #4329: Find a way to remove combined prop from button
It would be nice if we could just pass button into `<ButtonGroup combined>...` without having to also pass combined to the children also.

sean-brydon:
This was easier than I thought ...

Before
CleanShot 2022-09-09 at 11 02 05@2x

After:
CleanShot 2022-09-09 at 11 02 43@2x

Results are exactly the same
CleanShot 2022-09-09 at 11 02 25@2x


emrysal (approved):
Nice!

PeerRich:
can you also add all instances where this is obsolete and remove it?

PeerRich (approved):
good stuff

PR #4344, opened 2022-09-09
/event-types make switch async

### LINKED ISSUE #4130: [CAL-102] /event-types make switch async
right now it seems like the switch state is based on the return of the server, i think we shouldnt wait for the response and fake the state switch

https://user-images.githubusercontent.com/8019099/188223838-c9cfcc82-3967-4a5c-9362-292279cde06c.mp4

gitstart:
## What does this PR do?

Previously when the hide switch is clicked we wait for a response from the server before the switch is toggled. These changes enable optimistic updates of the hidden property on a switched eventType, and roll-backs the changes if the switching fails. So now the switch and hidden label no longer wait for the server response.

Fixes 

## Type of change

## Demo
https://www.loom.com/share/7907fbdf0bf646ccbe8b27b8d10375a4


sean-brydon (changes requested):
Thanks for your contribution. 

We believe this should be extracted into it's own component so we don't need to refetch all eventTypes for the user. 

e.g Extract Into a EventTypeListItem component for each item in this list - then handle the state locally per component.

gitstart:
Hi @sean-brydon! Thank for the review, will get back to you with the requested changes asap

gitstart:
closing this PR for now, will push a different PR with required changes

PR #4361, opened 2022-09-10
fix: toggle alligment

Udit-takkar:
Screenshot 2022-09-10 at 10 18 58 PM

Now it's fixed for all screen sizes


PeerRich:
thank you!

PR #4327, opened 2022-09-09
V2 password reset design

PeerRich:
haven't tested locally because I dont have a mailer installed.

can someone test locally? 

before:
CleanShot 2022-09-09 at 09 56 28@2x

Updated UI : 

https://user-images.githubusercontent.com/55134778/189358908-7157f6b6-4239-455e-a7c2-39ef0ffdf9ff.mp4

PR #4369, opened 2022-09-11
Feat/3796 new UI for signin

### LINKED ISSUE #3796: 2.0 Sign in {View}
[View in Figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=15301%3A134828)

Image

JeroenReumkens:
## What does this PR do?

Updated v2 design for sign in page. 

Fixes #3796
Screenshot 2022-09-11 at 21 30 35

 Loom Video: https://www.loom.com/share/9416030d53434ecda931db41a2639a6f

## Type of change

## How should this be tested?
- [ ] Make sure design matches Figma design as mentioned in #3796
- [x] https://github.com/calcom/cal.com/pull/4327 has a small overlap with this branch. As soon as that PR is merged to main, I need to retest my work with these changes. 

## Other notes

@sean-brydon  This PR also includes a change for the password peek button. However, personally I'm not really happy with the way I've implemented it (usage of position absolute). However, with the current way the input components are build, I didn't see any other solution. Wondering if you already had an idea in mind? 😄 

PS @PeerRich  As soon as this PR get's merged, the forgot password page will also have this peek button 👍


packages/ui/v2/core/form/fields.tsx
```diff
@@ -246,7 +247,39 @@ export const PasswordField = forwardRef<HTMLInputElement, InputFieldProps>(funct
...
+          type="button"
+          onClick={() => toggleIsPasswordVisible()}>
+          {isPasswordVisible ? (
+            <EyeOff className="h-4 stroke-[2.5px]" />
+          ) : (
+            <Eye className="h-4 stroke-[2.5px]" />
+          )}
+          <span className="sr-only">{textLabel}</span>
+        </button>
+      </Tooltip>
```
sean-brydon, line 280: I believe we can just pass this into `addOnSuffix` prop that is attached to the input field to achieve the same functionality without needing "absolute" @JeroenReumkens

sean-brydon (changes requested):

sean-brydon:
Overall looks amazing - great job! Lets see if we can use the `addonSuffix` props as mentioned in the review to make this a little bit cleaner :)

packages/ui/v2/core/form/fields.tsx
JeroenReumkens: Yeah I tried that. Only thing is the addons are squared off and have a divider line, so that would look quite different than the design. So not sure what @Jaibles thinks about that? 😄  See attached screenshot where the addon is used. Personally I like the design in Figma better.  

Screenshot 2022-09-12 at 11 16 07
Screenshot 2022-09-12 at 11 16 20

Jaibles: Definitely not ideal having the line for this.

sean-brydon: We could push a change to the
CleanShot 2022-09-12 at 10 27 24
the addonFilled option to set border-l-0 ? or create another prop I guess?

JeroenReumkens: Absolutely. I could definitely add another prop if that's the way to go. It does make the component a bit more complex by having all these different props 🤔  Personally I always try to stay away from that if possible. Composability is usually what I opt for rather that 'config' via props. That's why I was a bit hesitant to change this 😛  

At least by having the position absolute now we don't make the input component itself more complicate — so looking at it that way it might not be a very bad choice.

sean-brydon: After talking on slack - we've decided to stick with the current implementation due to the fact this design isn't used in the app that often. If this borderless suffix design keeps coming up we can think of implementing into this component or creating a wrapper like our `EmailInputs`

Awesome work!

sean-brydon: Agreed - not worth it in this instance

sean-brydon (approved):
LGTM - looking super clean!

PR #4386, opened 2022-09-12
Fix event type availability gap 

sean-brydon:
Note to self: Don't blame @emrysal before looking at who caused the issue ;') 

CleanShot 2022-09-12 at 11 44 06@2x

_It was me_

CleanShot 2022-09-12 at 11 43 47@2x


emrysal (approved):
🤣 LGTM

PR #4305, opened 2022-09-08
Fixing app details disconnect button and number of apps

leog:
## What does this PR do?

Trying to fix #4269 and #4257

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

Go to an installed app details page and see the Disconnect button to remove it. Also, trying to fix number of apps in categories in v2 where it should show more apps under certain categories.

Issue #3780, opened 2022-08-10
2.0 Apps / Installed Apps


Jaibles:
[View in Figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=1440%3A65827)

![Image](https://user-images.githubusercontent.com/4536123/185592862-998bf670-4f1f-4f13-8435-26d056eda520.png)

Udit-takkar:
@Jaibles  can you assign this to me?

PR #4397, opened 2022-09-12
Fixes rounded corners in app store

CarinaWolli:
## What does this PR do?

Before: 
Screenshot 2022-09-12 at 10 16 40

After: 
Screenshot 2022-09-12 at 10 16 47

Before: 
Screenshot 2022-09-12 at 10 13 44

After: 
Screenshot 2022-09-12 at 10 09 33

## Type of change

### DIFF
--- a/packages/ui/v2/core/apps/Categories.tsx
+++ b/packages/ui/v2/core/apps/Categories.tsx
@@ -33,7 +33,7 @@ export default function AppStoreCategories({
           <Link key={category.name} href={"/apps/categories/" + category.name}>
             <a
               data-testid={`app-store-category-${category.name}`}
-              className="relative flex rounded-sm bg-gray-100 px-6 py-4 sm:block">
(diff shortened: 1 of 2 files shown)


ciaranha (commented):
nice 👌

leog:
Nice catch @CarinaWolli. It would be best to apply these changes in #4305, which will also have changes for #3780 that may have already have fixed them. I think the features category element is fixed, the other one is not.

CarinaWolli:
> Nice catch @CarinaWolli. It would be best to apply these changes in https://github.com/calcom/cal.com/pull/4305, which will also have changes for https://github.com/calcom/cal.com/issues/3780 that may have already have fixed them. I think the features category element is fixed, the other one is not.

Added changes to #4305 (https://github.com/calcom/cal.com/pull/4305/commits/c0ee624badcb5a4e19b6a199263fcd92baabff53)

PR #4403, opened 2022-09-12
Reload on language change to immediately see result

noahflk:
Before

CleanShot 2022-09-12 at 22 51 03

After

CleanShot 2022-09-12 at 22 52 51

closes  #4104


hariombalhara:
This is already fixed by https://github.com/calcom/cal.com/pull/4431

PR #4406, opened 2022-09-12
Fix event edit link styling issue in Safari

### LINKED ISSUE #4377: V2 - When href is applied to dropdown menu item styling is wrong
When "href" gets applied to menu item it loses all styling

CleanShot 2022-09-12 at 09 43 36

This seems to only happen in safari :')

noahflk:
Before

CleanShot 2022-09-13 at 00 03 06@2x

After

CleanShot 2022-09-13 at 00 02 36@2x

Fixes  #4377

### DIFF
--- a/apps/web/pages/v2/event-types/index.tsx
+++ b/apps/web/pages/v2/event-types/index.tsx
@@ -317,7 +317,6 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
                             <DropdownMenuContent>
                               <DropdownMenuItem>
                                 <DropdownItem
-                                  type="button"
                                   href={"/event-types/" + type.id}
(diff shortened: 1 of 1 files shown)


apps/web/pages/v2/event-types/index.tsx
```diff
@@ -317,7 +317,6 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
                             <DropdownMenuContent>
                               <DropdownMenuItem>
                                 <DropdownItem
-                                  type="button"
```
hariombalhara, line 320: type = button tells browser to not submit the containing form. So, this is required.

hariombalhara (changes requested):

PR #4408, opened 2022-09-12
Bugfix: dynamic import rainbow

hexcowboy:
## What does this PR do?

Previous issue was Rainbow was not dynamically imported so Walletconnect RPCs were sending requests on behalf of users who didn't have Rainbow app installed.

image

With the fix it now dynamically imports RainbowKit so users who don't have RainbowKit app installed (or even enabled) will not see these requests come through:

image

Also I tested to make sure it loads properly when Rainbow app is installed and enabled ✅

## Type of change


hexcowboy:
something looks wrong with this branch lemme retry

PR #4418, opened 2022-09-13
Fix/daterangepicker css

### LINKED ISSUE #4389: [component] simple date selector - calendar arrows layout broken
Image

[View in Figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=13738%3A97458)

Image

alishaz-polymath:
## What does this PR do?

Loads CSS from external CSS file instead of the component css file (which is not loading on prod for some reason)

Fixes #4389 

## Type of change

### DIFF
--- a/packages/ui/v2/core/form/date-range-picker/styles.css
+++ b/packages/ui/v2/core/form/date-range-picker/styles.css
@@ -1,158 +1,250 @@
-.react-daterange-picker > .react-daterange-picker__wrapper {
-  padding: .625rem .75rem;
-  font-size: .875rem;
-  line-height: 1.25rem; 
-  border-radius: .375rem; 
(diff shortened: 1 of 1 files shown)


alishaz-polymath:
Codacy is misidentifying `.react-calendar__navigation button` as `.react-calendar button` which is why the error regarding specificity. We can ignore it. The other error is regarding the use of `!important` which is necessary to over-ride styling set by JS directly

sean-brydon (approved):
LGTM - not a fan of the rule of not using !important so doesn't bother me :D  - we use it in a lot of use reusable components in tailwind and it never complains (probably can't check them due to it not being a CSS rule)

PR #4426, opened 2022-09-13
[V2]Fix Embed for Event Type Pages.

### LINKED ISSUE #4060: /event-types: Embed button has no action
Reimplement Embed button within /event-type/[type] as the current implementation was clashing with the tab navigation that we now use in V2

`tabName` prop should ideally be renamed in the embed component to something else to prevent the clash

CleanShot 2022-08-29 at 13 51 30@2x

hariombalhara:
## What does this PR do?

Fixes #4060
Also, fixes following
- Uses V2 Toast component for Embed notifications
- Adds support in Navbar components to decide from which key tabName should be read from or written to. This allows EventSetup tab(controlled by `tabName`) to be able to show HTML/React/Preview tab(controlled by `embedTabName`)

**Environment**: Production

## Type of change

## How should this be tested?

- [ ] Got to /event-types and check that different types of embed can be selected and their HTML, React code can be copied
- [ ] Do the same for /event-types/[type] page

## Checklist


packages/ui/v2/core/navigation/tabs/VerticalTabItem.tsx
```diff
@@ -1,20 +1,20 @@
...
+import { Fragment, MouseEventHandler } from "react";
 
 // import { ChevronRight } from "react-feather";
 import classNames from "@calcom/lib/classNames";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { SVGComponent } from "@calcom/types/SVGComponent";
 import { Icon } from "@calcom/ui/Icon";
 
-export type VerticalTabItemProps = {
+export type VerticalTabItemProps<T extends string = "tabName"> = {
```
zomars, line 12: Why is this needed? @hariombalhara

packages/ui/v2/core/layouts/BookingLayout.tsx
```diff
@@ -45,10 +45,10 @@ export default function BookingLayout({
     <Shell {...rest}>
       <div className="flex flex-col sm:space-x-2 xl:flex-row">
         <div className="hidden xl:block">
-          <VerticalTabs<"tabName"> tabs={tabs} sticky />
+          <VerticalTabs tabNameKey="tabName" tabs={tabs} sticky />
```
sean-brydon, line 48: Minor nit - should tabName not be the default? and we only override if tabNameKey value is passed int?

hariombalhara:
Incorporated in https://github.com/calcom/cal.com/pull/4477

PR #4428, opened 2022-09-13
[V2] Fix /settings/conferencing - 4421

### LINKED ISSUE #4421: /settings /conferencing - UI issues
- [ ] Weird spacing below app card
- [ ] Button in the app card doesn't look like a square
- [ ] Border radius appearing top & bottom

Current
CleanShot 2022-09-13 at 09 12 14@2x

Should be
CleanShot 2022-09-13 at 09 13 51@2x

sean-brydon:
Before:
CleanShot 2022-09-13 at 12 07 52

After:
CleanShot 2022-09-13 at 12 07 07

### DIFF
--- a/packages/ui/v2/modules/List.tsx
+++ b/packages/ui/v2/modules/List.tsx
@@ -3,18 +3,31 @@ import { createElement } from "react";
 
 import classNames from "@calcom/lib/classNames";
 
-export function List(props: JSX.IntrinsicElements["ul"]) {
+export type ListProps = {
+  roundContainer?: boolean;
+} & JSX.IntrinsicElements["ul"];
+
+export function List(props: ListProps) {
   return (
-    <ul {...props} className={classNames("-mx-4 rounded-sm sm:mx-0 sm:overflow-hidden", props.className)}>
+    <ul
+      {...props}
+      className={classNames(
+        "-mx-4 rounded-sm sm:mx-0 sm:overflow-hidden ",
+        // Add rounded top and bottome if roundContainer is true
+        props.roundContainer && "[&>*:first-child]:rounded-t-md [&>*:last-child]:rounded-b-md ",
+        props.className
+      )}>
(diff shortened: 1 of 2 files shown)


packages/ui/v2/modules/List.tsx
```diff
@@ -3,18 +3,31 @@ import { createElement } from "react";
...
+
+export function List(props: ListProps) {
   return (
-    <ul {...props} className={classNames("-mx-4 rounded-sm sm:mx-0 sm:overflow-hidden", props.className)}>
+    <ul
+      {...props}
+      className={classNames(
+        "-mx-4 rounded-sm sm:mx-0 sm:overflow-hidden ",
+        // Add rounded top and bottome if roundContainer is true
+        props.roundContainer && "[&>*:first-child]:rounded-t-md [&>*:last-child]:rounded-b-md ",
```
sean-brydon, line 17: It would be ideal if we could pass `[&_a]:rounded-none` here but for some reason specificity doesn't like it and ignores.

packages/ui/v2/modules/List.tsx
```diff
@@ -3,18 +3,31 @@ import { createElement } from "react";
...
 }
 
-export type ListItemProps = { expanded?: boolean } & ({ href?: never } & JSX.IntrinsicElements["li"]);
+export type ListItemProps = { expanded?: boolean; rounded?: boolean } & ({
+  href?: never;
+} & JSX.IntrinsicElements["li"]);
 
 export function ListItem(props: ListItemProps) {
-  const { href, expanded, ...passThroughProps } = props;
+  const { href, expanded, rounded = true, ...passThroughProps } = props;
```
sean-brydon, line 30: See comment above 

If the above approached worked we would be able to get away without this additional prop

ciaranha (approved):
Looks great 👏 Keep knocking these out 🚀

PR #4438, opened 2022-09-13
fix: merge order of calendar freebusy

ee07b570:
## What does this PR do?

Bug: availability of lark calendar doesn't working right. 

Why:

const { i18n, isLocaleReady } = useLocale();
  const { slots: _1 } = useSlots({
    eventTypeId: eventType.id,
    usernameList: users,
    startTime: selectedDate?.startOf("day"),
    endTime: selectedDate?.endOf("day"),
    timeZone,
  });
  const { slots: _2, isLoading } = useSlots({
    eventTypeId: eventType.id,
    usernameList: users,
    startTime: browsingDate?.startOf("month"),
    endTime: browsingDate?.endOf("month"),
    timeZone,
  });

const slots = useMemo(() => ({ ..._1, ..._2 }), [_1, _2]);

As shown above, the final available slots are the results of merging slots of query by month and slots of query by day. if _1, and _2 have the same key,  result of _2(i.e. the available slots from querying by month) will overwrite the value of _1(i.e. the available slots from querying by day)

However, query of availability of lark calendar is limited in 14 days. If the dates range is more than 14 days, the api will return parameters error. In this case, available slots are calculated as there are no events occupied and return more slots.

Query of availability of lark calendar by day will response correctly, and the available slots are right. however, it's overwrited by that of query of month.

Therefore, exchange the order of merging ( query by day overwrites query by month ) will correct this bug.

## Type of change

## How should this be tested?

- [ ] Availability from lark and other calendars should be right


joeauyeung:
@emrysal looking to you for review as this touches availability and I know you did a major refactoring

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -169,7 +169,7 @@ const SlotPicker = ({
     timeZone,
   });
 
-  const slots = useMemo(() => ({ ..._1, ..._2 }), [_1, _2]);
+  const slots = useMemo(() => ({ ..._2, ..._1 }), [_1, _2]);
```
joeauyeung, line 172: @emrysal change the order here so merging the single day availability into the monthly one.

emrysal (approved):
Can't fault this, lgtm

PR #4468, opened 2022-09-14
feat: disable arrow button up on first and down on last

### LINKED ISSUE #4460: /event-types - Reordering - disabled up on first and down on last
When hovering over the first event, the up icon button should be disabled; hover over the last event, the down icon button should be disabled.

Current

Image

G3root:
## What does this PR do?

Fixes #4460
cal-disabled

## Type of change

### DIFF
--- a/apps/web/pages/v2/event-types/index.tsx
+++ b/apps/web/pages/v2/event-types/index.tsx
@@ -203,6 +203,8 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
     }
   }, []);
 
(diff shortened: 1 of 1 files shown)


emrysal (approved):
I wonder if there is a neater way of doing this using first: and last: but this works! thanks for this.

emrysal (changes requested):
Actually having run a test; could you modify your PR to instead of disabling just remove the element altogether? Ran a test and it looks great that way:

image

emrysal:
Merged as part of #4481

PR #4472, opened 2022-09-14
Fix #4457 (remove event type action buttons)

### LINKED ISSUE #4457: /event-types /detail - We have an Update and a Save button, should be one
From a design perspective we only need one of these. Ideally just the update button on the top right. Currently are these all separate forms so we can't remove the update button? Would like to know so we can at least remove one of these. It's super confusing right now.

Image

emrysal:
## What does this PR do?

Ditch all the buttons from the /event-types/[type].tsx pages and use the top save button only, as well as implements backPath for providing an alternative to Cancel. Future improvement IMO is stickying the top bar but don’t think that’s blocking.

Fixes #4457

Issues:
* Not blocking / Minor: The top save button should be stickied so it stays visible even on the bigger configuration pages. 
image

* FIXED Mayor: No way to cancel - arrow is not implemented yet.
  image


emrysal:
Using backPath parameter in the Shell this is now allowing returning to the /event-types page. opening PR as the scrolling issue only matters on smaller screens.

pumfleet (approved):
Works great, thank you

PR #4478, opened 2022-09-14
fix: positioning of arrows

### LINKED ISSUE #4464: /routing-forms - reordering alignment is not centered
CleanShot 2022-09-14 at 09 57 48@2x

Udit-takkar:
Fixes: #4464 

Screenshot 2022-09-15 at 2 39 19 AM


emrysal:
It works but the use of -mt-10 does not sit right with me, do you reckon it would be possible to always align to the HR like `<flex items-center><buttons><hr /></endflex>` ? Thinking out loud here

Udit-takkar:
@emrysal no just tried this. the <divider> in FormCard is under route 2 heading in image but we want to center it according to second divider which is in children.

Screenshot 2022-09-20 at 12 02 52 AM

Udit-takkar:
Screenshot 2022-09-20 at 12 10 01 AM


this is how it would look like

If i use 
<div items-center>
   <buttonsDiv>
   {children}
</div>

hariombalhara (approved):

hariombalhara:
Thanks @Udit-takkar

PR #4486, opened 2022-09-15
Sets DB default of weekStart to Monday

### LINKED ISSUE #4380: /availability/single-view - start date doesn't reflects settings
Currently it always seems to start on a Sunday even if my week start date is Monday.

If there's going to be a default I would argue that it should be Monday. Especially since in our onboarding we have it as unavailable by default. This implies Monday is the logical start date and would make much more sense with our copy to all button.

CleanShot 2022-09-12 at 10 59 35@2x

gitstart:
## What does this PR do?
Adds Monday as the start day for week 

## Demo
https://www.loom.com/share/6f82a73d41134add9d9a1b8e700e3505

## Type of change

## How should this be tested?

- [ ] Create new user and verify default value start week


pumfleet (approved):

emrysal:
Thanks for this, but the is talking about applying it to the availability page, for the default is a different discussion but will not fix the issue; @baileypumfleet

emrysal (changes requested):
May require migration, checking

baileypumfleet:
Ah okay makes sense

gitstart:
@baileypumfleet @emrysal Thank you for the review. I guess we misunderstood the issue. Will raise a different PR for fixing #4380

hariombalhara:
Issue fixed in a different PR

PR #4490, opened 2022-09-15
fix prop in react svg Icon

hussamkhatib:
## What does this PR do?

- use correct prop for stoke width 
:x: `stroke-width` 
:heavy_check_mark: `strokeWidth`

## Steps to reproduce 
- I noticed this error in console at `/bookings/upcoming` page.

side note: Not sure why this is not caught by linters


pumfleet (approved):
Thank you for spotting this!

PR #4496, opened 2022-09-15
Add slot interval

sean-brydon:
Adds slot interval into V2 was missing :( 

CleanShot 2022-09-15 at 15 43 33@2x
CleanShot 2022-09-15 at 15 43 49@2x


alishaz-polymath (approved):
Let's go 🚀

pumfleet (approved):
Code looks good to me, haven't tested

PR #4551, opened 2022-09-17
Removing update table user plan

leog:
## What does this PR do?

This is to avoid having the wrong conversion being applied each deployment now that we figured that we don't need every user to be with a FREE plan.

## Type of change


alannnc (approved):
One line change LGTM

zomars (changes requested):
Looks simple enough but I'd rather create a new migration. Blocking until coming up with a fix.

zomars (approved):
done

PR #4559, opened 2022-09-17
2.0 prod hotfixes

PeerRich:


apps/web/components/team/TeamCreateModal.tsx
```diff
@@ -50,7 +49,7 @@ export default function TeamCreate(props: Props) {
               </div>
             </div>
           </div>
-          <form onSubmit={createTeam}>
+          <form>
```
alannnc, line 52: Form is no longer needed as it's submit prop got removed.

packages/features/ee/workflows/pages/v2/workflow.tsx
```diff
@@ -189,45 +187,36 @@ function WorkflowPage() {
       <Shell
         title={workflow && workflow.name ? workflow.name : "Untitled"}
         CTA={
-          !isFreeUser && (
-            <div>
-              <Button type="submit">{t("save")}</Button>
-            </div>
-          )
+          <div>
```
alannnc, line 190: I believe this div is no longer necessary, if you need a wrapper it can be a fragment instead <></>

PR #4566, opened 2022-09-18
fix: DestinationCalendar htmlescape() data.label

agustif:
## What does this PR do?
Hotfix


zomars (changes requested):
Why is this necessary? Let's try to fill the PR template please so we can get full context for reviewing

PR #4577, opened 2022-09-18
Use PremiumTextField in v2

hariombalhara:
- Use PremiumTextField in V2 profile again
- V2 UI fixes as well for PremiumTextField component
- If a user is premium or not, this info is obtained from stripe only now from stripe customer metadata. _This was part of previous PR merge related to signup._
[Loom ](https://www.loom.com/share/356c28a87d3a487eb97d9777d4311882)

### Important
In current flow, if a username is already premium it is assumed that 
- The user might have already paid for it or 
- He has been given premium username for free(like a DB update). 

So, short usernames that are in our seed are considered Premium and they don't require payment to change to a new premium username. This is the reason @alannnc that for pro username it wasn't considering it as an upgrade when you try to choose a premium username. 
_This was done because I analyzed some of the metadata that we have in DB and found out that some of the premium usernames don't have user.metadata.isPremium set to true. It might be due to the reason that user.metadata.isPremium logic might have been added later on_ 

**Follow up**
- Show proper inline messages on successful payment or a failure in payment. They were working earlier but broken in V2, will figure out what needs changing.


apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
```diff
@@ -62,6 +60,13 @@ const obtainNewUsernameChangeCondition = ({
   if (!userIsPremium && isNewUsernamePremium && !stripeCustomer?.paidForPremium) {
     return UsernameChangeStatusEnum.UPGRADE;
   }
+
+  // This code requires payment from already premium usernames(if they don't have payment info in stripe under paidForPremium)
+  // If we are sure that metadata.isPremium is set to true for all existing premium users we can uncomment this code with an additional check of `&& !metadata.isPremium`
+  // if (userIsPremium && isNewUsernamePremium && !stripeCustomer?.paidForPremium) {
+  //   return UsernameChangeStatusEnum.UPGRADE;
+  // }
```
alannnc, line 68: I think there were several attempts to achieve this, so to my knowledge everyone should have it already.
What we did back then was:
- Get list of premium users on STRIPE
- Add metadata key to the user DB with isPremium: true

alannnc: Also every time a subscription was renewed or the username was trying to downgrade we always read from stripe for highest subscription tier active for the single user.

hariombalhara: Thanks @alannnc. I have started rechecking user.metadata.isPremium to decide if payment is required or not.
I am not aware of the entire flow of Stripe payments and webhooks. Can you please test it at your end now.

apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
```diff
@@ -202,6 +195,20 @@ const PremiumTextfield = (props: ICustomUsernameProps) => {
     }
   };
 
+  let paymentMsg = !currentUsername ? (
+    <span className="text-xs text-orange-400">
+      You need to reserve your premium username for {getPremiumPlanPriceValue()}
```
emrysal, line 200: Should this be untranslated?

hariombalhara: Yeah, it makes sense to translate it. Will translate this and other strings in a followup PR, once we have this important PR merged.

zomars (approved):
Ship it

PR #4601, opened 2022-09-19
add German "Teams Translation"

paul-hph:
## What does this PR do?

Fixes # (issue)

notlocalized1
notlocalized

**Environment**: Production

## Type of change

Localization

## How should this be tested?

Let a German review.


pumfleet (approved):

apps/web/public/static/locales/de/common.json
```diff
@@ -1166,5 +1166,7 @@
   "theme_dark": "Dunkel",
   "back_to_signin": "Zurück zur Anmeldung",
   "password_updated": "Passwort aktualisiert!",
-  "pending_payment": "Zahlung ausstehend"
+  "pending_payment": "Zahlung ausstehend",
+  "add_a_team": "Team hinzufügen",
+  "add_a_team": "Team löschen",
```
CarinaWolli, line 1171: @paul-siteway I think this should be "delete_team"

CarinaWolli (changes requested):

apps/web/public/static/locales/de/common.json
paul-siteway: yes! 🙈

CarinaWolli (approved):

CarinaWolli (changes requested):
we should not commit yarn.lock @paul-siteway can you remove the changes?

emrysal:
@paul-siteway The lockfile shouldn't be committed at all, just left as is :)

```bash
git checkout origin/main -- yarn.lock
git commit -m "Undo yarn.lock changes"
```

PR #4615, opened 2022-09-20
Fixes /availability/single-view - start date doesn't reflects settings

### LINKED ISSUE #4380: /availability/single-view - start date doesn't reflects settings

emrysal:
Fixes #4380

### DIFF
--- a/apps/web/pages/v2/availability/[schedule].tsx
+++ b/apps/web/pages/v2/availability/[schedule].tsx
@@ -9,6 +9,7 @@ import { availabilityAsString } from "@calcom/lib/availability";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { stringOrNumber } from "@calcom/prisma/zod-utils";
 import { trpc } from "@calcom/trpc/react";
+import useMeQuery from "@calcom/trpc/react/hooks/useMeQuery";
 import type { Schedule as ScheduleType } from "@calcom/types/schedule";
 import { Icon } from "@calcom/ui";
 import TimezoneSelect from "@calcom/ui/form/TimezoneSelect";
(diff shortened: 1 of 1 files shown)


apps/web/pages/v2/availability/[schedule].tsx
```diff
@@ -132,7 +134,17 @@ export default function Availability({ schedule }: { schedule: number }) {
...
                 </h3>
-                <Schedule control={control} name="schedule" />
+                {typeof me.data?.weekStart === "string" && (
+                  <Schedule
+                    control={control}
+                    name="schedule"
+                    weekStart={
+                      ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"].indexOf(
+                        me.data?.weekStart
+                      ) as 0 | 1 | 2 | 3 | 4 | 5 | 6
```
Comment (hariombalhara) [inline apps/web/pages/v2/availability/[schedule].tsx:144]:
```suggestion
                      )
```

Comment (hariombalhara) [reply apps/web/pages/v2/availability/[schedule].tsx:144]: weekStart prop is a `number` type and thus it doesn't need type assertion
Also, the result can be -1 as well so assertion isn't correct.

hariombalhara (approved):

apps/web/pages/v2/availability/[schedule].tsx
Comment (emrysal) [reply apps/web/pages/v2/availability/[schedule].tsx:144]: This is fair, it's a copy paste from similar places (I'm debating with myself whether to move weekStart logic into the schedule component itself).

Comment (hariombalhara) [reply apps/web/pages/v2/availability/[schedule].tsx:144]: maybe we can have MySchedule Higher Order Component that can have user specific settings preconfigured.

PR #4616, opened 2022-09-20
Google Cal return 401 if permissions not given

joeauyeung:
## What does this PR do?

This PR returns a 401 error when a user is connecting a Google Calendar but does not give the proper permissions. This was creating null credentials and breaking the connected calendars screen.

Fixes # (issue)

## Type of change

## How should this be tested?

- Go to connect a Google Calendar
- Do not check off the permissions and hit continue

## Checklist

### DIFF
--- a/packages/app-store/googlecalendar/api/callback.ts
+++ b/packages/app-store/googlecalendar/api/callback.ts
@@ -38,6 +38,8 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     const token = await oAuth2Client.getToken(code);
 
     key = token.res?.data;
+
+    if (!key) res.status(401).json({ message: "Permissions not granted" });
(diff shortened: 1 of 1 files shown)


packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -38,6 +38,8 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     const token = await oAuth2Client.getToken(code);
 
     key = token.res?.data;
+
+    if (!key) res.status(401).json({ message: "Permissions not granted" });
```
zomars, line 42:
```suggestion
    if (!key) return res.status(401).json({ message: "Permissions not granted" });
```

zomars (approved):

Issue #4621, opened 2022-09-20
[v2 Component] Input prefix & suffix components spacing is too large


Jaibles:
Currently
![Image](https://user-images.githubusercontent.com/4536123/191232554-34ced2ab-cf64-4cc7-81e5-03bc065d7c1e.png)

Should be

![Image](https://user-images.githubusercontent.com/4536123/191234387-b87ebf25-beb2-48cb-9c7f-0d2a2b9b4761.png)

imkrunal:
#4623 @Jaibles

PR #4623, opened 2022-09-20
fix: spacing issue in suffix and prefix

krunalcodes:
## What does this PR do?

Fixes # (issue)

#4621 
Also added border radius to suffix check screenshot.

image

## Type of change

## How should this be tested?

- [x] Check the spacing of input suffix and prefix in multiple pages.
- [x] Check border radius of suffix


packages/ui/v2/core/form/fields.tsx
```diff
@@ -208,15 +208,15 @@ const InputField = forwardRef<HTMLInputElement, InputFieldProps>(function InputF
             className={classNames(
               "h-9 border border-gray-300",
               addOnFilled && "bg-gray-100",
-              addOnLeading && "rounded-l-md border-r-0 px-3",
-              addOnSuffix && "border-l-0 px-3"
+              addOnLeading && "rounded-l-md border-r-0 bg-gray-100 px-3",
```
hariombalhara, line 211: Why add `bg-gray-100` specifically for addOnLeading.

I think if addOnLeading or addOnSuffix is true then addOnFilled should be set to true and then bg-gray-100 will be automatically added.

hariombalhara (changes requested):

packages/ui/v2/core/form/fields.tsx
imkrunal: My bad i thoguht i wasn't applied. Update the pr.

PR #4634, opened 2022-09-20
fix: skeleton loader issue in settings/api-keys

### LINKED ISSUE #4585: Skeleton loader for settings/api-keys empty state looks wrong
### Issue Summary

That's how it currently looks: 
Screenshot 2022-09-18 at 16 33 48

But it should look the same as the empty state of settings/webhooks

G3root:
## What does this PR do?

Fixes #4585

https://www.loom.com/share/0a3d932cf17c49fab70aee33c47e25d9


apps/web/pages/v2/settings/developer/api-keys.tsx
```diff
@@ -45,10 +45,10 @@ const ApiKeysView = () => {
         <>
           {isLoading && <SkeletonLoader />}
           <div>
-            {data?.length ? (
+            {isLoading && data === undefined ? null : !isLoading && data && data.length > 0 ? (
```
emrysal, line 48:
```suggestion
            {isLoading && !!data?.length ? (
```

hariombalhara, line 48:
```suggestion
            {isLoading ? null : data?.length ? (
```

hariombalhara (dismissed):

apps/web/pages/v2/settings/developer/api-keys.tsx
hariombalhara: It can be simplified as above.

G3root: that's better, i'll push the requested changes.

emrysal (approved):

hariombalhara (approved):

PR #4644, opened 2022-09-21
Fixes password prompt race condition

emrysal:
## What does this PR do?

FormMethods is created with default values that can sometimes be "" due to the user not being loaded. Because tests run very fast this can cause the form to be submitted before the default value had a chance to update from the initial draw. When this happens the form value != the email and the password prompt is shown.

### DIFF
--- a/apps/web/pages/v2/settings/my-account/profile.tsx
+++ b/apps/web/pages/v2/settings/my-account/profile.tsx
@@ -1,5 +1,4 @@
 import crypto from "crypto";
-import { reset } from "mockdate";
 import { signOut } from "next-auth/react";
 import { useRef, useState, BaseSyntheticEvent, useEffect } from "react";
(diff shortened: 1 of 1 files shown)


apps/web/pages/v2/settings/my-account/profile.tsx
```diff
@@ -124,15 +124,26 @@ const ProfileView = () => {
...
-  const formMethods = useForm({
-    defaultValues: {
-      avatar: user?.avatar || "",
-      username: user?.username || "",
-      name: user?.name || "",
-      email: user?.email || "",
-      bio: user?.bio || "",
-    },
-  });
+  const formMethods = useForm<{
```
hariombalhara, line 127: Going forward, suspense mode might be a better approach to avoid unnecessary useEffect.

hariombalhara (approved):

PR #4646, opened 2022-09-21
Run available user check if event type does not have seats

joeauyeung:
## What does this PR do?

Available users should be run if there are no seats on the event type.

Fixes # (issue)

## Type of change

## How should this be tested?

- Add seats to an event type
- Book the same time slot twice to take up two seats

## Checklist


apps/web/components/v2/eventtype/EventAdvancedTab.tsx
```diff
@@ -308,6 +308,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
                 checked={seatsInputVisible}
                 onCheckedChange={(e) => {
                   setSeatsInputVisible(e);
+                  formMethods.setValue("seatsPerTimeSlot", seatsPerTimeSlot);
```
joeauyeung, line 311: Set default seats on event type when switch is clicked

apps/web/pages/api/book/event.ts
```diff
@@ -327,20 +327,20 @@ async function handler(req: NextApiRequest) {
...
-    },
-    {
-      dateFrom: reqBody.start,
-      dateTo: reqBody.end,
-    }
-  );
-
-  // Add an if conditional if there are no seats on the event type
-  // Assign to only one user when ROUND_ROBIN
   if (!eventType.seatsPerTimeSlot) {
```
joeauyeung, line 330: We should only run `ensureAvailableUsers` if there are no seats. Currently since the user that made the event type is counted as "booked" if there is already one attendee.

emrysal (approved):

PR #4675, opened 2022-09-23
Hotfix: Be more strict with safeRedirectUrl check

hariombalhara:
It assumes that the following URLs are in this format https://example.com (which is origin of a website) . .env.example confirms it. But if someone has accidentally used something like https://example.com/ (note terminating slash), then safeRedirectUrl would return WebApp URL.

CONSOLE_URL
WEBAPP_URL
WEBSITE_URL

Screenshot 2022-09-23 at 12 31 23 PM

For more details refer to internal thread: https://twist.com/a/195135/ch/589886/t/3786075/


packages/lib/getSafeRedirectUrl.ts
```diff
@@ -9,8 +9,10 @@ export const getSafeRedirectUrl = (url = "") => {
     throw new Error("Pass an absolute URL");
   }
 
+  const urlParsed = new URL(url);
+
   // Avoid open redirection security vulnerability
-  if (![CONSOLE_URL, WEBAPP_URL, WEBSITE_URL].some((u) => url.startsWith(u))) {
+  if (![CONSOLE_URL, WEBAPP_URL, WEBSITE_URL].some((u) => u === urlParsed.origin)) {
```
emrysal, line 15:
```suggestion
  if (![CONSOLE_URL, WEBAPP_URL, WEBSITE_URL].some((u) => new URL(u).origin === urlParsed.origin)) {
```

emrysal (changes requested):

packages/lib/getSafeRedirectUrl.ts
hariombalhara: It's better this way. Thanks

emrysal (approved):

PR #4676, opened 2022-09-23
fix: fixes multiple issues in routing form

G3root:
## What does this PR do?
this PR fixes multiple issues in routing forms. issues fixed:

- [x] revalidate existing data after adding a form.
- [x] revalidate existing data after deleting a form and update optimistically.
- [x] add correct translation key in new form modal description.

issues:
https://www.loom.com/share/f9bce957b58b441ea1330f94e0ad801c
https://www.loom.com/share/2ebd792efeb2448983168f08bfa0db15

after:
https://www.loom.com/share/a827a228b3e84681815a2d5a9b2ce549

## Type of change


hariombalhara:
Thanks @G3root for the untranslated content fix and the PR.

This PR seems to address 2 issues
1. Untranslated content which is working perfectly fine with your fix.
2. Forms list not updated with the just created Form. _This particular issue doesn't replicate on local._ See this loom https://www.loom.com/share/cab4b60caa1f40b594ecc6e40bef1837 for main branch on my local.

But I do see that issue 2 does occur on production but only when a Form is created for the first time. On, consecutive attempts forms list update just fine. See loom https://www.loom.com/share/94d7a3b051194dda88d79f89cc633d70

Forms list updates automatically because the page is SSR and whenever someone goes to the page a .json request is sent by Next.js automatically that updates the forms list page. But this behaviour doesn't seem to trigger for the very first Form creation.


Theoretically, your fix of switching to tRPC(with SSR'd data as initial data) would work, but I don't want to involve tRPC on SSR pages which should already have the data we are fetching through tRPC

hariombalhara:
If you can open a PR for untranslated content, that can be merged quickly

G3root:
> Theoretically, your fix of switching to tRPC(with SSR'd data as initial data) would work, but I don't want to involve tRPC on SSR pages which should already have the data we are fetching through tRPC

yeah you are right, involving tRPC here is might not be the best. the reason i went on with this approach is because, this path `viewer.app_routing_forms.forms` is revalidated every time a form is created or deleted even though tRPC is not involved.

another advantage involving tRPC is that we can make use of optimistic-updates. what's your thought's on that ?

hariombalhara:
> this path viewer.app_routing_forms.forms is revalidated every time a form is created or deleted even though tRPC is not involved.
Yeah, that's dead code. It should be removed. 

> another advantage involving tRPC is that we can make use of optimistic-updates.
That's a valid point. We are not using optimistic updates in general in the app. But I do think they would provide good UX.

I will take a look at this PR again for the Optimistic updates approach.

hariombalhara (approved):

PR #4679, opened 2022-09-23
Fix min booking notice

### LINKED ISSUE #4678: /event type /limits - Minimum notice period does not appear to be saving
https://user-images.githubusercontent.com/4536123/191956436-ad177e6c-2582-464d-b3c4-75f31ab456dd.mp4

Also had report of a user saying it was capped at 120 minutes. But I suspect the placeholder threw him off. Will update that placeholder design to be lighter.

> That being said, it looks like the minimum booking notice is now capped at 120 minutes. When I had last used the product and set up my event types, I had this set to at least three days in advance.

sean-brydon:
Fixes min booking notice 

Adds minute field as a v2 field component

After:

CleanShot 2022-09-23 at 13 00 58@2x

GIF:

CleanShot 2022-09-23 at 13 01 36

### DIFF
--- a/apps/web/components/v2/eventtype/EventLimitsTab.tsx
+++ b/apps/web/components/v2/eventtype/EventLimitsTab.tsx
@@ -6,7 +6,7 @@ import { useFormContext, Controller, useWatch } from "react-hook-form";
 import { classNames } from "@calcom/lib";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { PeriodType } from "@calcom/prisma/client";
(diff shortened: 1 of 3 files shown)


emrysal (changes requested):
TextField was saving but there was an uncommitted change in the event-types/[type].tsx (defaultValues) - This field should be uncontrolled, will make amends.

emrysal (approved):

emrysal:
* The MinutesField is now uncontrolled
* defaultValue properly supplied (cause of the error)
* Ring offset has been adjusted to mirror the normal TextField (We may need to offset-1 the Selects)

PR #4682, opened 2022-09-23
Feature/deprecate title translations

emrysal:
Removes the _title translation for the sake of consistency. It's not needed as all languages support a shared format.


zomars (approved):
LGTM 

Loving a good cleanup

Issue #4643, opened 2022-09-20
[CAL-123] /settings sidebar doesn't match figma designs


Jaibles:
There are some subtleties to spacing in the settings sidebar that make it easier to scan, particularly the sectioning of different settings areas that are not reflected from the Figma designs.

[View in Figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=17937%3A171647)

<img width="1148" alt="CleanShot 2022-09-20 at 22 06 16@2x" src="https://user-images.githubusercontent.com/4536123/191364539-ab7a6433-c505-4667-a28e-e807f8b2114e.png">

harshsinghatz:
Can i work on this?

Jaibles:
Yep go ahead @harshsinghatz

harshsinghatz:
Sure

Udit-takkar:
<img width="593" alt="Screenshot 2022-09-22 at 2 04 25 AM" src="https://user-images.githubusercontent.com/53316345/191605040-f8e99ff8-ca63-45a9-8703-7f94e4e07b61.png">

one more thing in mobile sidebar the team logo is not centered in teams.

harshsinghatz:
I will take a look at that as well!

AdityaPainuli:
Hey is anyone working on this issue or can I work on this one ?

harshsinghatz:
> Hey is anyone working on this issue or can I work on this one ?
> 

Currently, I am working on it

AdityaPainuli:
> > Hey is anyone working on this issue or can I work on this one ?
> 
> Currently, I am working on it

Ok no worries 👍

Jaibles:
There's also a line-height issue with the nav items causing it to not be vertically centered.

[35e7a9ac-dc46-473b-ab9b-9a3ca5bf6fb2](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/ae93d516-cbea-4285-b0f0-d8d924b7afa2/35e7a9ac-dc46-473b-ab9b-9a3ca5bf6fb2)

harshsinghatz:
@Jaibles Got it!, is it a priority fix? or can i do this in a few days?
Because currently my exams are going on, so i would be active after the exams are done!

Jaibles:
in a few days is ok! thanks @harshsinghatz  🙏 and good luck in the exams :)

[...]

PR #4705, opened 2022-09-25
/settings sidebar doesn't match figma designs fix

harshsinghatz:
## What does this PR do?

changes made:
- fix: the spacing in the settings sidebar (according to the figma design given in the issue mentioned below)
- fix: team logo positioning in mobile settings sidebar and also give round shape (according to design)

Fixes # (issue)
[Related Issue](https://github.com/calcom/cal.com/issues/4643) - Related issue

image

image

## Type of change

## How should this be tested?

- [ ] E2E
- [ ] Local testing ( navigate to settings tab and verify for the figma design changes )


packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -120,44 +120,46 @@ const SettingsSidebarContainer = ({ className = "" }) => {
...
             textClassNames="text-md font-medium leading-none text-black"
+            className="my-6"
           />
         </div>
         {tabsWithPermissions.map((tab) => {
           return tab.name !== "teams" ? (
             <React.Fragment key={tab.href}>
-              <div>
-                <div className="group flex h-9 w-64 flex-row items-center rounded-md px-3 py-[10px] text-sm font-medium leading-none text-gray-600 hover:bg-gray-100  group-hover:text-gray-700 [&[aria-current='page']]:bg-gray-200 [&[aria-current='page']]:text-gray-900">
+              <div className={`${!tab.children?.length ? "!mb-3" : ""}`}>
```
sean-brydon, line 133: Overall looks good - great job. However, I can't figure out this line is doing -> I've tried removing it and have noticed no difference.

packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -120,44 +120,46 @@ const SettingsSidebarContainer = ({ className = "" }) => {
...
                     disableChevron
                   />
                 ))}
               </div>
             </React.Fragment>
           ) : (
             <React.Fragment key={tab.href}>
-              <div>
-                <div className="group mt-2 flex h-9 w-64 flex-row items-center rounded-md px-3 py-[10px] text-sm font-medium leading-none text-gray-600 hover:bg-gray-100  group-hover:text-gray-700 [&[aria-current='page']]:bg-gray-200 [&[aria-current='page']]:text-gray-900">
+              <div className={`${!(tab.children ?? []) && "!mb-3"}`}>
```
sean-brydon, line 157: Same as above

packages/ui/v2/core/layouts/SettingsLayout.tsx
harshsinghatz: For example, if there are no children in the tab, we would add the "mb-3" class, but if there are some children, we don't want the extra margin. The same logic applies below

sean-brydon: Okay sweet - is there any reason it is set as important? There is nothing else assigning mb on this div so I don't believe it is needed

packages/ui/v2/core/layouts/SettingsLayout.tsx
harshsinghatz: Here I think there is a mistake, I should check for the length and not the array. I will fix this!!!

packages/ui/v2/core/layouts/SettingsLayout.tsx
harshsinghatz: Actually i tried removing it, but it doesn't work that way (not sure about the reason)

packages/ui/v2/core/layouts/SettingsLayout.tsx
harshsinghatz: I will remove the important here

packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -120,44 +120,46 @@ const SettingsSidebarContainer = ({ className = "" }) => {
...
+                    className={`my-0.5 h-7 ${tab.children && index === tab.children?.length - 1 && "!mb-3"}`}
                     disableChevron
                   />
                 ))}
               </div>
             </React.Fragment>
           ) : (
             <React.Fragment key={tab.href}>
-              <div>
-                <div className="group mt-2 flex h-9 w-64 flex-row items-center rounded-md px-3 py-[10px] text-sm font-medium leading-none text-gray-600 hover:bg-gray-100  group-hover:text-gray-700 [&[aria-current='page']]:bg-gray-200 [&[aria-current='page']]:text-gray-900">
```
sean-brydon, line 156: Also I'm pretty sure we need these styles still 

So we can have active/hover states on the items

sean-brydon (changes requested):
Please also remove yarn.lock from this commit - not needed

packages/ui/v2/core/layouts/SettingsLayout.tsx
harshsinghatz: Every other styles are still present

harshsinghatz:
@sean-brydon Done with changes!, lmk if there are any other feedback changes

packages/ui/v2/core/layouts/SettingsLayout.tsx
harshsinghatz: I reverted the yarn.lock with the master version

sean-brydon: > Every other styles are still present

Must have been a git bug - showed the line as deleted for me :)

sean-brydon (changes requested):
LGTM - nice job man :) - 

Yarn.lock still seems to be present for some reason - looks like a new line

Will approve once this has been fixed

harshsinghatz:
> LGTM - nice job man :) -
> 
> Yarn.lock still seems to be present for some reason - looks like a new line
> 
> Will approve once this has been fixed

Thanks:)
Changes are done yarn.lock is not present now

sean-brydon (approved):
Awesome :)

harshsinghatz:
Sean, just a doubt. When will this PR be merged I am unfamiliar with the process (in cal.com)

sean-brydon:
> Sean, just a doubt. When will this PR be merged I am unfamiliar with the process (in cal.com)

This will be merged into main once all checks have passed and branch is up to date - then it will be in production in our next release :) Thanks again

Issue #4700, opened 2022-09-24
update "yarn app-store create" to use the new v2 /event-types section


PeerRich:
right now adding apps to this section requires manual effort:
<img width="1301" alt="CleanShot 2022-09-24 at 12 35 25@2x" src="https://user-images.githubusercontent.com/8019099/192098275-25a83177-18d0-42b2-b8fd-b44d1ca977a3.png">

that should be covered by the CLI (or at least pre-generated and easy to edit afterwards)

PeerRich:
the entire app sections should also live in packages/app-store/ and not hard coded into /event-types

PeerRich:
relevant for https://github.com/calcom/cal.com/pull/4701

PR #4727, opened 2022-09-28
Introduce EventTypeAppCard in app-store and make it super easy to add it through CLI - Also adds Fathom app

### LINKED ISSUE #4221: /event-types-detail - Active apps count is wrong
Currently says 2 active when 0 are active. Should say "2 apps, 0 active"

Image

### LINKED ISSUE #4792: Rainbow App not visible under Apps in EventType even if installed
This happens when it's the only installed app(Stripe and Giphy are not installed)

hariombalhara:
## What does this PR do?

Fixes #4221 #4700 #4791 #4792

Also, review https://app.gitbook.com/o/6snd8PyPYMhg0wUw6CeQ/s/VXRprBTuMlihk37NQgUU/~/changes/6xkqZ4qvJ3Xh9k8UaWaZ/engineering/product-specs/app-store#user-apps docs. Will keep on updating it as and when required.

_Note: This PR also deletes some V1 files whose V2 versions I was modifying._

## Important
- Because we are parsing metadata schema with zod now, it's possible that some users might have metadata(already existing) that might not be meeting zod schema and thus that might fail eventTypes loading.  Though zod schema is as per the current code but it's possible that code might have evolved and there might be data not meeting current code expectations.
 **Recovery Plan in case the issue comes**
  - If many people are being impacted, revert. Because there are no migrations in it, it is completely safe to revert this change. Any event types that make an update to Apps in the mean time, wouldn't have that change reflected but as soon as we make the change live again, they should be able to see their configurations back.
  - If few cases appear, go and fix their data as per the schema

## Design Principle-1:  One App shouldn't break another App. 
- Thanks to ErrorBoundaries, error in one AppCard won't impact another. 
	 e.g. why should Stripe App issue make other Apps unavailable?
	Screenshot 2022-10-01 at 3 06 36 PM
- Each app can only modify its respective slot under the `metadata.apps` namespace. This avoids the chances of one app breaking the other app with name conflicts on properties.
- Moves rainbow, stripe and giphy app-related code from EventAppsTab to their respective app folders.
	- Deprecates eventType.price and eventType.currency as these are specific to stripe only. Migration to drop the columns will follow later.
- Replaced all occurrences of` eventType.price` and `eventType.currency` to `stripeAppData.price` and  `stripeData.currency` - This requires testing of all pages involved as payment is an important flow.

## Design Principle-2: While adding a new app, the developer shouldn't worry about code beyond the app directory
- So, the entire app code exists in app-store/APP_NAME only.

## Design Principle-3: Adding new apps and maintenance should be easy. See [Loom](https://www.loom.com/share/f9c104cdcb8d4ae8baedd0a9a9e7e9ca)
- Adding a new EventType app should be super easy and quick.
- Limited the ways app data can be accessed. getEventTypeAppData is the only accepted entry point to access event-type app data. This allows capabilities like denying access to data if the app is disabled.
- AppStore CLI has the support to instantly create such an app which can be toggled
- Introduced the concept of `extensions` in AppStore(to extend EventTypes, Routing Forms and other features) that can be specified using `extendsFeature` property. Currently it supports only EventType.
		-` EventTypeAppCard.tsx` file would automatically be generated for Apps with `extendsFeature` set to `EventType`
		- `zod.ts` would also be generated that would allow the app to expose the data it wants to store in eventType.metadata as `appDataSchema` export
		- apps.browser.generated.tsx would have the required exports automatically generated for ease of use.

## Design Principle-4:  Easy discoverability of apps(See [Loom](https://www.loom.com/share/aedeeeb71abf4e1eb9d16f06fba07c1e))
- Screenshot 2022-10-03 at 3 59 11 PM
- Not installed apps, would also show now under AppsTab allowing users to easily discover apps relevant to EventType within that view itself.
- User would be able to do mostly a one-click install of apps using `OmniInstallAppButton` from the AppsTab page itself. Giphy and Rainbow would be one click but because stripe requires authentication it can't be one click.
- AppCard now links to the app in app-store so that users can get more details about the app quickly.

## Other Fixes
- UX Improvement: Disabling doesn't discard app data now, that you would have spent time in configuring. See this [loom](https://www.loom.com/share/62c8409174454956b74f6aa6bc33e4c9) for a demo
- Updated EventTypeDescription to V2 on event-types listing page. This is consistent with booking pages now.
Before:
Screenshot 2022-10-01 at 2 42 47 PM
After:
Screenshot 2022-10-01 at 2 44 33 PM

Before:
Screenshot 2022-10-01 at 3 42 12 PM

After:
Screenshot 2022-10-01 at 3 41 43 PM

## Type of change

## How should this be tested?

- [x] Install Rainbow, Stripe, Giphy apps and verify their impact on booking pages.
- [x] Enable these apps and then verify their impact
- [x] Configure these apps with various options.
- [x] Test that stripe installation n correctly resets price to 0 in metadata.apps

## Checklist


packages/ui/ErrorBoundary.tsx
```diff
@@ -20,7 +20,7 @@ class ErrorBoundary extends React.Component<
       // Error path
       return (
         <div>
-          <h2>Something went wrong.</h2>
+          <h2>{this.props.message || "Something went wrong."}</h2>
```
hariombalhara, line 23: Support custom message in ErrorBoundary

apps/web/components/v2/eventtype/EventRecurringTab.tsx
```diff
@@ -1,25 +1,18 @@
...
 
-export const EventRecurringTab = ({
-  eventType,
-  hasPaymentIntegration,
-}: Pick<EventTypeSetupInfered, "eventType" | "hasPaymentIntegration">) => {
-  const requirePayment = eventType.price > 0;
-  const [recurringEventDefined, setRecurringEventDefined] = useState(
-    eventType.recurringEvent?.count !== undefined
-  );
+export const EventRecurringTab = ({ eventType }: Pick<EventTypeSetupInfered, "eventType">) => {
```
hariombalhara, line 8: Removed some unused code

apps/web/components/v2/eventtype/RecurringEventController.tsx
```diff
@@ -11,13 +11,11 @@ import { Label, Select, Switch } from "@calcom/ui/v2";
 type RecurringEventControllerProps = {
   recurringEvent: RecurringEvent | null;
   paymentEnabled: boolean;
-  onRecurringEventDefined: (value: boolean) => void;
 };
 
 export default function RecurringEventController({
   recurringEvent,
   paymentEnabled,
-  onRecurringEventDefined,
```
hariombalhara, line 20: unused

hariombalhara (commented):
Self review done

packages/ui/ErrorBoundary.tsx
leog: Should "Something went wrong." be translated? 🤔

hariombalhara: I think it should be. Not sure how to re-use a hook(useLocale) code in class-based component

leog (approved):
Left some comments, nothing blocking I think. Great changes!

apps/web/pages/v2/event-types/[type]/index.tsx
```diff
@@ -159,9 +164,20 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
       },
       schedulingType: eventType.schedulingType,
       minimumBookingNotice: eventType.minimumBookingNotice,
+      metadata: eventType.metadata,
     },
   });
 
+  const appsMetadata = formMethods.getValues("metadata")?.apps;
```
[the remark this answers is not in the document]
Comment (hariombalhara) [reply apps/web/pages/v2/event-types/[type]/index.tsx:171]: Agreed, eventType.metadata.apps should be called appData

zomars (approved):
This PR is a bold move. But a welcome one nonetheless. Let's ship this to staging and test the heck out of it before going to prod on Monday.

Impressive work @hariombalhara 🙏

Ship it

PR #4731, opened 2022-09-28
Fixes for advanced tabs form logic

zomars:
## What does this PR do?

- Adds disabled cursor for disabled switches
- Fixes logic for seated events

https://user-images.githubusercontent.com/3504472/192757971-749b3f1c-6d58-47a9-8040-641732168aa7.mov

## Type of change

## How should this be tested?

- [ ] Go to an event type edit view and play with the advanced type values. Should be working as expected


apps/web/components/v2/eventtype/EventAdvancedTab.tsx
```diff
@@ -42,7 +42,6 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
   const [hashedLinkVisible, setHashedLinkVisible] = useState(!!eventType.hashedLink);
   const [hashedUrl, setHashedUrl] = useState(eventType.hashedLink?.link);
   const [seatsInputVisible, setSeatsInputVisible] = useState(!!eventType.seatsPerTimeSlot);
-  const [seatsPerTimeSlot, setSeatsPerTimeSlot] = useState(eventType.seatsPerTimeSlot || 2);
```
zomars, line 45: When using form, we should not rely on `useState` as much as possible. We already have a "Form State" we should take advantage of it as much as possible.

apps/web/components/v2/eventtype/EventAdvancedTab.tsx
```diff
@@ -313,45 +312,57 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
       />
       <hr />
       <Controller
-        name="seatsPerTimeSlot"
+        name="seatsPerTimeSlotEnabled"
```
zomars, line 315: Added a pseudo-field that won't be send to the backend but will allow us to keep track of this state locally.

apps/web/components/v2/eventtype/EventAdvancedTab.tsx
```diff
@@ -313,45 +312,57 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
...
+              onCheckedChange={(e) => {
+                setSeatsInputVisible(e);
+                // Enabling seats will disable guests and requiring confimation until fully supported
+                if (e) {
+                  formMethods.setValue("disableGuests", true);
+                  formMethods.setValue("requiresConfirmation", false);
+                  formMethods.setValue("seatsPerTimeSlot", 2);
+                } else {
+                  formMethods.setValue("seatsPerTimeSlot", null);
+                }
```
zomars, line 332: Here's the actual logic, we don't need to manipulate external state, just the actual form values.

apps/web/components/v2/eventtype/EventAdvancedTab.tsx
```diff
@@ -313,45 +312,57 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
...
+            <div className="flex flex-col">
+              <Label className="text-sm font-semibold leading-none text-black">{t("offer_seats")}</Label>
+              <p className="-mt-2 text-sm leading-normal text-gray-600">{t("offer_seats_description")}</p>
+            </div>
+          </div>
+        )}
+      />
+      {seatsInputVisible && (
+        <Controller
+          name="seatsPerTimeSlot"
```
zomars, line 346: The Switch (which is a boolean) was manipulating a number value. This makes it right.

packages/ui/v2/core/Switch.tsx
```diff
@@ -21,7 +21,8 @@ const Switch = (
     <div className={classNames("flex h-auto w-auto flex-row items-center", props.fitToHeight && "h-fit")}>
       <PrimitiveSwitch.Root
         className={classNames(
-          props.checked ? "bg-gray-900" : "bg-gray-200 hover:bg-gray-300",
+          props.checked ? "bg-gray-900" : "bg-gray-200",
+          primitiveProps.disabled ? "cursor-not-allowed" : "hover:bg-gray-300",
```
zomars, line 25: At least for desktop users, disabled switches will display a "non allowed" cursor. Tried adding opacity as well but it becomes confusion between on-disabled/off-disabled/on-enabled/off-enabled cc @Jaibles any suggestions for mobile users?

emrysal (approved):
I don't really like pseudo-fields, at that point you're better off with state again, which Controller under the hood also does. Minor nit.

zomars:
> I don't really like pseudo-fields, at that point you're better off with state again, which Controller under the hood also does. Minor nit.

We could make it an actual field and move this logic to the backend tho. "If input. seatsPerTimeSlotEnabled is false then set the seatsPerTimeSlot to null"

emrysal:
> We could make it an actual field and move this logic to the backend tho. "If input. seatsPerTimeSlotEnabled is false then set the seatsPerTimeSlot to null"

I'd say no as it's purely FE logic, if I ask myself "What if I update through the API" that doesn't make sense. But asking myself the same question the following logic does come to mind as belonging to the BE.

```javascript
// Enabling seats will disable guests and requiring confirmation until fully supported
if (e) {
  formMethods.setValue("disableGuests", true);
  formMethods.setValue("requiresConfirmation", false);
  formMethods.setValue("seatsPerTimeSlot", 2);
} else {
  formMethods.setValue("seatsPerTimeSlot", null);
}
```

PR #4732, opened 2022-09-28
Separate and paginate team query for members

joeauyeung:
## What does this PR do?

Since Vercel only allows 4MB max for API payloads, this PR separates the "viewer.teams.get" call into one that gets the team info and another get gets the team members. The get team members query is also paginated so we should avoid the 4MB limit that Vercel imposes. 

Fixes # (issue)

https://www.loom.com/share/12ee4dc070d94bb28adb363944994014

## Type of change

## How should this be tested?

- Sign into the teampro account
- In settings go to the Seeded team page
    - Click load more memebers. If there are only 6 members then the button should disappear
    - Invite a team member
    - Delete a member
    - Change the role of a member
- Go to the team's profile page
    - Make changes and save
- Go to the teams appearance page
    - Disable branding and save

## Checklist


apps/web/components/team/MemberList.tsx
```diff
@@ -4,7 +4,7 @@ import MemberListItem from "./MemberListItem";
 
 interface Props {
   team: inferQueryOutput<"viewer.teams.get">;
-  members: inferQueryOutput<"viewer.teams.get">["members"];
+  members: inferQueryOutput<"viewer.teams.getMembers">["members"];
```
joeauyeung, line 7: Since we are separating the "get" and "getMembers" endpoint we need to change the types here

apps/web/components/team/MemberListItem.tsx
```diff
@@ -92,11 +92,11 @@ export default function MemberListItem(props: Props) {
           </div>
           <div className="mt-2 flex ltr:mr-2 rtl:ml-2 sm:mt-0 sm:justify-center">
             {/* Tooltip doesn't show... WHY????? */}
-            {props.member.isMissingSeat && (
+            {/* {props.member.isMissingSeat && (
               <Tooltip side="top" content={t("hidden_team_member_message")}>
                 <TeamPill color="red" text={t("hidden")} />
               </Tooltip>
```
joeauyeung, line 98: We don't need this anymore right? Billing for teams changes based on the number of people automatically

apps/web/pages/settings/teams/[id]/index.tsx
```diff
@@ -95,7 +95,7 @@ export function TeamSettingsPage() {
               )
             }>
             {!!errorMessage && <Alert className="-mt-24 border" severity="error" title={errorMessage} />}
-            {team && (
+            {/* {team && (
               <>
                 <div className="block sm:flex md:max-w-5xl">
                   <div className="w-full ltr:mr-2 rtl:ml-2 sm:w-9/12">
```
Comment (joeauyeung) [inline apps/web/pages/settings/teams/[id]/index.tsx:101]: GH is being weird and not letting me comment the whole block. Because this is a V1 screen I just commented this out since it was giving type errors

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -9,7 +9,7 @@ import { Button, Dialog, DialogContent, DialogFooter, Select, TextField } from "
 
 type MemberInvitationModalProps = {
   isOpen: boolean;
-  team: TeamWithMembers | null;
+  teamId: number | null;
```
joeauyeung, line 12: This component only needed the teamId and not the whole team object

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -35,7 +35,7 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
 
   const inviteMemberMutation = trpc.useMutation("viewer.teams.inviteMember", {
     async onSuccess() {
-      await utils.invalidateQueries(["viewer.teams.get"]);
+      await utils.invalidateQueries(["viewer.teams.getMembers"]);
```
joeauyeung, line 38: Since this component changes the team members we have to invalidate the new "getMembers" query

packages/features/ee/teams/components/MemberListItem.tsx
```diff
@@ -100,7 +100,7 @@ export default function MemberListItem(props: Props) {
               <div className="mb-1 flex">
                 <span className="mr-1 text-sm font-bold leading-4">{name}</span>
 
-                {props.member.isMissingSeat && <TeamPill color="red" text={t("hidden")} />}
+                {/* {props.member.isMissingSeat && <TeamPill color="red" text={t("hidden")} />} */}
```
joeauyeung, line 103: We don't need this for V2 right?

packages/features/ee/teams/components/v2/AddNewTeamMembers.tsx
```diff
@@ -30,10 +30,10 @@ const AddNewTeamMembers = (props: { teamId: number }) => {
   const { t } = useLocale();
   const utils = trpc.useContext();
 
-  const { data: team, isLoading } = trpc.useQuery(["viewer.teams.get", { teamId: props.teamId }]);
+  const { data: team, isLoading } = trpc.useQuery(["viewer.teams.getMembers", { teamId: props.teamId }]);
```
joeauyeung, line 33: We don't need the team info just the members query. Team id is passed down as a prop

packages/features/ee/teams/pages/team-members-view.tsx
```diff
@@ -1,20 +1,20 @@
 import { MembershipRole } from "@prisma/client";
 import { useSession } from "next-auth/react";
 import { useRouter } from "next/router";
-import { useState } from "react";
+import { useState, Fragment } from "react";
```
joeauyeung, line 4: @zomars I noticed on other PRs you've used <Fragment> over <>. Is that the convention we should be aiming for?

packages/features/ee/teams/pages/team-members-view.tsx
```diff
@@ -27,13 +27,25 @@ const MembersView = () => {
...
+  const {
+    data: teamMemebers,
+    isLoading: memberLoading,
+    hasNextPage,
+    fetchNextPage,
+    isFetchingNextPage,
+  } = trpc.useInfiniteQuery(["viewer.teams.getMembers", { teamId: Number(router.query.id), limit: 5 }], {
+    getNextPageParam: (lastPage) => lastPage.nextCursor,
+  });
+
```
joeauyeung, line 39: Created a new tRPC endpoint for members. This bug came up with a team that had 20 members. By breaking by paginating this query we should avoid hitting Vercel's 4MB limit.

packages/features/ee/teams/pages/team-members-view.tsx
```diff
@@ -56,7 +68,7 @@ const MembersView = () => {
                     ]}
                   />
                 )}
-                {team.membership.role === MembershipRole.OWNER &&
+                {/* {team.membership.role === MembershipRole.OWNER &&
                 team.membership.isMissingSeat &&
                 team.requiresUpgrade ? (
                   <Alert
```
joeauyeung, line 74: Do we need this for V2 teams?

packages/lib/server/queries/teams/index.ts
```diff
@@ -4,44 +4,55 @@ import prisma, { baseEventTypeSelect } from "@calcom/prisma";
...
+  },
+});
+
+export const getTeam = async (id?: number, slug?: string) => {
+  const team = await prisma.team.findUnique({
+    where: id ? { id } : { slug },
+    include: {
+      members: true,
     },
   });
```
joeauyeung, line 50: Create a new function that only grabs the team without the member user info to reduce size

packages/trpc/server/routers/viewer/teams.tsx
```diff
@@ -35,20 +41,93 @@ export const viewerTeamsRouter = createProtectedRouter()
...
+            select: {
+              userId: true,
+            },
+          },
+        },
+      });
+
+      if (!team?.members.some((member) => member.userId !== ctx.user.id)) {
+        throw new TRPCError({ code: "UNAUTHORIZED", message: "You are not a member of this team." });
+      }
```
joeauyeung, line 84: Use this query to confirm that the user making the call is actually a part of the team.

apps/web/components/team/MemberListItem.tsx
zomars: It is not implemented yet

packages/features/ee/teams/components/MemberListItem.tsx
zomars: We do until team billing is properly implemented

packages/features/ee/teams/components/v2/AddNewTeamMembers.tsx
```diff
@@ -104,8 +104,9 @@ const AddNewTeamMembers = (props: { teamId: number }) => {
           <MemberInvitationModal
             isOpen={memberInviteModal}
             onExit={() => setMemberInviteModal(false)}
-            team={team}
-            currentMember={team?.membership.role}
+            teamId={props.teamId}
+            // We set this as owner since this is the screen that appears when a team is first made
+            currentMember="OWNER"
```
zomars, line 109: Can we get this from the session role?

packages/features/ee/teams/pages/availability.tsx
```diff
@@ -30,9 +30,9 @@ export function TeamAvailabilityPage() {
   });
 
   // prevent unnecessary re-renders due to shell queries
-  const TeamAvailability = useMemo(() => {
-    return <TeamAvailabilityScreen team={team} />;
-  }, [team]);
+  // const TeamAvailability = useMemo(() => {
+  //   return <TeamAvailabilityScreen team={team} />;
+  // }, [team]);
```
zomars, line 35: Why?

packages/features/ee/teams/pages/availability.tsx
```diff
@@ -54,11 +54,11 @@ export function TeamAvailabilityPage() {
       <LicenseRequired>
         {!!errorMessage && <Alert className="-mt-24 border" severity="error" title={errorMessage} />}
         {isLoading && <Loader />}
-        {isFreeUser ? (
+        {/* {isFreeUser ? (
```
zomars, line 57: This is technically free now right @PeerRich We don't need to gate this anymore. We should implement team billing properly tho.

packages/features/ee/teams/pages/team-members-view.tsx
zomars: There's no convention. But I use `Fragment` when need to pass the `key` prop for arrays.

packages/features/ee/teams/pages/team-members-view.tsx
zomars: There might be teams with missing seats in V2. So yes.

packages/features/ee/teams/components/TeamAvailabilityScreen.tsx
```diff
@@ -33,86 +33,87 @@ export default function TeamAvailabilityScreen(props: Props) {
...
-                <span className="-mt-1 block truncate text-sm text-gray-400">{member?.email}</span>
-              </div>
-            </div>
-          }
-        />
-      </div>
-    );
-  };
+    //   return (
+    //     <div key={member.id} style={style} className="flex border-r border-gray-200 pl-4 ">
```
zomars, line 37: Why?

packages/features/ee/teams/pages/availability.tsx
joeauyeung: This page isn't being used in V2 and it was causing type errors

packages/features/ee/teams/components/TeamAvailabilityScreen.tsx
joeauyeung: Also not being used in V2 and was causing type errors

apps/web/components/team/MemberListItem.tsx
alannnc: Can we add context as a note?
```suggestion
            /* @NOTE: not implemented yet */
            {/* {props.member.isMissingSeat && (
              <Tooltip side="top" content={t("hidden_team_member_message")}>
                <TeamPill color="red" text={t("hidden")} />
              </Tooltip>
```

packages/features/ee/teams/components/MemberListItem.tsx
alannnc: add context note

packages/features/ee/teams/pages/team-members-view.tsx
```diff
@@ -109,13 +121,29 @@ const MembersView = () => {
...
               </div>
             )}
-            <div>
-              <ul className="divide-y divide-gray-200 rounded-md border ">
-                {team?.members.map((member) => {
-                  return <MemberListItem key={member.id} team={team} member={member} />;
-                })}
-              </ul>
-            </div>
+            {teamMemebers && team && (
```
alannnc, line 124:
```suggestion
            {teamMembers && team && (
```

hariombalhara (changes requested):
There are some bugs that I found while testing the PR. I have opened a separate PR to fix those.
https://github.com/calcom/cal.com/pull/4994

Bug that I have not fixed:
- On refresh of members page, [sometimes an error is thrown by tRPC endpoint]( https://www.loom.com/share/4c0f903f1c8d4eaaa0454e718e635904) because teamId is NaN.

Screenshot 2022-10-14 at 11 08 35 AM

packages/features/ee/teams/pages/team-members-view.tsx
```diff
@@ -25,6 +25,18 @@ const MembersView = () => {
     onError: () => {
       router.push("/settings");
     },
+    enabled: !!router.query.id,
```
joeauyeung, line 28: Adding this will only run the query when `useRouter` has time to parse the query params

@hariombalhara

zomars: Why not use router.isReady ?

joeauyeung: Good call

packages/features/ee/teams/pages/team-members-view.tsx
```diff
@@ -25,6 +25,18 @@ const MembersView = () => {
...
+
+  const {
+    data: teamMembers,
+    isLoading: memberLoading,
+    hasNextPage,
+    fetchNextPage,
+    isFetchingNextPage,
+  } = trpc.useInfiniteQuery(["viewer.teams.getMembers", { teamId: Number(router.query.id), limit: 5 }], {
+    getNextPageParam: (lastPage) => lastPage.nextCursor,
+    enabled: !!router.isReady,
```
hariombalhara, line 39: The problem with this approach is that router.query.id isn't a number(we are calming typescript by doing an explicit typecast to a number). So, we aren't getting compile time safety. If we use zod over query it would fail.

One of the approaches that I use is to, make a wrapper component that renders this component only when router.isReady is true. After that in this component we should parse the schema using zod, because we now know that it has to be a parseable number.

hariombalhara: cc @zomars for review

hariombalhara (approved):
Looks good from my side now.

hariombalhara:
There is one more thing that should be fixed, Team's menu should auto expand on refresh to show currently selected tab.
https://www.loom.com/share/2db958f98ff84177920611276d970691

packages/features/ee/teams/pages/team-members-view.tsx
joeauyeung: @hariombalhara can you point me to a file where you used the wrapper component?

hariombalhara: Screenshot 2022-10-17 at 9 09 05 PM
Here is a similar approach, this approach solves the problem of making sure useForm defaultValues has non null value.

Similar approach can be used for this usecase.

packages/features/ee/teams/pages/team-appearance-view.tsx
```diff
@@ -29,16 +29,20 @@ const ProfileView = () => {
...
+    {
+      onError: () => {
+        router.push("/settings");
+      },
+      onSuccess: (team) => {
+        if (team) {
+          form.setValue("hideBranding", team.hideBranding);
+        }
+      },
+      enabled: !!router.isReady,
```
joeauyeung, line 43: Good catch @hariombalhara about refreshing the page. Had to add the enabled to both the profile and appearance page.

zomars:
Closing since the payload limit got resolved in 6e9849e173c5dd9420c7da6932249ad73f19b33e

hariombalhara:
@joeauyeung there are other fixes as well in it. Let's open a PR for the remaining one ?

PR #4739, opened 2022-09-28
Close menu and prevent overscroll

joeauyeung:
## What does this PR do?

This PR closes the settings side container when the route changes and prevents scrolling past the overlay.

Fixes # (issue)

## Type of change

## How should this be tested?

- Navigate to settings on a small screen size
- Open the container and choose an option
- With the container open try to scroll past the overlay

## Checklist


packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -311,6 +313,12 @@ export default function SettingsLayout({
     };
   }, []);
 
+  useEffect(() => {
+    if (sideContainerOpen) {
+      setSideContainerOpen(!sideContainerOpen);
+    }
+  }, [router.asPath]);
```
PeerRich, line 320: missing dependency?

joeauyeung: I understand the linting suggestion but I only want this to run when the router path changes. When both dependencies are added, the menu always closes because it triggers when `sideContainerOpen` changes but if it's open it'll just close automatically.

https://www.loom.com/share/7d453d125c6e4d758e9dcf0447db4345

zomars (approved):
LGTM

PR #4743, opened 2022-09-28
Fix connected calendar list

leog:
## What does this PR do?

Fixed an issue with the calendar list where an error showed up regarding primary calendar, which was referring to non-calendar apps not being filtered correctly.

Before:
<kbd></kbd>

After: 
<kbd></kbd>

## Type of change

## How should this be tested?

Have Hubspot installed along with a Google Calendar (with or without proper permissions.


emrysal (approved):
This fixes the immediate issue but we need context-based errors in the future.

🚀

alannnc:
> This fixes the immediate issue but we need context-based errors in the future.
> 
> 🚀

I'm working on a Fix for this. Catch the error and return connectedCalendar with an error flag. We can give this to production while I work on my solution as it will require a lot of changes.

zomars (approved):
Ship it

playwright.config.ts
```diff
@@ -59,6 +59,10 @@ const config: PlaywrightTestConfig = {
       timeout: 120000,
       use: {
         ...devices["Desktop Chrome"],
+        viewport: {
+          width: 1920,
+          height: 1280,
+        },
```
alannnc, line 65: Why this change was needed?

alannnc (approved):
Code looks good, couldn't tested it properly. Just a note about playwright config being changed.

playwright.config.ts
emrysal: Suspected that viewport size was the reason the end to end tests were failing which turned out to be wrong, it's back to normal now.

PR #4750, opened 2022-09-29
Ensure currentUsername is set before comparing it with username

### LINKED ISSUE #4749: “Username already taken” message for accounts that own that username

hariombalhara:
## What does this PR do?

Fixes #4749

**Environment**: Production

## Type of change

## How should this be tested?

- Impersonate a user and refresh multiple times the profile page to see the issue.
## Checklist


zomars (approved):

alannnc (approved):
I did once run on this error and looks like the fix worked.
NIT: @hariombalhara 
On first render I'm able to see this text on yellow appears and then it's gone.
"You need to reserve your premium username for ..."
I don't know when it's triggering but can we add a flag to don't show? Till it's needed.

PR #4763, opened 2022-09-29
Improved team layout when blocks have unequal content length.

JeroenReumkens:
## What does this PR do?

Was just looking at the [cal.com team page](https://cal.com/team/cal?members=1) and noticed that the layout was a bit skewed when some people had more content than others. While working on this I also noticed that the component kept on rerendering when clicking around on this page. This happened because all the components were defined inside the render, fixed that as well in this PR. 

**Before:** 
CleanShot 2022-09-29 at 16 17 04@2x
CleanShot 2022-09-29 at 16 17 21@2x

**After this PR:**
CleanShot 2022-09-29 at 16 18 13@2x
CleanShot 2022-09-29 at 16 18 31@2x

## Type of change

### DIFF
--- a/apps/web/components/team/screens/Team.tsx
+++ b/apps/web/components/team/screens/Team.tsx
@@ -13,46 +13,48 @@ type TeamType = TeamPageProps["team"];
 type MembersType = TeamType["members"];
 type MemberType = MembersType[number];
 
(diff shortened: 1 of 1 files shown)


emrysal (approved):
Tried it and looks good to me 👍

PR #4771, opened 2022-09-30
#4533: Show correct time format on booking page

### LINKED ISSUE #4533: US-based timezone showing 24-hour time by default
This might be similar to #3908 -- but see my URL at https://cal.com/tim/15?date=2022-09-17

As an American on a US IP, I'd expect this to be 12-hour time. It's been like this for awhile.

Screenshot_2022-09-16_09-09-16_2okMmw1P

JeroenReumkens:
## What does this PR do?

Store timeformat preferences in localstorage when changing profile. This way you see the correct format on the booking page. 

I've also abstracted the reuse of the localstorage key for time format into a small util function, to prevent typos in the future.

Fixes #4533

https://www.loom.com/share/5e2c10f918a749fcb11ef157268d7ded

## Type of change

## How should this be tested?

- [ ] Verify that when you update the time format in your settings the format on the bookings page updates accordingly.

### DIFF
--- a/apps/web/components/booking/pages/AvailabilityPage.tsx
+++ b/apps/web/components/booking/pages/AvailabilityPage.tsx
@@ -224,7 +224,7 @@ function TimezoneDropdown({
   const [isTimeOptionsOpen, setIsTimeOptionsOpen] = useState(false);
 
   useEffect(() => {
-    handleToggle24hClock(getIs24hClockFromLocalStorage() === true);
+    handleToggle24hClock(!!getIs24hClockFromLocalStorage());
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
--- a/apps/web/components/booking/pages/AvailabilityPage.tsx
+++ b/apps/web/components/booking/pages/AvailabilityPage.tsx
@@ -24,7 +24,7 @@ import useTheme from "@calcom/lib/hooks/useTheme";
 import notEmpty from "@calcom/lib/notEmpty";
 import { getRecurringFreq } from "@calcom/lib/recurringStrings";
 import { collectPageParameters, telemetryEventTypes, useTelemetry } from "@calcom/lib/telemetry";
-import { detectBrowserTimeFormat } from "@calcom/lib/timeFormat";
+import { detectBrowserTimeFormat, getIs24hClockFromLocalStorage } from "@calcom/lib/timeFormat";
 import { localStorage } from "@calcom/lib/webstorage";
 import { trpc } from "@calcom/trpc/react";
 import { Icon } from "@calcom/ui/Icon";
@@ -225,7 +225,7 @@ function TimezoneDropdown({
   const [isTimeOptionsOpen, setIsTimeOptionsOpen] = useState(false);
 
   useEffect(() => {
-    handleToggle24hClock(localStorage.getItem("timeOption.is24hClock") === "true");
+    handleToggle24hClock(getIs24hClockFromLocalStorage() === "true");
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
--- a/apps/web/components/booking/pages/AvailabilityPage.tsx
+++ b/apps/web/components/booking/pages/AvailabilityPage.tsx
@@ -224,7 +224,7 @@ function TimezoneDropdown({
   const [isTimeOptionsOpen, setIsTimeOptionsOpen] = useState(false);
 
   useEffect(() => {
-    handleToggle24hClock(getIs24hClockFromLocalStorage() === "true");
+    handleToggle24hClock(getIs24hClockFromLocalStorage() === true);
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
--- a/apps/web/components/booking/pages/AvailabilityPage.tsx
+++ b/apps/web/components/booking/pages/AvailabilityPage.tsx
@@ -25,7 +25,6 @@ import notEmpty from "@calcom/lib/notEmpty";
 import { getRecurringFreq } from "@calcom/lib/recurringStrings";
 import { collectPageParameters, telemetryEventTypes, useTelemetry } from "@calcom/lib/telemetry";
 import { detectBrowserTimeFormat, getIs24hClockFromLocalStorage } from "@calcom/lib/timeFormat";
-import { localStorage } from "@calcom/lib/webstorage";
 import { trpc } from "@calcom/trpc/react";
 import { Icon } from "@calcom/ui/Icon";
 import DatePicker from "@calcom/ui/v2/modules/booker/DatePicker";
--- a/apps/web/lib/clock.ts
+++ b/apps/web/lib/clock.ts
@@ -1,6 +1,10 @@
 // handles logic related to user clock display using 24h display / timeZone options.
 import dayjs from "@calcom/dayjs";
-import { isBrowserLocale24h } from "@calcom/lib/timeFormat";
+import {
+  getIs24hClockFromLocalStorage,
+  isBrowserLocale24h,
+  setIs24hClockInLocalStorage,
+} from "@calcom/lib/timeFormat";
 import { localStorage } from "@calcom/lib/webstorage";
 
 interface TimeOptions {
@@ -20,8 +24,8 @@ const initClock = () => {
     return;
   }
   // This only sets browser locale if there's no preference on localStorage.
-  if (!localStorage.getItem("timeOption.is24hClock")) set24hClock(isBrowserLocale24h());
-  timeOptions.is24hClock = localStorage.getItem("timeOption.is24hClock") === "true";
+  if (!getIs24hClockFromLocalStorage()) set24hClock(isBrowserLocale24h());
+  timeOptions.is24hClock = getIs24hClockFromLocalStorage() === "true";
   timeOptions.inviteeTimeZone = localStorage.getItem("timeOption.preferredTimeZone") || dayjs.tz.guess();
 };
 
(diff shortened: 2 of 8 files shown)


apps/web/pages/v2/settings/my-account/general.tsx
```diff
@@ -110,6 +111,8 @@ const GeneralView = ({ localeProp }: GeneralViewProps) => {
     <Form
       form={formMethods}
       handleSubmit={(values) => {
+        setIs24hClockInLocalStorage(values.timeFormat.value === 12 ? "false" : "true");
```
JeroenReumkens, line 114: This is the main fix for this issue. Making sure that we update the localstorage when saving your preferences in the profile.

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -24,7 +24,7 @@ import useTheme from "@calcom/lib/hooks/useTheme";
 import notEmpty from "@calcom/lib/notEmpty";
 import { getRecurringFreq } from "@calcom/lib/recurringStrings";
 import { collectPageParameters, telemetryEventTypes, useTelemetry } from "@calcom/lib/telemetry";
-import { detectBrowserTimeFormat } from "@calcom/lib/timeFormat";
+import { detectBrowserTimeFormat, getIs24hClockFromLocalStorage } from "@calcom/lib/timeFormat";
 import { localStorage } from "@calcom/lib/webstorage";
```
zomars, line 28:
```suggestion
```

zomars: Removing unused import

zomars (approved):
LGTM

hariombalhara:
@JeroenReumkens Can you also check the behaviour in Chrome incognito in embed mode on a third party website. A sample codesandbox is here https://d1zlcn.csb.app/

zomars:
> This is important because in Chrome incognito, in third party context, localStorage access throws error and it doesn't store anything.

I forgot about that! Cookies do work on incognito right @hariombalhara ?

hariombalhara:
Cookies also don't work. You can't have any kind of persistence in a third-party context. Though, cookies fail silently without throwing errors.

packages/lib/timeFormat.ts
```diff
@@ -6,8 +6,16 @@
  */
 import { localStorage } from "@calcom/lib/webstorage";
 
+const is24hLocalstorageKey = "timeOption.is24hClock";
+
+export const setIs24hClockInLocalStorage = (is24h: "true" | "false") =>
+  localStorage.setItem(is24hLocalstorageKey, is24h);
```
alannnc, line 12: Is there a reason that we can't receive true/false as a boolean and just save it as is24h.toString()?

packages/lib/timeFormat.ts
```diff
@@ -6,8 +6,16 @@
...
+
+export const getIs24hClockFromLocalStorage = () =>
+  localStorage.getItem(is24hLocalstorageKey) as "true" | "false" | null;
```
alannnc, line 15: Also here we can create the logic to return boolean.

zomars:
> Cookies also don't work. You can't have any kind of persistence in a third-party context. Though, cookies fail silently without throwing errors.

If that were true we couldn't login at all in incognito mode. Cookies do work. But they're not persisted after closing the window.

alannnc (commented):
Probably we can improve get/set localStorage function to use booleans as parameters and return values, so this can be easily used by others and eliminate redundant ternary operations or comparisons.

JeroenReumkens:
> Probably we can improve get/set localStorage function to use booleans as parameters and return values, so this can be easily used by others and eliminate redundant ternary operations or comparisons.

I gave it another thought, and I think if I do all the converting in the util there's little confusion around this indeed. I'll change this now.

packages/lib/timeFormat.ts
alannnc: I think we should be safe as anything with cookies, localStorage, and query params is expected to be string value saved. We can aim for ease of use in this case. You code once and use it everywhere, instead of using === "true" when being used in other places.

pumfleet (approved):
Looks great to me, I am impartial about @alannnc's comment, so I'll leave it down to you guys if it's worth changing or not, but otherwise all good!

JeroenReumkens:
Changed strings to booleans 😄 
Would love a recheck from you @alannnc. Commit: https://github.com/calcom/cal.com/pull/4771/commits/ef3c78c6fcbf613e68191c5d9eb2270bd771033c

hariombalhara:
> If that were true we couldn't login at all in incognito mode. Cookies do work. But they're not persisted after closing the window.

@zomars  So, there is an additional factor here about Third Party Context. Cal Embeds would usually open in third party context, where the website embedding cal.com won't be cal.com and thus cal.com embed would be in third party context.

See this loom where I try to explain the difference https://www.loom.com/share/b381e498e6be48f6916714f9e5444cf5

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -225,7 +224,7 @@ function TimezoneDropdown({
   const [isTimeOptionsOpen, setIsTimeOptionsOpen] = useState(false);
 
   useEffect(() => {
-    handleToggle24hClock(localStorage.getItem("timeOption.is24hClock") === "true");
+    handleToggle24hClock(getIs24hClockFromLocalStorage() === true);
```
alannnc, line 227:
```suggestion
    handleToggle24hClock(getIs24hClockFromLocalStorage());
```

alannnc: I believe === true becomes redundant as we should now return true|false|null.

apps/web/lib/clock.ts
```diff
@@ -20,8 +24,8 @@ const initClock = () => {
     return;
   }
   // This only sets browser locale if there's no preference on localStorage.
-  if (!localStorage.getItem("timeOption.is24hClock")) set24hClock(isBrowserLocale24h());
-  timeOptions.is24hClock = localStorage.getItem("timeOption.is24hClock") === "true";
+  if (getIs24hClockFromLocalStorage() === null) set24hClock(isBrowserLocale24h());
```
alannnc, line 27: === null should be valid as it's a special case I think.

apps/web/lib/clock.ts
```diff
@@ -20,8 +24,8 @@ const initClock = () => {
...
+  timeOptions.is24hClock = getIs24hClockFromLocalStorage() === true;
```
alannnc, line 28:
```suggestion
  timeOptions.is24hClock = !!getIs24hClockFromLocalStorage();
```

apps/web/pages/v2/success.tsx
```diff
@@ -201,7 +201,7 @@ export default function Success(props: SuccessProps) {
       // TODO: Add payment details
     });
     setDate(date.tz(localStorage.getItem("timeOption.preferredTimeZone") || dayjs.tz.guess()));
-    setIs24h(!!localStorage.getItem("timeOption.is24hClock"));
+    setIs24h(!!getIs24hClockFromLocalStorage());
```
alannnc, line 204: Thumbs up

packages/lib/timeFormat.ts
```diff
@@ -6,23 +6,35 @@
...
+const is24hLocalstorageKey = "timeOption.is24hClock";
+
+export const setIs24hClockInLocalStorage = (is24h: boolean) =>
+  localStorage.setItem(is24hLocalstorageKey, is24h.toString());
+
+export const getIs24hClockFromLocalStorage = () => {
+  const is24hFromLocalstorage = localStorage.getItem(is24hLocalstorageKey);
+  if (is24hFromLocalstorage === null) return null;
+
+  return is24hFromLocalstorage === "false" ? false : true;
```
alannnc, line 18:
```suggestion
  return is24hFromLocalstorage === "true";
```

alannnc: With only comparing to true should return true|false as null it's being handled before.

packages/lib/timeFormat.ts
```diff
@@ -6,23 +6,35 @@
...
+};
+
 export const isBrowserLocale24h = () => {
-  const localStorageTimeFormat = localStorage.getItem("timeOption.is24hClock");
+  const localStorageTimeFormat = getIs24hClockFromLocalStorage();
```
alannnc, line 22:
```suggestion
  const localStorageTimeFormat = !!getIs24hClockFromLocalStorage();
```

packages/lib/timeFormat.ts
```diff
@@ -6,23 +6,35 @@
...
   // If time format is already stored in the browser then retrieve and return early
-  if (localStorageTimeFormat === "true") {
+  if (localStorageTimeFormat === true) {
     return true;
-  } else if (localStorageTimeFormat === "false") {
+  } else if (localStorageTimeFormat === false) {
     return false;
   }
```
alannnc, line 28: return localStorageTimeFormat; Its already converted with the !! from above

alannnc:
Re-reviewed, handling the comments and we should be good!

packages/lib/timeFormat.ts
JeroenReumkens: Yes smart!!

packages/lib/timeFormat.ts
JeroenReumkens: In this case we don't want to convert null into the boolean as well. If it's null we explicitly want to skip to the logic of trying to infer this from the browsers language.

alannnc (approved):

PR #4777, opened 2022-09-30
Update Close.com info

joeauyeung:
## What does this PR do?

This PR updates Close.com's app information.

https://www.loom.com/share/5134a40d3f7b43e8bf7ab6d43b3f2817

## Type of change


shirazdole:
Is is possible to say that the integration is free, but the actual service cost for that product costs x amount. 

Maybe 

"Cal.com Integration" 
"Free"

joeauyeung:
> Is is possible to say that the integration is free, but the actual service cost for that product costs x amount.
> 
> Maybe
> 
> "Cal.com Integration" "Free"

Maybe instead of just "Pricing" we should change the title to "Service Pricing" or "Provider Pricing". I can't think of the exact wording but something that conveys that Cal.com isn't the one charging for the service but the app provider is.

leog:
I'm wondering why do we need to include the service pricing information. If anyone is interested in using Close or any other paid app, they should be informed in the app official page. We just publish we have some kind of integration with that app and how we integrate, and that should be enough. Otherwise we would need to keep track of the app pricing in their official website and continuously update as they update their pricing, which seems very inconvenient.

PeerRich:
> I'm wondering why do we need to include the service pricing information. If anyone is interested in using Close or any other paid app, they should be informed in the app official page. We just publish we have some kind of integration with that app and how we integrate, and that should be enough. Otherwise we would need to keep track of the app pricing in their official website and continuously update as they update their pricing, which seems very inconvenient.



agree -- we should use the price section only for our price not theirs

in this case, our close app is free

closing-- sorry @joeauyeung

PeerRich:
lets just add this in the description

PeerRich (approved):

PR #4778, opened 2022-09-30
Fix destination calendar overflow on installed and move DestinationCalendarSelector to feature package

joeauyeung:
## What does this PR do?

This PR fixes the destination calendar overflowing when the ID was too long. I have also moved the DestinationCalendarSelector to the feature/calendar package.

Before
image

After
image

Fixes # (issue)

## Type of change

## How should this be tested?

- Change the destination calendar in the following places
    - Installed
    - A single event type
    - Settings -> Calendars

## Checklist


leog:
> image

@joeauyeung That looks a bit off, as the combobox now doesn't include details of the calendar (to avoid a long ID overflowing the UI), it now looks too big for its content. Maybe a better solution would be to truncate that ID, as the idea is to provide a way to identify which calendar provider was selected aside from the name. 

Also, the long ID comes from a CalDav calendar right? When a Google Calendar is selected, the Google email associated to the Google Calendar should show up as the ID of that Google Calendar. The goal here is, as you may have multiple Google Calendars and in the case you have identical calendars in them, you couldn't know which was selected as destination.

joeauyeung:
@leog what if we render the integration headers that are present when choosing a destination calendar? Here's a Loom with my thoughts

https://www.loom.com/share/d7664d33a53b40c5b7e26c3754772d46

leog:
> @leog what if we render the integration headers that are present when choosing a destination calendar? Here's a Loom with my thoughts
> 
> 
> https://www.loom.com/share/d7664d33a53b40c5b7e26c3754772d46

Very well thought, I like it 👍 I wonder if we can improve that lag when loading the name to make it perfect. Anyway, I think it works, we could ask @Jaibles just to confirm.

Jaibles:
@joeauyeung @leog this looks good to me in the Loom!

Would it be awkward/weird to remove the "calendar" in these particular instances just to shorten this?
CleanShot 2022-10-11 at 17 56 36@2x

I know technically it's correct to leave it in, to match the app name but it does really make the labelling longer and it's obvious here it's a calendar.

packages/app-store/office365calendar/_metadata.ts
```diff
@@ -3,10 +3,10 @@ import type { AppMeta } from "@calcom/types/App";
 import _package from "./package.json";
 
 export const metadata = {
-  name: "Office 365 / Outlook.com Calendar",
+  name: "Outlook Calendar",
   description: _package.description,
   type: "office365_calendar",
-  title: "Office 365 / Outlook.com Calendar",
+  title: "Outlook Calendar",
```
joeauyeung, line 9: I propose shortening the app name to just "Outlook Calendar"

packages/features/calendars/DestinationCalendarSelector.tsx
```diff
@@ -44,31 +50,63 @@ const DestinationCalendarSelector = ({
...
+  };
+
+  const Option = (props) => {
+    const { label } = props.data;
+    return (
+      <components.Option {...props}>
+        <span>{label}</span>
+      </components.Option>
+    );
+  };
```
joeauyeung, line 69: Use custom components for select to render which integration the calendar is from

packages/features/calendars/DestinationCalendarSelector.tsx
```diff
@@ -44,31 +50,63 @@ const DestinationCalendarSelector = ({
...
         value: `${selected.integration}:${selected.externalId}`,
-        label: selected.name || "",
+        label: `${selected.name} ` || "",
+        subtitle: `(${selectedIntegration?.integration.title?.replace(/calendar/i, "")} - ${
+          selectedIntegration?.primary?.name
+        })`,
       });
     }
-  }, [query.data?.connectedCalendars, value]);
+  }, [query.data?.connectedCalendars]);
```
joeauyeung, line 92: Let's display the actual destination calendar rather than "Selected destination calendar"

packages/features/calendars/DestinationCalendarSelector.tsx
```diff
@@ -44,31 +50,63 @@ const DestinationCalendarSelector = ({
...
+        selectedCalendar.primary?.name
+      })`,
       options: (selectedCalendar.calendars ?? [])
         .filter((cal) => cal.readOnly === false)
         .map((cal) => ({
-          label: cal.name || "",
+          label: ` ${cal.name} `,
+          subtitle: `(${selectedCalendar?.integration.title?.replace(/calendar/i, "")} - ${
+            selectedCalendar?.primary?.name
+          })`,
```
joeauyeung, line 109: Move where which integration the calendar comes from into the subtitle

joeauyeung (commented):
@leog @Jaibles I made where the calendar comes as a subtitle so we can style it differently.

image

leog (approved):
Looks good to me! Nice work Joe

apps/web/pages/settings/my-account/calendars.tsx
```diff
@@ -3,6 +3,8 @@ import Link from "next/link";
 import { useRouter } from "next/router";
 import { Fragment } from "react";
 
+
```
zomars, line 6:
```suggestion
```

packages/app-store/office365calendar/_metadata.ts
zomars: This should be another PR but I'll allow it

zomars (approved):
LGTM

zomars:
We do have some type-check errors tho.

PR #4800, opened 2022-10-03
Hotfix: Fix connected calendar list (#4743)

leog:
Cherry picking #4743 as a hotfix.


joeauyeung (approved):
If it's good enough for @zomars it's good enough for me

PR #4807, opened 2022-10-03
Hotfix: Fix Broken webhooks

hariombalhara:
## What does this PR do?

Fixes # (issue)
Bug: when you try to create a new webhook when Giphy or any other App is installed.
Screenshot 2022-10-04 at 12 08 40 AM
Reported by @shirazdole 

We have tests for webhook and they didn't catch the bug because the setup doesn't have an app installed.

**Environment**: Production

## Type of change

## How should this be tested?

- [x] Make sure that you have Giphy app installed and go and try to add a webhook


alishaz-polymath (approved):
🚢

PR #4819, opened 2022-10-04
Embed Improvements: Add auto-scroll support back, reduce CLS, remove loader for Routing Form embed

### LINKED ISSUE #4805: Routing Forms issues observed on cal.com/sales
- Loading forms/xyzzzz visible on the page. It is shown because Cal embed.js isn't loaded yet.
	- We shouldn't show such text. We can immediately move to rendering null in that case and work on a skeleton for near future.
	- We can preload the link https://developer.cal.com/embed/set-up-your-embed#preload
- Unnecessary CLS on the page. 2 things seem to be happening
	1. Form description isn't visible in b/w and then it comes
	2. Form has top spacing in b/w which goes away
- Ensure the content is always visible even if the Routing Form has a lot of content causing user to scroll.

hariombalhara:
## What does this PR do?

Fixes #4805
cal.com/sales [After]: https://www.loom.com/share/c7d2cbccb7004abcb08b9f60b4a2a088
cal.com/sales[Before]: https://www.loom.com/share/1fad874d648a4129a241871188139ab8

## What changed
- Add auto-scroll support back to focus where the iframe is. It now tries to avoid refocussing when iframe is already mostly in focus.
- Reduces CLS by detecting embed at backend for Routing Forms. It also allows disabling loader and giving a faster experience. For other pages, there is an additional delay of 500ms to allow embed UI changes to be committed to DOM.
- Added `uiDebug` support that right now adds a border to mark where the iframe is exactly.

**Environment**: Production

## Type of change


emrysal (approved):
I wonder if there is a better way to wait for react component update, but setTimeout:500 is good 👍 - lgtm.

hariombalhara:
> I wonder if there is a better way to wait for react component update, but setTimeout:500 is good 👍 - lgtm.

Working on a followup PR(almost ready) in which all the embeddable links are pre-rendered and thus setTimeout can be removed.

PR #4821, opened 2022-10-04
fix/connected calendar list different states

alannnc:
## What does this PR do?

Fixes /settings/my-account/calendars view, when a single credential calendar failed it stopped entire execution of consecutive code. Now we try catch and return error on connectedCalendar: { error } so the UI can handle single error when loading a single calendar connection.
Also added onError when something goes wrong when fetching calendars and a button link to try to refetch query.

Fixes #

https://www.loom.com/share/b5bc8b7f89ab453d8686a21d48851903

## Type of change

## How should this be tested?

- [ ] Integrate any calendar, but remove any given permission e.g. in google account remove third party permission. Go to settings/my-account/calendars and now you should see single calendar error handling on the UI

## Checklist


joeauyeung (approved):

emrysal (changes requested):
Needs a few updates:

* translation for the error message "Calendar connection broke"

Ideally we'd show the other calendar connections as working and be specific about which connection broke.

alannnc:
> Ideally we'd show the other calendar connections as working and be specific about which connection broke.
I tried this, but the way we save credentials leave us without a specific identifier when something broke.
Some things that I came up with:

- In the calendar permission add email as another permission so we can have that to identify.
- When a new calendar it's added, save that credential with a custom identifier from user input, e.g. "Calendar from my work email"

apps/web/pages/v2/settings/my-account/calendars.tsx
```diff
@@ -95,7 +94,7 @@ const CalendarsView = () => {
                       <Alert
                         severity="warning"
                         key={item.credentialId}
-                        title="Calendar connection broke"
+                        title={t("calendar_connection_broke")}
```
emrysal, line 97: "calendar_connection_fail" with translation "Calendar connection failed" reads a little better I think

emrysal (approved):

Issue #4508, opened 2022-09-15
use new v2 Dropdown in booking page


PeerRich:
<img width="683" alt="CleanShot 2022-09-15 at 22 14 33@2x" src="https://user-images.githubusercontent.com/8019099/190518107-f2297abd-88a3-47e5-8573-65a187ad267f.png">

JeroenReumkens:
@sean-brydon  Just came across this issue now —  I will already do the majority of this work when converting this component to support dark mode (and thus implement select v2). Just so you know.

PR #4824, opened 2022-10-04
#4306: Added darkmode support for select by styling via tailwind

### LINKED ISSUE #4306: dark mode for <Select>
CleanShot 2022-09-08 at 18 32 07@2x

https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS-v2.0?node-id=385%3A32867

Image

JeroenReumkens:
## What does this PR do?

This PR implements **dark mode for all selects** (had to rebuild to tailwind to properly use dark mode styles), and converted the **timezoneselect component to use the new select v2**, so this also looks nice and supports dark mode (part of #4508).
Also added a sexy custom scroll bar (css only) to this dropdown, just looks so much better 😉 

Fixes #4306 #4508

CleanShot 2022-10-04 at 15 53 33@2x

## Loom video

In the below loom video I'll show you the different variants of the select that are still around, including the new dark mode in the beginning of the video. At the end of the video I also show a local testpage with a multiselect variant that has also been styled to match dark mode.

https://www.loom.com/share/bad49b3209264ccdb2dce864ebdd4b7a

## Type of change

## How should this be tested?

This changed touched all v2 selects. It is very important to click through the platform and double check we didn't break any of them.

- [ ] Ensure we didn't break any selects by consolidating them to one.


apps/web/styles/globals.css
```diff
@@ -197,11 +197,6 @@ button[role="switch"][data-state="checked"] {
     box-shadow: 0 0 0 4px rgba(0, 0, 0, 0.2);
   }
 
-  /* React Select V2 */
```
JeroenReumkens, line 200: Global styles have been moved to the component itself.

packages/ui/v2/core/form/select/Select.tsx
```diff
@@ -1,106 +1,83 @@
...
-    <components.Input
-      // disables our default form focus hightlight on the react-select input element
-      inputClassName={classNames("focus:ring-0 focus:ring-offset-0", inputClassName)}
-      {...props}
-    />
-  );
-};
-
-const OptionComponent = <Option, IsMulti extends boolean, Group extends GroupBase<Option>>({
+export const getReactSelectProps = <
```
JeroenReumkens, line 32: I'm not happy with all the generics defined in this file, but otherwise `react-select` wouldn't stop complaining. Tried to abstract it into a function, but that would make it a lot more complicated for a few lines less code.. If anyone has a great idea — always welcome!

packages/ui/v2/core/form/select/components.tsx
```diff
@@ -0,0 +1,141 @@
...
+  MenuProps,
+  OptionProps,
+  SingleValueProps,
+  ValueContainerProps,
+} from "react-select";
+
+import { classNames } from "@calcom/lib";
+import { Icon } from "@calcom/ui/Icon";
+
+export const InputComponent = <
```
JeroenReumkens, line 16: I'm not happy with all the generics defined in this file, but otherwise react-select wouldn't stop complaining. Tried to abstract it into a function, but that would make it a lot more complicated for a few lines less code.. If anyone has a great idea — always welcome!

PeerRich:
did you make sure it's not applying dark mode in our logged-in app? since we don't have dark mode yet in the logged in app we need to specifically only assign dark mode to components on the public booking page

sean-brydon (dismissed):
Overall LGTM excited for this to be a tailwind controlled component instead of some awful CSS :) 

few NITS
* API submodule doesn't need to be included here - neither does website 
*

JeroenReumkens:
@PeerRich  Sean is correct. Darkmode only applies on booking pages, not on the app itself yet. But IF we ever decided to enabled darkmode in the app as well, the select will auto adjust to the theme select :)

packages/ui/v2/core/form/select/Select.tsx
```diff
@@ -1,106 +1,83 @@
...
-    </components.Option>
-  );
-};
+  components,
+}: {
+  className?: string;
+  components: SelectComponentsConfig<Option, IsMulti, Group>;
+}) => ({
+  className: classNames(
+    "cal-react-select-container block h-[36px] w-full min-w-0 flex-1 rounded-md",
```
hariombalhara, line 44: I don't see class 'cal-react-select' or 'cal-react-select-container' being used now.

hariombalhara:
@JeroenReumkens 
MultiSelect in routing forms is still looking weird.
Screenshot 2022-10-10 at 3 55 58 PM

packages/ui/v2/core/TimezoneSelect.tsx
```diff
@@ -1,48 +1,25 @@
...
+  }, [className, components]);
 
-function TimezoneSelect({ className, ...props }: SelectProps) {
   return (
     <BaseSelect
-      theme={(theme) => ({
-        ...theme,
-        borderRadius: 6,
-        colors: {
-          ...theme.colors,
```
hariombalhara, line 19: It is so much cleaner now 🙌

packages/app-store/ee/routing-forms/pages/routing-link/[...appPages].tsx
```diff
@@ -170,7 +170,11 @@ function RoutingForm({ form, profile }: inferSSRProps<typeof getServerSideProps>
                       );
                     })}
                     <div className="mt-4 flex justify-end space-x-2 rtl:space-x-reverse">
-                      <Button loading={responseMutation.isLoading} type="submit" color="primary">
+                      <Button
```
Comment (hariombalhara) [inline packages/app-store/ee/routing-forms/pages/routing-link/[...appPages].tsx:173]: Changed by me to fix the Button in dark mode(found during testing this PR)

packages/ui/v2/modules/event-types/DestinationCalendarSelector.tsx
```diff
@@ -4,7 +4,7 @@ import React, { useEffect, useState } from "react";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { DestinationCalendar } from "@calcom/prisma/client";
 import { trpc } from "@calcom/trpc/react";
-import Select from "@calcom/ui/v2/core/form/Select";
```
hariombalhara, line 7: Why have we moved it to smallcase? Actual files still seems to be Capital case. This might break on non unix systems

hariombalhara (changes requested):

packages/ui/v2/modules/event-types/DestinationCalendarSelector.tsx
JeroenReumkens: It's a directory now with an `index.ts`, a `Select.tsx` (PascalCase) and `components.tsx`. The directory itself is lowercase since it isn't a component file? I think that makes most sense, right?

hariombalhara: Oooooops. Didn't notice that. Yeah totally makes sense.

packages/ui/v2/core/form/select/Select.tsx
JeroenReumkens: Removed 👍

JeroenReumkens:
@hariombalhara  Fixed the styles of the multi select. Let me know if this is approvable 😁  Thanks again! 
CleanShot 2022-10-10 at 14 59 23@2x
CleanShot 2022-10-10 at 14 59 05@2x

hariombalhara (approved):
Awesome !!

PeerRich (approved):
great work!

PR #4826, opened 2022-10-04
Some tweaks to the ui - use button - use flex justify-between

emrysal:
## What does this PR do?

Before:
image

After
image


PeerRich (approved):
easy

PR #4830, opened 2022-10-04
added ?tabName=setup to event-type detail link to give it active status

### LINKED ISSUE #4829: /event-types set default tab when opening details
append `?tabName=setup` to the link

PeerRich:
we use the query to assign active state to the nav, which means there is no initial active state, hence we added the query to the detail page link

before
CleanShot 2022-10-04 at 16 33 45@2x

after
CleanShot 2022-10-04 at 16 33 31@2x

### DIFF
--- a/apps/web/pages/v2/event-types/index.tsx
+++ b/apps/web/pages/v2/event-types/index.tsx
@@ -53,7 +53,7 @@ const Item = ({ type, group, readOnly }: { type: EventType; group: EventTypeGrou
   const { t } = useLocale();
 
   return (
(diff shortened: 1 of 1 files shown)


emrysal (approved):

zomars:
This PR shouldn't have been merged since it caused E2E to fail. Reverting for now.


Quoting Twist chat:

> We didn’t use to include query params on those links so before it gave use the eventype ID only. and now it returns a 123?tabName=setup
> image
> Will revert it since is not a proper solution anyways. If there’s no query parameter we should fallback to the first one and display it as active.

PR #4848, opened 2022-10-05
Feat/improve animation on public booking page

JeroenReumkens:
## What does this PR do?

Improved the animation of showing the timeslots on the public booking page. In the old version the columns will jump around a bit when animating, making it feel a bit janky and slower than it should be. I opted to set a few more fixed widths, which help a lot with the animation.

## Old (first showing the animation on click, after that shows the animation on load)
CleanShot 2022-10-05 at 15 14 20

## New (first showing the animation on click, after that shows the animation on load)
CleanShot 2022-10-05 at 15 16 08

## Type of change

## How should this be tested?

- [x] In the old version of the code the first column of the booking page (title, duration, timezone select, etc) was duplicate in the page. I removed it in the PR since it seems there was little need to have it duplicated as we could also solve this with css only. But we have to be 100% sure we didn't remove any additional functionality.
- [x] Test the booking page as embed
- [x] Make sure there's no broken styling introduced on any breakpoint


apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -1,10 +1,11 @@
 // Get router variables
+import autoAnimate from "@formkit/auto-animate";
 import { EventType } from "@prisma/client";
```
JeroenReumkens, line 3: Reading this file's diff is a lot easier if you click 'hide whitespace' and then 'apply and reload' at the top of this page:

CleanShot 2022-10-05 at 15 25 37@2x

PeerRich:
this looks great, whats missing to go "ready for review"?

JeroenReumkens:
@PeerRich mainly the last two bullets under how to test. I need to test it one final time myself, and didn’t test it at all as an embed yet. If anyone else does that it’s also good to go.

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -402,203 +394,109 @@ const AvailabilityPage = ({ profile, eventType }: Props) => {
...
-          <div
-            style={availabilityDatePickerEmbedStyles}
-            className={classNames(
-              isBackgroundTransparent
-                ? ""
-                : "dark:bg-darkgray-100 sm:dark:border-darkgray-300 bg-white pb-4 md:pb-0",
-              "border-bookinglightest overflow-hidden rounded-md md:border",
-              isEmbed ? "mx-auto" : maxWidth
-            )}>
-            {/* mobile: details */}
```
hariombalhara, line 419: Screenshot 2022-10-10 at 4 44 39 PM
Back button wasn't there earlier on mobile(should be a bug), but it is showing now. 

Location of Back button should improve for mobile.Seems ill placed.

hariombalhara (changes requested):
Small but very impactful PR.
- Makes the component more dry. No more mobile specific JSX.
- Reduces complexity of the component

Found just 1 concern while testing. Seems important to me, so blocking merge

apps/web/components/booking/pages/AvailabilityPage.tsx
JeroenReumkens: Attention to detail, nice!! The breakpoint was wrong indeed, updated it from sm to md like it should've been. Thanks! 🙏

hariombalhara (approved):

PR #4849, opened 2022-10-05
Ensure we have to correct tab order for addons.

JeroenReumkens:
## What does this PR do?
Ensures we have to correct tab order for addons. Before addons AFTER …the input still would get tab focus before the input.

https://www.loom.com/share/0e078015bddb48b4bb7b1d27986f934e

## Type of change

## How should this be tested?

- [ ] Confirm that password peek gets focus after the input got focus
- [ ] Verify other inputs' layout with leading and suffixed addons - see Loom and eg the event setup page.

### DIFF
--- a/packages/ui/v2/core/form/fields.tsx
+++ b/packages/ui/v2/core/form/fields.tsx
@@ -164,6 +164,26 @@ type InputFieldProps = {
     labelClassName?: string;
   };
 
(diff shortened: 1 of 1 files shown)


sean-brydon (approved):
LGTM

emrysal (approved):
Works! 🚀

PR #4871, opened 2022-10-06
Add successful booking redirect URL to booking page

### LINKED ISSUE #4868: URL redirection field missing from the advanced settings of the event type
This didn't make the 2.0

sean-brydon:
Adds success redirect URL 

Currently redirect url is editable by all users - not sure if this is the current implementation that we want. 

CleanShot 2022-10-06 at 12 17 07@2x


pumfleet (approved):
Looks good to me

PR #4876, opened 2022-10-06
fixes to booking page

### LINKED ISSUE #4872: There is no need to display this redundant information.
### Discussed in https://github.com/calcom/cal.com/discussions/4853

<sup>Originally posted by **JoeyFenny** October  5, 2022</sup>
Anyone knowledgeable about landing pages and conversion rates will tell you, the last thing you should be doing is throwing too much irrelevant information at your user. You want to keep it clear and concise.

image

Displaying my URL (e.g. joey-fenny-egimdb/15-minute-meeting) gives too much information to the user, makes the page uglier (the excess whitespace on the right of the row items also makes it uglier), and serves little to no purpose:
The user more likely than not does not need or care for this information and if they did, most people would know to look in the browser URL bar or right-click and copy the link address from the context menu. I have never seen another website display the link behind a title, alongside the title. It is not a common practice.

I vote to clean this screen up and stop displaying it.

Also, I'm not sure the extra 5-pixel indentation of 15m 1-on-1 compared to the row title's starting point was intentional.

---

This one is a little more aggressive, but while we're at it, the user also doesn't need to know the available meeting types in the left column. Before they weren't displayed and I think we should go back to that. It is too much information and can cause confusion (the user needs to stop to process that information only to realize there is nothing for them to do with it) and it is redundant because this information would be shown on the next screen once they select a time anyway. Once again, I vote to remove it (Zoom, phone, Google Meet)

image

---
While we're at it still, this page looks cluttered with not enough whitespace if the checkbox label extends to two lines.
image

</div>

PeerRich:
fixed line break for checkbox on booking page and removed event type URL from profile page

closes https://github.com/calcom/cal.com/issues/4872

CleanShot 2022-10-06 at 13 35 09@2x

__

before:

CleanShot 2022-10-06 at 13 37 09@2x

after:

CleanShot 2022-10-06 at 13 34 32@2x

CleanShot 2022-10-06 at 13 34 54@2x

### DIFF
--- a/apps/web/components/booking/pages/BookingPage.tsx
+++ b/apps/web/components/booking/pages/BookingPage.tsx
@@ -712,23 +712,25 @@ const BookingPage = ({
                         />
                       )}
                       {input.type === EventTypeCustomInputType.BOOL && (
-                        <div className="flex h-5 items-center">
-                          <input
-                            type="checkbox"
-                            {...bookingForm.register(`customInputs.${input.id}`, {
-                              required: input.required,
-                            })}
-                            required={input.required}
-                            id={"custom_" + input.id}
-                            className="h-4 w-4 rounded border-gray-300 text-black focus:ring-black disabled:bg-gray-200 ltr:mr-2 rtl:ml-2 disabled:dark:text-gray-500"
-                            placeholder=""
-                            disabled={disabledExceptForOwner}
-                          />
-                          <label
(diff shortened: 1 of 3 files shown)


pumfleet (commented):
code changes are fine but not sure on submodules

emrysal (approved):
lgtm

PR #4888, opened 2022-10-06
Updating API submodule to latest

leog:
In order to reference the latest changes regarding booking API supporting recurring event types.


leog:
Not needed.

PR #4906, opened 2022-10-07
Delete Dialog once closed, makes the entire app unresponsive.

### LINKED ISSUE #4864: Delete Dialog once closed, makes the entire app unresponsive.
See the loom https://www.loom.com/share/18065a8a4e92472887f0fa28fa6e3235. 

Doesn't happen with embed modal box

gitstart:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/4864

 Loom Video: https://www.loom.com/share/579d91612ec0443a9190d585b2f75e6d

## Type of change


apps/web/pages/v2/event-types/index.tsx
```diff
@@ -278,7 +278,7 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
                               combined
                             />
                           </Tooltip>
-                          <Dropdown>
+                          <Dropdown modal={false}>
```
hariombalhara, line 281: Thanks for this @gitstart
Are we losing some functionality of Dropdown if we use modal={false}? Otherwise it looks good.

gitstart: > Thanks for this @gitstart Are we losing some functionality of Dropdown if we use modal={false}? Otherwise it looks good.

Hi @hariombalhara! Thank you for the review.
The only difference is that when the modal is set to true it disables the interaction outside elements.

hariombalhara (approved):

PR #4907, opened 2022-10-07
Reintroduce SAML SSO

### LINKED ISSUE #4880: reintroduce SAML SSO
it existed in 1.9 but got lost in transition to 2.0

devkiran:
## What does this PR do?

Fixes #4880

## Type of change


PeerRich:
@devkiran can you do a self review of your changes? https://handbook.cal.com/engineering/self-reviews

apps/web/pages/v2/settings/security/sso.tsx
```diff
@@ -0,0 +1,39 @@
...
+const SAMLSSO = () => {
+  const router = useRouter();
+
+  if (HOSTED_CAL_FEATURES) {
+    router.replace("/404");
+  }
+
+  const teamsView = false;
+
+  const { data: saml, isLoading } = trpc.useQuery(["viewer.saml.access", { teamsView }], {
```
emrysal, line 19: The access endpoint does not throw when access is denied, instead the response is success:

image

Ideally the access menu item is hidden when the server does not support it, if the server does support it but the user can set something up that allows access, the menu should be shown with the steps needed taking to obtain access.

Currently if access is false an empty saml page is shown:

image

apps/web/public/static/locales/en/common.json
```diff
@@ -1234,7 +1235,7 @@
   "theme_dark": "Dark",
   "theme_system": "System default",
   "add_a_team": "Add a team",
-  "saml_config": "SAML Config",
+  "saml_config": "SAML SSO",
```
emrysal, line 1238: Checked with @PeerRich and agreed the menu item should be called "Single Sign-On".

emrysal (changes requested):
See points mentioned in review; but overall this feature deserves some love, I'd like to know about the current users and if the two menu items are necessary. Also there is still a pro plan check in `canAccess` that probably needs to change. Did you have any thoughts on this?

apps/web/public/static/locales/en/common.json
PeerRich, line 1238:
```suggestion
  "saml_config": "Single Sign-On",
```

apps/web/pages/v2/settings/security/sso.tsx
devkiran: @emrysal Thanks for your feedback. I'll make the changes based on your feedback.

There are no significant changes compared to our initial implementation. Just some refactoring and UI changes.

The SAML SSO was there in 2 places in the original implementation. So kept it as it was. The Pro plan check was there already. I assumed only PRO users could access the SAML SSO. Am I right?

I'll discuss the above points with @deepakprabhakara and will make to a plan.

apps/web/public/static/locales/en/common.json
devkiran: Noted.

PeerRich:
@devkiran can you unblock this branch/PR? https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md#making-a-pull-request

otherwise we can't update with the base branch 
CleanShot 2022-10-10 at 09 29 38@2x

apps/web/pages/v2/settings/security/sso.tsx
deepakprabhakara: @emrysal We already make the right check and show/hide the SSO menu item. We'll handle the access denied case and show an error message to contact the admin (which is the only logical step possible in this case, let us know if you'd like to handle it in another way).

We'll make changes to hide the SSO menu item in Teams for self-hosted and vice versa for hosted version (hide SSO menu item in Settings -> Security) since there needs to only be one place for this in both scenarios.

Please let us know if the PRO user check for access regarding SSO has changed and we'll make the right check.

devkiran:
@PeerRich I can't see that option; unfortunately, I think it is because I made the fork under the organization boxyhq. If you want, I can open a new PR after closing this.

apps/web/pages/v2/settings/security/sso.tsx
emrysal: @PeerRich Could you update with the specifics on the PRO plan and how it impacts SAML post 2.0?

PeerRich:
> @PeerRich I can't see that option; unfortunately, I think it is because I made the fork under the organization boxyhq. If you want, I can open a new PR after closing this.

yes please reopen.🙏

devkiran:
@PeerRich Here is the new PR https://github.com/calcom/cal.com/pull/4938

PR #4943, opened 2022-10-11
#4252 Use vercel og to generate og images for meetings and apps.

### LINKED ISSUE #4252: og-image (social media preview) for apps (app.cal.com/apps/zoom)
CleanShot 2022-09-07 at 19 43 03@2x

like raycast: https://www.raycast.com/eluce2/cal-com-share-meeting-links

CleanShot 2022-09-07 at 19 49 55@2x

JeroenReumkens:
## What does this PR do?

Generates OG images for apps and meetings with the new Vercel OG. 
Pages it generates images for: 
* `/apps/...`
* `/[user]`
* `/[user]/[type]`
* `/team/[slug]`
* `/team/[slug]/[type]`

## Small tweaks needed after a check in prod

It seems that for the avatars there's quite a few ways in which they might be used; users do/don't have avatars, teams do/don't, and sometimes it seems to return an avatar while other times it could return a base64 string (which @zomars  I believe is removing soon). So I can imagine there's some edge cases that I couldn't find locally (image upload locally also doesn't work on my end), so I think we also have to test this heavily in production, and see which cases aren't handled yet. If an image breaks, it's the api route that breaks for that one image, so the image simply won't load. It should not impact the page itself. Curious what others think about this 😄 

### Single user

image

### Collection, lots of people and long title

image

### Team of 2 (collection)
image

### Round robin
image

### Dynamic collective

image

### App

image

Fixes #4252 

## Type of change

## How should this be tested?

- [ ] Ensure OG images render properly on the pages mentioned at the beginning of this PR. You find the OG images in the source of the page, look for a meta tag with the name `og:image`


packages/app-store/dailyvideo/static/icon.svg
```diff
@@ -1,4 +1,4 @@
-<svg id="svg" version="1.1" xmlns="http://www.w3.org/2000/svg" width="1.25em" height="1.25em" viewBox="0, 0, 400,400">
+<svg id="svg" xmlns="http://www.w3.org/2000/svg" width="40" height="40" viewBox="0,0,400,400">
```
JeroenReumkens, line 1: Had to tweak a few SVG's because otherwise vercel/og wouldn't render them.

zomars (approved):
LFG

Ship it

PR #4954, opened 2022-10-11
Hotfix: Fix Routing Form Saving related issue

### LINKED ISSUE #4950: [CAL-54] SIU routing forms not saving
[https://app.birdeatsbug.com/sessions/4QAk-rDvlDH5nGJO9DTXvUtB6Ka7WaYSoplDNrFA-Vic](https://app.birdeatsbug.com/sessions/4QAk-rDvlDH5nGJO9DTXvUtB6Ka7WaYSoplDNrFA-Vic)

hariombalhara:
Fixes https://github.com/calcom/cal.com/issues/4950

The issue seems to occur with a combination of following:
- Code should be compiled(not in dev mode. This is the reason this bug doesn't replicate normally on local as we use dev mode)
- Middleware should be used(which we started using a month or 2 back)

Solution, is to use tRPC fetched form for rendering, which would be invalidated on every save and works correctly.

### DIFF
--- a/packages/app-store/ee/routing-forms/components/SingleForm.tsx
+++ b/packages/app-store/ee/routing-forms/components/SingleForm.tsx
@@ -175,11 +175,7 @@ const Actions = ({
   );
 };
 
-export default function SingleForm({
-  form,
-  appUrl,
-  Page,
-}: {
+type SingleFormComponentProps = {
   form: RoutingFormWithResponseCount;
   appUrl: string;
   Page: React.FC<{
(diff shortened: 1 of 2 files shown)


PeerRich (approved):

packages/app-store/ee/routing-forms/components/SingleForm.tsx
```diff
@@ -262,3 +259,20 @@ export default function SingleForm({
     </Form>
   );
 }
+
+export default function SingleFormWrapper({ form: _form, ...props }: SingleFormComponentProps) {
+  const { data: form, isLoading } = trpc.useQuery(["viewer.app_routing_forms.formQuery", { id: _form.id }], {
```
hariombalhara, line 264: Because we are using tRPC now, it is important that useForm is called in a child component which is rendered only when `form` is available. THis is to avoid initialization of useForm with defaultValues:undefined

packages/app-store/ee/routing-forms/components/SingleForm.tsx
```diff
@@ -175,35 +175,32 @@ const Actions = ({
   );
 };
```
hariombalhara, line 177: Will fix Eslint warning in a follow up PR>

hariombalhara (commented):
Self reviewed

PR #4961, opened 2022-10-12
Handle double quotes situation in checkLicense function

emrysal:
## What does this PR do?

This fixes the likely culprit to our invalid `CALCOM_LICENSE_KEY` API issues. When you create a key and save the .env file on windows there may be wrapping `""`'s. 

Another situation where this happens is in linux when you accidentally add a trailing space after the .env file.
`CALCOM_LICENSE_KEY="xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"<space>`

Also adds validation for the `CALCOM_LICENSE_KEY` format.

### DIFF
--- a/packages/features/ee/common/server/checkLicense.ts
+++ b/packages/features/ee/common/server/checkLicense.ts
@@ -1,12 +1,26 @@
 import cache from "memory-cache";
+import { z } from "zod";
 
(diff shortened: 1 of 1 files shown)


zomars (approved):
We might need to validate every single env variable 🙃

PR #4966, opened 2022-10-12
added new license required /signup page

### LINKED ISSUE #3600: [CAL-76] Make app signup page work if calcom license is present
Replace current signup to be dynamic, and detect presence of license and allow signups if EE

PeerRich:
- [x] remove apps/api/pages/signup.tsx and use rewrites instead @emrysal 

CleanShot 2022-10-12 at 09 54 54@2x


apps/web/pages/signup.tsx
```diff
@@ -0,0 +1,187 @@
+// code duplicate of apps/api/auth/signup.tsx, could not find a way to import without breaking SSR
```
PeerRich, line 1: when importing from /api/
CleanShot 2022-10-12 at 09 53 11@2x

we get this

CleanShot 2022-10-12 at 09 53 00@2x

PeerRich: likely because /api is a private submodule?

PeerRich:
I think we can replace /apps/api/pages/signup with this one, too

emrysal:
Redirect installed, just making some modifications to getServerSideProps and testing invite flow.

emrysal (approved):
I tested it through fully so merging, I'll deal with the consequences.

zomars:
@PeerRich two cents here. We might want to make this opt-in by default. So we don't disrupt our hosted and self-hosted users.

PeerRich:
> @PeerRich two cents here. We might want to make this opt-in by default. So we don't disrupt our hosted and self-hosted users.

why would it disrupt it?

zomars:
> > @PeerRich two cents here. We might want to make this opt-in by default. So we don't disrupt our hosted and self-hosted users.
> 
> why would it disrupt it?

Let's say a self-hosted user wants to control who can signup (like we did pre-v2) and suddenly after an update they have a new public signup page for anyone to use.

kevinscholz:
Just happened 🙈 Could you give me a hint on how to deactivate the sign-up page or to limit it to a specific Google SSO domain?

zomars:
@kevinscholz you could just delete the `apps/web/pages/signup.tsx` file and rebuild.

kevinscholz:
@zomars Thanks for the hint. That also makes the API route `/api/auth/signup` unusable?

zomars:
> @zomars Thanks for the hint. That also makes the API route `/api/auth/signup` unusable?

If you want total control over it then you would need to remove that as well.

apps/web/pages/signup.tsx
```diff
@@ -0,0 +1,194 @@
...
+            callbackUrl: (`${WEBAPP_URL}/${router.query.callbackUrl}` || "") as string,
+          })
+      )
+      .catch((err) => {
+        methods.setError("apiError", { message: err.message });
+      });
+  };
+
+  return (
+    <LicenseRequired>
```
zomars, line 69: Two extra cents here @emrysal @PeerRich This page is mostly going to be accessed publicly and LicenseRequired needs a logged in user in order to work. So all unauthed users will se the "license required" screen.

PR #4971, opened 2022-10-12
Feat/3334 show times in timezones for bookingpage

### LINKED ISSUE #3334: [CAL-74] Always list Time Zones

Problem: Since teams are becoming more and more global, consumers are always juggling timezones in their head. 

Wherever there is a time listed in Cal.com's settings, email, scheduler link, etc. list the time zone associated with it. The timezone of any booking should always be listed.
CleanShot 2022-07-11 at 17 20 40@2x

JeroenReumkens:
## What does this PR do?

Adds a popup/tooltip to view the meeting time in the different timezones of all attendees. 

### Behaviour: 
* Only shows unique timezones, doesn't show duplicates
* If everyone is in the same timezone popup won't show
* Timezones are sorted by their GMT timezone (earliest to latest)
* If someone meets a day before / after we show -1 / +1
* As discussed with Ciaran we don't show avatars for now

Fixes #3334

https://www.loom.com/share/2caee7c38aa74a588f83ccec9af5672b

## Type of change

## How should this be tested?

- [ ] Create different meetings and verify that the popup shows the correct times in the respective timezones.


CarinaWolli:
@JeroenReumkens I just tested it and it shows +1, even when it is the same day: 
Screenshot 2022-10-18 at 17 09 08

leog:
@JeroenReumkens I was testing it in https://cal-auu1mnfp0-cal.vercel.app/bookings/upcoming and clicking in the popover was actually sending me to the booking success page 🤔

JeroenReumkens:
@leog  Thanks, will also retest that tomorrow! In theory [this line](https://github.com/calcom/cal.com/pull/4971/files#diff-178e89024734f552c5fa0953359ded5880f0e6975ae88e7bb085b3908c762abdR194) should prevent that, but then it doesn't seem to do that according to your tests 🤔  Will recheck.

zomars:
@JeroenReumkens let's also do [a self-review from now](https://handbook.cal.com/engineering/self-reviews).

apps/web/components/booking/BookingListItem.tsx
```diff
@@ -185,7 +186,7 @@ function BookingListItem(booking: BookingItemProps) {
 
   const location = booking.location || "";
 
-  const onClick = () => {
+  const onClickTableData = (ev: React.MouseEvent<HTMLDivElement>) => {
```
alannnc: @JeroenReumkens Renamed to better understanding and don't collide with onClick.

packages/ui/v2/core/MeetingTimeInTimezones.tsx
```diff
@@ -0,0 +1,85 @@
...
+  if (times.length === 1) return null;
+
+  return (
+    <Popover.Root>
+      <Popover.Trigger className="popover-button ml-2 inline-flex h-5 w-5 items-center justify-center rounded-sm text-gray-900 transition-colors hover:bg-gray-200 focus:bg-gray-200">
+        <Icon.FiGlobe />
+      </Popover.Trigger>
+      <Popover.Portal>
+        <Popover.Content
+          onClick={(e) => e.stopPropagation()}
```
alannnc, line 60: This should stop navigating to /success @JeroenReumkens

packages/lib/date-fns/index.ts
```diff
@@ -6,3 +6,73 @@ export const yyyymmdd = (date: Date | Dayjs) =>
...
+    .utcOffset();
+
+  return timeZone
+    ? dayjs
+        .utc(date)
+        .utcOffset(utcOffset)
+        .format(timeFormat === 12 ? "h:mma" : "HH:mm")
+    : dayjs(date).format(timeFormat === 12 ? "h:mma" : "HH:mm");
+};
+/**
```
alannnc, line 30:
```suggestion

/**
```
we add some space between functions here

alannnc (approved):
Code works now. 
NIT: What do you think of adding this so you know which one is yours.

image

zomars (approved):
LGTM

apps/web/components/booking/BookingListItem.tsx
JeroenReumkens: @alannnc  Cool! Main reason I added it in here instead of the default behavior of the Popover is because the click behavior is actually something the BookingListItem tries to control, and not something the Popover should know about. But I sticked with your suggestion and added it to the tooltip button as well. Abstracted it to a function to so I could add a comment as to why this is needed 👍

JeroenReumkens:
Fixes mentioned in PR are done. Also make the `sortByTimezone` function simpler by utilising the `utcOffset` in `dayjs`, which I did not discover during my search for something like this before 😅 

@CarinaWolli  There indeed was a bug in comparing prev / next day 🙏  Thanks so much for discovering this. This is fixed now. Could you please retest it? 

@leog Clicking the popup content now doesn't redirect you to the detail page anymore 😄  Thanks for spotting that, I actually only tested clicking the button, not the content I realised!  Could you please retest? 

@alannnc I reverted your date format change for #5050 as you told me to. I did NOT revert all the other changes in that PR, was that actually the plan? 

Thanks all!

PR #4992, opened 2022-10-13
Fix/fallback to calvideo

alannnc:
## What does this PR do?

If createMeeting failed with user credentials we fallback to create video with cal.video

Fixes # (issue)

 Loom Video: https://www.loom.com/share/99c739ae26f64f34976b081d57913a2b

## Type of change

## How should this be tested?

- [ ] Install zoom, setup event with location with zoom.
- [ ] Book your event, and see zoom link in the email.
- [ ] In DB change expire_date to the past and change any character in access_token, refresh_token to invalidate them.
- [ ] Book yourself again and you should receive Email with Integration error and owner and attendees emails with cal.video link as a fallback.

## Checklist


packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -45,6 +46,7 @@ export const zoomMeetingsSchema = z.object({
 });
 
 // Successful API response
+// @TODO: add link to the docs
```
alannnc, line 49: This both TODO can wait a little, updating them in the next iteration of this bugfix.

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -72,24 +84,33 @@ const zoomAuth = (credential: Credential) => {
...
-          },
-          data: {
-            key: responseBody,
-          },
-        });
-        return responseBody.access_token;
-      });
+    });
+    const responseBody = await handleErrorsJson(response);
+    await handleErrorResponseFromZoom(responseBody, credential.id);
```
alannnc, line 89: Function that explicitly looks for error object in zoom response.

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -98,12 +119,8 @@ const zoomAuth = (credential: Credential) => {
       try {
         credentialKey = zoomTokenSchema.parse(credential.key);
       } catch (error) {
-        // If the parse fails, it means we have an invalid credential saved. We should delete it.
-        await prisma.credential.delete({ where: { id: credential.id } });
-        // We reject the promise, and should handle the error on the UI
-        return Promise.reject(
-          "This Zoom credential was malformed so we've removed it from the DB. Please add a new Zoom credential"
-        );
+        await invalidateCredential(credential.id);
```
alannnc, line 122: Instead of raw deleting credential we now mark it as invalid key.

packages/core/videoClient.ts
```diff
@@ -44,18 +44,28 @@ const createMeeting = async (credential: Credential, calEvent: CalendarEvent) =>
...
 
-  if (!createdMeeting) {
-    return {
-      type: credential.type,
-      success: false,
-      uid,
-      originalEvent: calEvent,
-    };
+    // Default to calVideo
+    const defaultMeeting = await createMeetingWithCalVideo(calEvent);
```
alannnc, line 64: If error is throw before getting meeting result object we fallback to cal video

PeerRich:
do we inform the user that the link has changed due to an issue with {integration}?

zomars (approved):
Looks good enough for a hot fix. Let's fix those type errors before merging first tho.

alannnc:
> do we inform the user that the link has changed due to an issue with {integration}? @PeerRich 

I think we can add that text in several places e.g. UI and broken integration email.
For now this should do as a hotfix so we don't broke anymore zoom credentials while refreshing the tokens.

packages/app-store/applecalendar/lib/CalendarService.ts
```diff
@@ -1,9 +1,9 @@
-import { Credential } from "@prisma/client";
+import { Prisma } from "@prisma/client";
 
 import CalendarService from "@calcom/lib/CalendarService";
 
 export default class AppleCalendarService extends CalendarService {
-  constructor(credential: Credential) {
+  constructor(credential: Prisma.CredentialCreateInput) {
```
alannnc, line 6: @zomars I believe this was the root of the type check errors.

packages/app-store/caldavcalendar/api/add.ts
```diff
@@ -32,7 +32,6 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
 
     try {
       const dav = new CalendarService({
-        id: 0,
```
alannnc, line 35: This had to go since probably was put there to comply with type check

packages/app-store/applecalendar/lib/CalendarService.ts
zomars: How? It was working well before.

packages/app-store/caldavcalendar/api/add.ts
zomars: Let's keep this PR as a mere hotfix. We might need to revert the migration so we don't have to deal with typings. Let's keep this fix simple.

alannnc: reverting

alannnc: It was working well before but it just showed me that the type it's not open to changes as it should be.

Issue #4970, opened 2022-10-12
[CAL-82] /settings/teams/new: auto-add team URL when typing in team name


PeerRich:
![CleanShot 2022-10-12 at 12 41 38](https://user-images.githubusercontent.com/8019099/195333794-6368dce5-8766-4586-b050-f2ce37212bd1.gif)

same as in event-types:

https://user-images.githubusercontent.com/8019099/195334237-8ce50699-e742-41e6-a913-6ae281dfe0fe.mp4

harshsinghatz:
Can I work on this??

PeerRich:
> Can I work on this??

sure!

Udit-takkar:
PR has been merged. Closing this

PR #4997, opened 2022-10-14
Harsh/auto add url

harshsinghatz:
## What does this PR do?
- It auto completes the event-url/ team-url for the first time 
- It only works till the time the url field is not updated because we not completely binding the two fields

refer [this issue](https://github.com/calcom/cal.com/issues/4970)

Fixes # (issue)

[Loom video link](https://www.loom.com/share/d4458a4bbe2b4cbe9c3fa8349fe51c66)

## Type of change

## How should this be tested?

- [ ] e2e
- [ ] tested locally


packages/ui/v2/modules/event-types/CreateEventType.tsx
```diff
@@ -202,7 +204,15 @@ export default function CreateEventTypeButton(props: CreateEventTypeBtnProps) {
...
-            <TextField label={t("title")} placeholder={t("quick_chat")} {...register("title")} />
+            <TextField
+              label={t("title")}
+              placeholder={t("quick_chat")}
+              {...register("title")}
+              onChange={(e) => {
+                form.setValue("title", e?.target.value);
+                setTitleCopy(e?.target.value);
+              }}
+            />
```
PeerRich, line 215: why do we need to do this here too?

alannnc (changes requested):
Added a function that converts text to url friendly.
Removed states as they were not necessary in this case.

NIT: branch name with action/description in the future
feature/auto-add-url-on-team-creation

Just undo changes on CreateEventType.tsx and should be ready to go.
@harshsinghatz

harshsinghatz:
> Added a function that converts text to url friendly. Removed states as they were not necessary in this case.
> 
> NIT: branch name with action/description in the future feature/auto-add-url-on-team-creation
> 
> Just undo changes on CreateEventType.tsx and should be ready to go. @harshsinghatz

The problem with the approach that you have implemented is that you're binding the slug value to the name 

That means if the name changes the slug value changes as well. If this is an expected behaviour then your approach works well

harshsinghatz:
> Added a function that converts text to url friendly. Removed states as they were not necessary in this case.
> 
> NIT: branch name with action/description in the future feature/auto-add-url-on-team-creation
> 
> Just undo changes on CreateEventType.tsx and should be ready to go. @harshsinghatz

Also there is util func named "slugify"

packages/ui/v2/modules/event-types/CreateEventType.tsx
harshsinghatz: This was here because in the first approach the updated value of title was needed. But i could have used watch. So this step was unnecessary.

harshsinghatz:
Can someone review the changes?

PeerRich:
> Can someone review the changes?

will do monday, enjoy your weekend!

apps/web/components/getting-started/components/Steps.tsx
```diff
@@ -22,12 +22,12 @@ const Steps = (props: ISteps) => {
               key={`step-${index}`}
               onClick={() => navigateToStep(index)}
               className={classNames(
-                "h-1 w-1/4 rounded-[1px] bg-black dark:bg-white",
+                "h-1 w-full rounded-[1px] bg-black dark:bg-white",
```
alannnc, line 25: Fixed width

packages/features/ee/teams/components/v2/CreateNewTeam.tsx
```diff
@@ -45,7 +46,11 @@ const CreateANewTeamForm = (props: { nextStep: () => void; setTeamId: (teamId: n
               value={value}
               onChange={(e) => {
                 formMethods.setValue("name", e?.target.value);
+                if (formMethods.formState.touchedFields["slug"] === undefined) {
```
alannnc, line 49: @harshsinghatz We can add this touchedFields validation

alannnc (approved):

PR #4982, opened 2022-10-13
Add links to Zapier templates in installed apps

CarinaWolli:
## What does this PR do?

- Adds new "Automation" tab to Installed apps
- Adds a new component for App specific settings and used it for Zapier to show the templates and link them to the actual Zapier templates. 
- Small mobile design fixes for installed apps 

Fixes #4048

Loom: https://www.loom.com/share/f84a6e5e1f7c4d7fbf61c03612bbfe6b

**Environment**: Staging(main branch)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How should this be tested?

- Install Zapier 
- Go to installed apps -> automation
- See the new Zapier templates
- Click on 'Use zap' and check if you are redirected to the correct template

PR #4886, opened 2022-10-06
added weather in calendar app 

PeerRich:
CleanShot 2022-10-06 at 18 14 29@2x

https://user-images.githubusercontent.com/8019099/194724607-fe848508-0fcd-4f45-af2d-2991424c5085.mp4

todo:
- [ ] needs app-wide settings @hariombalhara  :

CleanShot 2022-10-06 at 18 15 08@2x

PR #4998, opened 2022-10-14
Add AppStore General Settings support

hariombalhara:
## What does this PR do?

Fixes # (issue)
- Recognizes AppSettings.tsx as a special component that allows adding App Settings to any App.
- Weather App comes from https://github.com/calcom/cal.com/pull/4886 and utilizes App Settings.

This is based on @CarinaWolli's work here https://github.com/calcom/cal.com/pull/4982. This PR is mergeable after #4982 is merged.

**Environment**: Production

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


hariombalhara (commented):
Self reviewed.

apps/web/pages/v2/apps/installed/[category].tsx
```diff
@@ -96,19 +96,18 @@ interface IntegrationsListProps {
...
         <IntegrationListItem
           name={item.name}
           slug={item.slug}
-          key={item.name}
+          key={item.title}
           title={item.title}
           logo={item.logo}
           description={item.description}
-          separate={separateItems}
+          separate={true}
```
Comment (hariombalhara) [inline apps/web/pages/v2/apps/installed/[category].tsx:110]: Always show in separate mode as AppSettings are supported by all apps now.

packages/app-store/_components/AppSettings.tsx
```diff
@@ -1,11 +1,9 @@
-import dynamic from "next/dynamic";
+import { AppSettingsComponentsMap } from "@calcom/app-store/apps.browser.generated";
 
 import { DynamicComponent } from "./DynamicComponent";
 
-export const AppSettingsMap = {
```
hariombalhara, line 5: It is autogenerated now in apps.browser.generated.ts

hariombalhara (commented):
Reviewed again

apps/web/pages/v2/apps/installed/[category].tsx
Comment (CarinaWolli) [reply apps/web/pages/v2/apps/installed/[category].tsx:110]: I think the separate prop is not needed anymore now. So we should remove it

packages/app-store/weather_in_your_calendar/components/AppSettings.tsx
```diff
@@ -0,0 +1,38 @@
+import { useState } from "react";
+
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+import Button from "@calcom/ui/v2/core/Button";
+import { TextField } from "@calcom/ui/v2/core/form/fields";
+
+/*
```
CarinaWolli, line 7: Can we remove the comments? Or are they still needed for some future work?

CarinaWolli (commented):
Looks good to me 👍 Just added some minor comments! And should we add the top gray-200 border for all appSettings to keep the design consistent? So the weather app would then look like this:
Screenshot 2022-10-15 at 10 44 58

packages/app-store/weather_in_your_calendar/components/AppSettings.tsx
hariombalhara: I think it's needed for future work. So, will keep them as is

hariombalhara:
Added border. Makes sense to me.

apps/web/pages/v2/apps/installed/[category].tsx
Comment (hariombalhara) [reply apps/web/pages/v2/apps/installed/[category].tsx:110]: I still see one usage of IntegrationListItem where separate property isn't passed to be true.So, it can't be removed yet.

PeerRich (approved):

PR #5016, opened 2022-10-14
follow up: 4978 cal 87 re add tip in sidebar

PeerRich:


hariombalhara (approved):
Looks good except that re-rendering of Tips(which can only be fixed if all routes move to using getLayout approach)

hariombalhara:
@PeerRich I think this is mergeable except for the conflict?

PR #5026, opened 2022-10-15
Fixes double bottom border in Integration List Item

hariombalhara:
## What does this PR do?

Fixes # (issue)

Notice the bottom border in case there are no App Settings
image

It was introduced here [`11fee3a` (#4998)](https://github.com/calcom/cal.com/pull/4998/commits/11fee3a415cd2ba4a2af17cfabb58cf1148dde50)

### DIFF
--- a/packages/app-store/_components/DynamicComponent.tsx
+++ b/packages/app-store/_components/DynamicComponent.tsx
@@ -1,9 +1,17 @@
-export function DynamicComponent<T extends Record<string, any>>(props: { componentMap: T; slug: string }) {
+export function DynamicComponent<T extends Record<string, any>>(props: {
+  componentMap: T;
+  slug: string;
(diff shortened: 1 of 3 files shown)


packages/app-store/_components/DynamicComponent.tsx
```diff
@@ -1,9 +1,17 @@
...
+}) {
   const { componentMap, slug, ...rest } = props;
 
   if (!componentMap[slug]) return null;
 
   const Component = componentMap[slug];
 
-  return <Component {...rest} />;
+  return (
+    <div className={props.wrapperClassName || ""}>
```
hariombalhara, line 13: Adds the className support through wrapper so that the app's AppSettings component never requires to read className

zomars (approved):
Ship it

PR #4738, opened 2022-09-28
Fix long url and show password

leog:
## What does this PR do?

Long Website URLs were breaking the admin onboarding of the app aside from a misplaced "Show Password" icon, see:

<kbd>image</kbd>

Now fixed detecting Website URLs greater than 30 characters:

<kbd></kbd>

Fixes #4703 

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?
Bring up the app having deleted all users in the database to let the Admin Onboarding appear when trying to login for the first time in the app.

PR #3171, opened 2022-06-27
Team event type scroll, manage team invite from mobile

leog:
## What does this PR do?

Fixes a couple of UI bugs.

#### Original Mobile:
image

#### Original Desktop:
Teams_accept_buttons

#### Fixed Mobile:
Teams_accept_with_dropdown
Screenshot 2022-06-27 at 12 39 17

#### Original Mobile
image

#### Fixed Mobile
https://user-images.githubusercontent.com/467258/176039791-f7fe4724-5045-4e32-967e-d726b6df95e5.mov

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

PR #5031, opened 2022-10-15
fixed team URL in settings

### LINKED ISSUE #5029: [CAL-161] wrong link in teams
instead of cal.com/team-name it needs to be cal.com/team/team-name in settings

PeerRich:
before
CleanShot 2022-10-15 at 21 29 22@2x

after
CleanShot 2022-10-15 at 21 29 55@2x


leog (commented):
Would it make sense to apply the fallback I introduced somewhere where we show the web app URL differently when it surpasses certain character limit? Otherwise if the instance has a long URL, it might break or become unusable.

PeerRich:
> Would it make sense to apply the fallback I introduced somewhere where we show the web app URL differently when it surpasses certain character limit? Otherwise if the instance has a long URL, it might break or become unusable.

do you know where we use it?

leog:
> do you know where we use it?

[This](https://github.com/calcom/cal.com/pull/4738)  is the last variation I worked on, there is another version[here](https://github.com/calcom/cal.com/pull/3171)  that goes for having the url in the label. We should probably settle on one strategy only 😅

PeerRich:
> > do you know where we use it?
> 
> [This](https://github.com/calcom/cal.com/pull/4738) is the last variation I worked on, there is another one [here](https://github.com/calcom/cal.com/pull/3171) that goes for having the url in the label. We should probably settle on one strategy only 😅

lets merge this and open a ticket to consolidate all of this into /lib/

hariombalhara (approved):

PR #5034, opened 2022-10-16
Apps Status + Updates to non-traditional calendars

leog:
## What does this PR do?

Implements Apps Status on emails for organizers. It also fixes updating meetings when rescheduling for non-traditional calendars. Recurring events get a (xNumber) alongside the status for each app to let the user know if every booking was successfully created. If any of them fails, it shows a cross emoji.

Some screenshots:
<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

## Type of change

## How should this be tested?

Book a regular or recurring event and see as the organizer an email sent to you with the apps status.


leog (commented):
Self-review done

PeerRich (approved):

apps/web/public/static/locales/en/common.json
```diff
@@ -543,6 +543,7 @@
   "link_shared": "Link shared!",
   "title": "Title",
   "description": "Description",
+  "apps_status": "Apps Status",
```
zomars, line 546: Would "Apps results" make more sense in this context?

packages/emails/email-manager.ts
```diff
@@ -28,7 +28,7 @@ export const sendScheduledEmails = async (calEvent: CalendarEvent) => {
     ...calEvent.attendees.map((attendee) => {
       return new Promise((resolve, reject) => {
         try {
-          const scheduledEmail = new AttendeeScheduledEmail(calEvent, attendee);
+          const scheduledEmail = new AttendeeScheduledEmail({ ...calEvent, appsStatus: undefined }, attendee);
```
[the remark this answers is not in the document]
zomars: Would it make more sense to simply "not add this unless is the organizer email" ? Instead of removing it everywhere?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -744,6 +784,7 @@ async function handler(req: NextApiRequest & { userId?: number }) {
           metadata.hangoutLink = updatedEvent.hangoutLink;
           metadata.conferenceData = updatedEvent.conferenceData;
           metadata.entryPoints = updatedEvent.entryPoints;
+          handleAppsStatus(results, booking);
```
zomars, line 787: Nice!

zomars (commented):
Very nice addition @leog left some feedback to address first 🙏🏽

apps/web/public/static/locales/en/common.json
leog: We need you @Jaibles!

packages/emails/email-manager.ts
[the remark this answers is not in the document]
leog: Makes sense. We actually rely on the Calendar Event everywhere as the only parameter to base any email decision so I went that way, include it there. I can always not include it in the Calendar Event and have it as a separate parameter to the email interface, such as `sendScheduledEmails` and the others, to then in the implementation only pass it down in the Calendar Event for organizer emails only.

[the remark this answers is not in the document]
zomars: Or delegate the appStatus logic the email template itself. Checks inside the template if the receipient is the organizer and render the `AppsStatus.tsx` component based on that condition.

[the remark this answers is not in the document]
leog: Done! Much better, thanks Omar!

packages/emails/src/templates/BaseScheduledEmail.tsx
```diff
@@ -70,6 +72,7 @@ export const BaseScheduledEmail = (
       <LocationInfo calEvent={props.calEvent} t={t} />
       <Info label={t("description")} description={props.calEvent.description} withSpacer />
       <Info label={t("additional_notes")} description={props.calEvent.additionalNotes} withSpacer />
+      {props.includeAppsStatus && <AppsStatus calEvent={props.calEvent} t={t} />}
```
leog, line 75: Here you go @zomars, great idea! Works like a charm

packages/emails/src/templates/OrganizerScheduledEmail.tsx
```diff
@@ -33,6 +33,7 @@ export const OrganizerScheduledEmail = (
       t={t}
       subject={t(subject)}
       title={t(title)}
+      includeAppsStatus={true}
```
zomars, line 36:
```suggestion
      includeAppsStatus
```

zomars (approved):
LGTM

PR #5036, opened 2022-10-16
Disable user errors from stdout

emrysal:
User errors like 400 Bad Request, 401 Unauthorized, etc, should not show up in server logs. These errors would quickly flood our logs and are not useful.

A future possibility is to add a DEBUG=1 flag of sorts to running the API in case you do want to have these errors in stdout.

Not sure this file should be located in @calcom/lib/server 🤔 .


packages/lib/server/defaultResponder.ts
```diff
@@ -15,8 +15,11 @@ function defaultResponder<T>(f: Handle<T>) {
       ok = true;
       if (result) res.json(result);
     } catch (err) {
-      console.error(err);
       const error = getServerErrorFromUnknown(err);
+      // user errors should not end up in stdout.
+      if (error.statusCode < 400 || error.statusCode >= 500) {
```
hariombalhara, line 20: There can be non-HTTP errors as well e.g. JSON parsing failure, that won't be logged this way.

zomars: Can we move the login alotgether into `getServerErrorFromUnknown`? There at least we can know for example if it's a ZodError and we may want to log it to cover new edge-cases.

PeerRich: @emrysal

PeerRich:
closing after talking to @emrysal

PR #5039, opened 2022-10-16
Adds a iso8601 helper function to parse a ISO Date String into a Date

emrysal:
For use inside Zod schemas. Turns a valid ISO string into a date object and otherwise shows: 

image


packages/prisma/zod-utils.ts
```diff
@@ -58,6 +58,20 @@ export const recurringEventType = z
   })
   .nullable();
 
+// dayjs iso parsing is very buggy - cant use :( - turns ISO string into Date object
+export const iso8601 = z.string().transform((val, ctx) => {
```
hariombalhara, line 62: We should have some usage of it in place as well. Is it being used somewhere?

hariombalhara (dismissed):
See the comment

packages/prisma/zod-utils.ts
emrysal: API 👍

hariombalhara (approved):

PR #5046, opened 2022-10-17
Fix disabled locations when rescheduling

### LINKED ISSUE #4981: [CAL-150] Event types with Attendee Phone Number as location can not be rescheduled
### Issue Summary

Bookings with Attendee Phone Number as the location can not be rescheduled because the input field for phone number is empty and disabled.

### Steps to Reproduce

1. Create an event type with Attendee Phone Number as the location 
2. Book event
3. Reschedule the same event

CarinaWolli:
## What does this PR do?

When rescheduling an event the radio buttons for choosing a location were disabled without anything selected. With this PR the disabled radio buttons are not shown anymore but the location of the previous booking which will also be the location of the rescheduled booking is shown instead. 

Also fixes: The input field for the phone number is not shown when Attendee Phone Number is selected as the first location. Only appears after choosing Organizer Phone Number and then again Attendee Phone Number. It also can be booked like that (so the booking location is then 'phone') (https://github.com/calcom/cal.com/issues/4981#issuecomment-1277623403)

Fixes #4981

 Loom Video: https://www.loom.com/share/53fde81383a84f8aa973fefd9668f505

## Type of change

## How should this be tested?

1. Reschedule a booking of an event type with several locations
2. See that it shows the location of the previous booking and not the radio buttons

1. Book an event with several locations and Attendee Phone Number should be the first location 
2. See that phone input is shown when Attendee Phone Number is default selected as the first location


apps/web/components/booking/BookingDescription.tsx
```diff
@@ -84,7 +85,7 @@ const BookingDescription: FC<Props> = (props) => {
             {t("requires_confirmation")}
           </div>
         )}
-        {!isBookingPage ? (
+        {!isBookingPage && !props.rescheduleUid ? (
```
CarinaWolli, line 88: Don't show available locations of event type when rescheduling (User can't choose between them, so no need to show them)

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -578,37 +577,48 @@ const BookingPage = ({
...
-                            {locationKeyToString(location)}
-                          </span>
-                        </label>
-                      );
-                    })}
-                  </div>
-                )}
+                    {rescheduleUid ? (
+                      <p className="mt-1 text-sm text-gray-500">
+                        {getHumanReadableLocationValue(booking?.location, t)}
```
CarinaWolli, line 587: Show the location of the previous booking when rescheduling

hariombalhara (approved):
Looks good as a bug fix 👍 

But I think we should work on allowing location change using reschedule as currently there is no way for the booker to change location. Only organizer can change the location.

PR #5050, opened 2022-10-17
Uses utcOffset to display bookings according to current time & timeZone

emrysal:
## What does this PR do?

Recurring booking fixes

image


zomars (approved):
I'm trusting your time wizardry

Ship it

PR #5065, opened 2022-10-18
Fix App Redirect logic to allow EventTypeApps requiring auth to correctly redirect

hariombalhara:


packages/app-store/wipemycalother/components/index.ts
```diff
@@ -1,2 +1 @@
-export { default as InstallAppButton } from "./InstallAppButton";
```
hariombalhara, line 1: @sean-brydon There still was InstallAppButton file for stripe app and thus isOmniInstall was being treated as false for it always. This was the reason, even with wrong logic stripe app was working correctly 🤦‍♂️

sean-brydon: lol I guess that makes sense - thanks for solving the head scratcher

sean-brydon (approved):
LGTM

PR #5068, opened 2022-10-18
Fix Zapier setup page

CarinaWolli:
## What does this PR do?

Fixes that Zapier setup page is always showing that Zapier is not installed. 

## Type of change

### DIFF
--- a/packages/app-store/zapier/pages/v2/setup/index.tsx
+++ b/packages/app-store/zapier/pages/v2/setup/index.tsx
@@ -18,7 +18,7 @@ export default function ZapierSetup(props: IZapierSetupProps) {
   const [newApiKey, setNewApiKey] = useState("");
   const { t } = useLocale();
   const utils = trpc.useContext();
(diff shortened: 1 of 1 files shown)


hariombalhara (approved):

packages/app-store/zapier/pages/v2/setup/index.tsx
```diff
@@ -18,7 +18,7 @@ export default function ZapierSetup(props: IZapierSetupProps) {
   const [newApiKey, setNewApiKey] = useState("");
   const { t } = useLocale();
   const utils = trpc.useContext();
-  const integrations = trpc.useQuery(["viewer.integrations", { variant: "other" }]);
+  const integrations = trpc.useQuery(["viewer.integrations", { variant: "automation" }]);
```
hariombalhara, line 21: There is useApp hook available now which accepts appId and tells if an app is installed or not plus gives it's data. I think that's a better way to access app data going forward.

PR #5085, opened 2022-10-18
Fix 404 error when editing default availability & create new availability from event type page

### LINKED ISSUE #4608: [CAL-122] Editing default availability from an event type causes a 404

### Issue Summary

When editing the availability of an event type with the default availability it navigates to a 404 page because the default availability does not contain an id. 																																																																															

### Steps to Reproduce

1. Go to an event type with the default availability.
2. Go to the availability tab and edit the availability
3. It should redirect to a 404 page

joeauyeung:
## What does this PR do?

This PR fixes the 404 error when editing the default availability from an event type page. When editing the default availability a new availability will also be created linked to that event type.

Fixes #4608

## Type of change

## How should this be tested?

- Go to an event type with the default availability
- Go to the availability tab and click "Edit availability"
- You should be navigated to a new availability with the default named {eventTypeName} availability
- Check the DB or navigate back to the event type page to check that the new availability is connected to that event type

## Checklist


packages/ui/v2/core/Shell.tsx
```diff
@@ -170,7 +170,7 @@ type LayoutProps = {
   TopNavContainer?: ReactNode;
   drawerState?: DrawerState;
   HeadingLeftIcon?: ReactNode;
-  backPath?: string; // renders back button to specified path
+  backPath?: string | boolean; // renders back button to specified path
```
joeauyeung, line 173: backPath will either navigate to the route if it is a string or just do router.back.

packages/ui/v2/core/Shell.tsx
```diff
@@ -773,7 +773,9 @@ export function ShellMain(props: LayoutProps) {
           <Button
             size="icon"
             color="minimal"
-            onClick={() => router.push(props.backPath as string)}
+            onClick={() =>
+              typeof props.backPath === "string" ? router.push(props.backPath as string) : router.back()
+            }
```
joeauyeung, line 778: If a new availability is created from an event type it should bring them back to that event type.

apps/web/components/v2/eventtype/AvailabilityTab.tsx
```diff
@@ -62,16 +66,54 @@ const format = (date: Date) =>
...
+      showToast(t("schedule_created_successfully", { scheduleName: schedule.name }), "success");
+    },
+    onError: (err) => {
+      if (err instanceof HttpError) {
+        const message = `${err.statusCode}: ${err.message}`;
+        showToast(message, "error");
+      }
+
+      if (err.data?.code === "UNAUTHORIZED") {
+        const message = `${err.data.code}: You are not able to create this event`;
```
JeroenReumkens, line 95: Did you mean "edit this event" in the error message, or should it say "create availability for this event"?

JeroenReumkens (approved):

JeroenReumkens:
Tested and works great @joeauyeung ! Approved. 

@Jaibles  I think it's not really clear if an event has selected the default availability. Currently that looks like this:
CleanShot 2022-10-19 at 09 48 33@2x

Especially the select being empty makes it confusing. Should we create a custom state when the default is selected, highlighting even more that the default is selected, and that you have 3 options; Keeping the default, selection a new one from the dropdown (so default should be put eg next to the dropdown I think?) or editing the availability (which will actually create a new availability if you currently have the default selected!, and will edit it if you've selected an availability from the dropdown)

Jaibles:
@JeroenReumkens yes this should never be empty unless you have no availability schedule created at all.

JeroenReumkens:
@Jaibles  
I'm all for adding the default label (new task I think). As for the editing of the availability: Makes total sense that people might want to edit their default instead of creating a new one. So.. Should we instead of the current change do the following: 

1. When default is selected: Show 2 buttons instead of 1; "Edit default availability" and "Create custom availability for this event" (content not final of course 😛 ) 
2. When custom is selected: Show 1 button "edit availability"

Jaibles:
@JeroenReumkens the custom availability in my screenshot is for a separate feature not yet built that would enable you to create a one off availability schedules for an event type.
CleanShot 2022-10-19 at 09 42 29@2x

joeauyeung:
@Jaibles @JeroenReumkens I like your idea with labeling that the selected availability is the default one and having the option to create a new one in the dropdown menu. The fear I had was if someone edits a default availability then other event types using that default availability will also be changed.

Jaibles:
> The fear I had was if someone edits a default availability then other event types using that default availability will also be changed.

This is my expectation. Edit one availability across all my event types.

PR #5088, opened 2022-10-19
Adds impersonation settings

### LINKED ISSUE #5061: [CAL-175] Impersonation settings need to implemented in V2
During the V2 migration we seem to have lost impersonation settings

They should live in security - potentially a new tab. 

The impersonation on a team level is already implemented 

sean-brydon:
Adds impersonation settings to V2 settings as it was missed in the transition 
Before:
N/A

After: (Fixed title but not in this SC :D)
CleanShot 2022-10-19 at 09 38 17

Fixes: #5061

### DIFF
--- a/apps/web/pages/settings/security/impersonation.tsx
+++ b/apps/web/pages/settings/security/impersonation.tsx
@@ -33,7 +33,7 @@ const ProfileImpersonationView = () => {
 
   return (
     <>
-      <Meta title="Two-Factor Authentication" description="Manage settings for your account passwords" />
+      <Meta title="Impersonation Settings" description="" />
       <Form
         form={formMethods}
         handleSubmit={({ disableImpersonation }) => {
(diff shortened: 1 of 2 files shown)


apps/web/pages/settings/security/impersonation.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Form
+        form={formMethods}
+        handleSubmit={({ disableImpersonation }) => {
+          mutation.mutate({ disableImpersonation });
+          utils.invalidateQueries(["viewer.me"]);
+        }}>
+        <div className="flex space-x-3">
+          <Switch
+            {...formMethods.register("disableImpersonation")}
+            defaultChecked={!user?.disableImpersonation}
```
sean-brydon, line 46: Flip logic here as true means its disabled -> show switch as toggled when disabled===false

hariombalhara (approved):
looks good codewise

JeroenReumkens (approved):

pumfleet (approved):

PR #5102, opened 2022-10-19
Added improved designs for OG images.

JeroenReumkens:
## What does this PR do?

Adds updated designs for Apps and Meeting OG Images.

### Why draft? 
Because it's not 100% according to design yet, and Peer needs to decide whether he wants to give this a go or let me tweak this still a bit.

### Screenshots

image

image

image

image

image

## Type of change

## How should this be tested?

- [ ] Go to meeting and app pages and check the images in the `og:image` meta tag and ensure they look nice <3


packages/app-store/riverside/config.json
```diff
@@ -3,8 +3,8 @@
   "name": "Riverside",
   "slug": "riverside",
   "type": "riverside_video",
-  "imageSrc": "/api/app-store/riverside/icon.svg",
-  "logo": "/api/app-store/riverside/icon.svg",
+  "imageSrc": "/api/app-store/riverside/icon-dark.svg",
```
JeroenReumkens, line 6: We've decided to add "dark" to a few icons which are mainly of a dark color. In the render of the OG images we check this name and don't render a gradient blob in the background if it's a dark image.

packages/lib/OgImages.tsx
```diff
@@ -127,46 +127,77 @@ export const Meeting = ({ title, users = [], profile }: MeetingImageProps) => {
...
         </div>
       </div>
     </Wrapper>
   );
 };
 
+const VisualBlur = ({ visualSlug }: { visualSlug: string }) => {
+  // Making a blur of a dark logo is very ugly. We use the filename to indicate,
+  // when we don't want to render these blurry blob backgrounds.
+  if (visualSlug.indexOf("dark") > -1) return null;
```
JeroenReumkens, line 149: This checks if it's a dark image and doesn't render a gradient blob.

PeerRich (approved):
this is awesome, great fixes!

only NIT: the "more" circle is not the same size and the "+2" is not horizontally, vertically centered
CleanShot 2022-10-19 at 17 00 00@2x

JeroenReumkens:
@PeerRich  Thanks! 🙏  you want to have the 'more' circle the same size as the avatars you mean?

JeroenReumkens:
@PeerRich  updated the size of the circle. Also offset the +2 a little bit to the top and left — it was actually centered before, so that was an optical illusion. But I think it now visually looks like it's centered better, while it actually isn't anymore, haha. 

image

zomars (approved):
LGTM

PR #5120, opened 2022-10-19
Feature/invalid credentials

alannnc:
## What does this PR do?

Adds a way to invalidate credentials, so we can now let the user now when an integration has broken.

- Add migration that creates new field on Credential table, invalid.
- Display alert error message on UI when credentials has been flag as invalid.
- When certain authentication response is received for zoom we flag that credential as invalid.
- Fix a lot of types from Prisma with the correct interface.

image

## Type of change

## How should this be tested?

- [ ] Install any app and from the DB update its invalid field to true.
- [ ] Now in the installed app page we should receive a alert error message, so our user can manually reinstall again.

## Checklist


packages/app-store/_utils/getCalendar.ts
```diff
@@ -1,13 +1,12 @@
...
 import logger from "@calcom/lib/logger";
 import type { Calendar } from "@calcom/types/Calendar";
+import { CredentialPayload } from "@calcom/types/Credential";
 
 import appStore from "..";
 
 const log = logger.getChildLogger({ prefix: ["CalendarManager"] });
 
-export const getCalendar = (credential: Credential | null): Calendar | null => {
+export const getCalendar = (credential: CredentialPayload | null): Calendar | null => {
```
alannnc, line 9: Most of the changes includes changing the type that was being used for Credential.

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -86,8 +86,13 @@ const zoomAuth = (credential: Credential) => {
...
     });
 
-    const responseBody = await handleZoomResponse(response);
+    const responseBody = await handleZoomResponse(response, credential.id);
 
+    if (responseBody.error) {
+      if (responseBody.error === "invalid_grant") {
+        return Promise.reject(new Error("Invalid grant for Cal.com zoom app"));
+      }
+    }
```
alannnc, line 95: If invalid_grant is received we should invalidate the credential right away.

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -280,42 +292,73 @@ const ZoomVideoApiAdapter = (credential: Credential): VideoApiAdapter => {
...
+    if ((response && response.status === 124) || responseBody.error === "invalid_grant") {
+      await invalidateCredential(credentialId);
+    }
     throw Error(response.statusText);
   }
 
   return response.json();
 };
 
+const invalidateCredential = async (credentialId: Credential["id"]) => {
```
alannnc, line 345: Our function that invalidate certain credential

packages/types/Credential.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  select: {
+    id: true;
+    appId: true;
+    type: true;
+    userId: true;
+    key: true;
+  };
+}>;
+
+export type CredentialWithAppName = CredentialPayload & { appName: string };
```
alannnc, line 18: A help type definition to use when requiring Credential Type from DB.

leog (approved):
LGTM, great job, very useful for the users!

I was just wondering, is this a retroactive change for existing credentials? If not, would it be worth in executing some script? Also, perhaps it would be worth exploring running such a script with an email notification added, so users can be aware their apps will not work as expected, before anyone books a new event and they don't get the desired behavior out of their installed apps.

On another note, as a user I would prefer the app facilitates an action to let me delete and reinstall all in one, instead of providing a delete button to then having to go search the app again and reinstall it. Within this context, we know the app already, we could just add a delete & reinstall alongside the delete button, WDYT @Jaibles?

packages/prisma/schema.prisma
```diff
@@ -100,6 +100,7 @@ model Credential {
   // How to make it a required column?
   appId                String?
   destinationCalendars DestinationCalendar[]
+  invalid              Boolean?
```
emrysal, line 103: Can we delete the credential if invalid?

alannnc: Yes! Let me elaborate a little more on this: This flag should help us on the backend server when we find out that some credential it's no longer gonna work we can flag it. We can add triggers to this and let know the user to make an action on this.

alannnc:
> I was just wondering, is this a retroactive change for existing credentials? If not, would it be worth in executing some script? Also, perhaps it would be worth exploring running such a script with an email notification added, so users can be aware their apps will not work as expected, before anyone books a new event and they don't get the desired behaviour out of their installed apps.

About retroactive this should work with existing credential, we just missing a way to invalidate them. (Zoom its the only one right now).
Also for notifications and such, we should add:
- Notification via backend that something broke.


> On another note, as a user I would prefer the app facilitates an action to let me delete and reinstall all in one, instead of providing a delete button to then having to go search the app again and reinstall it.
This should be reinstall button you're right there we can do that of course, but probably next iteration.

PR #5141, opened 2022-10-21
Hotfix: Unconfirmed bookings marked as confirmed in DB

### LINKED ISSUE #5142: [CAL-217] Bookings that require confirmation appear confirmed automatically on platform

hariombalhara:
Fixes #5142
- It was happening only if the booker isn't logged into his Cal account or doesn't have a Cal account.

I can see other inconsistencies here
- Webhook and Confirmation Email doesn't rely on the same variable that decides if booking should be marked Accepted. this can create a situation where booking is accepted in DB but webhook doesn't say so. Similarly, Email would go for an Accepted booking. This would arise only if the owner of the event is rescheduling the booking.

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -245,7 +245,7 @@ async function ensureAvailableUsers(
   return availableUsers;
 }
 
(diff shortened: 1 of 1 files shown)


alishaz-polymath (approved):
Looks good. Deployment is failing though, probably unrelated, but can you check please. 🙏

PR #5147, opened 2022-10-21
#5131 -Event Type Settings Inconsistency 

### LINKED ISSUE #5131: [CAL-210] Inconsistency between how toggle on settings appear in event type edit
When you're editing an event type and toggle on any settings, the behaviour for how and where they appear is very inconsistent.

Some animate in and some just appear, some are disabled and then become enabled. Then in terms of position some are correctly aligned with the text above, and some are aligned to the far left. Then some are half way between.

The correct appearance is in Figma

[   ↳  Event Type Single ✓ - ❖ Cal DS (Figma)](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=1413%3A65854)

sean-brydon:
Fixes the settings inconsistency throughout the Event Type pages

https://www.loom.com/share/0f436a1c729242f984b1328ebd22c320


packages/ui/v2/core/SettingsToggle.tsx
```diff
@@ -0,0 +1,57 @@
...
+              disabled={disabled}
+            />
+
+            <div className="">
+              <Label className="text-sm font-semibold leading-none text-black">{title}</Label>
+              {description && <p className="-mt-2 text-sm leading-normal text-gray-600">{description}</p>}
+            </div>
+          </div>
+          {children && (
+            <div className="mt-4 lg:ml-14" ref={animateRef}>
```
sean-brydon, line 47: We render an empty div with the animation ref so we can slowly fade in the content if `!checked`

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -159,149 +159,107 @@ export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">)
...
-                                      endDate,
-                                    });
-                                  }}
-                                />
-                              )}
+      <Controller
+        name="periodType"
+        control={formMethods.control}
+        render={({ field: { value } }) => (
+          <SettingsToggle
```
sean-brydon, line 190: Moves bulk of code into the component - children are passed through to be rendered on `true`

packages/ui/package.json
```diff
@@ -23,6 +23,7 @@
     "react": "^18.2.0",
     "react-colorful": "^5.6.0",
     "react-feather": "^2.0.10",
+    "@formkit/auto-animate": "^1.0.0-beta.1",
```
sean-brydon, line 26: Why was this not in UI already :D

Jaibles:
Looks like a great solution to me 👏

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -159,149 +159,106 @@ export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">)
...
-          </div>
-          <div className="mt-4 lg:ml-14">
+      <Controller
+        name="bookingLimits"
+        control={formMethods.control}
+        render={({ field: { value } }) => (
+          <SettingsToggle
+            title={t("limit_booking_frequency")}
+            description={t("limit_booking_frequency_description")}
+            checked={Object.keys(value ?? {}).length > 0}
```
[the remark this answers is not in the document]
sean-brydon: I'm not sure here tbs - `checked` is also shared with out switch component so if we were to change it one place I'd deffo saw we should change it there too

packages/ui/v2/core/SettingsToggle.tsx
```diff
@@ -0,0 +1,57 @@
+import { useAutoAnimate } from "@formkit/auto-animate/react";
```
emrysal, line 1: Does this have to be in v2?

sean-brydon: No it doesn't - but I believe it should live here until we fully migrate everything from V2 -> v1 files so we don't accidentally remove anything we are meant to in the clean up

zomars (approved):
LGTM

PR #5155, opened 2022-10-21
[CAL-103] All Bookings show up under Today and Today should be uppercase

### LINKED ISSUE #4111: [CAL-103] All Bookings show up under Today and Today should be uppercase
Task should probably just be to remove "Today"?

If the quickest thing to do here is not group things that's a perfectly ok fix. Grouping can be a future task as it likely requires complex logic.

Screenshot 2022-09-01 at 16 56 27

gitstart:
## What does this PR do?

Now "Today "are showing only today bookings

Fixes #4111

https://www.loom.com/share/c050ec5c5c224adc943486a0096fcaa4

## Type of change


gitstart:
Hello @PeerRich! Please review the PR

leog:
Hey @gitstart. Thanks a lot for your contribution! It really is useful when the user has a long list of bookings.

I have a couple of points to highlight. I found an issue when testing locally. Seems like when more bookings load as part of the infinite scrolling, it creates a new "ALL" group each infinite scroll comes with results. See:
<kbd></kbd>

Also, I would like to summon @Jaibles to check the position and margins on the design for this feature, also taking the liberty to suggest having a sticky header for the current scrolled group so the user can know to which one the scrolled bookings belong to, like: 
<kbd></kbd>

gitstart:
Hey @leog we are working on the scroll issue, however not sure if we want the sticky header. Can you please give confirmation on the design @Jaibles? Thank you

PeerRich:
also apps/web/pages/bookings/[status].tsx seems to have a merge conflict

gitstart:
Sorry for the delay, PR is ready for review now cc: @leog @Jaibles @PeerRich

zomars:
Does it make sense to have a "Today" section for any other that "upcoming" bookings? Also have the pagination issue been resolved?

Jaibles:
> Does it make sense to have a "Today" section for any other that "upcoming" bookings?

@zomars Not too sure I understand the question vs the current behaviour, but in the designs I have a Today section for anything happening today. An "Unconfirmed" booking that's requested for "Today" would also appear there.

CleanShot 2022-11-15 at 07 05 28@2x

zomars:
> > Does it make sense to have a "Today" section for any other that "upcoming" bookings?
> 
> @zomars Not too sure I understand the question vs the current behaviour, but in the designs I have a Today section for anything happening today. An "Unconfirmed" booking that's requested for "Today" would also appear there.
> 
> CleanShot 2022-11-15 at 07 05 28@2x

Agreed. But this means that there shouldn't be any "Today" section in past bookings right?

zomars (changes requested):
Tests are failing due a duplicate "Wipe today" button. Blocking while I come up with a fix.

image

zomars (approved):
Fixed test and approving for merge. Thank you for your contribution 🙏

zomars (changes requested):
Fixes some tests but there's still some remaining broken tests. Blocking until those are fixed as well.

Jaibles:
> Agreed. But this means that there shouldn't be any "Today" section in past bookings right?

Ah yes for sure.

gitstart:
> > Agreed. But this means that there shouldn't be any "Today" section in past bookings right?
> 
> Ah yes for sure.

So this means, the "Today" section should only be in the "Upcoming" tab and it should display the `Upcoming`, `Unconfirmed` and `Recurring` meetings for today. And `Cancelled` and `Past` meetings should not be included.

gitstart:
Hey @zomars PR updated with required changes, please take a look

zomars (approved):

PR #5170, opened 2022-10-22
Segment Control

### LINKED ISSUE #5151: [CAL-220] Build Segment component
[Segments - ❖ Cal DS (Figma)](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=15212%3A117411)

Also look for opportunities to introduce this in the app. The rule will likely be that it should be used to always replace dropdowns with 2 options, and 3 options if the labels are short (ie 1 or 2 words). To be refined with testing & evaluation.

sean-brydon:
## What does this PR do?

Fixes: #5151 

Added example to `sandbox` I know this isn't really used but radix tailwind doesn't like storybook as it doesn't generate the styles. 

@Jaibles Visit localhost:3000/sandbox/form or use the preview URL https://cal-git-segmentcontroll-cal.vercel.app/sandbox/form for the same thing :) 

CleanShot 2022-10-22 at 13 50 21@2x


PeerRich (approved):
looks great

JeroenReumkens:
Oh that's interesting. I just stumbled on this issue, and I'm building a very similar toggle for the new am/pm toggle: 

CleanShot 2022-10-24 at 13 24 54@2x


However, in my case semantically it shouldn't be a radio button but actual buttons, so I opted for the toggle group. Although in other place in the design a radio button definitely is the right choice. So I think that in the future we should see if we could somehow share the styles, but give the dev the option to use it either as a radio button or as a native button toggle. What do you think?

My WIP commit: 
https://github.com/calcom/cal.com/commit/1ccecbc90ae61c76fffcd06d7dd3ca32951a73a3

sean-brydon:
> Oh that's interesting. I just stumbled on this issue, and I'm building a very similar toggle for the new am/pm toggle:
> 
> CleanShot 2022-10-24 at 13 24 54@2x
> 
> However, in my case semantically it shouldn't be a radio button but actual buttons, so I opted for the toggle group. Although in other place in the design a radio button definitely is the right choice. So I think that in the future we should see if we could somehow share the styles, but give the dev the option to use it either as a radio button or as a native button toggle. What do you think?
> 
> My WIP commit: [1ccecbc](https://github.com/calcom/cal.com/commit/1ccecbc90ae61c76fffcd06d7dd3ca32951a73a3)

Although I agree with what you are saying - I believe the functionality of both is different. One is a form input and the other isn't ? that just doesn't sit right with me in some ways

JeroenReumkens:
> Although I agree with what you are saying - I believe the functionality of both is different. One is a form input and the other isn't ? that just doesn't sit right with me in some ways

Yes agree. It is visually similar, but the implementation is different indeed. Not sure if we can merge it somehow, at least styling wise. But it should still be able to be a button or a radio.  For now lets keep it in the back of our minds and see if we need to merge it one day. I did for example add an animation when switching the toggle, would be nice if those things stayed consistent.

Jaibles:
@sean-brydon looks perfect

(@JeroenReumkens I'd actually set Seán this task after designing that am/pm task thinking it could be reused so good to know to types are needed)

pumfleet (approved):

PR #5171, opened 2022-10-23
Bringing back recurring dates info tooltip

leog:
## What does this PR do?
There was a TODO pending regarding bringing back the recurring dates info tooltip as it was removed due to the usage of a heavy-weight dependency. This implementation uses `dayjs-recur` plugin which is light-weight.

Taking advantage of this implementation, now the tooltip shows date & times taking into consideration the user's preferences for timezone and format. Also, I fixed a small bug regarding showing the correct number for "remaining events" when one recurring event occurrence is in the past, shown in the following screenshot.

<kbd>image</kbd>

 - [ ] Removing RRule.js completely replacing it with manual calculations using `dayjs-recur` if needed

## Type of change

## How should this be tested?

Take a look at the recurring tabs in bookings, the info tooltip is now there again showing dates with correct timezone and time format. Using the added "Seeded Yoga Class" bookings you can see the bug fixed for a past occurrence.


leog:
Closing in favor of #5172.

PR #5172, opened 2022-10-24
Bugfix/recurring dst change

emrysal:
This PR fixes the DST issues regarding recurring events.

Also, it fixes a couple of issues:

  1.  "events remaining" on recurring events now reflect past events within the series
<kbd></kbd>

  2. Booking page was not applying user's preferences to display time
<kbd></kbd>
<kbd></kbd>


leog:
Hey @emrysal. I was actually working on replacing RRule.js with a dayjs plugin here: #5171. Maybe you could apply your changes to support DST in that PR to solve that issue and also we can remove the heavyweight dependency at the same time. WDYT?

emrysal:
@leog I had a look at dayjs-recur but I would say (looking at both repo's) that rrule is better. Recur is a dead project (last activity was in March) - and has a total of 8 stars. It also only supports a very limited subset of features compared to rrule - and also very importantly rrule is a port of the python project for recurring rules which is good for debugging (you can use python docs to learn about rrule.js features).

Rrule also has some bugs that need fixing but the performance issues we've been seeing can be resolved on our end. It's a much more stable project.

leog (approved):
Works great. Added 2 more relevant fixes brought from the closed PR #5171.

zomars (approved):
Ship it

PR #5175, opened 2022-10-24
Feature/improved timeformat selection booking page

### LINKED ISSUE #4340: [CAL-104] /public-booking - 12/24h time format: Can we check users locale settings (Intl JS API)
Image
- As an American (quote) I basically never use 24h time so it feels really unfamiliar/unfriendly.
- Similarly, the date format "7 September" in the top right feels unfamiliar. I suggest looking at the user's locale settings (Intl JS API)
- Okay now just for fun I clicked to expand the time zone dropdown and I just noticed that there is a way to toggle am/pm — but (a) again, it should do ampm by default for en-us locales, (b) it's in a place I'd never normally look since the default timezone was correct for me already.
- The ampm option wasn't remembered/respected on the following screen (but the date format does look better here):

At the very least we can do "September 7" here instead of "7 September" which would be a slight improvement.

JeroenReumkens:
## What does this PR do?

Improves time format behavior. I changed the following: 
* Time format changes in profile won't affect the public booking page anymore (reverted that change)
* Time format toggle isn't hidden in timezone dropdown anymore, rather directly above time slots
* Toggling time format will save it in localstorage and use that setting from that point on
* Added hint in profile settings telling user that changing this won't affect booking page for you or your attendees.

### Loom:

https://loom.com/share/5b66c0d2a42d40f58ea285da29e4379e

Closes #4340

## Type of change

## How should this be tested?

- [ ] Go to a booking page and check the time slots.  If you don't have the localstorage item 'is24clock', we will check your browser locale and based on that try to guess 12/24h time format. After that we store this in localstorage. If you open a specific date you should now see the toggle above the time slots. Toggling this should remember the setting in localstorage on refresh. 
- [ ] Verify that the hint explaining that changing timeformat in your profile doesn't affect booking page, is present on settings page. 
- [ ] I'm not a content editor, we need to come up with a better text for the helper text 😅


apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -48,14 +52,25 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
...
+        <div className="mr-4">
+          <span className="text-bookingdarker dark:text-darkgray-800 font-semibold text-gray-900">
+            {nameOfDay(i18n.language, Number(date.format("d")), "short")}
+          </span>
+          <span className="text-bookinglight font-medium">
+            , {date.toDate().toLocaleString(i18n.language, { month: "long", day: "numeric" })}
+          </span>
+        </div>
+        <div className="ml-auto">
+          <ToggleGroup
```
JeroenReumkens, line 65: The timeformat toggle has been added here, above the timeslots on the booking page.

apps/web/components/booking/TimeOptions.tsx
```diff
@@ -1,25 +1,19 @@
 import { FC, useEffect, useState } from "react";
 
 import { useLocale } from "@calcom/lib/hooks/useLocale";
-import { Switch } from "@calcom/ui/v2";
 import TimezoneSelect, { ITimezoneOption } from "@calcom/ui/v2/core/TimezoneSelect";
```
JeroenReumkens, line 4: Removed the timeformat from the timezone dropdown in this file.

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -213,35 +215,20 @@ const SlotPicker = ({
 };
 
 function TimezoneDropdown({
-  onChangeTimeFormat,
```
JeroenReumkens, line 216: also here, timezonedropdown doesn't need to know about timeformat anymore.

packages/lib/timeFormat.ts
```diff
@@ -8,6 +8,11 @@ import { localStorage } from "@calcom/lib/webstorage";
 
 const is24hLocalstorageKey = "timeOption.is24hClock";
 
+export enum TimeFormat {
```
JeroenReumkens, line 11: I found `h:mma` etc very confusing throughout the code, so I converted in to an enum for way better readability.

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -317,7 +292,12 @@ const AvailabilityPage = ({ profile, eventType, ...restProps }: Props) => {
   const isBackgroundTransparent = useIsBackgroundTransparent();
 
   const [timeZone, setTimeZone] = useState<string>();
-  const [timeFormat, setTimeFormat] = useState<string>("HH:mm");
+  const [timeFormat, setTimeFormat] = useState<TimeFormat>(detectBrowserTimeFormat);
```
JeroenReumkens, line 295: I found `h:mma` etc very confusing throughout the code, so I converted in to an enum for way better readability.

In this file we were also using 3 notations: string (`h:mma` etc), numbers (12/24) and `is24Hours` boolean. I got rid of the number variant to have less variants and confusion. Toggle uses the boolean now, all the rest uses string format (with new enum).

apps/web/public/static/locales/en/common.json
```diff
@@ -1110,6 +1112,7 @@
   "error_updating_settings": "Error updating settings",
   "personal_cal_url": "My personal Cal URL",
   "bio_hint": "A few sentences about yourself. this will appear on your personal url page.",
+  "timeformat_profile_hint": "Changes to this timezone will not affect your public booking pages, nor will it effect the way attendees see your booking page.",
```
JeroenReumkens, line 1115: This content can probably be improved! @PeerRich  @Jaibles Any suggestions?

Jaibles: What about: 
"This is an internal setting and will not affect how times are displayed on public booking pages.

zomars, line 1115:
```suggestion
  "timeformat_profile_hint": "This is an internal setting and will not affect how times are displayed on public booking pages for you or anyone booking you.",
```

zomars (approved):
LGTM

Issue #4205, opened 2022-09-06
Rescheduling event created on CalDav calendar throws an error/unable to reschedule and get a new confirmation


milospuac:
## Caldav - [Infomaniak](https://www.infomaniak.com/en/hosting/service-mail/)

Bug: Rescheduling creates a new meeting at the scheduled date and removes the old one. BUT: It doesn't send any confirmation email when someone reschedules a meeting with me, nor does it updates calendars synced with WebDAV. The meetings are visible on the app.cal.com Bookings tab. People also told me it doesn't work and it then removes the slot (which is sensible because they are in fact, booked) so they cannot even reschedule to the time they want.

Steps:

- Book a meeting (on someone who has a WebDAV/CalDAV calendar)

- Try rescheduling it with the link at the bottom of the confirmation mail you got

## CalDav - Fastmail 

The booking in question created an event on the Fastmail calendar. Rescheduling the same meeting using the link on the confirmation page gave me this error.

<img width="764" alt="Screenshot 2022-09-06 at 14 06 26" src="https://user-images.githubusercontent.com/97884287/188636566-b53363a5-7ab2-4f24-aeb4-4fc17202fbbc.png">

alishaz-polymath:
Fastmail is fixed and Infomaniac is not currently fully supported due to lack of requests. If anyone wishes for us to consider developing support for Infomaniac, please let us know below and upvote it 🙏🙂

PR #5193, opened 2022-10-25
Hotfix/caldav reschedule jsonerror

### LINKED ISSUE #5221: [CAL-254] Nextcloud CalDAV reschedule fails
Nextcloud CalDAV rescheduling fails (reschedules in Cal but not in the connected calendar) and cancellation doesn't remove the event from the calendar

alishaz-polymath:
## What does this PR do?

- Fixes json end of line error when rescheduling (can be easily recreated using a CalDAV service that doesn't return any content when using update)
- Fixes reschedule for NextCloud (and potentially other services which function on similar iCal string standard
- Fixes reschedule for Fastmail

Fixes #5221 #4205

[Loom](https://www.loom.com/share/a62fd169f66a4fb49db3bb2846e4ebd7)

## Type of change


packages/lib/CalendarService.ts
```diff
@@ -189,21 +189,46 @@ export default abstract class BaseCalendarService implements Calendar {
       }
 
       const eventsToUpdate = events.filter((e) => e.uid === uid);
+      let ev = eventsToUpdate[0];
```
alannnc, line 192:
```suggestion
      let [event] = eventsToUpdate;
```
NIT: We can try to be more explicit with variable naming.

packages/lib/CalendarService.ts
```diff
@@ -189,21 +189,46 @@ export default abstract class BaseCalendarService implements Calendar {
...
       return Promise.all(
         eventsToUpdate.map((e) => {
+          ev = e;
```
alannnc, line 195:
```suggestion
        eventsToUpdate.map(({eventItem) => {
          event = eventItem;
```
Here we change e for -> eventItem as some people can use e as error sometimes. With this change we add more readability.

packages/lib/CalendarService.ts
```diff
@@ -189,21 +189,46 @@ export default abstract class BaseCalendarService implements Calendar {
...
               url: e.url,
-              data: iCalString,
+              data: iCalString?.replace(/METHOD:[^\r\n]+\r\n/g, ""),
               etag: e?.etag,
             },
             headers: this.headers,
           });
         })
-      ).then((p) => p.map((r) => r.json() as unknown as NewCalendarEventType));
+      ).then((parameter) =>
```
alannnc, line 205: we receive a parameter but what kind of parameter we receive?

alannnc (dismissed):
Whops it's a draft. Left some nit comments.

alishaz-polymath:
> Whops it's a draft. Left some nit comments.

Haha, that's okay. The points were valid nonetheless, however, it was far from finished 😉

packages/lib/CalendarService.ts
```diff
@@ -190,30 +190,56 @@ export default abstract class BaseCalendarService implements Calendar {
 
       const eventsToUpdate = events.filter((e) => e.uid === uid);
       return Promise.all(
-        eventsToUpdate.map((e) => {
+        eventsToUpdate.map((eventItem) => {
+          event = eventItem;
```
alishaz-polymath, line 194: Update the value of `event` to persist the properties and use them in `.then`

packages/lib/CalendarService.ts
```diff
@@ -190,30 +190,56 @@ export default abstract class BaseCalendarService implements Calendar {
...
           return updateCalendarObject({
             calendarObject: {
-              url: e.url,
-              data: iCalString,
-              etag: e?.etag,
+              url: event.url,
+              data: iCalString?.replace(/METHOD:[^\r\n]+\r\n/g, ""),
```
alishaz-polymath, line 198: Adds the adjustment to comply with iCal2.0 standard which various service providers like nextcloud and fastmail expect

packages/lib/CalendarService.ts
```diff
@@ -190,20 +190,44 @@ export default abstract class BaseCalendarService implements Calendar {
...
+              etag: event?.etag,
             },
             headers: this.headers,
           });
         })
-      ).then((p) => p.map((r) => r.json() as unknown as NewCalendarEventType));
+      ).then((responses) =>
+        responses.map((response) => {
+          if (response.status > 199 && response.status < 207) {
```
alishaz-polymath, line 206: Confirms if the update call went through successfully. Some providers return updated data while the others choose not to (sending 204 instead)

packages/lib/CalendarService.ts
```diff
@@ -190,20 +190,44 @@ export default abstract class BaseCalendarService implements Calendar {
...
+            const ret = {
+              uid,
+              type: this.credentials.type,
+              id: typeof event.uid === "string" ? event.uid : "-1",
+              password: "",
+              url: event.url,
+              additionalInfo: event.additionalInformation,
+            } as NewCalendarEventType;
```
alishaz-polymath, line 214: Since some providers return updated data while the others choose not to (sending 204 instead), we can't really use the return data directly and form a `NewCalendarEventType` object from it, as at times it will be null, thus causing JSON errors. Instead, we build it by utilizing the data we already have direct access to.

packages/lib/CalendarService.ts
```diff
@@ -275,7 +298,7 @@ export default abstract class BaseCalendarService implements Calendar {
     const events: { start: string; end: string }[] = [];
 
     objects.forEach((object) => {
-      if (object.data == null) return;
+      if (object.data == null || JSON.stringify(object.data) == "{}") return;
```
alishaz-polymath, line 301: when the object doesn't contain any data, we return. There will be a follow up on this condition to consider situations when the service providers **do not** send data directly, but that's a solution to a different problem and outside of the scope of this PR.

alishaz-polymath (commented):
Self Review Added

alishaz-polymath:
Loom video also added to the description

packages/lib/CalendarService.ts
```diff
@@ -187,23 +187,48 @@ export default abstract class BaseCalendarService implements Calendar {
...
+              data: iCalString?.replace(/METHOD:[^\r\n]+\r\n/g, ""),
+              etag: calendarEvent?.etag,
             },
             headers: this.headers,
           });
         })
-      ).then((p) => p.map((r) => r.json() as unknown as NewCalendarEventType));
+      ).then((responses) =>
+        responses.map((response) => {
+          if (response.status > 199 && response.status < 207) {
```
emrysal, line 206:
```suggestion
          if (response.status >= 200 && response.status < 300) {
```

Everything between 200 and 299 should be considered success.

emrysal (approved):
Minor feedback, overall approved.

PR #5198, opened 2022-10-25
Fix: avoid creating multiple schedules on getting-started/setup-availability

leplay:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


packages/trpc/server/routers/viewer/availability.tsx
```diff
@@ -131,11 +131,13 @@ export const availabilityRouter = createProtectedRouter()
           })),
         },
       };
-
+      
+      const hasDefaultScheduleId = await hasDefaultSchedule(user, prisma);
```
hariombalhara, line 135: I might be confused but isn't the code behaving the same way unless we check for `hasDefaultScheduleId` to prevent `prisma.schedule.create`

hariombalhara (changes requested):

packages/trpc/server/routers/viewer/availability.tsx
leplay: The logic is right. But in `hasDefaultSchedule` function, it will also check if `schedule` db has a record.
```js
const hasDefaultSchedule = async (user: Partial<User>, prisma: PrismaClient) => {
  const defaultSchedule = await prisma.schedule.findFirst({
    where: {
      userId: user.id,
    },
  });
  return !!user.defaultScheduleId || !!defaultSchedule;
};
```
If `prisma.schedule.create` run before this check, user will never get a `defaultScheduleId`.

Or I can just skip the schedule db check in `hasDefaultSchedule` function?

hariombalhara: cc @emrysal

PR #5203, opened 2022-10-25
Add organizer's email address as reply-to for workflow emails

### LINKED ISSUE #5092: [CAL-191] Allows adding "reply-to" header in email workflows

It would be great if people could reply to emails sent using the workflow. Their emails should not go to [hello@cal.com](mailto:hello@cal.com) instead these replys should go to my email address.

I would like to be able to set a "reply-to" header inside workflows. This is a very simple way to solve the problem.

CarinaWolli:
## What does this PR do?

Adds the organizer's email address as reply-to for workflow emails. 

Screenshot 2022-10-25 at 14 17 11

Fixes #5092


packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -154,6 +154,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
             html: emailContent.emailBody.html,
             batchId: batchId,
             sendAt: dayjs(reminder.scheduledDate).unix(),
+            replyTo: reminder.booking?.user?.email || senderEmail,
```
hariombalhara, line 157: I think if we reminder.booking?.user?.email isn't there, that is actually an error because we couldn't add the correct  `replyTo`.

So, along with falling back to senderEmail, we should also log this as an error for observation

hariombalhara (approved):
Looks good codewise. 

Tested that on event booking, evt.organizer.email has correct detail.

PR #5204, opened 2022-10-25
Cleaned up v1 for app store pages.

JeroenReumkens:
## What does this PR do?

Cleans up old v1 code for app-store package and its implementation in web module.

## Type of change

## How should this be tested?

- [ ] Ensure that app overview and detail pages don't show any errors
- [ ] Ensure adding an app with a setup page (exchange, caldavcalendar, apple calendar) still shows this setup page


packages/app-store/applecalendar/pages/setup/index.tsx
```diff
@@ -4,9 +4,7 @@ import { useForm } from "react-hook-form";
 import { Toaster } from "react-hot-toast";
 
 import { useLocale } from "@calcom/lib/hooks/useLocale";
```
JeroenReumkens, line 6: The v2 pages above say they're deleted, but they're actually moved to the location of v1 (so without the v2 prefix), the diffs you see in those original v1 pages are actually the changes that were applied in v2.

hariombalhara (approved):

PR #5208, opened 2022-10-25
Duplicate meeting within non-traditional calendars

leog:
## What does this PR do?

When user does not have any calendar but there is a non-traditional calendar app installed such as Close.com or HubSpot, the bookings were going through those apps twice.

<kbd>image</kbd>

## Type of change

## How should this be tested?

Have no calendar installed and install a CRM. Book a meeting and see the meeting created just once in the CRM while also seeing the confirmation email with just a single app status for it.


packages/core/EventManager.ts
```diff
@@ -329,11 +329,10 @@ export default class EventManager {
        *  Not ideal but, if we don't find a destination calendar,
        * fallback to the first connected calendar
        */
-      const [credential] = this.calendarCredentials;
-      if (!credential) {
-        return [];
+      const [credential] = this.calendarCredentials.filter((cred) => cred.type === "calendar");
```
JeroenReumkens, line 332: Hmm, actually it feels weird that the array is called `CALENDARcredentials`, but still potentially could include a credential that is not of type `calendar`. Should we explain this somewhere in a comment or rename the property?

packages/core/EventManager.ts
```diff
@@ -329,11 +329,10 @@ export default class EventManager {
...
+      if (credential) {
+        createdEvents.push(await createEvent(credential, event));
```
JeroenReumkens, line 334: Had to look up if the early returned wouldn't be necessary, but looking at the default on line 309 it seems not. Really smart change, a lot less confusing 🥇

JeroenReumkens (approved):

packages/core/EventManager.ts
leog: well, they are all calendar credentials, for apps that manage events, the thing is that we actually treat calendar apps per se differently from apps that are not calendars but manage events either way, like CRMs

I guess we can explain it better in the code, yeah

packages/core/EventManager.ts
```diff
@@ -79,6 +79,9 @@ export default class EventManager {
     const appCredentials = getApps(user.credentials).flatMap((app) =>
       app.credentials.map((creds) => ({ ...creds, appName: app.name }))
     );
+    // This includes all calendar-related apps, traditional calendars such as Google Calendar
+    // (type google_calendar) and non-traditional calendars such as CRMs like Close.com
+    // (type closecom_other_calendar)
```
leog, line 84: Here you go @JeroenReumkens, thanks for the suggestion 🙂

packages/core/EventManager.ts
alannnc: I would move the await outside of this .push() sentence.
```suggestion
        const createdEvent = await createEvent(credential, event);
        if (createdEvent) createdEvents.push(createdEvent);

```

alannnc: And probably add a try/catch for that await.

alannnc (changes requested):
Code looks fine, just some NIT changes. Couldn't replicate it tho.

alannnc (approved):
Couldn't replicate it but code now looks fine!

PR #5216, opened 2022-10-25
Fix/event location when fallback video

alannnc:
## What does this PR do?

Fix reference type when fallback to calVideo. This created a row in BookingReferences with cal video credentials but zoom_video as type. This was because it wasn't even considering that when fallback to cal video we should update references to create.

Fixes # (issue)

https://www.loom.com/share/df667a4e21fd4d8f9d016596874bbacd

## Type of change

## How should this be tested?

- [ ] Force an exception while doing a booking with zoom video
- [ ] Open email with cal.video link and it should be working now.

## Checklist

### DIFF
--- a/packages/core/EventManager.ts
+++ b/packages/core/EventManager.ts
@@ -98,9 +98,11 @@ export default class EventManager {
     // If and only if event type is a dedicated meeting, create a dedicated video meeting.
     if (isDedicated) {
       const result = await this.createVideoEvent(evt);
+
       if (result.createdEvent) {
         evt.videoCallData = result.createdEvent;
         evt.location = result.originalEvent.location;
+        result.type = result.createdEvent.type;
       }
 
       results.push(result);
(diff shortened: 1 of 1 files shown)


packages/core/EventManager.ts
```diff
@@ -98,8 +98,11 @@ export default class EventManager {
     // If and only if event type is a dedicated meeting, create a dedicated video meeting.
     if (isDedicated) {
       const result = await this.createVideoEvent(evt);
+
       if (result.createdEvent) {
         evt.videoCallData = result.createdEvent;
+        evt.location = result.originalEvent.location;
```
alannnc, line 104: originalEvent.location gets updated in case there is a fallback to cal video

packages/core/EventManager.ts
```diff
@@ -98,8 +98,11 @@ export default class EventManager {
...
+        result.type = result.createdEvent.type;
```
alannnc, line 105: This should only run on video references.

PeerRich (approved):

PR #5217, opened 2022-10-26
Minimum booking notice will allow hours and days

### LINKED ISSUE #3962: Minimum booking notice should also allow hours and days

The "Minimum booking notice" field can only be specified in minutes.  This means that if I want a minimum notice period of hours or days, I need to multiply by 60 and maybe also by 24 in my head to get the right number of minutes:

image

Whilst as an ex-mathematician I enjoy little opportunities to keep my mental arithmetic sharp, this does not seem like ideal UI/UX design!

Change the fixed `mins` text to a dropdown where the user can select between `mins`, `hours`, and `days`.  Then the app will automatically calculate the corresponding number of minutes behind the scenes.

om-ray:
## What does this PR do?

Now you can use minutes, hours, and calendar days as duration types for the minimum booking notice

Fixes #3962 

https://www.loom.com/share/56f3cb20a5974b11a940aac042655819

## Type of change


om-ray:
@aspiers @zomars @hariombalhara

PeerRich:
hey @om-ray awesome! 

do you have a loom or screenshots?

om-ray:
@PeerRich Ill record one and edit the message

PeerRich:
should we just call it "days" vs "calendar days"?

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -114,13 +114,99 @@ export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">)
...
+                {
+                  label: t("minutes"),
+                  value: "minute",
+                },
+                {
+                  label: t("hours"),
+                  value: "hour",
+                },
+                {
+                  label: t("calendar_days"),
```
PeerRich, line 167:
```suggestion
                  label: t("days"),
```

apps/web/lib/hooks/useSlots.ts
```diff
@@ -27,6 +27,7 @@ type UseSlotsProps = {
   eventLength: number;
   eventTypeId: number;
   minimumBookingNotice?: number;
+  minimumBookingNoticeType?: string | undefined;
```
PeerRich, line 30: instead of string, maybe 

"minute" | "hour" | "day" ?

packages/prisma/schema.prisma
```diff
@@ -30,60 +30,61 @@ enum PeriodType {
 }
 
 model EventType {
-  id                      Int                     @id @default(autoincrement())
+  id                       Int                     @id @default(autoincrement())
```
PeerRich, line 33: odd change here

PeerRich: this whole migration looks scary

packages/prisma/selects/event-types.ts
```diff
@@ -84,6 +84,7 @@ export const availiblityPageEventTypeSelect = Prisma.validator<Prisma.EventTypeS
   userId: true,
   slug: true,
   minimumBookingNotice: true,
+  minimumBookingNoticeType: true,
```
PeerRich, line 87: not quite sure why we need a boolean here

PeerRich:
@om-ray left a couple comments. waiting for more reviewers

packages/prisma/selects/event-types.ts
om-ray: This is to tell prisma that we want that value to go to the backend

apps/web/lib/hooks/useSlots.ts
om-ray: I kept it as string so that if we wanted to add another value such as `business day` we could add that

packages/prisma/schema.prisma
om-ray: I added the space so that the formatting for the file would be consistent

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -114,13 +114,99 @@ export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">)
...
+                    placeholder="120"
+                    className="mr-2 w-full"
+                    defaultValue={eventType.minimumBookingNotice}
+                    {...formMethods.register("minimumBookingNotice", { valueAsNumber: true })}
+                  />
+                  <Select
+                    isSearchable={false}
+                    className="mb-2 ml-2 w-1/3"
+                    defaultValue={durationTypeOptions.find(
+                      (option) => option.value === eventType.minimumBookingNoticeType
```
JeroenReumkens, line 194: If we only say we allow whole numbers in the duration field (eg whole minutes, whole hours or whole days), I think we could make the `noticeType` a client side only select. That would make this change a lot smaller since we don't need that additional field in the DB. 

When we load this page we check the number of minutes, then check whether it rounds to a full amount of days, if not, we do the same for hours, if not, we show minutes. 

What do you think of this idea? I think it makes the solution quite a bit easier to use throughout the rest of our app.

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -114,13 +114,99 @@ export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">)
...
-            {...formMethods.register("minimumBookingNotice", { valueAsNumber: true })}
+        <div className="flex w-full items-end">
+          <Controller
+            name="minimumBookingNotice"
+            control={formMethods.control}
+            render={() => {
+              const durationArray = [eventType.minimumBookingNoticeType];
+              const minBookingValue = formMethods.watch("minimumBookingNotice");
+              let newMinBookingValue: number;
+              const convertToNewDurationType = function (
```
JeroenReumkens, line 125: Could you define this function definition outside of this render method, or maybe even move it to `packages/lib`?  There's no reason for this function to be redefined on every render, plus we keep the renders a bit clearer. 

And you will also notice that you use the `eventType.minimumBookingNoticeType` inside your function, instead of expecting it as an argument. Using these parameters outside of your function can have very unpredictable side effects, so I will really advice against doing that, and use them as an argument instead.

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -148,6 +149,17 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
     endDate: new Date(eventType.periodEndDate || Date.now()),
   });
 
+  const convertMinimumBookingNoticeToMinutes = (type: string, minNotice: number) => {
```
Comment (JeroenReumkens) [inline apps/web/pages/event-types/[type]/index.tsx:152]: I think we can move this to a similar util file in `packages/lib` as the one I mentioned above.

packages/lib/slots.ts
```diff
@@ -39,9 +40,31 @@ const splitAvailableTime = (
   return result;
 };
 
-const getSlots = ({ inviteeDate, frequency, minimumBookingNotice, workingHours, eventLength }: GetSlots) => {
+const convertMinimumBookingNoticeToMinutes = (type: string, minNotice: number) => {
```
JeroenReumkens, line 43: Same util here as well.

packages/ui/v2/core/form/fields.tsx
```diff
@@ -453,6 +453,9 @@ export function InputGroupBox(props: JSX.IntrinsicElements["div"]) {
   );
 }
 
-export const MinutesField = forwardRef<HTMLInputElement, InputFieldProps>(function MinutesField(props, ref) {
-  return <InputField ref={ref} type="number" min={0} {...props} addOnSuffix="mins" />;
+export const DurationField = forwardRef<HTMLInputElement, InputFieldProps>(function DurationField(
```
JeroenReumkens, line 456: I think there is very little reason to keep this component in here now, since it's actually just an input field with `type=number`. What do you think about removing it here and just use the `InputField` on the page?

JeroenReumkens (changes requested):
Added a few suggestions which I think would make this even more awesome 🔥  
Thanks for the super nice work @om-ray!

apps/web/components/v2/eventtype/EventLimitsTab.tsx
om-ray: @JeroenReumkens I'll add this asap!

om-ray:
@JeroenReumkens I've made all the changes you asked for, please review when you can! thank you so much for the awesome suggestions!

packages/lib/convertMinimumBookingNoticeToMinutes.ts
```diff
@@ -0,0 +1,10 @@
+export default function convertMinimumBookingNoticeToMinutes(type: string, minNotice: number) {
+  if (type == "minute") {
+    return minNotice;
+  } else if (type == "hour") {
+    return minNotice * 60;
+  } else if (type == "day") {
+    return minNotice * 1440;
```
alannnc, line 7:
```suggestion
  if (type === "minute") {
    return minNotice;
  } else if (type === "hour") {
    return minNotice * 60;
  } else if (type === "day") {
    return minNotice * 1440;
```

packages/lib/convertToNewDurationType.ts
```diff
@@ -0,0 +1,25 @@
+export default function convertToNewDurationType(prevType: string, newType: string, prevValue: number) {
```
alannnc, line 1: NIT:
Declare any integer as an exportable const with a definition as its name and reuse it. Probably we already had them.

export const MINUTES_IN_HOUR = 60;
export const MINUTES_IN_DAY = 1440;
export const HOURS_IN_DAY = 24;

packages/lib/convertMinimumBookingNoticeToMinutes.ts
om-ray: Thank you so much for this, I need to remember to use `===` instead of `==` more often!

alannnc (changes requested):
Committed some changes for mobile.

image
Something we can improve its do not show days as float point. Instead if we receive 0.9 or below change value to 1 day. I'm not fan of 0.988261661 as day value.
Also looks like minutes box is 2 px smaller on its height than the select box.
One of minute box or select box its border color looks different from each other.

emrysal (dismissed):
Thank you for spearheading this feature!

It's heading in the right direction but I've got some suggestions;

Re. the styling:

image

I suggest we can create a component that embeds a select within the addOnSuffix, much like so (although this demonstrates a prefix)

image

(Tagging @Jaibles too so he is aware)

I'd like to propose renaming `minimumBookingNoticeType` to `minimumBookingNoticeUnit` as this better describes its function. But the other question is whether we should have this field at all given the unit does not change the function. We can store everything as minutes in the current implementation and simply show hours or days if % 60 = 0 or % 1440 = 0.

om-ray:
@emrysal The reason for the field of `minimumBookingNoticeType` is so that the user doesn't need to do the conversions between minutes, hours, and days when entering the time they want, if we get rid of the field, the user would need to do the conversion themselves instead of letting the app do it for them. I will work on the component with the `addOnSuffix` though as I initially thought of using that as the design! Thank you so much for your input!

JeroenReumkens:
@om-ray Amazing improvements you did based on my first review. I really like the simplicity this shows now 🙏

PeerRich:
@JeroenReumkens wanna approve?

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -254,6 +261,7 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
             periodStartDate: periodDates.startDate,
             periodEndDate: periodDates.endDate,
             periodCountCalendarDays: periodCountCalendarDays === "1",
+            minimumBookingNotice: minimumBookingNoticeInMinutes,
```
Comment (JeroenReumkens) [inline apps/web/pages/event-types/[type]/index.tsx:264]: @om-ray  Locally I still see "0" being saved when I save the minimum booking notice. Could you test this and see if it also occurs on your side?

JeroenReumkens (changes requested):

JeroenReumkens:
@om-ray @PeerRich  Did a final check, and noticed that it doesn't save the value for me. Could you please recheck this? 
Besides that super happy with this change, works perfectly. Even the rounding when changing between hours, minutes, days. 👏

apps/web/pages/event-types/[type]/index.tsx
Comment (om-ray) [reply apps/web/pages/event-types/[type]/index.tsx:264]: @JeroenReumkens I'll check right now!

om-ray:
@JeroenReumkens I just checked and it seems to work perfectly for me. If this issue keeps happening please let me know and I'll see if I can fix it!

om-ray (commented):
@PeerRich Should I change the design in the way @emrysal suggested?

Jaibles:
@om-ray I can design that component, we can decide then whether it needs to be part of this pr or whether it can be a follow up.

JeroenReumkens:
@om-ray  I've just retested this, and it doesn't seem to send 0 actually, but it IS always sending the old value. So it is not updating correctly. Could you retest and see if you can reproduce this? See this loom:

https://www.loom.com/share/5845fdcb815f4c8fb7c4625d4efa6e52

JeroenReumkens (changes requested):
See above loom, we first have to fix that before we can merge.

om-ray:
@JeroenReumkens that's really strange, I dont remember that happening before, ill fix it asap! Thank you so much for your review!

om-ray:
@JeroenReumkens I've fixed the issue, please review it again! Thank you so much for your review!

packages/lib/convertToNewDurationType.ts
```diff
@@ -0,0 +1,35 @@
+export const MINUTES_IN_HOUR = 60;
+export const MINUTES_IN_DAY = 1440;
+export const HOURS_IN_DAY = 24;
+
+export default function convertToNewDurationType(prevType: string, newType: string, prevValue: number) {
```
JeroenReumkens, line 5: Shall we create a union for `minutes, hours, days`? This way can't make that typo anymore! 😁

packages/lib/convertToNewDurationType.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (newType === "minutes") {
+    if (prevType === "hours") {
+      console.log("hours to minutes", prevValue * MINUTES_IN_HOUR);
```
JeroenReumkens, line 8: Could you please remove this logs again? 😄

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -150,6 +151,14 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
     endDate: new Date(eventType.periodEndDate || Date.now()),
   });
 
+  const minimumBookingNoticeInMinutes = convertToNewDurationType(
+    durationArray[0],
+    "minutes",
+    eventType.minimumBookingNotice
+  );
+
+  console.log(minimumBookingNoticeInMinutes, durationArray, eventType.minimumBookingNotice);
```
Comment (JeroenReumkens) [inline apps/web/pages/event-types/[type]/index.tsx:160]:
```suggestion
```

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -1,21 +1,31 @@
...
+import findDurationType from "@calcom/lib/findDurationType";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { PeriodType } from "@calcom/prisma/client";
 import type { BookingLimit } from "@calcom/types/Calendar";
 import { Icon } from "@calcom/ui";
-import { Select, Switch, Label, Input, MinutesField, Button, SettingsToggle } from "@calcom/ui/v2";
+import { Select, Switch, Label, Input, Button, SettingsToggle, InputField } from "@calcom/ui/v2";
 import DateRangePicker from "@calcom/ui/v2/core/form/date-range-picker/DateRangePicker";
 
+export const durationArray: string[] = ["minutes"];
```
JeroenReumkens, line 17: I didn't notice the usage of this array outside of the component before. I think accessing changing variables outside the render function is very prone to breaking. I would suggest adding a `useRef` variable inside your component where you track it's previous value. In that case you can also use a string instead of an array by eg setting `previousValueRef.current = ....`. In my opinion that also makes it again a bit more readable 🥰

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -1,21 +1,31 @@
...
+
 export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">) => {
   const { t } = useLocale();
   const formMethods = useFormContext<FormValues>();
   const { eventType } = props;
+  let minimumBookingNoticeType = findDurationType(eventType.minimumBookingNotice);
```
JeroenReumkens, line 23: I think we should also change this to a ref, since also changing normal variables in a component is not a guarantee that the component wil rerender. Now I'm thinking of it — perhaps we don't even need this to be a changing value anymore as soon as we make the previousValue a ref, since you're only using it to set an initial value and shifting the array (which will be gone when using a ref). Let me know if you want more input from me on this.

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -31,6 +41,11 @@ export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">)
     },
   ];
 
+  useEffect(() => {
+    durationArray.unshift(minimumBookingNoticeType);
+    formMethods.setValue("minimumBookingNotice", displayValue);
```
JeroenReumkens, line 46: Could we set this as a `defaultValue` on the `minimumBookingNotice` controller? Or wouldn't that work? Would be nice if we could get rid of this effect all together (since the unshift will already be removed if the above comments are fixed).

JeroenReumkens (changes requested):
Hi @om-ray 

It is working now! 👏  Amazing! 
I did spot a few other things while reviewing it now which I think will make it even simpler to understand. Could you please have a look at them still? Do let me know if you have any questions around this! 

Thanks again — really appreciate all the effort you put in this!

apps/web/components/v2/eventtype/EventLimitsTab.tsx
om-ray: @JeroenReumkens I tried using a `defaultValue` but since react-hook-form makes it so that default values aren't set if they are undefined,  the default value isn't set properly and you see the value as minutes

packages/lib/convertToNewDurationType.ts
om-ray: @JeroenReumkens Oh right! I'm so sorry about that! I completely forgot about them! Thank you so much for reminding me!

packages/lib/convertToNewDurationType.ts
om-ray: @JeroenReumkens This is a great idea! ill add this now! Thank you so much

apps/web/components/v2/eventtype/EventLimitsTab.tsx
om-ray: @JeroenReumkens If I use a `useRef` how would I access the value of the ref outside of this component? I need to use the current duration type to convert the `minimumBookingNotice` to minutes before sending it to prisma.

om-ray: @JeroenReumkens I'm not very familiar with using refs, but I will try and figure this out! Thank you for the suggestion!

om-ray:
@JeroenReumkens Thank you so so much for these amazing suggestions, I've implemented them and now the code is so much more readable! Thank you again for your amazing review!

apps/web/components/v2/eventtype/EventLimitsTab.tsx
```diff
@@ -1,23 +1,34 @@
...
 import { PeriodType } from "@calcom/prisma/client";
 import type { BookingLimit } from "@calcom/types/Calendar";
 import { Icon } from "@calcom/ui";
 import { Button } from "@calcom/ui/components";
-import { Label, Input, MinutesField } from "@calcom/ui/components/form";
+import { Label, Input, InputField } from "@calcom/ui/components/form";
 import { Select, SettingsToggle } from "@calcom/ui/v2";
 import DateRangePicker from "@calcom/ui/v2/core/form/date-range-picker/DateRangePicker";
 
+export let CurrentDurationType: string | "minutes" | "hours" | "days";
```
JeroenReumkens, line 19: @om-ray Why are we exactly changing this variable and importing this elsewhere? Changing this variable from within a component and reusing it somewhere else is not a react standard, this variable might not update. Could we also use a ref for this? 

If we need this value in the parent component we should probably work with a change handler that we pass in as a prop I think?

JeroenReumkens (dismissed):

JeroenReumkens:
> @JeroenReumkens Thank you so so much for these amazing suggestions, I've implemented them and now the code is so much more readable! Thank you again for your amazing review!

Amazing work! I've added one more remark in the code where we change a variable outside of the component. I think we still have to fix that as well. Thanks for sticking with me through these changes, really appreciated 🙏

om-ray:
@JeroenReumkens Thank you so much for your reviews! They've really increased the quality of the code and I've learned a lot from implementing these changes! I fixed the issue you mentioned and I would love if you could review it again!

alannnc (approved):

JeroenReumkens:
@om-ray  Thanks again for all the changes 🙏  
I did still see some weird behaviour with the form validation where I got an error telling me the current value was not a valid duration, it was then saying I filled a veeeery long decimal number, which wasn't visible to the user at all 🤔  

So I started to tinker with this a bit and actually changed quite a few things in this commit: https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b

* The default value of the input is now set in the [defaults of the form](https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b#diff-d75f96ff093473eefda3b50645504e299af40b26538bf2b235dc067293d3af2cR172) rather than in [the default of the field](https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b#diff-c85c0dfd7aa6ecafe494a7ede752ad7198998a96bbd5584bafa519f2f8163ddfL170)
* I changed the onchange functions for the fields to [functions outside of the render](https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b#diff-c85c0dfd7aa6ecafe494a7ede752ad7198998a96bbd5584bafa519f2f8163ddfL170). This way we can memoize them, to prevent any weird side effects not updating values that actually should.
* In the old code we were still in a few places updating the `eventType` object like [eg here](https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b#diff-c85c0dfd7aa6ecafe494a7ede752ad7198998a96bbd5584bafa519f2f8163ddfL175). However, this is called 'two-way-binding', a principle that does not exist in React. You can not rely on React updating the values if you change a prop this way. So I removed that and actually [used `formMethods.setValue`](https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b#diff-c85c0dfd7aa6ecafe494a7ede752ad7198998a96bbd5584bafa519f2f8163ddfR88) to update this value. The original value won't change, but the value of the form does, which is actually what we want to achieve. Before this probably (partially) worked because React was triggering a rerender, which still updated the values. However, in production this behaviour could be very different than in dev.
* Because the [value in minutes in the form](https://github.com/calcom/cal.com/pull/5217/commits/6d75920c7050bd2c7992d57976920c8c55d7c48b#diff-d75f96ff093473eefda3b50645504e299af40b26538bf2b235dc067293d3af2cL157) is now correct and updates all the time, we don't need the change handler in the page anymore.
* Finally I've also added some comments to explain the behaviour of converting the types. 

I hope my changes make sense, and you can agree with the changes I made 😄 

Could you and @emrysal please both test this one more time? I did some extensive testing myself, but having more pair of eyes on this is really valuable!

Thanks 🙏

om-ray:
@JeroenReumkens I looked through the changes and it made sense to me! I like how you did it, it's a lot better and it's definitely more readable! Thank you so much for putting so much time and care into this!

JeroenReumkens:
Thank you so much for you kind words @om-ray . I've requested a new review with the team so we can get this merged asap 🙏  Thanks for all the time you put in as well!

apps/web/components/eventtype/EventLimitsTab.tsx
```diff
@@ -1,23 +1,34 @@
...
+import { Label, Input, InputField } from "@calcom/ui/components/form";
 import { Select, SettingsToggle } from "@calcom/ui/v2";
 import DateRangePicker from "@calcom/ui/v2/core/form/date-range-picker/DateRangePicker";
 
-export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">) => {
+type EventLimitType = {
+  eventType: Pick<EventTypeSetupInfered, "eventType">;
+};
+
+export const EventLimitsTab = (props: EventLimitType) => {
```
hariombalhara, line 23:
```suggestion
export const EventLimitsTab = (props: {eventType: Pick<EventTypeSetupInfered, "eventType">}) => {
```

hariombalhara: I find this approach better. Plus, EventLimitType name seems wrong.

packages/lib/convertMinimumBookingNoticeToMinutes.ts
```diff
@@ -0,0 +1,12 @@
+import { MINUTES_IN_DAY, MINUTES_IN_HOUR } from "./convertToNewDurationType";
+
+export default function convertMinimumBookingNoticeToMinutes(type: string, minNotice: number) {
```
hariombalhara, line 3: Seems unused

packages/lib/convertToNewDurationType.ts
```diff
@@ -0,0 +1,43 @@
+export const MINUTES_IN_HOUR = 60;
+export const MINUTES_IN_DAY = 1440;
+export const HOURS_IN_DAY = 24;
+
+export default function convertToNewDurationType(
```
hariombalhara, line 5: `"string" | "minutes" | "hours" | "days"` is equivalent to `string`
Screenshot 2022-11-11 at 5 00 53 PM

hariombalhara (approved):
Works well as per code. Small suggestions(nits).

Though, I think if it is acceptable in UI to just discard the previous value, it would made it much less complex. 

The automatic conversions that we from one type to another are pretty simple and user can do themselves as well.

Another thing is probably the conversions aren't that helpful. Say I enter 15 hrs. When I switch to Days, it would show 1. I am not sure how that is helpful.

PeerRich:
@hariombalhara should your comments be addressed or is this safe to merge?

JeroenReumkens:
> The automatic conversions that we from one type to another are pretty simple and user can do themselves as well.
> 
> Another thing is probably the conversions aren't that helpful. Say I enter 15 hrs. When I switch to Days, it would show 1. I am not sure how that is helpful.

Good question to ask if this is really necessary. Personally I actually like having the automatic conversion, but I can also imagine that user find it very confusing, since it's not really logical that we round up to the nearest value (so 15 hours rounds to 1 day if you change from hours to days). @Jaibles Could you test this in the test env as soon as it is deployed, and let us know whether you find the automatic conversion a good addition or actually bringing more confusion than benefit?

hariombalhara:
It's mergeable. It's mostly UX wr can improve upon

apps/web/components/eventtype/EventLimitsTab.tsx
```diff
@@ -1,23 +1,34 @@
...
+import { Label, Input, InputField } from "@calcom/ui/components/form";
 import { Select, SettingsToggle } from "@calcom/ui/v2";
 import DateRangePicker from "@calcom/ui/v2/core/form/date-range-picker/DateRangePicker";
 
-export const EventLimitsTab = (props: Pick<EventTypeSetupInfered, "eventType">) => {
+type EventLimitType = {
+  eventType: Pick<EventTypeSetupInfered, "eventType">;
+};
+
+export const EventLimitsTab = (props: {eventType: Pick<EventTypeSetupInfered, "eventType">}) => {
```
zomars, line 23:
```suggestion
export const EventLimitsTab = (props: EventLimitType) => {
```

apps/web/components/eventtype/EventLimitsTab.tsx
```diff
@@ -1,23 +1,34 @@
...
   const { t } = useLocale();
   const formMethods = useFormContext<FormValues>();
-  const { eventType } = props;
+  const { eventType } = props.eventType;
```
zomars, line 26: Shouldn't this be like this?
```suggestion
  const { eventType } = props;
```

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -199,7 +207,7 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
         currentUserMembership={props.currentUserMembership}
       />
     ),
-    limits: <EventLimitsTab eventType={eventType} />,
+    limits: <EventLimitsTab eventType={{ eventType: eventType }} />,
```
Comment (zomars) [inline apps/web/pages/event-types/[type]/index.tsx:210]: Why?

apps/web/test/lib/getSchedule.test.ts
```diff
@@ -477,6 +477,7 @@ describe.skip("getSchedule", () => {
         eventType: {
           id: 1,
           minimumBookingNotice: 0,
+
```
zomars, line 480:
```suggestion
```

apps/web/test/lib/getSchedule.test.ts
```diff
@@ -552,6 +553,7 @@ describe.skip("getSchedule", () => {
         eventType: {
           id: 2,
           minimumBookingNotice: 0,
+
```
zomars, line 556:
```suggestion
```

packages/lib/convertToNewDurationType.ts
```diff
@@ -0,0 +1,43 @@
...
+  prevType: string | "minutes" | "hours" | "days",
+  newType: string | "minutes" | "hours" | "days",
```
zomars, line 7: Agreed with @hariombalhara. Adding `string` defeats the purpose of setting an intersection.

```suggestion
  prevType: "minutes" | "hours" | "days",
  newType: "minutes" | "hours" | "days",
```

packages/lib/convertMinimumBookingNoticeToMinutes.ts
zomars: Agreed. Let's remove this file

packages/lib/convertToNewDurationType.ts
```diff
@@ -0,0 +1,43 @@
...
+      return round(prevValue);
+    }
+    if (prevType === "minutes") {
+      return round(prevValue / MINUTES_IN_DAY);
+    }
+    if (prevType === "hours") {
+      return round(prevValue / HOURS_IN_DAY);
+    }
+  }
+  return round(prevValue);
```
zomars, line 42: Legibility can be further improved upon. Also let's try to avoid as much nested conditions as possible.

```suggestion
  /** Convert `prevValue` from `prevType` to `newType` */
  const newDurationTypeMap = {
    minutes_minutes: () => prevValue,
    minutes_hours: () => prevValue * MINUTES_IN_HOUR,
    minutes_days: () => prevValue * MINUTES_IN_DAY,
    hours_minutes: () => prevValue / MINUTES_IN_HOUR,
    hours_hours: () => prevValue,
    hours_days: () => prevValue * HOURS_IN_DAY,
    days_minutes: () => prevValue / MINUTES_IN_DAY,
    days_hours: () => prevValue / HOURS_IN_DAY,
    days_days: () => prevValue,
  };
  const getNewValue = newDurationTypeMap[`${prevType}_${newType}`];
  const newValue = getNewValue();
  return Math.ceil(newValue);
```

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -341,7 +341,3 @@ export function InputGroupBox(props: JSX.IntrinsicElements["div"]) {
     </div>
   );
 }
-
-export const MinutesField = forwardRef<HTMLInputElement, InputFieldProps>(function MinutesField(props, ref) {
```
zomars, line 345: Is this not needed any more?

packages/lib/findDurationType.ts
```diff
@@ -0,0 +1,11 @@
+import { MINUTES_IN_DAY, MINUTES_IN_HOUR } from "./convertToNewDurationType";
+
+export default function findDurationType(value: number) {
+  if (value % MINUTES_IN_DAY === 0) {
+    return "days";
+  } else if (value % MINUTES_IN_HOUR === 0) {
+    return "hours";
+  } else {
+    return "minutes";
+  }
```
zomars, line 10: No need for `if/else` if we're returning anyways
```suggestion
  if (value % MINUTES_IN_DAY === 0) return "days";
  if (value % MINUTES_IN_HOUR === 0) return "hours";
  return "minutes";
  }
```

zomars (changes requested):
Agreed with @hariombalhara that is mergeable. But will block to improve upon some legibility suggestions.

apps/web/pages/event-types/[type]/index.tsx
Comment (om-ray) [reply apps/web/pages/event-types/[type]/index.tsx:210]: Since I was using `const { eventType } = props.eventType;` i had to pass it in as an object, but now that that's fixed we can use this

packages/ui/components/form/inputs/Input.tsx
om-ray: No, the `minimumBookingNotice` was the only place in the codebase this was used

zomars:
WE have a type error

zomars (approved):
Fixed type. Nice job @om-ray 

LGTM

alannnc:
image
Small type fixes

PR #5224, opened 2022-10-26
using WEBAPP_URL instead of BASE_URL in constants.ts

servusrene:
## What does this PR do?

Using WEBAPP_URL instead of BASE_URL in constants.ts

## Checklist


PeerRich:
i believe there is a reason for this, asking @zomars why

zomars:
Is this is not fixing something specifically, then there's no reason to merge.

rene-mueller:
@zomars true this does not fix a bug, but the use of a variable marked as deprecated
fyi @PeerRich

zomars:
> @zomars right this does not fix an error, but the use of a variable marked as deprecated
> fyi @PeerRich

Thank you for your contribution 🙏 In this case it would make sense to handle all deprecation replacements in one single PR at once.

PR #5234, opened 2022-10-26
use LOGO constant instead of fixed string in OgImages

servusrene:
## What does this PR do?

 use LOGO constant instead of fixed string in OgImages 

## Type of change


PeerRich:
asked @JeroenReumkens to review, also tehre is a tiny merge conflict. great idea to use consts here

packages/lib/OgImages.tsx
```diff
@@ -105,7 +105,7 @@ export const Meeting = ({ title, users = [], profile }: MeetingImageProps) => {
     <Wrapper>
       <div tw="h-full flex flex-col justify-start">
         <div tw="flex items-center justify-center" style={{ fontFamily: "cal", fontWeight: "300" }}>
-          <img src={`${CAL_URL}/cal-logo-word-black.svg`} width="350" alt="Logo" />
+          <img src={`${CAL_URL}/${LOGO}`} width="350" alt="Logo" />
```
hariombalhara, line 108: LOGO constant has different value than this. 
@rene-mueller can you share a before and after screenshot

rene-mueller: @hariombalhara the name is different, but the content is the same:
* https://cal.com/calcom-logo-white-word.svg
* https://cal.com/cal-logo-word-black.svg

hariombalhara: @PeerRich Should we just have a single one of these?

JeroenReumkens: Hey @rene-mueller  @hariombalhara  
The logo in the `-black.svg` file is black, where in the `white-word.svg` the logo is not fully black. So there's actually a difference in color here. I don't think we can really change this with CSS if we're using an `` tag, right? Let me know if you have another idea on using this SVG but having the right color as well 👌

rene-mueller: @hariombalhara by the way, there is another logo that is double to the ones above:
* cal.com/cal-logo-word.svg

rene-mueller: @JeroenReumkens no, with css you can't change the color of an svg in a . but it would be possible to embed the svg directly as svg, but this way you lose the possibility to swap the logo via a variable. wouldn't it just be possible to use the same logo? my goal is to change the logo to make it even easier to whitelabel.

PeerRich: i actually dont know if we can remove any of those, we may link them in some places. but we should consolidate all logos into one

PeerRich (approved):

packages/lib/OgImages.tsx
JeroenReumkens: Embedding them as an SVG is definitely an option. Only wondering how we're going to import them, without having to turn them into react components first? Feel free to experiment with this 🙏  

Agree with Peer indeed, those logo's being in the public folder makes it hard to know where they're really used (even outside the platform).

rene-mueller: Including an SVG in React is actually straightforward. It is simply imported and used like a ts or tsx component.

```
import React from 'react';
import {ReactComponent as ReactLogo} from './logo.svg';

const App = () => {
  return (
    <div className="App">
      <ReactLogo />
    </div>
  );
}
export default App;
```

That's true … i think there has to be a hard cut at some point. currently i just swap the string locally with a constant. so i think it should just be changed.

JeroenReumkens: @rene-mueller  If by importing them that way we could still change the color, I'm all for it 🙏  
Like Peer said before though, I don't think we can remove the svg's from the public folder yet, because we don't know all the places where they are used. But removing them from the OG image if possible would already be a great start. Do you want to try to tackle that?

PR #5240, opened 2022-10-27
Stop long urls from overflowing on booking page.

JeroenReumkens:
## What does this PR do?

Prevents long urls (or other very long texts for that matter) from overflowing on the public booking page. It will only break long words if they're longer than one sentence.  

It's not the most beautiful thing, but for edge cases this works and is way better than the overflowing. Probably down the line we can add the HTML editor @CarinaWolli is working on here as well, and give people to option to add real anchors? 

### Before

CleanShot 2022-10-27 at 09 49 34@2x

### After 

CleanShot 2022-10-27 at 09 49 04@2x

## Type of change

-  Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [ ] Create an event with a long discription with either a very long word or url in there (without dashes, . etc, since it will break there too), and notice that it now breaks it over multiple lines.

### DIFF
--- a/apps/web/components/booking/BookingDescription.tsx
+++ b/apps/web/components/booking/BookingDescription.tsx
@@ -70,7 +70,9 @@ const BookingDescription: FC<Props> = (props) => {
                 )}
               />
             </div>
-            <EventTypeDescriptionSafeHTML eventType={eventType} />
+            <div className="max-w-[calc(100%_-_2rem)] flex-shrink break-words">
+              <EventTypeDescriptionSafeHTML eventType={eventType} />
+            </div>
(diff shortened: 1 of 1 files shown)


harshsinghatz:
Tested on mobile viewport?

JeroenReumkens:
@harshsinghatz  Jep! 
On wider mobiles it will even show more of it: 

CleanShot 2022-10-27 at 09 57 44@2x

On smaller screens it overflows the same: 
CleanShot 2022-10-27 at 09 58 31@2x

pumfleet (approved):

PR #5241, opened 2022-10-27
Fix Zapier sample data and payload

CarinaWolli:
## What does this PR do?

The sample data (`/listBookings` endpoint) shown to the user when testing a Zapier trigger should show the same data as the `sendPayload` function sends once a zap triggers. 

## Type of change


hariombalhara (approved):
Looks good.

PR #5248, opened 2022-10-27
Buffer limits fix

sean-brydon:
This PR fixes the buffer limit logic for before + after limits.

I will record a Loom for this tomorrow if needed. Please checkout and test this branch on preview to ensure the correct behaviour is indented.


packages/core/getBusyTimes.ts
```diff
@@ -39,12 +42,19 @@ export async function getBusyTimes(params: {
         startTime: true,
         endTime: true,
         title: true,
+        eventType: {
+          select: {
+            afterEventBuffer: true,
```
CarinaWolli, line 47: we could add afterEventBuffer also as a parameter like beforeEventBuffer, to keep it consistent

packages/core/getBusyTimes.ts
```diff
@@ -39,12 +42,19 @@ export async function getBusyTimes(params: {
...
+          },
+        },
       },
     })
     .then((bookings) =>
-      bookings.map(({ startTime, endTime, title, id }) => ({
-        end: endTime,
+      bookings.map(({ startTime, endTime, title, id, eventType }) => ({
         start: startTime,
+        end: dayjs(endTime)
```
CarinaWolli, line 55: wouldn't it make more sense to leave it without buffer times here and add it in getUserAvailability in bufferedBusyTimes?

packages/core/getUserAvailability.ts
```diff
@@ -134,14 +136,15 @@ export async function getUserAvailability(
...
 
   const bufferedBusyTimes: EventBusyDetails[] = busyTimes.map((a) => ({
     ...a,
-    start: dayjs(a.start).subtract(currentUser.bufferTime, "minute").toISOString(),
-    end: dayjs(a.end)
-      .add(currentUser.bufferTime + (afterEventBuffer || 0), "minute")
+    start: dayjs(a.start)
+      .subtract(afterEventBuffer || 0, "minutes")
       .toISOString(),
+    end: dayjs(a.end).toISOString(),
```
CarinaWolli, line 147: so adding the beforeEventBuffer and afterEventBuffer here instead

apps/web/test/lib/slots.test.ts
```diff
@@ -5,109 +5,14 @@ import dayjs from "@calcom/dayjs";
...
-          startTime: MINUTES_DAY_START,
-          endTime: MINUTES_DAY_END,
-        },
-      ],
-      eventLength: 60,
-    })
-  ).toHaveLength(11);
-});
-
-it("adds buffer time", async () => {
```
CarinaWolli, line 107: can we update the tests instead of deleting them?

packages/core/getUserAvailability.ts
```diff
@@ -134,14 +136,15 @@ export async function getUserAvailability(
...
+    beforeEventBuffer,
   });
 
   const bufferedBusyTimes: EventBusyDetails[] = busyTimes.map((a) => ({
     ...a,
-    start: dayjs(a.start).subtract(currentUser.bufferTime, "minute").toISOString(),
-    end: dayjs(a.end)
-      .add(currentUser.bufferTime + (afterEventBuffer || 0), "minute")
+    start: dayjs(a.start)
+      .subtract(afterEventBuffer || 0, "minutes")
```
CarinaWolli, line 145: I think we need to add the beforeEventBuffer here too. I was able to book this with a beforeEventBuffer of 15 minutes:
Screenshot 2022-10-28 at 15 33 52

CarinaWolli (changes requested):
I did not fully follow the conversation about buffers but I think there is something still not right yet 🤔 I have a before buffer of 15 minutes, so I don't think I should be able to make my bookings like that: 

Screenshot 2022-10-28 at 15 38 42

apps/web/test/lib/slots.test.ts
sean-brydon: I believe this should be picked up in another PR

packages/core/getBusyTimes.ts
sean-brydon: Not exactly - we need this event types afterEventBuffer not the one we are currently booking.

packages/core/getUserAvailability.ts
sean-brydon: I will investigate this on Monday :) Although this is different behaviour from what myself and @Jaibles have experienced.

apps/web/test/lib/slots.test.ts
sean-brydon: These tests were not failing btw - these tests were testing code that is never used within our app.

packages/core/getUserAvailability.ts
emrysal: Confirmed this is indeed a bug @CarinaWolli - nice spot. It happens when you book 2:30 before you book 2:15. beforeBuffer needs to be taken into account in the busyTime

Jaibles:
> I did not fully follow the conversation about buffers but I think there is something still not right yet 🤔 I have a before buffer of 15 minutes, so I don't think I should be able to make my bookings like that: 
> 
> 
> 
> Screenshot 2022-10-28 at 15 38 42
> 
> 

Yea the conversation was largely about ensuring you cannot end up with back to back meetings (and preventing it actually requires adding a buffer before and after even if you've just set before)

alishaz-polymath:
Pretty sure in this logic we can still end up with back to back bookings by simply booking through an event type which doesn't have buffer. @sean-brydon @Jaibles

packages/core/getBusyTimes.ts
```diff
@@ -39,12 +51,19 @@ export async function getBusyTimes(params: {
...
+          .add((eventType?.afterEventBuffer || afterEventBuffer || 0) + (beforeEventBuffer || 0), "minute")
```
CarinaWolli, line 65: I think we only need to add eventType.afterEventBuffer here. If there is none, we don't need to add the afterEventBuffer from the booking.

CarinaWolli (commented):
I left you one comment. However, I found another bug that has nothing to do with your PR that makes it unable to even test this. I tried to book 2 different event types (one has only a before buffer and one has before and after), but I realized that `busyTimes` only includes the bookings from the event type that is currently booked. I think we need to remove the `eventTypeId` when querying the `busyTimes`, so we also include all other bookings from different event types in the availabilities

packages/core/getBusyTimes.ts
```diff
@@ -27,7 +39,6 @@ export async function getBusyTimes(params: {
     .findMany({
       where: {
         userId,
-        eventTypeId,
```
sean-brydon, line 30: Nice spot @CarinaWolli  - Looks like we were only getting busy events for this particular event type and not all event types.  Potentially allowing a double booking (same time) as each other

apps/web/test/lib/slots.test.ts
```diff
@@ -5,109 +5,14 @@ import dayjs from "@calcom/dayjs";
 import getSlots from "@calcom/lib/slots";
 
 import { MINUTES_DAY_END, MINUTES_DAY_START } from "@lib/availability";
-import { getFilteredTimes } from "@lib/hooks/useSlots";
 
 MockDate.set("2021-06-20T11:59:59Z");
 
-it("can fit 24 hourly slots for an empty day", async () => {
```
sean-brydon, line 12: Remove tests on code that isn't used in product anymore

packages/core/getBusyTimes.ts
```diff
@@ -39,12 +50,22 @@ export async function getBusyTimes(params: {
...
+        },
       },
     })
     .then((bookings) =>
-      bookings.map(({ startTime, endTime, title, id }) => ({
-        end: endTime,
-        start: startTime,
+      bookings.map(({ startTime, endTime, title, id, eventType }) => ({
+        start: dayjs(startTime)
+          .subtract((eventType?.beforeEventBuffer || 0) + (afterEventBuffer || 0), "minute")
```
sean-brydon, line 64: Thanks @CarinaWolli  for this fix :) 

Always subtract the beforeEventBuffer for this specific booking eventType - then add afterEventBuffer for the buffer you are trying to book. 

Preventing different event types from having a clash

packages/core/getBusyTimes.ts
```diff
@@ -55,7 +76,19 @@ export async function getBusyTimes(params: {
   if (credentials?.length > 0) {
     const calendarBusyTimes = await getBusyCalendarTimes(credentials, startTime, endTime, selectedCalendars);
 
-    busyTimes.push(...calendarBusyTimes); /*
+    busyTimes.push(
+      ...calendarBusyTimes.map((value) => ({
+        ...value,
+        end: dayjs(value.end)
```
sean-brydon, line 82: Always add beforeEventBuffer to the end time so we get separation on calendar events 

Without this we could book a meeting straight after a normal calendar event.

packages/core/getBusyTimes.ts
```diff
@@ -55,7 +76,19 @@ export async function getBusyTimes(params: {
...
+          .add(beforeEventBuffer || 0, "minute")
+          .toDate(),
+        start: dayjs(value.start)
+          .subtract(afterEventBuffer || 0, "minute")
```
sean-brydon, line 86: Always subtract afterEventBuffer if it exists

packages/core/getUserAvailability.ts
```diff
@@ -134,14 +136,14 @@ export async function getUserAvailability(
...
     userId: currentUser.id,
     selectedCalendars,
+    beforeEventBuffer,
+    afterEventBuffer,
   });
 
   const bufferedBusyTimes: EventBusyDetails[] = busyTimes.map((a) => ({
     ...a,
-    start: dayjs(a.start).subtract(currentUser.bufferTime, "minute").toISOString(),
-    end: dayjs(a.end)
```
sean-brydon, line 142: This logic is handled in getBusyTimes now as bookings/calendar events require different logic

packages/trpc/server/routers/viewer/slots.tsx
```diff
@@ -50,13 +50,11 @@ const checkIfIsAvailable = ({
   time,
   busy,
   eventLength,
-  beforeBufferTime,
   currentSeats,
 }: {
   time: Dayjs;
-  busy: (TimeRange | { start: string; end: string } | EventBusyDate)[];
```
sean-brydon, line 57: These types just simply are EventBusyDate - not sure why we replicated this

packages/trpc/server/routers/viewer/slots.tsx
```diff
@@ -67,7 +65,7 @@ const checkIfIsAvailable = ({
   const slotStartTime = time.utc();
 
   return busy.every((busyTime) => {
-    const startTime = dayjs.utc(busyTime.start).subtract(beforeBufferTime, "minutes").utc();
+    const startTime = dayjs.utc(busyTime.start).utc();
```
sean-brydon, line 68: Handled in getBusyTimes

alishaz-polymath (approved):
Looks good to me. Buffer specific testing went well
Awesome job @sean-brydon

CarinaWolli (approved):
Tested and works as expected 👏🏻

PR #5269, opened 2022-10-28
Sendgrid app and code simplification

leog:
## What does this PR do?

Brings a new app for Sendgrid, creating attendees for an event as contacts in Sendgrid.

Also took the opportunity to simplify the code regarding utils being in their own file that were adding unneeded complexity.

 Loom Video: https://www.loom.com/share/0e234a2910894a10be73a4f6c5b7cf9b

## Type of change

## How should this be tested?

Install Sendgrid providing an API Key and receive a booking. You should be seeing the attendees created as contacts in your Sendgrid account.


packages/app-store/sendgridothercalendar/components/InstallAppButton.tsx
```diff
@@ -0,0 +1,18 @@
+import type { InstallAppButtonProps } from "@calcom/app-store/types";
```
hariombalhara, line 1: InstallAppButton.tsx is not required now. This file can be deleted.

packages/app-store/sendgridothercalendar/_metadata.ts
```diff
@@ -0,0 +1,26 @@
+import type { AppMeta } from "@calcom/types/App";
```
hariombalhara, line 1: @leog Can you try using app-store -cli to create the app. You can refer to https://developer.cal.com/guides/how-to-build-an-app. It would not add much code that's not required.
e.g. metadata is supposed to be pretty lean now.


Would love to fix any issues that come up with app-store-cli usage.

leog: Oh sure, I can create a new app using the CLI and then port what I have to it, so I leverage the latest structure you put together. It was not that I'm too lazy, as Sendgrid behaves the exact same way as closecom app, I copied the code from it 😅

packages/app-store/sendgridothercalendar/README.mdx
```diff
@@ -0,0 +1,10 @@
+---
+description: SendGrid delivers your transactional and marketing emails through the world's largest cloud-based email delivery platform.
+items:
+  - /api/app-store/sendgridothercalendar/1.png
+---
+
+<Slider items={items} />
```
JeroenReumkens, line 7: Since this week I cleaned up all v1 appstore related things, which also means this `Slider` component doesn't need to be rendered anymore. I guess you started this before the cleanup was merged, because it should also be gone from the template. Could you remove it here place? 😁

PeerRich:
nice!!

packages/app-store/sendgridothercalendar/README.mdx
leog: That came from the CLI if I'm not wrong cc: @hariombalhara

PeerRich:
can you record a loom video @leog ?

packages/app-store/sendgridothercalendar/README.mdx
leog: Also @hariombalhara, now with the absence of the Install Button component, for apps like this that require a previous step before installing it (setup page to get an API key) when the Install button of the app gets clicked, a toast message wrongly appears stating the app was successfully installed when it stills needs to redirect first.

leog:
> can you record a loom video @leog ?

Done: https://www.loom.com/share/0e234a2910894a10be73a4f6c5b7cf9b

packages/app-store/sendgridothercalendar/api/_postAdd.ts
```diff
@@ -0,0 +1,39 @@
...
+import prisma from "@calcom/prisma";
+
+import checkSession from "../../_utils/auth";
+import getInstalledAppPath from "../../_utils/getInstalledAppPath";
+
+export async function getHandler(req: NextApiRequest, res: NextApiResponse) {
+  const session = checkSession(req);
+
+  const { api_key } = req.body;
+  if (!api_key) throw new HttpError({ statusCode: 400, message: "No Api Key provoided to check" });
```
JeroenReumkens, line 16:
```suggestion
  if (!api_key) throw new HttpError({ statusCode: 400, message: "No Api Key provided to check" });
```

packages/app-store/sendgridothercalendar/api/_postAdd.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const encrypted = symmetricEncrypt(JSON.stringify({ api_key }), process.env.CALENDSO_ENCRYPTION_KEY || "");
+
+  const data = {
+    type: "sendgrid_other_calendar",
+    key: { encrypted },
+    userId: session.user?.id,
```
JeroenReumkens, line 23: Just thinking out loud here, since I see in other places we do this as well. But isn't it actually weird that the `userId` is optional here? Because what would happen if you added null/undefined into that field? That wouldn't result in a successful addition of the credentials, right? Even though Prisma accepts it and won't error.

packages/app-store/sendgridothercalendar/pages/setup/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+import z from "zod";
+
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+import { Button, Form, showToast, TextField } from "@calcom/ui/v2";
+
+const formSchema = z.object({
+  api_key: z.string(),
+});
+
+export default function CloseComSetup() {
```
JeroenReumkens, line 16:
```suggestion
export default function SendgridSetup() {
```

JeroenReumkens:
@leog  I see that you also did some changes to the CloseCom app. Would it make sense to add this in a separate PR? Otherwise it would be nice if you could add what / why you change about this in the PR description as well, and probably include that in the Loom? This way we won't forget to test this too 🙏

packages/app-store/sendgridothercalendar/api/_postAdd.ts
sean-brydon: Looking at this too - I agree it is weird and its strange that typescript doesn't recognise this as being not nullable	as in the `checkSession` function we do this https://github.com/calcom/cal.com/blob/6238b6d6b9d4f39d702c67cdf7788423ab4d546a/packages/app-store/_utils/auth.ts#L5

PeerRich:
> @leog I see that you also did some changes to the CloseCom app. Would it make sense to add this in a separate PR? Otherwise it would be nice if you could add what / why you change about this in the PR description as well, and probably include that in the Loom? This way we won't forget to test this too 🙏

@leog whats the status here? can you chase reviews and get this merged?

leog:
> > @leog I see that you also did some changes to the CloseCom app. Would it make sense to add this in a separate PR? Otherwise it would be nice if you could add what / why you change about this in the PR description as well, and probably include that in the Loom? This way we won't forget to test this too 🙏
> 
> @leog whats the status here? can you chase reviews and get this merged?

oh yeah, this PR was pending review before I went OOO and got reviews during; I can work on in some time today to have it merged for tomorrow’s release

leog:
> @leog I see that you also did some changes to the CloseCom app. Would it make sense to add this in a separate PR? Otherwise it would be nice if you could add what / why you change about this in the PR description as well, and probably include that in the Loom? This way we won't forget to test this too 🙏

Hey @JeroenReumkens! Good catch. Actually I mentioned in the PR, it’s just refactoring, moving things from their own utils to their corresponding expanded class. I tested myself already, otherwise you could check the Closecom app and the internal sync service too that uses that.

packages/app-store/sendgridothercalendar/api/_postAdd.ts
leog: Yeah, we might need to review this behavior to make sure there is no edge case that might force to an undesired app state. Resolving for this PR but @JeroenReumkens @sean-brydon feel free to open an issue to track it down.

sean-brydon: Looking at it - I think its an impossible edge case to ever hit. Should be okay on this front

packages/app-store/sendgridothercalendar/pages/setup/index.tsx
```diff
@@ -0,0 +1,157 @@
+import { zodResolver } from "@hookform/resolvers/zod";
+import { useRouter } from "next/router";
+import { useState, useEffect } from "react";
+import { Check, X } from "react-feather";
```
sean-brydon, line 4: Can we import these from "Calcom/ui/Icon` just for constancy sake

sean-brydon (approved):
LGTM one little NIT comment but code looks good - nice job :) The testing the api key was satisfying 

Tests are failing but im not sure if its relevant to this PR or not.

packages/app-store/sendgridothercalendar/pages/setup/index.tsx
leog: Done, thanks!

packages/app-store/sendgridothercalendar/README.mdx
hariombalhara: I think the slider think isn't there in CLI. Not sure how it came. But I expect it to not come for an app created using CLI.

@leog what should happen according to you. According to me, the app is installed, now the configuration is pending. Because at this moment, even if you skip the setup page configuration, it would show up as installed everywhere.

packages/prisma/seed-app-store.ts
```diff
@@ -240,6 +240,8 @@ export default async function main() {
   }
   // No need to check if environment variable is present, the API Key is set up by the user, not the system
   await createApp("closecom", "closecomothercalendar", ["other"], "closecom_other_calendar");
+  // No need to check if environment variable is present, the API Key is set up by the user, not the system
+  await createApp("sendgrid", "sendgridothercalendar", ["other"], "sendgrid_other_calendar");
```
hariombalhara, line 244: @leog I don't think it's required. `seed-app-store.config.json` has the entry already.

PR #5279, opened 2022-10-30
Fixes #5278 - wrong profile picture

### LINKED ISSUE #5278: [CAL-282] Colleague profile picture shown on booking page
All of my colleagues have my profile picture included in their link. This is getting messy. In their profiles they have their own pictures; it’s just their booking links.

emrysal:
## What does this PR do?

Fixes #5278

The reason this was happening is due to team events being included in the user list selection due to a duplicate slug, as the affected user created the team before the user event was created it was preferred - but obfuscated due to use of the profile object.

The PR removes the secondary query for fetching eventTypes in the [user]/[type].tsx file, instead fetching them directly through one single prisma query. It also adds an explicit check to make sure there is no teamId set on the EventType as these are included through the many-to-many relationship, comments added where useful.


alishaz-polymath (approved):
Awesome 🙏

PR #5286, opened 2022-10-31
🧹 Removed old v1 app components in web.

JeroenReumkens:
## What does this PR do?

Cleanup: Removes v1 for app components in web app. 

## Type of change

## How should this be tested?

- [ ] Ensure build still works
- [ ] Check app pages for any weird behaviour. Especially since <Slider> component removed from MDX (wasn't used anymore), double check the images as well.


packages/app-store/applecalendar/README.mdx
```diff
@@ -3,6 +3,4 @@ items:
   - /api/app-store/applecalendar/1.jpg
 ---
 
-<Slider items={items} />
```
JeroenReumkens, line 6: This slider was only used in V1. The new version uses the items array in the MDX meta.

sean-brydon (approved):
LGTM - So happy to be starting the migration to V2 files now :)

PR #5288, opened 2022-10-31
🧹  Cleaned up older v1 components for availability, settings and auth.

JeroenReumkens:
## What does this PR do?

More cleanup. Removed old v1 components for availability, settings and auth container. 

## Type of change

## How should this be tested?

Main things touched are listed below. There should be only one page that could have a functionality change (`/auth/error`), because all other pages were correctly using v2 pages, which are now moved outside the v2 folder after the v1 components (that weren't used anymore) were deleted. 

- [ ] /availability (skeleton loader)
- [ ] Page `old-getting-started` is removed as I don't believe it was still used? 
- [ ] settings/security — ensure 2factor auth modal still shows up when toggling
- [ ] /auth/error page (this component was still using v1, I've checked it with the new version now, and it still looks good (better actually 😉  )). 
- [x] Build is still working


apps/web/components/availability/SkeletonLoader.tsx
```diff
@@ -1,8 +1,8 @@
+import classNames from "classnames";
```
JeroenReumkens, line 1: This diff is here since the v1 and v2 file had the same name, I moved the v2 to this place and deleted v1. So here you see the differences between these two 🙂

apps/web/components/ui/AuthContainer.tsx
```diff
@@ -17,27 +17,25 @@ interface Props {
 
 export default function AuthContainer(props: React.PropsWithChildren<Props>) {
   return (
-    <div className="flex min-h-screen flex-col justify-center border-gray-200 bg-neutral-50 py-12 sm:px-6 lg:px-8">
```
JeroenReumkens, line 20: This diff is here since the v1 and v2 file had the same name, I moved the v2 to this place and deleted v1. So here you see the differences between these two 🙂

apps/web/components/availability/SkeletonLoader.tsx
PeerRich: i dont believe we need this package anymore since we have `@lib/classnames`

JeroenReumkens: Oh that's right indeed. I will already add that fix in this PR 👍

sean-brydon (approved):
LGTM

PR #5293, opened 2022-10-31
Fix

hariombalhara:


zomars:
Let's revert this please. This is not a critical vulnerability. Basically don't "Approve and run" workflows from external contributors without making sure they haven't modified any workflow files.

Issue #5053, opened 2022-10-17
[CAL-173] Booking success query refactor


zomars:
Basically, right now a booking success url looks like this:

```
https://cal.dev/success?date=2022-06-30T02%3A00%3A00-06%3A00&type=3&eventSlug=30min&user=pro&reschedule=false&name=zomars&email=zomars%40cal.com&location=&eventName=&bookingId=14&isSuccessBookingPage=true
```

Since this a server side rendered page, we need to accomplish two things:

1. Simplify the URL to only need the bookings UID as a parameter `https://cal.dev/success?uid={bookingUID}`
   * This will prevent security leaks like previously since you need to know the UID before hand
   * From UID we should be able to query all needed booking data
2. we should try to migrate this page from SSR to SSG by having a a pre-rendered empty Skeleton and fetches data on client only, this will allow us to get the benefits of SSR but with extra performance benefits. Also we can have more control on when do want to return the booking data and when not. (Maybe we don’t want to render success pages from certain passed time or the booking got rescheduled)

> From [Linear-GitHub Sync](https://synclinear.com)

G3root:
@zomars shall i give it a try ?

mischarouleaux:
Hi Everyone! @zomars @PeerRich @CarinaWolli 

I'm a friend of @JeroenReumkens and he introduced me to your awesome platform! So I wanted to help you guys a bit out!

I think I have a solution for part of this issue. I have rewritten the code to work with an UID of a booking in the query.
So I haven't migrated it from SSR to SSG.
This should resolve the security issues with this page. As I was able to see bookings without being logged in to the system, with this change it is only possible if you know the UID.
I can see the same issue on the cancel success, but I see already a mention of this in [CAL-184]!

I'm now in the stage of finalising the code (doing tests etc.), I hope to have a PR out at the end of the day. But I would like to have some extra eyes on testing as there are changes in the payment part!

PeerRich:
@mischarouleaux great to have you!

PeerRich:
@mischarouleaux any update here?

PR #5298, opened 2022-10-31
Booking success query refactor

mischarouleaux:
## What does this PR do?

Previously the query used for the success page was quite long. All the parameters were provided in the url instead of being taken from the db. This results that somebody can see the actual booking without being logged in. So there was no check at all if the booking could be seen by the user.

This PR fixes this by using the UID of a booking in the query and fetching data needed from the db. This makes it harder for people to see an unauthorised booking!

There are 3 components which had an indirect connection to the success page:
1. BookingListItem.tsx
2. BookingPage.tsx
3. Payment.tsx

There are still 2 parameters left in the query and I don't think we can take them out as they are not stored in the db:
1. isSuccessBookingPage, I believe used on the create booking page as an indicator that it's a new booking.
2. listingStatus --> This is not a value saved on a booking, but seems to be determined by some logic in the trpc viewer. The only way to take this value out of the url, is to place logic for this out of the trpc code. I also don’t believe the succes.tsx is a good place to store this kind of logic as it should be generic —> viewer.tsx —> query(”bookings”).

Furthermore, I can see that cancel success page has the same issue. This can be resolved in a similar way as this PR.

Example of an old url: `http://localhost:3000/success?date=2022-11-01T08%3A30%3A00.000Z&type=10&eventSlug=yoga-class&username=pro&name=Pro+Example&email=pro%40example.com&location=&eventName=&bookingId=45&recur=eW9nYS1jbGFzcw%3D%3D&reschedule=true&listingStatus=upcoming&status=ACCEPTED`
Example of an url with this change: `http://localhost:3000/success?uid=v119fisvhrPYo6m5S1uzKC&listingStatus=upcoming`

Screenshot 2022-10-31 at 17 08 23
Screenshot 2022-10-31 at 17 11 15

Fixes part of #5053 
This only fixes the query and doesn't change the SSR to SSG.

## Type of change

## How should this be tested?

This PR should have multiple eyes and tests as it touches the payment process, as in it adds the UID as parameter to the push to success.tsx. 

- [x] Create a new booking and check if the success page is correct.
- [x] Reschedule a booking and check if the success page is correct.
- [ ] Create a new booking using a payment and check if the success page is correct.
- [x] Open a upcoming/past etc. booking and check if the success page is correct.
- [x] Make sure the following components are tested: BookingListItem.tsx, BookingPage.tsx and Payment.tsx

## Checklist

- I have checked the unit tests, and 3 seem to be failing. But it doesn't seem to me this is caused by the new code, but more the lack of some keys. Please double check the unit tests are working with a correct working env.

## Remarks
- The cancel success page has the same issue as in this PR. Please make sure that this resolved in another ticket!
- The booking UUID in the db are not always according the UUID format. I wanted to add a UUID check on the success.tsx page, but this was not possible because of this. An example of this is: "qRNBz6LFS6LeXA9KE32BJb" as UID. This can be by design!


JeroenReumkens:
@mischarouleaux  Congrats on your first PR 👏  Awesome work! We'll review it soon.

apps/web/pages/success.tsx
```diff
@@ -763,42 +761,62 @@ const getEventTypesFromDB = async (id: number) => {
...
+  });
+
+  if (!bookingInfo || !bookingInfo.eventType?.slug)
+    return {
+      notFound: true,
+    };
+
+  // @NOTE: had to do this because Server side cant return [Object objects]
+  // probably fixable with json.stringify -> json.parse
+  bookingInfo["startTime"] = (bookingInfo?.startTime as Date)?.toISOString() as unknown as Date;
```
JeroenReumkens, line 815: Correct! Anything returned in getserversideprops should be serializable 👍  Perhaps towards the future we should convert this into something reusable, since I see we're actually doing this conversion (including the comment) in multiple places.

JeroenReumkens: I actually see that when you create custom events and try to book those, the `periodeStartDate` is not null and will throw a similar error. I think unfortunately we have to convert the `periodStartDate` and `periodEndDate` in the `eventType` object as well. Could you check and change this? 🙏  

CleanShot 2022-11-01 at 08 47 25@2x

JeroenReumkens:
@mischarouleaux  I tested it locally and it works perfect!  Thank you so much for this contribution 🙏  
Only thing I didn't test yet is the payment page for enterprise users. @hariombalhara  Do you know what's the best way to test this?

hariombalhara:
> @mischarouleaux I tested it locally and it works perfect! Thank you so much for this contribution 🙏 Only thing I didn't test yet is the payment page for enterprise users. @hariombalhara Do you know what's the best way to test this?
> 
> That test is the only thing pending for an approval from my side.

The easiest way to test is supposed to actually hardcode the values required by payment page to render. Considering the change is in fetching those values itself, it won't make sense to use that approach.

The ideal way is to setup Stripe and actually create an event with payment enabled. That is a bit cumbersome. If you want to set that up, you can follow this doc https://app.gitbook.com/o/6snd8PyPYMhg0wUw6CeQ/s/yln2maO5KoPmX2Iy6dxo/self-hosting/install-apps/stripe#setting-up-stripe

I think I should have that setup already(did it sometime back). I will be happy to  test the payment page if you want


Also, I have a question here didn't review the PR, but have we tested old success and payment URLs? They should also keep on working. cc @mischarouleaux

JeroenReumkens:
@hariombalhara If it's very quick for you to test locally, it would be awesome if you could 🙏 
As far as the old urls go — I don't think we send out emails with the success page in it, right? 🤔  I believe only way to get back there is via the booking-page in Cal itself, and those links all work.

hariombalhara:
That's a valid point @JeroenReumkens. I am not too sure about success. AFAIK you are right about the success page.
But I see payment page link being sent in emails.
Screenshot 2022-11-01 at 1 52 36 PM

Also, if we keep getting the old links working then it defeats the purpose anyway. So, I think it might be okay not to support old links. But we should test what happens with old links, they shouldn't error out.

hariombalhara:
@mischarouleaux Payment page doesn't seem to get bookingUid param
See this loom where I try to book an event with Payment enabled https://www.loom.com/share/41bfbe99a72541d0b616f22869af7577
cc @JeroenReumkens

mischarouleaux:
@hariombalhara I saw your video, but that url is not changed. But what should be working is if go to that payment page and click "pay now". It should go to the success page with the correct UID. Because I did change the payment component to make use of that bookingUid! The bookingUid in that component comes from a prop inside of that component and not for the query in the url. Could you please double check that?

Screenshot 2022-11-01 at 10 12 24

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -47,7 +48,7 @@ type States =
 export default function PaymentComponent(props: Props) {
   const { t, i18n } = useLocale();
   const router = useRouter();
-  const { email, name, date } = router.query;
+  const { bookingUid } = router.query;
```
hariombalhara, line 51: `bookingUid` is being read from router.query here.
See this loom where I try to make the payment and reach a page with empty uid

mischarouleaux: Sorry I misunderstood your previous comment. I added a new commit where the bookingUid comes from the props instead of the router.query. 

I'm not able to test this locally as I'm missing a key, will look into that a bit later. Does this fix your issue? @hariombalhara

hariombalhara:
Looks good overall to me. Waiting for jobs to get fixed.

JeroenReumkens (approved):

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -136,16 +136,7 @@ const BookingPage = ({
...
-          type: eventType.id,
-          eventSlug: eventType.slug,
-          username: profile.slug,
-          reschedule: !!rescheduleUid,
-          name: bookingForm.getValues("name"),
-          email: bookingForm.getValues("email"),
-          location: responseData.location,
-          eventName: profile.eventName || "",
-          bookingId: id,
+          uid,
```
CarinaWolli, line 139: Same needs to be done for the recurringMutation a few lines below. We are now getting 'page not found' when booking a recurring meeting

CarinaWolli (changes requested):

apps/web/components/booking/pages/BookingPage.tsx
mischarouleaux: Hi @CarinaWolli! Thank you for the review and good catch. I have added an extra commit for the recurringMutation to also use the UID. Could you recheck that page?


@JeroenReumkens I also noticed you added two properties (periodStartDate and periodEndDate) but were not in the db select, so I added them as well

CarinaWolli (approved):
Great work @mischarouleaux 👏🏻

zomars:
Double checking the checks locally...

zomars:
Got some rescheduling tests failing locally. Will investigate further 
image

PeerRich (dismissed):
waiting for @zomars review

mischarouleaux:
@zomars I see the unit tests are still failing, is there anything I can support you with? If I run the forked main branch locally, the same tests seem to be failing!

@alannnc There is a merge conflict. I just wanted to elaborate it with you a bit more before resolving it. It is about this code in the success.tsx

`if (bookingInfo !== null && email) {
    handleSeatsEventTypeOnBooking(eventType, bookingInfo, email);
  }`
  
  Previously the email property was coming from the url query, that has been changed in this PR. Could you maybe check which email property to use now, as example: `bookingInfo.user.email`?

alannnc:
@mischarouleaux Yes, I think right now for booking with seats the only identifier we have is that email in the URL. If we can omit email for other booking without seats we should be okay, but leave it as required if the booking has seats enabled.

CarinaWolli:
@mischarouleaux I fixed the merge conflicts and added the email to the URL query

hariombalhara:
@JeroenReumkens Payment flow works well.

JeroenReumkens:
I've approved the deployment. As soon as that gives a green check I think we're good to merge 🔥

JeroenReumkens (approved):

PR #5299, opened 2022-10-31
Fix avatar infinite redirect

### LINKED ISSUE #5213: [CAL-252] Avatar icon not loading properly

hariombalhara:
- Don't set avatar to dynamic avatar endpoint in session because that would endup overriding user avatar 
- Avoid infinite redirect. If avatar is set to the dynamic avatar route, consider serving the default route.

Fixes #5213

Steps to replicate the issue
- Simply, go to the settings profile page and save without changing the avatar.
	- 	Because getUserFromSession now returns /avatar.png endpoint, it get's saved to DB.


PeerRich (approved):

apps/web/pages/api/user/avatar.ts
```diff
@@ -26,6 +28,9 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       email: user?.email,
       avatar: user?.avatar,
     };
+    linksToThisRoute =
```
hariombalhara, line 31: This is to fix existing avatar's that are doing infinite redirects(because of the bug due to `getUserFromSession` and makes the logic more stable.

hariombalhara (commented):
Self review done

apps/web/pages/api/user/avatar.ts
```diff
@@ -40,14 +45,18 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
   }
 
   const emailMd5 = crypto
     .createHash("md5")
     .update((identity?.email as string) || "guest@example.com")
     .digest("hex");
   const img = identity?.avatar;
-  if (!img) {
+  // If image isn't set or links to this route itself, use default avatar
+  if (!img || linksToThisRoute) {
```
harshsinghatz, line 59: @hariombalhara Can we not write this as 
`
if( !linksToThisRoute)
`
because in the case where username/teamname are falsy values, linksToThisRoute will be false and also while setting the identity we are indirectly checking for img

apps/web/pages/api/user/avatar.ts
harshsinghatz: @hariombalhara what exactly was the bug from getUserFromSession?

packages/trpc/server/createContext.ts
```diff
@@ -87,7 +87,8 @@ async function getUserFromSession({
     return null;
   }
   // This helps to prevent reaching the 4MB payload limit by avoiding base64 and instead passing the avatar url
-  if (user.avatar) user.avatar = `${CAL_URL}/${user.username}/avatar.png`;
```
hariombalhara, line 90: This is the big @harshsinghatz

apps/web/pages/api/user/avatar.ts
hariombalhara: It has to execute for `linksToThisRoute` and not  `!linksToThisRoute`

apps/web/pages/api/user/avatar.ts
```diff
@@ -11,6 +12,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   const username = req.query.username as string;
   const teamname = req.query.teamname as string;
   let identity;
+  let linksToThisRoute = false;
```
alannnc, line 15:
```suggestion
  let isAvatarSourceWebAppUrl = false;
```

apps/web/pages/api/user/avatar.ts
```diff
@@ -26,6 +28,9 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       email: user?.email,
       avatar: user?.avatar,
     };
+    linksToThisRoute =
+      identity.avatar === `${CAL_URL}/${username}/avatar.png` ||
+      identity.avatar === `${WEBAPP_URL}/${username}/avatar.png`;
```
alannnc, line 33: CAL_URL its WEBAPP_URL at the end, so we should only look for the second statement right?
image

alannnc (dismissed):
Couldn't replicate from instructions, some NIT changes.

apps/web/pages/api/user/avatar.ts
hariombalhara: Nope. It is WEBAPP_URL for previews only(which ends in .vercel.app). For the production app it is WEBSITE_URL.

Also, considering that we use WEBAPP_URL and WEBSITE_URL interchangeably many times, it is safer to consider both endpoints

apps/web/pages/api/user/avatar.ts
hariombalhara: This might not be a better variable name considering it can be WEBSITE_URL

packages/trpc/server/createContext.ts
zomars: @hariombalhara does this change returns the whole base64 avatar in the response? If it does we need to address this ASAP.

hariombalhara: Only viewer.me endpoint is having the size issue now. Opened a PR to fix that.
https://github.com/calcom/cal.com/pull/5392

zomars:
The problem with this is that we're exceeding the payload limits due we're returning the whole image as codified data. We should query another endpoint for avatars specifically.

PR #5304, opened 2022-10-31
fix: use matcher in middleware

Udit-takkar:
To not execute middleware on every request https://threads.com/34432119721?s=tHyYikDxFkSDz1W7SfGdxC


apps/web/middleware.ts
```diff
@@ -39,6 +39,10 @@ const middleware: NextMiddleware = async (req) => {
   return NextResponse.next();
 };
 
+export const config = {
+  matcher: ["/api/collect-events/:path*", "/api/auth/:path*", "/apps/routing_forms/:path*", "/:path*/embed"],
```
Udit-takkar, line 43: where is /embed path used for?

Udit-takkar: cc: @hariombalhara

hariombalhara: @Udit-takkar Does it fix the issue for you? I will also test locally.

hariombalhara: Screenshot 2022-11-01 at 5 14 37 PM

Udit-takkar: yeah i tested locally . the middleware only executed for these paths. just test the embed url once

hariombalhara: I could replicate locally that in main, updateProfile request is stuck and in this branch it immediately completes

hariombalhara: @Udit-takkar Right now it's possible to be in a situation where someone adds logic in middleware fn to do something for a URL, but the logic never executes. This is because there is a separate array governing which URLs can invoke middleware.
@emrysal  and  I were thinking how to solve this. Can you try something like this and see if it solves the problem and still let us use middleware the current way. 
image

hariombalhara: @emrysal shared and I agree that this isn't something that we need right now and makes things complex. So, we can skip it for now. 
Just ignore the last comment @Udit-takkar

hariombalhara (approved):

PR #5311, opened 2022-11-01
Fix profile page data reset on re-focus of window

### LINKED ISSUE #4783: [CAL-266] Cannot update email from settings

### Issue Summary

When I attempt to update my email on the [https://app.cal.com/settings](https://app.cal.com/settings) page, the network request sent to the `viewer.updateProfile` endpoint does not contain the new URL, so the value does not change. I was able to manually resend the request with the updated email, which appears to have been successful.

### Steps to Reproduce

1. Log in and go to the settings page
2. Type in a new email, click update, type in password to confirm
3. See that the email does not update

### Technical details

* Browser version: Firefox 105.0.1

hariombalhara:
## What does this PR do?

Fixes #4783
See before and after in this loom https://www.loom.com/share/cb51ffacdacb41e69ab7abf8fa77fd19

**Environment**: Production

## Type of change

## How should this be tested?

- See loom
## Checklist


apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -158,16 +158,20 @@ const ProfileView = () => {
...
   const { reset } = formMethods;
-
+  const formInitializedRef = useRef(false);
   useEffect(() => {
-    if (user)
+    // The purpose of reset is to set the initial value obtained from tRPC.
+    // `user` would change for many reasons (e.g. when viewer.me automatically fetches on window re-focus(a react query feature))
+    if (user && !formInitializedRef.current) {
+      formInitializedRef.current = true;
       reset({
```
JeroenReumkens, line 167: Do you know why this form uses the `reset` method? Shouldn't we set it via `initialvalues` on the form component? I think that would also solve the issue, right?

hariombalhara: initialValues as in use form defaultValues? 
The problem here is that when useForm hook is called first time, user isn't set.

JeroenReumkens: Oh sorry, I meant `defaultValues` indeed. That also doesn't update when the `user` gets updated? I thought it did actually. 

If that doesn't work indeed, than this change is totally fine by me! 🙌

hariombalhara: Yeah it doesn't update. I hate this about React Hook Forms. Using them with tRPC forces us to use `reset`

Screenshot 2022-11-01 at 1 39 46 PM

JeroenReumkens: Fair! Let's go with this then instead 😄  Perhaps towards the future we could create a custom hook for this that would improve readability. Like `useSetFormValuesBasedOnData` or something like that.

JeroenReumkens (approved):

apps/web/pages/settings/my-account/profile.tsx
hariombalhara: Yeah, exactly what I had in my mind.

PR #5319, opened 2022-11-01
Provide update feedback on event type settings

sean-brydon:
CleanShot 2022-11-01 at 12 40 23

The gif says it all :)


JeroenReumkens (approved):

PeerRich (approved):

hariombalhara:
@sean-brydon Does it show the error as well in case of failure?

PR #5321, opened 2022-11-01
Fixed description by manually implementing asSafeHTML

emrysal:


JeroenReumkens (approved):
@emrysal  Tested locally and works perfect! Seems like a great solution to me until we improve the middleware 🙏

PR #5326, opened 2022-11-01
feature/sentry.io for web

alannnc:
## What does this PR do?

Install sentry.io for web app

## Type of change


apps/web/next.config.js
```diff
@@ -228,6 +229,18 @@ const nextConfig = {
 
     return redirects;
   },
+  sentry: {
+    hideSourceMaps: true,
+  },
+};
```
alannnc, line 235: This shouldn't affect self hosted.

apps/web/next.config.js
```diff
@@ -228,6 +229,18 @@ const nextConfig = {
...
+
+const sentryWebpackPluginOptions = {
+  silent: true, // Suppresses all logs
```
alannnc, line 238: Same as above

apps/web/next.config.js
```diff
@@ -228,6 +229,18 @@ const nextConfig = {
...
 };
 
-module.exports = () => plugins.reduce((acc, next) => next(acc), nextConfig);
+const moduleExports = () => plugins.reduce((acc, next) => next(acc), nextConfig);
+
+// Sentry should be the last thing to export to catch everything right
+module.exports = process.env.NEXT_PUBLIC_SENTRY_DSN
+  ? withSentryConfig(moduleExports, sentryWebpackPluginOptions)
```
alannnc, line 245: If they don't have any sentry env we should fallback to default export.

JeroenReumkens:
LGTM @alannnc !

JeroenReumkens (approved):

hariombalhara (dismissed):
Blocking this till crash problem is solved. 
https://threads.com/34432396033?s=hDF7phTfHe6ecRWvnwWRV6
https://www.loom.com/share/f3fa835c5d7f4695be1121f271f8dfaa

I have tried cleaning up node_modules and re-installation but this error persists. I am worried that this might break infra users build.

alannnc:
> LGTM @alannnc !
> 
> Only wondering if the huge update to yarn.lock is a mistake or not?

Thats what I think too just did a clean and fresh install of everything, probably it's not being updated correctly from others and someone has to take it. Help me out and let me know if you generate the same lock.
@JeroenReumkens

JeroenReumkens:
@alannnc  The only thing if you delete the yarn.lock is that lots of packages get upgraded to a new version, so in theory that could result in stuff being broken because of a package update. I my opinion you usually don't want purge a full yarn lock file, and only update one dependency at a time (which will update that dependency in the lock file as well).

hariombalhara:
@JeroenReumkens I did a checkout of yarn.lock from main and run a `yarn update` again. The file has pretty less changes now.

hariombalhara:
Screenshot 2022-11-03 at 4 58 08 PM
Tested a sample error and it's recorded in Sentry.
merging it now.

PR #5330, opened 2022-11-02
Make SMS workflow actions only available for teams plan

CarinaWolli:
## What does this PR do?

Make SMS workflow actions only available for teams plan. As discussed with @baileypumfleet  and @PeerRich 


packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -515,6 +520,9 @@ export const workflowsRouter = createProtectedRouter()
           });
           //step was edited
         } else if (JSON.stringify(oldStep) !== JSON.stringify(newStep)) {
+          if (user.plan === "FREE" && !isSMSAction(oldStep.action) && isSMSAction(newStep.action)) {
```
CarinaWolli, line 523: So attackers can not add SMS actions to existing steps. It checks if the old action was already a SMS action and if not the new action also can not be a SMS action (for users on the FREE plan).

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -652,6 +660,9 @@ export const workflowsRouter = createProtectedRouter()
       //added steps
       const addedSteps = steps.map((s) => {
         if (s.id <= 0) {
+          if (user.plan === "FREE" && isSMSAction(s.action)) {
```
CarinaWolli, line 663: So attackers can not add a new step with a SMS action as a free user

leog (approved):
After the whole "made all users FREE by mistake" fiasco, I'm not sure if the intended behavior will occur @CarinaWolli. Are we sure we are targeting the correct users taking that into account? Otherwise, the changes look good to me

PR #5349, opened 2022-11-03
fix: add hour12

### LINKED ISSUE #4885: Event type availability does not respect 12h/24h account setting.
On an account with 12-hour time format display

image

Udit-takkar:
Fixes: #4885 

Screenshot 2022-11-03 at 4 22 22 PM


apps/web/components/v2/eventtype/AvailabilityTab.tsx
```diff
@@ -91,15 +92,16 @@ const AvailabilitySelect = ({
...
+  Intl.DateTimeFormat(undefined, { hour: "numeric", minute: "numeric", hour12 }).format(
     new Date(dayjs.utc(date).format("YYYY-MM-DDTHH:mm:ss"))
   );
 
 export const AvailabilityTab = () => {
   const { t, i18n } = useLocale();
   const { watch } = useFormContext<FormValues>();
-
+  const me = useMeQuery();
+  const { timeFormat } = me.data || { timeFormat: null };
```
JeroenReumkens, line 104: Perhaps it's just me, but I don't really like the formatting this way 🤔  
I'd rather see something like `const timeFormat = me?.data?.timeFormat`, which also makes it more obvious to me that this data could be undefined. What do you think?

JeroenReumkens (approved):
Added one NIT, besides that awesome change!

apps/web/components/v2/eventtype/AvailabilityTab.tsx
Udit-takkar: done

PR #5350, opened 2022-11-03
Fixes recurring hard crash in two places to ensure rrule gets valid p…

emrysal:
RecurringInfo could be null in some situations when the event type hasn't always been a recurring event, this caused undefined values passed to rrule - triggering an infinite loop in the library.


alannnc (approved):
Code looks good, couldn't replicate the error tho

PR #5366, opened 2022-11-04
fix/google-calendar-removes-attendees-when-seats-5336

### LINKED ISSUE #5336: [CAL-307] Google calendar connection sends cancellation emails as new seats are booked
When a second seat is booked the first seat receives an email informing them of the event cancellation. This happens when the organizer has a google calendar connected.

This also leaks additional notes of the second seat as part of the cancellation email.

alannnc:
## What does this PR do?

Fixes #5336 

## Type of change

## How should this be tested?

- [x] Update an event type with seats available
- [x] Install google calendar in that event type
- [x] Book that event multiple times
- [x] Users already booked shouldn't been remove from calendar Event thus they won't receive an email about event cancellation.

## Checklist


apps/web/components/v2/eventtype/EventAdvancedTab.tsx
```diff
@@ -329,9 +326,10 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
                     label={t("number_of_seats")}
                     type="number"
                     defaultValue={value || 2}
+                    min={1}
                     addOnSuffix={<>{t("seats")}</>}
                     onChange={(e) => {
-                      onChange(Number(e.target.value));
+                      onChange(Math.abs(Number(e.target.value)));
```
alannnc, line 332: NIT improvement to don't allow negative numbers on input field. (Should validate on form and server too probably)

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -160,6 +160,10 @@ export default class GoogleCalendarService implements Calendar {
   async updateEvent(uid: string, event: CalendarEvent, externalCalendarId: string): Promise<any> {
     return new Promise(async (resolve, reject) => {
       const myGoogleAuth = await this.auth.getToken();
+      const eventAttendees = event.attendees.map(({ id, ...rest }) => ({
+        ...rest,
+        responseStatus: "accepted",
+      }));
```
alannnc, line 166: Just clean up

packages/core/EventManager.ts
```diff
@@ -294,6 +294,13 @@ export default class EventManager {
   }
 
   public async updateCalendarAttendees(event: CalendarEvent, booking: PartialBooking) {
+    const attendees = booking.attendees.map((attendee) => {
+      return {
+        email: attendee.email,
+        name: attendee.name,
+      };
+    });
+    event.attendees = attendees;
```
alannnc, line 303: This was the problem, it looks like code that copies attendees to event was missing here.

packages/core/EventManager.ts
```diff
@@ -430,11 +437,14 @@ export default class EventManager {
     try {
       // Bookings should only have one calendar reference
       calendarReference = booking.references.filter((reference) => reference.type.includes("_calendar"))[0];
-      if (!calendarReference) throw new Error("bookingRef");
-
+      if (!calendarReference) {
+        throw new Error("bookingRef");
+      }
```
alannnc, line 442: Clean up like the rest of the code below this change

packages/core/EventManager.ts
emrysal: Is there any situation where event.attendees would be set? (And this would overwrite that existing value?)

emrysal (changes requested):
We've got to figure out why https://github.com/calcom/cal.com/blob/352ce35b40cc68b3d7830619198e78e58946c259/packages/features/bookings/lib/handleNewBooking.ts#L485 does not work.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -511,8 +517,13 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
...
     const newSeat = booking.attendees.length !== 0;
-
-    await sendScheduledSeatsEmails(evt, invitee[0], newSeat, !!eventType.seatsShowAttendees);
+    /**
+     * Remember objects are passed into functions as references
+     * so if you modify it in a inner function it will be modified in the outer function
+     * deep cloning evt to avoid this
+     */
+    const copyEvent = cloneDeep(evt);
+    await sendScheduledSeatsEmails(copyEvent, invitee[0], newSeat, !!eventType.seatsShowAttendees);
```
alannnc, line 526: @emrysal Turns out this function was modifying attendees due to using the object as reference. Now using a deepCloned object to send the email and updateEvent with the original event.

emrysal (approved):

PR #5369, opened 2022-11-04
[CAL-211] Confirmed booking page for rescheduled event should show previous time and reschedule reason

### LINKED ISSUE #5134: [CAL-211] Confirmed booking page for rescheduled event should show previous time and reschedule reason
[Scheduled - Cal - Live (Figma)](https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/Cal---Live?node-id=68%3A95398)

gitstart:
## What does this PR do?
show previous time and reschedule reason on confirmed booking page

Fixes #5134

**Before any rescheduling**
image

**With reschedule reason**
Screenshot 2022-11-07 at 2 16 33 PM

**Without reschedule reason**
Screenshot 2022-11-07 at 2 16 56 PM

## Type of change

### DIFF
--- a/apps/web/components/booking/pages/BookingPage.tsx
+++ b/apps/web/components/booking/pages/BookingPage.tsx
@@ -143,6 +143,7 @@ const BookingPage = ({
           isSuccessBookingPage: true,
           email: bookingForm.getValues("email"),
           eventTypeSlug: eventType.slug,
+          formerTime: booking?.startTime.toString(),
         },
       });
     },
@@ -158,6 +159,7 @@ const BookingPage = ({
           allRemainingBookings: true,
           email: bookingForm.getValues("email"),
           eventTypeSlug: eventType.slug,
+          formerTime: booking?.startTime.toString(),
         },
       });
     },
--- a/apps/web/pages/booking/[uid].tsx
+++ b/apps/web/pages/booking/[uid].tsx
@@ -168,6 +168,7 @@ const querySchema = z.object({
   changes: stringToBoolean,
   reschedule: stringToBoolean,
   isSuccessBookingPage: z.string().optional(),
+  formerTime: z.string().optional(),
 });
 
 export default function Success(props: SuccessProps) {
@@ -179,6 +180,7 @@ export default function Success(props: SuccessProps) {
     isSuccessBookingPage,
     cancel: isCancellationMode,
     changes,
+    formerTime,
   } = querySchema.parse(router.query);
 
   if ((isCancellationMode || changes) && typeof window !== "undefined") {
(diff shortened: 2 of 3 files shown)


PeerRich:
can you move "reschedule reason" above "What"?

apps/web/pages/success.tsx
```diff
@@ -355,6 +368,12 @@ export default function Success(props: SuccessProps) {
                         is24h={is24h}
                       />
                     </div>
+                    {reschedule && !!rescheduleReason?.length && (
+                      <>
+                        <div className="font-medium">{t("reschedule_reason_success_page")}</div>
+                        <div className="col-span-2 mb-6 last:mb-0">{rescheduleReason}</div>
+                      </>
+                    )}
```
PeerRich, line 376: can you move this above "what"? @gitstart

gitstart: sure thing

gitstart:
> can you move "reschedule reason" above "What"?

Hey @PeerRich it's done, please take a look

PeerRich (approved):

gitstart:
Hey @PeerRich, can we get this PR merged?

JeroenReumkens (dismissed):
Hi @gitstart 

Thanks for this change, I really like it 🙏  
In the mean time the success page actually changed a bit, we're not getting all the details from the URL anymore, rather we're getting a UUID and fetching all the data via the API. Are you able to rework this PR so it is in line with the recent change? 

Thanks a lot!

zomars:
> Hi @gitstart
> 
> Thanks for this change, I really like it 🙏 In the mean time the success page actually changed a bit, we're not getting all the details from the URL anymore, rather we're getting a UUID and fetching all the data via the API. Are you able to rework this PR so it is in line with the recent change?
> 
> Thanks a lot!

Also adding to this this might be soon change to `/booking/uid`

gitstart:
> Hi @gitstart
> 
> Thanks for this change, I really like it 🙏 In the mean time the success page actually changed a bit, we're not getting all the details from the URL anymore, rather we're getting a UUID and fetching all the data via the API. Are you able to rework this PR so it is in line with the recent change?
> 
> Thanks a lot!

Sure, we can. Will get back to you with the requested changes

gitstart:
> Also adding to this this might be soon change to `/booking/uid`

@zomars Do you suggest we work on the PR after this change is done?

PeerRich:
> > Also adding to this this might be soon change to `/booking/uid`
> 
> @zomars Do you suggest we work on the PR after this change is done?

we're investigating now!

gitstart:
@JeroenReumkens PR has been updated, please take a look

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -119,6 +119,10 @@ const BookingPage = ({
         collectPageParameters("/book", { isTeamBooking: document.URL.includes("team/") })
       );
     }
+    localStorage.removeItem("formerTime");
```
JeroenReumkens, line 122: I am thinking whether it makes sense to still pass the formerTime via a query parameter. I think the localstorage solution is a bit prone to breaking, because if you're able to book an event without this component, the storage wouldn't get cleared, and you would see an incorrect/old "before" time. Perhaps as of right now that's not possible, but in the future it might, and you need to keep the fact that this localstorage items needs to be cleared in your mental model to think about it when building new functionality, in order to not break this. 

If we would turn this into a query param however, you can also forget to pass this param, but in that case it's not broken since we still show the correct new date and time, only the previous date isn't shown. So I would suggest changing this to a query param to be more future proof 🙏

JeroenReumkens (dismissed):
Thanks @gitstart! The changes look really nice! 🔥  I've requested one change though — instead of using localstorage resorting to a query parameter, since I think this will be more resilient towards the future. Would you be able and have the time to implement that? 

Thanks again for your contributions, greatly appreciated!

gitstart:
Hi @JeroenReumkens! Implemented the query parameter now, please take a look
https://www.loom.com/share/2394c06aea4f46299d5f5348230d083f

zomars (approved):
Good enough to merge? @JeroenReumkens

PR #5371, opened 2022-11-04
Hotfix:Fixes zoom tokens not being refreshed properly

### LINKED ISSUE #5367: [CAL-316] Zoom link not generated in the confirmation
It falls back to calvideo even after disconnecting and installing the zoom app again

Created via Threads. See full discussion: [https://threads.com/34432760520](https://threads.com/34432760520)

hariombalhara:
Make sure body being jsoned is fresh
Fixes #5367

As, we were refreshing the token from Zoom's side but not able to save in our DB. People who faced this issue would have to do a re-installation once more.

@alannnc I think we should update the token in our DB before erroring out.


packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -339,7 +340,7 @@ const handleZoomResponse = async (response: Response, credentialId: Credential["
     throw Error(response.statusText);
   }
 
-  return response.json();
+  return responseClone.json();
```
hariombalhara, line 343: This is a dirty hack to create a new clone of the response, acceptable as a Hotfix.
Ideally, we need to send _response. But _response is sometimes of type Response and sometimes a POJO it seems.

alannnc:
> I think we should update the token in our DB before erroring out.

My take too here but I think it was discussed at the time if we didn't received what we wanted we wouldn't save anything. And throw error.

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
zomars: Can we mention in the code comments why this hack is needed?

alannnc (approved):

alannnc:
This works in the meanwhile we investigate further.

PR #5372, opened 2022-11-04
fix: callback url

Udit-takkar:
## What does this PR do?

Just after signing up with email and password and logging in you are directed to 404.

Fixes # (issue):- https://calendso.slack.com/archives/C01NXDSRY9W/p1667485320197439?thread_ts=1667480237.056219&cid=C01NXDSRY9W

Video :- 

https://user-images.githubusercontent.com/53316345/199982027-2a40ee01-3df4-4a51-a5c5-7b1b2fb7d46c.mov

### DIFF
--- a/apps/web/pages/signup.tsx
+++ b/apps/web/pages/signup.tsx
@@ -54,12 +54,11 @@ export default function Signup({ prepopulateFormValues }: inferSSRProps<typeof g
       method: "POST",
     })
       .then(handleErrors)
-      .then(
-        async () =>
-          await signIn("Cal.com", {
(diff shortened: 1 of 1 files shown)


apps/web/pages/signup.tsx
```diff
@@ -143,7 +142,7 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   // no token given, treat as a normal signup without verification token
   if (!token) {
     return {
-      props,
+      props: JSON.parse(JSON.stringify(props)),
```
Udit-takkar, line 145: It was throwing error getServerSideProps cannot be serialized as JSON when we return only props.

emrysal (approved):

PR #5390, opened 2022-11-05
fix/success-page-seats-description-5338

### LINKED ISSUE #5338: [CAL-308] Success page always shows the additional notes of the first seat
When booking the second (or more) seat of an event the additional notes of the first seat is shown, this happens also when **Hide notes in calendar** is selected and has the potential to leak confidential information of the first seat member.

alannnc:
## What does this PR do?

- Removed description from bookingInfo if eventType has seats. This is due we don't save every description input on the DB for every user with a seat.
- Added server side validation that if booking has seats we don't leak any user's info to the client if seatsShowAttendees is false;

Fixes #5338 

## Type of change

## How should this be tested?

- [ ] Add seats to booking and turn off/on Show attendees list
- [ ] While booking success make sure any description is shown and proper user list should be displayed if showAttendees its true|false.

## Checklist

### DIFF
--- a/apps/web/pages/success.tsx
+++ b/apps/web/pages/success.tsx
@@ -788,7 +788,7 @@ const schema = z.object({
 
 const handleSeatsEventTypeOnBooking = (
   eventType: {
-    seatsPerTimeSlot: boolean | null;
+    seatsPerTimeSlot?: boolean | null;
     seatsShowAttendees: boolean | null;
     [x: string | number | symbol]: unknown;
   },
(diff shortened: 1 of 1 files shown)


om-ray (approved):
Looks good to me! Are we ever planning to include additional notes from every user?

emrysal:
Made a small amend to remove the filter from the frontend, no longer needed - great idea pulling the logic into the backend @alannnc .

emrysal (approved):

PR #5398, opened 2022-11-07
Fail lint check on error

emrysal:
It seems like yarn build is no longer part of yarn e2e - causing automerges with failed lint config. Then on yarn build an error is thrown and the build fails.

This PR fixes all outstanding lint errors & fails the lint check on error.


.github/workflows/lint.yml
```diff
@@ -32,7 +32,6 @@ jobs:
 
       - name: Lint
         run: yarn lint:report
-        continue-on-error: true
```
zomars, line 35: I don't think this is an acceptable solution. We need this to actually upload the linting report to the GitHub annotations.

zomars (dismissed):

.github/workflows/lint.yml
zomars: Maybe. But this was working before. What changed?

emrysal: Maybe it wasn't working - e2e always failed due to build errors which may have made this check redundant all along until now? Is there any reason for `yarn lint:report` over `yarn lint` - looking at the implementation yarn lint just dumps everything into stdout.

emrysal: Confirmed it was broken at least since Sep 8 - https://github.com/calcom/cal.com/actions/runs/3370385128/jobs/5592719743

emrysal:
@zomars Did further research and according to https://github.com/ataylorme/eslint-annotate-action you indeed need to run both.

zomars (approved):
Nice find @emrysal 

LGTM

PR #5400, opened 2022-11-07
Migrates V2 Select to new ui/components

sean-brydon:
This components moves select V2 into the new `ui/components` folder and also creates documentation behind it.
CleanShot 2022-11-07 at 13 57 38

WIP flag on variants is due to react select not handling IconStart like expected. Needs more investigation outside this PR


packages/ui/components/form/select/select.stories.mdx
```diff
@@ -0,0 +1,115 @@
...
+  </Example>
+  <Example title="Focus">
+    <SelectField options={options} className="sb-pseudo--focus"/>
+  </Example> */}
+</Examples>
+
+
+## Select Story
+<Canvas>
+  <Story name="Default">
```
JeroenReumkens, line 112: Are we able to still use the variants table here? Makes the layout just a bit nicer and adds the dark mode demo as well 👍

JeroenReumkens (dismissed):
Looks great! Added one small remark to the story!

PeerRich (approved):

packages/ui/components/form/select/select.stories.mdx
PeerRich: is this blocking or can this be merged @sean-brydon @JeroenReumkens

sean-brydon: I think this should be good to merge - will solve merge conflicts.

PR #5405, opened 2022-11-07
Using RHF to decrease state logic in profile.tsx for UsernameAvailabi…

emrysal:
## What does this PR do?

Removes username related state in favour of the component / user?.username


sean-brydon (approved):
LGTM :) Works as expected

PR #5406, opened 2022-11-07
Salesforce app

### LINKED ISSUE #1545: [CAL-84] Salesforce App
Send cal data to salesforce

leog:
## What does this PR do?

Salesforce app to add new contacts and events from bookings. 

Also, this PR:
 - Renames HubSpot and Close.com apps to clean up URLs, which is why there is an additional task to take care of those renamed apps
 - Introduces a way to surfice errors and warnings from installed apps runs into the Apps Status in the organizer email. And to avoid surfacing errors that could not be as descriptive or updated as we want, only errors within a `calError` property will be surfaced. Warnings are totally new and also require a `calWarning` property to be  defined.

### Pending tasks 
 - [ ] Go into HubSpot Developer Account holding the Cal.com app and modify part of the Redirect URI in it from "hubspotothercalendar" to "hubspot"
 - [x] Create Salesforce app for Staging and for Production copying the one created for Dev, credentials for team@cal.com in Bitwarden
 - [x] Add instructions to the README to create a Salesforce app for this Cal app to work

Loom Video: https://www.loom.com/share/f0d65ef09a5b474d85a7c62343bf9b9f

## Type of change

## How should this be tested?

Install Salesforce app, receive a booking, and see the attendee(s) be created in Salesforce as contacts with an event representing the booking.


.env.appStore.example
```diff
@@ -11,6 +11,10 @@
 #   - ZOOM
 #   - GIPHY
 #   - VITAL
+#   - ZAPIER
+#   - LARK
+#   - WEB3
+#   - SALESFORCE
```
leog, line 17: Missing apps in the index added

apps/web/components/apps/App.tsx
```diff
@@ -43,8 +43,10 @@ const Component = ({
   const router = useRouter();
 
   const mutation = useAddAppMutation(null, {
-    onSuccess: () => {
-      showToast(t("app_successfully_installed"), "success");
+    onSuccess: (data: any) => {
+      if (data.showToast !== undefined && data.showToast) {
```
leog, line 47: This is to fix the false positive "App successfully installed" toast message
cc: @hariombalhara

apps/web/components/apps/OmniInstallAppButton.tsx
```diff
@@ -18,11 +18,13 @@ export default function OmniInstallAppButton({ appId, className }: { appId: stri
   const utils = trpc.useContext();
 
   const mutation = useAddAppMutation(null, {
-    onSuccess: () => {
+    onSuccess: (data: any) => {
       //TODO: viewer.appById might be replaced with viewer.apps so that a single query needs to be invalidated.
       utils.invalidateQueries(["viewer.appById", { appId }]);
       utils.invalidateQueries(["viewer.apps", { extendsFeature: "EventType" }]);
-      showToast(t("app_successfully_installed"), "success");
+      if (data.showToast !== undefined && data.showToast) {
```
leog, line 25: This is to fix the false positive "App successfully installed" toast message
cc: @hariombalhara

packages/ui/v2/core/apps/AppCard.tsx
```diff
@@ -16,8 +16,10 @@ interface AppCardProps {
 export default function AppCard({ app, credentials }: AppCardProps) {
   const { t } = useLocale();
   const mutation = useAddAppMutation(null, {
-    onSuccess: () => {
-      showToast(t("app_successfully_installed"), "success");
+    onSuccess: (data: any) => {
+      if (data.showToast !== undefined && data.showToast) {
```
leog, line 20: This is to fix the false positive "App successfully installed" toast message
cc: @hariombalhara

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -40,16 +40,18 @@ function useAddAppMutation(_type: App["type"] | null, options?: Parameters<typeo
 
     const json = await res.json();
 
+    const externalUrl = !json.url.startsWith(window.location.origin);
+
     if (!isOmniInstall) {
       window.location.href = json.url;
-      return;
+      return { showToast: !externalUrl && !json.url.endsWith("/setup") };
```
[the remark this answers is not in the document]
hariombalhara: So, we are assuming here that if /setup redirection is happening, then there is something more required to do before successful installation.
This is true for Sendgrid but not for other apps like Zapier. In Zapier's case, the app is installed immediately and the user needs to configure it inside Zapier.

But I like the idea of being able to do it automatically without explicit configuration by each app. 👍👍

Also, I think it would be better to rename `showToast` to something more generic because we can choose the characteristic to do more things.
e.g. `showToast` -> Tells that you can show the toast
If we rename it to something like `setupPending`, we can use it show a message on the redirected page as well that 'Your installation is pending till this setup is complete.'

apps/web/components/apps/OmniInstallAppButton.tsx
hariombalhara: I think we can remove `data.showToast !== undefined`

apps/web/components/apps/IntegrationListItem.tsx
```diff
@@ -27,6 +28,10 @@ function IntegrationListItem(props: {
   const [highlight, setHighlight] = useState(hl === props.slug);
   const title = props.name || props.title;
 
+  if (highlight) {
+    showToast(t("app_successfully_installed"), "success");
```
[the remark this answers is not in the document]
hariombalhara: I think a comment would be appropriate in the code as to why highlight is deciding if toast would be shown or not.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -40,16 +40,18 @@ function useAddAppMutation(_type: App["type"] | null, options?: Parameters<typeo
 
     const json = await res.json();
 
+    const externalUrl = !json.url.startsWith(window.location.origin);
```
hariombalhara, line 43: `json.url` can be a relative path like '/setup' and in that case it would be considered external URL wrongly.
We need to copy `/https?:\/\//.test(json.url)` logic here as well.

PeerRich:
awesome job! funny to hear tennis in your loom video too.🤣

PeerRich:
do we have screenshots for the detail page?

leog:
> awesome job! funny to hear tennis in your loom video too.🤣

@PeerRich Oh shoot! Didn't know it took my connected iPad where I see it as an audio input, sorry 😅

> do we have screenshots for the detail page?

I can add for sure, which one? Salesforce app details page you mean?

packages/app-store/_utils/getCalendar.ts
```diff
@@ -9,8 +9,8 @@ const log = logger.getChildLogger({ prefix: ["CalendarManager"] });
 export const getCalendar = (credential: CredentialPayload | null): Calendar | null => {
   if (!credential || !credential.key) return null;
   let { type: calendarType } = credential;
-  if (calendarType === "sendgrid_other_calendar") {
-    calendarType = "sendgrid";
+  if (calendarType?.endsWith("_other_calendar")) {
+    calendarType = calendarType.split("_other_calendar")[0];
```
leog, line 13: @hariombalhara making this as a general rule instead for specific apps as I renamed all other `_other_calendar` apps within this PR.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -36,8 +36,8 @@ function useAddAppMutation(_type: App["type"] | null, options?: UseAddAppMutatio
       isOmniInstall = variables.isOmniInstall;
       type = variables.type;
     }
-    if (type === "sendgrid_other_calendar") {
-      type = "sendgrid";
+    if (type?.endsWith("_other_calendar")) {
+      type = type.split("_other_calendar")[0];
```
leog, line 40: @hariombalhara making this as a general rule instead for specific apps as I renamed all other `_other_calendar` apps within this PR.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -73,7 +73,7 @@ function useAddAppMutation(_type: App["type"] | null, options?: UseAddAppMutatio
       gotoUrl(json.url, json.newTab);
     }
 
-    return { setupPending: !externalUrl && json.url.endsWith("/setup") };
+    return { setupPending: externalUrl || json.url.endsWith("/setup") };
```
leog, line 76: This somehow was wrong, corrected it. We want to let the UI know there is setup pending when a redirection is going to be done to an external URI or the URL to redirect URL ends with "/setup", which means there is an internal setup page pending to be filled in order to consider the app successfully installed.

leog:
@PeerRich here you go:
<kbd>image</kbd>

zomars (changes requested):
Nice work @leog I'll block since this is a big change and this is not a "Salesforce app" only PR. We should try to make all the prep work in one PR and make the Salesforce PR into another so we can revert safely if needed.

PR #5407, opened 2022-11-07
Add data?.isDefault to disabled={} prop

### LINKED ISSUE #5397: [CAL-325] Make it impossible to uncheck a default availability
Currently at the top right of an availability you can 'uncheck' the 'set to default' of an availability. When the user does this, it won't work, because there's not other default selected, so we won't update anything. 

Goal: 

In this issue we want to NOT show the toggle when the current availability is the default. When it is **not** the default the toggle should be visible, and you should be able to make this availability the default by checking the toggle. 

Note: In this design by Ciaran the toggle is moved as well, we're not implementing that as part of this story yet.

emrysal:
Fixes #5397


sean-brydon (approved):
LGTM - should save a few headaches :D

PR #5408, opened 2022-11-07
Cleanup + migrate away from process.browser

### LINKED ISSUE #5360: process.browser is deprecated and browser.utils.ts is duplicated
`process.browser` is deprecated
There are two references in the repo.
image

https://github.com/calcom/cal.com/blob/436f88a48e0de81a0bcbd33a32c113bd093904d5/packages/lib/browser/browser.utils.ts#L10

```ts
  if (typeof window === "undefined") {
```

it seems to be a duplication between browser.info.ts
image

[packages/lib/browser/browser.utils.ts](https://github.com/calcom/cal.com/blob/179a9b7139f2a1117bab896a3330a220fae405fb/packages/lib/browser/browser.utils.ts)
[apps/web/lib/core/browser/browser.utils.ts](https://github.com/calcom/cal.com/blob/855aa56e7766e37773c07e8e77924fe144c171f5/apps/web/lib/core/browser/browser.utils.ts)

emrysal:
## What does this PR do?

- Fixes #5360
- Cleans up dead files


packages/lib/browser/browser.utils.ts
```diff
@@ -7,7 +7,7 @@ type BrowserInfo = {
 };
 
 export const getBrowserInfo = (): Partial<BrowserInfo> => {
-  if (!process.browser) {
+  if (typeof window === "undefined") {
```
JeroenReumkens, line 10: Perhaps it might make sense to add a small util `isBrowser()`? Makes that we have this logic only in one place, and also reads a bit easier. The change from NOT to now strict equals almost threw me off thinking it was reverted 😁

JeroenReumkens (approved):

PR #5409, opened 2022-11-08
Sendgrid App fixes

hariombalhara:
## What does this PR do?

Fixes
- Changed dir name to `sendgrid` which matches App URL and is cleaner.
- Wasn't showing app in app-store for installation. Reason being expectation of `sendgrid` dirName
- Sendgrid logo wasn't loading. Fixed that
- Fixes wrong 'Installation successful' message 
- Removed an unused duplicate App.tsx component.
- Remove duplicate seeding.


packages/app-store/sendgrid/config.json
```diff
@@ -3,8 +3,8 @@
   "name": "Sendgrid",
   "slug": "sendgrid",
   "type": "sendgrid_other_calendar",
-  "imageSrc": "/api/app-store/sendgridothercalendar/icon.png",
-  "logo": "/api/app-store/sendgridothercalendar/icon.png",
```
hariombalhara, line 7: Didn't see file with icon.png. Instead it had logo.png

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -20,6 +28,13 @@ function useAddAppMutation(_type: App["type"] | null, options?: Parameters<typeo
       isOmniInstall = variables.isOmniInstall;
       type = variables.type;
     }
+
+    //Type _other_calendar apps aren't supported yet with CLI properly.
+    // If we change their type to the same as slug the app logic that depends on checking the presence of _calendar breaks and if we keep the type as is, it is different from slug.
+    // TODO: Work on removing dependency of having type contain _calendar
```
leog, line 34: There are more than a few apps with this characteristic. I would prefer to tackle them together in a future fix than having to tackle just one introducing logic to handle just one exception. Thoughts?

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -42,17 +56,20 @@ function useAddAppMutation(_type: App["type"] | null, options?: Parameters<typeo
...
-      window.location.href = json.url;
+      // Check first that the URL is absolute, then check that it is of different origin from the current.
+      if (/https?:\/\//.test(json.url) && !json.url.startsWith(window.location.origin)) {
+        // TODO: For Omni installation to authenticate and come back to the page where installation was initiated, some changes need to be done in all apps' add callbacks
+        window.location.href = json.url;
+      }
     }
+
+    return {
+      setupRequiredForInstallation: json.url.includes("setupRequiredForInstallation=true"),
```
leog, line 71: Actually, this kind of solution will be needed in the case we need OAuth2 authorization by the user, redirecting the user to an external site, not just an internal setup page.

packages/prisma/seed-app-store.ts
```diff
@@ -240,8 +240,6 @@ export default async function main() {
   }
   // No need to check if environment variable is present, the API Key is set up by the user, not the system
   await createApp("closecom", "closecomothercalendar", ["other"], "closecom_other_calendar");
-  // No need to check if environment variable is present, the API Key is set up by the user, not the system
-  await createApp("sendgrid", "sendgridothercalendar", ["other"], "sendgrid_other_calendar");
```
leog, line 244: Not sure why this was removed? @hariombalhara

leog (changes requested):
Left some comments, I think the changes I introduced in the Salesforce app PR to tackle the false-positive successful toast message will be better if you ask me, as it covers both use cases, external and internal redirects. The renaming of the app is not actually required for it to work. I would take the icon vs logo fix only 😬

packages/app-store/_utils/useAddAppMutation.ts
hariombalhara: Yeah, those are old apps that are not created using CLI. It should be fixed for all of those together, till then we can make a whitelist here.

packages/prisma/seed-app-store.ts
hariombalhara: seed-app-store-config.json has the same value which is automatically added. So, a manual addition is not required.

hariombalhara:
Cool, Thanks for a quick review. I will check your approach tomorrow @leog 🙏 

Renaming to `sendgrid` is required because of the following reasons:
- **Try deleting the app from DB completely and then do a fresh seed and fresh installation. App won't show up in list.**
- Makes the 'type' thing limited to just type. We don't need to name the directory sendgrid_other_calendar when we can name it just sendgrid
- It is consistent with app URL. App URL is /apps/sendgrid.
- It makes the slug the same as dirName, simplifying things. All app-store-cli created apps have this characteristic.

hariombalhara:
> Left some comments, I think the changes I introduced in the Salesforce app PR to tackle the false-positive successful toast message will be better if you ask me, as it covers both use cases, external and internal redirects. The renaming of the app is not actually required for it to work. I would take the icon vs logo fix only 😬

@leog I too think that your approach is better apart from one limitation that I could see with it.
https://github.com/calcom/cal.com/pull/5406/files#r1017451708

Renaming of the app isn't a requirement due to this but for other reasons I mentioned above.

shirazdole:
@leog - had a customer asking for a SendGrid app, but basically they want emails to program emails through cal.com to be sent through their sendgrid (notifications, workflows, etc)

leog:
> @leog - had a customer asking for a SendGrid app, but basically they want emails to program emails through cal.com to be sent through their sendgrid (notifications, workflows, etc)

SendGrid app was reverted, not available now until we get it back. And what they are looking for is possible if they use their own instance.

leog:
@hariombalhara should I work on the name changes and fixing the logo in the upcoming PR to bring back Sendgrid that was reverted? We can close this one.

hariombalhara:
Sure @leog

leog:
Closing in favor of #5501

PR #5389, opened 2022-11-05
Downgrade vite

emrysal:
Downgrades vite temporarily as the builds are failing

PR #5410, opened 2022-11-08
Downgrade plugin to match vite version

hariombalhara:
Continuation of https://github.com/calcom/cal.com/pull/5389


JeroenReumkens (approved):
@hariombalhara  This changes makes a lot of sense. Tested it locally and works flawlessly! 🙏

PR #5412, opened 2022-11-08
Removed unused ui components and updated skeleton components

JeroenReumkens:
## What does this PR do?

Removed unused ui components and updated skeleton components throughout the app to use new version.

## Type of change

## How should this be tested?

- [ ] Verify the build is successful
- [ ] All components deleted were unused **except one**, so there shouldn't be any components to test
- [ ] Except: The skeleton components — I have updated the few components that were still using the old skeleton to use the new one. The features between these two components were very equal, so I think a quick look at a few of the changed pages would suffice. Pages include: Public booking page (available times), app store detail page (loading state for toggling an app on/off), loading state of /teams page


apps/web/components/apps/App.tsx
```diff
@@ -143,6 +143,7 @@ const Component = ({
             </div>
           ) : existingCredentials.length > 0 ? (
             <DisconnectIntegration
+              buttonProps={{ color: "secondary" }}
```
JeroenReumkens, line 146: Sneaked this change in as per design 😬

apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -152,13 +152,13 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
         {isLoading && !slots.length && (
           <>
             <SkeletonContainer className="mb-2">
-              <SkeletonText width="full" height="20" />
+              <SkeletonText className="h-5 w-full" />
```
JeroenReumkens, line 155: The new skeleton components don't support the width and height props anymore, rather they expect you to use className for it. Added that everywhere.

apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -7,7 +7,7 @@ import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { TimeFormat } from "@calcom/lib/timeFormat";
 import { nameOfDay } from "@calcom/lib/weekday";
 import type { Slot } from "@calcom/trpc/server/routers/viewer/slots";
-import { SkeletonContainer, SkeletonText } from "@calcom/ui";
+import { SkeletonContainer, SkeletonText } from "@calcom/ui/v2";
```
zomars, line 10: Would it make sense to get rid of v1 Skeletons altogether?

zomars (approved):
Seems like a step in the right direction. Nice work @JeroenReumkens 

LGTM

apps/web/components/booking/AvailableTimes.tsx
JeroenReumkens: Yes, asap indeed. Eg console is still using it, so I'm working hard om removing them from there as well, so I can delete all of those v1 components. 🙏

PR #5426, opened 2022-11-08
Fixing count and redundant checks for recurring

leog:
## What does this PR do?

When creating a recurring event, the user selects the number of occurrences, which was not taken into consideration when ensuring availability for the organizer, the number of occurrences used was the one initially configured by the organizer as the maximum number of occurrences. Also, for each recurring booking, the user availability check was run redundantly for unneeded dates. 

For example, given a recurring event with a maximum of 12 occurrences setup by the organizer, a user wants to book 5 occurrences, the checks were checking 12 occurrences either way, and for each of those 12 occurrences, 12 more occurences in the future.

This fixes both issues, also improving the performance avoiding the hard calculation using the heavy-weight RRule.js dependency for it.

My other finding regarding the blocking issue preventing booking a recurring event was fixed in [this commit](https://github.com/calcom/cal.com/commit/69100a0fe8ba756b7b2f0c4022473ea8a35093fc), thanks @emrysal for taking care of changing it as a super hotfix 😁

## Type of change

## How should this be tested?

Try to book a recurring event, being able to successfully book it, probably cutting some time of delay.

### DIFF
--- a/apps/web/lib/mutations/bookings/create-recurring-booking.ts
+++ b/apps/web/lib/mutations/bookings/create-recurring-booking.ts
@@ -3,10 +3,18 @@ import { AppsStatus } from "@calcom/types/Calendar";
 import * as fetch from "@lib/core/http/fetch-wrapper";
 import { BookingCreateBody, BookingResponse } from "@lib/types/booking";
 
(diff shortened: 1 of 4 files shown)


emrysal (approved):
Tested locally, works

PR #5443, opened 2022-11-09
Feat/top banner

### LINKED ISSUE #5287: [CAL-287] Create top banner component
[Top banner - ❖ Cal DS (Figma)](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=21311%3A144646)

Udit-takkar:
Fixes #5287 

Screenshot 2022-11-09 at 7 15 26 PM


packages/ui/components/topbanner/TopBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const variantClassName = {
+  default: "bg-gradient-primary",
+  warning: "bg-orange-400",
+  error: "bg-red-400",
+};
+
+export function TopBanner(props: TopBannerProps) {
+  const { variant = "default", text } = props;
+  const [isOpen, setOpen] = useState(true);
```
sean-brydon, line 20: I know this is still a draft and WIP but I'll quickly leave me concern here. 
Using internal state on this component will mean we also need to create a `useState` in the parent component to toggle the functionality. 

My thoughts would be to take in `isOpen & setOpen` from props. This would allow us to conditionally render the component rather than use css to show/hide it.

Udit-takkar: yeah @sean-brydon  we should take isOpen from props and maybe we create a context wrapper so that we can invoke setOpen from anywhere.

sean-brydon: Good shout :)

packages/ui/components/topbanner/TopBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <div
+      id="banner"
+      className={classNames(
+        " z-50 flex  h-[40px] w-full items-start justify-between gap-8 border border-b border-gray-200 bg-gray-50 px-4 text-center sm:items-center",
+        variantClassName[variant],
+        isOpen ? "sticky" : "hidden"
+      )}>
+      <div className="flex flex-1 items-center justify-center gap-2">
+        <p className="flex items-center justify-center gap-2 font-sans text-sm font-medium leading-4 text-gray-900">
+          {variant === ("warning" || "error") && (
```
zomars, line 32: Does this actually work? IIRC this will never reach error.

```suggestion
          {["warning", "error"].includes(variant) && (
```

packages/ui/components/topbanner/TopBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { variant = "default", text } = props;
+  const [isOpen, setOpen] = useState(true);
+
+  return (
+    <div
+      id="banner"
+      className={classNames(
+        " z-50 flex  h-[40px] w-full items-start justify-between gap-8 border border-b border-gray-200 bg-gray-50 px-4 text-center sm:items-center",
+        variantClassName[variant],
+        isOpen ? "sticky" : "hidden"
```
zomars, line 28: Let's conditionally render instead of hiding with styles

packages/ui/components/topbanner/TopBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+      )}>
+      <div className="flex flex-1 items-center justify-center gap-2">
+        <p className="flex items-center justify-center gap-2 font-sans text-sm font-medium leading-4 text-gray-900">
+          {variant === ("warning" || "error") && (
+            <ExclamationIcon className={classNames("h-5 w-5 text-black")} aria-hidden="true" />
```
zomars, line 33:
```suggestion
            <ExclamationIcon className="h-5 w-5 text-black" aria-hidden="true" />
```

packages/ui/components/topbanner/TopBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+          {text}
+        </p>
+        {props.actions && <div className="text-sm">{props.actions}</div>}
+      </div>
+      <button
+        data-collapse-toggle="banner"
+        type="button"
+        onClick={() => setOpen((open) => !open)}
+        className="hover:bg-gray-20 flex items-center rounded-lg p-1.5 text-sm text-gray-400">
+        <XIcon className={classNames("h-4 w-4 text-black")} />
```
zomars, line 44:
```suggestion
        <XIcon className="h-4 w-4 text-black" />
```

Let's use the `classNames` function only when needed.

packages/ui/components/topbanner/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./TopBanner";
```
zomars, line 1: I'd rather be more explicit on what we're actually importing here.
```suggestion
export { TopBanner, TopBannerProps } from "./TopBanner";
```

packages/ui/components/topbanner/topbanner.stories.mdx
```diff
@@ -0,0 +1,67 @@
+import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs';
+import { Examples, Example, Note, Title,CustomArgsTable,VariantsTable, VariantRow } from '@calcom/storybook/components'
+import { Icon } from "@calcom/ui/Icon";
+import { TopBanner } from './TopBanner.tsx'
```
zomars, line 4: Do we need the explicit `.tsx` here? tagging the Storybook expert @sean-brydon 

```suggestion
import { TopBanner } from './TopBanner'
```

packages/ui/v2/core/Shell.tsx
```diff
@@ -143,12 +144,26 @@ const Layout = (props: LayoutProps) => {
...
+              className="border-b border-black text-sm"
+              onClick={() => {
+                console.log("test");
+              }}>
+              Action
+            </button>
+          }
+          variant="warning"
+          text="An app wide warning is occuring that should be resolved."
+        />
```
zomars, line 160: On which conditions will this component be rendered/hidden?

Udit-takkar: discussion about this here https://threads.com/34433601299?s=v45qV4JNriP5dLaqWvNqNm. this is just for testing how the component would look like

packages/ui/components/topbanner/TopBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+  error: "bg-red-400",
+};
+
+export function TopBanner(props: TopBannerProps) {
+  const { variant = "default", text } = props;
+  const [isOpen, setOpen] = useState(true);
+
+  return (
+    <div
+      id="banner"
```
zomars, line 24: If this is for testing the it should be a `data-testid` otherwise let's remove it since there may be multiple banners and create an ID conflicts

```suggestion
      data-testid="banner"
```

packages/ui/components/index.ts
```diff
@@ -4,3 +4,4 @@ export * from "./buttonGroup";
 export * from "./badge";
 export * from "./breadcrumb";
 export * from "./form";
+export * from "./topbanner";
```
zomars, line 7:
```suggestion
export { TopBanner, TopBannerProps } from "./topbanner";
```

zomars (approved):
Fixed and approved since I need this component for another PR. Thanks again @Udit-takkar 🙏🏽 

LGTM

PR #5445, opened 2022-11-09
Disable ADMIN role when password requirements aren't met

emrysal:
## What does this PR do?

Downgrades the role of a privileged user when the password is less than 15 characters long. Doesn't disable the account but does prevent access of admin features.


pumfleet (approved):
Code looks good to me

zomars (approved):
Is this is a temporal downgrade until the user changes the password? Kinda genius if you ask me

PR #5169, opened 2022-10-21
Teams billing on creation

joeauyeung:
## What does this PR do?

This PR adds Stripe billing when creating a new team. Also when deleting a team, the team on Stripe will be deleted.

Fixes #5166 

**Environment**: Staging(main branch) / Production

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How should this be tested?

For both monthly and annual billing
- Create a team
- Invite members and choose the appropriate billing frequency
- In the Stripe checkout session, ensure that the total adds up and the frequency is correct
- Go to the Stripe dashboard to confirm everything is correct
    - Invoice total and frequency
    - Customer information (name & email)
- In the team settings, visit the billing page and the customer portal
    - Visit the user billing page and go to the billing portal. It should be different from the team.
- On the team profile page, delete the team
    - The team should also be deleted on Stripe

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #5453, opened 2022-11-09
Team billing

### LINKED ISSUE #5166: [CAL-238] Create subscription when creating a team & delete subscription

zomars:
## What does this PR do?

fixes #5166

Following up efforts done in #5169 

To minimize functionality differences between Cal.com and self-hosed we now allow for "unpublished" teams to exist. 

These are teams that have a `null` slug. Which renders the publicly unaccessible (cannot permalink nor book events).

You can still see them on you dashboard and add/remove members. But until you "publish" it the team will become publicly accessible. 

On Cal this means that you need to checkout in order to publish. Self hosted can just publish a team without checkout.

## Type of change

## How should this be tested?

- [ ] `yarn e2e apps/web/playwright/teams.e2e.ts`


packages/prisma/migrations/20221107201132_add_team_subscription_cols/migration.sql
```diff
@@ -0,0 +1,4 @@
+-- AlterTable
+ALTER TABLE "Team" ADD COLUMN     "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+ADD COLUMN     "metadata" JSONB,
+ALTER COLUMN "slug" DROP NOT NULL;
```
zomars, line 4: Needed to have "unpublished" teams

packages/ui/v2/core/Dialog.tsx
```diff
@@ -80,7 +80,7 @@ type DialogContentProps = React.ComponentProps<typeof DialogPrimitive["Content"]
 };
 
 export const DialogContent = React.forwardRef<HTMLDivElement, DialogContentProps>(
-  ({ children, title, Icon, actionProps, ...props }, forwardedRef) => {
+  ({ children, title, Icon, actionProps, useOwnActionButtons, ...props }, forwardedRef) => {
```
zomars, line 83: Fixes a console warning

zomars (commented):
Ready for review

apps/web/playwright/teams.e2e.ts
```diff
@@ -0,0 +1,61 @@
...
+      expect(await page.locator('[data-testid="pending-member-item"]').count()).toBe(1);
+    });
+
+    await test.step("Can publish team", async () => {
+      await page.locator("text=Publish team").click();
+      await page.waitForURL(/\/settings\/teams\/(\d+)\/profile$/i);
+    });
+
+    await test.step("Can disband team", async () => {
+      await page.locator("text=Delete Team").click();
```
alannnc, line 55: most of text locators should go when translation its added to buttons

packages/features/ee/teams/components/v2/CreateNewTeam.tsx
```diff
@@ -1,112 +1,128 @@
...
-                  formMethods.setValue("slug", slugify(e?.target.value));
-                }
-              }}
-              autoComplete="off"
-            />
-          )}
-        />
-      </div>
+  const validateTeamSlug = async () => {
+    console.log("This triggers");
```
alannnc, line 35: remove

apps/web/pages/api/auth/signup.ts
```diff
@@ -71,6 +71,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   });
 
   // If user has been invitedTo a team, we accept the membership
+  // TODO check memberships instead of single invitedTo
```
JeroenReumkens, line 74: Is this a todo we have to fix before merge?

apps/web/pages/settings/teams/new/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import CreateNewTeam from "@calcom/features/ee/teams/components/v2/CreateNewTeam";
+import { getLayout } from "@calcom/ui/v2/core/layouts/WizardLayout";
+
+const CreateNewTeamPage = () => {
+  return (
+    <>
+      <Head>
+        <title>Create a new Team</title>
+        <meta name="description" content="Create a new team to ease your organisational booking" />
```
JeroenReumkens, line 11: Should we add translations here?

packages/app-store/stripepayment/api/portal.ts
```diff
@@ -1,23 +1,37 @@
...
-      customer: customerId,
-      return_url,
-    });
+  let return_url = `${WEBAPP_URL}/settings/billing`;
 
-    res.redirect(302, stripeSession.url);
+  // If accessing a team's portal if referrer has /settings/team/[:teamId]/billing
+  if (/settings\/teams\/\d+\/billing/g.test(referer)) {
+    // Grab the teamId by just matching /settings/teams/[:teamId]/billing and getting third item in array after split
+    const teamId = referer.match(/\/(settings.+)/g) || "";
```
JeroenReumkens, line 26: This is not splitting the team id, right? 🤔  So.... teamId here equals an array with one item being the full string of the path?  Regex is not my strong suit, but might be wise to double check this ✌️

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -1,127 +1,137 @@
...
+      }}>
       <DialogContent
         type="creation"
         useOwnActionButtons
         title={t("invite_new_member")}
         description={
           <span className=" text-sm leading-tight text-gray-500">
-            Note: This will <span className="font-medium text-gray-900">cost an extra seat ($12/m)</span> on
-            your subscription if this invitee does not have a TEAM account.
+            Note: This will <span className="font-medium text-gray-900">cost an extra seat ($15/m)</span> on
```
JeroenReumkens, line 64: Translations?

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -1,127 +1,137 @@
...
+              render={({ field: { onChange } }) => (
+                <div>
+                  <label
+                    className="mb-1 block text-sm font-medium tracking-wide text-gray-700"
+                    htmlFor="role">
+                    {t("role")}
+                  </label>
+                  <Select
+                    defaultValue={options[0]}
+                    options={options.slice(0, 2)}
```
JeroenReumkens, line 104: Would it be safer and perhaps clearer if you'd do something like `options.filter(option => option.value !== 'OWNER')`? This prevents owner appearing in the options by accident if someone where to change the order of the array (something people wouldn't expect when changing the order). Also makes it immediately clear what you're doing here.

packages/features/ee/teams/components/v2/CreateNewTeam.tsx
```diff
@@ -1,112 +1,128 @@
...
-        </Button>
-        <Button color="primary" type="submit" EndIcon={Icon.FiArrowRight} className="w-full justify-center">
-          {t("continue")}
-        </Button>
-      </div>
+            )}
+          />
+        </div>
 
-      {createTeamMutation.isError && <p className="mt-4 text-red-700">{createTeamMutation.error.message}</p>}
```
JeroenReumkens, line 108: I don't see any error handling for this added back in. I believe right now we don't show any error when the mutation fails, is that correct?

JeroenReumkens (dismissed):
Looking good in general! Added a few small remarks / requests. 
Will wait with testing locally until those are resolved. ✌️

hariombalhara:
Screenshot 2022-11-10 at 2 46 51 PM
Able to create events for an Unpublished team(which don't work). May be we can filter-out such teams from showing up here Screenshot 2022-11-10 at 2 49 40 PM

hariombalhara:
I think it doesn't make sense for self-hosted:
Screenshot 2022-11-10 at 3 07 44 PM

hariombalhara:
Screenshot 2022-11-10 at 3 09 08 PM
There are only 2 steps for self-hosted.

packages/trpc/server/routers/viewer/teams.tsx
```diff
@@ -149,17 +149,28 @@ export const viewerTeamsRouter = createProtectedRouter()
...
+        bio: input.bio,
+        hideBranding: input.hideBranding,
+      };
+
+      if (
+        input.slug &&
+        IS_TEAM_BILLING_ENABLED &&
+        /** If the team doesn't have a slug we can assume that it hasn't been published yet. */ !prevTeam.slug
+      ) {
+        data.metadata = { requestedSlug: input.slug };
```
hariombalhara, line 166: Shouldn't we throw an error here, that you can't add a slug without publishing? This is silently discarding the slug which user can consider a bug

hariombalhara (commented):
Left some comments.

apps/web/pages/settings/teams/new/index.tsx
PeerRich: follow up PR

packages/features/ee/teams/components/MemberInvitationModal.tsx
PeerRich: agree, missing

apps/web/playwright/teams.e2e.ts
zomars: Agreed that it can be improved. Alright all e2e tests will always run in English

apps/web/pages/api/auth/signup.ts
zomars: Not needed anymore

packages/trpc/server/routers/viewer/teams.tsx
zomars: Could be. Also we would need a way to publish from the teams view

packages/app-store/stripepayment/api/portal.ts
zomars: Added `returnTo` query instead 🙏

packages/features/ee/teams/components/v2/CreateNewTeam.tsx
zomars: re-added

.env.example
```diff
@@ -92,15 +92,10 @@ TWILIO_MESSAGING_SID=
 NEXT_PUBLIC_IS_E2E=
 
 # Used for internal billing system
-NEXT_PUBLIC_STRIPE_PRO_PLAN_PRICE=
-NEXT_PUBLIC_STRIPE_PREMIUM_PLAN_PRICE=
-NEXT_PUBLIC_IS_PREMIUM_NEW_PLAN=0
-NEXT_PUBLIC_STRIPE_PREMIUM_NEW_PLAN_PRICE=
```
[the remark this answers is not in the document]
zomars: nice catch!

apps/web/playwright/teams.e2e.ts
alannnc: Ah yes all test run on english, but I found some sometimes test run so fast that when fetching certain text it doesn't found it because it loaded the translation key.

alannnc (approved):
Main path worked and I didn't run to any error. Didn't tested payment flow.

NIT:
- Would add a way to go back in create team flow.
- I didn't publish but I lost track of my entered slug, we could probably use another input instead of displaying it empty.

apps/web/pages/api/auth/signup.ts
alannnc: Should  we remove it then?
```suggestion

```

packages/trpc/server/routers/viewer/teams.tsx
zomars: Added error instead

apps/web/playwright/teams.e2e.ts
zomars: Also sidetone that playwright will auto-wait for the text to be visible. So untranslated string shouldn't affect this unless the translation don't load at all.

zomars:
Fixed tests and ready to re-review @calcom/reviewers

PR #5455, opened 2022-11-10
Correction of unclear Japanese translation

geeknees:
## What does this PR do?

'最低頻度の通知` means Minimum Frequency Notification. I understand that the intent here is the minimum time for advance notice of meetings, so I have changed the Japanese translation.

## Type of change

## How should this be tested?

- [x] Check in development environment

スクリーンショット 2022-11-10 15 23 44


alannnc (approved):
Looks good and even google translate agrees on this! Thank you!

PR #5457, opened 2022-11-10
Self-hosted: Team Event Created without explicitly setting locations crashes on /booking page

hariombalhara:
In self-hosted, calvideo isn't default and thus an event can be created without locations

This error comes right now on local, while booking such an event.

Screenshot 2022-11-10 at 2 33 22 PM


sean-brydon (approved):
LGTM - Good catch

PR #5462, opened 2022-11-10
App-Store/Delete Slack App

### LINKED ISSUE #5322: [CAL-301] Remove slack from prod app
Slack app is currently broken and needs removed from prod DB and seed.

hariombalhara:
Slack App needs to be completely redone due to major Slack API changes I guess @sean-brydon ?
It isn't a priority right now, so just remove the app so that people don't use it.
Closes #5322


sean-brydon:
Yeah this isn't necessarily due to the fact slack has had API changes - it's the fact the approach we were taking before was quiet hacky and should ideally be used with their BoltJS framework made for making slack bots. However, this requires another server to be spun up to run.

sean-brydon (approved):
LGTM - we should deffo find time / another person to pick this up @Jaibles

Issue #5435, opened 2022-11-09
Message: Only absolute URLs are supported


rene-mueller:
### Issue Summary

i installed release 2.1.3 after starting and sign in there occurred a error in the console: Only absolute URLs are supported

### Steps to Reproduce

1. install release 2.1.3
2. start application
3. sign in

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

### Technical details

```
@calcom/web:start: [next-auth][error][CLIENT_FETCH_ERROR] 
@calcom/web:start: https://next-auth.js.org/errors#client_fetch_error Only absolute URLs are supported {
@calcom/web:start:   error: {
@calcom/web:start:     message: 'Only absolute URLs are supported',
@calcom/web:start:     stack: 'TypeError: Only absolute URLs are supported\n' +
@calcom/web:start:       '    at getNodeRequestOptions (/opt/app.meedle.me/node_modules/next/dist/compiled/node-fetch/index.js:1:63531)\n' +
@calcom/web:start:       '    at /opt/app.meedle.me/node_modules/next/dist/compiled/node-fetch/index.js:1:65062\n' +
@calcom/web:start:       '    at new Promise (<anonymous>)\n' +
@calcom/web:start:       '    at Function.fetch [as default] (/opt/app.meedle.me/node_modules/next/dist/compiled/node-fetch/index.js:1:64996)\n' +
@calcom/web:start:       '    at fetchWithAgent (/opt/app.meedle.me/node_modules/next/dist/server/node-polyfill-fetch.js:53:39)\n' +
@calcom/web:start:       '    at _callee$ (/opt/app.meedle.me/node_modules/next-auth/client/_utils.js:52:20)\n' +
@calcom/web:start:       '    at tryCatch (/opt/app.meedle.me/node_modules/regenerator-runtime/runtime.js:63:40)\n' +
@calcom/web:start:       '    at Generator.invoke [as _invoke] (/opt/app.meedle.me/node_modules/regenerator-runtime/runtime.js:294:22)\n' +
@calcom/web:start:       '    at Generator.next (/opt/app.meedle.me/node_modules/regenerator-runtime/runtime.js:119:21)\n' +
@calcom/web:start:       '    at asyncGeneratorStep (/opt/app.meedle.me/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)\n' +
[...]

PR #5479, opened 2022-11-11
Bugfix/revert e5ba8616

emrysal:
## What does this PR do?

Reverts e5ba8616
Reverts the Sentry revert


zomars (changes requested):
Thank you Alex. I got some observations:

- Let's add a self review
- Need context on why we're removing sendgrid
- Need context on what was causing the 500 error on user/type

emrysal:
Let's add a self review - purely reverts / adds; no code authored.

* Need context on why we're removing sendgrid
* Need context on what was causing the 500 error on user/type

At this moment we don't know yet, @leog and I will tear that commit apart piece by piece until we've found the reason for a crash that seems unrelated. This commit will allow us to deploy again however something we've been unable to since v2.1.3.

zomars:
Let's fix tests as well

zomars (approved):

hariombalhara:
Wow, crash is gone🤯
 Seems like some unintended import happening in [user]/[type]

rene-mueller:
is this related to https://github.com/calcom/cal.com/issues/5435?

emrysal:
@rene-mueller - could potentially fix your issue too - yes.

Issue #5424, opened 2022-11-08
[CAL-334] disable Update button unless you make a change that requires changing


Udit-takkar:
Follow up [https://github.com/calcom/cal.com/issues/5401](https://github.com/calcom/cal.com/issues/5401)

> From [SyncLinear.com](https://synclinear.com)

PR #5482, opened 2022-11-11
fix: use .watch() to disable update btn

Udit-takkar:
refs: #5424

https://user-images.githubusercontent.com/53316345/201348407-6136c4ab-4702-4389-95f9-86219027c257.mov

this was suggested by @Jaibles here https://threads.com/34433127474?s=ucqt4HpZQcM3tsQmy5m8G5


apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -156,7 +156,7 @@ const ProfileView = () => {
   const {
     reset,
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
zomars, line 158:
```suggestion
```

zomars (changes requested):
No need for use `watch` we can just use the `isDirty` property from `formState`

zomars (commented):
LGTM

zomars (approved):

Udit-takkar:
@zomars  we don't want the update button to be enabled when username is changed because there is already a separate update button for it. isDirty is checking all the fields (including username)

  is there any other way other than .watch() to achieve this?

Screenshot 2022-11-12 at 3 47 12 PM

zomars:
We should encapsulate the username field in another form. So it can have it's own formstate

PR #5493, opened 2022-11-11
Update engines to >=16.x in apps/web and monorepo

emrysal:
## What does this PR do?

Dropping support for 14.x going forward unless reasons exist for persisting our support.

Allow 18.x and potential future versions, In my opinion shouldn't should lock unless we know for a fact a version does not work. So far my observations of 18.x are positive.


zomars (approved):
Let's go!

PR #5514, opened 2022-11-14
Update UI components to be able to use them in Console

JeroenReumkens:
## What does this PR do?

Updates UI components to be able to use them in Console.

Two main changes have been done: 
1. Changed ID of `wizard-form` steps to prefix with `wizard` instead of `setup` to make it more generic. 
2. Remove cloning of form elements in `<Form>` component, since this would break the usage of any field if NOT wrapped in a div. The reason it is not broken in `web` is because in `web` we always wrap input fields in a div. See PR comment in file for more detail. 

## Type of change

## How should this be tested?

- [ ] Check the wizard form in `auth/setup` and verify that it is still looking good and working
- [ ] Verify the usage of forms, the data should still be submitted correctly. For example the API key form in the settings page or the login form uses this new form.


packages/ui/v2/core/WizardForm.tsx
```diff
@@ -52,8 +52,8 @@ function WizardForm<T extends DefaultStep>(props: { href: string; steps: T[]; co
               loading={currentStep.isLoading}
               type="submit"
               color="primary"
-              form={`setup-step-${step}`}
```
JeroenReumkens, line 55: "setup" felt really tailored to the usage in `/auth/setup`. Changed it to `wizard` so it's a reusable id in other forms as well.

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -296,25 +296,7 @@ const PlainForm = <T extends FieldValues>(props: FormProps<T>, ref: Ref<HTMLForm
             });
         }}
         {...passThrough}>
-        {
-          /* @see https://react-hook-form.com/advanced-usage/#SmartFormComponent */
-          React.Children.map(props.children, (child) => {
```
JeroenReumkens, line 301: This was actually breaking all our forms 🤯  
I've been searching everywhere why my new forms in `Console` where NOT working (state got lost of the form fields), but all the fields in web actually WERE working. 

Well.... It turns out that the fields in `Web` also don't work if you put them as a direct child of the form. In `Web` we seem to have gotten used to wrap all our inputs at least in a `div` or `fieldset` to get it to work. So instead of also wrapping my fields in `Console` in `divs` as well, I opted to remove it. 

Would love your opinions on this to know fur sure I didn't went crazy 😁

sean-brydon:
All looks good - not 100% sure on the reasoning behind the Input createElement stuff to begin with - maybe @emrysal has any ideas why we were doing this?

JeroenReumkens:
> All looks good - not 100% sure on the reasoning behind the Input createElement stuff to begin with - maybe @emrysal has any ideas why we were doing this? 

The main reason I believe this was done, was so we don't need to add register to all inputs. Making it easier to render inputs. However, that didn't work at all I noticed. It seems to add wrong parameters. And while diving deeper I found out that we actually don't use it anywhere in web, since we wrap all components in an extra Div, and this createElement only looped through the top level children. If you added any input in the top level in web, you notice that these forms break (lose their data when submitting) as well.

packages/ui/components/form/inputs/Input.tsx
emrysal: Looking at https://codesandbox.io/s/react-hook-form-smart-form-component-forked-iq89z?file=/src/components.js:109-117 there was a mistake made in the manner we handle non-array situations. 

Adopting the following logic seems ideal

```javascript
{Array.isArray(children)
    ? children.map((child) => {
        return child.props.name
          ? React.createElement(child.type, {
              ...{
                ...child.props,
                register,
                key: child.props.name
              }
            })
          : child;
      })
    : children}
```

emrysal:
I've added a code snippet and yes that seems to be the idea @JeroenReumkens - not that we ever used that capability. I do like the idea of auto-registering fields.

JeroenReumkens:
I do like it as well @emrysal. But for now it is broken. So I think removing makes most sense for now. 
We can always add a working version later 😁

packages/ui/components/form/inputs/Input.tsx
JeroenReumkens: I tried that as well. Didn't work either 🤷‍♂️

Also replaced createElement with cloneElement, since that also keeps references for the children. That actually DID work. But I find that quite a risky change to add randomly in this PR. That's why I opted to actually delete it since it is broken now and we don't use it.

emrysal (approved):
Approving with the minor sidenote that Form probably should not be living in `components/form/inputs/Input.tsx` - before it was living in `form/fields.tsx` which also isn't ideal (I think form is where a `index.ts` would make sense)

JeroenReumkens:
Agree @emrysal. Just created an issue for this to do that soon.

pumfleet (approved):

packages/ui/components/form/inputs/Input.tsx
zomars: Yeah I think this abstraction isn't clear enough and we should get rid of it meanwhile we come up with a better one.

PR #5515, opened 2022-11-14
Techdebt/upgrade node18 deps

emrysal:
## What does this PR do?

* Don't support NodeJS 17
* Upgrade all libraries not compatible with node 18 to versions that are.


zomars:
Let's see why tests are failing @emrysal will come back to it once these are passing.

emrysal:
Continued in https://github.com/calcom/cal.com/pull/5622 which is currently being blocked by a typescript error in latest next-auth.

PR #5519, opened 2022-11-14
Use Intl.DateTimeFormat logic to determine locale.

emrysal:
I believe in ChromeHeadless `window.navigator.language` === "": this is throwing errors.

image

(Shared IP in screenshot is AWS, not PII)


leog (approved):
LGTM nice tweak!

PR #5520, opened 2022-11-14
fix: update new team cancel button url

G3root:
## What does this PR do?

before:
https://www.loom.com/share/b1817147a30f4e48acbea1364da1836d

after:
https://www.loom.com/share/137a20734758426b8cfe915f1d342b27

## Type of change


emrysal:
@baileypumfleet @zomars where do we want this to go? Going to the main setting page is not ideal as a possible flow ends up being "/teams" > add a new team > cancel > settings.

zomars:
> @baileypumfleet @zomars where do we want this to go? Going to the main setting page is not ideal as a possible flow ends up being "/teams" > add a new team > cancel > settings.

Ideally we could set a `returnTo` query so we can conditionally return to where we were before starting the new team flow IMO.

G3root:
@emrysal @zomars pushed some changes to set link based on the `returnTo` query param

zomars (approved):
Thank you for your contribution 🙏 @G3root

PR #5521, opened 2022-11-14
Removed an invalidate onDelete + updated test

emrysal:
## What does this PR do?

* Fixes invalidate `await utils.viewer.teams.get.invalidate();` holding up the delete. It tried 3 times to invalidate all resulting in 404's as the resource was gone at that point. 
* Test would break occasionally if any other test for any reason created a team, so did a minor update to fix that potential flakiness.


leog (approved):
LGTM nice catch!

PR #5522, opened 2022-11-14
fix: resetting input on cancel and refactoring input fiedl

Udit-takkar:
Fixes:-
 - After clicking cancel the input has to reset
 - refactored ( as discussed here https://github.com/calcom/cal.com/pull/5482#issuecomment-1312532555 )  input field to have it's own form state  because updating username is different  than other things in profile. username field has it's own inline update button and is independent of other fields.
 
 

https://user-images.githubusercontent.com/53316345/201755837-f665876f-cdd8-46f7-9823-40995c6cbe83.mov


zomars (approved):
Nice job @Udit-takkar 🙏🏽

PR #5533, opened 2022-11-15
Event row style improvements

JeroenReumkens:
## What does this PR do?

Made full row of event type clickable, and fixed event type skeleton dimensions. Skeleton was a regression from removing old v1 skeleton and me not copying over the right dimensions. 

CleanShot 2022-11-15 at 14 40 30@2x

## Type of change

## How should this be tested?
- [ ] Verify that the event types row looks good on /event-types and /pro. The previous change had to do with the fact that very long meeting descriptions (see screenshot) should run over multiple lines. Verify that this is still the case as well (also responsive).
- [ ] Verify that the loading state of event types looks good (it used to be way too small) again!


CarinaWolli (approved):
Looks good now, thank you 🙏

PR #5541, opened 2022-11-15
Adds team upgrade banner

zomars:
## What does this PR do?

fixes #5541

Only applicable to Cal.com, self-hosted shouldn't be affected.

Adds a conditional TopBanner component that will ask the user to upgrade their missing teams.

Loom demo: https://www.loom.com/share/9da1bc4f5b854155a28a04675baa4f3e

## Type of change

## How should this be tested?

- [ ] Set `IS_TEAM_BILLING_ENABLED` to `true`
- [ ] Create a team avoid publishing it. 
- [ ] You should see the banner and allow you to upgrade.
- [ ] The banner should disappear after upgrade.


packages/features/ee/teams/api/upgrade.ts
```diff
@@ -39,22 +39,25 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
   if (!team) {
     const prevTeam = await prisma.team.findFirstOrThrow({ where: { id } });
     const metadata = teamMetadataSchema.parse(prevTeam.metadata);
-    if (!metadata?.requestedSlug) throw new HttpError({ statusCode: 400, message: "Missing requestedSlug" });
-
+    /** Legacy teams already have a slug, this will allow them to upgrade as well */
+    const slug = prevTeam.slug || metadata?.requestedSlug;
+    if (!slug) throw new HttpError({ statusCode: 400, message: "Missing team slug" });
```
zomars, line 44: Explained in comment

packages/features/ee/teams/api/upgrade.ts
```diff
@@ -39,22 +39,25 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
     try {
-      team = await prisma.team.update({
+      /** We save the metadata first to prevent duplicate payments */
```
zomars, line 46: Important note

packages/features/ee/teams/api/upgrade.ts
```diff
@@ -39,22 +39,25 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
           metadata: {
             paymentId: checkoutSession.id,
             subscriptionId: subscription.id || null,
             subscriptionItemId: subscription.items.data[0].id || null,
           },
         },
       });
+      /** Then we try to upgrade the slug, which may fail if a conflict came up since team creation */
+      team = await prisma.team.update({ where: { id }, data: { slug } });
     } catch (error) {
```
zomars, line 59: Had to split the update so we don't miss saving payment data in case of slug conflict.

packages/features/ee/teams/components/TeamListItem.tsx
```diff
@@ -7,6 +7,7 @@ import { getPlaceholderAvatar } from "@calcom/lib/getPlaceholderAvatar";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { RouterOutputs, trpc } from "@calcom/trpc/react";
 import { Icon } from "@calcom/ui/Icon";
+import { Avatar } from "@calcom/ui/components";
```
zomars, line 10: For some reason this wasn't failing earlier

packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
```diff
@@ -0,0 +1,38 @@
...
+    onSuccess(data) {
+      router.push(data.url);
+    },
+    onError: (error) => {
+      showToast(error.message, "error");
+    },
+  });
+  if (!data) return null;
+  const [team] = data;
+  if (!team) return null;
```
zomars, line 22: Instead of displaying all unpublished teams at once. We get the first one and ask for upgrade. Rinse and repeat until all teams are upgraded or deleted.

packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <TopBanner
+      text={`Hey! your team "${team.team.name}" needs to be upgraded.`}
```
zomars, line 25: cc @Jaibles @baileypumfleet some better copy would be nice here. We need to make clear that if users don't upgrade let's say in 30 days their teams will be unpublished.

packages/features/ee/teams/lib/payments.ts
```diff
@@ -30,6 +48,7 @@ export const purchaseTeamSubscription = async (input: { teamId: number; seats: n
       },
     },
   });
+  return { url: session.url };
```
zomars, line 51: We only need the URL really.

packages/ui/v2/core/Shell.tsx
```diff
@@ -141,6 +142,8 @@ const Layout = (props: LayoutProps) => {
         <Toaster position="bottom-right" />
       </div>
 
+      <TeamsUpgradeBanner />
```
zomars, line 145: Maybe later we can make a `DynamicBanners` component that will auto display banners according to each specific User/System status/etc... cc @Udit-takkar

zomars (commented):
Checks are passing, demo and self review done.

packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
alannnc: Do we have a value for that already? @zomars

packages/trpc/server/routers/viewer/teams.tsx
```diff
@@ -607,6 +606,9 @@ export const viewerTeamsRouter = router({
         return { url: checkoutSession.url };
       }
 
+      if (!metadata.data?.requestedSlug)
+        throw new TRPCError({ code: "BAD_REQUEST", message: "Can't publish team without `requestedSlug`" });
```
alannnc, line 610: two line conditionals are better with {}

packages/ui/v2/core/Shell.tsx
sean-brydon: I think this was the plan - we just weren't 100% sure the best approach to handle the data. If I remember correctly a similar thing to how ToastContext works and just have the ability to push into and render all banners.

packages/trpc/server/routers/viewer/teams.tsx
alannnc, line 610:
```suggestion
      if (!metadata.data?.requestedSlug) {
        throw new TRPCError({ code: "BAD_REQUEST", message: "Can't publish team without `requestedSlug`" });
      }
```

alannnc (approved):
Code looks good, some NIT changes and more details about team free usage expiry.

packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
JeroenReumkens: Should we also add a translation in here?

packages/ui/v2/core/Shell.tsx
JeroenReumkens: Yeah @Udit-takkar  is already thinking about this. My idea was to create a separate table in the DB where we push all these notifications to, so we have a single place where all of them live, and then have one root notification banner which pulls all of them from the DB on every page, and shows them. Also see the thread in Threads about this for more opinions 😄

JeroenReumkens (approved):

packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
zomars: Once we have final copy yes

PR #5551, opened 2022-11-16
CAL-300 Ensure we show the default calendar name based in the advance…

### LINKED ISSUE #5320: [CAL-300] /event-types: advanced settings show "Default calendar()" and not the first entry
this should by default take the first calendar

(sorry for the potato quality, screengrabbed this from a loom)

JeroenReumkens:
## What does this PR do?

CAL-300 Ensure we show the default calendar name based in the advanced tab for event types.

Fixes #5320 

CleanShot 2022-11-16 at 14 55 47@2x

## Type of change

## How should this be tested?

- [ ] Add a new event type and ensure you have a Google Calendar connected. After that, verify that the default calendar is now visible instead of empty brackets
- [ ] Verify that if you select a different calendar that this is remembered and shown correctly as well. 
- [ ] I've also added an if statement around the brackets show we don't show them if we can not fetch any email. Makes it look a bit nicer, just don't think you can test that.


apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -74,6 +74,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
               defaultValue={eventType.destinationCalendar || undefined}
               render={({ field: { onChange, value } }) => (
                 <DestinationCalendarSelector
+                  destinationCalendar={eventType.destinationCalendar}
```
JeroenReumkens, line 77: We didn't even pass in the calendar for the event to component 😬  So actually even if you selected something else, you would still see "default ()". That should be fixed now too.

packages/features/calendars/DestinationCalendarSelector.tsx
```diff
@@ -115,6 +115,11 @@ const DestinationCalendarSelector = ({
           value: `${cal.integration}:${cal.externalId}`,
         })),
     })) ?? [];
+
+  // Get primary calendar, which is shown in the placeholder since this is the calendar that will
+  // be used when no destination calendar is selected.
+  const primaryCalendar = query.data.connectedCalendars.filter((cal) => Boolean(cal.primary))[0];
```
JeroenReumkens, line 121: If there is NO calendar connected, there's not destinationCalendar in the `eventType` object. We rather need to get the default from the calendars in that case. That is what this change does.

zomars (approved):
Simple enough. Thanks @JeroenReumkens 🙏🏽 

LGTM

PR #5552, opened 2022-11-16
fix: ui of button

### LINKED ISSUE #5548: [CAL-379] Inline banner needs to be responsive
Right = Current

Left = Preferred

> From [SyncLinear.com](https://synclinear.com) | [CAL-379](https://linear.app/calcom/issue/CAL-379/inline-banner-needs-to-be-responsive)

Udit-takkar:
Fixes: #5548 
Screenshot 2022-11-16 at 7 57 24 PM


packages/ui/v2/core/banner.tsx
```diff
@@ -38,29 +35,28 @@ const Banner = (props: BannerProps) => {
...
-          className={buttonStyle}
-          onClick={(e) => {
-            setShow(false);
-            props.onDismiss(e);
-          }}>
-          Dismiss
-        </Button>
+        <div className="mt-4 flex w-full items-center gap-4 self-start text-sm font-medium">
+          {props.actionText && (
+            <button className="text-sm font-medium leading-4" onClick={() => props.onAction}>
```
alannnc, line 47: is there a reason why we changed Button for button?

alannnc (commented):
Try to fill the data from the PR placeholder, that way others can quickly jump and test what is being fix on the PR.

packages/ui/v2/core/banner.tsx
Udit-takkar: image

left one is what the ui we want. Button in /components has py and px and background change on hover which we don't want so it would require over ridding a lot of styles if we use Button or should i just add another variant prop and use Button ?

alannnc (approved):

Issue #5555, opened 2022-11-16
[CAL-382] Minimum booking notice conversion are incorrect


zomars:
### Issue Summary

It seems like the Minute / Hour / Days conversion for the minimum booking notice are incorrect:

[https://user-images.githubusercontent.com/3504472/202286359-fb0cba3c-cc84-4f4a-9569-35a0f87fd023.mov](https://user-images.githubusercontent.com/3504472/202286359-fb0cba3c-cc84-4f4a-9569-35a0f87fd023.mov)

### Steps to Reproduce

1. Go to an event type page
2. Go to limits tab
3. Play around with the minimum booking notice values and see how conversion is wrong.

> From [SyncLinear.com](https://synclinear.com)

JeroenReumkens:
Closed by #5557

PR #5557, opened 2022-11-17
Fixed wrong calculation of duration type on event type page.

JeroenReumkens:
## What does this PR do?

Calculation of the event type duration was not correct, eg 90 mins was converted into a very big number of hours. 

## Type of change

## How should this be tested?

- [ ] Go to an event type page, go to the advanced tab, fill in a number of minutes (eg 300). Save it, and see that after a refresh we've converted into the correct amount of hours or days. It will only show it as hours / days when it rounds to a full amount of hours / days. No decimals. If it does not round, we still show minutes.


hariombalhara (approved):

JeroenReumkens:
Fixes #5555

PR #5561, opened 2022-11-17
🧹 One `calcom/ui` import to rule them all

JeroenReumkens:
Required for:  

https://github.com/calcom/website/pull/398

https://github.com/calcom/console/pull/44

## What does this PR do?

Removes more v1 components and replaces them with the new versions. 

This PR includes changes to: 
* Removal of old EmptyScreen component. Was used in:
- [ ] License required page shown on the workflow page
- [ ] Workflow overview page, empty screen shown when you have 0 workflows.
* Removed v1 of workflow pages. They weren't used anymore. Any diff you see in these pages are because v2 files had the same name. 
- [ ] Confirm workflow pages still look okay
* Removed 3 out of 4 unused components from `/apps/web/components/integrations`. (CalendarListContainer, DisconnectIntegration, IntegrationListItem). All of them had duplicates in other places that were used instead of these.
* Removed v1 of `Switch` component and replace it with v2. Was still used at:
- [ ] API key dialog form in `/settings/developer/api-keys` modal
* Deleted v1 components of API-keys section in settings
- [ ] Confirm the modal still looks okay
* Removed v1 version of `List` component. Was still used at:
- [ ] `/getting-started` flow in the step where you connect calendars. The list of calendars is the `List` component. 

## Type of change

## How should this be tested?

Use the checks in the description above to test and verify that the changes didn't break anything 🙏


apps/web/pages/sandbox/List.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
```
JeroenReumkens, line 2: Do we still use these sandbox pages actually? Otherwise I'd be more than happy to delete them 😁

packages/features/ee/api-keys/components/ApiKeyDialogForm.tsx
```diff
@@ -2,48 +2,54 @@ import { useState } from "react";
 import { Controller, useForm } from "react-hook-form";
```
JeroenReumkens, line 2: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/api-keys/components/ApiKeyListItem.tsx
```diff
@@ -1,100 +1,96 @@
 import dayjs from "@calcom/dayjs";
```
JeroenReumkens, line 1: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/workflows/components/AddActionDialog.tsx
```diff
@@ -1,8 +1,3 @@
-/**
```
JeroenReumkens, line 1: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/workflows/components/AddVariablesDropdown.tsx
```diff
@@ -1,58 +1,63 @@
-/**
```
JeroenReumkens, line 1: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/workflows/components/WorkflowDetailsPage.tsx
```diff
@@ -1,19 +1,18 @@
-import { ArrowDownIcon } from "@heroicons/react/outline";
 import { WorkflowActions, WorkflowTemplates } from "@prisma/client";
```
JeroenReumkens, line 1: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/workflows/components/WorkflowListPage.tsx
```diff
@@ -1,108 +1,169 @@
+import { Workflow, WorkflowStep } from "@prisma/client";
```
JeroenReumkens, line 1: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/workflows/pages/index.tsx
```diff
@@ -1,45 +1,72 @@
 import { useSession } from "next-auth/react";
+import { useRouter } from "next/router";
```
JeroenReumkens, line 2: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

packages/features/ee/workflows/pages/workflow.tsx
```diff
@@ -13,18 +13,21 @@ import { useEffect, useState } from "react";
 import { useForm } from "react-hook-form";
 import { z } from "zod";
```
JeroenReumkens, line 14: All diffs in this page are because the v1 version had the same name, and I moved v2 to the same place. So this is actually the difference between v1 and v2.

apps/web/pages/sandbox/List.tsx
zomars: I would say go for it. The less code to maintain the better

PeerRich: agree

zomars (approved):
Let's go big or go home!

- Fixed all `@calcom/ui` imports
- Removed `sandbox` pages.
- Removed more deprecated components
- Removed implicit UI indexes over explicit ones (no more `export * from...` unless necessary.

Ship it

packages/ui/Kbar.tsx
```diff
@@ -1,19 +1,19 @@
...
+  useMatches,
 } from "kbar";
 import { useRouter } from "next/router";
 
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { isMac } from "@calcom/lib/isMac";
-import { Icon } from "@calcom/ui/Icon";
-import Tooltip from "@calcom/ui/v2/core/Tooltip";
+
+import { Icon, Tooltip } from ".";
```
zomars, line 16: IMO all `calcom/ui` imports inside `calcom/ui` should be relative. If we're finding too many relative imports here is a mostly a symptom that there are many components that don't belong here. For example `feature` specific components should go into `calcom/features` instead.

packages/ui/components/badge/Badge.tsx
```diff
@@ -3,7 +3,7 @@ import { GoPrimitiveDot } from "react-icons/go";
 
 import classNames from "@calcom/lib/classNames";
 
-export const badgeClassNameByVariant = {
+const badgeClassNameByVariant = {
```
zomars, line 6: This was being exported to infer types. Let's use the exported type instead.

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -1,16 +1,12 @@
...
 import { Eye, EyeOff } from "react-feather";
 import { FieldValues, FormProvider, SubmitHandler, useFormContext, UseFormReturn } from "react-hook-form";
 
 import classNames from "@calcom/lib/classNames";
 import { getErrorFromUnknown } from "@calcom/lib/errors";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
-import { Alert } from "@calcom/ui/v2/core/Alert";
-import { Tooltip } from "@calcom/ui/v2/core/Tooltip";
-import showToast from "@calcom/ui/v2/core/notifications";
-import { Skeleton } from "@calcom/ui/v2/core/skeleton";
```
zomars, line 12: No `@calcom/ui` imports inside `@calcom/ui`

packages/ui/components/index.ts
```diff
@@ -1,8 +1,27 @@
...
+  Input,
+  InputField,
+  InputGroupBox,
+  InputLeading,
+  Label,
+  PasswordField,
+  TextArea,
+  TextAreaField,
+  TextField,
+} from "./form";
```
zomars, line 25: I'd rather concentrate the complexity here instead of each individual file. If we stop exporting a file then we expect an error and will help us find old/deprecated code and replace It easily.

JeroenReumkens:
@zomars Thanks for pushing this PR even further over the finish line 😁 Go big or go home definitely applies now 🤣❤️

Nice to see you were also already able to update all of the imports to /ui. I've tried that before and had quite a lot of issues with circular imports. Amazing to see that this is actually working now 😁

Thanks!

PR #5562, opened 2022-11-17
fix: return when success status 204

### LINKED ISSUE #5553: [CAL-381] Zoom link disappears/breaks when you reschedule a meeting
**Steps to reproduce**

* Book a call with me [https://cal.com/ciaran/catchup](https://cal.com/ciaran/catchup)
  * The zoom link gets added and Zoom is included
* Reschedule the meeting
  * The Zoom link drops off the meeting

I got an email about this but Zoom is correctly set in the event type.

Is removing and adding the app the only way to fix this?

> From [SyncLinear.com](https://synclinear.com) | [CAL-381](https://linear.app/calcom/issue/CAL-381/zoom-link-disappearsbreaks-when-you-reschedule-a-meeting)

Udit-takkar:
Fixes: #5553 

updating event in zoom returns 204 so we cannot use .json() on its response

Screenshot 2022-11-17 at 4 33 03 PM


emrysal (changes requested):
An error thrown in an async function is automatically a Promise rejection, so wrapping in try {} catch {} is redundant. 

This fix is also not the right place - this should be handled in the handleZoomResponse.

Udit-takkar:
@emrysal  ok made the changes and tested it.

emrysal (approved):

PR #5566, opened 2022-11-17
Feature/Routing Forms: Add the concepts of reusable Router and Router Linked Fields.

### LINKED ISSUE #5508: [CAL-368] RFC: Routing Forms - Reusability and Extensibility of Router
* Allow a Router to be created outside of a Form - Call it a **Global Router**
* User can create a **Global Router** which would have just Form and Routing tabs.
  * This is because a Router can't be created without Form Fields.
* When a new Form is created user can optionally select which **Global Router** to use. We can call such a Form as **Global Router Linked Form**
  * It would automatically add all the fields required by the Router in addition to Routing Rules
* Characteristics of **Global Router Linked Forms**
  * Such a form has a Global Router which can be edited at one place and all Global Router Linked Forms linked to that Global Router would have the effect.
  * would have **Global Router Fields** locked
  * would have **Global Router Rules** locked
  * User can add more fields.
  * User would be able to add more Routing Rules
  * ~The way user can click Add a rule, user can do 'Add a Global Router'. This allows a scenario like Having a Global Router as 'Enterprise Client'(reusable across multiple forms) and make it AND/OR with other form specific rules.~ The ~striked off~ approach won't work as a Router also has the corresponding Route and not just the AND/OR condition. The only thing that's possible is to decide where exactly Global Router would come in ordered Routes list. So, there can be
    * R1, R2, GR(Global Route)
    * GR, R1, R2
    * R1, GR, R2
* **Global Routers** should list which Forms are linked to it.
* User can save an existing form as `Global Router` as well
* **Global Router Fields** still can be re-ordered inside a Global Router Linked Form. That reordering would impact only that Form and not other Global Router Linked Forms or the Global Router itself. It can also be said that Global Router doesn't care about the ordering of fields. Global Router Linked Form cares about that.
* User would be able to delete Global Router fields(Would be a soft delete only)

Legend:

* Router is the Routing Rules which are available under the **Routing** tab

From [SyncLinear.com](https://synclinear.com/) | [CAL-368](https://linear.app/calcom/issue/CAL-368/rfc-routing-forms-reusability-and-extensibility-of-router)

hariombalhara:
## What does this PR do?

Closes #5508

- This is a live connection and editing the fields and routing in the form being used as Router should be reflected in the new form as well.

Loom Video: https://www.loom.com/share/d0ef3bcaa77c4c84b8159c255e744a14

**Environment**: Production

## Type of change

## How should this be tested?

- [x] See the loom

## Documentation
- Terminology
    - **Router** : A form that has been picked as a route.
	- **F1 ← F2 :** F2 is using F1 as a Router. We can call F1 a router in this context. Fields in F2 coming from F1 are called _Router Linked Fields_.
	- **Connected Forms:** For a relationship F1 ← F2, F2 is Connected Form

- **Restrictions**
    - You can’t delete fields connected to a Router. To delete, you can remove the connection(by removing the Router from routes list). That would make all the Router fields local. After that you can delete them.
    - A multi-level relationship like F1<-F2<-F3 is not allowed only one level like F1<-F2 is allowed. This is to prevent creation of circular dependencies like F3<-F1<-F2<-F3 

- **Actions**
    - [ ] **Disconnect** → [Not yet Implemented]You can disconnect a router. When a Router is disconnected, the Router fields would become local and won’t get removed. The routing would also become local.
    - [x] **Removing a Router →** Deleting a router removes the routing logic and corresponding fields of that Router from the connected form.
    - [x] **Adding after Removing** → Adding after removing would re-add the fields. It simply marks the fields as non-delet
    - [x] **Updating Field →** Modifying **Type/Label/Required** options of a field in a Router, updates them in the connected forms.
    - [x] **Delete Field →** For a relationship **F1 ← F2**, if a field is deleted in F1, it won’t be deleted from F2 but become local in F2, so that it can be modified in F2.

- **Behaviours**
	- [x] If a form F2 picks F1 as a Router - F1’s deleted fields won’t appear in F2 anywhere.
	- [x] If a relationship like F1 ← F2 exists, then F1 can’t have F2 in ‘Add a new Route’ listing. This is to avoid a circular relationship.
	- [x] A form can be picked as a router only once in a given form.

- [x] Playright Tests


packages/lib/logger.ts
```diff
@@ -3,7 +3,7 @@ import { Logger } from "tslog";
 import { IS_PRODUCTION } from "./constants";
 
 const logger = new Logger({
-  minLevel: "warn",
```
hariombalhara, line 6: Remove it

packages/prisma/migrations/20221114095200_add_router_table/migration.sql
```diff
@@ -0,0 +1,16 @@
+-- CreateTable
```
hariombalhara, line 1: Remove all migrations.No migration is required now.

PeerRich (commented):
i wish i knew what was going on here 🤣

CarinaWolli:
@hariombalhara If F2 has F1 as a router, then I would imagine that when I rename a field in F1 or add a new field in F1, that the changes would also appear in F2. But that's not happening for me, let me know if I am misunderstanding something here

hariombalhara:
@CarinaWolli yes, that should have worked. Looks like my recent cleanup broke it ;( Will get back to you soon.

hariombalhara:
@CarinaWolli  Fixed that issue. Thanks for reporting. Really appreciate it 🙏 I will be testing things more and will also add Playwright tests to avoid such regressions.

hariombalhara (commented):
Self review done

CarinaWolli (changes requested):
Great new feature, it's approved from my side! Requesting changes for now to block merge until after holidays 🎄

CarinaWolli:
@hariombalhara Let's fix the merge conflicts and then I'll approve 🙏

hariombalhara:
@CarinaWolli Resolved.

CarinaWolli (approved):
Great work 🙏

PR #5592, opened 2022-11-20
fix: disable update button when nothing changed

### LINKED ISSUE #5589: [CAL-397] Buttons in forms should be disabled until something changes
* Our buttons in all forms are all enabled by default. 
* They should start disabled, then become enabled when there's a change.
* This helps act as a visual hint/reminder to hit Save (until we have autosaving)

> From [SyncLinear.com](https://synclinear.com) | [CAL-397](https://linear.app/calcom/issue/CAL-397/buttons-in-forms-should-be-disabled-until-something-changes)

Udit-takkar:
Signed-off-by: Udit Takkar <udit.07814802719@cse.mait.ac.in>

Fixes: https://github.com/calcom/cal.com/issues/5589

https://user-images.githubusercontent.com/53316345/202919517-303f5aa5-b513-41ab-b7c1-c3262810a42c.mov


apps/web/pages/settings/my-account/appearance.tsx
```diff
@@ -37,21 +37,35 @@ const SkeletonLoader = () => {
...
 
-  const formMethods = useForm();
-
   if (isLoading) return <SkeletonLoader />;
 
-  if (user)
+  const fields = formMethods.watch();
+
+  if (user) {
+    const isDisabled =
```
Udit-takkar, line 62: in /profile we used isDirty property https://github.com/calcom/cal.com/blob/main/apps/web/pages/settings/my-account/profile.tsx#L199. but here isDirty behaviour is a bit weird it should  return true when form value change but it is only work fine in case of theme else all fields the value doesn't change. I have tried passing defaultValues in form or  resetting with keepDirtyValues true but it is not working for some reason

alannnc (approved):
Added defaultValues to form and also prop shouldDirty: true when needed.

Udit-takkar:
@alannnc  thanks it works fine now

PR #5601, opened 2022-11-21
Booking Limits checking ALL events not the event the limit is applied too

sean-brydon:
This PR ensure we only check the event type the user is booking and not all bookings of OUR bookings in the users calendar.

**_Should_** fix the issue where we were getting an inconsistent result between availability being hidden on the calendar and the booking error message 

Translate error string on booking page if any. If there is no translation found we will just display the English error as per the normal usage of `useLocale`

CleanShot 2022-11-21 at 11 42 49@2x

### DIFF
--- a/packages/core/getUserAvailability.ts
+++ b/packages/core/getUserAvailability.ts
@@ -202,10 +202,12 @@ export async function getUserAvailability(
         const startDate = dayjs(booking.start).startOf(filter);
         // this is parsed above with parseBookingLimit so we know it's safe.
         const endDate = dayjs(startDate).endOf(filter);
-
+        const bookingEventTypeId = booking.source?.split("-")[1];
+        console.log({ bookingEventTypeId, eventTypeId, booking });
         if (dayjs(booking.start).isBetween(startDate, endDate)) total++;
-        if (total >= limit)
+        if (total >= limit && bookingEventTypeId === eventType?.id?.toString()) {
           bufferedBusyTimes.push({ start: startDate.toISOString(), end: endDate.toISOString() });
+        }
       });
     }
   }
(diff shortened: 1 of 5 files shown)


packages/core/getUserAvailability.ts
```diff
@@ -202,10 +202,11 @@ export async function getUserAvailability(
         const startDate = dayjs(booking.start).startOf(filter);
         // this is parsed above with parseBookingLimit so we know it's safe.
         const endDate = dayjs(startDate).endOf(filter);
-
+        const bookingEventTypeId = booking.source?.split("-")[1];
         if (dayjs(booking.start).isBetween(startDate, endDate)) total++;
-        if (total >= limit)
+        if (total >= limit && bookingEventTypeId === eventType?.id?.toString()) {
```
sean-brydon, line 207: Only check OUR booking that matches the current eventTypeId - we don't care about another event type in this case

emrysal:
Looks good to me - waiting for feedback before approval on the following:

💡 All the Github comments are probably suitable code comments for the code

packages/lib/server/checkBookingLimits.ts
```diff
@@ -86,7 +86,7 @@ export async function checkLimit({
       }
 
       throw new HttpError({
-        message: `Booking limit of ${limitingNumber}/${filter} reached for this eventType`,
+        message: `Booking limit of ${limitingNumber}/${filter} reached for this Event Type`,
```
emrysal, line 89: This should be translatable - this is meant for the FE but as it is meant for display it needs a translation string. We should name this error so we can use the name as a translation key (I think that's the easiest approach)

sean-brydon: How would we handle passing in the limitingNumber and filter to the FE error?

emrysal: For now let's keep it simple, we can come up with a more global strategy in the future as we fix similar HttpError's

emrysal (approved):

emrysal (approved):

PR #5609, opened 2022-11-21
Trigger webhooks when booking is confirmed not submitted

CarinaWolli:
## What does this PR do?

Like for workflows, webhooks should also trigger when the booking is confirmed and not when an event that requires confirmation is booked. 
This PR also fixes that when a recurring event was confirmed it sent out workflow emails ('when new event is booked' trigger) for all events.

## Type of change

## How should this be tested?

- Test 'when new event is booked' trigger in workflows with a recurring event that requires confirmation
  - Only one email should be sent when the series of bookings is confirmed
- Test webhooks with events that require confirmation 
  - Should not trigger when event is booked, it should trigger once the event is confirmed


packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -944,13 +950,46 @@ export const bookingsRouter = router({
           triggerEvent: WebhookTriggerEvents.MEETING_ENDED,
         };
 
+        const subscriberOptionsBookingCreated = {
```
CarinaWolli, line 953: 'Booking Created' trigger should trigger when event is confirmed

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -635,6 +635,12 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
   if (rescheduleUid) {
     originalRescheduledBooking = await getOriginalRescheduledBooking(rescheduleUid);
   }
+  // If the user is not the owner of the event, new booking should be always pending.
+  // Otherwise, an owner rescheduling should be always accepted.
+  // Before comparing make sure that userId is set, otherwise undefined === undefined
+  const userReschedulingIsOwner = userId && originalRescheduledBooking?.user?.id === userId;
+  const isConfirmedByDefault =
+    (!eventType.requiresConfirmation && !stripeAppData.price) || userReschedulingIsOwner;
```
leog, line 643: Is there any way we can perhaps expand this into regular if-else? I think it's a bit difficult to read as it is

leog (approved):
LGTM, left a nit.

PR #5622, opened 2022-11-21
Upgrade deps + allow node 18.x

emrysal:
## What does this PR do?

* Add 18.x support for NodeJS


packages/trpc/server/routers/viewer/saml.tsx
```diff
@@ -30,7 +30,7 @@ export const samlRouter = router({
       }
 
       // Retrieve the SP SAML Config
-      const SPConfig = samlSPConfig.get();
+      const SPConfig = await samlSPConfig.get();
```
emrysal, line 33: https://github.com/boxyhq/jackson/pull/667#issuecomment-1322630427

emrysal:
This is currently blocked by next-auth@v4.17.0 which contains an (unrelated to Node v18) issue with the typings documented in https://github.com/nextauthjs/next-auth/issues/5576. 

The issue was introduced in v4.13.0 and unfortunately to support NodeJS v18 we need v4.15.1, upgrading is not possible at this time.

packages/trpc/server/routers/viewer/saml.tsx
deepakprabhakara: Thanks @emrysal, added the upgrade note here - https://boxyhq.com/docs/jackson/upgrade#upgrading-to-v136

emrysal: Loving the context for the change! 👍

emrysal:
@zomars It required quite a bit of changes (and investigation) - but Node v18 builds are working!

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -331,16 +331,16 @@ export default NextAuth({
     async signIn(params) {
       const { user, account, profile } = params;
 
-      if (account.provider === "email") {
+      if (account?.provider === "email") {
```
Comment (zomars) [inline apps/web/pages/api/auth/[...nextauth].tsx:334]: Why not check for account instead of `?` every instance?

```suggestion
      if (!account) return false;

      if (account.provider === "email") {
```

Comment (emrysal) [reply apps/web/pages/api/auth/[...nextauth].tsx:334]: I don't know the expected values of account in all login scenario's (what is it when the provider is google, when credentials, etc) and also this would require me to at least check the flow of this entire function.

This approach was much easier development wise - undefined !== "oath" == return false, so that is certainly the right behaviour. This approach prevented having to test everything through as the logic remains self contained.

Comment (zomars) [reply apps/web/pages/api/auth/[...nextauth].tsx:334]: Fair enough

zomars (approved):
LGTM

PR #5636, opened 2022-11-22
Attempt to fix vercel edge crash

alishaz-polymath:
## What does this PR do?

Likely fixes the edge function crash that started slightly after the previous function got merged, although some caching is going on making it hard to validate.


emrysal (approved):
If it works on windows, let's merge this 🚀

PR #5653, opened 2022-11-23
Calendar Weekly Scheduler 

### LINKED ISSUE #5590: [CAL-398] New booking layout - Weekly view
We're adding a new layout option for the public booking page, a weekly view.

[2. Weekly - Cal - New Features (Figma)](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/Cal---New-Features?node-id=451%3A65668&t=pgAbFenXbJx7vDAK-1)

> From [SyncLinear.com](https://synclinear.com) | [CAL-398](https://linear.app/calcom/issue/CAL-398/new-booking-layout-weekly-view)

sean-brydon:
## What does this PR do?

The time has come 

Loom: https://www.loom.com/share/4c38c3f0a58f435dbc29d96d6035c15b

Fixes #5590

## Type of change

## How should this be tested?
- Please pull down this PR
- `yarn install`
- `cd apps/storybook`
- `yarn dev`
- Test the first storybook -> Press `D` to open up the controls section - try changing around props. 
- View the documentation side of the storybook to see information 

---

## Todos

- [ ] Handle hover overlapping events 
- [ ] Current time line 
- [ ] Mobile Support
- [ ] Pixel push


packages/ui/components/scheduler/_storybookData.ts
```diff
@@ -0,0 +1,71 @@
+import dayjs from "@calcom/dayjs";
```
sean-brydon, line 1: Storybook data used in stories.

Sinde we use mdx files - its a lot easier defining this here than inside the file itself

packages/ui/components/scheduler/components/DateValues/DateValues.tsx
```diff
@@ -0,0 +1,60 @@
...
+export function DateValues({ days, containerNavRef }: Props) {
+  return (
+    <div
+      ref={containerNavRef}
+      className="sticky top-0 z-30 flex-none border-b border-b-gray-300 bg-white sm:pr-8">
+      <div className="flex text-sm leading-6 text-gray-500 sm:hidden" data-dayslength={days.length}>
+        {days.map((day) => {
+          const isToday = dayjs().isSame(day, "day");
+          return (
+            <button
```
sean-brydon, line 20: Mobile mode - render a header of all dates in the days array.

packages/ui/components/scheduler/components/DateValues/DateValues.tsx
```diff
@@ -0,0 +1,60 @@
...
+                  isToday && "rounded-full bg-gray-900 text-white"
+                )}>
+                {day.format("D")}
+              </span>
+            </button>
+          );
+        })}
+      </div>
+      <div className="-mr-px hidden auto-cols-fr  text-sm leading-6 text-gray-500 sm:flex ">
+        <div className="col-end-1 w-14" />
```
sean-brydon, line 37: Blank square for now - in future this will allow the user to change the timezone of their calendar to get a preview

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { SchedulerColumns } from "./grid";
+import { SchedulerHeading } from "./heading/SchedulerHeading";
+import { HorizontalLines } from "./horizontalLines";
+import { VeritcalLines } from "./verticalLines";
+
+export default function Scheduler(props: SchedulerComponentProps) {
+  const container = useRef<HTMLDivElement | null>(null);
+  const containerNav = useRef<HTMLDivElement | null>(null);
+  const containerOffset = useRef<HTMLDivElement | null>(null);
+  const initalState = useSchedulerStore((state) => state.initState);
```
sean-brydon, line 20: Refs used to get scroll height - offset height etc etc

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+  useEffect(() => {
+    initalState(props);
+  }, [props, initalState]);
+
+  return (
+    <MobileNotSupported>
+      <div
+        className="scheduler-wrapper flex h-full w-full flex-col overflow-y-scroll"
+        style={
+          { "--one-minute-height": `calc(1.75rem/(60/${usersCellsStopsPerHour}))` } as React.CSSProperties // This can't live in the css file because it's a dynamic value and css variable gets super
```
sean-brydon, line 46: Define a global CSS var we can use across the app. This has to live here because css vars don't really support dynamic variables and makes life a bit awkward

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+                <SchedulerColumns
+                  zIndex={50}
+                  offsetHeight={containerOffset.current?.offsetHeight}
+                  gridStopsPerDay={numberOfGridStopsPerDay}>
+                  <>
+                    {[...Array(days.length)].map((_, i) => (
+                      <li
+                        key={i}
+                        style={{
+                          gridRow: `2 / span ${numberOfGridStopsPerDay}`,
```
sean-brydon, line 80: we use `2` here due to the fact the first row is reserved for all day events (Not yet implemented)

Then we span all of the grid stops in the day

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+                          position: "relative",
+                        }}>
+                        {/* While startDate < endDate:  */}
+                        {[...Array(numberOfGridStopsPerDay)].map((_, j) => {
+                          const key = `${i}-${j}`;
+                          return (
+                            <EmptyCell
```
sean-brydon, line 87: Always render an empty cell 
This allows us to easily handle hover states as previously we were just using grid templates and it was impossible to handle preview hovers in a nice way.

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+                  <>
+                    {[...Array(days.length)].map((_, i) => (
+                      <li
+                        key={i}
+                        style={{
+                          gridRow: `2 / span ${numberOfGridStopsPerDay}`,
+                          position: "relative",
+                        }}>
+                        {/* While startDate < endDate:  */}
+                        {[...Array(numberOfGridStopsPerDay)].map((_, j) => {
```
sean-brydon, line 84: [Renders every day ](https://github.com/calcom/cal.com/pull/5653/files#r1030371078)

Then we render every single "time slot" an empty cell

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+                      </li>
+                    ))}
+                  </>
+                </SchedulerColumns>
+
+                <SchedulerColumns
+                  offsetHeight={containerOffset.current?.offsetHeight}
+                  gridStopsPerDay={numberOfGridStopsPerDay}>
+                  {/*Loop over events per day  */}
+                  {days.map((day, i) => {
```
sean-brydon, line 106: Goes through every single day in a separate grid to empty cells

packages/ui/components/scheduler/components/Scheduler.tsx
```diff
@@ -0,0 +1,134 @@
...
+                    return (
+                      <li key={day.toISOString()} className="relative" style={{ gridColumnStart: i + 1 }}>
+                        <EventList day={day} />
+                        <BlockedList day={day} containerRef={container} />
```
sean-brydon, line 110: Render event list + blocked list on this grid

packages/ui/components/scheduler/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,111 @@
...
+            top: `var(--one-minute-height)`,
+            zIndex: 60,
+            height: `calc(${(endHour + 1 - startHour) * 60} * var(--one-minute-height))`, // Add 1 to endHour to include the last hour that we add to display the last vertical line
+          }}>
+          <BlockedTimeCell />
+        </div>
+      )}
+    </>
+  );
+}
```
sean-brydon, line 40: Whenever you are looking at a `day` that is before `today` render a whole empty time cell for the entire grid.

packages/ui/components/scheduler/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,111 @@
+import { useId, useMemo } from "react";
+
+import dayjs from "@calcom/dayjs";
+
+import { useSchedulerStore } from "../../state/store";
+import { BlockedTimeCell } from "./BlockedTimeCell";
+
+type Props = {
+  day: dayjs.Dayjs;
```
sean-brydon, line 9: Day is the current grid "DAY" we are looking at as this is iterated over in the main scheduler component

packages/ui/components/scheduler/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,111 @@
...
+  return (
+    <>
+      {day.isBefore(dayjs(), "day") && (
+        <div
+          key={day.format("YYYY-MM-DD")}
+          className="absolute z-40 w-full"
+          style={{
+            top: `var(--one-minute-height)`,
+            zIndex: 60,
+            height: `calc(${(endHour + 1 - startHour) * 60} * var(--one-minute-height))`, // Add 1 to endHour to include the last hour that we add to display the last vertical line
```
sean-brydon, line 33: We can't span for a full 24 hours here so we need to work out how many hours the user actually has in their start/end config

packages/ui/components/scheduler/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,111 @@
...
+              nowComparedToDayStart,
+              60 / gridCellsPerHour
+            )} * var(--one-minute-height) - 2px)`, // We minus the border width to make it 🧹
+          }}>
+          <BlockedTimeCell />
+        </div>
+      )}
+    </>
+  );
+}
```
sean-brydon, line 63: This blocks up to the current time of the day and takes into account grid cells.
e.g. 

- Its currently 12:05 and your grid cells are every 15 minutes
- We block until 12:15 on the grid so you can't be booked outside of your time interval

packages/ui/components/scheduler/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,111 @@
...
+          const eventStart = dayjs(event.start);
+          const eventEnd = dayjs(event.end);
+          const eventDuration = eventEnd.diff(eventStart, "minutes");
+
+          const eventStartHour = eventStart.hour();
+          const eventStartDiff = (eventStartHour - (startHour || 0)) * 60;
+
+          if (!eventStart.isSame(day, "day")) {
+            return null;
+          }
```
sean-brydon, line 88: If the eventStart (in this case blockingEvent) is not the same day as the current `day` we are iterating over - we don't care about it in this context.

packages/ui/components/scheduler/components/blocking/BlockedTimeCell.tsx
```diff
@@ -0,0 +1,15 @@
+import { classNames } from "@calcom/lib";
+
+export function BlockedTimeCell() {
+  return (
+    <div
+      className={classNames("group absolute inset-0 flex h-full flex-col  hover:cursor-not-allowed")}
```
sean-brydon, line 6: `hover:cursor-not-allowed` to give some more indication that the user is not meant to be clicking on these spots.

packages/ui/components/scheduler/components/currentTime/index.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useEffect, useState } from "react";
+
```
sean-brydon, line 2: This component doesn't fully work as of right now but I will still explain the what's going on

packages/ui/components/scheduler/components/currentTime/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [currentTimePos, setCurrentTimePos] = useState<number>(0);
+  const { startHour, endHour } = useSchedulerStore((state) => ({
+    startHour: state.startHour || 0,
+    endHour: state.endHour || 23,
+  }));
+
+  useEffect(() => {
+    // Set the container scroll position based on the current time.
+    let currentMinute = new Date().getHours() * 60;
+    currentMinute = currentMinute + new Date().getMinutes();
```
sean-brydon, line 23: Need to figure out what the difference in minutes is between the current time and the time that is available in the grid

packages/ui/components/scheduler/components/event/Empty.tsx
```diff
@@ -0,0 +1,38 @@
...
+    selectionLength: props.selectionLength,
+    startHour: props.startHour,
+  });
+
+  return (
+    <div
+      className="group w-full"
+      style={{ height: "1.75rem", overflow: "visible" }}
+      onClick={() => onEmptyCellClick && onEmptyCellClick(cellToDate.toDate())}>
+      {hoverEventDuration !== 0 && (
```
sean-brydon, line 23: Ability to turn off hovers in Scheduler props

packages/ui/components/scheduler/components/event/Event.tsx
```diff
@@ -0,0 +1,49 @@
...
+      onClick={() => onEventClick?.(event)} // Note this is not the button event. It is the calendar event.
+      className={classNames(
+        "group  flex h-full w-full flex-col overflow-y-auto rounded-[4px]  py-1 px-[6px] text-xs font-semibold  leading-5 ",
+        event.status === "ACCEPTED" &&
+          !selected &&
+          "border-[1px] border-gray-900 bg-gray-100 text-gray-900 hover:bg-gray-200",
+        event.status === "PENDING" &&
+          !selected &&
+          "border-[1px] border-dashed border-gray-900 bg-white text-gray-900",
+        selected && "border-[1px] border-transparent bg-gray-900 text-white",
```
sean-brydon, line 36: basic styling based on what state the eventStatus is

packages/ui/components/scheduler/components/event/EventList.tsx
```diff
@@ -0,0 +1,93 @@
...
+    startHour: state.startHour,
+    events: state.events,
+    eventOnClick: state.onEventClick,
+  }));
+
+  return (
+    <>
+      {events
+        .filter((event) => {
+          return dayjs(event.start).isSame(day, "day") && !event.allDay; // Filter all events that are not allDay and that are on the current day
```
sean-brydon, line 23: All day events are going to be handled elsewhere in another component

packages/ui/components/scheduler/components/event/EventList.tsx
```diff
@@ -0,0 +1,93 @@
...
+        })
+        .map((event, idx, eventsArray) => {
+          let width = 90;
+          let marginLeft: string | number = 0;
+          let right = 0;
+          let zIndex = 61;
+
```
sean-brydon, line 30: Default styles for non overlapping events

packages/ui/components/scheduler/components/grid/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function SchedulerColumns({ offsetHeight, gridStopsPerDay, children, zIndex }: Props) {
+  return (
+    <ol
+      className="scheduler-grid-row-template col-start-1 col-end-2 row-start-1 grid auto-cols-auto sm:pr-8"
+      style={{ marginTop: offsetHeight || "var(--gridDefaultSize)", zIndex }}
+      data-gridstopsperday={gridStopsPerDay}>
+      {children}
+    </ol>
+  );
+}
```
sean-brydon, line 19: Reusable grid component as we were using it in a few places

packages/ui/components/scheduler/components/grid/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+  offsetHeight: number | undefined;
+  gridStopsPerDay: number;
+  children: React.ReactNode;
+  zIndex?: number;
+};
+
+export function SchedulerColumns({ offsetHeight, gridStopsPerDay, children, zIndex }: Props) {
+  return (
+    <ol
+      className="scheduler-grid-row-template col-start-1 col-end-2 row-start-1 grid auto-cols-auto sm:pr-8"
```
sean-brydon, line 13: `scheduler-grid-row-template` lives in styles.css and requires the data-gridstopsperday to function

packages/ui/components/scheduler/components/heading/SchedulerHeading.tsx
```diff
@@ -0,0 +1,55 @@
...
+          />
+          <Button
+            StartIcon={Icon.FiChevronRight}
+            size="icon"
+            color="secondary"
+            aria-label="Next Week"
+            onClick={() => {
+              handleDateChange("INCREMENT");
+            }}
+          />
```
sean-brydon, line 50: Handle incrementing the users start/EndDate

packages/ui/components/scheduler/components/horizontalLines/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+              {hour.format("h A")}
+            </div>
+          </div>
+        </>
+      ))}
+      <div key={`${id}-${finalHour}`}>
+        <div className="sticky left-0 z-20 -mt-2.5 -ml-14 w-14 pr-2 text-right text-xs leading-5 text-gray-400">
+          {finalHour}
+        </div>
+      </div>
```
sean-brydon, line 36: Might not be the best DX here - as the endHour always ends up being > than what they pass in by one. But this way we get the correct styling and have an easier way to indicate the end of the "Hours"

packages/ui/components/scheduler/components/verticalLines/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+          key={`Key_vertical_${i}`}
+          className="row-span-full"
+          style={{
+            gridColumnStart: i + 1,
+          }}
+        />
+      ))}
+    </div>
+  );
+};
```
sean-brydon, line 19: Handles the vertical lines to separate each day. 

In future this might be handled elsewhere and be handled on one of the many grids we have with a simple divide x but that approach required some messing around with zIndex

packages/ui/components/scheduler/state/store.ts
```diff
@@ -0,0 +1,82 @@
...
+  setEndDate: (endDate: SchedulerComponentProps["endDate"]) => set({ endDate }),
+  setEvents: (events: SchedulerComponentProps["events"]) => set({ events }),
+  // This looks a bit odd but init state only overrides the public props + actions as we don't want to override our internal state
+  initState: (state: SchedulerState & SchedulerPublicActions) => {
+    // Handle sorting of events if required
+    let events = state.events;
+
+    if (state.sortEvents) {
+      events = state.events.sort((a, b) => a.start.getTime() - b.start.getTime());
+    }
```
sean-brydon, line 36: Events need to be sorted - but if you don't pass in it sorted we offer a way to sort them on the fly

packages/ui/components/scheduler/styles/styles.css
```diff
@@ -0,0 +1,17 @@
+.scheduler-wrapper {
+  --gridDefaultSize: 1.75rem;
+  --hoursInDay: 24;
+  --minuteInHour:60;
+  --gridMobileSize: 100vw;
```
sean-brydon, line 5: Set some default CSS values that were coming up a lot across the app

packages/ui/components/scheduler/styles/styles.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.scheduler-grid-row-template{
+  grid-template-rows: var(--gridMobileSize) repeat(attr(data-gridstoperperday number), var(--gridMobileSize)) auto;
+}
+
+@media (min-width: 640px) {
+  .scheduler-grid-row-template {
+    grid-template-rows: var(--gridDefaultSize) repeat(attr(data-gridstoperperday number), var(--gridDefaultSize)) auto;
```
sean-brydon, line 14: Explaining awful CSS: 

Create a grid with rows - every 1.75 rem and repeat it for the number of grid stops in a day.

packages/ui/components/scheduler/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,114 @@
...
+    <>
+      <BlockedBeforeToday day={day} startHour={startHour} endHour={endHour} />
+      <BlockedToday gridCellsPerHour={gridCellsPerHour} day={day} startHour={startHour} endHour={endHour} />
+      {blockingDates &&
+        blockingDates.map((event, i) => {
+          const dayStart = day.startOf("day").hour(startHour);
+          const blockingStart = dayjs(event.start);
+          const eventEnd = dayjs(event.end);
+
+          const eventStart = dayStart.isAfter(blockingStart) ? dayStart : blockingStart;
```
sean-brydon, line 83: We make sure to not calculate the height and top values with anything that is before the start Date

This causes the weird event positioning seen in the loom video

packages/ui/components/scheduler/components/event/EventList.tsx
```diff
@@ -0,0 +1,96 @@
...
+              // figure out which event has the longest duration
+              const prevEventDuration = prevEventEnd.diff(prevEventStart, "minutes");
+              if (prevEventDuration > eventDuration) {
+                zIndex = 65;
+                marginLeft = "auto";
+                right = 8;
+                width = width / 2;
+                if (eventDuration >= 30) {
+                  width = 80;
+                }
```
sean-brydon, line 73: If the event is longer than 30 minutes we set width to 80% 
CleanShot 2022-11-23 at 15 56 21

zomars:
Lowering the endHours results in this column overflow:

image

packages/ui/components/scheduler/components/event/EventList.tsx
zomars: TODO: What happens with more than 3 overlaps?

zomars (dismissed):
LGTM 

Let's fix nitpicks and I can say it's about 90% done. (not counting mobile tho)

packages/ui/components/calendar/components/Calendar.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { HorizontalLines } from "./horizontalLines";
+import { VeritcalLines } from "./verticalLines";
+
+export function Calendar(props: CalendarComponentProps) {
+  const container = useRef<HTMLDivElement | null>(null);
+  const containerNav = useRef<HTMLDivElement | null>(null);
+  const containerOffset = useRef<HTMLDivElement | null>(null);
+  const initalState = useCalendarStore((state) => state.initState);
+
+  const { startDate } = useCalendarStore((state) => ({ startDate: state.startDate }));
```
JeroenReumkens, line 22: Wouldn't `const startDate = useCalendarStore(state => state.startDate)` be possible as well? Makes me think less because we won't have the creating of an object and the destructuring of it later.

packages/ui/components/calendar/components/Calendar.tsx
```diff
@@ -0,0 +1,135 @@
...
+    </MobileNotSupported>
+  );
+}
+
+/** @todo Will be removed once we have mobile support */
+const MobileNotSupported = ({ children }: { children: React.ReactNode }) => {
+  return (
+    <>
+      <div className="flex h-full flex-col items-center justify-center sm:hidden">
+        <h1 className="text-2xl font-bold">Mobile not supported</h1>
```
JeroenReumkens, line 129: Stupid NIT, but perhaps we should add "yet" -> Mobile not supported yet. Might prevent people from getting disappointed by assuming we're not going to build this 😄

packages/ui/components/calendar/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,128 @@
...
+    <>
+      <BlockedBeforeToday day={day} startHour={startHour} endHour={endHour} />
+      <BlockedToday gridCellsPerHour={gridCellsPerHour} day={day} startHour={startHour} endHour={endHour} />
+      {blockingDates &&
+        blockingDates.map((event, i) => {
+          const dayStart = day.startOf("day").hour(startHour);
+          const blockingStart = dayjs(event.start);
+          const eventEnd = dayjs(event.end);
+
+          const eventStart = dayStart.isAfter(blockingStart) ? dayStart : blockingStart;
```
JeroenReumkens, line 97: Should we apply this same logic for the day end? What would happen if the blocked time is longer than the day end?

packages/ui/components/calendar/components/blocking/BlockedList.tsx
```diff
@@ -0,0 +1,128 @@
...
+          const eventDuration = eventEnd.diff(eventStart, "minutes");
+
+          const eventStartHour = eventStart.hour();
+          const eventStartDiff = (eventStartHour - (startHour || 0)) * 60;
+
+          if (!eventStart.isSame(day, "day")) {
```
JeroenReumkens, line 103: Would it be more performant if we add a filter on line 92 where we filter only on blockedevents for today?

packages/ui/components/calendar/components/currentTime/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    endHour: state.endHour || 23,
+  }));
+
+  useEffect(() => {
+    // Set the container scroll position based on the current time.
+    let currentMinute = new Date().getHours() * 60;
+    currentMinute = currentMinute + new Date().getMinutes();
+
+    if (containerOffsetRef.current) {
+      const totalHours = endHour - startHour;
```
JeroenReumkens, line 26: This is unused?

packages/ui/components/calendar/components/currentTime/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      const currentTimePos = currentMinute;
+      setCurrentTimePos(currentTimePos);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [startHour, endHour]);
+
+  return (
+    <div
+      className="absoulte z-40 ml-3 flex h-1 items-center justify-center border-gray-800"
```
JeroenReumkens, line 35: 1. There is a typo in `absolute`
2. Should we make this `aria-hidden`, since I can imagine this time label might only confuse visually impaired users? 🤔  I don't think it will bring any value to them either.

packages/ui/components/calendar/components/event/Empty.tsx
```diff
@@ -0,0 +1,43 @@
...
+    selectionLength: props.selectionLength,
+    startHour: props.startHour,
+  });
+
+  return (
+    <div
+      className="group w-full"
+      style={{ height: "1.75rem", overflow: "visible" }}
+      onClick={() => onEmptyCellClick && onEmptyCellClick(cellToDate.toDate())}>
+      {hoverEventDuration !== 0 && (
```
JeroenReumkens, line 28: Perhaps return null early instead of rendering an empty div that in theory is still clickable?

packages/ui/components/calendar/components/event/EventList.tsx
```diff
@@ -0,0 +1,108 @@
...
+          const eventDuration = eventEnd.diff(eventStart, "minutes");
+
+          const eventStartHour = eventStart.hour();
+          const eventStartDiff = (eventStartHour - (startHour || 0)) * 60 + eventStart.minute();
+          const nextEvent = eventsArray[idx + 1];
+          const prevEvent = eventsArray[idx - 1];
+
+          // Check for overlapping events since this is sorted it should just work.
+          if (nextEvent) {
+            console.log("Next" + nextEvent.title);
```
JeroenReumkens, line 46: 🍕 console

packages/ui/components/calendar/components/event/EventList.tsx
```diff
@@ -0,0 +1,108 @@
...
+  );
+
+  return (
+    <>
+      {events
+        .filter((event) => {
+          return dayjs(event.start).isSame(day, "day") && !event.allDay; // Filter all events that are not allDay and that are on the current day
+        })
+        .map((event, idx, eventsArray) => {
+          let width = 90;
```
JeroenReumkens, line 29: I can imagine that if we'd add these calculations inside a child component (meaning we turn the event wrapper div into a component as well, could be in this file), that we actually improve the performance because React will then only recalculate one child if that changes. 

Perhaps down the line we can even memoize this more, but let's see how that performs first. What do you think?

JeroenReumkens (changes requested):
Amazing work Sean 🔥 Added a few minor comments that I think could be quick wins 🫶

packages/ui/components/calendar/components/Calendar.tsx
sean-brydon: Good shout - we cal also move to the shallow approach and keep all of this in one object

packages/ui/components/calendar/components/Calendar.tsx
sean-brydon: the plan is to not actually render this on mobile at all but we can for sure add this

packages/ui/components/calendar/components/blocking/BlockedList.tsx
sean-brydon: Yep good shout :D Agh good old time logic

packages/ui/components/calendar/components/currentTime/index.tsx
sean-brydon: Very true

packages/ui/components/calendar/components/event/Empty.tsx
sean-brydon: We still need this div to render the grid and also handle the onClick since we moved away from template cells

packages/ui/components/calendar/components/event/EventList.tsx
sean-brydon: 💩

packages/ui/components/calendar/components/event/EventList.tsx
sean-brydon: I agree I will create an issue once this a merged so we can do this in a followup

packages/ui/components/calendar/components/event/Empty.tsx
JeroenReumkens: Makes sense! I'll resolve this one.

JeroenReumkens (approved):
Amazing work Sean 🥰 
I've approved this, but it's wise to get one more approve since I was involved early.

apps/storybook/styles/storybook-styles.css
```diff
@@ -1,3 +1,5 @@
+@import url("../../../packages/features/calendars/weeklyview/styles/styles.css");
```
hariombalhara, line 1: Trying to learn a few things here while reviewing.
Why are we using URL import here whike we are not using it in globals.css

packages/features/calendars/weeklyview/components/Calendar.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { CalendarComponentProps } from "../types/state";
+import { getDaysBetweenDates, getHoursToDisplay } from "../utils";
+import { DateValues } from "./DateValues";
+import { BlockedList } from "./blocking/BlockedList";
+import { EmptyCell } from "./event/Empty";
+import { EventList } from "./event/EventList";
+import { SchedulerColumns } from "./grid";
+import { SchedulerHeading } from "./heading/SchedulerHeading";
+import { HorizontalLines } from "./horizontalLines";
+import { VeritcalLines } from "./verticalLines";
```
hariombalhara, line 14: nit: Typo here

apps/storybook/styles/storybook-styles.css
sean-brydon: Because storybook ... 

Webpack really had an issue when me and @JeroenReumkens were trying to get the base styles setup here. 

So we moved it here (techincally duplicated) but because of the way storybook imports CSS this was the only way i could find to get the styles to apply in the story

PR #5656, opened 2022-11-23
SMS workflow actions for all users

CarinaWolli:
## What does this PR do?

As discussed, we want to make SMS workflow actions free for all users again. Reverts changes from #5330


zomars (approved):
Ship it

PR #5675, opened 2022-11-25
fix: add team member dialog ui updated

### LINKED ISSUE #5611: [CAL-412] Teams Onboarding - UI quick wins
[https://www.loom.com/share/b1732998a2b84a2dacc8a2b9e2d73cbb](https://www.loom.com/share/b1732998a2b84a2dacc8a2b9e2d73cbb)

@Udit-takkar I'll provide designs for these but some of them are possible to tackle even without.

[Create a team → Step 1 - Cal - Live (Figma)](https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/Cal---Live?node-id=26%3A69689&t=hHQFqEMaQohJZFMR-1)

Udit-takkar:
Signed-off-by: Udit Takkar <udit.07814802719@cse.mait.ac.in>

Fixes https://github.com/calcom/cal.com/issues/5611

- fixed the ui of add team member dialog
- fixed the url in form while creating team  ( should cal.com/team/[slug] instead of app.cal.com/team/[slug])

Screenshot 2022-11-25 at 2 55 20 PM

Screenshot 2022-11-25 at 2 55 01 PM

## Type of change

### DIFF
--- a/packages/features/ee/teams/components/MemberInvitationModal.tsx
+++ b/packages/features/ee/teams/components/MemberInvitationModal.tsx
@@ -80,12 +80,7 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
             ""
           )
         }>
-        <Form
-          form={newMemberFormMethods}
-          handleSubmit={(values) => {
-            console.log(values);
-            props.onSubmit(values);
-          }}>
+        <Form form={newMemberFormMethods} handleSubmit={(values) => props.onSubmit(values)}>
           <div className="mt-6 space-y-6">
             <Controller
               name="emailOrUsername"
--- a/packages/features/ee/teams/components/MemberInvitationModal.tsx
+++ b/packages/features/ee/teams/components/MemberInvitationModal.tsx
@@ -33,7 +33,7 @@ type MembershipRoleOption = {
 
 export interface NewMemberForm {
   emailOrUsername: string;
-  role: MembershipRoleOption;
+  role: MembershipRole;
   sendInviteEmail: boolean;
 }
 
@@ -80,7 +80,12 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
             ""
           )
         }>
-        <Form form={newMemberFormMethods} handleSubmit={(values) => props.onSubmit(values)}>
+        <Form
+          form={newMemberFormMethods}
+          handleSubmit={(values) => {
+            console.log(values);
+            props.onSubmit(values);
+          }}>
           <div className="mt-6 space-y-6">
             <Controller
               name="emailOrUsername"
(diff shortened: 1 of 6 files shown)


packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -105,32 +106,34 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
...
             />
             <Controller
               name="sendInviteEmail"
               control={newMemberFormMethods.control}
-              defaultValue={false}
+              defaultValue={true}
               render={() => (
-                <div className="relative flex items-start">
+                <div className="relative  flex items-start">
                   <CheckboxField
```
sean-brydon, line 134: Minor NIT that isn't your fault - I don't think this Field needs to be inside of a controller

sean-brydon (changes requested):
LGTM nice job as usual :) 

Will approve once CI is passing

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -105,32 +106,34 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
             <Controller
               name="role"
               control={newMemberFormMethods.control}
-              defaultValue={options[0]}
-              render={({ field: { onChange } }) => (
+              defaultValue={options[1]}
+              render={({ field: { option, onChange } }) => (
```
sean-brydon, line 110:
```suggestion
              render={({ field: { value, onChange } }) => (
```
Not 100% sure if this is the solution to the type error but Im not sure option exists here

Udit-takkar: fixing it

Udit-takkar:
@sean-brydon what is the difference between owner and admin type ? 

@Jaibles  mentioned in the loom (https://www.loom.com/share/b1732998a2b84a2dacc8a2b9e2d73cbb) to replace admin with owner but i see that both owner and admin are different roles defined in db

export const MembershipRole: {
  MEMBER: 'MEMBER',
  ADMIN: 'ADMIN',
  OWNER: 'OWNER'
};

we should atleast mention what action can a user perform  with these roles.

sean-brydon:
> @sean-brydon what is the difference between owner and admin type ?
> 
> @Jaibles mentioned in the loom ([loom.com/share/b1732998a2b84a2dacc8a2b9e2d73cbb](https://www.loom.com/share/b1732998a2b84a2dacc8a2b9e2d73cbb)) to replace admin with owner but i see that both owner and admin are different roles defined in db
> 
> export const MembershipRole: { MEMBER: 'MEMBER', ADMIN: 'ADMIN', OWNER: 'OWNER' };
> 
> we should atleast mention what action can a user perform with these roles.

If I remember correctly - admin has all perms but unable to disband the team or deal with any payment related tasks

Udit-takkar:
ok should i change the role option admin role here in the invite dialog like before instead of owner?

sean-brydon:
> ok should i change the role option admin role here in the invite dialog like before instead of owner?

I believe so - I don't think it makes sense to add a team member as owner from the creation screen.

Admin/Member should be the default options here I believe

Udit-takkar:
done now

sean-brydon (approved):
🚀

PR #5711, opened 2022-11-27
Fixing calWarnings error on reschedule

leog:
## What does this PR do?

As reported by @CarinaWolli by Threads, there is an error when rescheduling a booking due to undefined `additionalInfo`. This fixes that error.

## Type of change

## How should this be tested?

Try rescheduling and you should not get an error message mentioning `calWarnings`


packages/core/CalendarManager.ts
```diff
@@ -266,7 +266,7 @@ export const updateEvent = async (
           })
       : undefined;
 
-  if (Array.isArray(updatedResult)) {
+  if (Array.isArray(updatedResult) && updatedResult?.additionalInfo) {
```
emrysal, line 269: updatedResult is an array in the && clause, so additionalInfo won't exist there.

packages/core/CalendarManager.ts
```diff
@@ -266,7 +266,7 @@ export const updateEvent = async (
...
     calWarnings = updatedResult.flatMap((res) => res.additionalInfo.calWarnings ?? []);
```
emrysal, line 270:
```suggestion
    calWarnings = updatedResult.flatMap((res) => res.additionalInfo?.calWarnings ?? []);
```

packages/core/CalendarManager.ts
```diff
@@ -266,7 +266,7 @@ export const updateEvent = async (
...
   } else {
     calWarnings = updatedResult?.additionalInfo.calWarnings;
```
emrysal, line 272:
```suggestion
    calWarnings = updatedResult?.additionalInfo?.calWarnings;
```

emrysal (changes requested):

packages/core/CalendarManager.ts
emrysal: Ignore my comments, without knowing exactly what calWarnings is when it is not an array I can't really provide the right suggestions - but it sounds to me like it should always be an array.

emrysal (approved):

PR #5713, opened 2022-11-28
Don't return admin_min error when we're not validating for admin password

JeroenReumkens:
## What does this PR do?

Ensures we don't return the `admin_error` key when we're not validating for the admin password. This broke the website's signup flow. 

## Type of change

## How should this be tested?

- [ ] Go to /signup and ensure the signup form works again (it should create your account). Also ensure the password validation (nr of chars) is still correct.


packages/lib/auth.ts
```diff
@@ -54,7 +54,14 @@ export function isPasswordValid(password: string, breakdown?: boolean, strict?:
       if (password[i] === password[i].toLowerCase()) low = true;
     }
   }
-  return !!breakdown ? { caplow: cap && low, num, min, admin_min } : cap && low && num && min;
```
JeroenReumkens, line 57: We were actually always returning the `admin_min` boolean, even when we shouldn't check for it. This caused the boolean to be always false when strict was disabled.

packages/lib/auth.ts
```diff
@@ -54,7 +54,14 @@ export function isPasswordValid(password: string, breakdown?: boolean, strict?:
...
+
+  if (!breakdown) return cap && low && num && min && (strict ? admin_min : true);
```
JeroenReumkens, line 58: I DID also add the admin comparison to the non breakdown return. I guess that was also missing and should be in there, right? Otherwise the `!breakdown` version might show true incorrectly.

packages/lib/auth.ts
```diff
@@ -54,7 +54,14 @@ export function isPasswordValid(password: string, breakdown?: boolean, strict?:
...
+
+  let errors: Record<string, boolean> = { caplow: cap && low, num, min };
```
JeroenReumkens, line 60: Split things a bit for readability 🥰

sean-brydon (approved):
LGTM - hurt my head trying to understand this but it works :D

JeroenReumkens:
@sean-brydon  Thanks! And yeah, I agree this function could definitely be improved later.

zomars (approved):

PR #5716, opened 2022-11-28
A Playwright test for "Copy Typeform Redirect URL link" action

### LINKED ISSUE #4570: A Playwright test for "Copy Typeform Redirect URL link" action
A test is needed to verify that the CTA is there if Typeform app is installed and it copies correct Redirect URL.

There are tests for Routing Forms already in https://github.com/calcom/cal.com/blob/main/packages/app-store/ee/routing_forms/playwright/tests/basic.e2e.ts 
This new test can go in the same file.

gitstart:
## What does this PR do?

Fixes #4570

image

## Type of change

### DIFF
--- a/packages/app-store/ee/routing-forms/playwright/tests/basic.e2e.ts
+++ b/packages/app-store/ee/routing-forms/playwright/tests/basic.e2e.ts
@@ -8,7 +8,7 @@ async function gotoRoutingLink(page: Page, formId: string) {
   await new Promise((resolve) => setTimeout(resolve, 500));
 }
 
-async function addForm(page: Page) {
+export async function addForm(page: Page) {
   await page.click('[data-testid="new-routing-form"]');
   await page.fill("input[name]", "Test Form Name");
   await page.click('[data-testid="add-form"]');
@@ -41,7 +41,7 @@ async function verifySelectOptions(
   };
 }
 
-async function fillForm(
+export async function fillForm(
   page: Page,
   form: { description: string; field?: { typeIndex: number; label: string } }
 ) {
--- a/packages/app-store/ee/routing-forms/components/FormActions.tsx
+++ b/packages/app-store/ee/routing-forms/components/FormActions.tsx
@@ -135,7 +135,7 @@ export const FormActionsDropdown = ({ form, children }: { form: RoutingForm; chi
   return (
     <dropdownCtx.Provider value={{ dropdown: true }}>
       <Dropdown>
-        <DropdownMenuTrigger asChild>
+        <DropdownMenuTrigger data-testid="form-dropdown" asChild>
           <Button
             type="button"
             size="icon"
--- a/packages/app-store/ee/routing-forms/components/SingleForm.tsx
+++ b/packages/app-store/ee/routing-forms/components/SingleForm.tsx
@@ -120,6 +120,7 @@ const Actions = ({
         {typeformApp?.isInstalled ? (
           <FormActionsDropdown form={form}>
             <FormAction
+              data-testid="copy-redirect-url"
               routingForm={form}
               action="copyRedirectUrl"
               color="minimal"
@@ -172,6 +173,7 @@ const Actions = ({
           </FormAction>
           {typeformApp ? (
             <FormAction
+              data-testid="copy-redirect-url"
               routingForm={form}
               action="copyRedirectUrl"
               color="minimal"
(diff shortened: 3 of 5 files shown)


PeerRich:
@hariombalhara i dont think we need a test for this

hariombalhara:
@PeerRich This is the only test we need to ensure Typeform app is working fine. So, I think we should have it.

@gitstart But the test should be inside Typeform app

gitstart:
> @PeerRich This is the only test we need to ensure Typeform app is working fine. So, I think we should have it.
> 
> @gitstart But the test should be inside Typeform app

Looks like a misunderstanding on our part, will make the required changes and get back to you @hariombalhara

gitstart:
@hariombalhara PR has been updated, please take a look

We updated the test with the Typeform install and hope we are now testing the correct button.
Is it necessary to test within https://www.typeform.com/? If necessary, we don't know how we could do that, as it requires logging into the site, and in addition, the related functionality is paid for.

hariombalhara:
@gitstart Oh sorry for confusion. I actually meant it should be placed inside[typeform app](https://github.com/calcom/cal.com/pull/5716) in our app-store.  I don't mean to actually test typeform webapp.

Right now the test is added in routing-forms app. But a routing form app isn't bundled with Typeform App. Typeform app is installed separately and thus the test needs to be part of that app in our app-store

gitstart:
> @gitstart Oh sorry for confusion. I actually meant it should be placed inside[typeform app](https://github.com/calcom/cal.com/pull/5716) in our app-store. I don't mean to actually test typeform webapp.
> 
> Right now the test is added in routing-forms app. But a routing form app isn't bundled with Typeform App. Typeform app is installed separately and thus the test needs to be part of that app in our app-store

Hey @hariombalhara, is the test still incorrect? Thought we can use Typeform only through the Routing Form
Screenshot 2022-12-13 at 1 27 03 PM

hariombalhara:
Yes, Typeform App can't be used without Routing Forms. But the feature that we are testing belongs to Typeform App only.
So, it would be the responsibility of Typeform App playwright code to install Routing Form and Typeform Apps during testing and  test the appropriate behaviour.

packages/app-store/ee/routing-forms/playwright/tests/basic.e2e.ts
```diff
@@ -106,6 +107,59 @@ test.describe("Routing Forms", () => {
...
+
+        await page.goto(`/apps/routing-forms/forms`);
+        const formId = await addForm(page);
+        await fillForm(page, {
+          description: "",
+          field: { label: "test", typeIndex: 1 },
+        });
+
+        await page.click('button[tooltip="Delete"] ~ button[data-state="closed"]');
+        await page.locator("text=Copy Typeform Redirect URL").click();
```
hariombalhara, line 129: It would be better to use data-testid attribute to identify the element being clicked

packages/app-store/ee/routing-forms/playwright/tests/basic.e2e.ts
```diff
@@ -106,6 +107,59 @@ test.describe("Routing Forms", () => {
...
+        });
+
+        await page.goto(`/apps/routing-forms/forms`);
+        const formId = await addForm(page);
+        await fillForm(page, {
+          description: "",
+          field: { label: "test", typeIndex: 1 },
+        });
+
+        await page.click('button[tooltip="Delete"] ~ button[data-state="closed"]');
```
hariombalhara, line 128: Shouldn't rely on the order of Delete being present before the button. This could target wrong element easily in future.
Let's try to add data-testid directly to the button being clicked

packages/app-store/ee/routing-forms/playwright/tests/basic.e2e.ts
```diff
@@ -106,6 +107,59 @@ test.describe("Routing Forms", () => {
...
+          return navigator.clipboard.readText();
+        });
+        expect(text).toBe(`${CAL_URL}/router?form=${formId}&test={Recalled_Response_For_This_Field}`);
+      });
+
+      test("should copy link in RoutingForms list", async ({ page, context }) => {
+        await page.waitForSelector('[data-testid="empty-screen"]');
+        context.grantPermissions(["clipboard-read", "clipboard-write"]);
+
+        await page.goto(`/apps/typeform`);
```
hariombalhara, line 140: typeform app installation can be abstracted out in a fn like `installTypeformApp` that can be reused

hariombalhara (changes requested):
I added a TODO and moved the tests to a Typeform App describe block. Apart from that added 3  comments that need to be addressed.

hariombalhara (approved):
Looks good. Thank you so much for your contribution 🙏

PR #5727, opened 2022-11-28
Prevent importing anything from /ui from breaking because of window usage in embed

JeroenReumkens:
## What does this PR do?

Because we're now importing a lot more from `/ui`, it seems that code from other files in UI also get's ran (at least in dev?). In this case the window definitions for embed where missing, causing it to error because it was undefined in for example the Console. By adding optional chaining to embed iframe using window object, we prevent it from throwing undefined errors. As a fallback I've added all the default values to false / null as the code would otherwise run as. 

## How should this be tested?

- [ ] The build should be succesful, and this should not have any impact on functionality. Might be wise to test a widget though.

### DIFF
--- a/packages/embeds/embed-core/src/embed-iframe.ts
+++ b/packages/embeds/embed-core/src/embed-iframe.ts
@@ -47,7 +47,7 @@ let isSafariBrowser = false;
 const isBrowser = typeof window !== "undefined";
 
 if (isBrowser) {
-  window.CalEmbed = window.CalEmbed || {};
+  window.CalEmbed = window?.CalEmbed || {};
   window.CalEmbed.embedStore = embedStore;
   const ua = navigator.userAgent.toLowerCase();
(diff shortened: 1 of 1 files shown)


packages/embeds/embed-core/src/embed-iframe.ts
```diff
@@ -228,7 +228,7 @@ export const useIsEmbed = (embedSsr?: boolean) => {
         "Looks like you have iframed cal.com but not using Embed Snippet. Directly using an iframe isn't recommended."
       );
     }
-    setIsEmbed(window.isEmbed());
+    setIsEmbed(window?.isEmbed?.() || false);
```
emrysal, line 231: TIL `?.()` is the way to call an undefinable function.

JeroenReumkens: Optional chaining for functions 🔥

zomars (approved):
Ship it

PR #5823, opened 2022-12-01
Add smsReminderNumber to webhook payload

CarinaWolli:
## What does this PR do?

Adds phone number from SMS workflow actions to webhook payload.


alishaz-polymath:
@CarinaWolli can this be used with custom template as well?

alannnc (approved):

PR #5825, opened 2022-12-01
Requires Confirmation Threshold

### LINKED ISSUE #3964: [CAL-95] FR: make confirmation requirement dependent on notice threshold
### Problem description

Ideally, I would never check the `Requires confirmation` checkbox for any of my event types, since this just adds extra overhead to the booking process, and the whole point of cal.com is to make booking as low friction as possible:

image

However, currently I need to check it, because I want to allow people the _chance_ to book slots with me at short notice (e.g. 3 hours):

image

but if I don't require confirmation then there is a risk that if someone requests a slot in 3 hours from now, I might be busy with other things and not notice the booking notification until after the slot - and of course a no show is embarrassing and unprofessional.

In contrast to this short notice scenario, if someone requests a slot at least 24 hours from now, I can be confident that I will notice the booking notification in time, therefore I don't need to play it safe by requiring a confirmation.

Extend the `Requires confirmation` to something like this:

image

In this example, if the booking is more than 24 hours away from the time it was created, no confirmation would be required, but if it's sooner than 24 hours, then confirmation would be required.

leog:
## What does this PR do?

Introduces the ability to add a threshold to the requires confirmation event type option.

Closes #3964

Loom Video: https://www.loom.com/share/e197de9655e0401c82673d6609107ed9

## Type of change

## How should this be tested?

Go to an event type configuration page under Advanced tab and check Requires Confirmation. It will show up 2 options, "Always" or to set the threshold.


apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -33,22 +33,10 @@ export const EventSetupTab = (
...
-    { value: 75, label: t("multiple_duration_mins", { count: 75 }) },
-    { value: 80, label: t("multiple_duration_mins", { count: 80 }) },
-    { value: 90, label: t("multiple_duration_mins", { count: 90 }) },
-    { value: 120, label: t("multiple_duration_mins", { count: 120 }) },
-    { value: 180, label: t("multiple_duration_mins", { count: 180 }) },
-  ];
+  const multipleDurationOptions = [5, 10, 15, 20, 25, 30, 45, 50, 60, 75, 80, 90, 120, 180].map((mins) => ({
+    value: mins,
+    label: t("multiple_duration_mins", { count: mins }),
+  }));
```
sean-brydon, line 39: 😍

apps/web/components/eventtype/RequiresConfirmationController.tsx
```diff
@@ -0,0 +1,156 @@
...
+    metadata?.requiresConfirmationThreshold
+  );
+  const defaultRequiresConfirmationSetup = { time: 30, unit: "minutes" as UnitTypeLongPlural };
+  const formMethods = useFormContext<FormValues>();
+
+  useEffect(() => {
+    if (!requiresConfirmation) {
+      formMethods.setValue("metadata.requiresConfirmationThreshold", undefined);
+    }
+  }, [requiresConfirmation]);
```
sean-brydon, line 37: Can we just do this in the default state of the form?

leog: We need to dynamically set the threshhold config to undefined in case there is one and requiresConfirmation get disabled

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -744,7 +744,7 @@ const BookingPage = ({
                           </div>
                         </div>
                       )}
-                      {input.options && input.type === EventTypeCustomInputType.RADIO && (
+                      {input.options && input.type === EventTypeCustomInputType.BOOL && (
```
sean-brydon, line 747: Is this right? Below is showing the radio inputs - not sure why its been changed to bool

sean-brydon: Also the input.options below only is available on RADIO type

PeerRich:
this is awesome!

sean-brydon (approved):
LGTM - works locally awesome job @leog

pumfleet (approved):

aspiers:
Great job, thanks everyone!

PR #5852, opened 2022-12-04
fix: use the new docker compose command if it exists

G3root:
## What does this PR do?

## Type of change


zomars (commented):
What happens if both docker-compose and `docker compose` are available?

G3root:
> What happens if both docker-compose and `docker compose` are available?

in that case i thinks only the first command runs. quoting from the [Man Page](https://www.gnu.org/software/bash/manual/html_node/Lists.html) 


“ `command1 || command2`
command2 is executed if, and only if, command1 returns a non-zero exit status.
The return status of AND and OR lists is the exit status of the last command executed in the list. "

zomars (approved):

PR #5859, opened 2022-12-05
BugFix: Link based Apps installation - Opens 2 tabs instead of 1

hariombalhara:
## What does this PR do?

Fixes the issue where installing the Pipedream app opens 2 tabs instead of 1.
 https://www.loom.com/share/bb15f730593c475b9d82990544ffe172

**Environment**: Production

## Type of change

## How should this be tested?

- Try installation of Pipedream app as shown in loom


packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -62,6 +63,7 @@ function useAddAppMutation(_type: App["type"] | null, options?: UseAddAppMutatio
 
     if (!isOmniInstall) {
       gotoUrl(json.url, json.newTab);
+      return;
```
hariombalhara, line 66: Early return. goToUrl is expected to redirect and flow should stop there.

pumfleet (approved):
nice

PR #5959, opened 2022-12-09
event type setting refactor

### LINKED ISSUE #5946: [CAL-569] Additional inputs bugs
As soon as I go to add a SECOND "additional inputs", the first entry gets overwritten BY the second entry. It seems to have something to do with the "REQUIRED" checkbox. See attached video
[https://photos.app.goo.gl/5Xvf3hRu46M37ZZW8](https://photos.app.goo.gl/5Xvf3hRu46M37ZZW8)

Also, "additional inputs" gets turned OFF even after I save it, leave the page, and come back to it. See attached video.

[https://photos.app.goo.gl/ecgW7ygs5y5RYSX18](https://photos.app.goo.gl/ecgW7ygs5y5RYSX18)

After I create "Additional Inputs"... I can't change the order of them. I have to DELETE them and reenter them in the correct order.

Re: Radio buttons... I actually need an option to turn them into checkboxes for MULTIPLE selections as well.

"Other:" should also be a radio/checkbox option, and then allowing a free form field.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-569](https://linear.app/calcom/issue/CAL-569/additional-inputs-bugs)</sub>

Udit-takkar:
Fixes https://github.com/calcom/cal.com/issues/5946

## Type of change


apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -401,7 +401,14 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
                     selectedCustomInput.type = customInput.type;
                     selectedCustomInput.options = customInput.options || undefined;
                     // Update by id
-                    const inputIndex = customInputs.findIndex((input) => input.id === values.id);
+                    const inputIndex = customInputs.findIndex(
+                      (input) =>
+                        input.id === values.id &&
```
hariombalhara, line 406: This seems hackish. id should uniquely identify an element, why do we need to match all props

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -99,33 +92,38 @@ const querySchema = z.object({
...
   });
 
-  const { eventType: dbEventType, locationOptions, team, teamMembers } = props;
-  // TODO: It isn't a good idea to maintain state using setEventType. If we want to connect the SSR'd data to tRPC, we should useQuery(["viewer.eventTypes.get"]) with initialData
-  // Due to this change, when Form is saved, there is no way to propagate that info to eventType (e.g. disabling stripe app doesn't allow recurring tab to be enabled without refresh).
-  const [eventType, setEventType] = useState(dbEventType);
-
-  const router = useRouter();
-  const { tabName } = querySchema.parse(router.query);
+  const { data, isLoading } = trpc.viewer.eventTypes.get.useQuery({ id: props.type });
```
Comment (hariombalhara) [inline apps/web/pages/event-types/[type]/index.tsx:107]: It would be better to move useQuery to a wrapper Component for `EventTypePage` component. So, loading logic would be limited to the wrapper component and thus EventTypePage would have non null eventType. 
This would avoid all the newly added `eventType?` and `eventType&&` things

Comment (hariombalhara) [reply apps/web/pages/event-types/[type]/index.tsx:107]: We should use initialData with useQuery that is obtained from SSR. That makes the first page load really fast with no loader and later navigations to eventType page would be client side rendered.

hariombalhara (changes requested):
Great attempt to improve EventType flow @Udit-takkar !! 
Just need a bit of work

apps/web/components/eventtype/EventAdvancedTab.tsx
Udit-takkar: so id is only assigned after user has clicked save button otherwise a default id is being used -1. so editing multiple custom inputs without saving would work something like this https://photos.google.com/share/AF1QipMCbabiovDIRlru0ASfB0uerBr-aoN4XLG0_xTwAerAzAm1FFeRA4Kp-yf-Yi4bWA/photo/AF1QipP42jgzOHOeraQ45gF0MFcHCRxcEhjfl1dfyMq1?key=ZkZ0ZnZhQ2d0UXNYdE8wSHFmaWJjNTdsaWdGRmp3

It would overwrite other inputs which have same id.

apps/web/pages/event-types/[type]/index.tsx
Comment (Udit-takkar) [reply apps/web/pages/event-types/[type]/index.tsx]: Done

Udit-takkar:
@hariombalhara  thanks for the review. made all the changes you suggested.

apps/web/components/eventtype/EventAdvancedTab.tsx
hariombalhara: Can we generate id(from uuid()) at frontend ? That should solve the problem.

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -1,5 +1,5 @@
 import { SchedulingType } from "@prisma/client/";
-import { EventTypeSetupInfered, FormValues } from "pages/event-types/[type]";
+import { EventTypeSetupProps, FormValues } from "pages/event-types/[type]";
```
hariombalhara, line 2: Let's clean these unused import as well.

packages/trpc/server/routers/viewer/eventTypes.tsx
```diff
@@ -434,15 +440,233 @@ export const eventTypesRouter = router({
       if (!user) {
         throw new TRPCError({ code: "INTERNAL_SERVER_ERROR" });
       }
-      return await ctx.prisma.eventType.findUnique({
+
+      const userSelect = Prisma.validator<Prisma.UserSelect>()({
```
hariombalhara, line 444: This logic is duplicate of getServerSideProps, let's reuse it. Maybe we can create a getEventType fn or something similar that is used by both getServerSideProps and tRPC.

The only difference I see is how 404 and other errors are handled.

hariombalhara (commented):
Looks better now. Left some more comments.

apps/web/components/eventtype/EventAdvancedTab.tsx
Udit-takkar: so i tried generating random id ( can't use uuid() because we want id of type number. i tried something else here).  but a lot of logic in https://github.com/calcom/cal.com/blob/main/packages/trpc/server/routers/viewer/eventTypes.tsx#L38 depends on id being <0 ( -1 which is default).

Udit-takkar: @hariombalhara  okay i was facing some bugs earlier now i am using Math.random to generate id because type required is number only.

added a new optional variable which is being used to detect if custom input has to be created or updated .Earlier we were using id<0 (-1 which is default) to check new custom inputs https://github.com/calcom/cal.com/blob/main/packages/trpc/server/routers/viewer/eventTypes.tsx#L38 

and lastly useEffect to update customInput after saving.

this should work now

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -386,20 +392,22 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
                 selectedCustomInput={selectedCustomInput}
                 onSubmit={(values) => {
                   const customInput: CustomInputParsed = {
-                    id: -1,
+                    id: Math.floor(Math.random() * 10000),
```
PeerRich, line 395: whats going on here? 🤣

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -386,20 +392,22 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
...
                     eventTypeId: -1,
```
PeerRich, line 396: what is this -1?

apps/web/components/eventtype/EventAdvancedTab.tsx
Udit-takkar: we can't use uuid of type string so had to generate id of type number. number of custom inputs won't be too many so there should not be any collision.

Btw this is just temporary id till user click save. real id would be assigned after user click save and custom inputs would re render with correct id

apps/web/components/eventtype/EventAdvancedTab.tsx
Udit-takkar: -1 was used to denote new custom input which was used here to detect whether we have to create a field or update it https://github.com/calcom/cal.com/blob/main/packages/trpc/server/routers/viewer/eventTypes.tsx#L38. i just added another property in the customInputParsed type to denote if we have to create a new input or update it.

Udit-takkar: -1 as id was causing bugs when you create more than one custom input and try to edit them without clicking the save button on top.

apps/web/components/eventtype/EventAdvancedTab.tsx
zomars: Can we keep the number negative? That way we can GUARANTEE avoiding collision

apps/web/components/eventtype/EventAdvancedTab.tsx
zomars: https://github.com/calcom/cal.com/pull/5959#discussion_r1052525718

apps/web/components/eventtype/EventAdvancedTab.tsx
Udit-takkar: It would cause this bug if we use id as -1
https://photos.google.com/share/AF1QipMCbabiovDIRlru0ASfB0uerBr-aoN4XLG0_xTwAerAzAm1FFeRA4Kp-yf-Yi4bWA/photo/AF1QipP42jgzOHOeraQ45gF0MFcHCRxcEhjfl1dfyMq1?key=ZkZ0ZnZhQ2d0UXNYdE8wSHFmaWJjNTdsaWdGRmp3 .

bug is also mentioned here https://github.com/calcom/cal.com/issues/5946. the likelihood of collision is very less as after saving we would get the correct id from server. if you want i can try to further improve the id generation  Math.floor(Math.random() * 10000).

Udit-takkar: @zomars ok added negative

apps/web/components/eventtype/EventAdvancedTab.tsx
hariombalhara: `-1 * Math.random()` could still give the same negative number for 2 different custom inputs.
The simplest solution that I can think of is to use `Date.now()` as 2 different custom inputs are manually added there would always be a time difference and thus 2 different `Date.now `call will have guaranteed different values.

Udit-takkar: @hariombalhara date.now() was the first thing i tried. there was some prisma error because of the size.

Udit-takkar: we can only send 32 bit signed integer as id just updated the random id generator function to this. This function would give the temporary id till the user clicks save and then we would get the correct id from the server which won't have any collision. 

```
const getRandomId = (length = 8) => {
  return (
    -1 *
    parseInt(
      Math.ceil(Math.random() * Date.now())
        .toPrecision(length)
        .toString()
        .replace(".", "")
    )
  );
};
```

packages/trpc/server/routers/viewer/eventTypes.tsx
hariombalhara: @Udit-takkar Can you take care of this as well.

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -51,7 +63,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
   const [selectedCustomInputModalOpen, setSelectedCustomInputModalOpen] = useState(false);
   const [requiresConfirmation, setRequiresConfirmation] = useState(eventType.requiresConfirmation);
   const placeholderHashedLink = `${CAL_URL}/d/${hashedUrl}/${eventType.slug}`;
-
+  console.log(customInputs);
```
alannnc, line 66:
```suggestion

```

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -386,20 +403,22 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
                 selectedCustomInput={selectedCustomInput}
                 onSubmit={(values) => {
                   const customInput: CustomInputParsed = {
-                    id: -1,
+                    id: getRandomId(),
```
alannnc, line 406: Can we use uuid() generators instead of random number?

Udit-takkar: No because the type of id is number and not string

alannnc (commented):
Can we do loom? Or self review? I'm having trouble to understand why some changes are needed.

Udit-takkar:
@alannnc anything specific do you wanna know about?

packages/lib/getEventTypeById.ts
```diff
@@ -0,0 +1,264 @@
+import { Prisma, PrismaClient } from "@prisma/client";
```
Udit-takkar, line 1: this is just a common function which is used in getServersideprops and trpc end point. refer here https://github.com/calcom/cal.com/pull/5959#discussion_r1050806695 on why we are using it twice.

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -64,6 +75,12 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
     !hashedUrl && setHashedUrl(generateHashedLink(eventType.users[0]?.id ?? team?.id));
   }, [eventType.users, hashedUrl, team?.id]);
 
+  useEffect(() => {
+    if (eventType.customInputs) {
+      setCustomInputs(eventType.customInputs.sort((a, b) => a.id - b.id));
```
Udit-takkar, line 80: this makes sure the custom inputs are re rendered when we click save button.

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -36,7 +36,19 @@ const generateHashedLink = (id: number) => {
   return uid;
 };
 
-export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered, "eventType" | "team">) => {
+const getRandomId = (length = 8) => {
```
Udit-takkar, line 39: customInput type requires the id to be of type number and within 32 bit integer so i created this random id generator function. This is the temporary id we'll be using until user clicks save and server save these custom inputs and return correct id.  

our current approach of using id as -1 was causing this bug https://photos.google.com/share/AF1QipMCbabiovDIRlru0ASfB0uerBr-aoN4XLG0_xTwAerAzAm1FFeRA4Kp-yf-Yi4bWA/photo/AF1QipP42jgzOHOeraQ45gF0MFcHCRxcEhjfl1dfyMq1?key=ZkZ0ZnZhQ2d0UXNYdE8wSHFmaWJjNTdsaWdGRmp3

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -386,20 +403,22 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupInfered
...
                   const customInput: CustomInputParsed = {
-                    id: -1,
+                    id: getRandomId(),
                     eventTypeId: -1,
                     label: values.label,
                     placeholder: values.placeholder,
                     required: values.required,
                     type: values.type,
                     options: values.options,
+                    hasToBeCreated: true,
```
Udit-takkar, line 413: this variable would be used in handleCustomInputs(https://github.com/calcom/cal.com/blob/main/packages/trpc/server/routers/viewer/eventTypes.tsx#L38) to detect if we have to create a custom input. we can also choose to not make this change this but this makes it more obvious that we are creating custom input.

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -101,48 +100,3 @@ export const EventTeamTab = ({
     </div>
   );
 };
-
-const TeamProfileCard = ({
```
Udit-takkar, line 105: A lot of stuff was not used anywhere. so i removed it

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -109,34 +105,34 @@ const querySchema = z.object({
...
+  const utils = trpc.useContext();
+  const router = useRouter();
+  const { tabName } = querySchema.parse(router.query);
+
   const { data: eventTypeApps } = trpc.viewer.apps.useQuery({
     extendsFeature: "EventType",
   });
 
-  const { eventType: dbEventType, locationOptions, team, teamMembers } = props;
-  // TODO: It isn't a good idea to maintain state using setEventType. If we want to connect the SSR'd data to tRPC, we should useQuery(["viewer.eventTypes.get"]) with initialData
```
Comment (Udit-takkar) [inline apps/web/pages/event-types/[type]/index.tsx:121]: i basically did this stuff which was also mentioned in TODO.

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -326,6 +315,18 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
   );
 };
 
+const EventTypePageWrapper = (props: inferSSRProps<typeof getServerSideProps>) => {
```
Comment (Udit-takkar) [inline apps/web/pages/event-types/[type]/index.tsx:318]: this wrapper makes sure EventTypePage is loaded only when data is not null or undefined else we would have to use too much of optional chaining and && operator. ( suggested by  hariom)

packages/trpc/server/routers/viewer/eventTypes.tsx
```diff
@@ -36,30 +37,29 @@ function handlePeriodType(periodType: string | undefined): PeriodType | undefine
 }
 
 function handleCustomInputs(customInputs: CustomInputSchema[], eventTypeId: number) {
-  const cInputsIdsToDelete = customInputs.filter((input) => input.id > 0).map((e) => e.id);
+  const cInputsIdsToDeleteOrUpdated = customInputs.filter((input) => !input.hasToBeCreated);
```
Udit-takkar, line 40: we were filtering same thing twice so i just created another variable

Udit-takkar (commented):
Self review done. @alannnc  do let me know if i missed something

packages/trpc/server/routers/viewer/eventTypes.tsx
Udit-takkar: @hariombalhara done

hariombalhara:
Screenshot 2022-12-22 at 1 15 53 PM
I see hydration errors in console for event-types/[type] page

Udit-takkar:
https://user-images.githubusercontent.com/53316345/209204464-d9ed4699-1f0d-4fdc-9c3f-f861d306ae15.mov

@hariombalhara  I don't get any hydration error

hariombalhara:
@Udit-takkar can you try refreshing the page

Udit-takkar:
@hariombalhara yeah on refreshing i get hydration error but this error is also on the main branch. what are possible fix for this?

hariombalhara:
You would need to identify the difference b/w  server rendered content and client rendered content. That difference is the cause for hydration error. 
Error message should help. 

Though if it's in main, it isn't a blocker. But as  we are improving event-type page, you can tackle it in this PR itself

hariombalhara (approved):
Approving it now. The only issue that I see now is hydration error which is already in production and can be fixed separately 

Awesome work @Udit-takkar 🙏

PR #5991, opened 2022-12-12
Feature/date overrides

### LINKED ISSUE #2199: [CAL-505] availability: date overwrite
the date overwrite feature needs more visuals, @Jaibles 

image

<sub>[CAL-505](https://linear.app/calcom/issue/CAL-505/availability-date-overwrite)</sub>

emrysal:
## What does this PR do?

Fixes #2199

* Implements UI for Date overrides + logic to disable/enable getUserAvailability
* Changes the "md" dialog size from `35` to `48`
* Fixes broken layout of the TeamAvailabilityModal

TODO:
- [x] There's a date shift on existing date overrides whilst saving additional date overrides
- [x] Automatically disable dates before minDate if includedDates is omitted
- [x] Implement month change
- [x] Better inclusion/exclusion (when is a date override disabling an existing available date, and when is it enabling an unavailable date)
- [x] Dates are being overridden in the booker
- [x] Fix issue that on save it creates separate list entries for each of the date overrides happening on the same day if you added multiple.


emrysal:
@Jaibles I'm going over the behaviour of inclusive/exclusive and struggling a bit; the issue is the following:

Given you have a working hour schedule from 09:00-17:00 and you want to enable 18:00-20:00 on the 20th, you'd do this:
image

In this situation what needs to happen is that it should be clear that you are adding an extra slot, extra availability.

On the same day however if you put in a date override, 15:00-16:00 - it becomes the other way around, and it's disabled instead.

image

I'm wondering if there is a way we can clarify the behaviour more explicitly through the UX.

Ps. Ignore the DatePicker open slots right now before the 13th - this was me playing about with visuals.

Jaibles:
Shared in Threads but sharing here for visibility also 
[Loom](https://www.loom.com/share/edf5ee5556334f598476d521cb9b4ac5) | [Figma](https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/Cal---Live?node-id=26%253A46001&t=Q6feRImAQlxcYNNW-1)

zomars (commented):
Added some suggestions 🙏🏽

apps/web/public/static/locales/en/common.json
```diff
@@ -1431,7 +1431,6 @@
   "disable_app": "Disable App",
   "disable_app_description": "Disabling this app could cause problems with how your users interact with Cal",
   "edit_keys": "Edit Keys",
-  "apps_description": "Enable apps for your instance of Cal",
```
emrysal, line 1434: Duplicate key in main.

sean-brydon:
Awesome work this works pretty much spot on 
Fixing a few UI bugs but here is some general issues i've found and will update if i find more

Adding two to one day breaks things 
CleanShot 2022-12-14 at 11 18 54@2x

emrysal:
@sean-brydon Implemented excludedDates which addresses the NIT; by disabling adding new dates on a given date there is no way to create entries on the same date.

apps/web/public/static/locales/en/common.json
emrysal: 😖 - undoing

sean-brydon (approved):
Works perfectly - LGTM :) 

Few nits as mentioned in dms but they can be picked up at a later date and are not blocking

packages/features/schedules/components/DateOverrideInputDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+  end: new Date(dayjs.utc().hour(0).minute(0).second(0).format()),
+};
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
+
+const DateOverrideForm = ({
+  value,
+  workingHours,
+  excludedDates,
```
joeauyeung, line 33: What if there's a use case where someone's availability is Monday - Thursday but they want to open up some dates on a Friday? Right now we can't choose dates that are not a part of the original schedule.

packages/features/schedules/components/DateOverrideInputDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+import { DayRanges, TimeRange } from "./Schedule";
+
+const ALL_DAY_RANGE = {
+  start: new Date(dayjs.utc().hour(0).minute(0).second(0).format()),
+  end: new Date(dayjs.utc().hour(0).minute(0).second(0).format()),
+};
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
```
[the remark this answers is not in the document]
PeerRich: lets open issues to follow up on this @emrysal

packages/features/schedules/components/DateOverrideInputDialog.tsx
emrysal: Yes, the scope of this feature is to override available dates - it's not intended to open up dates. That's a future improvement - I tried to embed that in the current design but honestly I shouldn't have and it didn't work right.

Jaibles: while it would be great if we could do what you're stating @joeauyeung, fundamentally we're overriding availability ... not overriding unavailability. If people want to make some Fridays available, we do have the ability to create multiple availability schedules, you can make a new one.

PR #5997, opened 2022-12-13
#5613 Optimise social images by utilising next image api.

### LINKED ISSUE #5613: [CAL-415] Social Image no longer working on WhatsApp after commit: File size too big
### Issue Summary

In the past, sharing my [Cal.com](http://Cal.com) link on WhatsApp showed the social image/`og:image` properly, until October 17. All my subsequent shares don't show the social image.

The problem appears to be because of file-size. [WhatsApp's Social Images are limited to a maximum of 300kb](https://stackoverflow.com/a/39182227) in size, where the current file size of the `og:image` that the `/api/social/og/image` API returns (at least, on my personal [Cal.com](http://Cal.com) link) clocks in at **771kb**.

Looks like the commits that caused the change was probably [5c01467](https://github.com/calcom/cal.com/commit/5c01467caab0e49ac517d1a624c6859587945861).

*Ultimately, the file size of the image produced by the* `/api/social/og/image` needs to be reduced to 300kb or less, or it won't show up when sharing on WhatsApp.

### Steps to Reproduce

1. Share personal [Cal.com](http://Cal.com) link on WhatsApp
2. Title and description information will correctly be pulled from meta-data, but social image will not show

*Why do you consider this a bug and what did you expect to happen instead?*

WhatsApp is used by 2 billion people. Would be great if the social images could show up again like they used to. Feature regression isn't super fun.

### Technical details

* Android 10
* WhatsApp version 2.22.23.77

JeroenReumkens:
## What does this PR do?

Optimises our OG Images by leveraging Next's Image API. This reduces the image size from 500KB to 80KB 🤯

Fixes #5613

## Type of change

## How should this be tested?

- [ ] Verify that the OG images are still generated properly by checking the OG Image meta tag and visiting that url on the following pages: Meeting booking page (pro/30min), App detail page, any other page on app.cal.com (should show a generic image).

The OG meta tag looks like this: 
CleanShot 2022-12-13 at 11 15 58@2x

To verify that the link is working you can double click that "content" attribute and copy+paste that url in a new tab. The image should show up 🔥


JeroenReumkens:
@shuding Would also love to hear your opinion about this! I couldn't find any other solution to optimise the images, but this seems to work really well! Thanks 🙏

shuding:
Wow this is an interesting usage, looks legit! Ideally the generated image itself should be smaller, we will improve that in future versions. 🙏

PeerRich (approved):

apps/web/pages/api/social/og/image.tsx
```diff
@@ -1,5 +1,5 @@
 import { ImageResponse } from "@vercel/og";
-import { NextApiRequest } from "next";
+import { NextApiRequest, NextApiResponse } from "next";
```
hariombalhara, line 2: Unused import

packages/lib/OgImages.tsx
```diff
@@ -49,29 +52,35 @@ export const constructMeetingImage = ({ title, users = [], profile }: MeetingIma
...
     `&description=${encodeURIComponent(description)}`,
     // Joining a multiline string for readability.
   ].join("");
+
+  return encodeUri ? encodeURIComponent(url) : url;
 };
 
-export const constructGenericImage = ({ title, description }: GenericImageProps) => {
-  return [
+export const constructGenericImage = ({ title, description }: GenericImageProps, encodeUri = true) => {
```
hariombalhara, line 75: I don't see `encodeUri` param being used.

yarn.lock
```diff
@@ -2,67 +2,13 @@
 # yarn lockfile v1
 
 
-"@47ng/cloak@^1.1.0":
```
hariombalhara, line 5: yarn.lock seems to have changed a lot. Are we sure there is nothing else here apart from og-image deps.

hariombalhara (commented):
This is definitely interesting use of _next  🙌  I wasn't aware of this difference b/w these 2 types of servings.

Left some comments.

pumfleet (approved):
Aside from Hariom's comments, code looks good to me. Great optimisation!

yarn.lock
JeroenReumkens: Oh you're right. Overlooked that. I'll run yarn again in a sec. Cloned stuff from scratch, that's probably why there's a lot of changes.

JeroenReumkens: Nope. Even after resetting the lock file and only updating `@vercel/og` it does apply **all** these changes in the yarn.lock. Weird. I guess the lock file simply hasn't been committed for a long time? 🤔

hariombalhara: Yeah it's possible.

packages/lib/OgImages.tsx
hariombalhara: It's a small change. I am okay either way.

My preference personally would have been to not add the parameter if it isn't being actually used

hariombalhara (approved):

PR #6010, opened 2022-12-13
fix: use callback url in signup

### LINKED ISSUE #4383: [CAL-105] /onboarding - View of app before steps appear
There's still a long view of the app before onboarding starts.. can we do anything about that?

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/4383

https://user-images.githubusercontent.com/53316345/207410279-a33d80da-c54a-4e7d-9765-56e2e4cc273c.mov

## Type of change

## How should this be tested?

1) Create an account in /signup 
2) then try logging in you should be redirected to /login


Udit-takkar:
@zomars  although this fixes the issue for all new users signing up. do suggest if we should go with another approach like using cookies here.

2) what is the purpose of login instead button ? both sign up and login instead seem to be working the same way.

PeerRich:
can we auto-log in people after account creation, too?

Udit-takkar:
@PeerRich yeah i can try adding login logic after sign up but what is the use of login instead button in signup page ?

zomars:
> can we auto-log in people after account creation, too?

Agreed we should go straight to onboarding after account creation.

apps/web/pages/signup.tsx
```diff
@@ -54,7 +54,9 @@ export default function Signup({ prepopulateFormValues }: inferSSRProps<typeof g
       .then(handleErrors)
       .then(async () => {
         await signIn("Cal.com", {
-          callbackUrl: router.query.callbackUrl ? `${WEBAPP_URL}/${router.query.callbackUrl}` : WEBAPP_URL,
+          callbackUrl: router.query.callbackUrl
```
hariombalhara, line 57: The way I understand, Onboarding should be shown only to users who haven't completed it. 

How does taking user to onboarding if `callbackUrl` isn't there, solves the problem.

PeerRich:
> > can we auto-log in people after account creation, too?
> 
> Agreed we should go straight to onboarding after account creation.

yup. straight to onboarding

apps/web/pages/signup.tsx
zomars: Yeah this is logic to be solved elsewhere. Not here

Udit-takkar: @hariombalhara we already have logic to check if user has completed onboarding or not and display onboarding steps. it's just that it sometimes takes few seconds to redirect to detect user has not onboarded and redirect to /getting-started.

Udit-takkar:
https://user-images.githubusercontent.com/53316345/208090654-bd033eef-d084-4d75-bab7-4bb2972d09f9.mov

@PeerRich  should auto login now

PeerRich (approved):

PeerRich:
approved, waiting for @zomars final review

zomars (approved):
Thanks @Udit-takkar any clues on why the hash-my-url tests are failing?

Udit-takkar:
@zomars  all tests are passing now

PR #6017, opened 2022-12-13
feature/rate-limit-middleware

alannnc:
## What does this PR do?

Adds new middleware and using it for test action workflow with custom parameters

## Loom
https://www.loom.com/share/3257b1e16d27469a8919cdc3a026e570

## Type of change

## How should this be tested?

- [ ] Create new workflow
- [ ] Go to Test action section and use it whether sms or email.
- [ ] Should trigger rate-limit error when using it 3 times in a row in 10 seconds

## Checklist


packages/features/ee/workflows/components/WorkflowStepContainer.tsx
```diff
@@ -134,10 +134,16 @@ export default function WorkflowStepContainer(props: WorkflowStepProps) {
       showToast(t("notification_sent"), "success");
     },
     onError: (err) => {
+      let message = t("unexpected_error_try_again");
```
alannnc, line 137: default message

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -801,7 +801,7 @@ export const workflowsRouter = router({
         workflow,
       };
     }),
-  testAction: authedProcedure
+  testAction: authedRateLimitedProcedure({ intervalInMs: 10000, limit: 3 })
```
alannnc, line 804: custom options for this trpc action

packages/trpc/server/trpc.ts
```diff
@@ -41,10 +43,41 @@ const isAdminMiddleware = t.middleware(({ ctx, next }) => {
...
+interface IRateLimitOptions {
+  intervalInMs: number;
+  limit: number;
+}
+const isRateLimitedByUserIdMiddleware = ({ intervalInMs, limit }: IRateLimitOptions) =>
+  t.middleware(({ ctx, next }) => {
+    // validate user exists
+    if (!ctx.user) {
+      throw new TRPCError({ code: "UNAUTHORIZED" });
+    }
```
zomars, line 55: Tagging @KATT to give some insight. Is there a way to avoid having to re-check this on every middleware? As you can see we already have an `isAuthedMiddleware` but when using the procedure the `ctx.user` inference is lost.

zomars (approved):
Ship it

packages/trpc/server/trpc.ts
```diff
@@ -41,10 +43,41 @@ const isAdminMiddleware = t.middleware(({ ctx, next }) => {
...
 export const router = t.router;
 export const mergeRouters = t.mergeRouters;
 export const middleware = t.middleware;
 export const publicProcedure = t.procedure.use(perfMiddleware);
 export const authedProcedure = t.procedure.use(perfMiddleware).use(isAuthedMiddleware);
+export const authedRateLimitedProcedure = ({ intervalInMs, limit }: IRateLimitOptions) =>
+  t.procedure
+    .use(perfMiddleware)
+    .use(isAuthedMiddleware)
+    .use(isRateLimitedByUserIdMiddleware({ intervalInMs, limit }));
```
KATT, line 81: @zomars just inline the logic here if you wanna skip the extra check, it'll be inferred

packages/trpc/server/trpc.ts
KATT: https://github.com/calcom/cal.com/pull/6017/files#r1047771652

packages/trpc/server/trpc.ts
zomars: WDYM?

KATT: Instead of having a standalone middleware, you can have an inline one

KATT: Just a callback fn instead of the isRateLimitedThing

zomars: I'm failing to understand how to use isAuthedMiddleware inline and keeping the ctx inference.

KATT: It just won't be reusable outside of this

PR #6024, opened 2022-12-14
fix: Parameterize Stripe Payment Fee configs

### LINKED ISSUE #6023: Stripe Payment Fee parameters are not accessible via environment parameters
### Issue Summary

Configuration of the default-seeded Stripe app is parametrizable via environment variables from here:
https://github.com/calcom/cal.com/blob/main/packages/prisma/seed-app-store.ts#L272-L287

However, it seems to be a bug that Payment Fee-related parameters are hardcoded and not also equivalently paramaterizable.

### Steps to Reproduce

1. Setup a standard Cal.com instance
2. Try to configure the Stripe app's Payment Fee parameters via environment variables, like other Stripe app parameters

**Expected Behavior:** This is possible
**Actual Behavior:** This is not possible

### Technical details

Cal.com [v2.3.8](https://github.com/calcom/cal.com/releases/tag/v2.3.8)

joshavant:
## What does this PR do?

This adds parameterization for Stripe Payment Fee-related configuration values, matching all other Stripe configuration value behavior.

Fixes #6023 

## Type of change

## How should this be tested?

I added the new parameters to an `env_file` passed into a Docker image generated from this branch.

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works (There doesn't seem to be existing tests for this)


.env.example
```diff
@@ -101,6 +101,8 @@ STRIPE_TEAM_MONTHLY_PRICE_ID=
 STRIPE_WEBHOOK_SECRET=
 STRIPE_PRIVATE_KEY=
 STRIPE_CLIENT_ID=
+STRIPE_PAYMENT_FEE_FIXED=
```
hariombalhara, line 104: @zomars I think we can remove thse variables from here. They should be part of .appStore.example only right?

.github/workflows/e2e-embed.yml
```diff
@@ -33,8 +33,8 @@ jobs:
       STRIPE_PRIVATE_KEY: ${{ secrets.CI_STRIPE_PRIVATE_KEY }}
       STRIPE_CLIENT_ID: ${{ secrets.CI_STRIPE_CLIENT_ID }}
       STRIPE_WEBHOOK_SECRET: ${{ secrets.CI_STRIPE_WEBHOOK_SECRET }}
-      PAYMENT_FEE_PERCENTAGE: 0.005
-      PAYMENT_FEE_FIXED: 10
+      STRIPE_PAYMENT_FEE_PERCENTAGE: ${{ secrets.CI_STRIPE_PAYMENT_FEE_PERCENTAGE }}
```
hariombalhara, line 36: We would probably need to set these secrets. Blocking till it's confirmed.

hariombalhara (changes requested):
Looks good to me. 
Blocking till the secrets related comment is resolved. Also, it being an important flow, I would wait for @zomars review before merge.

.github/workflows/e2e-embed.yml
PeerRich: @zomars I think we need those

zomars: I think he can keep them hardcoded in here. They're not affecting the actual app

.env.example
joshavant: @zomars If you let me know what to do here, happy to push the commits.

.github/workflows/e2e-embed.yml
PeerRich: yes

.github/workflows/e2e.yml
```diff
@@ -38,8 +38,8 @@ jobs:
       STRIPE_PRO_PLAN_PRODUCT_ID: ${{ secrets.CI_STRIPE_PRO_PLAN_PRODUCT_ID }}
       STRIPE_PREMIUM_PLAN_PRODUCT_ID: ${{ secrets.CI_STRIPE_PREMIUM_PLAN_PRODUCT_ID }}
       STRIPE_FREE_PLAN_PRODUCT_ID: ${{ secrets.CI_STRIPE_FREE_PLAN_PRODUCT_ID }}
-      PAYMENT_FEE_PERCENTAGE: 0.005
-      PAYMENT_FEE_FIXED: 10
+      STRIPE_PAYMENT_FEE_PERCENTAGE: ${{ secrets.CI_STRIPE_PAYMENT_FEE_PERCENTAGE }}
+      STRIPE_PAYMENT_FEE_FIXED: ${{ secrets.CI_STRIPE_PAYMENT_FEE_FIXED }}
```
zomars, line 42: Same here. hard-code

joshavant:
Addressed all feedback and rebased on `main`.

.env.appStore.example
```diff
@@ -53,12 +53,12 @@ SLACK_CLIENT_ID=
...
-STRIPE_WEBHOOK_SECRET=         # whsec_...
-STRIPE_CLIENT_ID=              # ca_...
-PAYMENT_FEE_FIXED=10           # Take 10 additional cents commission
-PAYMENT_FEE_PERCENTAGE=0.005   # Take 0.5% commission
+NEXT_PUBLIC_STRIPE_PUBLIC_KEY=        # pk_test_...
+STRIPE_PRIVATE_KEY=                   # sk_test_...
+STRIPE_WEBHOOK_SECRET=                # whsec_...
+STRIPE_CLIENT_ID=                     # ca_...
+STRIPE_PAYMENT_FEE_FIXED=10           # Take 10 additional cents commission
+STRIPE_PAYMENT_FEE_PERCENTAGE=0.005   # Take 0.5% commission
```
zomars, line 61: Can we revert these name changes? I know this should be named `STRIPE_` but we are going to prioritize backwards compatibility in this case. Not doing so will cause much unexpected behavior for self-hosted users. Doing so will allow us to merge without user action needed.

zomars (approved):
LGTM

packages/prisma/seed-app-store.ts
```diff
@@ -281,8 +281,8 @@ export default async function main() {
     await createApp("stripe", "stripepayment", ["payment"], "stripe_payment", {
       client_id: process.env.STRIPE_CLIENT_ID,
       client_secret: process.env.STRIPE_PRIVATE_KEY,
-      payment_fee_fixed: process.env.PAYMENT_FEE_FIXED,
-      payment_fee_percentage: process.env.PAYMENT_FEE_PERCENTAGE,
+      payment_fee_fixed: Number(process.env.PAYMENT_FEE_FIXED),
+      payment_fee_percentage: Number(process.env.PAYMENT_FEE_PERCENTAGE),
```
joshavant, line 285: Went to do a functional test of this and realized this was necessary, too.

Now everything is working. :)

PR #6034, opened 2022-12-15
Typed query hook

sean-brydon:
A hook that allows you to have type safe router queries. This was a nightmare - was working on something a lot messier than this for booking limits and decided to spend some time cleaning it up and learning a thing or two about conditional generics 

Pass in a schema to the hook that you would like to keep type safe
e.g. 
CleanShot 2022-12-15 at 11 30 32@2x

results will be fully typesafe
CleanShot 2022-12-15 at 11 31 52@2x
Setting values 
CleanShot 2022-12-15 at 11 32 31@2x
remove by key
CleanShot 2022-12-15 at 11 32 50@2x
push item to key
CleanShot 2022-12-15 at 11 33 08@2x
Remove by key and value
CleanShot 2022-12-15 at 11 33 36@2x


packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // push item to existing key
+  function pushItemToKey<J extends SchemaKeys>(
+    key: J,
+    value: InferedSchema[J] extends Array<unknown>
+      ? InferedSchema[J][0]
+      : InferedSchema[J] extends Array<unknown> | undefined
+      ? NonNullable<InferedSchema[J]>[0]
+      : InferedSchema[J]
+  ) {
```
sean-brydon, line 29: In short 
if its array use the first type if not use the type itsself 

oh and also checking if its undefined

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
+import { useRouter } from "next/router";
+import { z } from "zod";
+
+export function useTypedQuery<T extends z.Schema>(schema: T) {
+  type InferedSchema = z.infer<typeof schema>;
+  type SchemaKeys = keyof InferedSchema;
+  const { query: unparsedQuery, ...router } = useRouter();
+  const parsedQuery = schema.parse(unparsedQuery) as InferedSchema;
```
KATT, line 8: You likely want to use `safeParse` here.

You don't have full control of the query params - someone can edit them manually, failing this aggressively might not be what you want

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // Set the query based on schema values
+  function setQuery<J extends SchemaKeys>(key: J, value: InferedSchema[J]) {
+    router.replace({ query: { ...parsedQuery, [key]: value } }, undefined, { shallow: true });
+  }
```
KATT, line 13: `useCallback` might be an idea or overengineering

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+import { z } from "zod";
+
+export function useTypedQuery<T extends z.Schema>(schema: T) {
+  type InferedSchema = z.infer<typeof schema>;
+  type SchemaKeys = keyof InferedSchema;
+  const { query: unparsedQuery, ...router } = useRouter();
+  const parsedQuery = schema.parse(unparsedQuery) as InferedSchema;
+
+  // Set the query based on schema values
+  function setQuery<J extends SchemaKeys>(key: J, value: InferedSchema[J]) {
```
KATT, line 11: You might want to do this as a `Partial<InferedSchema[J]>`

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+    router.replace({ query: { ...parsedQuery, [key]: value } }, undefined, { shallow: true });
+  }
+
+  // Delete a key from the query
+  function removeByKey(key: SchemaKeys) {
```
KATT, line 16: This is not safe since you allow to remove keys that are required, you likely wanna filter and only allow removal of optional keys

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+    const { [key]: _, ...newQuery } = parsedQuery;
+    router.replace({ query: newQuery }, undefined, { shallow: true });
+  }
+
+  // push item to existing key
+  function pushItemToKey<J extends SchemaKeys>(
+    key: J,
+    value: InferedSchema[J] extends Array<unknown>
+      ? InferedSchema[J][0]
```
KATT, line 25:
```suggestion
      ? InferedSchema[J][number]
```

KATT: This won't work for tuples but that is prob okay

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // Delete a key from the query
+  function removeByKey(key: SchemaKeys) {
+    const { [key]: _, ...newQuery } = parsedQuery;
+    router.replace({ query: newQuery }, undefined, { shallow: true });
+  }
+
+  // push item to existing key
+  function pushItemToKey<J extends SchemaKeys>(
+    key: J,
```
KATT, line 23: You can filter this to only get the keys that rep arrays

zomars (commented):
Pretty impressive @sean-brydon will need a little bit of more time to review properly

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+// Take array as a string and return zod array
+export const queryNumberArray = z
+  .string()
+  .or(z.number())
+  .or(z.array(z.number()))
+  .transform((a) => {
+    if (typeof a === "string") return a.split(",").map((a) => Number(a));
+    if (Array.isArray(a)) return a;
+    return [a];
+  });
```
sean-brydon, line 14: Support number arrays as a final "type" but input can be string so we can parse it into a number array 

query params come in as strings before being parsed

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+  else if (!parsedQuerySchema.success) console.error(parsedQuerySchema.error);
+
+  // Set the query based on schema values
+  const setQuery = useCallback(
+    function setQuery<J extends SchemaKeys>(key: J, value: Partial<InferedSchema[J]>) {
+      // Remove old value by key so we can merge new value
+      const { [key]: _, ...newQuery } = parsedQuery;
+      const newValue = { ...newQuery, [key]: value };
+      const search = new URLSearchParams(newValue).toString();
+      router.replace({ query: search }, undefined, { shallow: true });
```
sean-brydon, line 51: Had to do this to prevent duplicate values Nextjs actually recommends you construct query params this way

packages/lib/hooks/useTypedQuery.ts
sean-brydon: I honestly couldnt get this working - I was trying the ArrayKeys type above but was getting an error about not being able to set J to typeof<t> 

Maybe @zomars has more TS knowledge to debug this

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+  type ArrayOnlyKeys = {
+    [K in keyof InferedSchema]: InferedSchema[K] extends Array<unknown> & undefined ? K : never;
+  };
+
+  const { query: unparsedQuery, ...router } = useRouter();
+  const parsedQuerySchema = schema.safeParse(unparsedQuery);
+
+  let parsedQuery: InferedSchema = useMemo(() => {
+    return {} as InferedSchema;
+  }, []);
```
sean-brydon, line 39: Preventing infinite renders

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  // Remove item by key and value
+  function removeItemByKeyAndValue<J extends SchemaKeys>(
+    key: J,
+    value: InferedSchema[J] extends Array<unknown> | undefined
+      ? NonNullable<InferedSchema[J]>[number]
+      : NonNullable<InferedSchema[J]>
+  ) {
+    const existingValue = parsedQuery[key];
+    console.log(existingValue);
```
sean-brydon, line 86:
```suggestion

```

zomars (approved):
Merry Xmas @sean-brydon 🙏🏽

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -20,52 +24,51 @@ export const queryStringArray = z
...
-  type SchemaKeys = keyof InferedSchema;
-  type OptionalKeys = {
-    [K in keyof InferedSchema]: undefined extends InferedSchema[K] ? K : never;
-  }[keyof InferedSchema];
+export const schema = z.object({
+  teamIds: queryNumberArray.optional(),
+  userIds: queryNumberArray.optional(),
+  status: z.enum(["upcoming", "recurring", "past", "cancelled", "unconfirmed"]),
+  eventTypeIds: queryNumberArray.optional(),
+});
```
zomars: Already fixed this

PR #6059, opened 2022-12-16
Fixing wizard validation

leog:
## What does this PR do?

Fixing wizard validation for https://github.com/calcom/console/issues/52

Also fixes https://github.com/calcom/console/issues/56

> **Warning**
> Please merge this before merging Console PR https://github.com/calcom/console/pull/55

## Type of change

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -15,6 +15,7 @@ const withTM = require("next-transpile-modules")([
   "@calcom/prisma",
   "@calcom/trpc",
   "@calcom/ui",
+  "@calcom/console",
 ]);
 
 const { withAxiom } = require("next-axiom");
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -3,6 +3,7 @@ const CopyWebpackPlugin = require("copy-webpack-plugin");
 const { withSentryConfig } = require("@sentry/nextjs");
 const os = require("os");
 const withTM = require("next-transpile-modules")([
+  "@calcom/console",
   "@calcom/app-store",
   "@calcom/core",
   "@calcom/dayjs",
@@ -15,7 +16,6 @@ const withTM = require("next-transpile-modules")([
   "@calcom/prisma",
   "@calcom/trpc",
   "@calcom/ui",
-  "@calcom/console",
 ]);
 
 const { withAxiom } = require("next-axiom");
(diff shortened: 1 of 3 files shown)


packages/ui/v2/core/WizardForm.tsx
```diff
@@ -1,7 +1,7 @@
 import { useRouter } from "next/router";
 
+import { useLocale } from "@calcom/console/modules/common/hooks/useLocale";
```
leog, line 3: This is to be able to use this component either from Web or Console

packages/ui/v2/core/WizardForm.tsx
```diff
@@ -42,7 +42,7 @@ function WizardForm<T extends DefaultStep>(props: {
           <p className="text-sm text-gray-500">{currentStep.description}</p>
         </div>
 
-        <div className="print:p-none px-4 py-5 sm:p-6">{currentStep.content}</div>
+        <div className="print:p-none max-w-3xl px-4 py-5 sm:p-6">{currentStep.content}</div>
```
leog, line 45: The fix for https://github.com/calcom/console/issues/56

apps/web/next.config.js
```diff
@@ -15,6 +15,7 @@ const withTM = require("next-transpile-modules")([
   "@calcom/prisma",
   "@calcom/trpc",
   "@calcom/ui",
+  "@calcom/console",
```
leog, line 18: This is to be able to import useLocale from `console` which in turn uses the one from web

apps/web/next.config.js
```diff
@@ -3,6 +3,7 @@ const CopyWebpackPlugin = require("copy-webpack-plugin");
 const { withSentryConfig } = require("@sentry/nextjs");
 const os = require("os");
 const withTM = require("next-transpile-modules")([
+  "@calcom/console",
```
zomars, line 6: Apps should never import between apps. I'm afraid I can't allow this.

packages/ui/v2/core/WizardForm.tsx
zomars: Why?

zomars (changes requested):
Sorry @leog but this is breaking a rule of not importing between apps in the monorepo. That's what packages are for. Let's either extract the common logic to a package or find another way to fix the issue.

zomars:
Also adding to this. This will break for ALL self-hosted users since they don't have access to the console repo.

leog:
You are right, not my finest hour. I will find a solution, thanks @zomars

zomars:
Closing for now

leog:
@zomars better approach here: https://github.com/calcom/cal.com/pull/6068

PR #6061, opened 2022-12-16
Fixing permission for admin wizard

### LINKED ISSUE #6060: [CAL-606] Admin Wizard visible for non-admin users
### Issue Summary

Admin Wizard visible for non-admin users.

> **Note**
> This is not an urgent matter, the admin creation on step 1 is already disabled when there are users in the DB, and the second step to enable apps does not return any app for non-admin users

### Steps to Reproduce

1. Log in as a regular user in any environment
2. Go to `/auth/setup` and see the admin wizard
3. Go to `/auth/setup?step=2` and see the app selection (not returning options luckily)

<sub>[CAL-606](https://linear.app/calcom/issue/CAL-606/admin-wizard-visible-for-non-admin-users)</sub>

leog:
## What does this PR do?

Adding a redirection to 400 error when logged in user's role is not ADMIN.

Fixes #6060 

## Type of change

## How should this be tested?

1. Log in as a regular user and go to `/auth/setup` and you will get redirected to the 400 error


apps/web/pages/auth/setup/index.tsx
```diff
@@ -38,8 +41,22 @@ export default function Setup(props: inferSSRProps<typeof getServerSideProps>) {
   );
 }
 
-export const getServerSideProps = async () => {
+export const getServerSideProps = async (context: GetServerSidePropsContext) => {
   const userCount = await prisma.user.count();
+  const { req } = context;
+  const session = await getSession({ req });
+
+  console.log({ role: session?.user.role });
```
hariombalhara, line 49: nit: remove console.log

apps/web/pages/auth/setup/index.tsx
```diff
@@ -38,8 +41,22 @@ export default function Setup(props: inferSSRProps<typeof getServerSideProps>) {
...
+
+  if (session?.user.role && session?.user.role !== UserPermissionRole.ADMIN) {
+    return {
+      redirect: {
+        destination: `/400`,
```
hariombalhara, line 54: 403 or 404 seems better option to me.
403 - Clearly tells that you are not authorized
404 - We want to hide that the user is accessing a valid route

hariombalhara (approved):

apps/web/pages/auth/setup/index.tsx
leog, line 49:
```suggestion
```

apps/web/pages/auth/setup/index.tsx
leog: @hariombalhara I think we only have error 400 and 500 at the moment

zomars: @leog we don't have a 400 error page tho

leog: My bad! 😅

PeerRich (approved):
LGTM

JeroenReumkens (approved):

PR #6068, opened 2022-12-16
Fixing wizard validation

leog:
## What does this PR do?

Fixing wizard validation for https://github.com/calcom/console/issues/52 and https://github.com/calcom/console/pull/57

Also fixes https://github.com/calcom/console/issues/56

## Type of change


packages/ui/v2/core/Stepper.tsx
```diff
@@ -11,8 +12,12 @@ function Stepper<T extends DefaultStep>(props: {
   step: number;
   steps: T[];
   disableSteps?: boolean;
+  t?: TFunction;
 }) {
-  const { t } = useLocale();
+  let { t } = useLocale();
+  if (props.t) {
+    t = props.t;
+  }
```
leog, line 20: Making the translation configurable from outside in order to be able to use this component from different packages resolving translations differently.

packages/ui/v2/core/WizardForm.tsx
```diff
@@ -18,9 +19,13 @@ function WizardForm<T extends DefaultStep>(props: {
...
   disableNavigation?: boolean;
   containerClassname?: string;
+  t?: TFunction;
 }) {
   const { href, steps } = props;
-  const { t } = useLocale();
+  let { t } = useLocale();
+  if (props.t) {
+    t = props.t;
+  }
```
leog, line 28: Making the translation configurable from outside in order to be able to use this component from different packages resolving translations differently.

packages/ui/v2/core/WizardForm.tsx
```diff
@@ -42,7 +47,7 @@ function WizardForm<T extends DefaultStep>(props: {
           <p className="text-sm text-gray-500">{currentStep.description}</p>
         </div>
 
-        <div className="print:p-none px-4 py-5 sm:p-6">{currentStep.content}</div>
+        <div className="print:p-none max-w-3xl px-4 py-5 sm:p-6">{currentStep.content}</div>
```
leog, line 50: The fix for https://github.com/calcom/console/issues/56

packages/ui/v2/core/WizardForm.tsx
```diff
@@ -63,10 +68,7 @@ function WizardForm<T extends DefaultStep>(props: {
                   type="submit"
                   color="primary"
                   form={`wizard-step-${step}`}
-                  className="relative ml-2"
-                  onClick={() => {
-                    setStep(step + 1);
-                  }}>
```
leog, line 69: This is the fundamental part of this PR, fixes validation by not switching to the next step on click, not letting the UI show errors to let the use correct them.

packages/ui/v2/core/WizardForm.tsx
```diff
@@ -76,7 +78,7 @@ function WizardForm<T extends DefaultStep>(props: {
       </div>
       {!props.disableNavigation && (
         <div className="print:hidden">
-          <Stepper href={href} step={step} steps={steps} disableSteps />
+          <Stepper href={href} step={step} steps={steps} disableSteps t={t} />
```
leog, line 81: Making `Stepper` component configurable in terms of translation.

packages/ui/v2/core/WizardForm.tsx
PeerRich: interesting, havent seen this before. is this a good pattern @zomars @emrysal ?

PeerRich (approved):

packages/ui/v2/core/WizardForm.tsx
leog: I would say it’s definitely not ideal, as co sole should have official i18n support, so any cross-package ui component that needs translation can be natively translated instead of artificially like this, but as of now, unless we are ready to introduce i18n in Console, it’s just a monkey patch

JeroenReumkens (approved):

packages/ui/v2/core/WizardForm.tsx
zomars: I have my doubts. Also I think WizardForm should not handle translations at all. It should receive everything pre translated.

PR #6081, opened 2022-12-17
Update constants.ts

PeerRich:


emrysal:
For some reason even though we don't seem to be using it, our tests do fail without - I brought it back for now.

PR #6098, opened 2022-12-19
Ensure topbanner is always on top of page layout rather than inside main content

JeroenReumkens:
## What does this PR do?

Ensures top banner is always on top of the page, instead of nested inside the main content (next to the nav).

The way it works now:
* The top banner is always on top of the page, if the user scrolls down it scrolls up as well, so we have as much space on the page as possible. Especially important on mobile. 
* If the left nav has more content than the height of the page, the left menu will get a scrollbar + a transparent gradient behind the profile link, showing that the contents of the nav is scrollable (see Loom).
* As for the navigation on the settings pages I removed the duplication — we had 2 navs, one for mobile and one for desktop. Now they share the same div. Please test that this works properly a well. 

Also included the following tweaks for the top banner:
* Made CTA text font medium as per design
* Added info icon for warnings and exclamation mark for errors, instead of always exclamation mark.

Loom Video: https://www.loom.com/share/ad1e322a78a84ad39b388a1144a465b0

## Type of change

## How should this be tested?

- [ ] Test pages that both have content longer than your current page as well as content lower than your current page. Both on desktop and mobile. Ensure that navigation looks correctly and is always accessible. Also ensure that the top banner isn't overlapping the navigation or content.
- [ ] Since I merged the mobile + desktop nav on the settings pages into one, we also need to test this with extra care to make sure the nav still works on all breakpoints on the settings pages.


packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
```diff
@@ -16,9 +16,12 @@ export function TeamsUpgradeBanner() {
       showToast(error.message, "error");
     },
   });
-  if (!data) return null;
```
JeroenReumkens, line 19: TODO: Need to revert this after it's tested. 

- [ ] TODO

packages/ui/components/top-banner/TopBanner.tsx
```diff
@@ -23,17 +25,20 @@ export function TopBanner(props: TopBannerProps) {
     <div
       data-testid="banner"
       className={classNames(
-        "flex min-h-[40px] w-full items-start justify-between gap-8 bg-gray-50 py-2 px-4 text-center lg:items-center",
+        "alert-banner flex min-h-[40px] w-full items-start justify-between gap-8 bg-gray-50 py-2 px-4 text-center lg:items-center [&+.alert-banner]:border-t [&+.alert-banner]:border-black",
```
JeroenReumkens, line 28: Added a divider line between to banners in the edge case where 2 banners are visible at the same time.

packages/ui/components/top-banner/TopBanner.tsx
```diff
@@ -23,17 +25,20 @@ export function TopBanner(props: TopBannerProps) {
...
         variantClassName[variant]
       )}>
       <div className="flex flex-1 flex-col items-start justify-center gap-2 p-1 lg:flex-row lg:items-center">
         <p className="flex flex-col items-start justify-center gap-2 text-left font-sans text-sm font-medium leading-4 text-gray-900 lg:flex-row lg:items-center">
-          {["warning", "error"].includes(variant) && (
-            <ExclamationIcon className="h-5 w-5 text-black" aria-hidden="true" />
+          {variant === "error" && (
```
JeroenReumkens, line 33: Show correct icons depending on banner type.

packages/ui/v2/core/Shell.tsx
```diff
@@ -153,12 +153,12 @@ const Layout = (props: LayoutProps) => {
 
       {/* todo: only run this if timezone is different */}
       <TimezoneChangeDialog />
-      <div className="h-screen overflow-hidden">
-        <div className="flex h-screen overflow-hidden" data-testid="dashboard-shell">
+      <div className="flex min-h-screen flex-col">
```
JeroenReumkens, line 156: This is actually the main change, moving the banners outside the page container and on top of everything. This did however cascade into a lot of other changes that needed to be made in our Shell to even be able to acommodate this  😅 — those are the changes following next.

packages/ui/v2/core/Shell.tsx
```diff
@@ -742,59 +738,59 @@ function SideBarContainer() {
 
 function SideBar() {
   return (
-    <aside className="desktop-transparent hidden w-14 flex-col border-r border-gray-100 bg-gray-50 md:flex lg:w-56 lg:flex-shrink-0 lg:px-4">
-      <div className="flex h-0 flex-1 flex-col overflow-y-auto pt-3 pb-4 lg:pt-5">
-        <header className="items-center justify-between md:hidden lg:flex">
+    <div className="relative">
```
JeroenReumkens, line 741: Main change here is that everything is wrapped in this relative div. Besides that you see additional changes on the new lines 771 and 772 highlighted in darker green/red.

packages/ui/v2/core/Shell.tsx
```diff
@@ -742,59 +738,59 @@ function SideBarContainer() {
...
-        <span className="hidden lg:inline">
-          <UserDropdown />
-        </span>
-        <span className="hidden md:inline lg:hidden">
-          <UserDropdown small />
-        </span>
-      </div>
-      <DeploymentInfo />
-    </aside>
+        <div className="fixed left-1 bottom-0 w-4 bg-gray-50 before:absolute before:left-0 before:-top-20 before:h-20 before:w-48 before:bg-gradient-to-t before:from-gray-50 before:to-transparent lg:left-4 lg:w-48">
```
JeroenReumkens, line 781: Again, little change, besides the user dropdown + deployment info wrapped inside this fixed positioned div. The content of these divs hasn't changed.

packages/ui/v2/core/Shell.tsx
```diff
@@ -855,35 +851,17 @@ export function ShellMain(props: LayoutProps) {
...
   ...props
 }: LayoutProps) {
-  const [sideContainerOpen, setSideContainerOpen] = props.drawerState || [false, noop];
-
   return (
-    <main className="relative z-0 flex flex-1 flex-col overflow-y-auto bg-white focus:outline-none">
+    <main className="relative z-0 flex-1 bg-white focus:outline-none">
       {/* show top navigation for md and smaller (tablet and phones) */}
       {TopNavContainerProp}
-      {/* The following is used for settings navigation on medium and smaller screens */}
```
JeroenReumkens, line 872: I've merged the mobile and desktop setting navs into one. However, even before that I couldn't find a use case for this empty div at all. Had the feeling this already was unused. @joeauyeung  Anything specific I overlooked? Although there isn't even a separate placeholder for the mobile nav anymore, but just want to make sure I didn't overlook anything.

packages/ui/v2/core/Shell.tsx
```diff
@@ -855,35 +851,17 @@ export function ShellMain(props: LayoutProps) {
...
 function MainContainer({
-  SettingsSidebarContainer: SettingsSidebarContainerProp = <SettingsSidebarContainerDefault />,
   MobileNavigationContainer: MobileNavigationContainerProp = <MobileNavigationContainer />,
   TopNavContainer: TopNavContainerProp = <TopNavContainer />,
   ...props
 }: LayoutProps) {
-  const [sideContainerOpen, setSideContainerOpen] = props.drawerState || [false, noop];
-
   return (
-    <main className="relative z-0 flex flex-1 flex-col overflow-y-auto bg-white focus:outline-none">
```
JeroenReumkens, line 869: removed overflow — the full page should be scrollable, not a portion of the page, can result in very weird behaviour and eg breaking scroll restoration when going back a page.

packages/ui/v2/core/Shell.tsx
```diff
@@ -855,35 +851,17 @@ export function ShellMain(props: LayoutProps) {
   );
 }
 
-const SettingsSidebarContainerDefault = () => null;
-
 function MainContainer({
-  SettingsSidebarContainer: SettingsSidebarContainerProp = <SettingsSidebarContainerDefault />,
```
JeroenReumkens, line 861: No custom settingssidebar anymore, it is injected via the default SidebarContainer prop.

packages/ui/v2/core/Shell.tsx
```diff
@@ -855,35 +851,17 @@ export function ShellMain(props: LayoutProps) {
...
-        )}
-        onClick={() => {
-          setSideContainerOpen(false);
-        }}
-      />
-      {SettingsSidebarContainerProp}
       <div className="max-w-full px-4 py-2 lg:py-8 lg:px-12">
         <ErrorBoundary>
-          {/* add padding to top for mobile when App Bar is fixed */}
-          <div className="pt-14 sm:hidden" />
```
JeroenReumkens, line 886: Seems redundant with our new positioning. Plus I love getting rid of these empty divs that only add padding — we shouldn't use those 😁

packages/ui/v2/core/Shell.tsx
```diff
@@ -906,7 +884,7 @@ function TopNav() {
     <>
       <nav
         style={isEmbed ? { display: "none" } : {}}
-        className="fixed z-40 flex w-full items-center justify-between border-b border-gray-200 bg-gray-50 bg-opacity-50 py-1.5 px-4 backdrop-blur-lg sm:relative sm:p-4 md:hidden">
+        className="sticky top-0 z-40 flex w-full items-center justify-between border-b border-gray-200 bg-gray-50 bg-opacity-50 py-1.5 px-4 backdrop-blur-lg sm:p-4 md:hidden">
```
JeroenReumkens, line 887: No fixed positioning anymore, sticky takes way better care of the job — sticking the nav to the top when the user scrolls down, but not sticking it when there's a topbar above the navigation,  until that is scrolled out of view.

packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -127,7 +135,13 @@ const SettingsSidebarContainer = ({ className = "" }) => {
 
   return (
     <nav
-      className={`no-scrollbar flex w-56 flex-col space-y-1 overflow-x-hidden overflow-y-scroll py-3 px-2 ${className}`}
+      className={classNames(
```
JeroenReumkens, line 138: Instead of having 2 instances of the nav (for mobile and desktop), we now use this boolean to add a few conditional styles to toggle the div between its desktop and mobile state 🔥

packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -295,9 +309,12 @@ const MobileSettingsContainer = (props: { onSideContainerOpen?: () => void }) =>
 
   return (
     <>
-      <nav className="fixed z-20 flex w-full items-center justify-between border-b border-gray-100 bg-gray-50 p-4 sm:relative lg:hidden">
+      <nav className="sticky top-0 z-20 flex w-full items-center justify-between border-b border-gray-100 bg-gray-50 p-4 sm:relative lg:hidden">
```
JeroenReumkens, line 312: Again position sticky ftw.

packages/ui/components/top-banner/TopBanner.tsx
```diff
@@ -23,17 +25,20 @@ export function TopBanner(props: TopBannerProps) {
...
+            <Icon.FiAlertTriangle className="h-4 w-4 stroke-[2.5px] text-black" aria-hidden="true" />
+          )}
+          {variant === "warning" && (
+            <Icon.FiInfo className="h-4 w-4 stroke-[2.5px] text-black" aria-hidden="true" />
           )}
           {text}
         </p>
-        {actions && <div className="text-sm">{actions}</div>}
+        {actions && <div className="text-sm font-medium">{actions}</div>}
```
JeroenReumkens, line 41: Correct font weight as per design 🥰

packages/ui/components/top-banner/TopBanner.tsx
sean-brydon: but but `divide-y`

JeroenReumkens: HAHAHA. This guy. You're totally right. It's my way too old brain trying to use old skool CSS where the new kid on the block (tailwind) does it way better. I'll change this indeed, great suggestion 😁

JeroenReumkens: Fixed!

Udit-takkar:
@JeroenReumkens  just tried running it. it works great.

packages/features/ee/teams/components/TeamsUpgradeBanner.tsx
JeroenReumkens: Removed.

packages/ui/v2/core/layouts/SettingsLayout.tsx
```diff
@@ -340,18 +358,21 @@ export default function SettingsLayout({
...
       {...rest}
-      SidebarContainer={<SettingsSidebarContainer className="hidden lg:flex" />}
+      SidebarContainer={
+        <>
+          {/* Mobile backdrop */}
+          {sideContainerOpen && (
+            <button
+              onClick={() => setSideContainerOpen(false)}
+              className="fixed top-0 left-0 z-10 h-full w-full bg-black/50">
+              <span className="sr-only">{t("hide_navigation")}</span>
```
PeerRich, line 368: do we wanna translate sr-only strings, too?

PeerRich (approved):

packages/ui/v2/core/layouts/SettingsLayout.tsx
zomars: Yes

PR #6099, opened 2022-12-19
Feature/booking filters

sean-brydon:
## What does this PR do?

- [x] #6034 needs to be merged first ideally

Booking filters working for teams and users

 Loom Video: https://www.loom.com/share/1924b26e62e540ca81a870b2ce80b684

Mobile mode has been fixed: 
* Detect which side of the screen the popover is opening on 
* Open on the correct start/end alignment based on which 1/2 the trigger is present on.

Left side -> Opens on the correct side to not overflow
CleanShot 2022-12-22 at 09 58 58@2x

Right side
CleanShot 2022-12-22 at 09 59 33@2x

## Type of change

## How should this be tested?

- [ ] Add multiple bookings to teams / users
- [ ] Filter them using the popovers 

TODOS:
- [x] Mobile View
- [ ] Add filter button 
- [ ] Support more filters
- [ ] Move from checkboxes to a custom "Select style" component


packages/features/bookings/components/EventTypeFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  {
+    team: {
+      id: number;
+      name: string;
+    } | null;
+    id: number;
+    title: string;
+    slug: string;
+  }[]
+>;
```
sean-brydon, line 24: This filter is not used just yet - need to figure out the add filter button

packages/features/bookings/groupBy.ts
```diff
@@ -0,0 +1,18 @@
...
+        acc[key].push(item);
+      } else {
+        // did not find key, create new array
+        acc[key] = [item];
+      }
+      return acc;
+    },
+    {} // start with empty object
+  );
+}
```
sean-brydon, line 18: This groups a array of objects by a single key. 

And will be used in the eventtype filter to show all event types per team

packages/features/bookings/layout/BookingLayout.tsx
```diff
@@ -40,12 +34,10 @@ export default function BookingLayout({
 }: { children: React.ReactNode } & ComponentProps<typeof Shell>) {
   return (
     <Shell {...rest}>
-      <div className="flex flex-col sm:space-x-2 xl:flex-row">
-        <div className="hidden xl:block">
-          <VerticalTabs tabs={tabs} sticky />
-        </div>
-        <div className="block xl:hidden">
```
sean-brydon, line 47: Always horizontal tabs for now

packages/features/bookings/lib/useFilterQuery.tsx
```diff
@@ -0,0 +1,15 @@
...
+export const filterQuerySchema = z.object({
+  teamIds: queryNumberArray.optional(),
+  userIds: queryNumberArray.optional(),
+  status: z.enum(["upcoming", "recurring", "past", "cancelled", "unconfirmed"]),
+  eventTypeIds: queryNumberArray.optional(),
+});
+
+export function useFilterQuery() {
+  return useTypedQuery(filterQuerySchema);
+}
```
sean-brydon, line 15: Saves typing this every time

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+// Take array as a string and return zod  number array
+
+// Take string and return return zod string array - comma separated
+export const queryStringArray = z
+  .preprocess((a) => z.string().parse(a).split(","), z.string().array())
+  .or(z.string().array());
+
+export function useTypedQuery<T extends z.Schema>(schema: T) {
```
sean-brydon, line 23: This file comes from https://github.com/calcom/cal.com/pull/6034 it was just not merged before starting this.

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -93,7 +93,12 @@ export const bookingsRouter = router({
   get: authedProcedure
     .input(
       z.object({
-        status: z.enum(["upcoming", "recurring", "past", "cancelled", "unconfirmed"]),
+        filters: z.object({
+          teamIds: z.number().array().optional(),
+          userIds: z.number().array().optional(),
+          status: z.enum(["upcoming", "recurring", "past", "cancelled", "unconfirmed"]),
+          eventTypeIds: z.number().array().optional(),
```
sean-brydon, line 100: Need a new filter here as they are sent as arrays - the one from typequery isnt a true filter for the result and can create errors in the backend

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -104,7 +109,7 @@ export const bookingsRouter = router({
       const take = input.limit ?? 10;
       const skip = input.cursor ?? 0;
       const { prisma, user } = ctx;
-      const bookingListingByStatus = input.status;
+      const bookingListingByStatus = input.filters.status;
```
sean-brydon, line 112: Moved status to filters also

packages/trpc/server/trpc.ts
```diff
@@ -1,4 +1,5 @@
 import superjson from "superjson";
+import { ZodError } from "zod";
```
sean-brydon, line 2:
```suggestion
```

packages/ui/components/avatar/Avatar.tsx
```diff
@@ -9,48 +9,56 @@ import { Maybe } from "@trpc/server";
...
+      <>
+        <AvatarPrimitive.Image src={imageSrc ?? undefined} alt={alt} className={rootClass} />
+        <AvatarPrimitive.Fallback delayMs={600} asChild={props.asChild}>
+          <>
+            {props.fallback && !gravatarFallbackMd5 && props.fallback}
+            {gravatarFallbackMd5 && (
+              <img src={defaultAvatarSrc({ md5: gravatarFallbackMd5 })} alt={alt} className={rootClass} />
+            )}
+          </>
+        </AvatarPrimitive.Fallback>
```
sean-brydon, line 49: Adding an as child prop to fallback - as for some stupid reason fallback renders this like 

```
<span>
  <span>
    <img/>
  </span>
</span>
```

packages/ui/components/popover/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <div className="max-w-36 flex">
+            <div className="truncate ">
+              {text}
+              {count && count > 0 && (
+                <div className="flex h-4 w-4 items-center justify-center rounded-full">{count}</div>
+              )}
+            </div>
+            <Icon.FiChevronDown
+              className={classNames("mt-auto ml-2 transition-transform duration-150", open && "rotate-180")}
+            />
```
sean-brydon, line 33: Fancy animations :)

packages/ui/components/popover/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+          </div>
+        </div>
+      </Popover.Trigger>
+      <Popover.Content side="bottom" align="end" asChild>
+        <div className="absolute z-50 mt-2 w-56 -translate-x-[228px] rounded-md bg-white  shadow-sm ring-1 ring-black ring-opacity-5 focus-within:outline-none">
```
sean-brydon, line 38: `-translate-x-[228px]` translate the size of the filter so its always the correct alligment

PeerRich:
awesome feature, somehow a ton of merge conflicts.

PeerRich:
how does this look on tablet and mobile?

PeerRich:
these dropdowns have no purpose if you are not in a team

CleanShot 2022-12-19 at 12 56 02
CleanShot 2022-12-19 at 12 56 10

sean-brydon:
> these dropdowns have no purpose if you are not in a team
> 
> CleanShot 2022-12-19 at 12 56 02 CleanShot 2022-12-19 at 12 56 10

Correct - i thought i set them to hidden - let me check that out will push a fix

PeerRich:
was changing the layout on purpose?

before

CleanShot 2022-12-19 at 13 02 34


after

CleanShot 2022-12-19 at 13 02 54

packages/trpc/server/routers/viewer/teams.tsx
```diff
@@ -642,4 +642,45 @@ export const viewerTeamsRouter = router({
...
+            },
+          },
+        },
+      });
+      type UserMap = Record<number, typeof teams[number]["members"][number]["user"]>;
+      // flattern users to be unique by id
+      const users = teams
+        .flatMap((t) => t.members)
+        .reduce((acc, m) => (m.user.id in acc ? acc : { ...acc, [m.user.id]: m.user }), {} as UserMap);
+      return Object.values(users);
```
zomars, line 684: Made some perf improvements

packages/features/bookings/components/TeamFilter.tsx
```diff
@@ -0,0 +1,75 @@
+import { useLocale } from "@calcom/lib/hooks/useLocale";
```
JeroenReumkens, line 1: This file is 99% similar to the previous filter. Do you agree that it makes sense to abstract this into a reusable component? Makes adding more filters super easy 🔥 
Let me know if there's any difference between them I overlooked!

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -194,7 +238,7 @@ export const bookingsRouter = router({
               },
             },
           ],
-          AND: [passedBookingsFilter],
+          AND: [passedBookingsStatusFilter, ...filtersCombined],
```
JeroenReumkens, line 241: You're potentially spreading undefined here, right? Not sure if that would work or you need to default to an empty array on line 206.

packages/ui/components/form/select/components.tsx
```diff
@@ -158,3 +158,28 @@ export const MenuListComponent = <
     )}
   />
 );
+
+export const OptionComponentWithIcon = <
```
JeroenReumkens, line 162: Why didn't you use the current optioncomponent and added an icon to that?

JeroenReumkens (changes requested):

packages/features/bookings/components/TeamFilter.tsx
sean-brydon: #5590  created an issue along with other things that need to be picked up in the future

PeerRich:
#6034 was merged

JeroenReumkens (dismissed):
Great work @sean-brydon , also on the fixes 🙏 
I just checked it locally as well, and still found a few things we need to tackle: 

1. The filters on mobile current don't have an overflow scroll. So if we add another one, we need to build something for that. Not something we have to fix now, but good to keep in mind.
2. The filters should detect boundaries on mobile, now they are out of view: 
CleanShot 2022-12-20 at 18 28 47@2x
3. The users filter now acts as an OR filter. I'm unsure whether we would be expecting it to work like "AND"? The latter was my expectation, and in that case my screenshot should return 0 results. What do @PeerRich  and @Jaibles  think? 
CleanShot 2022-12-20 at 18 29 22@2x


Apart from that amazing work and almost ready to merge 🔥

sean-brydon:
> 3. The users filter now acts as an OR filter. I'm unsure whether we would be expecting it to work like "AND"? The latter was my expectation, and in that case my screenshot should return 0 results. What do @PeerRich  and @Jaibles  think?
>    CleanShot 2022-12-20 at 18 29 22@2x
> 
> Apart from that amazing work and almost ready to merge 🔥

Thank you @JeroenReumkens i'll act on the feedback tomorrow.

However, i really disagree with this:
If im on amazon each filter acts as a OR 
CleanShot 2022-12-20 at 17 57 31@2x
If i am using the teams filter - each also would have to be an OR there also. as we don't support multiple teams per events.
@Jaibles

PeerRich:
CleanShot 2022-12-21 at 00 16 08

reminder to not commit those

Jaibles:
> The users filter now acts as an OR filter. I'm unsure whether we would be expecting it to work like "AND"? The latter was my expectation, and in that case my screenshot should return 0 results. What do @PeerRich and @Jaibles think?

@JeroenReumkens based on checking out @sean-brydon's loom this works exactly as expected. If you're checking all, you want "all". If you check any other individual or team, you want to filter by an individual therefore taking you out of the "all" filter. 

When you say the "AND" was your expectation, were you expecting clicking "All" and another person / team would be an "AND" in that scenario? While I accept that this might be slightly different behaviour than what you'd expect if all things were equal in the list from a UI perspective, I think the "All" labelling of it communicates that it's not. If we want to make this clearer, we could put a divider between the "all" filter and the others, do you think that would help?

JeroenReumkens:
@Jaibles  The "AND" expectation mainly had to do with selecting 2 persons from the team (second filter). My feeling was that it then would filter for meetings with person 1 AND person 2. Right now it shows you meetings with person 1 OR person 2 (and potentially others, as long as 1 OR 2 (or both) is included). 

I think it fully depends on how users would use these filters. Are they searching for meetings where they know these 2 people attended, or are they looking for meetings where either of these people attended, but not necessarily at the same meeting? 🤔

Jaibles:
@JeroenReumkens I get you. More about the results we're returning and not about the "All" in the filter. That does make sense. I've checked the behaviour of quite a few filters now and seems some behave like what we have, some like what you're talking about. If I was using a seperate filter, I would expect it to be an AND, but within the same set of filters I expect and OR. See [here](https://share.cleanshot.com/n4dvGSmx) and [here](https://share.cleanshot.com/h8Fr4L0D) for examples

sean-brydon:
@JeroenReumkens do you have any idea how we can detect the boundaries ?  

Over added the overflow scroll however, since we are using translate to move the popover content into position  - its super hard to control which side the content opens on from CSS

emrysal:
Spotted a strange (minor, non blocking bug)

1) Click on team dropdown, select a team.
2) Close dropdown
3) Open dropdown
4) Select "All Bookings"

image

image

emrysal:
Possibly not relevant to this PR, but not a fan of the double listing:

image

emrysal:
@JeroenReumkens I think the OR behaviour is acceptable, doesn't feel weird to me; but we can of course revisit at a later time. For now it's good to go if your other feedback has been addressed - @sean-brydon

emrysal (approved):

emrysal (changes requested):
Blocking until mobile is fixed.

emrysal (approved):
LGTM!

PR #6102, opened 2022-12-19
Removes EMAIL_ADDRESS action

CarinaWolli:
## What does this PR do?

Removes 'email to a specific email address' workflow action. 
Was originally implemented in #6075 but got reverted by #6085

### DIFF
--- a/packages/features/ee/workflows/lib/constants.ts
+++ b/packages/features/ee/workflows/lib/constants.ts
@@ -11,7 +11,7 @@ export const WORKFLOW_TRIGGER_EVENTS = [
 export const WORKFLOW_ACTIONS = [
   WorkflowActions.EMAIL_HOST,
   WorkflowActions.EMAIL_ATTENDEE,
-  //WorkflowActions.EMAIL_ADDRESS, Disabling for now due to abuse episode
+  WorkflowActions.EMAIL_ADDRESS,
   WorkflowActions.SMS_ATTENDEE,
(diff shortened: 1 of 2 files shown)


packages/features/ee/workflows/lib/constants.ts
```diff
@@ -11,7 +11,7 @@ export const WORKFLOW_TRIGGER_EVENTS = [
 export const WORKFLOW_ACTIONS = [
   WorkflowActions.EMAIL_HOST,
   WorkflowActions.EMAIL_ATTENDEE,
-  WorkflowActions.EMAIL_ADDRESS,
+  //WorkflowActions.EMAIL_ADDRESS, Disabling for now due to abuse episode
```
PeerRich, line 14:
```suggestion
```

PeerRich: not sure if // works. maybe just delete it?

JeroenReumkens: It does work. But I agree that removing might make more sense. Adding it back later is easy.

JeroenReumkens (approved):

PeerRich:
@CarinaWolli tests are failing because they rely on EMAIL_ADDRESS action. we'd need to remove that test, too

CarinaWolli:
@PeerRich @JeroenReumkens I am filtering out the action now, I think that's an easier approach as otherwise, we have type errors

PR #6114, opened 2022-12-19
Fixing multiple duration default issues

### LINKED ISSUE #6100: [CAL-619] multiple durations: if no default is set, days won't load
report: https://twitter.com/rich_cahill/status/1604792559112310785

if no default is set for multiple durations, it fails to load days:
image

<sub>[CAL-619](https://linear.app/calcom/issue/CAL-619/multiple-durations-if-no-default-is-set-days-wont-load)</sub>

leog:
## What does this PR do?

Default duration on multiple duration event-type was not well set nor checked upon saving, it's now fixed.

> **Warning**
> Candidate for a hotfix

Fixes #6100 

## Type of change


PeerRich:
when you select "allow booker to choose" for the first time, can we take 30 min as a default?

leog:
> when you select "allow booker to choose" for the first time, can we take 30 min as a default? that way it doesnt start empty

@PeerRich so we should always add 30 mins as an option which will set the default too?

apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -262,12 +262,14 @@ export const EventSetupTab = (
                   if (!newOptions.find((opt) => opt.value === defaultDuration?.value)) {
                     if (newOptions.length > 0) {
                       setDefaultDuration(newOptions[0]);
+                      formMethods.setValue("length", newOptions[0].value);
```
leog, line 265: When setting default duration, UI was showing a value but form was not impacted. Fixed.

apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -262,12 +262,14 @@ export const EventSetupTab = (
...
                     } else {
                       setDefaultDuration(null);
                     }
                   }
                   if (newOptions.length === 1 && defaultDuration === null) {
                     setDefaultDuration(newOptions[0]);
+                    formMethods.setValue("length", newOptions[0].value);
```
leog, line 272: When setting default duration, UI was showing a value but form was not impacted. Fixed.

emrysal (approved):
Maybe candidate for zod, looks good to me.

zomars (approved):
Approved due to urgency but I rather show an error if the user starts tinkering with the duration query. If duration=0 is set let's just show a 500 error. It's perfectly valid.

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -283,7 +283,7 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
             if (metadata?.multipleDuration.length < 1) {
               throw new Error(t("event_setup_multiple_duration_error"));
             } else {
-              if (input.length && !metadata?.multipleDuration?.includes(input.length)) {
+              if (!input.length && !metadata?.multipleDuration?.includes(input.length)) {
```
Comment (leog) [inline apps/web/pages/event-types/[type]/index.tsx:286]: We want to show the error if there is no event type duration and if the duration is not included as a value. This failed when duration was undefined. Fixed.

leog:
> Approved due to urgency but I rather show an error if the user starts tinkering with the duration query. If duration=0 is set let's just show a 500 error. It's perfectly valid.

@zomars done here: https://github.com/calcom/cal.com/pull/6146

emrysal:
> Approved due to urgency but I rather show an error if the user starts tinkering with the duration query. If duration=0 is set let's just show a 500 error. It's perfectly valid.

I disagree with this take, although valid (btw 500 isn't valid, but 400 would be). I think invalid values should just be ignored and not 400 at all. E.g. with GH - nothing happens:

image

PR #6117, opened 2022-12-20
fix/booking-page-query-search

alannnc:
## What does this PR do?

Change startTime to current date - 2 days if startTime it's on current month, so we don't load full calendar month if its not required.

## Type of change

## How should this be tested?

- [ ] Change months and the only time startTime should be different than 1 should be on current month

## Checklist


emrysal:
I understand this is a performance improvement, have you done any benchmarking on this?

alannnc:
> I understand this is a performance improvement, have you done any benchmarking on this?

It's a variable improvement since the later we go on the month the more milliseconds we save also with this if a user has heavy load of busy events we should be saving a few more ms. Also calendar fetching for any credentials should follow the same dateFrom rule to don't return past days so a few ms more.
After a couple of test run got this numbers:

- With fix AVG 17.1339782ms on checkForAvailability
- Without fix AVG 34.20797671ms

Probably nothing but hey we better save what we can.
@emrysal

emrysal (approved):

PR #6119, opened 2022-12-20
Fixing booking limits

sean-brydon:
## What does this PR do?
Well - this fixes a bug that should have never been a bug :') 

### DIFF
--- a/packages/core/getUserAvailability.ts
+++ b/packages/core/getUserAvailability.ts
@@ -207,18 +207,18 @@ export async function getUserAvailability(
           if (
             // Only check OUR booking that matches the current eventTypeId
             // we don't care about another event type in this case as we dont need to know their booking limits
-            bookingEventTypeId == eventType?.id &&
-            dayjs(booking.start).isBetween(startDate, endDate)
+            !(bookingEventTypeId == eventType?.id && dayjs(booking.start).isBetween(startDate, endDate))
           ) {
-            // increment total and check against the limit, adding a busy time if condition is met.
-            total++;
-            if (total >= limit) {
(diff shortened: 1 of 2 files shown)


packages/core/getUserAvailability.ts
```diff
@@ -196,21 +196,32 @@ export async function getUserAvailability(
...
-        // we don't care about another event type in this case as we dont need to know their booking limits
-        if (total >= limit && bookingEventTypeId === eventType?.id?.toString()) {
-          bufferedBusyTimes.push({ start: startDate.toISOString(), end: endDate.toISOString() });
+        const endDate = dayjs(date).endOf(filter);
+        for (const booking of ourBookings) {
+          const bookingEventTypeId = booking.source?.split("-")[1];
+          if (
+            // Only check OUR booking that matches the current eventTypeId
+            // we don't care about another event type in this case as we dont need to know their booking limits
+            !(bookingEventTypeId == eventType?.id && dayjs(booking.start).isBetween(startDate, endDate))
```
emrysal, line 210: Checking for `bookingEventTypeId` equality first, before checking against the more expensive isBetween operation.

packages/core/getUserAvailability.ts
```diff
@@ -196,21 +196,32 @@ export async function getUserAvailability(
...
+          ) {
+            continue;
+          }
+          // increment total and check against the limit, adding a busy time if condition is met.
+          total++;
+          if (total >= limit) {
```
emrysal, line 216: Limit check to when total can exceed the limit

emrysal (approved):
Some performance improvements + brought back some comments.

PR #6139, opened 2022-12-21
Fix/Embed - Fix for Duration query param getting removed

### LINKED ISSUE #6127: [CAL-633] /embed - Call duration is not set to the default
When I go to book a call in the embed, it does not preselect the default duration

I then end up with "undefined minutes"

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-633](https://linear.app/calcom/issue/CAL-633/embed-call-duration-is-not-set-to-the-default)</sub>

hariombalhara:
## What does this PR do?

Fixes #6127

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

### DIFF
--- a/apps/web/components/booking/BookingDescription.tsx
+++ b/apps/web/components/booking/BookingDescription.tsx
@@ -55,7 +55,7 @@ const BookingDescription: FC<Props> = (props) => {
         }
       }
     }
-  }, []);
+  }, [setDuration, eventType.length, eventType.metadata?.multipleDuration, duration]);
   let requiresConfirmation = eventType?.requiresConfirmation;
   let requiresConfirmationText = t("requires_confirmation");
   const rcThreshold = eventType?.metadata?.requiresConfirmationThreshold;


apps/web/components/booking/BookingDescription.tsx
```diff
@@ -55,7 +55,7 @@ const BookingDescription: FC<Props> = (props) => {
         }
       }
     }
-  }, []);
+  }, [setDuration, eventType.length, eventType.metadata?.multipleDuration, duration]);
```
hariombalhara, line 58: This works till I figure out why duration param is being removed.

hariombalhara: Using shallow:false on `useRouterQuery` also fixes the issue.

PeerRich (approved):

apps/web/components/booking/BookingDescription.tsx
PeerRich: so can we merge this?

hariombalhara: Yeah, it's safe to merge and also fixes the issue.

But the root cause for why it happens only in embed isn't clear.

PR #6142, opened 2022-12-21
remove cancel button for unconfirmed past bookings

### LINKED ISSUE #3845: Past bookings are cancellable
i am able to cancel past bookings. is this a bug or expected behavior? 

https://user-images.githubusercontent.com/53316345/184550214-9dd61978-c6f0-47be-8bc7-d760e8387051.mov

### LINKED ISSUE #6132: [CAL-635] /bookings: past events should not be editable or cancellable
CleanShot 2022-12-20 at 22 42 58

<sub>[CAL-635](https://linear.app/calcom/issue/CAL-635/bookings-past-events-should-not-be-editable-or-cancellable)</sub>

mocherfaoui:
## What does this PR do?

Fixes #6132
Fixes #3845

## Type of change

## How should this be tested?

1. Book an event that requires confirmation
2. When the event is in the past and still unconfirmed you won't see the "Cancel" button

### DIFF
--- a/apps/web/components/booking/BookingListItem.tsx
+++ b/apps/web/components/booking/BookingListItem.tsx
@@ -158,6 +158,10 @@ function BookingListItem(booking: BookingItemProps) {
     bookedActions = bookedActions.filter((action) => action.id !== "edit_booking");
   }
 
+  if (isPast && isPending && !isConfirmed) {
+    bookedActions = bookedActions.filter((action) => action.id !== "cancel");
+  }
+
   const RequestSentMessage = () => {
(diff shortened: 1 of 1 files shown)


emrysal (approved):

emrysal:
@Udit-takkar It does not, but both issues were limited to talking only about the /booking endpoint, if you have to take into account the behaviour of cancellation when hitting the endpoint you actually need additional views like "This booking has already happened" which makes it a lot more complex. We should track that however in a separate issue (if that does not happen right now).

PR #6146, opened 2022-12-21
Fixing event type duration query param

### LINKED ISSUE #6145: [CAL-638] Event type duration issue
### Issue Summary

Event type duration can be modified when using `duration` query parameter in non-multiple-duration event types.

### Steps to Reproduce

1. Go to book an event type without multiple duration
2. Add a query parameter `duration=22` and see how the offered times are using 22 minutes duration

<sub>[CAL-638](https://linear.app/calcom/issue/CAL-638/event-type-duration-issue)</sub>

leog:
## What does this PR do?

Non-multiple duration event type availability page now shows error when duration query parameter exists.

Before:
<kbd></kbd>

After:
<kbd></kbd>

Fixes #6145 

## Type of change

## How should this be tested?

Try to add duration query param to a non-multiple-duration event type.


emrysal:
I suggest we take an alternate approach where we just check whether the `duration=` param is included (`metadata.multipleDuration?.includes(router.query.duration)`) and otherwise just use the default / normal duration - no 500 error whatsoever, just treat it like regular messing with the URL (don't treat it at all)

leog:
> I suggest we take an alternate approach where we just check whether the `duration=` param is included (`metadata.multipleDuration?.includes(router.query.duration)`) and otherwise just use the default / normal duration - no 500 error whatsoever, just treat it like regular messing with the URL (don't treat it at all)

@emrysal that check is already in place somewhere else actually: https://github.com/calcom/cal.com/blob/main/apps/web/components/booking/BookingDescription.tsx#L49

leog:
@emrysal switched to ignoring the `duration` query param if it doesn't apply

emrysal (approved):

PR #5939, opened 2022-12-08
Direct links to accept/reject a booking

leog:
## What does this PR do?

Implement direct links in a email with required confirmation for a booking in order to accept or reject it, no need to log in.

Now when getting a booking that requires confirmation, it has 2 CTAs:
<kbd>image</kbd>

When clicking on accept:
<kbd></kbd>

When clicking on reject, it requires a reason to reject: 
<kbd></kbd>

Once the reason for reject is provided:
<kbd></kbd>

In case an already accepted/rejected booking is visited again:
<kbd></kbd>

In case a booking does not exist, using the 500 error with custom text:
<kbd></kbd>

In case the wrong URL is provided, hence the signature is not correct:
<kbd></kbd>

Closes #5807 

Internal Tech Specs for the implementation (WIP): https://app.gitbook.com/o/6snd8PyPYMhg0wUw6CeQ/s/VM7BFW5yP5ZxWKDW57nx/technical-specs/direct-link-actions-wip

 Loom Video: TBD

**Environment**: Staging(main branch) / Production

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How should this be tested?

Book required confirmation event-types and interact with the email CTAs.

PR #6151, opened 2022-12-21
Refactor/email confirm links

zomars:
## What does this PR do?

- Reverts some changes made in #5939
- Moves confirmation logic to `/api/link`
- Reuses Booking UI to display status
- Drops adding rejecting reason support (for now)
- Reuses tRPC logic
- Uses encryption/decryption in a single token instead of hashing

To do (maybe)

- Add backwards compatibility with old token (IDK if worth it)

## Type of change

## How should this be tested?

- [ ] Book a event with "requires confirmation"
- [ ] Test Accept/Reject buttons


zomars (commented):
Ready for review

apps/web/playwright/webhook.e2e.ts
```diff
@@ -57,6 +57,7 @@ test("add webhook & test that creating an event triggers a webhook call", async
     attendee.timeZone = dynamic;
     attendee.language = dynamic;
   }
+  body.payload.organizer.id = dynamic;
```
zomars, line 60: We now need this organizer id to generate the token

packages/lib/constants.ts
```diff
@@ -53,3 +53,4 @@ export const IS_STRIPE_ENABLED = !!(
 );
 /** Self hosted shouldn't checkout when creating teams unless required */
 export const IS_TEAM_BILLING_ENABLED = IS_STRIPE_ENABLED && (!IS_SELF_HOSTED || HOSTED_CAL_FEATURES);
+export const CALENDSO_ENCRYPTION_KEY = process.env.CALENDSO_ENCRYPTION_KEY || "";
```
emrysal, line 56: We could be easily leaking the CALENDSO_ENCRYPTION_KEY here, if there is any client side code including packages/lib/constants.ts - also this hardcodes it into the build (that's not that much of an issue and may be desirable)

emrysal (approved):
It looks good to me, nice cleanup, have a look at my comments and see if there's any merrit to my thought - will turn off automerge so you can eval.

alannnc (approved):
It works and couldn't find any security leak.

apps/web/pages/api/link.ts
```diff
@@ -0,0 +1,64 @@
...
+import { z } from "zod";
+
+import { CALENDSO_ENCRYPTION_KEY } from "@calcom/lib/constants";
+import { symmetricDecrypt } from "@calcom/lib/crypto";
+import { defaultResponder } from "@calcom/lib/server";
+import prisma from "@calcom/prisma";
+import { createContext } from "@calcom/trpc/server/createContext";
+import { viewerRouter } from "@calcom/trpc/server/routers/viewer";
+
+enum DirectAction {
```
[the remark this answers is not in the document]
zomars: just kept it as @leog made it. No particular reason

packages/lib/constants.ts
zomars: Good point

zomars: We use this pattern in multiple places tho 
image

zomars: Removed for now

emrysal: Those are OK - all those are serverside 👍  (😌)

leog (approved):
LGTM. Found some NITs comparing to old approach:
1. When an error happens (invalid token, booking already confirmed or rejected), we should redirect to page 500 with an error query param with the message
2. I think we should remove the following section of the booking page for this feature
<kbd>image</kbd>

apps/web/pages/api/link.ts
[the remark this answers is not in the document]
zomars: image

updated

zomars:
> LGTM. Found some NITs comparing to old approach:
> 
> 1. When an error happens (invalid token, booking already confirmed or rejected), we should redirect to page 500 with an error query param with the message

Agreed, we can follow up tho. Not critical IMO.

> 2. I think we should remove the following section of the booking page for this feature
>    image

We can discuss this. I don't think is that much of an issue.

leog:
> Agreed, we can follow up tho. Not critical IMO.

Agreed, which is why I approved it anyway. NITs shouldn't stop a PR.

> We can discuss this. I don't think is that much of an issue.

Not an issue, it just looks weird, an oversight. The person accepting/rejecting already has an account, it just looks sloppy IMO.

PR #6154, opened 2022-12-22
Hotfix/Event Type Save Error with Multiple Durations

hariombalhara:
## What does this PR do?

Fixes # (issue)
See [loom](https://www.loom.com/share/e040503589da4dbaa37a486cee5c3aa7)

**Environment**: Production

## Type of change

### DIFF
--- a/apps/web/components/eventtype/EventSetupTab.tsx
+++ b/apps/web/components/eventtype/EventSetupTab.tsx
@@ -290,7 +290,7 @@ export const EventSetupTab = (
                   setDefaultDuration(
                     selectedMultipleDuration.find((opt) => opt.value === option?.value) ?? null
                   );
-                  formMethods.setValue("length", option?.value);
+                  if (option) formMethods.setValue("length", option.value);
                 }}
               />
             </div>
--- a/apps/web/components/eventtype/EventSetupTab.tsx
+++ b/apps/web/components/eventtype/EventSetupTab.tsx
@@ -290,7 +290,7 @@ export const EventSetupTab = (
                   setDefaultDuration(
                     selectedMultipleDuration.find((opt) => opt.value === option?.value) ?? null
                   );
-                  formMethods.setValue("length", Number(option?.value));
+                  formMethods.setValue("length", option?.value);
                 }}
               />
             </div>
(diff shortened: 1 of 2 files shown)


apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -290,22 +290,19 @@ export const EventSetupTab = (
...
             required
-            name="length"
             type="number"
             label={t("duration")}
-            addOnSuffix={<>{t("minutes")}</>}
             defaultValue={eventType.length ?? 15}
-            onChange={(e) => {
-              formMethods.setValue("length", Number(e.target.value));
-            }}
+            {...formMethods.register("length")}
```
hariombalhara, line 304: Use standard RHF approach. This ensures that `length` is defined without onChange as well. But it causes length to be become a string and tRPC requires it to be a number. That has been handled by introducing zodResolver and adding a string to number transform.

Recent duration-related code checks `length` and it is  undefined if onchange hasn't occurred.

I could have fixed the duration code as well, but it seems unintuitive to have `length` be undefined.

apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -290,22 +290,19 @@ export const EventSetupTab = (
                   setDefaultDuration(
                     selectedMultipleDuration.find((opt) => opt.value === option?.value) ?? null
                   );
-                  formMethods.setValue("length", Number(option?.value));
+                  formMethods.setValue("length", option?.value);
```
hariombalhara, line 293: It was already a number here, because it is picked from options[number].value which is a number.

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -204,6 +205,16 @@ const EventTypePage = (props: inferSSRProps<typeof getServerSideProps>) => {
       ),
       metadata,
     },
+    resolver: zodResolver(
+      z
+        .object({
+          // Length if string, is converted to a number or it can be a number
+          // Make it optional because it's not submitted from all tabs of the page
+          length: z.union([z.string().transform((val) => +val), z.number()]).optional(),
```
Comment (hariombalhara) [inline apps/web/pages/event-types/[type]/index.tsx:213]: In Multiple durations case, it is already a number. In single duration case, it is a string and thus requires conversion to number.

hariombalhara (commented):
Self review done

emrysal (approved):

PR #6157, opened 2022-12-22
Chor/move table actions

sean-brydon:
Simple PR

Moves table actions to Calcom/UI

This is needed for an upcoming console PR for deleting users from a deployment


PeerRich (approved):
yessir

PR #6158, opened 2022-12-22
No more modules folder 🔥

sean-brydon:
Well new year new UI folder. 

Lets finally get rid of that v2 modules folder


zomars (commented):
Looking good. Will test first before approval

zomars (approved):
Noice

PR #6189, opened 2022-12-26
Make sure returnTo is correct URL

hariombalhara:
Details are in this private thread: https://threads.com/34439777102?s=d5UFfxEvSM6mMvqp1a5aGF


packages/lib/getSafeRedirectUrl.ts
```diff
@@ -5,15 +5,16 @@ export const getSafeRedirectUrl = (url = "") => {
...
-  if (![CONSOLE_URL, WEBAPP_URL, WEBSITE_URL].some((u) => new URL(u).origin === urlParsed.origin)) {
-    url = `${WEBAPP_URL}/`;
+  try {
+    const urlParsed = new URL(url);
+    // Avoid open redirection security vulnerability
+    if (![CONSOLE_URL, WEBAPP_URL, WEBSITE_URL].some((u) => new URL(u).origin === urlParsed.origin)) {
+      url = `${WEBAPP_URL}/`;
+    }
+  } catch (e) {
+    // If not an absolute URL
```
hariombalhara, line 16: Handle non absoluete URL case as well. Non absolute URLs would be safe to redirect to already.

zomars (approved):
LGTM

PR #6206, opened 2022-12-28
.env for root URL behavior+plausible #6193

### LINKED ISSUE #6193: Add control over root URL behavior

Currently, the root URL of a deployment (e.g. the value for `NEXT_PUBLIC_WEBAPP_URL`) redirects to `/auth/login`.

For self-hosters, this may not be a desired behavior.

For example, I'm interested in self-hosting a Cal.com deployment for my own, personal schedule. For my usecase, I _definitely_ don't want `NEXT_PUBLIC_WEBAPP_URL` to present a login screen, because that URL is easily guessable by the public (e.g. if I promote `https://schedule.josh.com/josh` people will probably try accessing `https://schedule.josh.com`).

It would be nice to have a first class feature to allow control over this behavior.

Some desired control functionality would be:
1. Allow me to redirect this URL to a specific user's page (e.g. from the example above `https://schedule.josh.com` > `https://schedule.josh.com/josh`) (easy, less preferable, straightforward with current architecture)
2. Allow me to show a specific user's page (e.g. `https://schedule.josh.com` shows the content of `https://schedule.josh.com/josh`) (harder, most preferable, not straightforward with current architecture?)

Editing source code to do Proposed Solution 1.

hybes:
## What does this PR do?

Adds 2 .env variables for Self Hosted Plausible URL, as well as root url to redirect users if they access the booking url without a session, this could be set to a particular users booking page, or to a landing page.

Fixes #6193 

## Type of change

- [✓ ] New feature (non-breaking change which adds functionality)
- [ ✓] This change requires a documentation update

## How should this be tested?

Configure the two new .env variables:

PLAUSIBLE_URL
ROOT_REDIR_URL

- You can see that traffic now sends to the self hosted Plausible URL, though you must include the full URL address, including the .js file.
- If you access the main WEBAPP_URL now without a session (incognito works), you should see you get redirected to the specified site.


.env.example
```diff
@@ -147,3 +147,9 @@ SENTRY_IGNORE_API_RESOLUTION_ERROR=
 NEXT_PUBLIC_APP_NAME="Cal.com"
 NEXT_PUBLIC_SUPPORT_MAIL_ADDRESS="help@cal.com"
 NEXT_PUBLIC_COMPANY_NAME="Cal.com, Inc."
+
+# Root URL redirection
+NEXT_PUBLIC_CUSTOM_ROOT_URL="https://cal.com"
```
PeerRich, line 152: this already exists

.env.example
```diff
@@ -147,3 +147,9 @@ SENTRY_IGNORE_API_RESOLUTION_ERROR=
...
+
+# Selfhost Plausible URL with js file
+PLAUSIBLE_URL="https://selfhosted-plausible.cal.com/js/plausible.js"
```
PeerRich, line 155: this should by default be the correct plausible url

Hybes:
updated to use env variables, not already in use, and to set default plausible URL

.env.example
```diff
@@ -147,3 +147,9 @@ SENTRY_IGNORE_API_RESOLUTION_ERROR=
 NEXT_PUBLIC_APP_NAME="Cal.com"
 NEXT_PUBLIC_SUPPORT_MAIL_ADDRESS="help@cal.com"
 NEXT_PUBLIC_COMPANY_NAME="Cal.com, Inc."
+
+# Root URL redirection
+ROOT_REDIR_URL="https://cal.com"
```
PeerRich, line 152: this exists already, no?

PeerRich:
@Hybes we're probably moving all the app settings into a database to not overload our .env

our goal is to have as little .env variables as possible and adding each app-settings into it doesn't seem like a good idea.

will keep this open but we're likely not merging it in this state -- instead we'll implement an "App Manager" where you can change the settings per instance

PeerRich (changes requested):
blocking until we have moved App-specific .env's into a database (App Manager cc @joeauyeung )

PR #6212, opened 2022-12-29
disables signups via env variable

### LINKED ISSUE #6192: Allow signups to be disabled

Currently, signups cannot be disabled.

This may be a problem for deployments that don't wish to allow new users (manually managed teams, individual people using a full deployment for their own schedule).

A first-class feature to disable signups.

Some of the things this would do include:
- Remove the link on the login page
- Make the form inaccessible if someone tried to access it directly
- Prevent any new user creation via API

Editing source files

phishy:
## What does this PR do?

Adds a new environment variable that disables signups:

-  remove link on login page
-  returns a 404 for /signup
-  returns a 403 for /api/auth/signup

Fixes #6192 

## Type of change

## How should this be tested?

- [ ] set env var `NEXT_PUBLIC_DISABLE_SIGNUP` to true
- [ ] ensure signup link is disabled
- [ ] ensure visiting /signup returns a 404
- [ ] ensure trying to make a new user via /api/auth/signup fails

## Checklist


PeerRich:
we don't wanna overload our .env, instead we're moving most modification settings to a database entry

phishy:
Yeah I saw you hit the 4K AWS limit, and the config table is in RFC. Is that right?

PeerRich:
> Yeah I saw you hit the 4K AWS limit, and the config table is in RFC. Is that right?

correct, in general its a much better UX to have a config table + UI + admin manager to make non-code changes to the product than having a convoluted .env file

PeerRich (dismissed):
blocking -- we'll move all of the non-code modifier to a database + admin settings page

phishy:
> > Yeah I saw you hit the 4K AWS limit, and the config table is in RFC. Is that right?
> 
> 
> 
> correct, in general its a much better UX to have a config table + UI + admin manager to make non-code changes to the product than having a convoluted .env file

Agreed. Env is for secrets.

ojengwa:
@phishy What's the status of this request?

PeerRich:
@ojengwa its currently blocked because we dont want to overload .env.

however, the database implementation is a few weeks away, @zomars should we merge this and later move the .env to a database / admin setup?

tofran:
@PeerRich I understand you might be having issues with your env size, but this feature should be opt-in, meaning you should not need to change your env variables in order to keep the app working as is.
Only users intending to block signups should need to add `NEXT_PUBLIC_DISABLE_SIGNUP` to their envs.
Thus, in my point of view, I don't see why this is blocked.

apps/web/pages/api/auth/signup.ts
```diff
@@ -12,6 +12,11 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     return;
   }
 
+  if (process.env.NEXT_PUBLIC_DISABLE_SIGNUP) {
```
tofran, line 15: We should do `if (process.env.NEXT_PUBLIC_DISABLE_SIGNUP === "true")` because as of the above code, any non empty string, including `"false"` is considered _truthy_.

Explanation: `!!"false"` and `Boolean("false")` is `true`.

Checking against the `true` string ensures backwards compatibility and that only users really intending to use this feature will disable signups.

tofran (changes requested):

PeerRich:
we will likely end up merging this and later extract this setting into an admin database @zomars @joeauyeung

.env.example
```diff
@@ -147,3 +147,4 @@ SENTRY_IGNORE_API_RESOLUTION_ERROR=
 NEXT_PUBLIC_APP_NAME="Cal.com"
 NEXT_PUBLIC_SUPPORT_MAIL_ADDRESS="help@cal.com"
 NEXT_PUBLIC_COMPANY_NAME="Cal.com, Inc."
+NEXT_PUBLIC_DISABLE_SIGNUP=false
```
zomars, line 150:
```suggestion
# Set this to true in to disable new signups
# NEXT_PUBLIC_DISABLE_SIGNUP=true
NEXT_PUBLIC_DISABLE_SIGNUP=
```

apps/web/pages/api/auth/signup.ts
zomars, line 15:
```suggestion
  if (process.env.NEXT_PUBLIC_DISABLE_SIGNUP === 'true') {
```

apps/web/pages/signup.tsx
```diff
@@ -145,6 +145,12 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
     prepopulateFormValues: undefined,
   };
 
+  if (process.env.NEXT_PUBLIC_DISABLE_SIGNUP) {
```
zomars, line 148:
```suggestion
  if (process.env.NEXT_PUBLIC_DISABLE_SIGNUP === 'true') {
```

apps/web/pages/auth/login.tsx
```diff
@@ -115,7 +115,9 @@ export default function Login({
         description={t("login")}
         showLogo
         heading={twoFactorRequired ? t("2fa_code") : t("welcome_back")}
-        footerText={twoFactorRequired ? TwoFactorFooter : LoginFooter}>
+        footerText={
+          twoFactorRequired ? TwoFactorFooter : !process.env.NEXT_PUBLIC_DISABLE_SIGNUP ? LoginFooter : null
```
zomars, line 119:
```suggestion
          twoFactorRequired ? TwoFactorFooter : process.env.NEXT_PUBLIC_DISABLE_SIGNUP !== 'true' ? LoginFooter : null
```

zomars (approved):

LGTM

PR #6221, opened 2022-12-30
Make cal.com version available on all pages, even on login page.

hariombalhara:
It is useful for debugging purposes to know what version is being used on a cal deployment.
Making it available in a variable allows it to be retrieved programmatically for other uses as well.

edit by @PeerRich:

- added to login page
- added "ee" (enterprise edtion) and "ce" (community edition) to string

CleanShot 2023-01-01 at 12 16 32


apps/web/pages/_app.tsx
```diff
@@ -32,7 +33,9 @@ function MyApp(props: AppProps) {
   }
   // Use the layout defined at the page level, if available
   const getLayout = Component.getLayout ?? ((page) => page);
-
+  const calComVersion = `${pkg.version}-${
+    process.env.NEXT_PUBLIC_WEBSITE_URL === "https://cal.com" ? "h" : "sh"
+  }`;
```
PeerRich, line 38: we have this in the sidebar footer, no? lets turn that into a helper function

apps/web/pages/_app.tsx
```diff
@@ -44,6 +47,10 @@ function MyApp(props: AppProps) {
       <Head>
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
       </Head>
+      <Script
+        id="calcom-version"
+        dangerouslySetInnerHTML={{ __html: `window.CalComVersion = '${calComVersion}'` }}
+      />
```
PeerRich, line 53: how would the user know the version then? they would need to log it themselves, right?

hariombalhara: Yeah, this is mostly for debugging. But if we can identify a place to show on all pages(public/non public), we can show it in UI as well

apps/web/pages/_app.tsx
hariombalhara: Yeah, thought about that. But skipped the idea because it was a small code. But let me do it now.

hariombalhara: Done

packages/lib/constants.ts
```diff
@@ -53,3 +56,6 @@ export const IS_STRIPE_ENABLED = !!(
 );
 /** Self hosted shouldn't checkout when creating teams unless required */
 export const IS_TEAM_BILLING_ENABLED = IS_STRIPE_ENABLED && (!IS_SELF_HOSTED || HOSTED_CAL_FEATURES);
+export const CalComVersion = `v.${pkg.version}-${
+  process.env.NEXT_PUBLIC_WEBSITE_URL === "https://cal.com" ? "h" : "sh"
+}`;
```
PeerRich, line 61: can we add here also if someone has a commercial license or not? we may need to move this from const into a proper component

`session.data?.hasValidLicense ? "l" : "n"` (l = license, n = no license)

PeerRich: i.e.: v.2.4.2-sh-l

PeerRich: the problem with session.data is that it's false when logged out because session is being destroyed when logging out

@zomars is there a way to know if someone has a EE license without checking session?

PeerRich: i am now checking for process.env.CALCOM_LICENSE_KEY is empty which should be enough. people could enter an invalid key but why would they? doesnt benefit them since the EE components need a license

packages/lib/constants.ts
```diff
@@ -53,3 +56,7 @@ export const IS_STRIPE_ENABLED = !!(
 );
 /** Self hosted shouldn't checkout when creating teams unless required */
 export const IS_TEAM_BILLING_ENABLED = IS_STRIPE_ENABLED && (!IS_SELF_HOSTED || HOSTED_CAL_FEATURES);
+
+export const CalComVersion = `v.${pkg.version}-${
+  process.env.NEXT_PUBLIC_WEBSITE_URL === "https://cal.com" ? "h" : "sh"
+}-${process.env.CALCOM_LICENSE_KEY === "" ? "ee" : "ce"}`;
```
hariombalhara, line 62: @PeerRich what does ce stand for?

packages/ui/v2/core/VersionInfo.tsx
```diff
@@ -0,0 +1,13 @@
+import { COMPANY_NAME, CalComVersion } from "@calcom/lib/constants";
+
+export default function VersionInfo() {
+  return (
+    <small
+      style={{
+        fontSize: "0.5rem",
```
JeroenReumkens, line 7: You can use an arbitrary value in tailwind for this; `text-[0.5rem]`. That way we don't need inline styles 😁

packages/ui/v2/core/VersionInfo.tsx
```diff
@@ -0,0 +1,13 @@
+import { COMPANY_NAME, CalComVersion } from "@calcom/lib/constants";
```
JeroenReumkens, line 1: Can you please add this component outside of the v2 folder? New components can be added inside `packages/ui/components`. We try to get rid of v2 asap.

JeroenReumkens (dismissed):
Small change requested, but overall great change!

packages/lib/constants.ts
PeerRich: community edition

packages/ui/v2/core/VersionInfo.tsx
PeerRich: i dont quite understand how I add it in ui/components

do i need an index.tsx?

PeerRich:
removed VersionInfo component and hardcoded string since its not likely to be reused again

apps/web/pages/_app.tsx
```diff
@@ -44,6 +44,10 @@ function MyApp(props: AppProps) {
       <Head>
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
       </Head>
+      <Script
+        id="calcom-version"
+        dangerouslySetInnerHTML={{ __html: `window.CalComVersion = '${CalComVersion}'` }}
+      />
```
PeerRich, line 50: @hariombalhara where do we use this?

hariombalhara: Not anywhere yet. But it might be useful for other tools on-page to know what version cal.com is running on.

PeerRich (approved):

PR #6226, opened 2022-12-31
Show admin password top banner

sean-brydon:
Adds top banner error if you are admin but do not have the right password length 

Admin with correct password length:
CleanShot 2022-12-31 at 12 57 16@2x

ADMIN without valid password:
CleanShot 2022-12-31 at 13 01 48@2x

Normal User:
CleanShot 2022-12-31 at 13 02 26@2x

### DIFF
--- a/packages/types/next-auth.d.ts
+++ b/packages/types/next-auth.d.ts
@@ -16,6 +16,6 @@ declare module "next-auth" {
     impersonatedByUID?: number;
     belongsToActiveTeam?: boolean;
     username?: PrismaUser["username"];
-    role?: PrismaUser["role"];
+    role?: PrismaUser["role"] | "INACTIVE_ADMIN";
   }
 }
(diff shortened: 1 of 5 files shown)


packages/types/next-auth.d.ts
```diff
@@ -16,6 +16,6 @@ declare module "next-auth" {
     impersonatedByUID?: number;
     belongsToActiveTeam?: boolean;
     username?: PrismaUser["username"];
-    role?: PrismaUser["role"];
+    role?: PrismaUser["role"] | "INACTIVE_ADMIN";
```
sean-brydon, line 19: We dont need this on a database level only on a session obj

packages/ui/v2/core/Shell.tsx
```diff
@@ -157,6 +158,7 @@ const Layout = (props: LayoutProps) => {
         <div className="divide-y divide-black">
           <TeamsUpgradeBanner />
           <ImpersonatingBanner />
+          <AdminPasswordBanner />
```
PeerRich, line 161: not important but what happens if you have multiple banners?

PeerRich (approved):

PR #6235, opened 2023-01-02
a11y: right-to-left improvements

### LINKED ISSUE #6195: [CAL-653] investigate rtl layout issues
and use ltr:ml-4 rtl:mr-4 etc.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-653](https://linear.app/calcom/issue/CAL-653/investigate-rtl-layout-issues)</sub>

PeerRich:
to test, change dir="rtl"

CleanShot 2023-01-02 at 15 57 48@2x

https://user-images.githubusercontent.com/8019099/210244585-bdf1183d-037f-4064-8fed-15a268dd1149.mp4


apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
```diff
@@ -110,26 +118,20 @@ const UsernameTextfield = (props: ICustomUsernameProps) => {
...
-        </span>
         <div className="relative w-full">
-          <Input
+          <TextField
             ref={usernameRef}
             name="username"
             value={inputUsernameValue}
+            addOnLeading={
+              <>{process.env.NEXT_PUBLIC_WEBSITE_URL.replace("https://", "").replace("http://", "")}/</>
+            }
```
PeerRich, line 129: replaced Input with Textfield

this needs to be tested

PeerRich:
@Udit-takkar can you test both rtl and ltr on the vercel preview?

JeroenReumkens:
If we want to start doing this I think it's good if we communicate with everyone that they should take this into account. Otherwise we'll soon have a situation where this partially works, but breaks for new functionality. What do you think? 

Besides that, Tailwind will soon (finally 😬) add support for `margin-inline-start` and `-end`. That's actually a more modern way to do this. If you use that instead of margin left/right, it will automatically be added on left/right side based on whether the user has rtl enabled. So that means you don't need to define the margin twice (for ltr and rtl). See https://github.com/tailwindlabs/tailwindcss/pull/10166 

I know you've already done a bunch of work looking at the PR, but maybe it's worth it to wait for this change to get merged soon?

packages/ui/v2/core/Shell.tsx
```diff
@@ -282,7 +282,7 @@ function UserDropdown({ small }: { small?: boolean }) {
                   </span>
                 </span>
                 <Icon.FiMoreVertical
-                  className="h-4 w-4 flex-shrink-0 text-gray-400 group-hover:text-gray-500 rtl:mr-2"
+                  className="h-4 w-4 flex-shrink-0 text-gray-400 group-hover:text-gray-500 ltr:mr-2 rtl:ml-2 rtl:mr-4"
```
JeroenReumkens, line 285: Is it correct that on ltr it only has a margin on right side, and on rtl it has a margin on both sides?

packages/ui/v2/core/navigation/tabs/VerticalTabItem.tsx
```diff
@@ -50,15 +50,17 @@ const VerticalTabItem = function ({
                 props.textClassNames || "text-sm font-medium leading-none text-gray-600",
                 "min-h-9 group flex w-64 flex-row items-center rounded-md px-2 py-[10px] hover:bg-gray-100 group-hover:text-gray-700 [&[aria-current='page']]:bg-gray-200 [&[aria-current='page']]:text-gray-900",
                 props.disabled && "pointer-events-none !opacity-30",
-                (isChild || !props.icon) && "ml-7 mr-5 w-auto",
+                (isChild || !props.icon) && "ml-7 w-auto ltr:mr-5 rtl:ml-5",
```
JeroenReumkens, line 53: There's still a `ml` here, should that be removed?

JeroenReumkens (approved):
Checked the code, apart from 2 small questions that looks good. 
So if we decide to go with this, instead of wait for https://github.com/tailwindlabs/tailwindcss/pull/10166, then only Udit needs to test the website on preview and we're good to go in my opinion :) 

cc @PeerRich

PR #6239, opened 2023-01-02
fix: use hasMounted to render calcom versionm

Udit-takkar:
## What does this PR do?

Fixes Hydration error in /event-types page

Ref: https://www.joshwcomeau.com/react/the-perils-of-rehydration/#the-solution

## Type of change


PeerRich:
i bet there is a better workaround. if not we should move this into a standalone component to prevent repeating code

PeerRich:
i think by turning the constant into a react component, it would not have hydration issues. wanna try that?

PeerRich:
```
export default function Credits() {
const CalComVersion = `v.${pkg.version}-${
  process.env.NEXT_PUBLIC_WEBSITE_URL === "https://cal.com" ? "h" : "sh"
}-${process.env.CALCOM_LICENSE_KEY === "" ? "ee" : "ce"}`;

 return <small className="mx-3 mt-1 mb-2 hidden text-[0.5rem] opacity-50 lg:block">
            &copy; {new Date().getFullYear()} {COMPANY_NAME} {CalComVersion}
 </small>
}
```

PeerRich (changes requested):
lets turn this into a component `<Credits>`

https://github.com/calcom/cal.com/pull/6239#issuecomment-1369117119

PeerRich (approved):
made some changes, created `<Credits>`

apps/web/pages/_app.tsx
```diff
@@ -44,10 +43,6 @@ function MyApp(props: AppProps) {
       <Head>
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
       </Head>
-      <Script
-        id="calcom-version"
-        dangerouslySetInnerHTML={{ __html: `window.CalComVersion = '${CalComVersion}'` }}
-      />
       {getLayout(
```
PeerRich, line 46: @hariombalhara i'm removing this until we need it

PeerRich (approved):

apps/web/pages/_app.tsx
hariombalhara: Cool

PR #6243, opened 2023-01-03
Fix lang and dir attributes not being set

### LINKED ISSUE #6237: [CAL-667] update lang="" in <html> when changing language
updating lang="<ISO>" is required so we can apply a different font for arabic. also helps with screen readers

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-667](https://linear.app/calcom/issue/CAL-667/update-lang=-in-html-when-changing-language)</sub>

hariombalhara:
## What does this PR do?

Fixes #6237

## Type of change

### DIFF
--- a/packages/lib/i18n.ts
+++ b/packages/lib/i18n.ts
@@ -2,10 +2,7 @@
 import parser from "accept-language-parser";
 import { IncomingMessage } from "http";
 
-import prisma from "@calcom/prisma";
 import { Maybe } from "@calcom/trpc/server";
-
-import { getSession } from "./auth";
 
 const { i18n } = require("@calcom/config/next-i18next.config");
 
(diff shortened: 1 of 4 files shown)


packages/lib/i18n.ts
```diff
@@ -17,33 +14,5 @@ export function getLocaleFromHeaders(req: IncomingMessage): string {
   return preferredLocale ?? i18n.defaultLocale;
 }
 
-export const getOrSetUserLocaleFromHeaders = async (req: IncomingMessage): Promise<string> => {
```
hariombalhara, line 20: Wasn't being used anywhere

PeerRich (approved):
perfect! that works! 👏

PR #6247, opened 2023-01-03
Move Skeleton to ui/components

sean-brydon:
Moves skeleton to ui/components instead of living in the V2 Directory.


packages/ui/v2/core/Shell.tsx
```diff
@@ -38,9 +38,9 @@ import {
 import ErrorBoundary from "../../ErrorBoundary";
 import { KBarContent, KBarRoot, KBarTrigger } from "../../Kbar";
 import Logo from "../../Logo";
+import { SkeletonText } from "../../components/skeleton";
```
PeerRich, line 41: long term this will be @calcom/components right?

JeroenReumkens: In general best practice within the UI folder itself is to use relative imports for everything. That will drastically reduce the amount of recursive imports. So in my opinion this is still up for debate.

JeroenReumkens (approved):
Looks good Sean, great work! Also checked it locally and couldn't find weird behaviour in Skeletons 🙏

PR #6261, opened 2023-01-04
🧹 Chore/move v2 components to components dir

JeroenReumkens:
## What does this PR do?

Removes quite a few root ui/v2/core/... files, and moves what's still used to new ui/components. 
This PR doesn't tackle all of them yet. Step by step. 

- [x] Change target branch once other branche is merged. 

## Type of change

## How should this be tested?

...


packages/app-store/ee/routing-forms/components/SingleForm.tsx
```diff
@@ -362,14 +362,14 @@ function SingleForm({ form, appUrl, Page }: SingleFormComponentProps) {
                     </Button>
                   </div>
                   {!form._count?.responses && (
-                    <Banner
```
JeroenReumkens, line 365: This was the only instance where the banner component was used, I replaced it with the `<Alert>` component that looks practically similar. However, it doesn't have the dismiss button (I didn't implement it). This dismiss button in the Banner was fake though, it would hide the message, but would be visible the next time. So I think we can do without here 😁

packages/ui/components/alert/Alert.tsx
```diff
@@ -2,13 +2,18 @@ import { CheckCircleIcon, ExclamationIcon, InformationCircleIcon, XCircleIcon }
 import classNames from "classnames";
 import { ReactNode } from "react";
 
+import { Icon } from "../icon";
+
 export interface AlertProps {
   title?: ReactNode;
+  // @TODO: Message should be children, more flexible?
```
JeroenReumkens, line 9: I think this component in general can still use some improvements (it's also not really according to design). So I left some todo's for when we are updating this component.

packages/ui/components/divider/Divider.tsx
```diff
@@ -1,6 +1,11 @@
 import { classNames } from "@calcom/lib";
 
-export default function VerticalDivider({ className, ...props }: JSX.IntrinsicElements["svg"]) {
+export function Divider({ className, ...props }: JSX.IntrinsicElements["hr"]) {
```
JeroenReumkens, line 3: Merged the divider and vertical divider into one file. Makes us have less files 🙂

packages/ui/components/index.ts
```diff
@@ -26,11 +26,15 @@ export {
...
   SelectField,
   SelectWithValidation,
+  TimezoneSelect,
 } from "./form";
+export type { ITimezone, ITimezoneOption } from "./form";
 export {
   AllApps,
   AppCard,
   AppSkeletonLoader,
+  // @deprecated: Please use AppSkeletonLoader instead
```
JeroenReumkens, line 36: @sean-brydon  Added this in temporary so we won't break console. Agree? 😄

sean-brydon: No - Places across the app are using AppSkeletonLoader (The skeleton loader for the app store) instead of the SkeletonLoader component we have. 

All instances of AppSkeletonLoader should be skeleton loader appart from the ones in the appstore files

```suggestion
```

packages/ui/components/divider/Divider.tsx
sean-brydon: Agreed - no reason to have two divider files - especially when the components are this small.

packages/ui/components/alert/Alert.tsx
```diff
@@ -2,13 +2,18 @@ import { CheckCircleIcon, ExclamationIcon, InformationCircleIcon, XCircleIcon }
...
   message?: ReactNode;
+  // @TODO: Provide action buttons so style is always the same.
   actions?: ReactNode;
```
sean-brydon, line 12: I like this - however, we had a convo about this in dialog and no one used it and preferred passing in actions themselves as children

sean-brydon (approved):
LGTM - few nits about comments. But not blocking

packages/ui/components/alert/Alert.tsx
JeroenReumkens: Yeah fair point. I think it's good if we still have it as a ReactNode, but I'm thinking about exporting a custom component `AlertButton` for example, which at least ensures the buttons are always styled the same. Because right now, everyone adds their own buttons in there, so it's guaranteed that stuff will never look the same.

Udit-takkar (approved):
LGTM. just one thing we should i18 in routing forms pages.

packages/app-store/ee/routing-forms/components/SingleForm.tsx
```diff
@@ -362,14 +362,14 @@ function SingleForm({ form, appUrl, Page }: SingleFormComponentProps) {
...
-                      description="Wait for some time for responses to be collected. You can go and submit the form yourself as well."
-                      Icon={Icon.FiInfo}
-                      onDismiss={() => console.log("dismissed")}
-                    />
+                    <>
+                      <Alert
+                        className="mt-6"
+                        severity="neutral"
+                        title="No responses yet"
+                        message="Wait for some time for responses to be collected. You can go and submit the form yourself as well."
```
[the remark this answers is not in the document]
sean-brydon: I believe this has just been moved from one place to another... 

However, you are completely right

PR #6293, opened 2023-01-05
Make WYSIWYG editor reusable

### LINKED ISSUE #6267: [CAL-676] Make WYSIWYG Editor reusable
Make the WYSIWYG Editor reusable. So we can use it for event type description and the about section ([https://github.com/calcom/cal.com/pull/6216](https://github.com/calcom/cal.com/pull/6216))

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-676](https://linear.app/calcom/issue/CAL-676/make-wysiwyg-editor-reusable)</sub>

CarinaWolli:
## What does this PR do?

- Makes text editor that is used for custom workflow templates reusable
  - By default the toolbar includes block type, bold, italic and link. The new prop `excludedToolbarItems: string[]` allows to exclude the ones not needed.
- Also makes the 'Add variable' dropdown reusable by introducing a new prop `variables: string[]` (if variables is undefined or empty the 'Add variable' button is not showing). 
  - renamed all translation keys for dynamic text variables form `_workflow` to `_variable`

- Block type dropdown had a broken design. We now use our own Dropdown component

Fixes #6267

 Loom Video 1. (shows that everything is working with new dropdown and the possibility to remove 'Add Variable' button): https://www.loom.com/share/a8c90b69264d45a1950bd14ca2637e2c

Loom Video 2 (shows how to remove items from the toolbar): https://www.loom.com/share/ed37bdfbef864bacb5024ace67337b29

## How should this be tested?

- Test all functionalities of the editor and see if everything is still working as expected


packages/features/ee/workflows/lib/variableTranslations.ts
```diff
@@ -21,14 +21,14 @@ export function getTranslatedText(text: string, language: { locale: string; t: T
     });
 
     variables?.forEach((variable) => {
-      const regex = new RegExp(variable, "g"); // .replaceAll is not available here for some reason
-      const translatedVariable = originalVariables.includes(variable.toLowerCase().concat("_workflow"))
-        ? language.t(variable.toLowerCase().concat("_workflow")).replace(/ /g, "_").toLocaleUpperCase()
-        : originalVariables.includes(variable.toLowerCase().concat("_name_workflow")) //for the old variables names (ORGANIZER_NAME, ATTENDEE_NAME)
-        ? language.t(variable.toLowerCase().concat("_name_workflow")).replace(/ /g, "_").toLocaleUpperCase()
+      const regex = new RegExp(`{${variable}}`, "g"); // .replaceAll is not available here for some reason
```
CarinaWolli, line 24: This (+ changes in line 31) fixed a bug in translating some variables in some languages

packages/ui/components/editor/plugins/ToolbarPlugin.tsx
```diff
@@ -220,51 +227,13 @@ function getSelectedNode(selection: RangeSelection) {
...
 }
 
-type BlockOptionsDropdownProps = {
-  editor: LexicalEditor;
-  blockType: string;
-  toolbarRef: RefObject<HTMLDivElement>;
-  setShowBlockOptionsDropDown: Dispatch<SetStateAction<boolean>>;
-};
-
-function BlockOptionsDropdownList({
```
CarinaWolli, line 230: Removing the BlockOptionsDropdownList as we are now using our `Dropdown` component

packages/ui/components/editor/plugins/ToolbarPlugin.tsx
```diff
@@ -457,69 +397,94 @@ export default function ToolbarPlugin(props: TextEditorProps) {
...
-              createPortal(
-                <BlockOptionsDropdownList
-                  editor={editor}
-                  blockType={blockType}
-                  toolbarRef={toolbarRef}
-                  setShowBlockOptionsDropDown={setShowBlockOptionsDropDown}
-                />,
-                document.body
-              )}
+            <Dropdown>
```
CarinaWolli, line 402: Using our `Dropdown` component here

packages/ui/components/editor/Editor.tsx
```diff
@@ -11,28 +11,32 @@ import { MarkdownShortcutPlugin } from "@lexical/react/LexicalMarkdownShortcutPl
...
 import { TableCellNode, TableNode, TableRowNode } from "@lexical/table";
-import { UseFormReturn } from "react-hook-form";
 
-import { FormValues } from "../../pages/workflow";
 import ExampleTheme from "./ExampleTheme";
 import AutoLinkPlugin from "./plugins/AutoLinkPlugin";
 import ToolbarPlugin from "./plugins/ToolbarPlugin";
 import "./stylesEditor.css";
 
+/*
```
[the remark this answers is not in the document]
CarinaWolli: Or @PeerRich @sean-brydon maybe?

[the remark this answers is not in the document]
sean-brydon: I don't see any issue with this living here :)

PeerRich:
testing right now.

i know the following changes are not part of this PR but could be addressed here:


not centered:
CleanShot 2023-01-09 at 14 55 20@2x

active and hover should not have border radius:
image

PeerRich:
i dont think we will bring back test action due to the potential for abuse so we should remove it

CleanShot 2023-01-09 at 14 57 59@2x

PeerRich (approved):
tested and works! added comments unrelated to this PR which can be addressed here or in a follow up PR

leog:
@CarinaWolli Looks good, agree with Peer's feedback on UI. The only NIT I would mention is maintaining two different lists of variables, one with "_variable" at the end and the other without that suffix, maybe one list could live in a constants file if makes sense, so it can be used in different places.

PR #6297, opened 2023-01-05
Add start and end time of rescheduled booking to webhook payload

### LINKED ISSUE #6223: [CAL-658] webhook rescheduling, keep UID for old appointment
from a user:

> I utilize webhooks heavily from your service to get data into a custom workflow. I have noticed that when rescheduling an appointment (instead of canceling it and then rebooking it for another time) that the payload for the webhook does not mention the old appointment time or UID it was rescheduled from.
> 
> It appears that the way that rescheduling works is that it deletes the old UID of the event and creates a new one, along with the new time. For my purposes, I need to know both the old and new UID’s when rescheduling the appointment. Is there a way to get this information (old and new UID) in the webhook data? If not, can this functionality be added?

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-658](https://linear.app/calcom/issue/CAL-658/webhook-rescheduling-keep-uid-for-old-appointment)</sub>

CarinaWolli:
## What does this PR do?

Adds start and end time of rescheduled booking to webhook payload. 

Fixes #6223 => booking UID of rescheduled booking is already included in payload (rescheduleUid), but when a booking is rescheduled the old booking is deleted

Screenshot 2023-01-05 at 17 04 16


zomars:
Do we need to update the webhook snapshot as well? @CarinaWolli

CarinaWolli:
@zomars There are only tests for the Booking Created webhook trigger and not the rescheduling trigger. The snapshot for creating a booking didn't change

zomars (commented):
LGTM

zomars (approved):

PR #6301, opened 2023-01-05
Fixes extension check for URL with dots in segments

zomars:
## What does this PR do?

Previous check didn't account for URLs with dots in segments like: `https://example.com/remote.php/dav/calendars/example/`

## Type of change

## How should this be tested?

- [ ] Add a CalDav calendar without a file extension, it should be ignored (for now)

### DIFF
--- a/packages/lib/CalendarService.ts
+++ b/packages/lib/CalendarService.ts
@@ -37,6 +37,22 @@ const DEFAULT_CALENDAR_TYPE = "caldav";
 
 const CALENDSO_ENCRYPTION_KEY = process.env.CALENDSO_ENCRYPTION_KEY || "";
 
+function hasFileExtension(url: string): boolean {
+  // Get the last portion of the URL (after the last '/')
+  const fileName = url.substring(url.lastIndexOf("/") + 1);
+  // Check if the file name has a '.' in it and no '/' after the '.'
+  return fileName.includes(".") && !fileName.substring(fileName.lastIndexOf(".")).includes("/");
+}
+
+function getFileExtension(url: string): string | null {
+  // Return null if the URL does not have a file extension
+  if (!hasFileExtension(url)) return null;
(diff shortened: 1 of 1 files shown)


packages/lib/CalendarService.ts
```diff
@@ -259,14 +275,14 @@ export default abstract class BaseCalendarService implements Calendar {
   }
 
   isValidFormat = (url: string): boolean => {
-    const acceptedFormats = ["eml", "ics"];
-    const urlFormat = url.split(".").pop();
```
emrysal, line 263: Not a fan of this rename, formats is a better descriptor than extension - NIT - in what scenario did this not work (what is the fail case?)

emrysal (approved):

packages/lib/CalendarService.ts
```diff
@@ -37,6 +37,22 @@ const DEFAULT_CALENDAR_TYPE = "caldav";
 
 const CALENDSO_ENCRYPTION_KEY = process.env.CALENDSO_ENCRYPTION_KEY || "";
 
+function hasFileExtension(url: string): boolean {
+  // Get the last portion of the URL (after the last '/')
+  const fileName = url.substring(url.lastIndexOf("/") + 1);
+  // Check if the file name has a '.' in it and no '/' after the '.'
+  return fileName.includes(".") && !fileName.substring(fileName.lastIndexOf(".")).includes("/");
```
emrysal, line 44:
```javascript
!fileName.substring(fileName.lastIndexOf(".")).includes("/");
// ^ This check is impossible, redundant I think
```

zomars: You're right. This also won't cover URL with query parameters like: `https://example.com/remote.php/dav/calendars/example/?someUrlQuery=/i-will-fail.ics`

PR #6307, opened 2023-01-06
Chor/v2 UI tidy - Dialog & Toasts

sean-brydon:
Tidies up Dialog and Toasts moves from v2 into ui/components

Few changes to toasts - extracted each variant into a component so we can demo it in storybook 
- Created storybook file for toasts

CleanShot 2023-01-06 at 11 33 09@2x


packages/features/settings/TimezoneChangeDialog.tsx
```diff
@@ -4,9 +4,7 @@ import { useEffect, useState } from "react";
 import dayjs from "@calcom/dayjs";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { trpc } from "@calcom/trpc/react";
-
-import { showToast } from ".";
-import { Dialog, DialogClose, DialogContent, DialogFooter } from "../ui/v2";
+import { Dialog, DialogClose, DialogContent, DialogFooter, showToast } from "@calcom/ui";
```
sean-brydon, line 7: File moved to features so we can import from `calcom/ui`

zomars (approved):
LGTM

zomars:
Let's check tests and fix conflicts tho

PR #6321, opened 2023-01-07
Fix/Next.js-13/Circular Dependency in Routing Forms

hariombalhara:
## What does this PR do?

Before:
image

Also, fixes an unstable test.

The issue **comes only in the dev** environment and is caused by an update to Next.js 13. Next.js-13's import related optimizations seem to be the cause of it.

## Type of change


PeerRich (approved):

PeerRich:
tested, works!

PR #6326, opened 2023-01-07
Replaces mdx-remote with markdown-it

zomars:
## What does this PR do?

Spent too much time trying to debug [why `next-mdx-remote`doesn't work with Next.js 13](https://github.com/hashicorp/next-mdx-remote/issues/307) in development mode.

Then I realized we're not using MDX full potential anymore since we're not using custom React components like `Slider` in mdx.

~~This PR switches the markdown rendered to `react-markdown` since it's more than enough for our use case that is rendering simple markdown and images.~~

- Got rid of MDX altogether as we don't need it anymore
- Uses markdown it instead
- Renamed all README.mdx files to DESCRIPTION.md so we don't conflicts with README.md from some apps (zapier, stripe, etc.) Also README files are tailored more for GitHub/devs.

## Type of change

## How should this be tested?

- [ ] Go to any app inside `/apps`
- [ ] Description and images should render properly in both development and production builds.


emrysal:
Could you run a quick comparison of markdown-it and react-markdown? (This introduces react-markdown but we already have markdown-it, I think we should choose one)

PeerRich:
> Could you run a quick comparison of markdown-it and react-markdown? (This introduces react-markdown but we already have markdown-it, I think we should choose one)

what @emrysal says + E2E is failing 🤔

zomars:
> Could you run a quick comparison of markdown-it and react-markdown? (This introduces react-markdown but we already have markdown-it, I think we should choose one)

I wasn't aware that we used markdown-it. Will test ASAP

zomars:
- Got rid of MDX altogether as we don't need it anymore
- Uses markdown it instead
- Renamed all README.mdx files to DESCRIPTION.md so we don't conflicts with README.md from some apps (zapier, stripe, etc.) Also README files are tailored more for GitHub/devs.

emrysal (approved):
👍 dangerouslySetInnerHTML is good to use here as we need to review incoming DESCRIPTION.md's anyway.

PR #6330, opened 2023-01-07
set useQueryParam to false for app categories

emrysal:
## What does this PR do?

See subject.


zomars (approved):

emrysal:
Back in PR - it's used in three places (1) the admin settings, without query param (2) the admin setup, with query param (3) the installed apps, without query param also. Addressed all cases.

PR #6367, opened 2023-01-10
Chor/bulk UI migration

sean-brydon:
Removes V2 folder - fixes shell - does pretty much everything :D 

I know this is massive im sorry. i got carried away


JeroenReumkens (approved):
Amazing work @sean-brydon So many cleanups, such wow.

PR #6396, opened 2023-01-11
Fixes ultrawide breaking booking limits selects

sean-brydon:
Before this page broke on ultrawides
image

After:
CleanShot 2023-01-11 at 10 05 00@2x


PeerRich:
good fix but in general we should not use w-1/2 at all

PeerRich (approved):

PR #6418, opened 2023-01-11
Removes 'Test Action' button in workflows

CarinaWolli:
## What does this PR do?

Removes the disabled 'Test Action' button in workflows.


CarinaWolli:
I commented out the code instead of deleting it in case we are trying to reintroduce that feature again in the future. @zomars Or what do you think about that, should I delete the code instead?

alannnc (approved):
I think commented it's good with a note why it was removed.

zomars (approved):
Ship it

PR #6343, opened 2023-01-09
feat: Adding OPTIMIZE_FAIRNESS as a distribution method

emrysal:
## What does this PR do?

Adds a new distribution method / algorithm to optimize fairness.

image

PR #6423, opened 2023-01-12
Feature/fixed hosts

### LINKED ISSUE #5605: [CAL-408] Fixed Round Robin
MVP of Fixed Round Robin

Person A is fixed Person B round robins through a pool of individuals selected in the event type

> From [SyncLinear.com](https://synclinear.com) | [CAL-408](https://linear.app/calcom/issue/CAL-408/fixed-round-robin)

emrysal:
## What does this PR do?

> **⚠️ This PR touches core booking logic**

Implements fixed round robin hosts; which brings COLLECTIVE behaviour into round-robin.

* Starts the deprecation process of the `users` relationship in `EventType` (in favour of a new table called `Host` which is only intended to be used for teams and group events).
* This is a requirement for finishing #6343

image

Resolves #5605


CarinaWolli:
@emrysal I found an issue when trying to set the hosts. The dropdown options are not updating, so this happens once you try to add the same user to both fixed hosts and round-robin hosts (without saving in between)

https://www.loom.com/share/edcdd44cbac94cd3a4aa26426f316c48

CarinaWolli (dismissed):
Awesome work @emrysal 👏🏻 Everything (except the dropdown issue I mentioned in my comment) seems to work. Code also looks good, just left one small comment. 
After the dropdown issue is resolved I can approve 🙏

packages/core/getAggregateWorkingHours.ts
```diff
@@ -7,16 +7,29 @@ import type { WorkingHours } from "@calcom/types/schedule";
...
   schedulingType: SchedulingType | null
 ): WorkingHours[] => {
-  if (schedulingType !== SchedulingType.COLLECTIVE) {
+  // during personal events, just flatMap.
+  if (!schedulingType) {
     return usersWorkingHoursAndBusySlots.flatMap((s) => s.workingHours);
   }
-  return usersWorkingHoursAndBusySlots.reduce((currentWorkingHours: WorkingHours[], s) => {
+  const looseHostWorkingHours = usersWorkingHoursAndBusySlots
+    .filter(({ user }) => schedulingType !== SchedulingType.COLLECTIVE && user?.isFixed !== true)
```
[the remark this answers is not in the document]
emrysal: getAggregateWorkingHours runs ~once per user; possibly more when they segment (e.g. availability like 9:30-10:00 10:30-11:00.. etc) - it's before timeslot looping. (So yes, but not needed)

zomars:
@CarinaWolli @emrysal is disabling instead of filtering a clearer approach? https://www.loom.com/share/a363658bff7c4e269be6a3171404a59c

apps/web/components/NavTabs.tsx
```diff
@@ -67,7 +67,7 @@ const NavTabs: FC<NavTabProps> = ({ tabs, linkProps, ...props }) => {
                   onClick={onClick}
                   className={classNames(
                     isCurrent
-                      ? "border-neutral-900 text-neutral-900"
+                      ? "border-neutral-900 text-gray-900"
```
zomars, line 70: It this from another PR? cc @sean-brydon

emrysal: > @CarinaWolli @emrysal is disabling instead of filtering a clearer approach? https://www.loom.com/share/a363658bff7c4e269be6a3171404a59c

@sean-brydon What do you think?

zomars: Scratch that. It was a merge artifact. But about disabling I think is clearer and less complex to code. Instead of options disappearing.

emrysal: I agree, the options disappearing is really complex really quickly.

zomars (changes requested):
~~Testing locally the pro@example account I'm getting a no users found error, blocking to investigate.~~

image

EDIT: Scratch that. The event was trying to book on a past time.

zomars (approved):
LGTM

PeerRich:
is one review enough for this? or should we continue to monitor and test, maybe @alannnc and @leog too

PR #6439, opened 2023-01-12
Session Timeout

### LINKED ISSUE #6380: [CAL-739] Session Timeout
We need build a feature to allow users to enable session timeout, so it boots them off their session after x amount of time. This is pretty important as it’s one of the last few things blocking HIPAA compliance.

**What constitutes as resetting the time to not session timeout the user? Navigating through pages is enough?**

* Yes

**Could the user choose the amount of time to execute a timeout?**

* Yes.

**Would there be a warning before timing out?**

* No.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-739](https://linear.app/calcom/issue/CAL-739/session-timeout)</sub>

leog:
## What does this PR do?

Implements Session Timeout setting.

https://user-images.githubusercontent.com/467258/212155421-f82320d2-9be2-45da-a449-4bd198dbe4ad.mov

Closes #6380 

## Type of change

## How should this be tested?

Go to Settings > Security > Session and set a session timeout of 5, 10 or 15 minutes. Go back to the app and wait the selected time to see the app logout.


apps/web/package.json
```diff
@@ -73,6 +73,7 @@
     "handlebars": "^4.7.7",
     "ical.js": "^1.4.0",
     "ics": "^2.37.0",
+    "jose": "^4.11.2",
```
leog, line 76: Used to encrypt and decrypt the JWT the way `next-auth` does it.

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -234,13 +243,41 @@ if (true) {
...
+        .encrypt(new Uint8Array(encryptionSecret));
+    },
+    decode: async ({ secret, token }) => {
+      const encryptionSecret = await getDerivedEncryptionKey(secret);
+      const { payload } = await jose.jwtDecrypt(token || "", new Uint8Array(encryptionSecret), {
+        clockTolerance: 15,
+      });
+      return payload;
+    },
+  },
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:280]: This way of encrypting and decrypting the token is the same as `next-auth` does it internally by default when no encode/decode functions are implemented.

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -234,13 +243,41 @@ if (true) {
...
+        sub: token?.sub,
+        name: token?.name,
+        email: token?.email,
+      })
+        .setProtectedHeader({
+          alg: "dir",
+          enc: "A256GCM",
+        })
+        .setIssuedAt()
+        .setExpirationTime(token && token.sessionTimeout ? `${token.sessionTimeout}m` : "30d")
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:269]: Using the `sessionTimeout` value or defaulting to what we had before, 30 days.

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -262,15 +299,20 @@ export default NextAuth({
...
+        if (!existingUserRaw) {
           return token;
         }
 
+        const { metadata, ...existingUser } = existingUserRaw;
+        const parsedMetadata = userMetadata.parse(metadata);
+
         return {
           ...existingUser,
+          sessionTimeout: parsedMetadata?.sessionTimeout,
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:315]: Making sure we have the `sessionTimeout` value to use it when encoding the session token

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -287,6 +329,7 @@ export default NextAuth({
           username: user.username,
           email: user.email,
           role: user.role,
+          sessionTimeout: user?.sessionTimeout,
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:332]: Making sure we have the `sessionTimeout` value to use it when encoding the session token

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -316,13 +359,16 @@ export default NextAuth({
...
+        const metadata = userMetadata.parse(existingUser.metadata);
+
         return {
           ...token,
           id: existingUser.id,
           name: existingUser.name,
           username: existingUser.username,
           email: existingUser.email,
           role: existingUser.role,
+          sessionTimeout: metadata?.sessionTimeout,
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:371]: Making sure we have the `sessionTimeout` value to use it when encoding the session token

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -341,6 +387,7 @@ export default NextAuth({
           name: token.name,
           username: token.username as string,
           role: token.role as UserPermissionRole,
+          sessionTimeout: session.user?.sessionTimeout as number,
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:390]: Making sure we have the `sessionTimeout` value to use it when encoding the session token

apps/web/pages/settings/security/session.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import { getLayout } from "@calcom/features/settings/layouts/SettingsLayout";
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+import { userMetadata } from "@calcom/prisma/zod-utils";
+import { trpc } from "@calcom/trpc/react";
+import { Button, Form, Meta, Select, SettingsToggle, showToast } from "@calcom/ui";
+
+import { ssrInit } from "@server/lib/ssr";
+
+const SessionConfigView = () => {
```
leog, line 13: Session page in Settings

packages/features/kbar/Kbar.tsx
```diff
@@ -208,6 +208,14 @@ export const KBarRoot = ({ children }: { children: React.ReactNode }) => {
...
       perform: () => router.push("/settings/security/impersonation"),
     },
+    {
+      id: "session",
+      name: "session_description",
+      section: "security",
+      shortcut: ["u", "s"],
+      keywords: "session",
+      perform: () => router.push("/settings/security/session"),
+    },
```
leog, line 218: Adding it to the KBar as well

packages/trpc/server/routers/viewer.tsx
```diff
@@ -19,7 +19,6 @@ import { samlTenantProduct } from "@calcom/features/ee/sso/lib/saml";
 import { isPrismaObjOrUndefined, parseRecurringEvent } from "@calcom/lib";
 import getEnabledApps from "@calcom/lib/apps/getEnabledApps";
 import { ErrorCode, verifyPassword } from "@calcom/lib/auth";
-import { IS_SELF_HOSTED, IS_TEAM_BILLING_ENABLED } from "@calcom/lib/constants";
```
leog, line 22: Unused constants

packages/trpc/server/routers/viewer.tsx
```diff
@@ -174,7 +173,7 @@ const loggedInViewerRouter = router({
       weekStart: user.weekStart,
       theme: user.theme,
       hideBranding: user.hideBranding,
-      metadata: user.metadata,
+      metadata: userMetadata.parse(user.metadata),
```
leog, line 176: Making sure we have the metadata available as the parsed JSON

packages/trpc/server/routers/viewer.tsx
```diff
@@ -1130,7 +1131,7 @@ const loggedInViewerRouter = router({
         roomName: z.string(),
       })
     )
-    .query(async ({ ctx, input }) => {
```
leog, line 1133: Unused var

packages/types/next-auth.d.ts
```diff
@@ -13,6 +13,7 @@ declare module "next-auth" {
     id: PrismaUser["id"];
     emailVerified?: PrismaUser["emailVerified"];
     email_verified?: boolean;
+    sessionTimeout?: number;
```
leog, line 16: Added to the `user` as we need it when dealing with the token

PeerRich:
does this really need a new section in settings? we can just add this to `security`

Jaibles:
Could be in with password. I wonder is there a better term we could put that covers both 🤔

apps/web/pages/settings/security/session.tsx
```diff
@@ -0,0 +1,129 @@
...
+          description={t("session_timeout_description")}
+          checked={sessionState !== undefined}
+          data-testid="session-check"
+          onCheckedChange={(e) => {
+            if (!e) {
+              setValue("sessionTimeout", undefined, { shouldDirty: true });
+              setSessionState(undefined);
+            } else {
+              setValue("sessionTimeout", 10, { shouldDirty: true });
+              setSessionState(10);
```
sean-brydon, line 88: Im confused why we need to useState here when its just duplicating what rhf does

apps/web/pages/settings/security/session.tsx
```diff
@@ -0,0 +1,129 @@
...
+            <div className="flex items-center">
+              <p className="text-neutral-900 ltr:mr-2 rtl:ml-2">{t("session_timeout_after")}</p>
+              <Select
+                options={timeoutOptions}
+                defaultValue={timeoutOptions[1]}
+                isSearchable={false}
+                className="block h-[36px] !w-auto min-w-0 flex-none rounded-md text-sm"
+                onChange={(event) => {
+                  setValue("sessionTimeout", event?.value, { shouldDirty: true });
+                  setSessionState(event?.value);
```
sean-brydon, line 103: >  Im confused why we need to useState here when its just duplicating what rhf does

Same here

apps/web/pages/settings/security/session.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { ssrInit } from "@server/lib/ssr";
+
+const SessionConfigView = () => {
+  const { t } = useLocale();
+  const utils = trpc.useContext();
+  const { data: user } = trpc.viewer.me.useQuery();
+  const mutation = trpc.viewer.updateProfile.useMutation({
+    onSuccess: () => {
+      showToast(t("profile_updated_successfully"), "success");
+      reset(getValues());
```
sean-brydon, line 20: We reset here but also dont reset local `sessionState`

apps/web/package.json
zomars: Ill Allow It Spanish GIF

apps/web/pages/api/auth/[...nextauth].tsx
Comment (alannnc) [reply apps/web/pages/api/auth/[...nextauth].tsx:280]: @leog Shouldn't maxAge option for next-auth do the trick here?

Comment (zomars) [reply apps/web/pages/api/auth/[...nextauth].tsx:280]: I was think the same thing @alannnc aren't we kinda over engineering this?

Comment (leog) [reply apps/web/pages/api/auth/[...nextauth].tsx:280]: @alannnc you mean the session.maxAge property in NextAuth config object? Couldn’t find a more straightforward way to configure it grabbing the value from the user metadata. Perhaps I can simplify the logic making a DB query from jwt.encode function given the token.sub as userId or token.email and just setting the expiration from it instead of going the user property approach.

apps/web/pages/settings/security/session.tsx
leog: I needed the `sessionState` to know when to show the dropdown section when it is turned on.

apps/web/pages/settings/security/session.tsx
leog: Same

apps/web/pages/settings/security/session.tsx
leog: On success we know the value for `sessionState` already so it is not set

apps/web/pages/api/auth/[...nextauth].tsx
Comment (alannnc) [reply apps/web/pages/api/auth/[...nextauth].tsx:280]: That's what I mean
> Perhaps I can simplify the logic making a DB query from jwt.encode function given the token.sub as userId or token.email and just setting the expiration from it instead of going the user property approach.

This sounds like a more easy way to do it.

Comment (leog) [reply apps/web/pages/api/auth/[...nextauth].tsx:280]: It would also mean an additional db query each time a token needs to be encoded, which is why I went for the user property approach. It would simplify but could take a toll in performance

zomars (changes requested):
Blocking to prevent merge until I can fully test this.

leog:
> Could be in with password. I wonder is there a better term we could put that covers both 🤔

I completely missed your reply @Jaibles @PeerRich. Would this work?

<kbd>image</kbd>

sean-brydon:
> > Could be in with password. I wonder is there a better term we could put that covers both 🤔
> 
> I completely missed your reply @Jaibles. @PeerRich would this work?
> 
> image

Could we swap the order of these and add a divider? Password reset is likely used by way more people than session toggle. Also the content on password form is larger so would make this page flow better.

Is there anyway we can turn this into one form and goto one Update button instead of the two?

PeerRich:
> > > Could be in with password. I wonder is there a better term we could put that covers both 🤔
> > 
> > 
> > I completely missed your reply @Jaibles. @PeerRich would this work?
> > image
> 
> Could we swap the order of these and add a divider? Password reset is likely used by way more people than session toggle. Also the content on password form is larger so would make this page flow better.
> 
> Is there anyway we can turn this into one form and goto one Update button instead of the two?

good feedback by @sean-brydon 

i dont mind the order, a divider makes sense and definitely only one update button

leog:
> good feedback by @sean-brydon
> 
> i dont mind the order, a divider makes sense and definitely only one update button

Thanks @sean-brydon @PeerRich @Jaibles for your feedback and confirmations, here is the revised UI:

<kbd>image</kbd>

apps/web/pages/api/auth/[...nextauth].tsx
Comment (zomars) [reply apps/web/pages/api/auth/[...nextauth].tsx:280]: We don't mind the extra query as it remains simple and robust. Also we're going to adopt DB sessions and accounts soon enough. So we're going to query anyways.

apps/web/pages/settings/security/password.tsx
```diff
@@ -1,51 +1,102 @@
...
+    value: mins,
+  }));
+
+  console.log({
+    isSubmitting: formMethods.formState.isSubmitting,
+    isDirty: formMethods.formState.isDirty,
+    sessionLoading: sessionMutation.isLoading,
+    passwordLoading: passwordMutation.isLoading,
+  });
+
```
zomars, line 97:
```suggestion
```

packages/features/kbar/Kbar.tsx
zomars, line 218: Not needed if session will coexist with password
```suggestion
```

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -45,6 +45,7 @@ const tabs: VerticalTabItemProps[] = [
       { name: "password", href: "/settings/security/password" },
       { name: "2fa_auth", href: "/settings/security/two-factor-auth" },
       { name: "impersonation", href: "/settings/security/impersonation" },
+      { name: "session", href: "/settings/security/session" },
```
zomars, line 48:
```suggestion
```

Same as above

packages/trpc/server/routers/viewer.tsx
zomars: This could break on prod. Not all users have metadata in the correct shape.

packages/trpc/server/routers/viewer.tsx
```diff
@@ -587,12 +586,14 @@ const loggedInViewerRouter = router({
...
         timeFormat: z.number().optional(),
         disableImpersonation: z.boolean().optional(),
+        metadata: userMetadata.optional(),
       })
     )
     .mutation(async ({ ctx, input }) => {
       const { user, prisma } = ctx;
       const data: Prisma.UserUpdateInput = {
         ...input,
+        metadata: input.metadata as Prisma.InputJsonValue,
```
zomars, line 596: Why not parse instead?

zomars (dismissed):

apps/web/pages/api/auth/[...nextauth].tsx
Comment (leog) [reply apps/web/pages/api/auth/[...nextauth].tsx:283]: Gotcha. Brought back my first implementation then.

PeerRich:
> Thanks @sean-brydon @PeerRich @Jaibles for your feedback and confirmations, here is the revised UI:

perfect! @leog

PeerRich:
> Thanks @sean-brydon @PeerRich @Jaibles for your feedback and confirmations, here is the revised UI:

perfect! @leog

PeerRich (approved):
code looks good, didnt test yet. waiting for @zomars to fully approve

layout is good!

CarinaWolli:
@leog e2e tests are failing and when I try to run it locally it is failing with the same error: 
Screenshot 2023-01-31 at 10 57 31

leog:
> @leog e2e tests are failing and when I try to run it locally it is failing with the same error

Can you test now @CarinaWolli 🙏

apps/web/pages/settings/security/password.tsx
```diff
@@ -68,24 +113,55 @@ const PasswordView = () => {
...
+                }
+              }}
+            />
+            {sessionTimeoutWatch && (
+              <div className="mt-4 text-sm">
+                <div className="flex items-center">
+                  <p className="text-neutral-900 ltr:mr-2 rtl:ml-2">{t("session_timeout_after")}</p>
+                  <Select
+                    options={timeoutOptions}
+                    defaultValue={timeoutOptions[1]}
```
CarinaWolli, line 148: on a new load, the select always shows '10 mins' as the selected value and not the value that is actually selected

leog: Fixed! please try again, and thank you for your feedback 🙏

apps/web/public/static/locales/en/common.json
```diff
@@ -939,6 +941,11 @@
   "current_location": "Current Location",
   "user_phone": "Your phone number",
   "new_location": "New Location",
+  "session": "Session",
+  "session_description": "Control your account session",
+  "session_timeout_after": "Timeout session after",
+  "session_timeout": "Session timeout",
+  "session_timeout_description": "Invalidate your session after certain amount of time.",
```
CarinaWolli, line 948:
```suggestion
  "session_timeout_description": "Invalidate your session after a certain amount of time.",
```

CarinaWolli (approved):
Works great, code looks good to me as well 🙌🏻  Left one small comment

PR #6443, opened 2023-01-12
Check for session in addition for isPublic prop

emrysal:
## What does this PR do?

Fixes woopsy that excluded the KBar on /apps entirely; Fixes by checking for "authenticated" instead of isPublic, returning early when isPublic is false.

### DIFF
--- a/packages/features/shell/Shell.tsx
+++ b/packages/features/shell/Shell.tsx
@@ -172,23 +172,38 @@ const CustomBrandingContainer = () => {
   return <CustomBranding lightVal={user?.brandColor} darkVal={user?.darkBrandColor} />;
 };
 
-export default function Shell(props: LayoutProps) {
-  useRedirectToLoginIfUnauthenticated(props.isPublic);
-  useRedirectToOnboardingIfNeeded();
-  useTheme("light");
-  // don't load KBar when unauthed
-  return props.isPublic ? (
-    <>
-      <CustomBrandingContainer />
-      <Layout {...props} />
-    </>
-  ) : (
+const PublicShell = (props: LayoutProps) => {
+  const { status } = useSession();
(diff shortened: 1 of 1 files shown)


packages/features/shell/Shell.tsx
```diff
@@ -172,23 +172,38 @@ const CustomBrandingContainer = () => {
...
+  // if not a public page, we can assume a session.
+  if (!props.isPublic) {
+    return (
+      <KBarRoot>
+        <CustomBrandingContainer />
+        <Layout {...props} />
+        <KBarContent />
+      </KBarRoot>
+    );
+  }
```
zomars, line 205: Is this check redundant since we do the same in PublicShell?

packages/features/shell/Shell.tsx
```diff
@@ -172,23 +172,38 @@ const CustomBrandingContainer = () => {
...
-  useTheme("light");
-  // don't load KBar when unauthed
-  return props.isPublic ? (
-    <>
-      <CustomBrandingContainer />
-      <Layout {...props} />
-    </>
-  ) : (
+const PublicShell = (props: LayoutProps) => {
+  const { status } = useSession();
```
sean-brydon, line 176: Could we do something like this instead?

Const wrapper = status === "authenticated" ? KBarRoot : Fragment

  <Wrapper>
      <CustomBrandingContainer />
      <Layout {...props} />
    </Wrapper>

sean-brydon: <Wrapper>
      <CustomBrandingContainer />
      <Layout {...props} />
  </Wrapper>

zomars (approved):
Updated, less code. Same result. No rerender.

image

emrysal:
image

PR #6453, opened 2023-01-13
booking page: only scroll to bottom on mobile

### LINKED ISSUE #6452: [CAL-774] Calendar widget on iPads
Short description of the issue:

For some devices on a page with embedded iFrame the timeslots become invisible after being chosen the second time.

**Environment:**
- Chrome/Safari on iPad Pro 11 2022 v16.0 Viewport 834 X 1075 px
- Chrome/Safari on iPad Pro 12.9 2022 v16.0 Viewport 1024 x 1292 px

**Steps to reproduce:**
1. Navigate to a page with embedded inline iframe from iPad (exact devices listed above)
3. Click on one of the dates from the calendar
4. Click on another date

**Actual Result:** When a date is selected  a second time,  the time slots component moves up and **becomes invisible**
**Expected Result**: The timeslots in iFrame should be visible

Recording: 
https://user-images.githubusercontent.com/33749906/212281901-a75f0a70-a39e-494b-85ac-a12ad5e6bf5d.mp4

<sub>[CAL-774](https://linear.app/calcom/issue/CAL-774/calendar-widget-on-ipads)</sub>

PeerRich:
fixes #6452 

previously we scrolled at any view point which causes issues on tablet if the outside container is too big when embedded.

todo:
- [ ] refactor and move useMediaQuery from /web/lib/hooks this into packages/lib/hooks/ and fix imports

### DIFF
--- a/packages/features/calendars/DatePicker.tsx
+++ b/packages/features/calendars/DatePicker.tsx
@@ -5,6 +5,7 @@ import { useEmbedStyles } from "@calcom/embed-core/embed-iframe";
 import classNames from "@calcom/lib/classNames";
 import { daysInMonth, yyyymmdd } from "@calcom/lib/date-fns";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
+import useMediaQuery from "@calcom/lib/hooks/useMediaQuery";
 import { weekdayNames } from "@calcom/lib/weekday";
 import { Button, Icon, SkeletonText } from "@calcom/ui";
 
@@ -135,6 +136,9 @@ const Days = ({
     const date = browsingDate.set("date", day);
     days.push(date);
   }
+
+  const isMobile = useMediaQuery("(max-width: 768px)");
+
   return (
     <>
       {days.map((day, idx) => (
@@ -153,12 +157,13 @@ const Days = ({
               date={day}
               onClick={() => {
                 props.onChange(day);
-                setTimeout(() => {
-                  window.scrollTo({
-                    top: 360,
-                    behavior: "smooth",
-                  });
-                }, 500);
+                isMobile &&
+                  setTimeout(() => {
+                    window.scrollTo({
+                      top: 360,
+                      behavior: "smooth",
+                    });
+                  }, 500);
               }}
               disabled={
                 (includedDates && !includedDates.includes(yyyymmdd(day))) ||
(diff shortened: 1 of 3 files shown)


apps/web/lib/hooks/useMediaQuery.ts
```diff
@@ -1,3 +1,4 @@
+// lets refactor and move this into packages/lib/hooks/
 import { useState, useEffect } from "react";
 
 const useMediaQuery = (query: string) => {
```
PeerRich, line 4: @zomars can you explain me why some lib/hooks are in /apps/web and others in packages/libs/hooks?

What is the decision making process of where the hook should go?

the useMediaQuery hook seems to be reusable in all of our apps so I am surprised to find it here.

I duplicated the code now because I am not sure if we want to refactor or keep it separate

PeerRich: if we want to refactor, lets move this into packages and adjust the imports

packages/ui/components/logo/Logo.tsx
```diff
@@ -2,14 +2,12 @@ import { LOGO_ICON, LOGO } from "@calcom/lib/constants";
...
-        )}
-      </strong>
-    </h1>
+    <strong>
+      {icon ? (
+        <img className="mx-auto w-9" alt="Cal" title="Cal" src={LOGO_ICON} />
+      ) : (
+        <img className={small ? "h-4 w-auto" : "h-5 w-auto"} alt="Cal" title="Cal" src={LOGO} />
+      )}
+    </strong>
```
PeerRich, line 11: unrelated but logo should not be `<h1>`

packages/features/calendars/DatePicker.tsx
```diff
@@ -135,6 +136,9 @@ const Days = ({
     const date = browsingDate.set("date", day);
     days.push(date);
   }
+
+  const isMobile = useMediaQuery("(max-width: 768px)");
+
```
PeerRich, line 141: note: i was initially looking for something that would use our tailwind config but sometimes you want to trigger this on a pixel-perfect level and not based on tailwind `sm` or `md` .. breakpoints so we'll keep this pixel based

hariombalhara (approved):

PR #6454, opened 2023-01-13
Fixing horizontal tabs 

sean-brydon:
Taking over from #6310  

Was still on V2 folder structure and this was a lot easier than migrating that PR 
CleanShot 2023-01-13 at 10 40 41@2x


PeerRich:
thank you for taking over!

hariombalhara (approved):

PR #6457, opened 2023-01-13
Chor/text colors+weights

sean-brydon:
Base all of our styling off html semantics. 

Creates a set of tailwind bases for headings and also talwind components (same as old @apply method) for when we need to override styles applied by bases

Main work for this is in `packages/config/tailwind-bases.js`

Please checkout the storybook instance locally to see the changes.

All h1,h2,h3 have base styles and any additional styling has been removed from them within the app. (I think i have corrected the ones where we have h2s looking like h3s etc 

`p` and `small` tags have been added with a base style but not changed in the app just yet as this PR would be massive.

`.multiline-normal` and `.multiline-medium` have been added in respect to the design system.


apps/web/components/Embed.tsx
```diff
@@ -629,7 +629,7 @@ const ChooseEmbedTypesDialogContent = () => {
   return (
     <DialogContent type="creation" size="lg">
       <div className="mb-4">
-        <h3 className="text-lg font-bold leading-6 text-gray-900" id="modal-title">
+        <h3 className="leading-6" id="modal-title">
```
JeroenReumkens, line 632: Can / should we also include default leadings in our base settings?

packages/config/tailwind-bases.js
```diff
@@ -0,0 +1,49 @@
+const plugin = require("tailwindcss/plugin");
+
+module.exports = plugin(function ({ addBase, theme, addComponents }) {
+  const h1Base = {
+    fontSize: "28px",
+    fontWeight: theme("fontWeight.semibold"),
+  };
+
+  const h2Base = {
+    fontSize: theme("fontSize.2xl"),
```
JeroenReumkens, line 10: Maybe it would be a good idea if we'd have a font size called `h1`, `h2`, etc, because that way we can easily set `font-h1` to override an h2; `<h2 className="font-h1">`. Also: If you make the font definition an array, the second value will be the leading 🔥

sean-brydon:
@Jaibles should we put this on the back burner for now?

Jaibles:
@sean-brydon maybe a good idea. I also don't mind going through it on a call, but basically right now everywhere I see headers I see issues could be easier to put on the backburner.

CleanShot 2023-02-08 at 12 42 05@2x
CleanShot 2023-02-08 at 12 41 49@2x

PeerRich:
CleanShot 2023-02-08 at 16 48 51@2x


CleanShot 2023-02-08 at 16 48 22@2x

sean-brydon:
Gonna put this on hold for a bit - needs some re-consideration

PeerRich:
yeah lets move forward with `dark:` for now

PR #6460, opened 2023-01-13
Fixing GCal emailing notif and sequential deleting

leog:
## What does this PR do?

Making sure Google Calendar does not send deleted events notifications when deleted by API, as we handle the notification for them. Also, Google Calendar delete was causing problems when using its API in parallel, changing it to be sequential.

## Type of change

## How should this be tested?

Try to cancel a recurring event with lots of instances and you shouldn't see Google sending cancelled notification for each instance nor a 500 error when doing so.


PeerRich:
i think some scheduling serivces still send the notification if I am not incorrect

zomars (approved):
LGTM

zomars:
I think is good enough. But is it possible to suppress notifications only in recurring events?

leog:
> I think is good enough. But is it possible to suppress notifications only in recurring events?

@zomars wouldn't we want to also avoid notifications for single events too? Cal already sends a notification for those too

zomars:
> > I think is good enough. But is it possible to suppress notifications only in recurring events?
> 
> @zomars wouldn't we want to also avoid notifications for single events too? Cal already sends a notification for those too

I don't know, hence the question. But merging for now. 🙏🏽

PR #6486, opened 2023-01-15
Don't update the field when changing the select

emrysal:
## What does this PR do?

https://www.loom.com/share/02f86161ebee491ebdef4b5de57904c3

(After UI fixes)
image


emrysal:
@Udit-takkar Added FiTrash to Availability list page here.

sean-brydon (approved):
LGTM - tested and its a much cleaner UX.  The things we gotta do when RHF just doesnt do what you want ;)

PR #6491, opened 2023-01-15
fix tooltip and modal stacking issues

### LINKED ISSUE #6467: [CAL-780] Tooltip is behind Dialog

### Issue Summary

Tooltip is shown behind dialog.
In this case, the upgrade button has a tooltip. You can see a little bit of the tooltip on the right side of the dialog:

Screenshot 2023-01-13 at 13 12 10

<sub>[CAL-780](https://linear.app/calcom/issue/CAL-780/tooltip-is-behind-dialog)</sub>

mocherfaoui:
## What does this PR do?

Fixes #6467 

## Type of change


mocherfaoui:
wouldn't be better to have one overlay and adjust its color and opacity?
https://github.com/calcom/cal.com/blob/7cdd4539535aee5ee0c1d58bf4eae45be7eae30c/packages/ui/components/dialog/Dialog.tsx#L58
https://github.com/calcom/cal.com/blob/7cdd4539535aee5ee0c1d58bf4eae45be7eae30c/packages/ui/components/dialog/Dialog.tsx#L77
since the modal and tooltip use `Portal` there is no need for `z-index`. I think it would be better to remove the first overlay (line 58)

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -165,7 +165,7 @@ const SettingsSidebarContainer = ({
   return (
     <nav
       className={classNames(
-        "no-scrollbar fixed bottom-0 left-0 top-0 z-10 flex max-h-screen w-56 flex-col space-y-1 overflow-x-hidden overflow-y-scroll bg-gray-50 px-2 pb-3 transition-transform lg:sticky lg:flex",
+        "no-scrollbar fixed bottom-0 left-0 top-0 flex max-h-screen w-56 flex-col space-y-1 overflow-x-hidden overflow-y-scroll bg-gray-50 px-2 pb-3 transition-transform max-lg:z-10 lg:sticky lg:flex",
```
PeerRich, line 168: what does max-lg:z-10 do?

PeerRich:
tested and works so far. @Udit-takkar can you test more?

packages/features/settings/layouts/SettingsLayout.tsx
Udit-takkar: +1

packages/ui/components/dialog/Dialog.tsx
```diff
@@ -75,12 +75,11 @@ export const DialogContent = React.forwardRef<HTMLDivElement, DialogContentProps
   ({ children, title, Icon, type = "creation", ...props }, forwardedRef) => {
     return (
       <DialogPrimitive.Portal>
-        <DialogPrimitive.Overlay className="fadeIn fixed inset-0 z-40 bg-gray-500 bg-opacity-75 transition-opacity" />
-        {/*zIndex one less than Toast */}
+        <DialogPrimitive.Overlay className="fadeIn fixed inset-0 bg-gray-500 bg-opacity-75 transition-opacity" />
         <DialogPrimitive.Content
           {...props}
           className={classNames(
-            "fadeIn fixed left-1/2 top-1/2 z-[9998] min-w-[360px] -translate-x-1/2 -translate-y-1/2 rounded bg-white text-left shadow-xl focus-visible:outline-none sm:w-full sm:align-middle",
```
Udit-takkar, line 83: why did you remove this ? did you test with the toast when dialog is open

Udit-takkar (changes requested):
It works but we can just change  the tooltip z -index to 9999 ( one greater than dialog ) instead of making changes everywhere.

PeerRich:
> It works but we can just change the tooltip z -index to 9999 ( one greater than dialog ) instead of making changes everywhere.

i actually like this more., can you open a PR for this so we can compare the preview?

packages/ui/components/dialog/Dialog.tsx
PeerRich: no i didnt

packages/features/settings/layouts/SettingsLayout.tsx
mocherfaoui: to fix a small issue with the sidebar in settings page
before:
Screenshot 2023-01-22 131224
after:
Screenshot 2023-01-22 131420
unless it was intended to appear that way 😅

packages/ui/components/dialog/Dialog.tsx
mocherfaoui: @Udit-takkar you mean if there is a visible toast and then I opened a modal right? 
I didn't test it, thanks for pointing it out

packages/features/settings/layouts/SettingsLayout.tsx
emrysal: Excellent spot 🏆

emrysal (approved):
Looks good to me, excellent work. Toasts stay in front of the dialog as they should. 

Also you are removing z-index values, this would mean that if we had any usage of z-index elsewhere the dialog would now be behind that, but I have checked that we have no usage of z-index where this may interfere with the dialog.

I prefer this approach over https://github.com/calcom/cal.com/pull/6617 as you should always lean towards not using z-index.

mocherfaoui:
@emrysal there is one small problem with my approach. there are two modal overlays, and one of them is why e2e tests are failing. I couldn't figure out why it fails in invite team members modal in /teams/:id/onboard-members
see my comment https://github.com/calcom/cal.com/pull/6491#issuecomment-1385625358

emrysal:
@mocherfaoui That L58 DialogPrimitive.Overlay shouldn't have been there, that's an artefact of the very early days and it just wasn't spotted. I removed it, let's see what the E2E tests do now.

Udit-takkar (approved):
LGTM

PR #6500, opened 2023-01-16
Added fallback to hostsFixed too which is used in COLLECTIVE

emrysal:
## What does this PR do?

See subject.


sean-brydon (approved):
Tested - Seems to work for me

PR #6519, opened 2023-01-17
Move exports up a folder

sean-brydon:
As per discussion in threads. Remove one level of exporting components into UI root. 

Makes life a lot easier when creating a component


PeerRich (approved):
tested preview, seems to work

PeerRich:
waiting for @zomars to confirm this is what we wanna do

zomars (approved):
LGTM. Will check why tests aren't running.

PR #6525, opened 2023-01-17
Icon small button variant

sean-brydon:
Super quick small variant for icons buttons

API has slightly changed for button - a few size props have been moved to variant -> icon and fab 
CleanShot 2023-01-18 at 09 00 04@2x

CleanShot 2023-01-17 at 13 10 46@2x
CleanShot 2023-01-17 at 13 10 42@2x


packages/ui/components/button/Button.tsx
```diff
@@ -46,6 +46,7 @@ const buttonClasses = cva(
         base: "h-9 px-4 py-2.5 ",
         lg: "h-[36px] px-4 py-2.5 ",
         icon: "flex justify-center min-h-[36px] min-w-[36px] ",
+        iconsm: "flex justify-center p-[4px]",
```
hariombalhara, line 49: Can it be icon-sm? That makes it clearer

packages/ui/components/button/Button.tsx
```diff
@@ -221,7 +226,16 @@ export const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, ButtonPr
...
           ) : (
-            <EndIcon className="inline h-5 w-5 ltr:-mr-1 ltr:ml-2 rtl:mr-2" />
+            <EndIcon
+              className={classNames(
+                "inline-flex",
+                size === "icon"
+                  ? "h-4 w-4 "
+                  : size === "iconsm"
+                  ? "h-3 w-3"
+                  : "h-4 w-4 stroke-[1.5px] ltr:mr-2 rtl:ml-2"
```
sean-brydon, line 236: We had incorrect styling on icon end buttons with text

packages/ui/components/button/Button.tsx
sean-brydon: For sure can

packages/ui/components/button/Button.tsx
```diff
@@ -190,7 +191,11 @@ export const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, ButtonPr
             <StartIcon
               className={classNames(
                 "inline-flex",
-                size === "icon" ? "h-4 w-4 " : "h-4 w-4 stroke-[1.5px] ltr:mr-2 rtl:ml-2"
+                size === "icon"
+                  ? "h-4 w-4 "
+                  : size === "iconsm"
```
hariombalhara, line 196: I find it hard to understand normally. Can we abstract it out into variables above this JSX.

sean-brydon: I would say yeah - but we  just loose the benefit of intelisense doing so

PeerRich:
i think eventually we need to refactor icon, size is not really an icon. its more of a variant. variant="icon" size="sm" makes more sense long term

sean-brydon:
> i think eventually we need to refactor icon, size is not really an icon. its more of a variant. variant="icon" size="sm" makes more sense long term

Maybe i just do this here? I don't see it being too much work

PeerRich:
> > i think eventually we need to refactor icon, size is not really an icon. its more of a variant. variant="icon" size="sm" makes more sense long term
> 
> Maybe i just do this here? I don't see it being too much work

sure if you want! good timing

JeroenReumkens (changes requested):
Amazing work @sean-brydon ! 

Could you please also add the icon to the canvas preview in storybook? (screenshot below). That's the only change I request. Apart from that, really happy!! 

CleanShot 2023-01-19 at 08 40 30@2x

JeroenReumkens (approved):

PR #6531, opened 2023-01-17
experiment: disco border 

emrysal:
## What does this PR do?

Prototypes:

image

image

### DIFF
--- a/apps/web/components/booking/pages/AvailabilityPage.tsx
+++ b/apps/web/components/booking/pages/AvailabilityPage.tsx
@@ -167,7 +167,7 @@ const SlotPicker = ({
       <DatePicker
         isLoading={isLoading}
         className={classNames(
-          "relative mt-8 px-4 pb-4 sm:mt-0 md:min-w-[300px] md:px-5 lg:min-w-[455px]",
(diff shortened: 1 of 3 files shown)


PeerRich:
uhh exciting!

PeerRich:
i wonder if we should only show it on loading. I would also slow it down a bit

emrysal:
image

Lightmode isn't offensive either :)

emrysal:
@PeerRich Depends if we want to show it on loading or always, I think on loading is more subtle but have a look; I slowed the animation down 6x.

emrysal:
There's another option to showing it always, if we show it always we should tune it down, like:

image

But my opinion is that we should only animate when something is in progress, not always.

PeerRich (approved):
looks awesom! do we have screenshots for light mode?

Jaibles:
@emrysal let's make sure we keep this as an experiment for now. While I think it's really cool, we can not turn this on for all users by default for a lot of reasons. This needs to be an opt-in feature - how exactly we make it opt-in (whether via themes or something else) we can discuss.

PR #6537, opened 2023-01-18
fix: remove annoying outline when <Button /> clicked

luiisca:
## What does this PR do?

This PR addresses a problem with the Button component where an unwanted outline, which should only be shown in focus, would appear when clicking on any button. So I did a bit of research and I found out focus-visible is a better way to focus elements, as it is only triggered ["via keyboard or some other non-pointing device”](https://hidde.blog/focus-visible-more-than-keyboard/) and not via clicks like focus.

Before: https://www.loom.com/share/f8b5d76024224e27889fd0af680166c1
Now: https://www.loom.com/share/0c1e91626d624943a84df3e26fd7211e

## Type of change


sean-brydon (changes requested):
Nice spot here bud - suprised i've never even noticed this happens on click!

Code looks good but could you go ahead and remove the yarn.lock from this PR? 

`git checkout origin/main -- yarn.lock`

sean-brydon:
Hey @luiisca it looks like yarn.lock was just deleted - if you follow the command i pasted above you will be able to restore it with the version that is on our main branch

PeerRich (approved):

sean-brydon (dismissed):
please run `git checkout origin/main -- yarn.lock` to restore yarn.lock in this PR

PeerRich (approved):

PeerRich:
hey, I am thinking about applying this fix for all focus:outline-none here: https://github.com/calcom/cal.com/pull/6604

emrysal:
Sorted the yarn.lock issue, ready to merge 👍 - thanks for your contribution!

PeerRich:
thank you, @luiisca !! i wonder if we should do this in other places of the app too?

PR #6544, opened 2023-01-18
Create local next.config.js, add storybook-static to .gitignore

emrysal:
## What does this PR do?

Adds a next.config.js which removes the dependency on web/ (and the associated env vars)


sean-brydon (approved):
Ship itttt

PR #6552, opened 2023-01-18
fix: case-sensitivity in apps path

### LINKED ISSUE #6548: [CAL-820] lowercase apps: cal.com/apps/Signal 404 (whereas /signal works)
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-820](https://linear.app/calcom/issue/CAL-820/lowercase-apps-calcomappssignal-404-whereas-signal-works)</sub>

G3root:
## What does this PR do?

Fixes #6548 

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/apps/web/pages/apps/[slug]/index.tsx
+++ b/apps/web/pages/apps/[slug]/index.tsx
@@ -53,7 +53,7 @@ export const getStaticPaths: GetStaticPaths<{ slug: string }> = async () => {
 
   return {
     paths,
-    fallback: false,
+    fallback: "blocking",
   };
 };
 
--- a/apps/web/pages/apps/[slug]/index.tsx
+++ b/apps/web/pages/apps/[slug]/index.tsx
@@ -61,7 +61,7 @@ export const getStaticProps = async (ctx: GetStaticPropsContext) => {
   if (typeof ctx.params?.slug !== "string") return { notFound: true };
 
   const app = await prisma.app.findUnique({
-    where: { slug: ctx.params.slug },
+    where: { slug: ctx.params.slug.toLowerCase() },
   });
 
   if (!app) return { notFound: true };


JeroenReumkens:
These pages are also indexed for SEO, right @PeerRich?  In that case we now have multiple URL's that could end up in Google. So it would be good to add a conical meta tag in that case, which tells google that there's only one single source of truth for every app page.

PeerRich:
> These pages are also indexed for SEO, right @PeerRich? In that case we now have multiple URL's that could end up in Google. So it would be good to add a conical meta tag in that case, which tells google that there's only one single source of truth for every app page.

it actually turns out we have "no-index, no-follow" in our meta tag which means google never indexed them

https://github.com/calcom/cal.com/pull/4943/files#diff-3042eafd6ef787392d368974c4283d202f74a510aa121a4fc8bfc892ceb2d893R312-R313

we should remove this

PeerRich (approved):

apps/web/pages/apps/[slug]/index.tsx
```diff
@@ -53,15 +53,15 @@ export const getStaticPaths: GetStaticPaths<{ slug: string }> = async () => {
 
   return {
     paths,
-    fallback: false,
+    fallback: "blocking",
```
Comment (PeerRich) [inline apps/web/pages/apps/[slug]/index.tsx:56]: what does this do?

Comment (G3root) [reply apps/web/pages/apps/[slug]/index.tsx:56]: all the paths returned by `getStaticPaths` are lower cased. with the fallback false, slug with the title case throws an 404 error. while fallback is blocking, it runs `getStaticProps` and check if it's matches any db record.

emrysal (approved):

Issue #5784, opened 2022-11-29
[CAL-718] Ability to collect First Name and Last Name inputs separately


aasargsian:
We are facing an issue with **Name** input collecting both First and Last name in one field.

In our database, first and last name are stored as separate attributes and used actively in email/SMS communication. And there is no good way to automatically split the full name into first and last names. This is especially challenging in locations where second names are commonly used.

We would like to be able to collect First and Last names as separate inputs.
Ideally, we would like to be able to control all Inputs in a separate section of Event settings.

---

Tagging @alishaz-polymath since we discussed this on a call today :) I, unfortunately, can't add labels to the issue.

---

Sharing a few screenshots from the previous booking solution that we used: 

<img width="264" alt="image" src="https://user-images.githubusercontent.com/33749906/204623440-dbdea85f-5fbf-446c-9070-81af9c869f8a.png">
<img width="695" alt="image" src="https://user-images.githubusercontent.com/33749906/204623538-e3983b1c-5a8f-4fad-ad4c-14c286fbc5fe.png">

<sub>[CAL-718](https://linear.app/calcom/issue/CAL-718/ability-to-collect-first-name-and-last-name-inputs-separately)</sub>

automationsClub:
I second that - It's a really needed feature.
Moreover, It will be great to be able to edit all the input fields and also control their labels.
As my audience isn't speaking English, it's funny to have the name,email,notes questions in English and the custom questions in Hebrew...

PeerRich:
we're rethinking the entire form builder right now and this could be a great option in the form builder. with a form builder you could make your own inputs and then just make an input for "First name" and "Last Name" separately 

cc @Jaibles

automationsClub:
@PeerRich 
That's awesome! 
Any ETA on that?

PeerRich:
not yet

aasargsian:
@PeerRich are there any designs on the form builder or an open issue? I would love to review / comment. This sounds very promising :)

Jaibles:
@aasargsian I need to do the designs first. I'll attach here when I have them.

aasargsian:
@Jaibles sounds great, lookimg forward to it!

Jaibles:
Designs for this are here

[Additional inputs - Cal - New Features (Figma)](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/Cal---New-Features?node-id=1807%3A120640&t=Aiw1Icg3cGdHMn8X-1)

[...]

PR #6560, opened 2023-01-19
Feature/ Manage Booking Questions

### LINKED ISSUE #2440: Add option to remove additional notes box
### Discussed in https://github.com/calcom/cal.com/discussions/924

<sup>Originally posted by **ianhyzy** October 12, 2021</sup>
I prefer to write my own version of the additional notes box, and would like to remove it, or at least have the option to edit it on a per meeting type basis. </div>

### LINKED ISSUE #3350: [CAL-91] For Multi-Location Event Type, required dependent "Additional Inputs"

Assume you can provide a client the choice if he want to be at our office, or at his office, or at video call, or meeting at a business restaurant

In all the cases you have different questions to ask.

* For Video call, no additional location question needed
* For on site meeting, the client needs the address details of our office branch
* For meeting at client location, the client needs to provide mandatory the location to visit
* For meeting at a business restaurant, the address is also mandatory. But the restaurant could be provided by us or by client
* 

in the first step it would be helpful if we could decide which Additional Inputs entry is depending to which selected location. So we could decide what to show in the case the the clients selection

Better would be the option, that we can somehow provide a link with a predefined address/location or the client can select and enter a predefined location. In both cases a geolocation lookup shall be done, to ensure there is no typo. (e.g. google geo coder api)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-91](https://linear.app/calcom/issue/CAL-91/for-multi-location-event-type-required-dependent-additional-inputs)</sub>

### LINKED ISSUE #4915: /event-types: be able to hide additional notes
Also is there a way to remove additional notes?

[Slack Message](https://calendso.slack.com/archives/C01NXDSRY9W/p1665165353348449?thread_ts=1665165353.348449&cid=C01NXDSRY9W)

### LINKED ISSUE #5245: Adding additional input shows wrong "... event type updated successfully" message
### Issue Summary

When adding an additional input field to an event type it shows a toast "... event type updated successfully" when clicking save even though it is not updated. You still need to click save on the top right. 

See here: https://www.loom.com/share/89c5abd6502c4e1098fd9f500c4624da

The pop-up also needs some updates in design. 

Figma Design: 
Screenshot 2022-10-27 at 13 09 18

How it currently looks: 
Screenshot 2022-10-27 at 13 11 23

### LINKED ISSUE #5536: [CAL-376] Move "Additional notes" into Additional inputs
Move this..

Into additional inputs

[4a46ad2f-0f83-4be8-9e04-da58714cac04](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/2d74d509-345c-42e0-a981-72f20cc27bcc/4a46ad2f-0f83-4be8-9e04-da58714cac04)

> From [SyncLinear.com](https://synclinear.com) | [CAL-376](https://linear.app/calcom/issue/CAL-376/move-additional-notes-into-additional-inputs)

### LINKED ISSUE #5564: [CAL-386] Control the order of fields on an event type
VP of Marketing from Daily asked for the ability to control the order of fields on an event type. Especially if they add additional fields such as "company name."

> From [SyncLinear.com](https://synclinear.com) | [CAL-386](https://linear.app/calcom/issue/CAL-386/control-the-order-of-fields-on-an-event-type)

### LINKED ISSUE #5783: Ability to remove Additional notes input
We are facing an issue with Additional notes input always being displayed on the booking page. We already gather a lot of information on the booking stage and yet another field can decrease booking conversions.

We would like to be able to remove Additional notes altogether from the event type inputs.

Ideally, we would like to be able to control all Inputs in a separate section of Event settings and be able to turn on/off any inputs.
---
Related issues: #5536 #4915 #2440
---
Tagging @alishaz-polymath since we discussed this on a call today :) I, unfortunately, can't add labels to the issue.

### LINKED ISSUE #6690: [CAL-906] Error when creating multiple additional inputs
Error when creating additional inputs

Created via Threads. See full discussion: [https://threads.com/34444168171](https://threads.com/34444168171)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-906](https://linear.app/calcom/issue/CAL-906/error-when-creating-multiple-additional-inputs)</sub>

hariombalhara:
## What does this PR do?

Fixes #2440 
Fixes #5536 
Fixes #5783 
Fixes #4915 
Fixes #5564 
Fixes #3350 
Fixes #5245
Fixes #6690
Implements https://github.com/calcom/cal.com/discussions/5840

## Looms
[Demo of sources and how Workflow adds smsReminderNumber now](https://www.loom.com/share/a59edc20ba874482ba86001f3416faf6)
[Introduction to Managing Booking Questions](https://www.loom.com/share/179f1d58940d4d03801c8dffca57f4e3)

## Short Term Follow-up
- #3350
- #5784
- [ ] Reuse Checkbox component cc @CarinaWolli 
- [ ] Support `responses` and `bookingFields` in API using feature detection(i.e. if `responses` are provided ignore `customInputs`). 
- [ ] Error Handling Improvements on Booking Page
	- Show Error Rings on fields
	- Validate on blur
- [x] Booking Success Page - User Fields to match with [Figma](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/Cal---New-Features?node-id=2519%3A135108&t=cLxDDR8ceL8T1DZz-1)
- [ ] `toString` implementation of each field to render responses differently for Email, Calendar, Booking Success Page. Currently, toString simply shows the response as a string. We have the context and we can render the responses with that context, e.g. Phone Number response should be a link that dials the phone number.
- [x] E2E Tests
	- Verify that the chosen location is actually working -> Verify the location shown on response page
	- Add a user field(one of each type) and check it is shown and can be filled on Booking Page and response is shown on booking success page
- [ ] Measure performance improvement of lazy loading FormBuilder on BookingPages( cc @roae Adding it to followup.)

## Long Term Follow-up
- Remove deprecated columns -` booking.customInputs`,` eventType.customInputs`, `eventType.disableGuests`, `eventType.metadata.additionalNotesRequired`
- **Tech Debt**: Make Routing Form use the same FormBuilder with a **variant** if needed.
- While Prefilling - `type`, `date`, `slug`, `count` are reserved query params for existing logic. More might be added. Adding a field with the same `name` would end up prefilling those with invalid information. Maybe restrict adding fields with those names or introduce a namespace for prefill query params?

**Note**: Migration introduces new columns only, so it's safe to be part of this PR itself. Instead of migrating customInputs to new schema for all existing rows, I am transforming from `customInputs` to `bookingFields` on the run, this allows us to avoid corrupting the data during migration.An eventType when saved, copies it's data to `bookingFields`, otherwise there is no modification in existing data. We can later plan to run a migration to do away with `customInputs` 

## Highlights
- Lesser number of Advanced Settings with the same configuration capability
   - **Disable Guests** - It isn't a separate setting now. It is controllable using the Booking Fields management
   - **Require Additional Notes** -  It isn't a separate setting now. It is controllable using the Booking Fields management 	
- **Enhancement**: Modifying Label for a question/field won't change the webhook payload or prefill query param. Earlier when you changed the label for a custom input, the payload received in Webhook and prefill query param name would change as they were generated from Label.
-  **BugFix**: There was a bug earlier that a Custom Input of Radio type wasn't prefilled. It is prefilled correctly now.
- The booking page would not bear the burden of pulling in fields from different places now. e.g. BookingPage was pulling  in data from workflow to determine if a field is to be shown. Now, workflow would push that field to bookingFields directly.

## Backward Compatibility
- Booking links with prefill params even prefill params for **customInputs** would keep on working . This is achieved by transforming `customInputs` into `bookingFields` using **slugified customInput label** as the **name** for the `bookingField`
- Any existing webhook endpoint would keep on receiving `customInputs` in the exact same format. This is achieved by transforming `booking.responses` into `customInputs` when a booking is done.
- Duplicating an existing EventType would still duplicate `customInputs`. This would allow the situation where an EventType which didn't have its `customInputs` saved into `bookingFields` would keep on working.
- A booking that was created before the EventType's `customInputs` were mapped to `bookingFields` would still be able to show responses by a mapping of `booking.customInputs` to `booking.responses`
- An EventType which is connected to an Attendee SMS workflow would add the `smsReminderNumber` field on the fly. It allows that EventType to still collect `smsReminderNumber` data without modifying the corresponding table row.

## Type of change

## Checklist


markflorkowski:
Per @PeerRich's suggestion, asking here --

are user responses to booking questions available in metadata for use in custom webhooks?

apps/web/pages/booking/[uid].tsx
```diff
@@ -574,15 +573,40 @@ export default function Success(props: SuccessProps) {
...
+                      if (
+                        !field ||
+                        field.name === "location" ||
+                        field.name === "guests" ||
+                        field.name === "notes" ||
+                        field.name === "name" ||
+                        field.name === "email"
+                      ) {
+                        return null;
+                      }
```
Comment (zomars) [inline apps/web/pages/booking/[uid].tsx:599]: https://handbook.cal.com/engineering/best-practices/avoid-comparing-multiple-values-with-includes

```suggestion
                      if (!field || ["location", "guests", "notes", "name", "email"].includes(field.name)) {
                        return null;
                      }
```

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -176,48 +177,59 @@ const EventTypePage = (props: EventTypeSetupProps) => {
...
     ),
   });
 
+  useEffect(() => {
+    if (!formMethods.formState.isDirty) {
+      //TODO: What's the best way to sync the form with backend
+      formMethods.setValue("bookingFields", defaultValues.bookingFields);
+    }
+  }, [defaultValues]);
+
```
Comment (zomars) [inline apps/web/pages/event-types/[type]/index.tsx:232]: Don't render the form until defaultValues is not undefined.

```suggestion
```

[Try the React Container approach](https://reactpatterns.js.org/docs/container-component)

packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        if (
+          bookingField.type === "address" ||
+          bookingField.type === "text" ||
+          bookingField.type === "select" ||
+          bookingField.type === "name" ||
+          bookingField.type === "number" ||
+          bookingField.type === "radio" ||
+          bookingField.type === "textarea"
+        ) {
```
zomars, line 143: https://handbook.cal.com/engineering/best-practices/avoid-comparing-multiple-values-with-includes

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -305,43 +312,145 @@ async function ensureAvailableUsers(
...
 
-async function handler(req: NextApiRequest & { userId?: number | undefined }) {
+function getBookingData({
+  req,
+  isNotAnApiCall,
+  eventType,
+}: {
+  req: NextApiRequest;
+  isNotAnApiCall: boolean;
+  eventType: Awaited<ReturnType<typeof getEventTypesFromDB>>;
```
zomars, line 322:
```suggestion
    args: { isNotAnApiCall: boolean;  eventType: Awaited<ReturnType<typeof getEventTypesFromDB>>; }
```

packages/lib/hooks/useHasPaidPlan.ts
```diff
@@ -13,7 +13,7 @@ export function useHasPaidPlan() {
 
   const hasPaidPlan = hasTeamPlan?.hasTeamPlan || isCurrentUsernamePremium;
 
-  return { isLoading, hasPaidPlan };
+  return { isLoading, hasPaidPlan: hasPaidPlan };
```
zomars, line 16:
```suggestion
  return { isLoading, hasPaidPlan };
```

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -338,6 +339,8 @@ const PlainForm = <T extends FieldValues>(props: FormProps<T>, ref: Ref<HTMLForm
           form
             .handleSubmit(handleSubmit)(event)
             .catch((err) => {
+              console.log(err);
```
zomars, line 342:
```suggestion
```

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -338,6 +339,8 @@ const PlainForm = <T extends FieldValues>(props: FormProps<T>, ref: Ref<HTMLForm
...
+              // FIXME: Booking Pages don't have toast, so this error is never shown
```
zomars, line 343: Can we dynamic load them? I don't want to increase booking pages bundle size any more.

packages/ui/form/PhoneInput.tsx
```diff
@@ -1,29 +1,21 @@
-import BasePhoneInput, { Props } from "react-phone-number-input/react-hook-form";
+import BasePhoneInput, { Props } from "react-phone-number-input";
```
zomars, line 1: Can we lazy load this component?

zomars (changes requested):
So far so good. Got some feedback and need to test more as well.

hariombalhara:
@markflorkowski Yes, they are available right now as `customInputs` and would be available after this PR in `responses`. `customInputs` would still remain there for backward compatibility.

hariombalhara:
@CarinaWolli The approach to ask for smsReminderNumber has changed in this PR, requesting your review on that.

packages/ui/form/PhoneInput.tsx
hariombalhara: Yup should be doable. Raised a separate issue.

packages/app-store/ee/routing-forms/components/react-awesome-query-builder/widgets.tsx
```diff
@@ -25,49 +79,70 @@ const TextAreaWidget = (props: TextWidgetProps) => {
...
+        className="dark:placeholder:text-darkgray-600 focus:border-brand dark:border-darkgray-300 dark:text-darkgray-900 block w-full rounded-md border-gray-300 text-sm focus:ring-black disabled:bg-gray-200 disabled:hover:cursor-not-allowed dark:bg-transparent dark:selection:bg-green-500 disabled:dark:text-gray-500"
+        value={textValue}
+        placeholder={placeholder}
+        disabled={readOnly}
+        onChange={onChange}
+        {...remainingProps}
+        {...customProps}
+      />
+    );
+  }
```
sean-brydon, line 124: we could simply this by just adding 

`labelSrOnly={noLabel}` to the text field below. This will visually disable the label for the TextField but also still provide the input name to the htmlFor so we still support the a11y requirements here 

(If im not understanding whats happening here let me know :D )

packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -0,0 +1,245 @@
...
+type Fields = z.infer<typeof eventTypeBookingFields>;
+
+const EventTypeCustomInputType = {
+  TEXT: "TEXT",
+  TEXTLONG: "TEXTLONG",
+  NUMBER: "NUMBER",
+  BOOL: "BOOL",
+  RADIO: "RADIO",
+  PHONE: "PHONE",
+};
```
sean-brydon, line 21: I believe this would provide better TS support when using this 
```suggestion
const EventTypeCustomInputType = {
  TEXT: "TEXT",
  TEXTLONG: "TEXTLONG",
  NUMBER: "NUMBER",
  BOOL: "BOOL",
  RADIO: "RADIO",
  PHONE: "PHONE",
} as const;
```

packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -0,0 +1,245 @@
...
+    },
+  ];
+
+  const missingSystemBeforeFields = [];
+  for (const field of systemBeforeFields) {
+    // Only do a push, we must not update existing system fields as user could have modified any property in it,
+    if (!bookingFields.find((f) => f.name === field.name)) {
+      missingSystemBeforeFields.push(field);
+    }
+  }
```
sean-brydon, line 194: NIT: I believe this could be a map (would save milliseconds)  / look a little cleaner

packages/features/form-builder/Components.tsx
```diff
@@ -0,0 +1,379 @@
...
+        }
+      >(
+        props: TProps
+      ) => JSX.Element;
+    };
+
+// TODO: Share FormBuilder components across react-query-awesome-builder(for Routing Forms) widgets.
+// There are certain differences b/w two. Routing Forms expect label to be provided by the widget itself and FormBuilder adds label itself and expect no label to be added by component.
+// Routing Form approach is better as it provides more flexibility to show the label in complex components. But that can't be done right now because labels are missing consistent asterisk required support across different components
+export const Components: Record<BookingFieldType, Component> = {
```
sean-brydon, line 59: This is awesome @hariombalhara 🔥

packages/ui/components/form/inputs/Label.tsx
```diff
@@ -4,7 +4,11 @@ export function Label(props: JSX.IntrinsicElements["label"]) {
   return (
     <label
       {...props}
-      className={classNames("mb-2 block text-sm font-medium leading-none text-gray-700", props.className)}>
+      // TODO: Test labels inside the app when dark theme is enabled as app doesn't support dark theme
+      className={classNames(
+        "mb-2 block text-sm font-medium leading-none text-gray-700 dark:text-white",
+        props.className
+      )}>
```
sean-brydon, line 11: Youre all good here we have a constant

`isThemeSupported` which is only enabled on booking pages - else we only allow light mode :)

hariombalhara: Thanks @sean. I tested as well. Removing the comment

CarinaWolli (dismissed):
Tested everything around SMS Reminder Number and I found the following issues: 
- Translation is missing for "SMS Reminder Number" (translation key is "number_sms_notifications")
- I think the input field used to be editable when rescheduling. I am also open to have it disabled, but then it should be greyed out to make it clear. 
- When the location of the event type is "Attendee Phone Number" the phone number input is not shown (has nothing to do with SMS Reminder Number)

packages/app-store/ee/routing-forms/components/react-awesome-query-builder/widgets.tsx
hariombalhara: @sean-brydon This is exactly what I need. Completely missed it. Thanks 🙏

packages/features/bookings/lib/getBookingFields.ts
sean-brydon: Makes sense :)

hariombalhara:
Thanks a lot @CarinaWolli, I fixed the issues you mentioned. Found that the placeholder was also not configured and fixed that too for SMS reminder.

@CarinaWolli Attendee Phone Input [seems to be working for me.]( https://www.loom.com/share/ee064d58a42943f3ac087c7158e79622) Can you share a loom of the issue if you can see it still.

Also, while testing it again, I found another case that I thought I had handled already(but I guess it didn't because I changed my approach midway). Earlier, phone number input was asked only once no matter it was for location(Attendee Phone input) or for `smsReminder`. Now, 2 separate inputs are being asked for.  I will see if I can handle it in the Form Builder.

hariombalhara (commented):
~~Blocker: Discovered a missing backward compatibility case. smsReminderNumber is not adding a new field for existing event types~~
Edit: Resolved

CarinaWolli:
@hariombalhara I tried it again and it shows the location 'Attendee Phone Number' correctly, but when booking it doesn't ask me for my phone number and the location shows a cal video link. Here is a loom, let me know if you need more details 🤔

https://www.loom.com/share/c406ce98029d4669b2672617a9530b53

hariombalhara:
Oh got it. That's an edge case I didn't think of when there is just attendee phone no location and nothing else. 
Thankyou 🙏🙏🙏

packages/features/bookings/lib/handleNewBooking.ts
zomars: I just don't want to pollute the function with extra arguments. I see these as ` args` or `options`

hariombalhara:
@CarinaWolli Fixed the Attendee Phone issue along with backward compatibility for `smsReminderNumber` field. All good now.

hariombalhara:
> @hariombalhara I tried it again and it shows the location 'Attendee Phone Number' correctly, but when booking it doesn't ask me for my phone number and the location shows a cal video link. Here is a loom, let me know if you need more details 🤔
> 
> [loom.com/share/c406ce98029d4669b2672617a9530b53](https://www.loom.com/share/c406ce98029d4669b2672617a9530b53)


@CarinaWolli Fixed the Attendee Phone issue along with backward compatibility for smsReminderNumber field. All good now.

hariombalhara (commented):
Self review done.

cc @zomars. All comments resolved as well. I would still want more people to test it out, it being the very critical part.

I am still doing more testing.

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,734 @@
...
+                        onChange(val);
+                      }}
+                      label="Required"
+                    />
+                  );
+                }}
+              />
+              <DialogFooter>
+                <DialogClose color="secondary">Cancel</DialogClose>
+                <Button type="submit">{isFieldEditMode ? t("save") : t("add")}</Button>
```
CarinaWolli, line 466: I can't edit fields. When clicking save this error is thrown:
Screenshot 2023-02-15 at 10 04 40

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,734 @@
...
+
+              {fieldType?.needsOptions ? (
+                <Controller
+                  name="options"
+                  render={({ field: { value, onChange } }) => {
+                    return <OptionsField onChange={onChange} value={value} className="mt-6" />;
+                  }}
+                />
+              ) : null}
+              <Controller
```
CarinaWolli, line 448: When editing a field I can't click 'yes' or 'no', it's working when creating a new field

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,734 @@
...
+                    </Button>
+                    {field.editable !== "system" && field.editable !== "system-but-optional" && (
+                      <Button
+                        className="!p-0"
+                        color="minimal"
+                        onClick={() => {
+                          removeField(index);
+                        }}
+                        StartIcon={FiTrash2}
+                      />
```
CarinaWolli, line 344:
```suggestion
                      <Button
                        color="minimal"
                        variant="icon"
                        onClick={() => {
                          removeField(index);
                        }}
                        StartIcon={FiTrash2}
                      />
```

packages/features/form-builder/FormBuilder.tsx
hariombalhara: @CarinaWolli I think you must be editing a system field, that doesn't allow marking itself optional. User fields can be marked yes or not in edit mode.

packages/features/form-builder/FormBuilder.tsx
hariombalhara: Ahh..regression due to recent name conflict check. Fixed it.

packages/features/form-builder/FormBuilder.tsx
CarinaWolli: I see, it works with the fields that I created. We should make that more clear that this is disabled and can't be edited

CarinaWolli:
I am getting this error when having a Select field, I also have options defined

Screenshot 2023-02-15 at 10 41 09

packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -0,0 +1,333 @@
...
+          id: "default",
+          type: "default",
+        },
+      ],
+    },
+    {
+      defaultLabel: "location",
+      type: "radioInput",
+      name: "location",
+      // Even though it should be required it is optional in production with backend choosing CalVideo as the fallback
```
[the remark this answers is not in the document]
hariombalhara: @CarinaWolli Yes, that should happen already. It is handled here https://github.com/calcom/cal.com/pull/6560/files#diff-d360af68e64bd72af3fbe53bea6bdfbf7a70e0e1366670d6adfd972c1059f88bR151

packages/features/form-builder/FormBuilder.tsx
hariombalhara: Yeah, I am not sure how to do it. But it is a valid point. @Jaibles I think a better disabled state for yes No button should help I guess.

packages/features/bookings/lib/getBookingFields.ts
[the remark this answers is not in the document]
CarinaWolli: I meant in the event-type settings. So this one: 
Screenshot 2023-02-15 at 10 46 39

But it might be ok to still have it here, was just a thought

hariombalhara:
> I am getting this error when having a Select field, I also have options defined
> 
> Screenshot 2023-02-15 at 10 41 09

@CarinaWolli Fixed this one as well 🙏 It was only when user didn't make any modification to options

packages/features/form-builder/FormBuilder.tsx
hariombalhara: Agreed, I too saw the problem. I was thinking that all buttons should be there but they should be disabled.

CarinaWolli: Yes I think I like that better 👍 I just starts looking pretty messy once you change the order and you have system fields and your fields mixed up

packages/features/bookings/lib/getBookingFields.ts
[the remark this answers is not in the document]
hariombalhara: Oh, so it should remain I think. The plan is to move location select to the right sidebar where we show locations anyway. There we always show the location if there is just one

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -305,43 +312,151 @@ async function ensureAvailableUsers(
...
     allRecurringDates,
     currentRecurringIndex,
     noEmail,
-    eventTypeSlug,
     eventTypeId,
+    eventTypeSlug,
     hasHashedBookingLink,
     language,
     appsStatus: reqAppsStatus,
-    ...reqBody
```
zomars, line 321: Isn't this the request body? Why the rename?

CarinaWolli:
Another small thing I found: The input fields for 'Attendee Address' and 'Attendee phone number' are required, however it is not maked with *
Screenshot 2023-02-15 at 15 50 26

Other than that everything is working for me now 👍

zomars:
Let's make it clearer that name CANNOT be optional:

image

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -77,22 +56,112 @@ import { HashLinkPageProps } from "../../../pages/d/[link]/book";
...
+  return (
+    // TODO: It might make sense to extract this logic into BookingFields config, that would allow to quickly configure system fields and their editability in fresh booking and reschedule booking view
+    <div>
+      {fields.map((field, index) => {
+        // During reschedule by default all system fields are readOnly. Make them editable on case by case basis.
+        // Allowing a system field to be edited might require sending emails to attendees, so we need to be careful
+        let readOnly =
+          (field.editable === "system" || field.editable === "system-but-optional") && !!rescheduleUid;
+        let noLabel = false;
+        let hidden = false;
```
zomars, line 83: Why not?

```suggestion
        let hidden = !!field.hidden;
```

zomars (changes requested):
Still got more tests to do but will submit my feedback so far 🙏🏽

apps/web/components/booking/pages/BookingPage.tsx
zomars: BTW this solved this https://github.com/calcom/cal.com/pull/6560#issuecomment-1432026165

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,737 @@
...
+      return typeof val === "string";
+    }
+    if (propsType === "multiselect") {
+      return val instanceof Array && val.every((v) => typeof v === "string");
+    }
+    if (propsType === "objectiveWithInput") {
+      return typeof value === "object" && value !== null ? "value" in value : false;
+    }
+    throw new Error(`Unknown propsType ${propsType}`);
+  };
```
zomars, line 564:
```suggestion
  const isValueOfPropsType = (val: unknown, propsType: typeof componentConfig.propsType) => {
    const propsTypeConditionMap = {
      boolean: typeof val === "boolean",
      multiselect: val instanceof Array && val.every((v) => typeof v === "string"),
      objectiveWithInput: typeof value === "object" && value !== null ? "value" in value : false,
      select: typeof val === "string",
      text: typeof val === "string",
      textList: val instanceof Array && val.every((v) => typeof v === "string"),
    } as const;
    if (!propsTypeConditionMap[propsType]) throw new Error(`Unknown propsType ${propsType}`);
    return propsTypeConditionMap[propsType];
  };
```

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,737 @@
...
+          value={value as { value: string; optionValue: string }}
+          setValue={setValue as (arg: typeof value) => void}
+          optionsInputs={field.optionsInputs}
+          options={field.options}
+        />
+      </WithLabel>
+    ) : null;
+  }
+
+  throw new Error(`Field ${field.name} does not have a valid propsType`);
```
zomars, line 674: I'm pretty sure we can improve a lot in terms of legibility and typing here.

packages/prisma/zod-utils.ts
```diff
@@ -119,23 +141,26 @@ export const stringOrNumber = z.union([
 
 export const stringToDayjs = z.string().transform((val) => dayjs(val));
 
-export const bookingCreateBodySchema = z.object({
+export const bookingCreateSchemaForApiOnly = z.object({
```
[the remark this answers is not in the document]
zomars: Let's keep `bookingCreateBodySchema` intact and add a new one so we don't break API until it is update.

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,737 @@
...
+      value: "radio",
+      needsOptions: true,
+      isTextType: false,
+    },
+    {
+      label: "Checkbox",
+      value: "boolean",
+      isTextType: false,
+    },
+  ];
```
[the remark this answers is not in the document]
hariombalhara: I am not sure if we can control the order of these types reliably now(across browsers).

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -77,22 +56,112 @@ import { HashLinkPageProps } from "../../../pages/d/[link]/book";
 import { TeamBookingPageProps } from "../../../pages/team/[slug]/book";
 
 type BookingPageProps = BookPageProps | TeamBookingPageProps | HashLinkPageProps;
+const BookingFields = ({
+  fields,
+  locations,
+  rescheduleUid,
+  isDynamicGroupBooking,
+}: {
+  fields: BookingPageProps["eventType"]["bookingFields"];
```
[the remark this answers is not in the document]
hariombalhara: I have a counter reasoning here. I am just passing on a prop created by getSSRProps here. If I end up generating a different variation of FieldsType in getSSRProps, the typecheck will pass, but it might crash at runtime.

apps/web/components/booking/pages/BookingPage.tsx
hariombalhara: Yeah it should be that way only 🙏 It's a recent regression while fixing another bug

packages/features/form-builder/FormBuilder.tsx
hariombalhara: Yeah I think so too. But couldn't get type narrowing to work and had to resort to type assertion. Will followup on it.

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara: Technically it is not reqBody now.
For an API request the body has `name` but for request from frontend the body has `responses.name` and this detail is normalized using `getBookingData` method so that `bookingData` has name in both the cases.

hariombalhara:
> Another small thing I found: The input fields for 'Attendee Address' and 'Attendee phone number' are required, however it is not maked with * Screenshot 2023-02-15 at 15 50 26
> 
> Other than that everything is working for me now 👍

Yeah that's known to me. I need to do some non trivial changes to get that working because that's not the field label you are seeing, that is actually the option label being shown as field label(for radioInput having just one option)

packages/features/form-builder/FormBuilder.tsx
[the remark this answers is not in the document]
hariombalhara: Okay, I just read that the order is reliable since many years now. So, if property names are strings, they would be iterated in the order they are created. So this should be fine now.

hariombalhara:
> Let's make it clearer that name CANNOT be optional:

As discussed [here ](https://threads.com/34446204540?s=gno6fDuRTrF2SA1UmtqREK) we should be showing an asterisk consistently across the labels for all fields. That would solve the problem. I would pick it up as a followup

apps/web/pages/booking/[uid].tsx
```diff
@@ -537,63 +533,39 @@ export default function Success(props: SuccessProps) {
...
+                            SystemField.Enum.name,
+                            SystemField.Enum.email,
+                            SystemField.Enum.guests,
+                            SystemField.Enum.notes,
+                            SystemField.Enum.rescheduleReason,
+                          ] as string[]
+                        ).includes(field.name)
+                      ) {
+                        return null;
+                      }
```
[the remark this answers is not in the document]
Comment (hariombalhara) [reply apps/web/pages/booking/[uid].tsx:557]: Wow. That safeParse approach is much better. I unnecessarily made it complex. 
Thanks @zomars 🙏

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -305,43 +312,151 @@ async function ensureAvailableUsers(
...
-
+  const dynamicUserList = Array.isArray(bookingData.user)
+    ? getGroupName(bookingData.user)
+    : getUsernameList(bookingData.user);
   if (!eventType) throw new HttpError({ statusCode: 404, message: "eventType.notFound" });
 
   const paymentAppData = getPaymentAppData(eventType);
 
-  // Check if required custom inputs exist
-  handleCustomInputs(eventType.customInputs as EventTypeCustomInput[], reqBody.customInputs);
```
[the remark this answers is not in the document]
hariombalhara: It's needed for API request and moved to `getBookingData` fn.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1246,7 +1359,7 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
   try {
     await scheduleWorkflowReminders(
       eventType.workflows,
-      reqBody.smsReminderNumber as string | null,
+      smsReminderNumber as string | null,
```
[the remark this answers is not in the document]
hariombalhara: Turns out smsReminderNumber is not expected to be undefined by `scheduleWorkflowReminders` but it can be undefined. Handled that and removed this type assertion

packages/prisma/zod-utils.ts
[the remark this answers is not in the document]
hariombalhara: This is not breaking API. It has the old props for backward compatibility only. This is merged with `extendedBookingCreateBody` to get final schema. Maybe the name is a bit misleading, I will change that.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -712,12 +832,6 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
     return booking;
   }
 
-  if (reqBody.customInputs.length > 0) {
```
hariombalhara, line 715: Moved to `getCustomInputsResponses` which handles backward compatibility as well.

hariombalhara:
Found another issue while testing **and it's fixed now.**
Steps:
1. You add no user fields and let the booker do a booking
2. Reach the booking/[uid] (success) page.
3. Add a user input and mark it required.
4. Refresh the already open booking/[uid] page. It would crash as it would expect that user field's response to be present but it will not be there because it's a booking that happened before the user field was added.

To fix this, I have created a partial version of bookingResponsesSchema. The non partial version should be used only when fresh booking is being created, otherwise we should use partial schema only.

cc @zomars @CarinaWolli for your review

apps/web/components/booking/pages/BookingPage.tsx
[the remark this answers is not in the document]
zomars: That's a problem of type signature differing from the runtime signature. It doesn't matter where this comes from as long as signature matches then there shouldn't be any problems.

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -0,0 +1,736 @@
...
+        render={({ field: { value, onChange } }) => {
+          return (
+            <div>
+              <ComponentForField
+                field={field}
+                value={value}
+                readOnly={readOnly}
+                setValue={(val: unknown) => {
+                  onChange(val);
+                }}
```
zomars, line 704: Why not let ComponentForField accept this signature?

```suggestion
                onChange={onChange}
```

apps/web/pages/team/[slug]/book.tsx
```diff
@@ -2,9 +2,10 @@ import type { GetServerSidePropsContext } from "next";
 
 import type { LocationObject } from "@calcom/app-store/locations";
 import { privacyFilteredLocations } from "@calcom/app-store/locations";
+import { getBookingFieldsWithSystemFields } from "@calcom/features/bookings/lib/getBookingFields";
```
Comment (roae) [inline apps/web/pages/team/[slug]/book.tsx:5]: In this page we can also load the dependencies that are only needed in getServerSideProps within the function, we must take care of the bundle size of the public pages.

packages/features/form-builder/Components.tsx
```diff
@@ -0,0 +1,385 @@
...
+  checkbox: {
+    propsType: "multiselect",
+    factory: ({ options, readOnly, setValue, value }) => {
+      value = value || [];
+      return (
+        <div>
+          {options.map((option, i) => {
+            return (
+              <label key={i} className="block">
+                <input
```
CarinaWolli, line 236: can we use our `<Checkbox/>` component here? Our checkboxes should have rounded corners

roae (dismissed):
There are some opportunities to decrease the bundle size of some public pages, and I think after that we can work on reducing the amount of tailwind alvitriary values.

apps/web/pages/team/[slug]/[type].tsx
```diff
@@ -2,6 +2,7 @@ import type { GetServerSidePropsContext } from "next";
 
 import type { LocationObject } from "@calcom/core/location";
 import { privacyFilteredLocations } from "@calcom/core/location";
+import { getBookingFieldsWithSystemFields } from "@calcom/features/bookings/lib/getBookingFields";
```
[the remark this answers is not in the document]
Comment (hariombalhara) [reply apps/web/pages/team/[slug]/[type].tsx:5]: I had thought that Next.js should automatically remove this through treeshaking. Because it knows it's used in getServerSideProps only.

Found an article on it https://gedasfx.medium.com/how-does-next-js-10-handle-imports-c51b9e4451c6#:~:text=All%20imports%2C%20which%20are%20then%20used%20ONLY,will%20be%20sent%20over%20to%20the%20client.

Also, I spent some time verifying that it is being removed automatically, but because of `getBookingFieldsWithSystemFields` being imported elsewhere(some indirect dependency), it's still being imported.

I will do followup work on some approach to remove it. 
I also checked Next.js bundle analyzer stats and the payload isn't changing much from what we have already.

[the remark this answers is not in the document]
Comment (hariombalhara) [reply apps/web/pages/team/[slug]/[type].tsx:5]: Btw, while testing it out I found something else that we fix too, filed an issue for that [here](https://github.com/calcom/cal.com/issues/7276)

packages/features/form-builder/Components.tsx
hariombalhara: Tried using <Checkbox> and it mandates a description field right now, which this component doesn't need because it is going to show a list of checkboxes. Need to think about making changes in the component. 
I am adding it to followup list in the PR and will pick it up. Don't want to do changes to more files in this PR, as it's already huge.

CarinaWolli: Sounds good, maybe you can make rounded corners to the checkboxes for now. That's why I saw that we don't reuse the component here :)

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -648,392 +574,23 @@ const BookingPage = ({
...
-                    onClick={() => {
-                      guestsField.append({ email: "" });
-                    }}
-                    className="mr-auto">
-                    {t("additional_guests")}
-                  </Button>
-                )}
-
-                <div className="flex justify-end space-x-2 rtl:space-x-reverse">
+              <Form form={bookingForm} noValidate handleSubmit={bookEvent}>
```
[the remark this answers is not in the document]
roae: Ok, you are right, loading it dynamically might slow down the LCP.

apps/web/pages/team/[slug]/[type].tsx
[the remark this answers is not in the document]
Comment (roae) [reply apps/web/pages/team/[slug]/[type].tsx:5]: Yes, I had already read that post, but in booking page it reduces -100kb when I tried async imports, maybe the reduction was for another reason.

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -1,34 +1,35 @@
 import { zodResolver } from "@hookform/resolvers/zod";
```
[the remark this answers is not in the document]
JeroenReumkens: This will be fun solving conflicts with the new booker component  😅 That new component merges the booker + booking form into one page. 

Not much we you can do for now, but it's good to be aware.

See this draft PR from this file down for more context: https://github.com/calcom/cal.com/pull/6792/files#diff-3314f3429ad01dbd0242361878990240275a062a8e0d564b2ec04a856ffae24b

packages/features/form-builder/Components.tsx
hariombalhara: Made it rounded.

zomars:
Shaved 16kb by lazy loading the payment data (not all event types use this)

image

image

zomars (approved):
I think we're on good position to merge. Let's test this troughly on staging.

apps/web/components/booking/BookingDescriptionPayment.tsx
```diff
@@ -5,8 +5,7 @@ import { FiCreditCard } from "@calcom/ui/components/icon";
 
 const BookingDescriptionPayment = (props: { eventType: Parameters<typeof getPaymentAppData>[0] }) => {
   const paymentAppData = getPaymentAppData(props.eventType);
-
-  if (!paymentAppData || paymentAppData.price > 0) return null;
+  if (!paymentAppData || paymentAppData.price <= 0) return null;
```
hariombalhara, line 8: @zomars FYI, have changed this check as it was rendering only when payment isn't required..

CarinaWolli:
The number field has 2 labels
Screenshot 2023-03-02 at 09 48 23

hariombalhara:
> The number field has 2 labels Screenshot 2023-03-02 at 09 48 23

What?? Why ?? :(. Let me fix it.

CarinaWolli (approved):
Amazing work @hariombalhara 🙌🏻

PR #6569, opened 2023-01-19
component: <UpdateScreen>

### LINKED ISSUE #6549: [CAL-821] create generic <UpdateScreen> for upgrading to TEAMS plan
* /teams
* /routing-forms
* /insights

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-821](https://linear.app/calcom/issue/CAL-821/create-generic-updatescreen-for-upgrading-to-teams-plan)</sub>

PeerRich:
the idea is to have a generic `<UpgradeScreen>` component, similar to `<EmptyScreen>` that is being used whenever a TEAM plan is needed. 

RFC: please comment on the underlying API before I move forward 

CleanShot 2023-01-19 at 14 58 36@2x

### DIFF
--- a/packages/features/ee/teams/components/TeamsListing.tsx
+++ b/packages/features/ee/teams/components/TeamsListing.tsx
@@ -22,7 +22,6 @@ export function TeamsListing() {
 
   const teams = data?.filter((m) => m.accepted) || [];
   const invites = data?.filter((m) => !m.accepted) || [];
-
   const features = [
     {
       icon: <Icon.FiUsers className="h-5 w-5 text-gray-700" />,
@@ -65,7 +64,65 @@ export function TeamsListing() {
           <TeamList teams={invites} />
         </div>
       )}
+
       {isLoading && <SkeletonLoaderTeamList />}
+
+      <UpgradeScreen
+        title="calcom_is_better_with_team"
+        description="add_your_team_members"
+        features={features}
+        background="/team-banner-background.jpg"
+        buttons={
+          <div className="space-y-2 rtl:space-x-reverse sm:space-x-2">
+            <ButtonGroup>
+              <Button color="primary" href={`${WEBAPP_URL}/settings/teams/new`}>
+                {t("create_team")}
+              </Button>
+              <Button color="secondary" href="https://go.cal.com/teams-video" target="_blank">
+                {t("learn_more")}
+              </Button>
+            </ButtonGroup>
+          </div>
+        }>
+        <>
+          <EmptyScreen
+            Icon={Icon.FiUsers}
+            headline={t("no_teams")}
+            description={t("no_teams_description")}
+            buttonRaw={
+              <Button color="secondary" href={`${WEBAPP_URL}/settings/teams/new`}>
+                {t("create_team")}
+              </Button>
+            }
+          />
+          {teams.length > 0 && <TeamList teams={teams} />}
+        </>
+      </UpgradeScreen>
+    </>
+  );
+}
+
+function UpgradeScreen({
+  title,
+  description,
+  background,
+  features,
+  buttons,
+  children,
+}: {
+  title: string;
+  description: string;
+  background: string;
+  features: Array<{ icon: JSX.Element; title: string; description: string }>;
+  buttons?: JSX.Element;
+  children: JSX.Element;
+}) {
+  const { data, isLoading } = trpc.viewer.teams.list.useQuery(undefined, {});
+  const teams = data?.filter((m) => m.accepted) || [];
+  const { t } = useLocale();
+
+  return (
+    <>
       {!teams.length && !isLoading && (
         <>
           {!isCalcom ? (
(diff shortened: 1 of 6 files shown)


packages/features/ee/teams/components/TeamsListing.tsx
```diff
@@ -65,31 +64,80 @@ export function TeamsListing() {
...
+            description={t("no_teams_description")}
+            buttonRaw={
+              <Button color="secondary" href={`${WEBAPP_URL}/settings/teams/new`}>
+                {t("create_team")}
+              </Button>
+            }
+          />
+          {teams.length > 0 && <TeamList teams={teams} />}
+        </>
+      </UpgradeScreen>
```
PeerRich, line 100: @sean-brydon @JeroenReumkens I hope to move as much of "team" logic into the upgrade screen so the pages dont need to call teams individually to check if someone has a team or not.

please comment on the props API used here

packages/features/ee/teams/components/TeamsListing.tsx
```diff
@@ -65,31 +64,80 @@ export function TeamsListing() {
...
+            headline={t("no_teams")}
+            description={t("no_teams_description")}
+            buttonRaw={
+              <Button color="secondary" href={`${WEBAPP_URL}/settings/teams/new`}>
+                {t("create_team")}
+              </Button>
+            }
+          />
+          {teams.length > 0 && <TeamList teams={teams} />}
+        </>
```
PeerRich, line 99: is this clear enough that the children is the "else" case when someone has a team plan?

PeerRich:
the objective here is to use the UpgradeScreen for Routing Forms too

PeerRich:
converted to draft until I have approval for the Props API. Will then add the wrapper to routing forms

JeroenReumkens:
@PeerRich Great work! I would improve a few things to make it more readable. Let me know what you think about it 😁  Also improves the children like you mentioned in your comment indeed. 

1. Perhaps we should name it `UpgradeToTeams`, so it's more clear what upgrading is?
2. I'm not a big fan of the `buttons` prop where we pass elements into a component. It makes it quite inflexible if you one day want to for example add text underneath the button. This most of the time results in way more props than necessary in the future. I'd suggest making it a separate component looking similar to: 

```
<UpgradeToTeams
  // Perhaps features could even be a separate component, since it's a reusable card thing? Only content differs.
  features={features}
  background="/team-banner-background.jpg"
>
  <h1>{t('calcom_is_better_with_team')}</h1>
  <p>{t('add_your_team_members')}</p>
  <UpgradeToTeams.Buttons>
    <Button color="primary" href={`${WEBAPP_URL}/settings/teams/new`}>
      {t("create_team")}
    </Button>
     <Button color="secondary" href="https://go.cal.com/teams-video" target="_blank">
       {t("learn_more")}
     </Button>
  </UpgradeToTeams.Buttons>
</UpgradeToTeams>
```

This way the buttongroup and wrapper div (with space-y) can also be part of the `UpgradeToTeams.Buttons`, so that becomes also a lot more consistent (people can't forget).

3. The children as an `else` case can be quite misleading indeed. People will overlook how that works. I'd suggest to: 
- Move the EmptyScreen when there ARE teams into the `teams.length > 0` check as well. 
- Always render the `UpgradeToTeams` component on top, but let the component itself return null when there's teams. This way it won't show itself. 
- As for performance, the query will be cached if you use it both on the team page as well as in the component, so that's not a problem.

PeerRich:
but if we do teams.length > 0 outside of the component than we always have to import trpc into the page, even if its not needed?

JeroenReumkens:
> but if we do teams.length > 0 outside of the component than we always have to import trpc into the page, even if its not needed?

That's correct. Although you are importing it anyways, either in the page, or in the component. The result is the same — the query will be done to check whether the user is in a team, right?

PeerRich:
`<UpgradeToTeams>` we may wanna use this in console to upgrade to a different platform plan, so for now I'm gonna keep this generic

packages/features/ee/teams/components/TeamsListing.tsx
```diff
@@ -67,23 +34,29 @@ export function TeamsListing() {
 
       {isLoading && <SkeletonLoaderTeamList />}
 
-      <UpgradeScreen
-        title="calcom_is_better_with_team"
-        description="add_your_team_members"
-        features={features}
-        background="/team-banner-background.jpg"
```
PeerRich: @JeroenReumkens  why did you remove this prop and hardcode `background: "url(/team-banner-background.jpg)` again?

the background has to be different for each UpgradeScreen

packages/features/teams/components/TeamFeatures.tsx
```diff
@@ -0,0 +1,50 @@
+import { APP_NAME } from "@calcom/lib/constants";
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+import { Icon } from "@calcom/ui";
+
+export const TeamFeatures = () => {
+  const { t } = useLocale();
+  const features = [
+    {
+      icon: <Icon.FiUsers className="h-5 w-5 text-red-500" />,
+      title: t("collective_scheduling"),
```
PeerRich, line 10: this file should not exist. UpgradeScreen should include the feature column as a required prop @JeroenReumkens

packages/features/teams/components/UpgradeScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { isInTeam, isLoading } = useIsInTeam();
+
+  if (isLoading || (!isLoading && isInTeam) || isCalcom) return null;
+
+  return (
+    <div className="-mt-6 rtl:ml-4 md:rtl:ml-0">
+      <div
+        className="flex w-full justify-between overflow-hidden rounded-lg pt-4 pb-10 md:min-h-[295px] md:pt-10"
+        style={{
+          background: background,
```
PeerRich, line 16: re-added background props

packages/features/teams/hooks/useIsInTeam.ts
```diff
@@ -0,0 +1,10 @@
+import { trpc } from "@calcom/trpc/react";
+
+export const useIsInTeam = () => {
+  const { data, isLoading } = trpc.viewer.teams.list.useQuery(undefined, {});
+  const teams = data?.filter((team) => team.accepted) || [];
+  return {
+    isLoading,
+    isInTeam: teams.length > 0,
+  };
+};
```
PeerRich, line 10: i like this. @CarinaWolli should we use this in other places where we check for teams?

packages/features/teams/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./components/UpgradeScreen";
+export * from "./components/TeamFeatures";
```
PeerRich, line 2: we wouldnt need this barrel if TeamFeatures would be in UpgradeScreen and UpgradeScreen would be a generic component like EmptyScreen.

delete this file

PeerRich:
@sean-brydon is taking over. we just hopped off a peer-programming call

PR #6570, opened 2023-01-19
fix: show only apps enabled by admin

Quentin-Guillemin:
## What does this PR do?

Applications not activated by the administrator were available at installation in the event types.

## Type of change

## How should this be tested?

With the administrator account disable applications and check if they are proposed in the event types 

## Checklist

### DIFF
--- a/apps/web/components/eventtype/EventAppsTab.tsx
+++ b/apps/web/components/eventtype/EventAppsTab.tsx
@@ -18,8 +18,8 @@ export const EventAppsTab = ({ eventType }: { eventType: EventType }) => {
     extendsFeature: "EventType",
   });
   const methods = useFormContext<FormValues>();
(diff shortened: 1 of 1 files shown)


emrysal:
Hi @Quentin-Guillemin; looking at the code the `eventTypeApps` variable should never contain apps without `enabled` - will close as this should never be necessary - but please reply if you encountered a bug with this.

```javascript
await prisma.app.findMany({ select: { slug: true, enabled: true } });
```

Quentin-Guillemin:
Hi @emrysal,

I've got a bug :

Here are my apps enabled

Capture d’écran du 2023-01-23 09-21-04

But in an event type, apps unavailable are listed and installable

Capture d’écran du 2023-01-23 09-20-36

Issue #6014, opened 2022-12-13
[CAL-590] RFC: Self-hosted instance metadata DB table


zomars:
### Is your proposal related to a problem?

We're getting to the point where we need to save some instance specific settings and values where using `.env` variables are not enough.

### Describe the solution you'd like

- Add an `Deployment` model to our DB `schema` that will allow us to save deployment specific settings.

We could use this table to save things like:
- Whitelabel data
  - App name
  - App Logo
  - Theme colors
- License data
  - `CALCOM_LICENSE_KEY` The actual license key for enterprise users.
  - `licenseConsentAt` A date of when the admin user has agreed to the Cal.com license
- Etc... ?

### Describe alternatives you've considered

- More environment variables? 

### Additional context

This would be a special table as would be a "Single record table". We can keep adding columns per needed but it should only have a single row. So the usage would be pretty simple. Example:

```ts
// Retrieve this deployment app logo
const deployment = await prisma.deployment.findFirst({
  where: { id: 1 },
  select: {
    appLogo: true,
  }
});
```

<sub>[CAL-590](https://linear.app/calcom/issue/CAL-590/rfc-self-hosted-instance-metadata-db-table)</sub>

zomars:
This could be used for #5998

PR #6574, opened 2023-01-19
Admin Wizard Choose License

### LINKED ISSUE #5998: [CAL-586] Admin Onboarding: Choose License
to replace "LICENSE_CONSTENT" and "LICENSE_KEY" in .env, we should move this into the database.

we should add a "Choose License" step to the admin onboarding (when User = 0)

next step if "enterprise is selected"

disabled if input is empty:

CleanShot 2022-12-13 at 12 53 57

[https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/Cal---New-Features?node-id=1426%3A116568](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/Cal---New-Features?node-id=1426%3A116568)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-586](https://linear.app/calcom/issue/CAL-586/admin-onboarding-choose-license)</sub>

leog:
## What does this PR do?

Implements 2 more steps in the Admin wizard for self-hosters to choose a license.

Closes #5998 #6014

Loom Video: UPDATED: https://loom.com/share/0d58e2fa5683478ebb52954b92eb6ee9

## Type of change

## How should this be tested?

Go to `/auth/setup` in a self-hosted Cal.com and see the wizard come up or If you don't have any user created in the database it will come up automatically when visiting the app. You can see two more steps or just one depending on the license chosen after setting up the first user (admin).


apps/web/pages/auth/setup/index.tsx
```diff
@@ -9,26 +10,83 @@ import prisma from "@calcom/prisma";
...
+  const [isLoadingAdminUser, setIsLoadingAdminUser] = useState(false);
+  const [isFreeLicense, setIsFreeLicense] = useState(
+    props.deployment?.licenseType !== DeploymentLicenseType.EE
+  );
+  const [isLoadingAdminApps, setIsLoadingAdminApp] = useState(false);
+  const [isLoadingChooseLicense, setIsLoadingChooseLicense] = useState(false);
+  const [isLoadingEnterpriseLicense, setIsLoadingEnterpriseLicense] = useState(false);
+  const [isEnabledEnterpriseLicense, setIsEnabledEnterpriseLicense] = useState(
+    !!props.deployment?.licenseKey
+  );
```
leog, line 29: These are states to keep track of various things among steps.

apps/web/pages/auth/setup/index.tsx
```diff
@@ -9,26 +10,81 @@ import prisma from "@calcom/prisma";
...
+      ) : (
+        <ChooseLicense
+          setIsLoading={setIsLoadingChooseLicense}
+          isFreeLicense={isFreeLicense}
+          setIsFreeLicense={setIsFreeLicense}
+        />
+      ),
+      isLoading: isLoadingChooseLicense,
     },
+    ...(!isFreeLicense
```
leog, line 57: If it's enterprise what's selected, add an additional step

packages/config/tailwind-preset.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   content: [
     "./pages/**/*.{js,ts,jsx,tsx}",
     "./components/**/*.{js,ts,jsx,tsx}",
-    "../../packages/app-store/**/{components,pages}/**/*.{js,ts,jsx,tsx}",
+    "../../packages/app-store/**/{_components,_pages}/**/*.{js,ts,jsx,tsx}",
```
leog, line 9: Components and pages folders were renamed and Tailwind process was not scanning them.

packages/ui/components/form/step/Stepper.tsx
```diff
@@ -16,12 +17,13 @@ function Stepper<T extends DefaultStep>(props: {
...
     stepLabel = (currentStep, totalSteps) => `Step ${currentStep} of ${totalSteps}`,
   } = props;
+  const [stepperRef] = useAutoAnimate<HTMLOListElement>();
   return (
     <>
       {steps.length > 1 && (
         <nav className="flex items-center justify-center" aria-label="Progress">
           <p className="text-sm font-medium">{stepLabel(props.step, steps.length)}</p>
-          <ol role="list" className="ml-8 flex items-center space-x-5">
+          <ol role="list" className="ml-8 flex items-center space-x-5" ref={stepperRef}>
```
leog, line 26: Animating adding more steps dynamically.

packages/ui/components/form/wizard/WizardForm.tsx
```diff
@@ -37,7 +39,7 @@ function WizardForm<T extends DefaultStep>(props: {
       <img className="mx-auto mb-8 h-8" src="https://cal.com/logo.svg" alt="Cal.com Logo" />
       <div
         className={classNames(
-          "mb-8 divide-y divide-gray-200 overflow-hidden rounded-lg bg-white shadow print:divide-transparent print:shadow-transparent",
+          "mb-8 w-[700px] divide-y divide-gray-200 overflow-hidden rounded-lg bg-white shadow print:divide-transparent print:shadow-transparent",
```
leog, line 42: Wizard was extending way more in width that it should, so limiting it from here.

apps/web/pages/auth/setup/steps/ChooseLicense.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <li>{t("required_to_keep_your_code_open_source")}</li>
+            <li>{t("cannot_repackage_and_resell")}</li>
+            <li>{t("no_enterprise_features")}</li>
+          </ul>
+        </div>
+        <div
+          className={classNames(
+            "cursor-pointer space-y-2 rounded-md border p-4 hover:border-black",
+            !isFreeLicense && "border-black"
+          )}
```
sean-brydon, line 54: Can we make this a button please

Clickable divs have no a11y support and require you to manually setup things like roles aria label and tab index aswell as a bunch of other things

apps/web/pages/auth/setup/steps/ChooseLicense.tsx
```diff
@@ -0,0 +1,70 @@
...
+          licenseType: isFreeLicense ? DeploymentLicenseType.FREE : DeploymentLicenseType.EE,
+        });
+      }}>
+      <div className="grid grid-cols-2 gap-4">
+        <div
+          className={classNames(
+            "cursor-pointer space-y-2 rounded-md border p-4 hover:border-black",
+            isFreeLicense && "border-black"
+          )}
+          onClick={() => setIsFreeLicense(true)}>
```
sean-brydon, line 40: Can we make this a button please

Clickable divs have no a11y support and require you to manually setup things like roles aria label and tab index aswell as a bunch of other things

sean-brydon: Also these appear to very simular to radio buttons under the hood - Use Radix if you need to do some fancy styling

apps/web/pages/auth/setup/steps/ChooseLicense.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { t } = useLocale();
+  const router = useRouter();
+  const { isFreeLicense, setIsFreeLicense } = props;
+  const mutation = trpc.viewer.auth.deploymentSetup.useMutation({
+    onSuccess: () => {
+      router.replace(`/auth/setup?step=3${isFreeLicense ? "&category=calendar" : ""}`);
+    },
+  });
+  return (
+    <form
```
sean-brydon, line 23: Use our Form component and use rhf to prevent some of the state tracking above

apps/web/pages/auth/setup/steps/EnterpriseLicense.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <TextField
+          className="mt-1"
+          placeholder="b6052119-1fee-42a6-90e3-55cjgjaaebb7"
+          labelSrOnly={true}
+          onChange={(e) => {
+            props.setIsEnabled(e.target.value !== "");
+            formMethods.setValue("licenseKey", e.target.value);
+          }}
+          id="licenseKey"
+          name="licenseKey"
```
sean-brydon, line 65:
```suggestion
          {...formMethods.register("licenseKey"),{
             onChange: (e) =>  props.setIsEnabled(e.target.value !== "")
          }}
```

apps/web/pages/auth/setup/index.tsx
zomars: That's a lot of useStates

packages/prisma/schema.prisma
```diff
@@ -643,6 +653,11 @@ enum WorkflowMethods {
   SMS
 }
 
+enum DeploymentLicenseType {
+  FREE
+  EE
+}
+
```
zomars, line 660: Can we split these migrations in a prep PR?

packages/config/tailwind-preset.js
zomars: This is an incomplete change. This is referring to each individual app's pages and components. You're replacing them with shared components. The correct solution should take both into account.

packages/ui/components/form/wizard/WizardForm.tsx
zomars: Should this be max width instead of a fixed one?

packages/app-store/_components/AppCategoryNavigation.tsx
```diff
@@ -31,7 +31,7 @@ const AppCategoryNavigation = ({
           tabs={appCategories}
           sticky
           linkProps={{ shallow: true }}
-          itemClassname={classNames(fromAdmin && "w-60")}
+          itemClassname={classNames(fromAdmin && "!w-48")}
```
zomars, line 34: what does the `!` do here?

packages/features/apps/AdminAppsList.tsx
```diff
@@ -183,26 +185,36 @@ const querySchema = z.object({
...
+  setIsLoading?: Dispatch<SetStateAction<boolean>>;
   useQueryParam?: boolean;
 }) => {
   const router = useRouter();
   return (
     <form
-      id="wizard-step-2"
-      name="wizard-step-2"
+      id={`wizard-step-${currentStep}`}
+      name={`wizard-step-${currentStep}`}
```
zomars, line 204: Why do we need currentStep here other than naming these?

packages/features/apps/AdminAppsList.tsx
```diff
@@ -183,26 +185,36 @@ const querySchema = z.object({
...
       onSubmit={(e) => {
         e.preventDefault();
+        setIsLoading && setIsLoading(true);
         router.replace("/");
       }}>
       <AppCategoryNavigation
         baseURL={baseURL}
-        fromAdmin
+        fromAdmin={fromAdmin}
```
zomars, line 212: Is this being used elsewhere when fromAdmin is not true?

packages/prisma/migrations/20230119020155_deployment_model/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+CREATE TABLE "Deployment" (
+    "id" SERIAL NOT NULL,
+    "logo" TEXT,
+    "theme" JSONB,
+    "licenseKey" TEXT,
+    "licenseType" "DeploymentLicenseType",
+    "licenseConsentAt" TIMESTAMP(3),
+
+    CONSTRAINT "Deployment_pkey" PRIMARY KEY ("id")
+);
```
zomars, line 14: https://github.com/calcom/cal.com/pull/6574/files#r1081847900

packages/trpc/server/routers/viewer/auth.tsx
```diff
@@ -86,6 +87,24 @@ export const authRouter = router({
...
+      })
+    )
+    .mutation(async ({ input }) => {
+      const data = {
+        licenseConsentAt: new Date(),
+        ...(input.licenseType ? { licenseType: input.licenseType } : {}),
+        ...(input.licenseKey ? { licenseKey: input.licenseKey } : {}),
+      };
+
+      await prisma.deployment.upsert({ where: { id: 1 }, create: data, update: data });
```
zomars, line 106: Since this is a special kind of table we should abstract these methods to limit the where clauses.  Since there will be only one and will be confusing for newcomers. could be something like `deploymentSettingsUpdate/Update` etc.

zomars (changes requested):
Blocking to address @sean-brydon and my feedback as well 🙏🏽

packages/ui/components/form/wizard/WizardForm.tsx
sean-brydon: Yes - this will likely break on mobile without max width

apps/web/pages/auth/setup/index.tsx
sean-brydon: Tracking some of these in rhf or even into a single state object would probably be the best here 

From what i can see - a lot of these are controlled by what are effectively radio buttons and would work directly in rhf

packages/prisma/schema.prisma
emrysal: Do self hosters really care about these models?

packages/features/apps/AdminAppsList.tsx
emrysal: Yeah, `fromAdmin` is false on the /apps page. It's true from the Settings pages.

emrysal:
@leog could you add a loom 🙏 ?

packages/prisma/schema.prisma
zomars: That's another good point @emrysal we don't want to mixup what should live in our console DB with here. We can also check for `CALCOM_LICENSE_KEY` for this.

leog:
> @leog could you add a loom 🙏 ?

@emrysal Totally forgot, will do.

> Do self hosters really care about these models?
> That's another good point @emrysal we don't want to mixup what should live in our console DB with here. We can also check for `CALCOM_LICENSE_KEY` for this.

@emrysal @zomars That's a good point, I just limited to implement what was needed from the ticket, which specifies creating those steps in the admin wizard and put things together following the `RFC: Self-hosted instance metadata DB table`. I was suggesting @PeerRich that we could probably send an API request to us to let us know when someone consents a license on their self-hosted instance to track down usages.

PeerRich:
how do I test this the best?

apps/web/public/static/locales/en/common.json
```diff
@@ -1512,5 +1514,21 @@
   "install_google_meet": "Install Google Meet",
   "install_google_calendar": "Install Google Calendar",
   "sender_name": "Sender name",
-  "no_recordings_found": "No recordings found"
+  "no_recordings_found": "No recordings found",
+  "agplv3_license": "AGPLv3 License",
+  "ee_enterprise_license": "“/ee” Enterprise License",
+  "enterprise_booking_fee": "$0.05/booking – min. $189/month",
```
PeerRich, line 1520: I am fine merging this but we need to create an issue to later get this from our API endpoint because pricing will change eventually. We need an endpoint that always returns the most recent pricing string

PeerRich: @zomars

PeerRich: also, our pricing is about to change in the next 2 days anyway

leog:
> how do I test this the best?

Locally, removing all users if you want to go through the admin wizard from scratch or signing in as admin and going to `/auth/setup`

apps/web/pages/auth/setup/steps/ChooseLicense.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <ul className="ml-4 list-disc text-xs text-gray-500">
+            <li>{t("no_need_to_keep_your_code_open_source")}</li>
+            <li>{t("repackage_rebrand_resell")}</li>
+            <li>{t("a_vast_suite_of_enterprise_features")}</li>
+          </ul>
+        </div>
+      </div>
+      <div>
+        <small>{t("you_can_upgrade_at_any_time_later")}</small>
+      </div>
```
PeerRich, line 71: fyi: @leog added this section

apps/web/pages/auth/setup/steps/ChooseLicense.tsx
```diff
@@ -0,0 +1,76 @@
...
+          </ul>
+        </div>
+        <div
+          className={classNames(
+            "cursor-pointer space-y-2 rounded-md border p-4 hover:border-black",
+            !isFreeLicense && "border-black"
+          )}
+          onClick={() => setIsFreeLicense(false)}>
+          <h2 className="font-cal text-xl text-black">{t("ee_enterprise_license")}</h2>
+          <p className="font-medium text-green-800">{t("enterprise_booking_fee", { enterprise_fee })}</p>
```
PeerRich, line 60: fyi: @leog made this a variable so we can pull it from an API endpoint to update later and not hardcode the price

apps/web/pages/auth/setup/steps/ChooseLicense.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { t } = useLocale();
+  const router = useRouter();
+  const { isFreeLicense, setIsFreeLicense } = props;
+  const mutation = trpc.viewer.auth.deploymentSetup.useMutation({
+    onSuccess: () => {
+      router.replace(`/auth/setup?step=3${isFreeLicense ? "&category=calendar" : ""}`);
+    },
+  });
+
+  const enterprise_fee = "$99"; // TODO: get this from a new API endpoint
```
PeerRich, line 23: @leog @zomars this needs to come from our backend so we can update them dynamically

PeerRich (approved):
submitted a commit, left some comments. rest looks good, very excited for this!

PeerRich:
oh i made a mistake, but should be an easy fix:
CleanShot 2023-01-20 at 11 22 52@2x

the {{enterprise_booking_fee}} is not being picked up correctly @leog

leog:
> the {{enterprise_booking_fee}} is not being picked up correctly @leog

@PeerRich Fixed. I will work on the suggestions from the other folks and will be ready Monday.

apps/web/pages/auth/setup/index.tsx
leog: Refactored to make the wizard handle it by the step content.

packages/prisma/schema.prisma
leog: The idea was to move out from environment variable, to let admins setup their instances as much as we can from the admin wizard, like we did with activating apps from it.

packages/app-store/_components/AppCategoryNavigation.tsx
leog: Makes it `!important` in order to overwrite any other similar CSS.

packages/features/apps/AdminAppsList.tsx
leog: Apps list in the wizard can be step 3 or 4 depending on the license you choose, so it needs to be dynamically set.

packages/prisma/schema.prisma
leog: @zomars I'm afraid that doing a prep PR with the migrations separately could introduce churn on why was the new model was introduced. Any particular reason for doing it? Just to understand the rationale, learn and do it differently next time.

leog:
Latest Loom: https://loom.com/share/0d58e2fa5683478ebb52954b92eb6ee9

sean-brydon:
I'm not sure if its just me but i kinda believe this should be one form ultimately and we handle the final submit on the last step of the wizard and not on every step. 

UX seems weird to me if you start completing this (without ever hitting a final step in the form) - close the tab and you need to resume from where you left off. Im not sure if my perception of a stepper form is off  but this just went against my intuition when testing it out. 

We do this elsewhere - eventTypes is a good example ... Each page "settings tab" isnt a unique form and the content is wrapped in a single form itself. 

@emrysal @zomars maybe im just being nit picky here but this just didnt click with me 

I know [RHF](https://react-hook-form.com/advanced-usage/) has a good example of a form stepper where each step is contained within one form

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -331,10 +331,14 @@ export default NextAuth({
       return token;
     },
     async session({ session, token }) {
-      const hasValidLicense = await checkLicense(process.env.CALCOM_LICENSE_KEY || "");
+      const deployment = await prisma.deployment.findFirst({ where: { id: 1 } });
+      const hasValidLicense = await checkLicense(deployment?.licenseKey || "");
```
Comment (leog) [inline apps/web/pages/api/auth/[...nextauth].tsx:335]: Now checking the new db model instead of environment variable

apps/web/pages/api/link.ts
```diff
@@ -39,7 +39,10 @@ async function handler(req: NextApiRequest, res: NextApiResponse<Response>) {
         username: "" /* Not used in this context */,
         role: UserPermissionRole.USER,
       },
-      hasValidLicense: true,
+      license: {
+        valid: true,
+        key: "",
+      },
```
leog, line 45: License validity has been moved to its own object together with the key in order to have it at hand for other usages.

apps/web/pages/auth/setup/steps/StepDone.tsx
```diff
@@ -1,20 +1,26 @@
...
+import { Dispatch, SetStateAction } from "react";
 
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { FiCheck } from "@calcom/ui/components/icon";
 
-const StepDone = () => {
+const StepDone = (props: {
+  currentStep: number;
+  nextStepPath: string;
+  setIsLoading: Dispatch<SetStateAction<boolean>>;
```
leog, line 10: StepDone could be shown at any completed step so it was parametrized to accommodate any step.

apps/web/public/static/locales/en/common.json
```diff
@@ -836,7 +836,7 @@
   "connect_metamask": "Connect Metamask",
   "create_events_on": "Create events on",
   "enterprise_license": "This is an enterprise feature",
-  "enterprise_license_description": "To enable this feature, get a deployment key at {{consoleUrl}} console and add it to your .env as CALCOM_LICENSE_KEY. If your team already has a license, please contact {{supportMail}} for help.",
+  "enterprise_license_description": "To enable this feature, have an administrator go to {{setupUrl}} to enter a license key. If a license key is already in place, please contact {{supportMail}} for help.",
```
leog, line 839: Changed the wording given the fact we moved out from environment variable to the new db model through the admin wizard.

apps/web/public/static/locales/en/common.json
```diff
@@ -1103,8 +1103,8 @@
   "close": "Close",
   "upgrade": "Upgrade",
   "upgrade_to_access_recordings_title": "Upgrade to access recordings",
-  "upgrade_to_access_recordings_description":"Recordings are only available as part of our teams plan. Upgrade to start recording your calls",
-  "recordings_are_part_of_the_teams_plan":"Recordings are part of the teams plan",
+  "upgrade_to_access_recordings_description": "Recordings are only available as part of our teams plan. Upgrade to start recording your calls",
+  "recordings_are_part_of_the_teams_plan": "Recordings are part of the teams plan",
```
leog, line 1107: i18n ally plugin corrections

apps/web/public/static/locales/en/common.json
```diff
@@ -1404,7 +1404,7 @@
   "team_url_required": "Must enter a team URL",
   "team_url_taken": "This URL is already taken",
   "team_publish": "Publish team",
-  "number_sms_notifications": "Phone number (SMS\u00a0notifications)",
+  "number_sms_notifications": "Phone number (SMS notifications)",
```
leog, line 1407: i18n ally plugin corrections

apps/web/public/static/locales/en/common.json
```diff
@@ -1486,11 +1488,11 @@
   "not_verified": "Not yet verified",
   "no_availability_in_month": "No availability in {{month}}",
   "view_next_month": "View next month",
-  "send_code" : "Send code",
+  "send_code": "Send code",
```
leog, line 1491: i18n ally plugin corrections

packages/config/tailwind-preset.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   content: [
     "./pages/**/*.{js,ts,jsx,tsx}",
     "./components/**/*.{js,ts,jsx,tsx}",
-    "../../packages/app-store/**/{components,pages}/**/*.{js,ts,jsx,tsx}",
+    "../../packages/app-store/**/*{components,pages}/**/*.{js,ts,jsx,tsx}",
```
leog, line 9: Correcting the glob to take into account "_components" and "_pages" within app-store and "components" and "pages" within app folders.

packages/features/apps/AdminAppsList.tsx
```diff
@@ -183,26 +185,36 @@ const querySchema = z.object({
...
         useQueryParam={useQueryParam}
-        containerClassname="w-full xl:mx-5 xl:w-2/3 xl:pr-5"
+        containerClassname={classNames(
+          "w-full xl:mx-5 xl:w-2/3 xl:pr-3.5",
+          fromAdmin ? " max-h-[400px] overflow-y-auto" : "max-w-[500px]"
```
leog, line 216: This is to have the app list within the admin wizard, contained inside the content section of the wizard and not expand its dimension, should be self-contained there.

packages/features/ee/api-keys/components/ApiKeyListItem.tsx
```diff
@@ -43,16 +43,8 @@ const ApiKeyListItem = ({
...
-              {t("expired")}
-            </Badge>
-          )}
-          {!isExpired && (
-            <Badge className="-p-2" variant="green">
-              {t("active")}
-            </Badge>
-          )}
+          {!neverExpires && isExpired && <Badge variant="red">{t("expired")}</Badge>}
+          {!isExpired && <Badge variant="green">{t("active")}</Badge>}
```
leog, line 47: Testing the API Keys section of the settings to show the "needs license" warning, couldn't help to notice the badges were wrongly padded, this is the fix.

packages/features/ee/common/components/LicenseRequired.tsx
```diff
@@ -32,7 +32,7 @@ const LicenseRequired = ({ children, as = "", ...rest }: LicenseRequiredProps) =
   const Component = as || Fragment;
   return (
     <Component {...rest}>
-      {session.data?.hasValidLicense ? (
+      {session.data?.license.valid ? (
```
leog, line 35: License validity has been moved to its own object together with the key in order to have it at hand for other usages.

packages/features/ee/common/components/v2/LicenseRequired.tsx
```diff
@@ -18,7 +18,7 @@ const LicenseRequired = ({ children, as = "", ...rest }: LicenseRequiredProps) =
   const session = useSession();
   const { t } = useLocale();
   const Component = as || Fragment;
-  const hasValidLicense = session.data ? session.data.hasValidLicense : null;
+  const hasValidLicense = session.data ? session.data.license.valid : null;
```
leog, line 21: License validity within the session object has been moved to its own object together with the key in order to have it at hand for other usages.

packages/features/ee/common/components/LicenseRequired.tsx
```diff
@@ -46,6 +46,7 @@ const LicenseRequired = ({ children, as = "", ...rest }: LicenseRequiredProps) =
                     consoleUrl: `<a href="https://go.cal.com/console" target="_blank" class="underline">
                 ${APP_NAME}
               </a>`,
+                    setupUrl: `<a href="${WEBAPP_URL}/auth/setup">/auth/setup</a>`,
```
leog, line 49: Text now refers to the admin wizard to be able to enter a license key instead of the environment variable.

packages/features/ee/common/components/v2/LicenseRequired.tsx
```diff
@@ -36,9 +36,9 @@ const LicenseRequired = ({ children, as = "", ...rest }: LicenseRequiredProps) =
                     consoleUrl: `<a href="${CONSOLE_URL}" target="_blank" rel="noopener noreferrer" class="underline">
                 ${APP_NAME}
               </a>`,
+                    setupUrl: `<a href="${WEBAPP_URL}/auth/setup" class="underline">/auth/setup</a>`,
```
leog, line 39: Text now refers to the admin wizard to be able to enter a license key instead of the environment variable.

packages/features/ee/common/components/v2/LicenseRequired.tsx
```diff
@@ -36,9 +36,9 @@ const LicenseRequired = ({ children, as = "", ...rest }: LicenseRequiredProps) =
...
                     supportMail: `<a href="mailto:${SUPPORT_MAIL_ADDRESS}" class="underline">
-                ${SUPPORT_MAIL_ADDRESS}
-              </a>`,
+                ${SUPPORT_MAIL_ADDRESS}</a>`,
```
leog, line 41: An extra underlined space after the email was shown, this fixes it.

packages/features/ee/common/server/checkLicense.ts
```diff
@@ -5,18 +5,9 @@ import { CONSOLE_URL } from "@calcom/lib/constants";
...
-  // .uuid() exists but I'd to fix the situation where the CALCOM_LICENSE_KEY is wrapped in quotes
-  .regex(/^\"?[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}\"?$/, {
-    message: "License key must follow UUID format: 8-4-4-4-12",
-  })
-  .transform((v) => {
-    // Remove the double quotes from the license key, as they 404 the fetch.
-    return v != null && v.length >= 2 && v.charAt(0) == '"' && v.charAt(v.length - 1) == '"'
-      ? v.substring(1, v.length - 1)
-      : v;
-  });
```
leog, line 19: We went back on using `.uuid` directly as quotes are not an option when entering it in the admin wizard form.

packages/features/ee/workflows/pages/index.tsx
```diff
@@ -43,7 +43,7 @@ function WorkflowsPage() {
       title={t("workflows")}
       subtitle={t("workflows_to_automate_notifications")}
       CTA={
-        session.data?.hasValidLicense && data?.workflows && data?.workflows.length > 0 ? (
+        session.data?.license.valid && data?.workflows && data?.workflows.length > 0 ? (
```
leog, line 46: License validity within the session object has been moved to its own object together with the key in order to have it at hand for other usages.

packages/features/ee/workflows/pages/workflow.tsx
```diff
@@ -251,7 +251,7 @@ function WorkflowPage() {
           </div>
         }
         heading={
-          session.data?.hasValidLicense &&
+          session.data?.license.valid &&
```
leog, line 254: License validity within the session object has been moved to its own object together with the key in order to have it at hand for other usages.

packages/prisma/zod-utils.ts
```diff
@@ -293,6 +293,24 @@ export const RoutingFormSettings = z
...
+    bookingLightest: z.string().default("#E1E1E1"),
+    bookingLighter: z.string().default("#ACACAC"),
+    bookingLight: z.string().default("#888888"),
+    bookingMedian: z.string().default("#494949"),
+    bookingDark: z.string().default("#313131"),
+    bookingDarker: z.string().default("#292929"),
+    fontName: z.string().default("Cal Sans"),
+    fontSrc: z.string().default("https://cal.com/cal.ttf"),
+  })
+  .optional();
```
leog, line 312: This is to support future theme customization within the admin wizard.

packages/trpc/server/routers/viewer.tsx
```diff
@@ -1148,6 +1148,7 @@ export const viewerRouter = mergeRouters(
     loggedInViewerRouter,
     public: publicViewerRouter,
     auth: authRouter,
+    deploymentSetup: deploymentSetupRouter,
```
leog, line 1151: As you suggested @zomars.

packages/ui/components/button/Button.tsx
```diff
@@ -210,7 +210,7 @@ export const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, ButtonPr
             <StartIcon
               className={classNames(
                 variant === "icon" && "h-4 w-4",
-                variant === "button" && "h-4 w-4 stroke-[1.5px] ltr:mr-2 rtl:ml-2"
+                variant === "button" && "h-4 w-4 stroke-[1.5px] ltr:ml-2 rtl:mr-2"
```
leog, line 213: I think this was the other way around as a bug, could see it in the "Purchase a License" button as anchor that includes am "external link" icon at the right of the button text.

packages/ui/components/button/Button.tsx
```diff
@@ -245,7 +245,7 @@ export const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, ButtonPr
               className={classNames(
                 "inline-flex",
                 variant === "icon" && "h-4 w-4",
-                variant === "button" && "h-4 w-4 stroke-[1.5px] ltr:mr-2 rtl:ml-2"
+                variant === "button" && "h-4 w-4 stroke-[1.5px] ltr:ml-2 rtl:mr-2"
```
leog, line 248: I think this was the other way around as a bug, could see it in the "Purchase a License" button as anchor that includes am "external link" icon at the right of the button text.

packages/ui/components/credits/Credits.tsx
```diff
@@ -23,7 +25,7 @@ export default function Credits() {
       </Link>{" "}
       {hasMounted && (
         <Link href="https://go.cal.com/releases" target="_blank" className="hover:underline">
-          {CalComVersion}
+          {CalComVersion(session.data?.license.valid)}
```
leog, line 28: Relying on the license validity instead of its existence.

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -117,7 +117,7 @@ export const InputField = forwardRef<HTMLInputElement, InputFieldProps>(function
         </Skeleton>
       )}
       {addOnLeading || addOnSuffix ? (
-        <div className="relative mb-1 flex items-center rounded-md focus-within:outline-none focus-within:ring-2 focus-within:ring-neutral-800 focus-within:ring-offset-1">
+        <div className="group relative mb-1 flex items-center rounded-md focus-within:outline-none focus-within:ring-2 focus-within:ring-neutral-800 focus-within:ring-offset-1">
```
leog, line 120: This allows to better control the hover states of the optional end or start icons of the field.

packages/ui/components/form/wizard/WizardForm.tsx
```diff
@@ -30,48 +33,52 @@ function WizardForm<T extends DefaultStep>(props: {
   const setStep = (newStep: number) => {
     router.replace(`${href}?step=${newStep || 1}`, undefined, { shallow: true });
   };
+  const [currentStepIsLoading, setCurrentStepIsLoading] = useState(false);
+
+  useEffect(() => {
+    setCurrentStepIsLoading(false);
+  }, [currentStep]);
```
leog, line 40: This is to be able to delegate the `isLoading` state to the step itself if it includes the callable element for that purpose. With this I was able to remove all `useStates` you guys were noticing @zomars @sean-brydon.

packages/ui/components/form/wizard/WizardForm.tsx
```diff
@@ -30,48 +33,52 @@ function WizardForm<T extends DefaultStep>(props: {
...
 
   return (
     <div className="mx-auto mt-4 print:w-full">
       {/* eslint-disable-next-line @next/next/no-img-element */}
-      <img className="mx-auto mb-8 h-8" src="https://cal.com/logo.svg" alt="Cal.com Logo" />
+      <img className="mx-auto my-8 h-8" src="https://cal.com/logo.svg" alt="Cal.com Logo" />
```
leog, line 45: Some space was needed also at the top of the Cal.com logo

sean-brydon (commented):
Also just a nit - i dont think this is the right stepper component? 

https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=1375%3A68066&t=MWFt6z1Mr9X67SHj-4

packages/ui/components/form/wizard/WizardForm.tsx
```diff
@@ -30,48 +33,52 @@ function WizardForm<T extends DefaultStep>(props: {
...
       <div
         className={classNames(
-          "mb-8 divide-y divide-gray-200 overflow-hidden rounded-lg bg-white shadow print:divide-transparent print:shadow-transparent",
+          "mb-8 divide-y divide-gray-200 overflow-hidden rounded-lg bg-white shadow print:divide-transparent print:shadow-transparent md:w-[700px]",
```
leog, line 48: Just applying wizard form width correction on medium+ screens

leog:
> Also just a nit - i dont think this is the right stepper component?

We are using the same one that was already there for the wizard in Console if I remember correctly. You should definitely create a follow up issue to correct this if you like @sean-brydon, and let's confirm with @Jaibles just in case.

Jaibles:
We should be using the same stepper component everywhere. The one that's in [the DS](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=1375%3A68066&t=skOTDLXzSf0UHM5W-1)

zomars:
How do FREE users finish this step?

image

apps/web/pages/auth/setup/steps/AdminUser.tsx
```diff
@@ -2,6 +2,7 @@ import { zodResolver } from "@hookform/resolvers/zod";
 import classNames from "classnames";
```
zomars, line 2: These files should not live here:

image

These should not be valid URLs

http://localhost:3000/auth/setup/steps/AdminUser

apps/web/pages/auth/setup/index.tsx
```diff
@@ -1,39 +1,86 @@
...
+          <AdminUser setIsLoading={setIsLoading} />
+        ),
+    },
+    {
+      title: t("choose_a_license"),
+      description: t("choose_license_description"),
+      contentEl: (setIsLoading: Dispatch<SetStateAction<boolean>>) =>
+        props.deployment?.licenseType ? (
+          <StepDone currentStep={2} nextStepPath="/auth/setup?step=3" setIsLoading={setIsLoading} />
+        ) : (
```
zomars, line 42: What' s the point of having a "prev step" if we cannot do anything about it?

image

IMO we should allow to go back and change the license. Makes sense to not do that with the first admin user creation tho.

packages/prisma/schema.prisma
zomars: Also we can assume that an empty key is a "FREE" license.

packages/trpc/server/routers/viewer/deploymentSetup.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import prisma from "@calcom/prisma";
+
+import { router, authedProcedure } from "../../trpc";
+
+export const deploymentSetupRouter = router({
+  update: authedProcedure
+    .input(
+      z.object({
+        licenseType: z.nativeEnum(DeploymentLicenseType).optional(),
```
zomars, line 12: What is `licenseType` being used for?

packages/types/environment.d.ts
```diff
@@ -1,7 +1,5 @@
 declare namespace NodeJS {
   interface ProcessEnv {
-    /** Needed to enable enterprise-only features */
-    readonly CALCOM_LICENSE_KEY: string | undefined;
```
zomars, line 4: We should create a migration for upsetting the value to the DB first.

packages/ui/components/credits/Credits.tsx
```diff
@@ -1,15 +1,17 @@
...
 
 import { COMPANY_NAME, IS_SELF_HOSTED } from "@calcom/lib/constants";
 import pkg from "@calcom/web/package.json";
 
-export const CalComVersion = `v.${pkg.version}-${!IS_SELF_HOSTED ? "h" : "sh"}-${
-  process.env.CALCOM_LICENSE_KEY === "" ? "ee" : "ce"
-}`;
+export const CalComVersion = (licenseKey: boolean | undefined) => {
+  return `v.${pkg.version}-${!IS_SELF_HOSTED ? "h" : "sh"}-${licenseKey ? "ee" : "ce"}`;
+};
```
zomars, line 10: This should come from a public endpoint

packages/ui/components/form/wizard/WizardForm.tsx
zomars: WDYM? I think I don't follow.

packages/ui/components/form/wizard/WizardForm.tsx
```diff
@@ -1,16 +1,19 @@
...
 type DefaultStep = {
   title: string;
+  containerClassname?: string;
+  contentClassname?: string;
   description: string;
-  content: JSX.Element;
-  enabled?: boolean;
-  isLoading: boolean;
+  content?: JSX.Element;
+  contentEl?: (setIsLoading: Dispatch<SetStateAction<boolean>>) => JSX.Element;
```
zomars, line 14: What does `contentEl` means in this context?

zomars (changes requested):
Overall nice progress @leog although I think this PR is not in a merge-able state yet. Left some feedback

packages/prisma/schema.prisma
```diff
@@ -615,6 +615,16 @@ model WorkflowsOnEventTypes {
   eventTypeId Int
 }
 
+model Deployment {
+  id               Int                    @id @default(1)
```
zomars, line 619: This change is missing an extra migration. Will open up another PR to migrate first.

packages/lib/telemetry.ts
```diff
@@ -35,24 +36,28 @@ export function collectPageParameters(
...
 const reportUsage: EventHandler = async (event, { fetch }) => {
-  const ets = telemetryEventTypes;
-  if ([ets.bookingConfirmed, ets.embedBookingConfirmed].includes(event.eventType)) {
-    const key = process.env.CALCOM_LICENSE_KEY;
-    const url = `${CONSOLE_URL}/api/deployments/usage?key=${key}&quantity=1`;
-    try {
-      return fetch(url, { method: "POST", mode: "cors" });
-    } catch (e) {
-      console.error(`Error reporting booking for key: '${key}'`, e);
+  const session = await getSession();
```
zomars, line 39: Why use session instead of prisma here? AFAIK `getSession` makes a request as well so there's no fetching benefit here

packages/lib/telemetry.ts
```diff
@@ -35,24 +36,28 @@ export function collectPageParameters(
...
   } else {
-    return Promise.resolve();
+    return undefined;
   }
 };
 
 export const nextCollectBasicSettings: CollectOpts = {
   drivers: [
-    process.env.CALCOM_LICENSE_KEY && process.env.NEXT_PUBLIC_IS_E2E !== "1" ? reportUsage : undefined,
+    reportUsage,
```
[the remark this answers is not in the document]
zomars: Do we want to report E2E usage?

packages/prisma/schema.prisma
zomars: https://github.com/calcom/cal.com/pull/6706

.env.example
```diff
@@ -1,24 +1,10 @@
...
-# Summary of terms:
-# - The codebase has to stay open source, whether it was modified or not
-# - You can not repackage or sell the codebase
-# - Acquire a commercial license to remove these terms by visiting: cal.com/sales
-#
-# To enable enterprise-only features, fill your license key in here.
-# @see https://console.cal.com
-CALCOM_LICENSE_KEY=
-# ***********************************************************************************************************
-
```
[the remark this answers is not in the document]
leog: Added a row creation instruction when visiting the admin setup to support the env var

apps/web/pages/api/link.ts
leog: I needed the key in the telemetry context, perhaps we can remove it from telemetry so I can prescind from including it in the session, which I don't really like

apps/web/pages/auth/setup/index.tsx
leog: Done!

packages/lib/telemetry.ts
leog: I tried, but this code seems to be executed client side. I could rely on tRPC perhaps, or we can just remove the license key from telemetry all-together to avoid relying on session.

packages/lib/telemetry.ts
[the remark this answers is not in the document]
leog: No sure 🤔

packages/trpc/server/routers/viewer/deploymentSetup.tsx
leog: to store the license type EE or FREE, if we choose to use `licenseKey` only, we can't distinguish between an admin that already went through the wizard or not

packages/types/environment.d.ts
leog: Did it here: https://github.com/calcom/cal.com/pull/6574/files#diff-da0845fc4b52d249c4f05f5150127d2d9bc00151a5a1272a69afa2b8a161bad9R114

packages/ui/components/credits/Credits.tsx
leog: Is this a blocker? Otherwise I can work on it in a follow up PR

packages/trpc/server/routers/viewer/deploymentSetup.tsx
zomars: How so? The wizard should display for empty license key

packages/ui/components/form/wizard/WizardForm.tsx
leog: Instead of having an external state for each step which the step itself needs to manipulate and the wizard needs to know the result, I just made the step manipulate its own loading state and allowing the wizard access to it.

packages/trpc/server/routers/viewer/deploymentSetup.tsx
leog: Ow, that's right, we don't need to remember which license was selected because we want to allow the admin to go back in the wizard to change it, making sense of the "Previous Step" button. Yeah, we can remove the license type I guess.

leog:
Applied all the feedback. Thanks everyone. @PeerRich @zomars @sean-brydon 

- Switched the stepper for the correct one
- Changed the wizard styles to match the design system
- Allowed to go back in steps without blocking choosing another option for license
- Removed the license type as we don't need to keep track of which license was selected to show already done steps
- Added "License" in Admin section of User settings and Cmd+k "Choose a license" option that gos to the admin wizard
- Dynamically removed the first step regarding creating the first (admin) user if user count was already different than 0
- Made sure console wizard also looked good with the wizard revamped styles to follow DS -> https://github.com/calcom/console/pull/64
- Also note there are changes in API to get the license key from the database instead of the env var -> https://github.com/calcom/api/pull/229

Here is what it looks like:

https://user-images.githubusercontent.com/467258/214700677-ac701a0c-b358-426c-9a9a-85e644dd804d.mov

PeerRich:
this is getting really close to completion! lots of great changes recently

sean-brydon:
> Applied all the feedback. Thanks everyone. @PeerRich @zomars @sean-brydon @Jaibles
> 
> * Switched the stepper for the correct one
> * Changed the wizard styles to match the design system
> * Allowed to go back in steps without blocking choosing another option for license
> * Removed the license type as we don't need to keep track of which license was selected to show already done steps
> * Added "License" in Admin section of User settings and Cmd+k "Choose a license" option that gos to the admin wizard
> * Dynamically removed the first step regarding creating the first (admin) user if user count was already different than 0
> * Made sure console wizard also looked good with the wizard revamped styles to follow DS -> [Revamped the wizard to accomodate new style console#64](https://github.com/calcom/console/pull/64)
> * Also note there are changes in API to get the license key from the database instead of the env var -> [Relying on deployment license key api#229](https://github.com/calcom/api/pull/229)
> 
> > Warning
> > There is still a pending discussion with @zomars related to having the license key in the session to be able to use it in telemetry.
> 
> Here is what admin wizard looks like now:
> 
>  Screen.Recording.2023-01-25.at.18.57.16.mov

Yoo this is looking good! Am i missing where the designs are for this or did @Jaibles pull a uno reverse card after the fact :D As these look a bit different https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=23849%3A147371&t=zWPdT0NqVfk8GfXA-4

leog:
> Yoo this is looking good! Am i missing where the designs are for this or did @Jaibles pull a uno reverse card after the fact :D As these look a bit different https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=23849%3A147371&t=zWPdT0NqVfk8GfXA-4

https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr?node-id=1426:116568&comments-enabled=1&viewer=1 😖

sean-brydon:
> > Yoo this is looking good! Am i missing where the designs are for this or did @Jaibles pull a uno reverse card after the fact :D As these look a bit different https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=23849%3A147371&t=zWPdT0NqVfk8GfXA-4
> 
> https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr?node-id=1426:116568&comments-enabled=1&viewer=1 😖

Lol nice - @Jaibles im gonna guess its the one in the DS file right??

@leog Happy to lend a hand refactoring the UI if it turns out to be the wrong one :D

Jaibles:
What @leog has done is perfect. There were old designs for the console there before. (Think even Peer or Bailey might have done them cause they did some console design awhile back but also could have been me along time ago.) I updated them today to match our current onboarding so it was easier to re-use our existing stepper component and so it's more visually inline. No major changes and either way would've been fine but looks great to me now.

apps/web/pages/auth/setup/index.tsx
```diff
@@ -1,47 +1,98 @@
...
+          fromAdmin
+          /*
+            | userCount === 0 | isFreeLicense | maxSteps |
+            |-----------------|---------------|----------|
+            | T               | T             |        3 |
+            | T               | F             |        4 |
+            | F               | T             |        2 |
+            | F               | F             |        3 |
+          */
+          currentStep={(props.userCount === 0) === isFreeLicense ? 3 : !isFreeLicense ? 4 : 2}
```
leog, line 77: Adding the rationale behind the condition just in case.

apps/web/public/static/locales/en/common.json
leog: Not restoring this so we can rely on the admin wizard steps for people to have a better experience regarding license key

leog:
Thanks @Jaibles for the support 🤗

@zomars restored the env var for backward compatibility, the PR shrieked quite a bit now.

Other pending PRs related to this one:
 - Made sure console wizard also looked good with the wizard revamped styles to follow DS -> https://github.com/calcom/console/pull/64
 - Also note there are changes in API to get the license key from the database instead of the env var -> https://github.com/calcom/api/pull/229

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -360,7 +360,9 @@ export default NextAuth({
       return token;
     },
     async session({ session, token }) {
-      const hasValidLicense = await checkLicense(process.env.CALCOM_LICENSE_KEY || "");
+      const deployment = await prisma.deployment.findFirst({ where: { id: 1 } });
+      const licenseKey = process.env.CALCOM_LICENSE_KEY || deployment?.licenseKey || "";
+      const hasValidLicense = await checkLicense(licenseKey);
```
Comment (zomars) [inline apps/web/pages/api/auth/[...nextauth].tsx:365]: Will update to self-contain. Also for public API.

apps/web/pages/auth/setup/index.tsx
```diff
@@ -62,9 +119,21 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
...
+        licenseKey: process.env.CALCOM_LICENSE_KEY,
+        agreedLicenseAt: new Date(),
+      },
+    });
+  }
+
   return {
     props: {
       userCount,
+      deployment,
```
zomars, line 136: We should never expose the licenseKey to the frontend. This will.

apps/web/pages/auth/setup/index.tsx
```diff
@@ -1,47 +1,100 @@
...
+import AdminUser from "@components/setup/AdminUser";
+import ChooseLicense from "@components/setup/ChooseLicense";
+import EnterpriseLicense from "@components/setup/EnterpriseLicense";
 
 export default function Setup(props: inferSSRProps<typeof getServerSideProps>) {
   const { t } = useLocale();
-  const [isLoadingStep1, setIsLoadingStep1] = useState(false);
-  const shouldDisable = props.userCount !== 0;
+  const [isFreeLicense, setIsFreeLicense] = useState(props.deployment?.licenseKey === "");
+  const [isEnabledEE, setIsEnabledEE] = useState(!!props.deployment?.licenseKey);
```
zomars, line 19: We should not read the licenseKey in the frontend. Also if the license KEY is set we cannot choose the FREE version without any explanation. We should allow to select FREE or remove it altogether. It creates unexpected behavior.

apps/web/components/setup/ChooseLicense.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Dispatch, SetStateAction } from "react";
+
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+
+const ChooseLicense = (props: {
+  currentStep: number;
+  isFreeLicense: boolean;
+  setIsLoading: Dispatch<SetStateAction<boolean>>;
+  setIsFreeLicense: Dispatch<SetStateAction<boolean>>;
+}) => {
```
zomars, line 13: This components is caring for way more that it needs to:

```suggestion
const ChooseLicense = (
  props: {
    value: string;
    onChange: (value: string) => void;
    onSubmit: (value: EventTarget) => void;
  } & JSX.IntrinsicElements["form"]
) => {
```

apps/web/pages/auth/setup/index.tsx
```diff
@@ -62,9 +119,21 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
     };
   }
 
+  // Check existant CALCOM_LICENSE_KEY env var and acccount for it
+  if (process.env.CALCOM_LICENSE_KEY !== "" && !deployment?.licenseKey) {
```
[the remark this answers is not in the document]
leog: Covered, thx!

apps/web/pages/auth/setup/index.tsx
leog: Fixed, thx!

apps/web/pages/auth/setup/index.tsx
leog: Fixed, thx!

apps/web/components/setup/ChooseLicense.tsx
leog: Actually, we need the `currentStep` to be able to let the form know which step is the submit button expecting to submit.

packages/ui/components/form/step/Steps.tsx
```diff
@@ -5,25 +5,32 @@ interface ISteps {
   maxSteps: number;
   currentStep: number;
   navigateToStep: (step: number) => void;
+  stepOfLabel?: (currentStep: number, maxSteps: number) => string;
 }
 
 const Steps = (props: ISteps) => {
-  const { maxSteps, currentStep, navigateToStep } = props;
   const { t } = useLocale();
```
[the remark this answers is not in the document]
zomars: Following the Stepper component example

apps/web/components/setup/ChooseLicense.tsx
zomars: Refactored to reduce prop drilling.

zomars: > Actually, we need the `currentStep` to be able to let the form know which step is the submit button expecting to submit.

We only need ´currenStep´ to form the new href. This can happen on the parent hence the use of onSubmit instead.

zomars (approved):
LGTM

leog:
We should have tested this better before merging @zomars, I had to create #6954 and #6977 afterwards.

apps/web/public/static/locales/en/common.json
PeerRich: this was a mistake. we need to keep `get a deployment key at {{consoleUrl}} console and`

PR #6585, opened 2023-01-20
Bye bye submodules

zomars:
## What does this PR do?

DEMO: https://www.loom.com/share/c8e4955578df4e50886f67dd82d48c49

This gets rids of private submodules by default.

Worked on a custom `git-setup.sh` script that will allow us to opt-in if we want to work on any specific submodule. We will remove `.gitmodules` from the main repo to avoid public users to auto-initialize it.

This will allow enterprise users to run `git-setup.sh api` to check if they have access to the repo and if not link them to our console. And if they have access, initialize the `.gitmodules` with only the api module setup so they don't get any console, website , etc. errors.

For the core team we can run `git-setup.sh api console website`  to get all available submodules to work locally.

Additionally added a `preinstall` `make` command that will check for the presence of a `.gitmodules` file and run `git-setup.sh api console website` only if not present.

### Downsides

- `yarn.lock` will be different between free, enterprise and core team members. Default will be public.
- Requires custom scripts for deploying modules in Vercel (which we already do anyways)

## Type of change

### As a public user

- [ ] Clone this branch
- [ ] Run `yarn`
- [ ] All should go normal and an empty `.gitmodules` should be created

### As an enterprise user with API access

- [ ] Clone this branch
- [ ] Run `yarn`
- [ ] All should go normal  you should have `apps/api` in `.gitmodules`

### As an core team memeber

- [x] Clone this branch
- [x] Run `yarn`
- [x] All should go normal  you should have api, website and console in `.gitmodules`


Makefile
```diff
@@ -0,0 +1,4 @@
+all: .gitmodules
+
+.gitmodules:
+	git-setup.sh api website console
```
zomars, line 4: If there's no `.gitmodules` file it will run the setup on `make`. This prevents running this command more than necessary.

package.json
```diff
@@ -50,6 +50,7 @@
     "lint": "turbo run lint",
     "postinstall": "turbo run post-install",
     "pre-commit": "lint-staged",
+    "preinstall": "make",
```
zomars, line 53: Before installing runs setup only if not has run once already.

git-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+  # Display usage and stop
+  echo "Usage: git-setup.sh <api,console,website>"
+  exit 1
+fi
+
+for module in "$@"; do
+  echo "Setting up '$module' module..."
+  # Set the project name, adding .git if necessary
+  project=$(echo "git@github.com:calcom/$module.git")
+  [ "$(git ls-remote $project 2>/dev/null)" ] && {
```
zomars, line 13: This makes a call to check if we have access to that repo

git-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "You have access to '${module}'"
+    ([ -e ".gitmodules" ] || touch ".gitmodules") && [ ! -w ".gitmodules" ] && echo cannot write to .gitmodules && exit 1
+    git submodule add --force git@github.com:calcom/$module.git apps/$module
+    git config -f .gitmodules --add submodule.apps/$module.branch main
+  } || {
+    if [ "$module" == "api" ]; then
+      echo "You don't have access to: '${module}' module. You can request access in: https://console.cal.com"
+      exit 0
```
zomars, line 21: We prompt to go to console only when getting api module

git-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+for module in "$@"; do
+  echo "Setting up '$module' module..."
+  # Set the project name, adding .git if necessary
+  project=$(echo "git@github.com:calcom/$module.git")
+  [ "$(git ls-remote $project 2>/dev/null)" ] && {
+    echo "You have access to '${module}'"
+    ([ -e ".gitmodules" ] || touch ".gitmodules") && [ ! -w ".gitmodules" ] && echo cannot write to .gitmodules && exit 1
```
zomars, line 15: Create a `.gitmodules` file if it doesn't exists

git-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+    git submodule add --force git@github.com:calcom/$module.git apps/$module
+    git config -f .gitmodules --add submodule.apps/$module.branch main
```
zomars, line 17: Initialize only the modules we have access to.

zomars (commented):
Self-review done

.gitignore
```diff
@@ -80,3 +80,13 @@ apps/storybook/build-storybook.log
...
+
+# Submodules
+.gitmodules
+apps/api
+apps/website
+apps/console
+/apps/api
+/apps/console
+/apps/website
+/.gitmodules
```
zomars, line 92:
```suggestion
```

git-setup.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+# If no project name is given
```
hariombalhara, line 2: Possible to use https://github.com/google/zx instead of shell script? 
Makes it easy to understand and contribute in it.

Makefile
```diff
@@ -0,0 +1,4 @@
+all: .gitmodules
```
hariombalhara, line 1: Maybe we can avoid the use of makefile by checking presence of  .gitmodules in the script itself?

git-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+    # Create the .gitmodules file if it doesn't exist
+    ([ -e ".gitmodules" ] || touch ".gitmodules") && [ ! -w ".gitmodules" ] && echo cannot write to .gitmodules && exit 1
+    # Prevents duplicate entries
+    git config -f .gitmodules --unset-all "submodule.apps/$module.branch"
+    # Add the submodule
+    git submodule add --force "git@github.com:calcom/$module.git" "apps/$module"
+    # Set the default branch to main
+    git config -f .gitmodules --add "submodule.apps/$module.branch" main
+    # Adding the subdmoule ignores the `.gitignore` so a reset is needed
+    git reset
```
hariombalhara, line 25: This would unstage any staged changed as well. Can we do it in a way that achieves what's needed without a sideffect

Makefile
```diff
@@ -0,0 +1,4 @@
...
+
+.gitmodules:
+	./git-setup.sh api website console
```
hariombalhara, line 4: It would add(try to add) all three submodules for non-core team members as well. Shouldn't we let the core team members run this command manually?

hariombalhara (commented):
Left some comments

Makefile
zomars: It has global support and I'd rather not reinvent the wheel. It's perfect for this exact use case

git-setup.sh
zomars: Not that I'm aware of. This will only run once tho. Suggestions always welcome.

Makefile
zomars: It will fail silently without disrupting anything. Did you try this locally?

emrysal (changes requested):
After checkout and running `yarn` I get the following error 🤔 

```
xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Makefile
zomars: Got rid of Makefile cc @emrysal

git-init.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# If there's a `.gitmodule` file skip this script
+[ -f .gitmodules ] && {
+  echo ".gitmodules already initializied"
```
emrysal, line 4: NIT typo

git-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+    echo "You have access to '${module}'"
+    # Create the .gitmodules file if it doesn't exist
+    ([ -e ".gitmodules" ] || touch ".gitmodules") && [ ! -w ".gitmodules" ] && echo cannot write to .gitmodules && exit 1
+    # Prevents duplicate entries
+    git config -f .gitmodules --unset-all "submodule.apps/$module.branch"
+    # Add the submodule
+    git submodule add --force "git@github.com:calcom/$module.git" "apps/$module"
+    # Set the default branch to main
+    git config -f .gitmodules --add "submodule.apps/$module.branch" main
+    # Adding the subdmoule ignores the `.gitignore` so a reset is needed
```
emrysal, line 24: Also typo NIT

emrysal (approved):
Approved, but if it can happen that there is a git reset with staged changes I would like a prompt that my staged changes are about to be reset, future improvement.

krumware:
FYI, the use of git@github.com in the repo paths requires users to have ssh keys setup properly in their shell for git authentication. Could be worth mentioning in the docs for folks who errantly receive `You don't have access to: 'api' module.`

krumware:
unexpected operator errors on Ubuntu (ignore the no-access, that's intentional)

```
Setting up 'api' module...
./git-setup.sh: 28: [: api: unexpected operator
You don't have access to: 'api' module.
Setting up 'website' module...
You don't have access to: 'website' module.
Setting up 'console' module...
./git-setup.sh: 28: [: website: unexpected operator
./git-setup.sh: 28: [: console: unexpected operator
You don't have access to: 'console' module.
```

emrysal:
@krumware Committed in a fix, should work now 👍

Issue #6577, opened 2023-01-19
[CAL-833] Improve Accepting Team Invite Flow


shirazdole:
As no team member received the confirmation email, I asked them to accept the invitation through the app, as they appeared as their status was “Pending”.

Some of our team members suggested that the option to accept the invitation could be in a more visible or direct place. Probably right after logging and entering to the app?

Created via Threads. See full discussion: [https://threads.com/34443144252](https://threads.com/34443144252)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-833](https://linear.app/calcom/issue/CAL-833/improve-accepting-team-invite-flow)</sub>

PR #6594, opened 2023-01-20
generic <UpgradeScreen> component

sean-brydon:
Upgrade screen with no teams
CleanShot 2023-01-20 at 12 19 23@2x

Upgrade with team a accepted team 
CleanShot 2023-01-20 at 12 18 43@2x

Upgrade screen with open invite
CleanShot 2023-01-20 at 12 18 04@2x


PeerRich:
ohh didnt think of open invites, smart!

PeerRich:
added routing forms: @sean-brydon 

CleanShot 2023-01-20 at 16 49 01@2x

packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx
```diff
@@ -27,12 +32,49 @@ export default function RoutingForms({
...
   const { t } = useLocale();
-  const { data: forms } = trpc.viewer.appRoutingForms.forms.useQuery(undefined, {
+
+  const { data: forms, isLoading } = trpc.viewer.appRoutingForms.forms.useQuery(undefined, {
     initialData: forms_,
   });
 
+  const { data } = trpc.viewer.teams.list.useQuery(); // TODO: use hasTeam hook
+  const teams = useMemo(() => data?.filter((m) => m.accepted) || [], [data]); // TODO: use hasTeam hook
+
```
Comment (PeerRich) [inline packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx:42]: @CarinaWolli turn this into a hook and in all other places where we check for teams not empty

Comment (CarinaWolli) [reply packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx:42]: working on it right now

packages/ui/components/badge/UpgradeTeamsBadge.tsx
```diff
@@ -1,12 +1,16 @@
...
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 
 import { Tooltip } from "../tooltip";
 import { Badge } from "./Badge";
 
 export const UpgradeTeamsBadge = function UpgradeTeamsBadge() {
   const { t } = useLocale();
+  const hasTeamPlan = useHasTeamPlan();
+
+  if (hasTeamPlan) return <></>;
```
sean-brydon, line 13:
```suggestion
  if (hasTeamPlan) return null;
```

packages/features/tips/UpgradeTip.tsx
```diff
@@ -0,0 +1,86 @@
...
+  buttons?: JSX.Element;
+  /**Chldren renders when the user is in a team */
+  children: JSX.Element;
+  isParentLoading?: ReactNode;
+}) {
+  const { data } = trpc.viewer.teams.list.useQuery();
+
+  const invites = useMemo(() => data?.filter((m) => !m.accepted) || [], [data]);
+  const { t } = useLocale();
+  const hasTeamPlan = useHasTeamPlan();
```
sean-brydon, line 38: My own thing with this is were calling the list query then has teams 🤔 They kinda do the same thing. Not sure what the best approach would be here

PeerRich:
@CarinaWolli can you fix the new merge conflict?
CleanShot 2023-01-20 at 21 52 01@2x

emrysal (dismissed):
Routing form tests are failing due to expectation it can be accessed.

PeerRich:
> Routing form tests are failing due to expectation it can be accessed.

ahhh thats why! great point

PeerRich:
> Routing form tests are failing due to expectation it can be accessed.

@hariombalhara can you help me out and change the test to include "hasTeam"

apps/web/pages/settings/my-account/appearance.tsx
```diff
@@ -74,8 +76,7 @@ const AppearanceView = () => {
     },
   });
 
-  if (isLoading || isLoadingHasTeamPlan)
-    return <SkeletonLoader title={t("appearance")} description={t("appearance_description")} />;
+  if (isLoading) return <SkeletonLoader title={t("appearance")} description={t("appearance_description")} />;
```
hariombalhara, line 79: We stopped using loading status from team plan query? This might let the content being shown for a moment, won't be a good UX.

packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx
```diff
@@ -15,10 +18,13 @@ import {
   List,
   ListLinkItem,
   Tooltip,
+  Button,
 } from "@calcom/ui";
 
 import { inferSSRProps } from "@lib/types/inferSSRProps";
 
+import SkeletonLoaderTeamList from "../../../../../features/ee/teams/components/SkeletonloaderTeamList";
```
Comment (hariombalhara) [inline packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx:26]: Should be @calcom/features

packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx
```diff
@@ -15,10 +18,13 @@ import {
...
+import { UpgradeTip } from "../../../../../features/tips";
```
Comment (hariombalhara) [inline packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx:27]: Same here

packages/features/tips/UpgradeTip.tsx
hariombalhara: We can simply do hasTeamPlan = data?.length

packages/features/tips/UpgradeTip.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  if (hasTeamPlan) return children;
+
+  if (!isCalcom)
+    return <EmptyScreen Icon={Icon.FiUsers} headline={title} description={description} buttonRaw={buttons} />;
+
+  if (isParentLoading) return <>{isParentLoading}</>;
```
hariombalhara, line 45: Seems not right to have isParentLoading here.
- I see that isParentLoading is the same component that we are rendering everywhere. So why not have that component rendered here directly.
- If there is a possibility of different components being rendered for different use cases, is it possible to achieve the same by putting this component behind that loader

hariombalhara (commented):
Added some comments.
Also, just reminding that the teams check exist only at UI side. User can still access the features using direct calls or API(probably?)

packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx
Comment (hariombalhara) [reply packages/app-store/ee/routing-forms/pages/forms/[...appPages].tsx]: Done

apps/web/pages/settings/my-account/appearance.tsx
hariombalhara: I added it back.

hariombalhara:
@PeerRich fixed tests

hariombalhara:
@PeerRich fixed tests

hariombalhara (approved):
Mergeable now. Some comments are there that we can take care of later maybe

zomars:
I don't think we should show this component in self-hosted instances. Let's follow up with a fix please.

PeerRich:
> I don't think we should show this component in self-hosted instances. Let's follow up with a fix please.

I think instead we should always return hasPlan = true on self-hosted

sean-brydon:
> I don't think we should show this component in self-hosted instances. Let's follow up with a fix please.
CleanShot 2023-01-24 at 09 49 45@2x
We do this already :)  - Looks like we merged with my dev isCalcom=true.

Fixing while working on #6577

packages/features/ee/teams/components/TeamsListing.tsx
```diff
@@ -59,66 +58,26 @@ export function TeamsListing() {
...
-                  {t("create_team")}
-                </Button>
-              }
-            />
-          )}
-        </>
-      )}
-      {teams.length > 0 && <TeamList teams={teams} />}
+      <UpgradeTip
+        title="calcom_is_better_with_team"
```
AaronPresley, line 62: @sean-brydon @PeerRich it seems this line (and `description` below) have resulted in the translation no longer being displayed. I've created [this issue](https://github.com/calcom/cal.com/issues/6838) as a way to track it.

PR #6626, opened 2023-01-22
Stripe App not showing in Event Type apps

### LINKED ISSUE #6615: [CAL-858] Stripe not installing
I've been trying to add stripe to my cal account, but unable to.

Here are the screenshots: image

image

Kindly help. My id is **anishjajodia**

<sub>[CAL-858](https://linear.app/calcom/issue/CAL-858/stripe-not-installing)</sub>

leog:
## What does this PR do?

Due to having two different slugs for the same app, Stripe was not showing up in Event Type Apps as installed, even though it was installed. Now it shows up:

<kbd>image</kbd>

This comes as a temporary fix, we should probably standardise using a single slug for it or perhaps using `dirName` to be sure we refer to the app internally where it is located. cc: @hariombalhara 

Fixes #6615 

## Type of change

## How should this be tested?

Go to an Event Type with payment enabled and see that Stripe is present in the Installed apps list for the event type settings.

### DIFF
--- a/packages/app-store/_components/EventTypeAppCardInterface.tsx
+++ b/packages/app-store/_components/EventTypeAppCardInterface.tsx
@@ -16,7 +16,11 @@ export const EventTypeAppCard = (props: {
   return (
     <ErrorBoundary message={`There is some problem with ${app.name} App`}>
       <EventTypeAppContext.Provider value={[getAppData, setAppData]}>
-        <DynamicComponent slug={app.slug} componentMap={EventTypeAddonMap} {...props} />
+        <DynamicComponent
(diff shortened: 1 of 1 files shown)


PeerRich:
indeed a hotfix but there gotta be a better fix with equally less work @hariombalhara @emrysal

PeerRich (approved):
approved and merged, tho we need to fix this in the next release @hariombalhara

hariombalhara:
Thank you so much for taking care of this on a weekend @leog 
The fix has to be like this only. This is a problem for apps that have slug different from dirName. All new apps can't have this in-consistency. 
It can be fixed by renaming `dirName` to stripe and rename `stripepayment` to `stripe`

PR #6640, opened 2023-01-23
Reduce bundle size by importing single icons at a time

JeroenReumkens:
## What does this PR do?

This PR reduces the bundle size (mainly in benefit to the Atoms later), but importing single icons at a time. Difference is over 100KB for the Atom. 

Before: 

CleanShot 2023-01-23 at 09 00 11@2x

After:

CleanShot 2023-01-23 at 09 00 41@2x

## Type of change

## How should this be tested?

The result should be exactly the same, since only the path of icons has been updated from `icons.[icon]` to `[icon]`.


zomars:
What happened? @JeroenReumkens

PR #6655, opened 2023-01-24
Modularize @heroicons

emrysal:
## What does this PR do?

Trials @heroicons/react/{outline,solid} as a modularised import. TODO see if this has any effect anywhere, if it doesn't close (no need to add this needlessly).

@JeroenReumkens could we have a quick look tomorrow? (as you've been benchmarking a lot)


emrysal:
Closing as I'm seeing a 50% bundle size increase with this change

PeerRich:
we also want to replace heroicons and feather icons with https://lucide.dev/ eventually

PR #6670, opened 2023-01-24
Fixes issues with workflows that have more than one action

### LINKED ISSUE #5940: [CAL-566] error in workflows
from a user:

When designing my workflows, I encountered the following error:
TypeError: e.steps[0] is undefined.

I've attached a screenshot of the situation below. Currently, I cannot edit or use any workflows which is annoying. Can you maybe reset my workflows so I can work on them again?

image

<sub>[CAL-566](https://linear.app/calcom/issue/CAL-566/error-in-workflows)</sub>

CarinaWolli:
## What does this PR do?

- Fixes issue that workflow with more than once action couldn't be saved
- Fixes data loading issues when an action was added. The sorting was wrong on the first load 
- Fixes error in /workflows when an untitled workflows doesn't have any steps

Fixes https://linear.app/calcom/issue/CAL-881/saving-workflow-fails-if-more-than-one-action)\
Fixes: #5940

## Type of change

## How should this be tested?

First two issues: 
1. Create a workflow with several actions
2. Click save 
3. Click on the workflow, and check if all actions are there and sorted correctly

Last issue:
1. Create a new workflow without a name
2. Delete all actions 
3. Click save
4. See that there is no error in /workflows and the workflow is called 'Untitled'

### DIFF
--- a/packages/features/ee/workflows/pages/workflow.tsx
+++ b/packages/features/ee/workflows/pages/workflow.tsx
@@ -100,7 +100,7 @@ function WorkflowPage() {
     data: workflow,
     isError,
     error,
-    isLoading,
+    dataUpdatedAt,
   } = trpc.viewer.workflows.get.useQuery(
     { id: +workflowId },
     {
@@ -111,7 +111,7 @@ function WorkflowPage() {
   const { data: verifiedNumbers } = trpc.viewer.workflows.getVerifiedNumbers.useQuery();
 
   useEffect(() => {
-    if (workflow && !isLoading) {
+    if (workflow && (workflow.steps.length === 0 || workflow.steps[0].stepNumber === 1)) {
       setSelectedEventTypes(
         workflow.activeOn.map((active) => ({
           value: String(active.eventType.id),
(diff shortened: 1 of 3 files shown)


packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -713,14 +713,14 @@ export const workflowsRouter = router({
...
+            const { senderName, ...newStep } = step;
             newStep.sender = getSender({
               action: newStep.action,
               sender: newStep.sender || null,
-              senderName: newStep.senderName,
+              senderName: senderName,
             });
             const createdStep = await ctx.prisma.workflowStep.create({
-              data: { ...step, numberVerificationPending: false },
+              data: { ...newStep, numberVerificationPending: false },
```
CarinaWolli, line 723: excluded senderName from newStep as it is not part of prisma.workflowSte. This caused the issue that workflow actions couldn't be added to a workflow anymore

packages/features/ee/workflows/pages/workflow.tsx
```diff
@@ -111,7 +111,7 @@ function WorkflowPage() {
   const { data: verifiedNumbers } = trpc.viewer.workflows.getVerifiedNumbers.useQuery();
 
   useEffect(() => {
-    if (workflow && !isLoading) {
+    if (workflow && (workflow.steps.length === 0 || workflow.steps[0].stepNumber === 1)) {
```
CarinaWolli, line 114: Temporary fix to make sure data is loaded correctly. Still need to find out why the order is wrong on first load when a step was added

emrysal (approved):
LGTM 👍 , I think this warrants an urgent deployment due to the multiple step saving issue.

PR #6691, opened 2023-01-25
fix teams text content to use from common

hybes:
## What does this PR do?

Teams text for users without a team, shows the placeholder and not the actual text.

## Type of change

## How should this be tested?

Used to show calcom_is_better_with_team and add_your_team_members - now it shows the actual content
## Checklist

### DIFF
--- a/packages/features/ee/teams/components/TeamsListing.tsx
+++ b/packages/features/ee/teams/components/TeamsListing.tsx
@@ -66,8 +66,8 @@ export function TeamsListing() {
       {!!errorMessage && <Alert severity="error" title={errorMessage} />}
 
       <UpgradeTip
-        title="calcom_is_better_with_team"
-        description="add_your_team_members"
+        title={t("calcom_is_better_with_team")}
+        description={t("add_your_team_members")}
         features={features}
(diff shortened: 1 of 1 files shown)


sean-brydon:
cal.com/packages/features/tips/UpgradeTip.tsx

Nice spot! We can actually make this change in the component its self so this doesnt happen in other places we use the component @Hybes are you up for picking this up?

Hybes:
> cal.com/packages/features/tips/UpgradeTip.tsx
> 
> Nice spot! We can actually make this change in the component its self so this doesnt happen in other places we use the component @Hybes are you up for picking this up?

not too sure on that, which part do you mean?

hariombalhara (approved):
I think @sean-brydon meant that the translation can be done in the Component itself and the Component user would just pass the translation keys.

But, I think this solves the immediate problem.

hariombalhara:
So, this one got fixed here https://github.com/calcom/cal.com/pull/6831/files#diff-d2343e642a5e3e3bb027fc0fd72ec76b0445c12b116325f3b40c742c15479af0 already

hariombalhara:
Closing it.

Issue #3457, opened 2022-07-20
[CAL-130] All CalDav Issues here!


alishaz-polymath:
## An updated CalDAV issue checklist

* [x] Nextcloud not syncing busy times #6756
* [x] Unclear error forwarding with CalDAV slow servers #6313
  https://github.com/calcom/cal.com/issues/6313
* [x] “e is undefined” error when adding CalDAV #5604
* [x] Meetings in Apple calendar are offset when calculating available time slots #5570
* [x] If a caldav server stops responding, it takes Installed Apps Page down. #2786
* [X] CalDAV integration with Baikal does not work #588
* [X] CalDAV availability checking is slow (30+ seconds) #479
* [X] CalDav/Lark Suite can't connect to Cal \[The Lark Suite isn't supported\] #2989
* [X] Nextcloud - JSON error on booking but it the booking goes through
* [X] CalDav/Zimbra - existing events (set to busy) don't reflect in the availability
* [X] Missing attendees in caldav events #3076
* [X] Availability fetching is bugged (atleast for CalDAV) #3589
* [X] Events with availability: free on my calendar still block booking availability #2236
* [X] Editing event location duplicates the event on my CalDAV calendar \[NextCloud\] (Note: this isn't limited to CalDAV but design choices. A fix is overdue) #3027
* [X] Caldav event is not updated on reschedule #2988
* [X] Busy booked events from CalDav (NextCloud) & Apple don't block scheduling availability #1588
* [X] Events created on kerio not considered by availability checker #4003
* [X] Apple Calendar/CalDav events are getting converted to GMT #2139
* [X] Rescheduling event created on CalDav calendar throws an error/unable to reschedule and get a new confirmation #4205
* [X] iCalendar timezone is off (doubles) CalDAV #5309

## Confirmed and verified list of all supported CalDAV integration calendar services

* [ ] 1CRM
* [ ] [A1.net](http://A1.net)
* [ ] [all-inkl.com](http://all-inkl.com)
* [ ] AOL Mail
* [X] Baïkal
* [ ] Bitrix24
* [x] bluemind (based on zimbra)
* [ ] Calendar and Contacts Server
* [ ] Confluence Calendar
* [ ] cPanel
* [ ] Cyrus IMAP
* [ ] DAViCal
* [ ] DavMail Gateway
* [ ] Davis
* [ ] Daylite
* [ ] DingTalk / 钉钉 / Dīngdīng
* [ ] Dolibarr
* [ ] EDIS
[...]

PR #6732, opened 2023-01-26
fix(calendar): recurring event fix

wzrdtales:
the loop was running way too often on every recurring object mostly due to the fact that the iterator was not checked for finalization, and in fact the iterator is not implemented correctly so it does not have the `done` property. instead it checks for undefined or not

also further note, this part is a very slow implementation, which will cause everything to become very slow and practically unusable, there are probably better ways to handle this.
Why? well simple, this iterates through every possible day, instead of just applying the current window to it. That means on the one hand this functionality is broken from the first place (b/c it wont work if it is a very long running event) and on the other it was implemented at a very high expense.

--

edit, added at least a start date to the iterator, that should increase the speed of this already.

Signed-off-by: Tobias Gurtzick <magic@wizardtales.com>

### DIFF
--- a/packages/lib/CalendarService.ts
+++ b/packages/lib/CalendarService.ts
@@ -360,7 +360,11 @@ export default abstract class BaseCalendarService implements Calendar {
 
         const start = dayjs(dateFrom);
         const end = dayjs(dateTo);
-        const iterator = event.iterator(ICAL.Time.fromDateTimeString(startISOString));
+        const startDate = ICAL.Time.fromDateTimeString(startISOString);
+        startDate.hour = event.startDate.hour;
+        startDate.minute = event.startDate.minute;
+        startDate.second = event.startDate.second;
+        const iterator = event.iterator(startDate);
         let current = iterator.next();
         let currentEvent;
         let currentStart = null;


PeerRich:
hey @wzrdtales thank you for the PR 

this touches core logic and requires a lot of eyeballs from our team. tagging @leog @emrysal and @zomars

wzrdtales:
> this touches core logic and requires a lot of eyeballs from our team.

agreed, this logic is broken in the first place though, so even more eyeballs to maybe suggest an even better fix. I validated the fix for now and it is indeed operational. I didn't include my hotfix for zoho calendars, see my comment in this issue regarding that.

https://github.com/calcom/cal.com/issues/3457

This is what you will get from me so far, I wont try to fix any code style issues, I don't have time for that really, you may want to introduce husky into your project and automatically adjust the CS to your standards, this would do contributors a big favor, b/c no one wants to adjust to someones random standards every 5 minutes when you switch working on a project. Auto Code style formatted with husky git hook do a long way here. Over and out

wzrdtales:
bump

PeerRich:
> bump

thank you for bumping, our lead dev @zomars  who would review this was on vacation for a week and will look at this monday

PeerRich:
there are a few other PRs working o recurring events too, so lets hope there wont be any merge conflicts cc @leog

zomars:
Hey @wzrdtales thanks for the contribution. I'm trying to get this branch up to date but I cannot push to it.

image

Can you enable "Allow edit by maintainers" as described in our [CONTRIBUTING guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md#making-a-pull-request)? Thank you

zomars:
We'll take it from here @wzrdtales Thank you for your contribution 🙏

wzrdtales:
so its off the radar again?

PR #6740, opened 2023-01-26
WIP: SIU: Maximum call stack size exceeded fix?

zomars:
## What does this PR do?

- It seems that we have a circular import/function/dependency that the TS wasn't catching before
- The main suspect relies on `@calcom/ui`
- No other PRs will pass the type checks until the issue is found and resolved.
- Tried to solve most imports that I've found but still getting the `Maximum call stack size exceeded` error

Here's a comment describing a way to debug this: https://github.com/microsoft/TypeScript/issues/40819#issuecomment-700379873

## Type of change

## How should this be tested?

- [ ] `yarn type-check --filter=@calcom/ui`
- [ ] `yarn type-check`

### DIFF
--- a/apps/web/components/booking/pages/BookingPage.tsx
+++ b/apps/web/components/booking/pages/BookingPage.tsx
@@ -38,6 +38,7 @@ import { useLocale } from "@calcom/lib/hooks/useLocale";
 import useTheme from "@calcom/lib/hooks/useTheme";
 import { HttpError } from "@calcom/lib/http-error";
 import { getEveryFreqFor } from "@calcom/lib/recurringStrings";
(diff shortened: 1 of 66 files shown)


zomars:
Turns our prisma client extensions is way too much for our TS codebase. Fixed by doing this

image

joeauyeung (approved):
Assuming that all the changes are just import fixes and it's just the schema file that needed changing, LGTM.

emrysal (changes requested):
Some instances remain of `import { Icon } from '../icon';` which shouldn't be used anymore (instead using the icons directly).

zomars:
Will fix types in a bit

emrysal (approved):

PR #6741, opened 2023-01-26
Fix embed modal leading inputs

sean-brydon:
Before
CleanShot 2023-01-26 at 20 21 41@2x

After:
CleanShot 2023-01-26 at 20 21 24@2x


alannnc (approved):

alannnc:
@sean-brydon It's failing because it has a prisma client but it's from previous commit no longer in main. Should we close and re upload PR with only necessary commit?

sean-brydon:
> @sean-brydon It's failing because it has a prisma client but it's from previous commit no longer in main. Should we close and re upload PR with only necessary commit?

Works for me - will pick this up tomorrow

sean-brydon:
Closing due to prisma issue - will reopen against main in the morning

PR #6769, opened 2023-01-27
Fix shift-enter in bio editor

CarinaWolli:
## What does this PR do?

Fixes error when using shift-enter in bio editor. 

**Thrown Error:** 
`Uncaught Error: insertNode: topLevelElement is root node`

## Type of change

## How should this be tested?

- test editor for user and team bio (use normal enter, shift-enter, formatting)

### DIFF
--- a/packages/lib/turndownService.ts
+++ b/packages/lib/turndownService.ts
@@ -2,19 +2,15 @@ import TurndownService from "turndown";
 
 const turndownService = new TurndownService();
 
-function isShiftEnter(node: HTMLElement) {
-  let currentNode: HTMLElement | null | ParentNode = node;
+function turndown(html: string | TurndownService.Node): string {
+  let result = turndownService.turndown(html);
+  result = result.replaceAll("[<p><br></p>]", "");
 
-  while (currentNode != null && currentNode.nodeType !== 1) {
-    currentNode = currentNode.parentElement || currentNode.parentNode;
+  if (result === "<p><br></p>") {
+    result = "";
   }
 
-  return (
-    currentNode &&
-    currentNode.nodeType === 1 &&
-    currentNode.parentElement &&
-    currentNode.parentElement.childNodes.length !== 1 // normal enter is <p><br><p> (p has exactly one childNode)
-  );
+  return result;
 }
 
 turndownService.addRule("shiftEnter", {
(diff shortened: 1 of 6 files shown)


packages/lib/turndownService.ts
```diff
@@ -2,11 +2,48 @@ import TurndownService from "turndown";
 
 const turndownService = new TurndownService();
 
-turndownService.addRule("newLine", {
-  filter: ["br"],
-  replacement: () => {
+function turndown(html: string | TurndownService.Node): string {
+  let result = turndownService.turndown(html);
+  result = result.replaceAll("[<p><br></p>]", "");
```
CarinaWolli, line 7: This is to avoid an error in the rare case someone tries to set a link in an empty line

packages/lib/turndownService.ts
```diff
@@ -2,11 +2,48 @@ import TurndownService from "turndown";
...
+
+  if (result === "<p><br></p>") {
+    result = "";
+  }
+
+  return result;
+}
+
+turndownService.addRule("shiftEnter", {
```
CarinaWolli, line 16: Add different rules for shift-enter and normal enter. This fixes the thrown error and the goal is to that shift-enter and normal enter should look the same

emrysal (approved):
Can confirm the shift+enter issue is fixed 🥂 

Found two minor issues for follow up (or to ignore):

* When I paste something that is multi-line but indented by tabs the new lines are not respected. Copy the following and try to save- you'll get this:

Before save:
```
⊂_ヽ
　 ＼＼
　　 ＼( ͡° ͜ʖ ͡°)
　　　 >　⌒ヽ
　　　/ 　 へ＼
　　 /　　/　＼＼
　　 ﾚ　ノ　　 ヽ_つ
　　/　/
　 /　/|
　(　(ヽ
　|　|、＼
　| 丿 ＼ ⌒)
　| |　　) /
ノ )　　Lﾉ
(_／
```
After:
```
`⊂_ヽ 　 ＼＼ 　　 ＼( ͡° ͜ʖ ͡°) 　　　 >　⌒ヽ 　　　/ 　 へ＼ 　　 /　　/　＼＼ 　　 ﾚ　ノ　　 ヽ_つ 　　/　/ 　 /　/| 　(　(ヽ 　|　|、＼ 　| 丿 ＼ ⌒) 　| |　　) / ノ )　　Lﾉ (_／`
```

* It doesn't seem to identify the tab character as an input value (the save button does not become clickable) 

This is what happens when you use spaces over tabs.

image

image

PR #6784, opened 2023-01-30
Add X-Frame-Options to page headers

emrysal:
## What does this PR do?

Uses regex to check if a page ends with /embed, if not we set the embed policy to DENY.

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -179,7 +179,7 @@ const nextConfig = {
   async headers() {
     return [
       {
-        source: "/auth/login",
+        source: "/:path*((?<!embed)$)",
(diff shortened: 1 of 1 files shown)


apps/web/next.config.js
```diff
@@ -176,6 +176,19 @@ const nextConfig = {
       }, */
     ];
   },
+  async headers() {
+    return [
+      {
+        source: "/:path*((?<!embed)$)",
```
hariombalhara, line 182:
```suggestion
        source: "/:path*((?<!(embed|embed\/preview.html))$)",
```

hariombalhara (approved):

PR #6810, opened 2023-01-31
Feat/select downshift  

sean-brydon:
https://www.loom.com/share/d7576baf010d48fb9c7454a035cdb285 

TODO: 
- [X] Fix input box being filled with [object Object] (I can't figure this out so if anyone sees something obvious please shout)
- [ ] Error in development ` The ref prop "undefined" from getComboboxProps was not applied correctly on your element.` however this only logs an error while not in production 🤔

### DIFF
--- a/packages/ui/components/form/select/SelectDownshift.tsx
+++ b/packages/ui/components/form/select/SelectDownshift.tsx
@@ -23,20 +23,11 @@ export function SelectDownshift() {
     const { getSelectedItemProps, getDropdownProps, addSelectedItem, removeSelectedItem, selectedItems } =
       useMultipleSelection<typeof books[number]>({ initialSelectedItems: [books[0]] });
     const items = books;
-    const {
-      isOpen,
-      selectedItem,
-      getToggleButtonProps,
-      getLabelProps,
-      getMenuProps,
-      highlightedIndex,
-      getItemProps,
-    } = useSelect({
-      selectedItem: null,
-      defaultHighlightedIndex: 0, // after selection, highlight the first item.
+    const { isOpen, getToggleButtonProps, getLabelProps, getMenuProps, getItemProps } = useSelect({
       items,
       stateReducer: (state, actionAndChanges) => {
         const { changes, type } = actionAndChanges;
+        console.log({ changes, type });
         switch (type) {
           case useSelect.stateChangeTypes.ItemClick:
             return {
@@ -45,17 +36,6 @@ export function SelectDownshift() {
             };
         }
         return changes;
-      },
-      onStateChange: ({ type, selectedItem }) => {
-        switch (type) {
-          case useSelect.stateChangeTypes.ItemClick:
-            if (selectedItem) {
-              addSelectedItem(selectedItem);
-            }
-            break;
-          default:
-            break;
-        }
       },
     });
 
@@ -97,17 +77,18 @@ export function SelectDownshift() {
               const isSelected =
                 selectedItems.filter((selectedItem) => selectedItem.title === item.title).length > 0;
               if (item.type === "divider") {
-                return <div className="h-[1px] w-full bg-gray-300" />;
+                return <div key={`${index}-divider`} className="h-[1px] w-full bg-gray-300" />;
               }
               return (
                 <li
                   className={classNames(
-                    highlightedIndex === index && "bg-gray-200",
-                    selectedItem === item && "font-bold",
-                    "flex cursor-pointer space-x-2 px-3 py-[10px] text-sm"
+                    "space-between flex cursor-pointer items-center space-x-2 px-3 py-[10px] text-sm hover:bg-gray-200"
                   )}
                   key={`${item.title}${index}`}
-                  {...getItemProps({ item, index })}>
+                  {...getItemProps({ item, index })}
+                  onClick={() => {
+                    isSelected ? removeSelectedItem(item) : addSelectedItem(item);
+                  }}>
                   <span className="no-highlight">{item.title}</span>
                   {isSelected && <span className="text-xs text-gray-400">✓</span>}
                 </li>
--- a/packages/ui/components/form/select/SelectDownshift.tsx
+++ b/packages/ui/components/form/select/SelectDownshift.tsx
@@ -1,0 +1,121 @@
+import { useMultipleSelection, useSelect } from "downshift";
+
+import { classNames } from "@calcom/lib";
+
+import { Label } from "../inputs/Label";
+
+export function SelectDownshift() {
+  const books = [
+    { author: "Harper Lee", title: "To Kill a Mockingbird" },
+    { author: "Lev Tolstoy", title: "War and Peace" },
+    { author: "Fyodor Dostoyevsy", title: "The Idiot" },
+    { author: "Oscar Wilde", title: "A Picture of Dorian Gray" },
+    { author: "George Orwell", title: "1984" },
+    { type: "divider" },
+    { author: "Jane Austen", title: "Pride and Prejudice" },
+    { author: "Marcus Aurelius", title: "Meditations" },
+    { author: "Fyodor Dostoevsky", title: "The Brothers Karamazov" },
+    { author: "Lev Tolstoy", title: "Anna Karenina" },
+    { author: "Fyodor Dostoevsky", title: "Crime and Punishment" },
+  ];
+
+  function MultipleSelect() {
+    const { getSelectedItemProps, getDropdownProps, addSelectedItem, removeSelectedItem, selectedItems } =
+      useMultipleSelection<typeof books[number]>({ initialSelectedItems: [books[0]] });
+    const items = books;
+    const {
+      isOpen,
+      selectedItem,
+      getToggleButtonProps,
+      getLabelProps,
+      getMenuProps,
+      highlightedIndex,
+      getItemProps,
+    } = useSelect({
+      selectedItem: null,
+      defaultHighlightedIndex: 0, // after selection, highlight the first item.
+      items,
+      stateReducer: (state, actionAndChanges) => {
+        const { changes, type } = actionAndChanges;
+        switch (type) {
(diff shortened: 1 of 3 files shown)


packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+import { classNames } from "@calcom/lib";
+
+import { Divider } from "../../divider";
+import { TextField } from "../inputs/Input";
+import { Label } from "../inputs/Label";
+
+type StringProperties<T> = {
+  [K in keyof T]: T[K] extends string ? K : never;
+}[keyof T];
```
sean-brydon, line 17: Typeof T that is only a string

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+interface MultiComboboxProps<T extends Record<string, unknown>> {
+  items: T[];
+  /** The field that will be used to display info */
+  labelKey: keyof T;
```
sean-brydon, line 22: Field that will display as label by default

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  // gets that can be used to search for items in text box
+  searchableKey?: StringProperties<T>;
```
sean-brydon, line 24: Any key on the data you wish to be searchable 

TODO: This could be an array of keys 🤔

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  dividerKey?: keyof T;
```
sean-brydon, line 25: If you wish to override the divider key from "divider"

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  initalSelectedItems?: T[];
```
sean-brydon, line 26: Any items that are initially  selected get passed here

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  onSelectedItemsChange?: (selectedItems: T[]) => void;
```
sean-brydon, line 27: Callback to get selected items

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  selectNode?: (
+    item: T,
+    index: number,
+    selection: { isActive: boolean; isKeyboardFocused: boolean },
+    // Downshift doesnt supply types here so we have to say as any
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    getItemProps: (options: UseComboboxGetItemPropsOptions<T>) => any
+  ) => ReactNode;
```
sean-brydon, line 35: Function to pass in your own selection node if you wish to have custom styles 

See storybook file

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  renderNode?: (
+    items: T[],
+    // Downshift doesnt supply types here so we have to say as any
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    getSelectedItemProps: (options: UseMultipleSelectionGetSelectedItemPropsOptions<T>) => any,
+    removeSelectedItem: (item: T) => void
+  ) => ReactNode;
```
sean-brydon, line 42: Function to pass in your own render of each item if you wish to have custom styles 

See storybook file

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  searchNode?: () => ReactElement<HTMLInputElement>;
```
sean-brydon, line 43: Full override the search box if needed

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  input?: {
+    className: string; // Nested so intelisense still picks up classNames
+  };
```
sean-brydon, line 46: Change the look of the input

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const [inputValue, setInputValue] = React.useState("");
+  const [selectedItems, setSelectedItems] = React.useState<T[]>(props.initalSelectedItems || ([] as T[]));
+
+  const items = React.useMemo(() => {
+    const lowerCasedInputValue = inputValue.toLowerCase();
+    return props.items.filter(function filterItem(item) {
+      const itemSearchable = props.searchableKey && (item[props.searchableKey] as string); // we know this is as string due to string properties type
+      return itemSearchable ? itemSearchable?.toLowerCase().includes(lowerCasedInputValue) : props.items;
+    });
+  }, [inputValue, props.items, props.searchableKey]);
```
sean-brydon, line 59: Memo the filter function otherwise it gets a canny few re-renders

TODO: maybe debounce this?

PeerRich:
is the search box like in figma?

sean-brydon:
> is the search box like in figma?

Not at the moment :)

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+                    key={`selected-item-${index}`}
+                    {...getSelectedItemProps({
+                      selectedItem: selectedItemForRender,
+                      index,
+                    })}>
+                    <span>
+                      <>{selectedItemForRender[props.labelKey]}</>
+                    </span>
+                    <span
+                      className="ml-2 cursor-pointer"
```
JeroenReumkens, line 151: I guess this should be a button with an a11y readable text?

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+  return (
+    <>
+      <div className="flex flex-col">
+        <Label {...getLabelProps()}>Pick some books:</Label>
+        <div
+          aria-label="menu-toggle"
+          {...getToggleButtonProps(getDropdownProps({ preventKeyAction: isOpen }))}
+          className="flex h-9 space-x-1 rounded-md border border-gray-300 p-1 focus-within:border-gray-900 hover:cursor-pointer hover:border-gray-400 focus-visible:border-gray-900">
+          {!props.renderNode
```
JeroenReumkens, line 137: I find the name 'renderNode' a bit confusing. Perhaps we can come up with a name that better explains what type of node it renders?

packages/ui/components/form/select/SelectDownshift.tsx
sean-brydon: I'm pretty sure the 
```
  {...getSelectedItemProps({
        selectedItem: selectedItemForRender,
        index,
      })}>
```
Adds an aria label to this div so likely not needed?

packages/ui/components/form/select/SelectDownshift.tsx
sean-brydon: Yeah 100%

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,248 @@
...
+                      &#10005;
+                    </span>
+                  </div>
+                );
+              })
+            : props.renderNode(selectedItems, getSelectedItemProps, removeSelectedItem)}
+        </div>
+      </div>
+      {/* @TODO: Make this UL (Button part) overrideable via props  */}
+      <ul
```
JeroenReumkens, line 165: It would be cool if we somehow could inject custom children in this select. 
By default you inject `<MultiSelect.List>`, which uses a context/zustand to get the data perhaps? And if you want to customise it, you can render your own component that uses the same hook to get the context, and render your own thing? 🤔 Makes the multiselect at least a lot easier from a props perspective. Just give it a thought 😁

JeroenReumkens: And looking a bit more into detail, it's perhaps a few more components instead of only one.

packages/ui/components/form/select/SelectDownshift.tsx
JeroenReumkens: Ooooh, that could very well be indeed. Isn't it a bit weird then that the onclick is on the span and not the div? I guess the div should be clickable then (and become the button)?

sean-brydon: Theres a weird reason why its a div and not a button if i remember what the docs were saying... Something to do with screen reader priority

And it needs to be on the whole div as this handles keyboard controls on the whole item which is as per the a11y spec 

e.g. pressing backspace/del will remove it when keyboard focused

packages/ui/components/form/select/SelectDownshift.tsx
```diff
@@ -0,0 +1,255 @@
...
+        case useCombobox.stateChangeTypes.ItemClick:
+          if (newSelectedItem) {
+            if (selectedItems.includes(newSelectedItem)) {
+              removeSelectedItem(newSelectedItem);
+              props.onSelectedItemsChange && props.onSelectedItemsChange(selectedItems);
+              return;
+            }
+            const newSelectedItems = [...selectedItems, newSelectedItem];
+            setSelectedItems(newSelectedItems);
+            props.onSelectedItemsChange && props.onSelectedItemsChange(newSelectedItems);
```
sean-brydon, line 119: propogate up to onchange handler

sean-brydon:
Closing in favour of #6865

PR #6822, opened 2023-01-31
Fix embed modal leading inputs

alannnc:
- From @sean-brydon original PR
Before
CleanShot 2023-01-26 at 20 21 41@2x

After:
CleanShot 2023-01-26 at 20 21 24@2x


sean-brydon (approved):
You are a star <3

PR #6823, opened 2023-01-31
I18N Caching

### LINKED ISSUE #6315: [CAL-698] RFC: Consider start splitting translation files per module/route/etc.
Right know we concentrate all of our translations string in a single `common.json` file which was fine for the time being. But as the app keeps growing we're starting to get some size warnings from next:

```
@calcom/web:dev: Warning: data for page "/auth/login" is 174 kB which exceeds the threshold of 128 kB, this amount of data can reduce performance.
@calcom/web:dev: See more info here: https://nextjs.org/docs/messages/large-page-data
```

The main culprit being that we download the whole `common.json` file on each page.

We could create namespaces for specific routes and call then on the get props methods as [described in the docs](https://github.com/i18next/next-i18next#4-declaring-namespace-dependencies).

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-698](https://linear.app/calcom/issue/CAL-698/rfc-consider-start-splitting-translation-files-per-modulerouteetc)</sub>

AaronPresley:
## What does this PR do?

Previously the I18N endpoint was intentionally being excluded from any caching. This PR updates the logic to add a 2 hour cache for the I18N endpoint. Additionally, it invalidates the i18n cache in the event a user successfully changes values within Settings > General.

Note that this PR _does not_ address the issue of the I18N endpoint being called too frequently, but simply adds caching to lessen the burden.

Fixes #6315

/claim #6315

## Type of change

## How should this be tested?

- [ ] Navigate the app with the browser's Inspector open. You should see the `viewer.public.i18n` endpoint now has a cache value from the Network tab
- [ ] Change your preferred language within Settings > General. Upon save you should see the page's language change without the need of a page reload

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works, because this code didn't have any tests previously. I will create new tests if requested to do so.


apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -30,34 +30,64 @@ export default trpcNext.createNextApiHandler({
    * @link https://trpc.io/docs/caching#api-response-caching
    */
   responseMeta({ ctx, paths, type, errors }) {
```
Comment (AaronPresley) [inline apps/web/pages/api/trpc/[trpc].ts:32]: @zomars @PeerRich I've reorganized this function a bit to prevent all logic from running each time the function is called. Instead I've split the large if statement into multiple smaller statements that hopefully make the code a bit easier to follow.

Additionally, the previous logic was to completely exclude a given path from being cached. I've updated the logic to instead of able to respond with different levels of caching depending on the value of `paths`.

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -30,34 +30,64 @@ export default trpcNext.createNextApiHandler({
...
-      );
-    }
+    // Some helpers relevant to this function only
+    const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+    const TWO_HOURS_IN_SECONDS = 60 * 60 * 2;
+
+    // Our response to indicate no caching
+    const noCacheResponse = {};
 
+    // TODO: Confirm this is doing what the original code intended
```
Comment (AaronPresley) [inline apps/web/pages/api/trpc/[trpc].ts:40]: Looking to get confirmation that the intent of this line from the original if statement was to not cache anything that's SSR? I'll remove this comment once I have clarity

- [ ] @zomars or @PeerRich to confirm caching is not wanted on SSR

PeerRich:
oh can you uncommit your yarn.lock? make sure to uncommit not delete it

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -30,34 +30,64 @@ export default trpcNext.createNextApiHandler({
...
+      return noCacheResponse;
+    }
+
+    // Never cache queries
+    if (type !== "query") {
+      return noCacheResponse;
     }
-    return {};
+
+    // TODO: Discuss whether this value achieves what the original comment below describes
```
Comment (AaronPresley) [inline apps/web/pages/api/trpc/[trpc].ts:64]: I obviously want to preserve the original caching logic, which was as follows:
```javascript
// cache request for 1 day + revalidate once every 5 seconds
let cacheValue = `s-maxage=5, stale-while-revalidate=${ONE_DAY_IN_SECONDS}`;
```
However I don't think this cache is doing what the comment describes. According the [the docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives), `s-max-age` is typically meant for CDN's or shared resources, which I don't think applies in this case? I might be wrong, but I feel the correct directive would be `max-age`.

Additionally, I think `max-age` should have the larger value, as `stale-while-revalidate` indicates how long the response is valid during the revalidation process.

Again, I could be completely misunderstanding here - but wanted to mention this.

- [ ] @zomars or @PeerRich to confirm that they want to preserve the original caching value

AaronPresley:
@PeerRich I'd appreciate some advice on the failing E2E Embed tests. I know E2E test are notoriously inconsistent, so should I run them again? I attempted to run them locally on my machine as well, but there seems to be an unrelated issue in getting them to run.

Happy to dig into it further, but figured I'd ask for your advice before spending hours tweaking on e2e tests 😬

PeerRich:
> @PeerRich I'd appreciate some advice on the failing E2E Embed tests. I know E2E test are notoriously inconsistent, so should I run them again? I attempted to run them locally on my machine as well, but there seems to be an unrelated issue in getting them to run.
> 
> Happy to dig into it further, but figured I'd ask for your advice before spending hours tweaking on e2e tests 😬

I think tests are borked right now, unrelated to this PR. @emrysal and @alannnc will investigate today!

PeerRich:
can you merge with main? i thikn tests are fixed now

AaronPresley:
@PeerRich Doesn't look like I can merge from my end. Seeing the following:
image

PeerRich:
just pull main and then git merge main

PeerRich:
i'm referring to your main branch not merging into ours

AaronPresley:
Think I got it? But also..
image

PeerRich:
@zomars will review tomorrow! 🙏

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -30,34 +30,64 @@ export default trpcNext.createNextApiHandler({
...
+    if (!paths || !paths.every((path) => path.startsWith("viewer.public."))) {
+      return noCacheResponse;
+    }
+
+    // No caching if we have any procedures errored
+    if (errors.length !== 0) {
+      return noCacheResponse;
+    }
+
+    // Never cache queries
```
Comment (zomars) [inline apps/web/pages/api/trpc/[trpc].ts:59]:
```suggestion
    // Never cache non-queries (aka mutations)
```

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -30,34 +30,64 @@ export default trpcNext.createNextApiHandler({
...
+    ];
+
+    // Making our paths a sorted string for easier comparison within our loop
+    const pathsVal = paths.sort().join(",");
+
+    // Find which element above is an exact match for this group of paths
+    const matchingIdx = cacheRules.findIndex((v) => (v[0] as Array<string>).sort().join(",") === pathsVal);
+
+    // Get the cache value of the matching element, if any
+    if (matchingIdx > -1) cacheValue = cacheRules[matchingIdx][1] as string;
```
Comment (zomars) [inline apps/web/pages/api/trpc/[trpc].ts:84]:
```suggestion
    // Our cache can change depending on our current paths value. Since paths is an array,
    // we want to create a map that can match potential paths with their desired cache value
    const cacheRules = {
      "viewer.public.i18n": `maxage=${TWO_HOURS_IN_SECONDS}, stale-while-revalidate=30`,
    } as const;

    // Find which element above is an exact match for this group of paths
    const matchedPath = paths.find((v) => v in cacheRules) as keyof typeof cacheRules;

    // Get the cache value of the matching element, if any
    if (matchedPath) cacheValue = cacheRules[matchedPath];
```

zomars (commented):
LGTM. Making some adjustments

zomars (approved):
LGTM

AaronPresley:
@zomars thanks for the review, let me know if you need anything else from me on this one 👍

hariombalhara:
@zomars Reschedule is failing because the test is able to pick the timeslot that's already booked by it in the previous step of the test. It's due to caching changes. Earlier `getSchedule` endpoint wasn't cached at all but now it is being cached.

On Production
Screenshot 2023-02-08 at 11 28 56 AM

On the branch
Screenshot 2023-02-08 at 11 29 36 AM

I am not sure why it wasn't cached earlier as per the code. But it mustn't be cached

hariombalhara (dismissed):
Preventing merge because of a genuine failure of reschedule test in embed

AaronPresley:
@zomars the endpoint pointed out by @hariombalhara is a non-mutation and was always intended to have a cache header response, correct? I had a really hard time getting the original code to _ever_ return my a cached header. I assumed it's because I'm just getting used to the code, but it's not surprising that there might be some pages that are suddenly being cached.

@zomars @PeerRich let me know how you'd like me to continue here. Happy to add `viewer.public.slots.getSchedule` to the list of paths to _not_ cache, or change the logic in some other way.

zomars:
getSchedule shouldn't be cached for two hours. A second to revalidate makes more sense tho.

AaronPresley:
@zomars as requested, I've update that path to have a revalidation time of 1 second.

Screenshot of it in action:
Screenshot 2023-02-08 at 1 07 45 PM

Let me know if anything is needed from my end 👍

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -30,34 +30,60 @@ export default trpcNext.createNextApiHandler({
...
+    }
+
+    // Never cache non-queries (aka mutations)
+    if (type !== "query") {
+      return noCacheResponse;
+    }
+
+    // TODO: Discuss whether this value achieves what the original comment below describes
+    // cache request for 1 day + revalidate once every 5 seconds
+    let cacheValue = `s-maxage=5, stale-while-revalidate=${ONE_DAY_IN_SECONDS}`;
```
Comment (zomars) [inline apps/web/pages/api/trpc/[trpc].ts:66]: Maybe I'm not familiar with the specs but isn't this "cache request for 5 seconds + revalidate once every 1 day" ?

Comment (AaronPresley) [reply apps/web/pages/api/trpc/[trpc].ts:66]: I agree with you. I called this out in [this comment](https://github.com/calcom/cal.com/pull/6823#discussion_r1092556306). I decided to preserve the original comment and code as it was before my change.

Comment (zomars) [reply apps/web/pages/api/trpc/[trpc].ts:66]: Let's fix it instead

Comment (AaronPresley) [reply apps/web/pages/api/trpc/[trpc].ts:66]: Will do - my comment is asking for clarity on what "fixed" means. Should the code be fixed to match the comment, or should the comment be fixed to match the code?

Comment (zomars) [reply apps/web/pages/api/trpc/[trpc].ts:66]: Let's match the comment instead. Caching for 5 seconds doesn't make much sense IMO

AaronPresley:
Apologies for the git churn 😅 attempted to rebase to `main` and had to fix some conflicts

zomars:
> Apologies for the git churn 😅 attempted to rebase to `main` and had to fix some conflicts

Can you revert those changes? You have 100+ files touched in changed files.

AaronPresley:
> Can you revert those changes? You have 100+ files touched in changed files.

Think I got it. Seeing 9 now, which is the number we were seeing after your modifications. Sorry about that, I'm new to the OSS fork workflow dealing with multiple remotes 😬

zomars (approved):
Thank you for your contribution 🙏  All good to merge now

LGTM

PR #6829, opened 2023-02-01
Fix empty screen tip

sean-brydon:
Translates empty screen on upgrade tip


PeerRich:
i am fixing this in a different PR but with different text

PR #6841, opened 2023-02-02
Beginning of Strict CSP Compliance

hariombalhara:
## What does this PR do?

**CSP Policy won't be added automatically and it is Opt In using CSP_POLICY env variable**

- Added 2 more security headers apart from Content-Security-Policy
	- X-Content-Type-Options
	- Referrer-Policy
- Enforces Strict CSP for the login page
	- If the user goes to any other route without a full page refresh, that would also follow strict CSP Policy restrictions. 
- Enables CSP in Report Only mode for all SSR Pages. So, nothing breaks there. It would just report the errors that come. Wherever a component is adding a script tag it would have to use `withNonce`.
- Introduces env variable CSP_POLICY with 2 possible values `strict`, `non-strict`
- Added a response header `x-nonce` with values `"ssr" | "initialPropsOnly"|"false"|"not-opted-in"`
	- `ssr` -> Page is an SSR Page and Component will not have `nonce` available as prop
	- `initialPropsOnly` ->  Page is probably an SSR Page but Components won't have `nonce` available as prop
	- `false` -> Due to some reason nonce wasn't available and thus CSP is disabled for the page.
	- `not-opted-in` -> CSP could be enabled but CSP_POLICY env variable isn't set.
	- If the header is not present, then CSP is not enabled for the page(e.g. SSG pages)

_Parser Inserted Script(without nonce) is blocked and img src that's not whitelisted is also blocked_
Screenshot 2023-02-02 at 6 38 52 PM

_CSP Evaluator Status for the Policy_
Screenshot 2023-02-02 at 6 51 50 PM

_securityheaders.com status for /auth/login_
Screenshot 2023-02-02 at 7 52 26 PM

# Important
**CSP_POLICY='non-strict' should be set to enable CSP in production. At any moment if we see any issues due to CSP, we need to simply remove this env variable and redeploy.**

## Followup
- [ ] Add a report-to directive that would log all CSP reports to Sentry.
- [ ] Find a way to fix [this error](https://reactjs.org/docs/error-decoder.html/?invariant=421) that comes only sometimes.
- [ ] Enforce CSP for other pages which has Report Only enabled.
- [ ] Figure out how to do it for SSG and ISR
- [ ] Add Permissions-Policy - We would probably be restricting all features like geolocation, camera access
- [ ] Explore require-trusted-types for directive
- [ ] Publish findings from https://www.notion.so/hariombalhara/CSP-Implementation-Cal-com-a9580b937fd441008dd4d81396bdc03b onto Github or gitbook under various sections like
	-  **Debugging CSP issues**
	- **Differences in CSP b/w local and production modes** 
	- **How to test CSP**

##
**Environment**:  Production

## Type of change

## How should this be tested?

Please ensure that you test the app in an instance with no browser extensions(usually incognito mode doesn't have browser extensions so that can be the quickest way to start testing). **This is important to avoid false failures arising due to resources requested by one of your extensions**

- [ ] Test that on Safari(<=15.3), Safari latest, Chrome and Firefox login page is working
- [ ] Also test that after a fresh login, when you browse the app all the pages are functional and all images are also loading. This is important because browsing after a fresh login keeps the entire app under CSP restrictions as all navigations happen in SPA flow. After refresh on any non login page, CSP won't apply for now.

## Checklist


apps/web/lib/withNonce.tsx
```diff
@@ -0,0 +1,82 @@
...
+    throw new Error(
+      "Strict CSP policy(for style-src) is not yet supported in production. You can experiment with it in Dev Mode"
+    );
+  }
+  return `
+	default-src 'self' ${isProduction ? "" : "data:"};
+	script-src 'self'  ${
+    isProduction
+      ? "'nonce-" + nonce + "' 'strict-dynamic' 'unsafe-inline' https:"
+      : // Note: We could use 'strict-dynamic' with 'nonce-..' instead of unsafe-inline but in dev mode there are some streaming related scripts that get blocked(because they don't have nonce on them)
```
hariombalhara, line 28: Found that such scripts are there for production build as well. The error doesn't come always and when it comes there is no visible functionality issue. So, we are okay with it for now.
Still keeping unsafe-inline for dev mode so that Next.js doesn't;t show that Error modal box that hinders browsing.

apps/web/lib/app-providers.tsx
```diff
@@ -77,6 +79,7 @@ const AppProviders = (props: AppPropsWithChildren) => {
           <TooltipProvider>
             {/* color-scheme makes background:transparent not work which is required by embed. We need to ensure next-theme adds color-scheme to `body` instead of `html`(https://github.com/pacocoursey/next-themes/blob/main/src/index.tsx#L74). Once that's done we can enable color-scheme support */}
             <ThemeProvider
+              nonce={props.pageProps.nonce}
```
hariombalhara, line 82: ThemeProvider adds a script and it supports a nonce that can be passed.

apps/web/lib/csp.ts
```diff
@@ -0,0 +1,82 @@
...
+	  child-src app.cal.com;
+	  style-src 'self' ${
+      IS_PRODUCTION ? (useNonStrictPolicy ? "'unsafe-inline'" : "") : "'unsafe-inline'"
+    } app.cal.com;
+	  font-src 'self';
+	  img-src 'self' https://www.gravatar.com https://img.youtube.com data:
+	`;
+}
+
+const isPagePathRequest = (url: URL) => {
```
hariombalhara, line 36: csp is to be shared with middleware to support SSG pages with CSP(it was being shared with it earlier but O removed that code for now) and there we need to differentiate whether it's a Page Request or an asset request.

apps/web/pages/_document.tsx
```diff
@@ -1,58 +1,36 @@
 import Document, { DocumentContext, Head, Html, Main, NextScript, DocumentProps } from "next/document";
 import Script from "next/script";
+import { z } from "zod";
 
 import { getDirFromLang } from "@calcom/lib/i18n";
 
-type Props = Record<string, unknown> & DocumentProps;
-
-function toRunBeforeReactOnClient() {
```
hariombalhara, line 8: Moved this to embed-init-iframe.ts so that inline code can be avoided.
Note that: This might not be necessary with strict-dynamic. But if we switch to non strict CSP policy it would come in handy.

hariombalhara (commented):
Self review done

emrysal:
Could you add `CSP_POLICY` the .env.example as well?

hariombalhara:
@emrysal Done

krumware:
Thanks for the heads up @emrysal!

hariombalhara:
Let's say that the cal.com deployment has added a script that isn't part of cal.com itself. If that script doesn't utilize the nonce it would be blocked potentially crashing the app. 

All scripts added in _document.tsx Head tag would automatically have nonce but scripts added elsewhere won't have that. e.g. see that ThemeProvider has to be provided nonce because it is added in _app.tsx and it adds a script tag.

PeerRich:
no what i mean is, why do we need a .env variable? sorry if that wasnt clear

emrysal (approved):
Can't fault it! LGTM

PR #6848, opened 2023-02-02
fix: Allow users to login with Google if they already have an account with the same email and password

devkiran:
## What does this PR do?

Allow users to login with Google if they already have an account with the same email address and password

## Type of change

## How should this be tested?

- Create an account with Email and Password, then login with Google using the same email address (should work)
- Create an account with Google, then sign up with Email and Password using the same email address (should not work)


deepakprabhakara:
Closed in favour of https://github.com/calcom/cal.com/pull/6874

PR #6849, opened 2023-02-02
feat: return `x-vercel-ip-timezone` in headers

### LINKED ISSUE #6679: [CAL-901] RFC: Consider adding timezone to request headers
This would allow us to compute most of needed data from the server.

We could use x-vercel-ip-timezone from Vercel by default with the option to override from the client if desired.

Some links about this:

[https://twitter.com/leeerob/status/1551273570713698304/photo/1](https://twitter.com/leeerob/status/1551273570713698304/photo/1)

[https://github.com/vercel/examples/blob/main/edge-functions/geolocation-country-block/middleware.ts](https://github.com/vercel/examples/blob/main/edge-functions/geolocation-country-block/middleware.ts)

[https://vercel.com/docs/concepts/edge-network/headers#x-vercel-ip-timezone](https://vercel.com/docs/concepts/edge-network/headers#x-vercel-ip-timezone)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-901](https://linear.app/calcom/issue/CAL-901/rfc-consider-adding-timezone-to-request-headers)</sub>

G3root:
## What does this PR do?

Fixes #6679

## Type of change


apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -47,17 +49,21 @@ export default trpcNext.createNextApiHandler({
...
-    // @ts-ignore ctx.req is available for SSR but not SSG
-    if (!!ctx?.req && allPublic && allOk && isQuery && !isThereANonCacheableQuery) {
+    if (!!ctx?.req && allOk && isQuery) {
       // cache request for 1 day + revalidate once every 5 seconds
       const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+      const timezone = z.string().safeParse(ctx.req.headers["x-cal-timezone"]);
       return {
         headers: {
-          "cache-control": `s-maxage=5, stale-while-revalidate=${ONE_DAY_IN_SECONDS}`,
+          ...(timezone.success && { "x-cal-timezone": timezone.data }),
```
Comment (G3root) [inline apps/web/pages/api/trpc/[trpc].ts:58]: pass time zone header to all queries

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -47,17 +49,21 @@ export default trpcNext.createNextApiHandler({
       );
     }
 
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore ctx.req is available for SSR but not SSG
-    if (!!ctx?.req && allPublic && allOk && isQuery && !isThereANonCacheableQuery) {
+    if (!!ctx?.req && allOk && isQuery) {
       // cache request for 1 day + revalidate once every 5 seconds
       const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+      const timezone = z.string().safeParse(ctx.req.headers["x-cal-timezone"]);
```
Comment (zomars) [inline apps/web/pages/api/trpc/[trpc].ts:55]: Shouldn't this be like this?

```suggestion
      const timezone = z.string().safeParse(ctx.req.headers["x-vercel-ip-timezone"]);
```

zomars (approved):
LGTM 

solved conflicts and updated to reflect changes introduced in #6823

Udit-takkar:
@zomars will this also work for self hosters who don't use vercel for deployment?

PeerRich:
> @zomars will this also work for self hosters who don't use vercel for deployment?

i dont think so

Udit-takkar:
how can we make it work for self hosters also . I am using similar headers here also https://github.com/calcom/cal.com/pull/6880#issuecomment-1419993533. 

Alternate approach is using this API ipapi.co like in this old commit https://github.com/calcom/cal.com/pull/6880/commits/5120fcf48a4a5715d0c6241b8cb943091f752027

G3root:
in this case we can rely on `dayjs.tz.guess()` for self hosters.

G3root:
@Udit-takkar  i don't think so 🤔 . any leads to this PR ?

PR #6850, opened 2023-02-02
Don't override but decorate the built-in encode function

emrysal:
## What does this PR do?

At the moment Google and other providers cannot login as the `token?.sub` value is not set for these providers (yet?) - this throws:

> if (!token || token.sub === undefined) throw new Error("Not valid token");

This simplifies the logic by decorating native, overriding maxAge only when applicable.


leog (approved):
Works, thanks Alex.

PR #6865, opened 2023-02-03
Feat/full custom select

### LINKED ISSUE #6593: [CAL-846] Improve multiselect component
Make multiselect component function more like this: [https://blueprintjs.com/docs/#select/multi-select2](https://blueprintjs.com/docs/#select/multi-select2)

**The main things are:**

* We want to show that an item is selected instead of removing it from the dropdown once selected
* The box needs to stay open
* We need search
* Toggle between showing all items (it currently expands onto a new line) or show a "+2 more items" at the end *(this isn't a necessity right now but would be ideal)*

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-846](https://linear.app/calcom/issue/CAL-846/improve-multiselect-component)</sub>

sean-brydon:
## What does this PR do?

Code probably needs a bit of cleanup but thought i'd put it out here to see if anyone has any suggestions/improvements first 

Full custom select component that matches 95% of WCAG a11y guidelines (I will work on improving in due course - react-select doesnt seem to even be 100%) 

This is not used within the app anywhere as of right now. Just been implemented in isolation in storybook. 

TODO (maybe in future PR) 
- [ ] Where does the focus go after you've select an item when it isn't a multiselect? It doesn't go back to the select itself at least, resulting in that I can not hit space/enter again to reopen the select. I think that's the desired behaviour. 

Fixes #6593 

 Loom Video: https://www.loom.com/share/436b048792514d4b8e6e8907e90bd5b5

## Type of change


packages/lib/hooks/useKeyPress.ts
```diff
@@ -0,0 +1,40 @@
...
+        placeHolderRef?.removeEventListener("keyup", upHandler);
+      };
+    } else {
+      window.addEventListener("keydown", downHandler);
+      window.addEventListener("keyup", upHandler);
+      // Remove event listeners on cleanup
+      return () => {
+        window.removeEventListener("keydown", downHandler);
+        window.removeEventListener("keyup", upHandler);
+      };
```
sean-brydon, line 35: Handle keypress on the red or when in a window 

Usefull if you only wanna do stuff when something is focused

packages/ui/components/form/selectimproved/components/Item.tsx
```diff
@@ -0,0 +1,71 @@
...
+  index?: number;
+  focused: boolean;
+}
+
+const Item: React.FC<ItemProps> = ({ item, index, focused }) => {
+  const { classNames, selectedItems, handleValueChange } = useSelectContext();
+  const isSelected =
+    (Array.isArray(selectedItems) && selectedItems?.some((selection) => selection.value === item.value)) ||
+    (!Array.isArray(selectedItems) && selectedItems?.value === item.value);
+
```
sean-brydon, line 20: Check if its selected based if we're using multiple or not

packages/ui/components/form/selectimproved/components/Item.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </div>
+
+          <div
+            className={cn(
+              "flex h-4 w-4 items-center justify-center rounded-[4px]  border ltr:mr-2 rtl:ml-2",
+              isSelected ? "bg-gray-800 text-gray-50" : "text-primary-600 border-gray-300 bg-gray-50"
+            )}>
+            {isSelected && <FiCheck className="h-3 w-3 text-current" />}
+          </div>
+        </li>
```
sean-brydon, line 65: Fake checkbox as the input item would highjack focus and also have weird on event handlers attached

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface OptionsProps<T extends Option> {
+  list: T[];
+  inputValue: string;
+  isMultiple: boolean;
+  selected: T | T[] | null;
+  searchBoxRef: React.RefObject<HTMLInputElement>;
+}
+
+type FlatternedOption = Option & { current: number; groupedIndex?: number };
```
sean-brydon, line 18: We flattern optional to include the count and also the index of the group it is attached to (If any)

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+const flatternOptions = (options: Option[], groupCount?: number): FlatternedOption[] => {
+  return options.reduce((acc, option, current) => {
+    if (option.options) {
+      return [...acc, ...flatternOptions(option.options, current + (groupCount || 0))];
```
sean-brydon, line 23: Recursivly call this is we have options -> current + group count keeps track of where we are in the inital list 

Needed to reference label and handle keyboard navigation correctly

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+function Options<T extends Option>({ list, inputValue, searchBoxRef }: OptionsProps<T>) {
+  const { classNames, handleValueChange } = useContext(SelectContext);
+  const [keyboardFocus, setKeyboardFocus] = useState(-1);
+  const downPress = useKeyPress("ArrowDown", searchBoxRef);
+  const upPress = useKeyPress("ArrowUp", searchBoxRef);
+  const enterPress = useKeyPress("Enter", searchBoxRef);
+
+  const flatternedList = useMemo(() => flatternOptions(list), [list]);
```
sean-brydon, line 36: flattern the list on the inital list passed in

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const totalOptionsLength = useMemo(() => {
+    return flatternedList.length;
+  }, [flatternedList]);
```
sean-brydon, line 40: Calculate the length of this - needed to get the min and max bounds of the array for keyboard navigation

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  useEffect(() => {
+    if (downPress) {
+      // Cycle to start of list if at end
+      setKeyboardFocus((prev) => (prev + 1) % totalOptionsLength);
```
sean-brydon, line 45: add one to previous take the %  of it 

e.g list size of 10 
 10+1 %  10 = 1

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }, [downPress, totalOptionsLength]);
+
+  useEffect(() => {
+    if (upPress) {
+      // Cycle to end of list if at start
+      setKeyboardFocus((prev) => (prev - 1 + totalOptionsLength) % totalOptionsLength);
```
sean-brydon, line 52: Same as above but in reverse

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  useEffect(() => {
+    if (enterPress) {
+      const item = filteredList[keyboardFocus];
+      if (!item || item.disabled) return;
+      handleValueChange(item);
+    }
+    // We don't want to re-run this effect when handleValueChange changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [enterPress, keyboardFocus, list]);
```
sean-brydon, line 64: Handle enter press - do nothing if we dont have an item or if item is disabled

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,109 @@
...
+        return (
+          <React.Fragment key={index}>
+            <div className="px-2.5">
+              {item.current === 0 && item.groupedIndex !== undefined && (
+                <Label>{list[item.groupedIndex].label}</Label>
+              )}
+              <Item item={item} index={index} focused={focused} />
+            </div>
+          </React.Fragment>
+        );
```
sean-brydon, line 103: Render all items - if its the first item in the list render a label before it

packages/ui/components/form/selectimproved/components/Select.tsx
```diff
@@ -0,0 +1,259 @@
...
+    list?: string;
+    listGroupLabel?: string;
+    listItem?: ({ isSelected }: { isSelected: boolean }) => string;
+    listDisabledItem?: string;
+    ChevronIcon?: ({ open }: { open: boolean }) => string;
+    searchContainer?: string;
+    searchBox?: string;
+    searchIcon?: string;
+    closeIcon?: string;
+  };
```
sean-brydon, line 42: WIP styles override if needed

packages/ui/components/form/selectimproved/components/Select.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  const onPressEnterOrSpace = useCallback(
+    (e: React.KeyboardEvent<HTMLDivElement>) => {
+      e.preventDefault();
+      if ((e.code === "Enter" || e.code === "Space") && !isDisabled) {
+        toggle();
+      }
+    },
+    [isDisabled, toggle]
+  );
```
sean-brydon, line 93: Handle open on space or enter if its not disabled

packages/ui/components/form/selectimproved/components/Select.tsx
```diff
@@ -0,0 +1,259 @@
...
+      }
+    },
+    [onChange, selectedItems]
+  );
+
+  const handleValueChange = useCallback(
+    (selected: Option) => {
+      function update() {
+        if (!isMultiple && !Array.isArray(selectedItems)) {
+          closeDropDown();
```
sean-brydon, line 112: Close when its a single array after selection

packages/ui/components/form/selectimproved/hooks/use-onclick-outside.ts
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect } from "react";
+
+export default function useOnClickOutside(
```
sean-brydon, line 3: TODO move to lib hooks

packages/ui/components/form/selectimproved/components/Options.tsx
JeroenReumkens: Great use of the modular operator!

packages/ui/components/form/selectimproved/components/Select.tsx
JeroenReumkens: Nice — I think this is a super simple approach of overriding the styles. Very easy to use, clear on how it works 👏

packages/ui/components/form/selectimproved/components/Select.tsx
JeroenReumkens: Can you add this + a because as a comment? I struggle the see why it should be closed in that case 😁

packages/ui/components/form/selectimproved/components/Select.tsx
```diff
@@ -0,0 +1,259 @@
...
+  );
+
+  const removeItem = useCallback(
+    (item: Option) => {
+      // remove the item from the selected items
+      if (Array.isArray(selectedItems)) {
+        const newSelectedItems = selectedItems.filter((selectedItem) => selectedItem.value !== item.value);
+        onChange(newSelectedItems);
+      } else {
+        onChange(null);
```
JeroenReumkens, line 102: Should it be `[]` if `isMultiple` is true?

packages/ui/components/form/selectimproved/components/SelectProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+  children: JSX.Element;
+  options: {
+    classNames?: ClassNames;
+  };
+}
+
+export const SelectContext = createContext<Store>({
+  selectedItems: null,
+  handleValueChange: (selected) => {
+    console.log("selected:", selected);
```
JeroenReumkens, line 23: Is this still (gonna be) used?

JeroenReumkens (commented):
Really great work so far @sean-brydon! 
I've left a few comments that I already spotted. On top of that I also tried to run it locally and encountered a few things that are worth looking into: 

1. When navigating through the options list by keyboard, and selecting an item with space, it also types a space in the input field — that triggers a new search, thus narrowing the items further down. After selecting 2 items, you've typed 2 spaces, and don't see any item in the dropdown anymore. I think the typing of space shouldn't be added to the search when selecting an item. 
2. I think we should use `scrollIntoView` on the option items when navigated to them. Since right now you can move the focus to an item that's not visible, which is confusing. 
3. Does a checkbox make sense when it isn't a multi select? 
4. Where does the focus go after you've select an item when it isn't a multiselect? It doesn't go back to the select itself at least, resulting in that I can not hit space/enter again to reopen the select. I think that's the desired behaviour. 
5. Can we hide the delete `x`  when no items are in the input? 

But, since selects are really hard to build, I'm absolutely amazed by the great work you've done. I think the could is really easy to understand for the amount of behaviour this have 👏 ❤️

sean-brydon:
> * Does a checkbox make sense when it isn't a multi select?

Probably not actually - will double check figma 

edit (Single should just be a check easy fix)

sean-brydon:
> 4\. Where does the focus go after you've select an item when it isn't a multiselect? It doesn't go back to the select itself at least, resulting in that I can not hit space/enter again to reopen the select. I think that's the desired behaviour.

Good spot - oh i love a11y :D

packages/lib/hooks/useKeyPress.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  // If released key is our target key then set to false
+  const upHandler = ({ key }: { key: string }): void => {
+    if (key === targetKey) {
+      setKeyPressed(false);
+    }
+  };
+  // Add event listeners
+  useEffect(() => {
+    if (ref && placeHolderRef) {
```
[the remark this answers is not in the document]
sean-brydon: I did try this approach but TS wasnt happy :D

packages/ui/components/form/selectimproved/components/SelectProvider.tsx
sean-brydon: No - this gets overriden on init of store - we just need it to always be defined

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,137 @@
...
+  list: T[];
+}) {
+  const focused = index === keyboardFocus;
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (ref.current && focused) {
+      ref.current.scrollIntoView({ behavior: "smooth" });
+    }
+  }, [ref, focused]);
```
sean-brydon, line 49: Since we're in a overflow-y-auto scrollIntoView scrolls the current item into view inside of that container :) 

Making it so you always can see the item you are hovering over

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,137 @@
...
+      }>
+      {filteredList?.map((item, index) => (
+        <FilteredItem
+          key={index}
+          item={item}
+          index={index}
+          keyboardFocus={keyboardFocus}
+          inputValue={inputValue}
+          list={list}
+        />
```
sean-brydon, line 131: Extracted so we can easily use ref behaviour

packages/ui/components/form/selectimproved/components/SearchInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+        classNames && classNames.searchContainer ? classNames.searchContainer : "relative py-1 px-2.5"
+      }>
+      <FiSearch
+        className={
+          classNames && classNames.searchIcon
+            ? classNames.searchIcon
+            : "absolute mt-2.5 ml-2 h-5 w-5 pb-0.5 text-gray-500"
+        }
+      />
+      <input
```
sean-brydon, line 34: This had different styles to our normal input

Wasnt worth building a new variant as this is probably one of a kind and didnt need the added RHF logic that comes with our input

packages/ui/components/form/selectimproved/components/Item.tsx
```diff
@@ -0,0 +1,80 @@
...
+const Item: React.FC<ItemProps> = ({ item, index, focused }) => {
+  const { classNames, selectedItems, handleValueChange } = useSelectContext();
+  const isMultiple = Array.isArray(selectedItems);
+  const isSelected =
+    (isMultiple && selectedItems?.some((selection) => selection.value === item.value)) ||
+    (!isMultiple && selectedItems?.value === item.value);
+
+  return (
+    <>
+      {item.disabled ? (
```
zomars, line 24:
```suggestion

  if (item.disabled) return ...

  return (
```

And like that.

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,137 @@
...
+  list: T[];
+  inputValue: string;
+  isMultiple: boolean;
+  selected: T | T[] | null;
+  searchBoxRef: React.RefObject<HTMLInputElement>;
+}
+
+type FlattenedOption = Option & { current: number; groupedIndex?: number };
+
+const flatternOptions = (options: Option[], groupCount?: number): FlattenedOption[] => {
```
zomars, line 20:
```suggestion
const flattenOptions = (options: Option[], groupCount?: number): FlattenedOption[] => {
```

packages/ui/components/form/selectimproved/components/SelectProvider.tsx
zomars, line 23:
```suggestion
```

packages/ui/components/form/selectimproved/components/Select.tsx
```diff
@@ -0,0 +1,260 @@
...
+  isMultiple?: boolean;
+  isClearable?: boolean;
+  isSearchable?: boolean;
+  isDisabled?: boolean;
+  loading?: boolean;
+  menuIsOpen?: boolean;
+  searchInputPlaceholder?: string;
+  noOptionsMessage?: string;
+  classNames?: {
+    menuButton?: ({ isDisabled }: { isDisabled: boolean }) => string;
```
zomars, line 27:
```suggestion
    menuButton?: (opts: { isDisabled: boolean }) => string;
```

zomars: Or args as well works

packages/ui/components/form/selectimproved/components/Select.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  useOnClickOutside(ref, () => {
+    closeDropDown();
+  });
+
+  useEffect(() => {
+    if (escapePressed) {
+      closeDropDown();
+    }
+  }, [escapePressed, closeDropDown]);
```
zomars, line 83: Maybe something similar as above 

```suggestion
   useOnEscKeypress(ref, () => {
     closeDropDown();
   });```

**Review by zomars — COMMENTED** [review]
So far so good @sean-brydon left some nits there and need to test locally before approving 🙏🏽

**Comment by JeroenReumkens** [comment]
> > 1. When navigating through the options list by keyboard, and selecting an item with space, it also types a space in the input field — that triggers a new search, thus narrowing the items further down. After selecting 2 items, you've typed 2 spaces, and don't see any item in the dropdown anymore. I think the typing of space shouldn't be added to the search when selecting an item.
> 
> Honestly not sure the right approach here - multi select none searchable should use space+enter to select items - however, with a search box its a bit of a gray area - and enter is seen as the most common way to select an item (You shouldnt ever hyjack an input from an input)

I also checked other comboboxes, and space does type in the field instead of select there as well. So I think using enter is a hard requirement instead of space. So no need to fix I feel.

File: packages/lib/hooks/useKeyPress.ts (hunk 18 above)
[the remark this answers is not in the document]
Comment (JeroenReumkens) [reply packages/lib/hooks/useKeyPress.ts:21]: Hmm, tried it right now myself as well, but indeed somehow the types of the `addEventListener` function change and starts trowing errors. Let's leave it like this then since the file is pretty small 😁

File: packages/ui/components/form/selectimproved/components/Options.tsx (hunk 19 above)
Comment (JeroenReumkens) [reply packages/ui/components/form/selectimproved/components/Options.tsx:49]: This works great now 🫶

**Comment by JeroenReumkens** [comment]
Great new commits @sean-brydon, very happy with what I see. One thing we might be missing still but I think is good if we add straight away is dark mode. Could you check with @Jaibles about this?

File: packages/ui/components/form/selectimproved/components/Options.tsx [hunk 26]
```diff
@@ -0,0 +1,126 @@
...
+  const flattenedList = useMemo(() => flattenOptions(list), [list]);
+
+  const totalOptionsLength = useMemo(() => {
+    return flattenedList.length;
+  }, [flattenedList]);
+
+  useKeyPress("ArrowDown", searchBoxRef, () => setKeyboardFocus((prev) => (prev + 1) % totalOptionsLength));
+  useKeyPress("ArrowUp", searchBoxRef, () =>
+    setKeyboardFocus((prev) => (prev - 1 + totalOptionsLength) % totalOptionsLength)
+  );
```
sean-brydon, line 75: Now this is cleaner ;) @zomars

packages/ui/components/form/selectimproved/components/Options.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  useEffect(() => {
+    if (enterPress) {
+      const item = filteredList[keyboardFocus];
+      if (!item || item.disabled) return;
+      handleValueChange(item);
+    }
+    // We don't want to re-run this effect when handleValueChange changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [enterPress, keyboardFocus, list]);
```
sean-brydon, line 85: Had to leave this as the old hook as we dont want this to run when handleValueChanges

PeerRich:
> Great new commits @sean-brydon, very happy with what I see. One thing we might be missing still but I think is good if we add straight away is dark mode. Could you check with @Jaibles about this?

~~not sure, i think the public booking page will continue to use react-select-timezone right?~~

nevermind. routing forms have selects and darkmdoe

sean-brydon:
> > Great new commits @sean-brydon, very happy with what I see. One thing we might be missing still but I think is good if we add straight away is dark mode. Could you check with @Jaibles about this?
> 
> ~not sure, i think the public booking page will continue to use react-select-timezone right?~
> 
> nevermind. routing forms have selects and darkmdoe

Will pick this up tomorrow should be a quick win to implement :)

sean-brydon:
CleanShot 2023-02-08 at 10 08 54@2x
@PeerRich @JeroenReumkens Darkmode been commited :)

JeroenReumkens:
Great work @sean-brydon! Just tested again and noticed two other small issues; 

1. I think the disabled states for section titles got lost when implementing dark mode? I also see that the section title has a checkbox even though you can not check that (or should it work as a "select all below"?)
2. When selecting options in a multi select, I see the dropdown itself also grows in size. Somehow it feels a bit annoying. What do you think? 🤔

sean-brydon:
@JeroenReumkens I have fixed the disabled state 

I can't replicate 
> When selecting options in a multi select, I see the dropdown itself also grows in size. Somehow it feels a bit annoying. What do you think? 🤔

At all - you got a demo of that happening?

JeroenReumkens (approved):
Great work @sean-brydon , approved!

PR #6873, opened 2023-02-03
Fix duplicating even types with active workflows

### LINKED ISSUE #6722: [CAL-931] Duplicating event-type with workflow duplicates a workflow that cannot be deactivated or seen in the general workflow view
When I duplicate an even type, the attached/activated workflow is also duplicated and shows on the workflow tab of the event. However, the workflow cannot be deactivated nor seen in the general workflow view. No I am unsure whether it is triggered on each booking as I cannot edit the workflow.

*submitted via feedback in the app*

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-931](https://linear.app/calcom/issue/CAL-931/duplicating-event-type-with-workflow-duplicates-a-workflow-that-cannot)</sub>

CarinaWolli:
## What does this PR do?

Fixes duplicating event types that have active workflows. 

Fixes #6722

## Type of change

## How should this be tested?

- Create a workflow and set it active for an event type
- Duplicate this event type
- Go to the event type settings (of the duplicated event type) and check in the workflow tab if the workflow is active

### DIFF
--- a/packages/trpc/server/routers/viewer/eventTypes.tsx
+++ b/packages/trpc/server/routers/viewer/eventTypes.tsx
@@ -770,20 +770,12 @@ export const eventTypesRouter = router({
       }
 
       if (workflows.length > 0) {
-        const workflowIds = workflows.map((workflow) => {
-          return { id: workflow.workflowId };
+        const relationCreateData = workflows.map((workflow) => {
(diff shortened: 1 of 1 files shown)


emrysal (approved):
Maybe it's also an idea to put a unique key on [eventTypeId, workflowId]?

PR #6874, opened 2023-02-03
Allow account linking for Google and SAML providers

deepakprabhakara:
## What does this PR do?

It is safe to allow account linking for Google and SAML providers since the email is verified in both cases. This functionality for self-hosted instance seems to have stopped working with a recent next-auth upgrade. This PR fixes this issue as well.

## Type of change

## How should this be tested?

- [x] Ensure a local instance with NEXT_PUBLIC_HOSTED_CAL_FEATURES not set so that it is treated as a self-hosted instance. Sign up with username/pass for a gmail address you have. Now sign back in using Google login, you should be allowed to login to the account you created earlier.
- [x] Repeat above with SAML instead of Google.
- [x] Repeat above with NEXT_PUBLIC_HOSTED_CAL_FEATURES=1

### DIFF
--- a/apps/web/pages/api/auth/[...nextauth].tsx
+++ b/apps/web/pages/api/auth/[...nextauth].tsx
@@ -164,6 +164,7 @@ if (IS_GOOGLE_LOGIN_ENABLED) {
     GoogleProvider({
       clientId: GOOGLE_CLIENT_ID,
       clientSecret: GOOGLE_CLIENT_SECRET,
+      allowDangerousEmailAccountLinking: !hostedCal,
     })
   );
 }
@@ -202,6 +203,7 @@ if (isSAMLLoginEnabled) {
       clientId: "dummy",
       clientSecret: "dummy",
     },
+    allowDangerousEmailAccountLinking: !hostedCal,
   });
 }
 
--- a/apps/web/pages/api/auth/[...nextauth].tsx
+++ b/apps/web/pages/api/auth/[...nextauth].tsx
@@ -164,7 +164,7 @@ if (IS_GOOGLE_LOGIN_ENABLED) {
     GoogleProvider({
       clientId: GOOGLE_CLIENT_ID,
       clientSecret: GOOGLE_CLIENT_SECRET,
-      allowDangerousEmailAccountLinking: !hostedCal,
+      allowDangerousEmailAccountLinking: true,
     })
   );
 }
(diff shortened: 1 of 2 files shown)


emrysal:
> I want to call out that you can also use allowDangerousEmailAccountLinking: true - while the config might look dangerous, most well-known providers are already secure and will only allow using verified emails.
https://github.com/nextauthjs/next-auth/discussions/2808#discussioncomment-4486932

With this in mind I think we can enable Google account linking across the board, I believe we have some issues open for this @PeerRich (I didn't know this used to work in previous next auth versions)

apps/web/pages/api/auth/[...nextauth].tsx
```diff
@@ -164,6 +164,7 @@ if (IS_GOOGLE_LOGIN_ENABLED) {
     GoogleProvider({
       clientId: GOOGLE_CLIENT_ID,
       clientSecret: GOOGLE_CLIENT_SECRET,
+      allowDangerousEmailAccountLinking: !hostedCal,
```
Comment (emrysal) [inline apps/web/pages/api/auth/[...nextauth].tsx:167]:
```suggestion
      allowDangerousEmailAccountLinking: true,
```

We can trust Google to work with verified email addresses 🙏

Comment (PeerRich) [reply apps/web/pages/api/auth/[...nextauth].tsx:167]: i think that should be fine @deepakprabhakara no?

Comment (deepakprabhakara) [reply apps/web/pages/api/auth/[...nextauth].tsx:167]: Yes, this should be fine but let me just double check.

Comment (deepakprabhakara) [reply apps/web/pages/api/auth/[...nextauth].tsx:167]: This works well and is only limited to the provider, so we can enable it for SAML as well. I'll push through the new changes

leog (dismissed):
Code looks good and it works. I just wanted to be sure we all agree on the following behavior that will happen accepting these changes:
 - If a user signups using an email that does not belong to that person and is powered by Google, then if the real person with that email signup using Google Sign-in, both will be able to use the account, one by email/password and the other by Google Sign-in

That could happen because we don't ask for email verification, and as Next-Auth docs [here](https://next-auth.js.org/configuration/providers/oauth#allowdangerousemailaccountlinking-option) and [here](https://next-auth.js.org/faq#security) explains, `allowDangerousEmailAccountLinking` could be dangerous (hence the name of the flag) when not all providers are trusted in terms of email validation.

deepakprabhakara:
@leog That's correct, this seems like an acceptable middle ground until the email verification is enforced.

leog:
@PeerRich as discussed, should we try to invalidate the password to avoid having a user login without Google? perhaps showing an error that they should login using Google or use the forget password to reset it?

PeerRich:
i think its a good idea to invalidate password once someone clicks "Sign in with Google"

zomars (approved):

zomars:
Still better that what be have now. Merging and let's follow up with the invalidation stuff.

PR #6677, opened 2023-01-24
feat/payment-service-6438-cal-767

alannnc:
## What does this PR do?

Abstract all stripe payment calls to a new PaymentService.
NOTE: should be merge after [6877]https://github.com/calcom/cal.com/pull/6877

Fixes #6438 

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Chore (refactoring code, technical debt, workflow improvements)
- [X] This change requires a documentation update

## How should this be tested?

So far it should allow payments to keep working without problem, but using new PaymentService

- [ ] Add Stripe for an eventType
- [ ] Configure payment for eventType
- [ ] Book previously eventType with payment required.
- [ ] Complete payment

## Checklist

- I haven't performed a self-review of my own code and corrected any misspellings
- I haven't added tests that prove my fix is effective or that my feature works

PR #6877, opened 2023-02-04
migration/remove-payment-type-enum-use-appId-instead

alannnc:
## What does this PR do?
- Run migration required for for [6677](https://github.com/calcom/cal.com/pull/6677)
- Removes payment Type enum as uses app Id instead

NOTE: should be merge before 6677 and along with it


alannnc:
Closing as running migration only will break deploy on staging.

PR #6880, opened 2023-02-04
feat: get country code from ip geolocation

### LINKED ISSUE #6621: [CAL-862] Assume international format when provided a regional phone number

Users sometimes provide a number in regional formatting, the phone number input fields expect an internationally formatted number.

image

There are a number of ways an international number can be inferred from a regional number, perhaps libphonenumber.

Alternatively a default country code and stripping the leading 0 from UK numbers as a soft solution for this specific use case.

Accepting that the recorded value isn't accurate & cleaning up the data myself.

<sub>[CAL-862](https://linear.app/calcom/issue/CAL-862/assume-international-format-when-provided-a-regional-phone-number)</sub>

### LINKED ISSUE #7049: [CAL-1043] Phone custom field automatically update country by IP
Is it possible to add an automatic IP recognition for phone custom field?

When this field is available, the person needs to choose the country and then type the number.

In mobile it may be not intuitive, getting the country by IP would be very good.

Eg: someone from Brazil's IP gets automatically Brazil selected with no need to choose the country.

<sub>[CAL-1043](https://linear.app/calcom/issue/CAL-1043/phone-custom-field-automatically-update-country-by-ip)</sub>

### LINKED ISSUE #7373: Set default country for phone number
Is it possible that you include the option to set the default country for the phone number beforehand? For example, that Germany (+49) is always selected by default? 90% of my customers come from Germany and if this is not preset, it quickly comes to confusion.

Bildschirm­foto 2023-02-24 um 15 54 00

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/6621

- set default country code in react-phone-number input from geolocation api for better user experience

TODO
- [x] test on preview deployment

https://user-images.githubusercontent.com/53316345/216760540-18b048fb-933d-4889-be8e-92ff01b9d306.mov

## Type of change

### DIFF
--- a/apps/web/pages/api/countrycode.ts
+++ b/apps/web/pages/api/countrycode.ts
@@ -1,0 +1,20 @@
+import type { NextRequest } from "next/server";
+
+export const config = {
+  runtime: "experimental-edge",
+};
+
+export default async function handler(req: NextRequest) {
+  const countryCode = req.headers.get("x-vercel-ip-city") ?? "";
+  return new Response(
+    JSON.stringify({
+      countryCode,
+    }),
+    {
+      status: 200,
+      headers: {
+        "content-type": "application/json",
+      },
+    }
+  );
+}
--- a/apps/web/pages/api/countrycode.ts
+++ b/apps/web/pages/api/countrycode.ts
@@ -5,7 +5,7 @@ export const config = {
 };
 
 export default async function handler(req: NextRequest) {
-  const countryCode = req.headers.get("x-vercel-ip-city") ?? "";
+  const countryCode = req.headers.get("x-vercel-ip-country") ?? "";
   return new Response(
     JSON.stringify({
       countryCode,
--- a/packages/ui/form/PhoneInput.tsx
+++ b/packages/ui/form/PhoneInput.tsx
@@ -21,17 +21,7 @@ function PhoneInput<FormValues>({
   onChange,
   ...rest
 }: PhoneInputProps<FormValues>) {
-  const [defaultCountry, setDefaultCountry] = useState("US");
-  useEffect(() => {
-    fetch("/api/countrycode")
-      .then((res) => res.json())
-      .then((res) => {
-        if (isSupportedCountry(res.countryCode)) setDefaultCountry(res.countryCode);
-      })
-      .catch((err) => {
-        console.log(err);
-      });
-  }, []);
+  const defaultCountry = useDefaultCountry();
   return (
     <BasePhoneInput
       {...rest}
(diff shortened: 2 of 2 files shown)


emrysal:
Alternative approach should be taken similar to #6849 to prevent the external API call instead.

emrysal (changes requested):
Alternative approach should be taken similar to https://github.com/calcom/cal.com/pull/6849 to prevent the external API call instead.

Udit-takkar:
@emrysal  is x-vercel-ip-city null when tested on localhost ?  we will have to test on preview deployment to make sure it is working fine

emrysal:
Checked with @G3root - yeah this would be the case for localhost - @zomars how do we do this nicely for self hosted?

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,10 +20,22 @@ function PhoneInput<FormValues>({
   onChange,
   ...rest
 }: PhoneInputProps<FormValues>) {
+  const [defaultCountry, setDefaultCountry] = useState("US");
```
emrysal, line 23: Make this a custom hook, `useDefaultCountryCode` - which should encapsulate the useEffect fetch logic.

In the pages/api call you should clearly specify how this will work for self-hosting 🙏 - e.g. on self hosting you return a `501 Not Implemented` error which you can then gracefully handle from the frontend.

emrysal (changes requested):

apps/web/pages/api/countrycode.ts
```diff
@@ -0,0 +1,20 @@
+import type { NextRequest } from "next/server";
+
+export const config = {
+  runtime: "experimental-edge",
+};
+
+export default async function handler(req: NextRequest) {
+  const countryCode = req.headers.get("x-vercel-ip-country") ?? "";
```
AaronPresley, line 8: I would strongly recommend using `x-vercel-ip-country-region` here ([docs](https://vercel.com/docs/concepts/edge-network/headers#x-vercel-ip-country-region)), as it's a closer match to the [country code standards](https://en.wikipedia.org/wiki/ISO_3166-2) used throughout the rest of our globalized codebase.

AaronPresley: Out of curiosity, I looked into how `libphonenumber-js` is mapping a country's code to a given format.

[This line](libphonenumber-js) confirms they base their region codes off of CLDR data, which gives me more confidence that `x-vercel-ip-country-region` is the best way to go here.

Udit-takkar: @AaronPresley BasePhoneInput from react-phone-number-input requires 2 character country code which is returned from x-vercel-ip-country

Udit-takkar:
@zomars 
this returns timezone dayjs.tz.guess() // America/Chicago and we need defaultCountry in this format https://www.npmjs.com/package/react-phone-number-input#country-code

Udit-takkar:
cc: @emrysal

emrysal:
This could help:
image

emrysal (approved):
As discussed need further improvement for self-hosting but it's a good place to start!

PR #6898, opened 2023-02-06
When unchecking the common schedule, schedule should be nulled

emrysal:
## What does this PR do?

Situation before was that untoggling this did leave the `schedule` intact, which later on (as the availability logic does not take into account `metadata.config.useHostSchedulesForTeamEvent`) results in an unrecoverable situation of a forced common schedule.

### DIFF
--- a/apps/web/components/eventtype/AvailabilityTab.tsx
+++ b/apps/web/components/eventtype/AvailabilityTab.tsx
@@ -99,18 +99,77 @@ const format = (date: Date, hour12: boolean) =>
     new Date(dayjs.utc(date).format("YYYY-MM-DDTHH:mm:ss"))
   );
 
-const EventTypeSchedule = () => {
-  const me = useMeQuery();
-  const timeFormat = me?.data?.timeFormat;
+const EventTypeScheduleDetails = () => {
+  const { data: loggedInUser } = useMeQuery();
+  const timeFormat = loggedInUser?.timeFormat;
   const { t, i18n } = useLocale();
   const { watch } = useFormContext<FormValues>();
 
   const scheduleId = watch("schedule");
-  const { isLoading, data: schedule } = trpc.viewer.availability.schedule.get.useQuery({ scheduleId });
+  const { isLoading, data: schedule } = trpc.viewer.availability.schedule.get.useQuery(
+    { scheduleId: scheduleId || loggedInUser?.defaultScheduleId || undefined },
+    { enabled: !!scheduleId || !!loggedInUser?.defaultScheduleId }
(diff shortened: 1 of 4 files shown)


apps/web/components/eventtype/AvailabilityTab.tsx
```diff
@@ -131,67 +190,14 @@ const EventTypeSchedule = () => {
...
-            href={`/availability/${schedule?.schedule.id}`}
-            color="minimal"
-            EndIcon={FiExternalLink}
-            target="_blank"
-            rel="noopener noreferrer">
-            {t("edit_availability")}
-          </Button>
-        </div>
-      </div>
+      <EventTypeScheduleDetails />
```
emrysal, line 193: Adding this as a separate component prevents the AvailabilitySelect from being re-rendered.

apps/web/components/eventtype/AvailabilityTab.tsx
```diff
@@ -200,6 +206,11 @@ export const AvailabilityTab = ({ isTeamEvent }: { isTeamEvent: boolean }) => {
           checked={!value}
           onCheckedChange={(checked) => {
             onChange(!checked);
+            if (checked) {
+              resetField("schedule");
```
emrysal, line 210: Go back to default value on re-check (or initial check)

packages/trpc/server/routers/viewer/eventTypes.tsx
```diff
@@ -564,6 +564,12 @@ export const eventTypesRouter = router({
         };
       }
     }
+    // allows unsetting a schedule through { schedule: null, ... }
+    else if (null === schedule) {
```
hariombalhara, line 568: I think in addition to explicitly asking schedule to be set as null, we should set it to null automatically on receiving false value of `useHostSchedulesForTeamEvent`. That should remove the frontend only check for explicitly setting null when `useHostSchedulesForTeamEvent` is false.

Not sure if it has an API call, if yes it should be handled there as well.

hariombalhara (approved):
Looks good to me apart from the change I requested.

PeerRich:
save to merge or should I wait @emrysal ?

packages/trpc/server/routers/viewer/eventTypes.tsx
emrysal: I had a look but it adds a bit of complexity, ideally we wouldn't have `useHostSchedulesForTeamEvent` at all but this is necessary, without it we can't default to a schedule user schedule with this feature enabled.

PR #6906, opened 2023-02-06
fix: admin password hint text

### LINKED ISSUE #6842: [CAL-981] Admin password change has incoherent/obsolete information
There are two issues with this:

1. The alert says minimum of 15 characters are required, but the description in the body says 7 characters, thus the information being invalid
2. Once the password is changed, the alert persists despite now the user having minimum 15 character password

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-981](https://linear.app/calcom/issue/CAL-981/admin-password-change-has-incoherentobsolete-information)</sub>

G3root:
## What does this PR do?

Fixes #6842 

to reproduce locally:
- [ ] create a admin user with a password lesser than the required length of 15
- [ ] try changing the password

issue:

[Screencast 2023-02-07 08:21:45.webm](https://user-images.githubusercontent.com/84864519/217136491-9391aef6-90e1-4c3e-b8a2-6b55110d37a6.webm)

after:

[Screencast 2023-02-07 07:25:36.webm](https://user-images.githubusercontent.com/84864519/217136729-5c90a71b-bab0-435d-ac50-7e242060c6b1.webm)

changes:

- [x] hide the banner after password is changed 
- [x] fix the text in password hint 

## Type of change


apps/web/public/static/locales/en/common.json
```diff
@@ -431,7 +431,7 @@
   "password_hint_min": "Minimum 8 characters long",
   "password_hint_admin_min": "Minimum 15 characters long",
   "password_hint_num": "Contain at least 1 number",
-  "invalid_password_hint": "The password must be a minimum of 7 characters long containing at least one number and have a mixture of uppercase and lowercase letters",
+  "invalid_password_hint": "The password must be a minimum of 15 characters long containing at least one number and have a mixture of uppercase and lowercase letters",
```
Udit-takkar, line 434: the 15 character minimum limit is only for the admin role. for other 7 characters are still allowed. maybe just check if the user role is admin then show the 15 character password hint else 7 chars.

Udit-takkar (changes requested):
can we improve the UX here ? if new password does not meet the requirements then message/input box should be in red.

Screenshot 2023-02-07 at 8 32 06 AM

apps/web/public/static/locales/en/common.json
G3root: i'll update those changes 🙌🏼 . do we still have the number, uppercase and uppercase requirement for a normal user ?

G3root: i'm planning on to convert the hardcoded number into template variable.

Udit-takkar: check lib/auth

export function validPassword(password: string) {
  if (password.length < 7) return false;

  if (!/[A-Z]/.test(password) || !/[a-z]/.test(password)) return false;

  if (!/\d+/.test(password)) return false;

  return true;
}

leog:
> can we improve the UX here ? if new password does not meet the requirements then message/input box should be in red.

I agree with @Udit-takkar, perhaps you can use the same hints shown when signing-up in website here: https://cal.com/signup. It is based on zod and uses our custom inputs to digest them. Happy to help there, I implemented it.

G3root:
pushed some changes:

- [x] used alert component for errors returned by the server.
- [x] added client side validation for new password and error messages
- [x] fix some inconsitencies in i18n


[Screencast 2023-02-07 22:52:25.webm](https://user-images.githubusercontent.com/84864519/217320489-bb7a2639-8dbc-4561-9567-a59a523944bc.webm)

cc: @Udit-takkar @leog

Udit-takkar (changes requested):
some issue with password validation(this should be a valid password). This is also the reason why e2e tests are failing. everything else is fine 👍


Screenshot 2023-02-08 at 12 49 12 PM

zomars (approved):
Nice work @G3root 🙏🏽 

Ship it

G3root:
@Udit-takkar  pushed a fix for the regex

[Screencast 2023-02-09 07:20:41.webm](https://user-images.githubusercontent.com/84864519/217696334-e4f882a2-b567-402f-a1d2-fc0f45d337f1.webm)

Udit-takkar (approved):

PR #6908, opened 2023-02-06
feature/alert-no-availability-selected

alannnc:
## What does this PR do?

Displays warning text on availability page and event-types

Fixes # (issue)
image
image

https://www.loom.com/share/325390112b074ce489d66b0a95a83de5

## Type of change

## How should this be tested?

- [ ] Unselect every option in default availability
- [ ] See alert on availability row and event types as a badge

## Checklist


PeerRich:
i dont think this is a good solution. quite a workaround. we should try to prevent people from saving this or find an alternative way what happens when someone has no availabilty.

PeerRich (changes requested):
i dont think this is a good solution. quite a workaround. we should try to prevent people from saving this or find an alternative way what happens when someone has no availabilty.

alannnc:
Closing as suggested, while we find another way to solve this.

PR #6909, opened 2023-02-06
Feature: Reserve slots currently being booked

alannnc:
## What does this PR do?
 
~~Uses cache as a reservation slot system. If a user visits book page with specific slot it should save for 30secs that slot if another user request for slots.~~ 

- Saves data in the DB when a booking is being booking to prevent other from selecting it

Fixes # (issue)
https://www.loom.com/share/1c84905827b347b2bc7559acad49bbf9

## Type of change

## How should this be tested?

- [ ] Navigate to book page after selecting time slot from event type.
- [ ] Use another window/navigator to see if previous selected slot its still available.

## Checklist


apps/web/pages/[user]/book.tsx
```diff
@@ -187,6 +188,15 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
         : eventType.recurringEvent.count)) ||
     null;
 
+  // We should mark on cache this slot as unavailable
+  const date = context.query.date as string;
+  const dateUTC = new Date(date).toISOString();
+  const bookingEventName = `${eventType.slug}-${eventType.id}`;
```
Comment (alannnc) [inline apps/web/pages/[user]/book.tsx:194]: Definitely we should improve bookingEventName to be unique for certain Slot

roae (dismissed):
I think we can fall into the same problem, when the 30 seconds that the cache lasts have passed, that slot will be released and it is very likely that the user is still in the booking form. We cannot depend on that cache in memory, because it is unreliable by the nature of lambda functions.

What do you think if we add a table that stores the "booking sessions", using the same logic, but using this table to identify the occupied slots?

PeerRich:
excited for this

.env.example
```diff
@@ -164,3 +164,5 @@ CSP_POLICY=
 
 # Vercel Edge Config
 EDGE_CONFIG=
+
+NEXT_PUBLIC_MINUTES_TO_BOOK=2 // Minutes
```
roae, line 168: This could be part of the event config.

emrysal:
`BookingReservation`?

.env.example
emrysal: Also, 5 is a better default, no user is going to finish in 2.

packages/trpc/server/createContext.ts
```diff
@@ -106,6 +106,7 @@ type CreateInnerContextOptions = {
   locale: string;
   user: Awaited<ReturnType<typeof getUserFromSession>>;
   i18n: Awaited<ReturnType<typeof serverSideTranslations>>;
+  ip: string;
```
roae, line 109: Add ip to identify the user

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -1070,8 +1080,14 @@ function ErrorMessage({ error }: { error: unknown }) {
...
-            {rescheduleUid ? t("reschedule_fail") : t("booking_fail")}{" "}
-            {error instanceof HttpError || error instanceof Error ? t(error.message) : "Unknown error"}
+            {typeof error === "string" ? (
+              error
+            ) : (
+              <>
+                {rescheduleUid ? t("reschedule_fail") : t("booking_fail")}{" "}
+                {error instanceof HttpError || error instanceof Error ? t(error.message) : "Unknown error"}
+              </>
+            )}
```
roae, line 1090: When the time to book is over it's show this message to user. I think, the ux can be improved


imagen

packages/trpc/server/createContext.ts
emrysal: If we can avoid this please do, what if we store a localhost reference to the booking in question? Reason being is that an ip is PII which means we'll need to do due diligence compliance wise.

.env.example
alannnc, line 168:
```suggestion
NEXT_PUBLIC_MINUTES_TO_BOOK=5 // Minutes
```

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -144,6 +147,11 @@ const BookingPage = ({
         collectPageParameters("/book", { isTeamBooking: document.URL.includes("team/") })
       );
     }
+    selectSlotMutation.mutate({ eventTypeId: eventType.id, slotUtcDate: dayjs(queryDate).utc().format() });
+    setTimeout(() => setTimeOver(true), parseInt(process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "2") * 60 * 1000);
```
alannnc, line 151:
```suggestion
    setTimeout(() => setTimeOver(true), parseInt(process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "5") * 60 * 1000);
```

packages/trpc/server/createContext.ts
zomars: Can we use the next-auth cookie for this?

packages/trpc/server/routers/viewer/slots.ts
```diff
@@ -44,6 +47,17 @@ const getScheduleSchema = z
     "Either usernameList or eventTypeId should be filled in."
   );
 
+const markSelectedSlotSchema = z
+  .object({
+    eventTypeId: z.number().int(),
+    // startTime ISOString
```
alannnc, line 53:
```suggestion
```

packages/trpc/server/routers/viewer/slots.ts
```diff
@@ -106,15 +120,49 @@ export const slotsRouter = router({
...
+    .mutation(({ ctx, input }) => markSelectedSlot(ctx, input)),
+  removeSelectedSlotMark: publicProcedure.mutation(({ ctx }) => removeSelectedSlotMark(ctx)),
 });
 
+async function markSelectedSlot(ctx: Context, input: z.infer<typeof markSelectedSlotSchema>) {
+  const { prisma, ip } = ctx;
+  const { slotUtcDate, eventTypeId } = input;
+  const releaseAt = dayjs
+    .utc()
+    .add(parseInt(process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "2"), "minutes")
```
alannnc, line 134:
```suggestion
    .add(parseInt(process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "5"), "minutes")
```

packages/trpc/server/createContext.ts
PeerRich: yes please dont save the IP

PeerRich: great catch @emrysal !

apps/web/components/booking/pages/BookingPage.tsx
PeerRich: we need a better error message even if this is quite rare

PeerRich: @emrysal can you write up something?

alannnc: Can we add an action on this warning that recalls reserveSlot?

PeerRich: is there any way we can prevent this error from ever happening? like smart re-routing back to the previous screen with a toast or someting? i know this error is already much more rare than the previous one

apps/web/public/static/locales/en/common.json
```diff
@@ -1619,5 +1619,6 @@
   "select_a_router": "Select a router",
   "add_a_new_route": "Add a new Route",
   "no_responses_yet": "No responses yet",
-  "this_will_be_the_placeholder": "This will be the placeholder"
+  "this_will_be_the_placeholder": "This will be the placeholder",
+  "time_to_booking_over": "This slot has already been released, it may be that the reservation does not materialize"
```
zomars, line 1623:
```suggestion
  "time_to_booking_over": "This slot has already been released, now it can be reserved by someone else."
```

packages/trpc/server/routers/viewer/slots.ts
alannnc: Not required startTime for this

apps/web/components/booking/pages/BookingPage.tsx
roae: We can increase the time automatically, if the user continues on this page.

packages/trpc/server/routers/viewer/slots.ts
```diff
@@ -106,15 +122,54 @@ export const slotsRouter = router({
...
+    create: {
+      eventTypeId,
+      slotUtcDate,
+      uid,
+      releaseAt,
+    },
+  });
+  res?.setHeader("Set-Cookie", serialize("uid", uid, { path: "/", sameSite: "lax" }));
+  return;
+}
```
roae, line 154: create a cookie with a unique id to differentiate public users

packages/trpc/server/routers/viewer/slots.ts
```diff
@@ -320,10 +375,31 @@ export async function getSchedule(input: z.infer<typeof getScheduleSchema>, ctx:
...
+  const slotBusy =
+    selectedSlots?.length > 0
+      ? selectedSlots.map((item) => {
+          ids.push(item.id);
+          return item.slotUtcDate.toISOString();
+        })
+      : [];
+  await prisma.selectedSlots.deleteMany({
+    where: { eventTypeId: { equals: eventType.id }, id: { notIn: ids } },
+  });
```
roae, line 396: Realease abandoned slots

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -144,6 +150,15 @@ const BookingPage = ({
...
     }
+    reserveSlot();
+    const interval = setInterval(
+      reserveSlot,
+      parseInt(process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "5") * 60 * 1000 - 2000
+    );
+    return () => {
+      clearInterval(interval);
+      releaseSlotMutation.mutate();
+    };
```
roae, line 161: this is to keep the slot accupied when the user stays in the booking form and release it when he leaves.

PeerRich:
@roae have your requested changes been addressed?

roae:
> @roae have your requested changes been addressed?

Yes, I made them because @alannnc  was busy with something else already.

roae:
I'm getting this error on reschedule, 

I tried to change the time from 5:00 to 5:30 but I got the error "Attendee not found", something strange too, the time of the previous reservation says 2:00


imagen

apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -41,6 +42,7 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
   seatsPerTimeSlot,
   ethSignature,
 }) => {
+  const mutation = trpc.viewer.public.slots.markSelectedSlot.useMutation();
```
hariombalhara, line 45: nit: We can be consistent and name it `reserveSlot` only

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -221,6 +227,15 @@ const BookingPage = ({
         collectPageParameters("/book", { isTeamBooking: document.URL.includes("team/") })
       );
     }
+    reserveSlot();
+    const interval = setInterval(
+      reserveSlot,
+      parseInt(process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "5") * 60 * 1000 - 2000
```
hariombalhara, line 233: We can abstract it into constants.ts like other env variables so that the default value is consistently set.

apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -61,6 +63,10 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
     [isMobile]
   );
 
+  const reserveSlot = (slot: Slot) => {
+    mutation.mutate({ slotUtcDate: slot.time, eventTypeId });
```
hariombalhara, line 67: I think we should reserve a user's timeslot instead of reserving it per event, as timeslots are shared across event types.

For Team Events with Round Robin scheduling, a slot can be booked multiple times and it can prevent booking a slot when in fact it is available.
Similarly, for events with multiple seats, I believe we need to consider the number of seats

Might be taken care of in a follow-up too.

zomars: Per event type is fine for now. this might open up a can of worms that we don't know yet.

PeerRich: do you think per event type is better vs for all links? i think there is a potential for more double bookings if its per event-type only

packages/trpc/server/routers/viewer/slots.ts
```diff
@@ -326,10 +368,31 @@ export async function getSchedule(input: z.infer<typeof getScheduleSchema>, ctx:
...
+    select: { id: true, slotUtcDate: true },
+  });
+  const ids: number[] = [];
+  const slotBusy =
+    selectedSlots?.length > 0
+      ? selectedSlots.map((item) => {
+          ids.push(item.id);
+          return item.slotUtcDate.toISOString();
+        })
+      : [];
```
zomars, line 386: No need to check length for an empty array. and findMany will always return an array. Not undefined.

```suggestion
  const slotBusy = selectedSlots.map((item) => {
    ids.push(item.id);
    return item.slotUtcDate.toISOString();
  });
```

zomars (commented):
Making some updates

zomars:
Tried to fix some unit tests but I think this needs updating. @roae can you verify this is related to the DST bug? 
image

apps/web/components/booking/AvailableTimes.tsx
roae: I think @hariombalhara is right, the timeslot must be reserved, it doesn't matter the event.

I propose add the userId to the table and a boolean field to indicate that this revervation is an event's seat, this way we can count the reserve seats and added to the real occupied seats.

What is your opinion about this ?@zomars @PeerRich @alannnc @emrysal

roae:
> Tried to fix some unit tests but I think this needs updating. @roae can you verify this is related to the DST bug? image

@zomars I think it is because of what @hariombalhara told us above.

> I think we should reserve a user's timeslot instead of reserving it per event, as timeslots are shared across event types.
> 
> For Team Events with Round Robin scheduling, a slot can be booked multiple times and it can prevent booking a slot when in fact it is available. Similarly, for events with multiple seats, I believe we need to consider the number of seats
> 
> Might be taken care of in a follow-up too.

apps/web/components/booking/AvailableTimes.tsx
zomars: Agreed

alannnc:
Should be good to go @roae, I can't approve it myself.

roae (approved):
Thank you @alannnc good team work 🙏

PR #6911, opened 2023-02-07
Collective events are now considered from getBusyTimes booking query

### LINKED ISSUE #6374: [CAL-736] Collective events are ignored from getBusyTimes booking query
### Issue Summary

In getBusyTimes.ts - conflicting bookings are found based on userId, this does not work on a Collective event as there is no single user assigned, so other event types remain bookable when there is no calendar connected.

```
const busyTimes: EventBusyDetails[] = await prisma.booking
    .findMany({
      where: {
        userId,
        startTime: { gte: new Date(startTime) },
        endTime: { lte: new Date(endTime) },
        status: {
          in: [BookingStatus.ACCEPTED],
        },
      },
      select: {
        id: true,
        startTime: true,
        endTime: true,
        title: true,
        eventType: {
          select: {
            id: true,
            afterEventBuffer: true,
            beforeEventBuffer: true,
          },
        },
      },
    })
```

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-736](https://linear.app/calcom/issue/CAL-736/collective-events-are-ignored-from-getbusytimes-booking-query)</sub>

AaronPresley:
## What does this PR do?

Fixes #6374

This PR updates the `getBusyTimes` query to consider a user busy within a given time period if:
- There is a booking owned by this user (previous functionality)
- There is a booking with an `eventType.id` in which this user is a host

## Type of change

## How should this be tested?

1. Clean your DB by running `yarn db-nuke` from within `packages/prisma`
2. Boot the app from the root with `yarn dx`
3. Login as `teampro@example.com` / `teampro`
4. Create a new Event Type for the `Team Pro` user
5. Edit `Collective Seeded Team Event` > Assignment > include the Team Pro user > Save
6. In an incognito browser, book a meeting under `Collective Seeded Team Event`
7. In an incognito browser, check the availability for the Event Type you created in Step 4

**In Production** you will see that the Team Pro user is considered _available_ for the time slot booked in Step 6

**In this PR** you will see that the Team Pro user is considered _unavailable_ for the time slot booked in Step 3

## Checklist

- I have added tests that prove my fix is effective or that my feature works
- I have updated existing unit tests to pass locally with my changes

### DIFF
--- a/packages/core/getBusyTimes.ts
+++ b/packages/core/getBusyTimes.ts
@@ -34,43 +34,97 @@ export async function getBusyTimes(params: {
       status: BookingStatus.ACCEPTED,
     })}`
   );
+
+  /**
+   * A user is considered busy within a given time period if there
+   * is a booking they own OR host.
+   *
+   * Therefore this query does the following:
+   * - Performs a query for all EventType id's where this user is a host
+   * - Performs a query for all bookings where:
+   *   - The given booking is owned by this user, or..
+   *   - The given booking's EventType is hosted by this user
+   *
+   * See further discussion within this GH issue:
+   * https://github.com/calcom/cal.com/issues/6374
+   *
+   * NOTE: Changes here will likely require changes to some mocking
+   *  logic within getSchedule.test.ts:addBookings
+   */
   performance.mark("prismaBookingGetStart");
-  const busyTimes: EventBusyDetails[] = await prisma.booking
-    .findMany({
-      where: {
-        userId,
-        startTime: { gte: new Date(startTime) },
-        endTime: { lte: new Date(endTime) },
-        status: {
-          in: [BookingStatus.ACCEPTED],
-        },
-      },
-      select: {
-        id: true,
-        startTime: true,
-        endTime: true,
-        title: true,
-        eventType: {
-          select: {
-            id: true,
-            afterEventBuffer: true,
-            beforeEventBuffer: true,
+  const busyTimes: EventBusyDetails[] =
+    // Getting all EventTypes ID's hosted by this user
+    await prisma.host
+      .findMany({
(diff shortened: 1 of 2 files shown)


AaronPresley:
@emrysal @PeerRich this one's ready for review 👍 It took a while to crawl through the schema to figure out the best way to gather this data, but I eventually figured out that the `Host` table is the best source. I've left a lot of comments in the code itself to explain the flow.

Also, I noticed there's some performance testing around this function (named `prismaBookingGetStart` and `prismaBookingGetEnd`). I'm guessing there will be a bit of time added to this query, but hoping it's minimal.

@emrysal please let me know if you envisioned this lookup happening a different way. Thanks!

packages/core/getBusyTimes.ts
```diff
@@ -34,43 +34,97 @@ export async function getBusyTimes(params: {
...
   performance.mark("prismaBookingGetStart");
-  const busyTimes: EventBusyDetails[] = await prisma.booking
-    .findMany({
-      where: {
-        userId,
-        startTime: { gte: new Date(startTime) },
-        endTime: { lte: new Date(endTime) },
-        status: {
-          in: [BookingStatus.ACCEPTED],
+  const busyTimes: EventBusyDetails[] =
```
emrysal, line 55: Excellent going! Needs one amend though as currently if you are not the host of any event type (non-team user) `prisma.host.findMany` will return an empty array, this will cause all following then's to be skipped, in which case the bookings owned by the user are no longer included.

emrysal (changes requested):

emrysal (approved):
Cannot re-create the problem (assume I was just looking wrong) - Approved @AaronPresley 🙇

PR #6930, opened 2023-02-07
Feature/maintenance mode

emrysal:
## What does this PR do?

Implements a MaintenancePage that is used by our Vercel Edge Config to show a maintenance window by setting the following variable.

```json
{"isInMaintenanceMode": true}
```

image

see: https://vercel.com/templates/next.js/maintenance-page


emrysal:
Confirmed on the preview environment that this works, even though this is not active on the majority of routes, when a 500 error is detected on the app the user is automatically redirected to /auth/login which does have this enabled, the result is the desired behaviour.

image

pumfleet (commented):
Code looks good to me, haven't tested but will defer to zomars for final approval

PeerRich:
interesting, tho i wanna make sure this only happens when actually something happens. i don't wanna show this to silent 500's and have false positives people thinking we're down

apps/web/middleware.ts
```diff
@@ -8,6 +9,22 @@ import { extendEventData, nextCollectBasicSettings } from "@calcom/lib/telemetry
 const middleware: NextMiddleware = async (req) => {
   const url = req.nextUrl;
 
+  if (!url.pathname.startsWith("/api")) {
+    try {
+      // Check whether the maintenance page should be shown
+      const isInMaintenanceMode = await get<boolean>("isInMaintenanceMode");
```
JeroenReumkens, line 15: Be aware that this middleware doesn't run on all routes. It only runs if the path is matched in the config on line 70. That config was added now so long ago, because before that the middleware was indeed ran on all of the routes, which was quite costly as far as I remember.

emrysal: I tried earlier on to have it without that config, but that also meant this logic was ran on non html pages like _next - brought back that config.

The reason this works although effectively the only page is /auth/login is because the user is redirected to the login page on trpc error (I think it has an unauth handler that triggers on error).

JeroenReumkens (approved):

zomars (approved):
Ship it

PR #6945, opened 2023-02-08
Fix trackingId required error

### LINKED ISSUE #6917: In the App page in an Event - the Google Analytics toggle generates an error even if deactivated

### Issue Summary

https://app.cal.com/event-types/{event-id}?tabName=apps
If you just toggle on and off the Google Analytics app (leaving it empty) even if it's empty, you get the error that the tracking ID is not set. It prevents you from saving.

### Steps to Reproduce

Screen Recording 2023-02-07 at 10 09 04 AM

hariombalhara:
Fixes #6917

Explaination:
If someone toggles an Analytics app without actually making a change to trackingId input, it creates an app metadata object with just `enabled` prop in it. `trackingId` is a required property and it couldn't be set because onChange event didn't fire.


emrysal (approved):
Looks good to me, that explains the type infer issue

packages/app-store/plausible/zod.ts
```diff
@@ -4,9 +4,7 @@ import { eventTypeAppCardZod } from "../eventTypeAppCardZod";
 
 export const appDataSchema = eventTypeAppCardZod.merge(
   z.object({
-    // FIXME: Due to some reason autogenerated zod schema at prisma/zod/eventtype.ts fails typecheck on using `.default()`. It somehow thinks that trackingId can be undefined
```
hariombalhara, line 7: The reason actually was that when using default you have to mark it optional as well which is weird.

PR #6959, opened 2023-02-09
fix(calendar): recurring event fix

zomars:
From #6732 

> the loop was running way too often on every recurring object mostly due to the fact that the iterator was not checked for finalization, and in fact the iterator is not implemented correctly so it does not have the `done` property. instead it checks for undefined or not
> 
> also further note, this part is a very slow implementation, which will cause everything to become very slow and practically unusable, there are probably better ways to handle this. Why? well simple, this iterates through every possible day, instead of just applying the current window to it. That means on the one hand this functionality is broken from the first place (b/c it wont work if it is a very long running event) and on the other it was implemented at a very high expense.
> 
> --
> 
> edit, added at least a start date to the iterator, that should increase the speed of this already.
> 
> Signed-off-by: Tobias Gurtzick [magic@wizardtales.com](mailto:magic@wizardtales.com)


packages/lib/CalendarService.ts
```diff
@@ -359,15 +360,24 @@ export default abstract class BaseCalendarService implements Calendar {
...
         const start = dayjs(dateFrom);
         const end = dayjs(dateTo);
-        const iterator = event.iterator();
-        let current;
+        const startDate = ICAL.Time.fromDateTimeString(startISOString);
+        startDate.hour = event.startDate.hour;
+        startDate.minute = event.startDate.minute;
+        startDate.second = event.startDate.second;
+        const iterator = event.iterator(startDate);
+        let current = iterator.next();
```
wzrdtales, line 368: to note this should be just 
```
let current;
```

otherwise the first entry gets skipped.

seems like this got missing.

wzrdtales:
so its off the radar again?

zomars:
@leog Since you made the original implementation for recurring bookings. Can you give this the final test/approval? 🙏🏽

leog (approved):
> @leog Since you made the original implementation for recurring bookings. Can you give this the final test/approval? 🙏🏽

Everything good on my side, tested recurring events and worked as expected.

PR #6976, opened 2023-02-09
Bugfix/fix spacing dialog header

emrysal:
## What does this PR do?

image

- [ ] check other components that have custom spacing applied to achieve "design consistency"


emrysal:
Was worth merging at some point.

PR #7043, opened 2023-02-10
Stops seeding DB on each Build

zomars:
## What does this PR do?

- Makes seeding opt-in
- We already have a UI to enable App Store apps and setting keys

## Type of change

## How should this be tested?

- [ ] yarn build
- [ ] yarn e2e


zomars:
@SocketSecurity ignore @prisma/client@4.10.1 @prisma/engines@4.8.1

leog (approved):
Makes sense 👍

PR #7048, opened 2023-02-10
Fix app fee for outside of US accounts

alannnc:
## What does this PR do?

- Only add application fee if defaultCurrency it's from US.


zomars (changes requested):
We support fees for multiple currencies, not only USD.

alannnc:
Closing as it needs more investigation.

PR #7051, opened 2023-02-11
Tree shake icons

zomars:
## What does this PR do?

Adds tree shaking support four our UI icons. [Taken from here](https://github.com/react-icons/react-icons/issues/154#issuecomment-895976123).

https://react-icons.github.io/react-icons/

## Type of change


packages/ui/components/icon/Icon.tsx
```diff
@@ -1,10 +1,3 @@
 // find all feather icons at https://feathericons.com/
 // github https://github.com/feathericons/feather
 export * from "react-icons/fi";
-
-// if feather icon is missing, use "@heroicons/react/outline";
-export { CollectionIcon } from "@heroicons/react/outline";
-export { ShieldCheckIcon } from "@heroicons/react/outline";
-export { BadgeCheckIcon } from "@heroicons/react/outline";
-export { ClipboardCopyIcon } from "@heroicons/react/outline";
-export { StarIcon as StarIconSolid } from "@heroicons/react/solid";
```
zomars, line 10: Removed to enable next.js plugin

packages/ui/package.json
```diff
@@ -23,6 +23,7 @@
     "@radix-ui/react-popover": "^1.0.2",
     "@radix-ui/react-portal": "^1.0.0",
     "@radix-ui/react-select": "^0.1.1",
+    "@react-icons/all-files": "^4.1.0",
```
zomars, line 26: Added so we can import individually after transform

JeroenReumkens:
Very nice @zomars! 
I did test for the previous setup if unused icons where in the bundle, and that wasn’t the case anymore. So I think they where already tree shaken properly. However, I think your change is even better since now we also can’t mess up the tree shaking by accident 🔥

I’m only wondering if we still need to have the icons file in our repo, since we directly map this to npm now? Although I can imagine that typescript would get confused if we’d import from a file that doens’t exist. 

Apart from that a great change! Let me know when it’s out of draft.

PeerRich (approved):
nice new next 13.1 feature 🥰

apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
```diff
@@ -1,3 +1,4 @@
+import { StarIcon as StarIconSolid } from "@heroicons/react/solid";
```
emrysal, line 1: Here is room for improvement, though, as we're including from a specific library again beyond the abstraction. NIT as we use so little hero-icons. I think you've looked into this already and this was not possible.

emrysal (approved):

emrysal:
Have had a look at the comment @JeroenReumkens left; if what he's seeing is correct and tree shaking is already happening, I would argue this PR needlessly complicates the setup.

apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
zomars: Yes, a good follow up would be to get rid of heroicons for good

PeerRich:
isnt this ready to merge? seems pretty good to me + approved by two people @zomars

zomars:
Waiting on bundle analyzer to give us the final veredict

zomars:
I'll take this as a win 
image

PR #7059, opened 2023-02-12
show upgrade badge and hide /team/null when team unpublished

### LINKED ISSUE #7058: [CAL-1045] make an "Upgrade" badge next to /null/ teams and hide slug
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1045](https://linear.app/calcom/issue/CAL-1045/make-an-upgrade-badge-next-to-null-teams-and-hide-slug)</sub>

PeerRich:
untested but i think this works

@zomars how can i test an unpublished team on localhost?


apps/web/pages/event-types/index.tsx
```diff
@@ -60,6 +60,8 @@ import SkeletonLoader from "@components/eventtype/SkeletonLoader";
 
 import { ssrInit } from "@server/lib/ssr";
 
+import { UpgradeTeamsBadge } from "../../../../packages/ui/components/badge/UpgradeTeamsBadge";
```
sean-brydon, line 63:
```suggestion
import { UpgradeTeamsBadge } from "@calcom/ui/components/badge/UpgradeTeamsBadge";
```

sean-brydon:
> untested but i think this works
> 
> @zomars how can i test an unpublished team on localhost?

You can create a unpublished team by going to step one of create a team -> Then just hit back or navigate back to localhost:3000

PeerRich:
actually the badge needs to link to stripe not to /teams 

we need to replace that @zomars

emrysal (approved):
Approved although it is not clear whether your request for feedback was intended to be a blocker @PeerRich (just approving and not auto-merge in case it was)

apps/web/pages/event-types/index.tsx
sean-brydon: Likely not to be honest @emrysal

PeerRich: true. its a smart component, better in features

PeerRich: but we actually dont wanna use this component at all haha

apps/web/pages/event-types/index.tsx
```diff
@@ -126,11 +128,17 @@ const Item = ({ type, group, readOnly }: { type: EventType; group: EventTypeGrou
...
-          className="hidden font-normal leading-4 text-gray-600 sm:inline"
-          data-testid={"event-type-slug-" + type.id}>{`/${group.profile.slug}/${type.slug}`}</small>
+        {group.profile.slug ? (
+          <small
+            className="hidden font-normal leading-4 text-gray-600 sm:inline"
+            data-testid={"event-type-slug-" + type.id}>
+            {`/${group.profile.slug}/${type.slug}`}
+          </small>
+        ) : (
+          <UpgradeTeamsBadge />
```
PeerRich, line 138: @zomars can we give this a props => upgradeLink={true} that uses the "upgrade" stripe api link vs the /team url

PeerRich: keep the logic with tooltip etc but link to the stripe checkout

zomars: No need to prop AFAIK. we can encapsulate that logic in the components itself.

PeerRich: not sure what that means but if you have a free minute may you wanna add that? you built the initial "upgrade here" top bar banner

PeerRich:
asked @CarinaWolli to help make a new prop that makes the Upgrade batch link to stripe instead fo /teams

CarinaWolli (dismissed):
@PeerRich the `<UpgradeTeamsBadge/>` only shows when the user doesn't already have a paid plan (so no team and not premium user name). So it's not working for that case as user can already have teams or a premium name and we still want the upgrade badge here to upgrade their new team

CarinaWolli:
@zomars or @alannnc could you check if the upgrade badge is doing what it is supposed to? I don't know how to test that locally and I don't have stripe set up

alannnc (approved):

PR #7060, opened 2023-02-12
chore: turn inline logo & shell headers into h3

p6l-richard:
## What does this PR do?

image

## Type of change

### DIFF
--- a/packages/features/shell/Shell.tsx
+++ b/packages/features/shell/Shell.tsx
@@ -797,13 +797,13 @@ export function ShellMain(props: LayoutProps) {
             {props.HeadingLeftIcon && <div className="ltr:mr-4">{props.HeadingLeftIcon}</div>}
             <div className={classNames("w-full ltr:mr-4 rtl:ml-4 md:block", props.headerClassName)}>
               {props.heading && (
-                <h1
+                <h3
                   className={classNames(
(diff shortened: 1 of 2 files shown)


emrysal:
I'd also like an opinion from @JeroenReumkens on this - I agree with this change but also question whether h3 is the right alternative.

sean-brydon:
> I'd also like an opinion from @JeroenReumkens on this - I agree with this change but also question whether h3 is the right alternative.

Putting my input here too -> I agree with this change - No reason at all for logo to be H1. H3 is fine from a a11y standpoint

emrysal (approved):
Approved as clarified by @sean-brydon

JeroenReumkens:
The `h1` in the shell definitely is a good one. And the logo indeed shouldn't be an `h1`, perhaps not even a heading at all. 
Headings work like book chapters, where `h1` is the book title (and the book is the page) — and that is not the Cal.com logo, but rather the title of the page. And `h2` is then chapter 1.1, 1.2, 2.1, etc (1 level deep), and the `h3` is 1.1.1, 1.2.1 2.1.1 etc. 

This also shows that you can't skip levels (so no h3 without an h2), and also proves the point that the logo isn't an h3, because it's not suddenly a sub chapter of anything. So it should actually be a paragraph 😁 
But since this PR is already merged, I think the biggest issue is solved, and that is making the right thing an h1. That's by far the most important part 🙂

emrysal:
Yeah that was my idea as well @JeroenReumkens - I wasn't sure if it was "best" to have h3, but the main problem was with the h1 so was good to merge.

Issue #7063, opened 2023-02-13
[CAL-1048] Lazy load Phone Number Input on BookingPages


hariombalhara:
Can we lazy load this component?

_Originally posted by @zomars in https://github.com/calcom/cal.com/pull/6560#discussion_r1103483352_
            

<sub>[CAL-1048](https://linear.app/calcom/issue/CAL-1048/lazy-load-phone-number-input-on-bookingpages)</sub>

esau-morais:
May I be assigned to work on this issue?

hariombalhara:
@esau-morais Sure. Done

esau-morais:
Correct If I'm wrong, but when I was taking a look closely to the code and trying to understand the component highlighted in this issue, I've found there's already a dynamic import for it, but it just is used nowhere. Can I then just implement it at the `BookingPage.tsx`?

![packages/ui/form/PhoneInputLazy.tsx](https://user-images.githubusercontent.com/55207584/218630525-01d5873e-e28d-4496-8bf2-c34158523fc4.png)

zomars:
Will close for now. This was an oversight of my part. Thanks again for you contribution 🙏🏽

PR #7083, opened 2023-02-14
Lazy load BasePhoneInput

esau-morais:
## What does this PR do?

Improve performance by lazy loading primitive Phone Number Input -> #7063

PS.: Type checks will depend on #6560 

## Type of change

## How should this be tested?

- [ ] Main use case: Create new Event Type with custom input of type "Phone Number"

## Checklist


roae (changes requested):
Need to import the isolate type

zomars:
Just noticed that we already have a LazyLoaded version of this component.

image

Also that's the only that is currently being exported by default

image

So this PR won't make a bundle size difference AFAIK.

zomars:
Thank you for your contribution 🙏 But will close this for now.

PR #7087, opened 2023-02-14
Implement URL validation on default videos links

sean-brydon:
## What does this PR do?
Adds validation to default links to prevent miss use 

https://www.loom.com/share/a27229dba13747a29bcd249e0c320453

### DIFF
--- a/packages/app-store/locations.ts
+++ b/packages/app-store/locations.ts
@@ -15,7 +15,7 @@ export type DefaultEventLocationType = {
   category: string;
 
   iconUrl: string;
-
+  urlRegExp?: string;
   // HACK: `variable` and `defaultValueVariable` are required due to legacy reason where different locations were stored in different places.
   variable: "locationType" | "locationAddress" | "address" | "locationLink" | "locationPhoneNumber" | "phone";
   defaultValueVariable: "address" | "attendeeAddress" | "link" | "hostPhoneNumber" | "phone";
--- a/packages/trpc/server/routers/viewer.tsx
+++ b/packages/trpc/server/routers/viewer.tsx
@@ -1195,7 +1195,7 @@ const loggedInViewerRouter = router({
       const credentials = ctx.user.credentials;
       const foundApp = getApps(credentials).filter((app) => app.slug === input.appSlug)[0];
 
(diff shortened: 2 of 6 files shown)


packages/app-store/locations.ts
```diff
@@ -15,7 +15,7 @@ export type DefaultEventLocationType = {
   category: string;
 
   iconUrl: string;
-
+  urlRegExp?: string;
```
sean-brydon, line 18: This existed in the object just wasnt typed

packages/trpc/server/routers/viewer.tsx
```diff
@@ -1192,6 +1192,24 @@ const loggedInViewerRouter = router({
     )
     .mutation(async ({ ctx, input }) => {
       const currentMetadata = userMetadata.parse(ctx.user.metadata);
+      const credentials = ctx.user.credentials;
+      const foundApp = getApps(credentials).filter((app) => app.slug === input.appSlug)[0];
+
+      if (!foundApp) throw new TRPCError({ code: "BAD_REQUEST", message: "App not found" });
```
sean-brydon, line 1198: Added a bit more validation so you cant break things

hariombalhara: Given the condition, I think the error message should probable be "App not installed"

sean-brydon:
```suggestion
      if (!foundApp) throw new TRPCError({ code: "BAD_REQUEST", message: "App not installed" });
```

packages/trpc/server/routers/viewer.tsx
```diff
@@ -1192,6 +1192,24 @@ const loggedInViewerRouter = router({
...
+
+      if (foundApp.appData?.location?.linkType === "static" && !input.appLink) {
+        throw new TRPCError({ code: "BAD_REQUEST", message: "App link is required" });
+      }
+
+      if (foundApp.appData?.location?.linkType === "static" && foundApp.appData.location.urlRegExp) {
```
hariombalhara, line 1204: nit: Maybe store  `foundApp.appData?.location?` in a variable

hariombalhara (approved):

JeroenReumkens (approved):
Great work!

PR #7098, opened 2023-02-14
Change quick setup to use docker compose plugin

### LINKED ISSUE #5819: yarn dx uses old "docker-compose" instead of current "docker compose"

### Issue Summary

When creating the initial database user using yarn dx, you will get an error about docker-compose. This is because current docker uses the syntax docker "space" compose, no hyphen.

### Steps to Reproduce

At https://github.com/calcom/cal.com#quick-start-with-yarn-dx is where the instructions are located.

### Technical details

This is on a vanilla Debian server with newest version of docker installed. It is a syntax issue though, not a hardware issue.

renancleyson-dev:
## What does this PR do?

Use the `docker compose` plugin as the first option at `yarn dx` related commands and `docker-compose` as a fallback.
Fixes #5819 

**Environment**: Development

## Type of change

## How should this be tested?

Test if the quick setup with `yarn dx` works the same way as before creating a locally populated DB successfully.

## Checklist


leog (approved):
Just to be clear, `docker compose` seems like an upgrade to the separated CLI v1 version `docker-compose` and this change will try to use the v2 and if it fails fallbacks to v1. v2 is supposed to behave almost the same as v1 but it may have differences. I just tested on macOS Ventura 13.0 and it worked with v2 as expected. I asked @alishaz-polymath to test on his Windows machine just to double check on another platform. Other than that, approving.

zomars (approved):

PR #7099, opened 2023-02-14
fix/premium-username-user-state-overriden

alannnc:
## What does this PR do?

Update from latest release overridden user.state being taken in consideration for Settings/Profile username field.
With this changes we should handle both scenarios.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Go to settings/profile and you no longer see warning about upgrading username to premium.- [ ] Test B

## Checklist

### DIFF
--- a/apps/web/components/ui/UsernameAvailability/index.tsx
+++ b/apps/web/components/ui/UsernameAvailability/index.tsx
@@ -1,3 +1,5 @@
+import { useRouter } from "next/router";
+import { useState } from "react";
 import { Controller, useForm } from "react-hook-form";
 
 import { IS_SELF_HOSTED } from "@calcom/lib/constants";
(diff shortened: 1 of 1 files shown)


apps/web/components/ui/UsernameAvailability/index.tsx
```diff
@@ -22,7 +24,13 @@ export const UsernameAvailabilityField = ({
   onErrorMutation,
   user,
 }: UsernameAvailabilityFieldProps) => {
-  const { username: currentUsername, setQuery: setCurrentUsername } = useRouterQuery("username");
+  const router = useRouter();
+  const [currentUsernameState, setCurrentUsernameState] = useState(user.username || "");
+  const { username: usernameFromQuery, setQuery: setUsernameFromQuery } = useRouterQuery("username");
+  const { username: currentUsername, setQuery: setCurrentUsername } =
+    router.query["username"] && user.username === null
```
alannnc, line 31: Having username as query param shouldn't save on DB actual value, it just place query username value and validate if user has that username or it needs to be upgraded.

leog (approved):
LGTM

PR #7102, opened 2023-02-14
Add/eslint plugin unused imports

zomars:
## What does this PR do?

- The bad: Unused imports will now error
- The good: We now have a way to auto-fix them
- The ugly: Big diff coming in a follow-up

## Type of change

## How should this be tested?

- [ ] `yarn lint:fix`


PeerRich (approved):
ah great. always disliked unused imports

PR #7104, opened 2023-02-15
Gradually increase the getSchedule refetch interval until it reaches 30 seconds

roae:
## What does this PR do?

increase refetch interval after each refetch to getSchedule until 30 seconds
 1. 3 seconds
 2. 5 seconds
 3. 10 seconds
 4. 20 seconds
 5. 30 seconds
 6. ...

## Type of change


baileypumfleet:
What do you think about this @emrysal @zomars @roae?

apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -75,10 +93,16 @@ const useSlots = ({
     },
     {
       enabled: !!startTime && !!endTime && enabled,
-      refetchInterval: 3000,
+      refetchInterval,
       trpc: { context: { skipBatch: true } },
     }
   );
+  useEffect(() => {
+    if (!!data && fetchStatus === "idle") {
```
emrysal, line 101: refetchCount should be it's own hook, otherwise other re-renders may interfere here

Also we probably want to useRef() this counter, no need to rerender 🙏

emrysal (changes requested):
refetchCount should be it's own hook, otherwise other re-renders may interfere here

Also we probably want to useRef() this counter, no need to rerender 🙏

zomars (approved):
Ship it

PR #7105, opened 2023-02-15
injects secrets with infisical

maidul98:
## What does this PR do?
Maidul, from team Infisical. This pull request adds [Infisical](https://infisical.com) to the root package.json so that secrets can be injected into all-consuming apps in the mono repo without the need for `.env` files. This allows the internal development team to have secrets synced across the whole team. 

There are several ways to inject secrets into a mono repo with Infisical. I have chosen this method based on my understanding of how `.env` and the `.env.appStore` are used in this repo.  I saw that in some packages such as Prisma both `.env` and the `.env.appStore` are utilized. As a result, I think it is best to create a single project in Infisical that holds secrets from both of these files, and tag the secrets according to the file they belong to. For example, all secrets imported from .env file can be tagged `web` and secrets from app store .env can be tagged `appstore` within the same Infisical project. 

### Steps to test this change out 
1. Create a project in Infisical and upload secrets from both `.env` and the `.env.appStore`. Tag secrets from .env with `web` and secrets from app store .env with `appstore`
2. Install infisical CLI on your machine by following the instructions here https://infisical.com/docs/cli/overview
3. Clone this repo with this change and run `infisical init` at the root 
4. Run `yarn run i-dx`

Example tagging: 
img

### Generate example env files 
To generate example files for `.env` and the `.env.appStore` with Infisical secrets, run 
`yarn run i-gen-web-example-env` `yarn run i-gen-app-store-example-env` respectively 

This will use the tags, comments, and secrets to create a consistent example .env file each time using your secrets in Infisical.

### Does adding Infisical affect contributors and existing self-host?
In short, it doesn't. Contributors will continue to use .envs to pass in secrets into the node process as they do while the internal team will use Infisical to pass in secrets into the node process. 

## Type of change

## How should this be tested?

No tests are needed, existing tests pass.

### Other notes 
Discussion to be continued on Thursday with Omar Lopez. Let me know if I can help in any way!


PeerRich:
hey, we have a lot of DX things we need to clean up first and will then look into this! thank you for the patience

package.json
```diff
@@ -39,6 +39,9 @@
     "docs-start": "turbo run start --scope=\"@calcom/docs\"",
     "build-storybook": "turbo run build --scope=\"@calcom/storybook\"",
     "dx": "turbo run dx",
+    "i-dx": "infisical run -- turbo run dx",
```
zomars, line 42:
```suggestion
    "i-dev": "infisical run -- turbo run dev --scope=\"@calcom/web\"",
    "i-dx": "infisical run -- turbo run dx",
```

zomars (approved):

zomars:
Ship it

PR #7112, opened 2023-02-15
refactor: usetheme hook

G3root:
## What does this PR do?

there are multiple reports of theme flashing, which is hard to reproduce this PR Aims to possibly fix it by removing some side effects

## Type of change


sean-brydon (changes requested):
Nice! tested re-renders with react-profiler and this seems to be a lot better. Can't replicate any flashing either even when switching system themes (Which is how i can break it in main)

Great job! Could we add the comment i suggested or something like it to explain why we dont have all dependancies in the array

sean-brydon (approved):
Approving but wont merge incase anyone else wants to take a look :)

PR #7118, opened 2023-02-15
Filter Timezones by cities

gitstart-calcom:
## What does this PR do?

Adds timezone filtering by cities

## Demo
https://www.loom.com/share/690a72cfe09d486594d6f2254194d1cb

Dropdown overlapping issue fixed
https://www.loom.com/share/32280b4fe3d249b5acbff46ba9927144


apps/web/components/booking/pages/AvailabilityPage.tsx
```diff
@@ -231,7 +231,7 @@ function TimezoneDropdown({
 
   return (
     <>
-      <div className="dark:focus-within:bg-darkgray-200 dark:bg-darkgray-100 dark:hover:bg-darkgray-200 -mx-[2px] !mt-3 flex w-fit items-center rounded-[4px] px-1 py-[2px] text-sm font-medium focus-within:bg-gray-200 hover:bg-gray-100 [&_svg]:focus-within:text-gray-900 dark:[&_svg]:focus-within:text-white [&_p]:focus-within:text-gray-900 dark:[&_p]:focus-within:text-white">
+      <div className="dark:focus-within:bg-darkgray-200 dark:bg-darkgray-100 dark:hover:bg-darkgray-200 -mx-[2px] !mt-3 flex w-fit max-w-[20rem] items-center rounded-[4px] px-1 py-[2px] text-sm font-medium focus-within:bg-gray-200 hover:bg-gray-100 md:max-w-[12rem] [&_svg]:focus-within:text-gray-900 dark:[&_svg]:focus-within:text-white [&_p]:focus-within:text-gray-900 dark:[&_p]:focus-within:text-white">
```
gitstart-calcom, line 234: This is to fix the overlapping of input 
Screenshot 2023-02-15 at 10 35 31 PM

gitstart-calcom:
Not sure why these tests are failing. They are passing locally. How can we get them fixed?

package.json
```diff
@@ -80,6 +80,7 @@
     "ts-jest": "^28.0.8"
   },
   "dependencies": {
+    "city-timezones": "^1.2.1",
```
zomars, line 83: You're adding a new dependency but the lock file didn't update

zomars (changes requested):

PeerRich:
NIT: can we replace "_" with " "? 

CleanShot%202023-02-15%20at%2019 00 06@2x


Los_Angeles => Los Angeles

zomars:
We cannot merge this @PeerRich 

We can create an endpoint to lazy load this data and reduce initial bundle size. But this is way out of our performance budget.

image

image

zomars (changes requested):
Blocking to address performance concerns

emrysal:
Simplest approach here is to use a static file and then load that on the select - not part of the bundle and cached 🙏

PeerRich:
> Simplest approach here is to use a static file and then load that on the select - not part of the bundle and cached 🙏

so what you are saying is we need a static: cities.json with all info right?

gitstart-calcom:
Not sure if we would want to keep this but we have added a loading state to the select since it takes some time for the first load
https://www.loom.com/share/690a72cfe09d486594d6f2254194d1cb

gitstart-calcom:
Hi @PeerRich, can you please help with the `yarn.lock` issue?

zomars:
Adding caching and preventing batching for this endpoint: 
image

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -65,6 +65,7 @@ export default trpcNext.createNextApiHandler({
       "viewer.public.i18n": `no-cache`,
       // Revalidation time here should be 1 second, per https://github.com/calcom/cal.com/pull/6823#issuecomment-1423215321
       "viewer.public.slots.getSchedule": `no-cache`, // FIXME
+      "viewer.public.cityTimezones": `max-age=${ONE_DAY_IN_SECONDS}, stale-while-revalidate`,
```
Comment (zomars) [inline apps/web/pages/api/trpc/[trpc].ts:68]: Added so we can cache the response

apps/web/test/lib/getTimezone.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  altName: "Pacific Standard Time",
+};
+
+it("should return empty array for an empty string", () => {
+  expect(filterByCities("", cityData)).toStrictEqual([]);
+});
+
+it("should filter cities for a valid city name", () => {
+  expect(filterByCities("San Francisco", cityData)).toStrictEqual(cityData);
+});
```
zomars, line 42: Seems like tests are failing:

image

zomars (changes requested):

packages/ui/components/form/timezone-select/TimezoneSelect.tsx
```diff
@@ -1,28 +1,44 @@
...
 
+export interface ICity {
+  city: string;
+  timezone: string;
+}
+
 export function TimezoneSelect({ className, components, ...props }: SelectProps) {
+  const [cities, setCities] = useState<ICity[]>([]);
+  const { data, isLoading } = trpc.viewer.public.cityTimezones.useQuery(undefined, {
+    trpc: { context: { skipBatch: true } },
```
zomars, line 18: Added so we can cache this endpoint individually

zomars (approved):
Updated tests to use inline snapshots since they can be auto-updated.

Ship it

PR #7141, opened 2023-02-16
Fix default app in main appstore list

sean-brydon:
Fixes CAL-1085

CleanShot 2023-02-16 at 11 02 59@2x


packages/ui/components/apps/AppCard.tsx
```diff
@@ -149,7 +149,7 @@ export function AppCard({ app, credentials, searchText }: AppCardProps) {
           <span className="rounded-md bg-red-100 px-2 py-1 text-sm font-normal text-red-800">Template</span>
         )}
 
-        {app.isGlobal && (
+        {(app.isDefault || (!app.isDefault && app.isGlobal)) && (
```
sean-brydon, line 152: If there is no default app and it is global show it (covers daily video from before default apps were created)

emrysal (approved):
:shipit:

hariombalhara (approved):

PR #7144, opened 2023-02-16
fix(app): disable edit availability link while fetching

### LINKED ISSUE #7124: [CAL-1076] Link to undefined when loading availability in event type page
The "Edit availablity" links to `/availablity/undefined` when loading.

Also unrelated but this is endpoint sometimes gets stuck on loading locally.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1076](https://linear.app/calcom/issue/CAL-1076/link-to-undefined-when-loading-availability-in-event-type-page)</sub>

Mythie:
## What does this PR do?

Disables the "Edit availability" link when editing an event type until the availability has finished fetching. Handles an issue where clicking the link otherwise would cause it to link to `/availbilities/undefined`

Fixes #7124 

## Type of change

## How should this be tested?

[ ] Edit an event type, click on the availability tab and verify that the "Edit availability" link is not clickable while the availability is still fetching.

### DIFF
--- a/apps/web/components/eventtype/AvailabilityTab.tsx
+++ b/apps/web/components/eventtype/AvailabilityTab.tsx
@@ -157,6 +157,7 @@ const EventTypeScheduleDetails = () => {
         </span>
         <Button
           href={`/availability/${schedule?.schedule.id}`}
+          disabled={isLoading}
           color="minimal"
           EndIcon={FiExternalLink}
           target="_blank"


apps/web/components/eventtype/AvailabilityTab.tsx
```diff
@@ -157,6 +157,7 @@ const EventTypeScheduleDetails = () => {
         </span>
         <Button
           href={`/availability/${schedule?.schedule.id}`}
+          disabled={isLoading}
```
sean-brydon, line 160: LGTM however,  - I think it would be best to introduce IsInitalLoading here 

So we dont disable this button on every refresh 

https://tanstack.com/query/v4/docs/react/guides/disabling-queries#isinitialloading

Mythie: Awesome suggestion, this only became available in `@tanstack/query` > 4.8.0, for context we are currently on 4.3.4 based on the `yarn.lock`. Would we also like to upgrade the package as part of this change? 🤔

ShaneMaglangit:
Hey @Mythie, can you have a look another issue that seems to be related to this.

When opening `https://app.cal.com/event-types/____?tabName=availability`, it seems like the list of available hours is stuck on its loading state. Until an option is selected within the select component. This is despite the fact that there's an initial value set.

This is on the same state for almost 5 minutes now:

image

Updates properly when reselected:

image

Mythie:
Yep, I can replicate this behaviour. I'll look into it shortly in addition to the already in-progress fix!

apps/web/components/eventtype/AvailabilityTab.tsx
emrysal: Nah; this is good for now 🙏 - I was unaware this was so new!

emrysal (approved):
LGTM

apps/web/components/eventtype/AvailabilityTab.tsx
sean-brydon: > Awesome suggestion, this only became available in `@tanstack/query` > 4.8.0, for context we are currently on 4.3.4 based on the `yarn.lock`. Would we also like to upgrade the package as part of this change? 🤔

I had no idea thank you :)

PR #7158, opened 2023-02-16
Enforces explicit type imports

zomars:
## What does this PR do?

- Enforces explicit type imports (this was hurting our bundle size as found out by @roae )
- Added a way to auto fix those
- Upgraded all eslint related packages and synced across packages
- Upgraded prettier (it seems like the typeof issue got resolved)

## Type of change

## How should this be tested?

- [ ] `yarn lint`
- [ ] `yarn lint:fix`
- [ ] `yarn format`


packages/config/eslint-preset.js
```diff
@@ -5,7 +5,6 @@ module.exports = {
   plugins: ["unused-imports"],
   parserOptions: {
     tsconfigRootDir: __dirname,
-    project: ["./tsconfig.json"],
```
zomars, line 8: This was being ignored

emrysal (approved):

PR #7165, opened 2023-02-17
Bugfix/improved assignment page

emrysal:
## What does this PR do?

* Replaces the `id` primary key in `Host` with a `userId_eventTypeId` foreign key.
* Change the check for multiple avatars, now always show user avatars for team events, even if just 1 user.
image

* Do not share fixed hosts state between `collective` and `round robin` - reset the form when the event type changes `schedulingType`.
* Keep initial state as reference (update on form submission) so the initial default value is shown when navigating back to the  original schedulingType.
* Remove `fixedHosts` from the form values, instead only have hosts like the BE.


apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -35,93 +35,197 @@ const sortByLabel = (a: ReturnType<typeof mapUserToValue>, b: ReturnType<typeof
...
-  control,
+const CheckedHostField = ({
   labelText,
   placeholder,
   options = [],
+  isFixed,
+  value,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  onChange = () => {},
   ...rest
```
sean-brydon, line 46:
```suggestion
  onChange,
  ...rest
```

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -35,93 +35,197 @@ const sortByLabel = (a: ReturnType<typeof mapUserToValue>, b: ReturnType<typeof
...
-                {...rest}
-              />
+        <CheckedTeamSelect
+          isOptionDisabled={(option) => !!value.find((host) => host.userId.toString() === option.value)}
+          onChange={(options) => {
+            onChange(
+              options.map((option) => ({
+                isFixed,
+                userId: parseInt(option.value, 10),
+              }))
```
sean-brydon, line 66:
```suggestion
            onChange && onChange(
              options.map((option) => ({
                isFixed,
                userId: parseInt(option.value, 10),
              }))
```

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -35,93 +35,197 @@ const sortByLabel = (a: ReturnType<typeof mapUserToValue>, b: ReturnType<typeof
...
+        type="number"
+        label={t("minimum_round_robin_hosts_count")}
+        defaultValue={1}
+        {...formMethods.register("minimumHostCount")}
+        addOnSuffix={<>{t("hosts")}</>}
+                />*/}
+              </>
+            );
+        }
+        return <></>;
```
sean-brydon, line 195: Maybe something like this is cleaner? I know we do this elsewhere in the app

```suggestion
      render={({ field: { onChange, value } }) => {
        const schedulingTypeRender = {
          COLLECTIVE: (
            <CheckedHostField
              value={value}
              onChange={onChange}
              isFixed={true}
              options={teamMembers.sort(sortByLabel)}
              placeholder={t("add_attendees")}
              labelText={t("team")}
            />
          ),
          ROUND_ROBIN: (
            <>
              <RoundRobinHosts teamMembers={teamMembers} onChange={onChange} value={value} />
              {/*<TextField
                required
                type="number"
                label={t("minimum_round_robin_hosts_count")}
                defaultValue={1}
                {...formMethods.register("minimumHostCount")}
                addOnSuffix={<>{t("hosts")}</>}
              /> */}
            </>
          ),
        };
        return !!schedulingType ? schedulingTypeRender[schedulingType] : <></>;
      }}
```

sean-brydon (commented):
Everything looks good - few suggestions/talking points

Will wait for @zomars to approve due to me not being 100% sure on the migration (what will happen with existing data)

emrysal:
@zomars 

Proof this is safe to execute (no dupes):
```sql
select h1.id, h1."userId" ppt, h2."userId" ppt2, h2."eventTypeId" from "Host" h1, "Host" h2 where h1."userId" = h2."userId" AND h1."eventTypeId" = h2."eventTypeId" AND h1.id <> h2.id
;
 id | ppt | ppt2 | eventTypeId 
----+-----+------+-------------
(0 rows)
```

apps/web/pages/event-types/index.tsx
```diff
@@ -335,7 +335,7 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
                   <MemoizedItem type={type} group={group} readOnly={readOnly} />
                   <div className="mt-4 hidden sm:mt-0 sm:flex">
                     <div className="flex justify-between space-x-2 rtl:space-x-reverse">
-                      {type.users?.length > 1 && (
+                      {type.team && (
```
Udit-takkar, line 338: Screenshot 2023-02-24 at 12 00 56 AM

sometimes avatars are not loaded (don't know why but they are present when i do inspect elements).

Everything else works fine  👍

emrysal: Could be something network related 🤔

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -35,93 +35,195 @@ const sortByLabel = (a: ReturnType<typeof mapUserToValue>, b: ReturnType<typeof
...
+  useEffect(() => {
+    // Handles init & out of date initial value after submission.
+    if (!initialValue.current || initialValue.current?.submitCount !== submitCount) {
+      initialValue.current = { hosts: getValues("hosts"), schedulingType, submitCount };
+      return;
+    }
+    resetField("hosts", {
+      defaultValue: initialValue.current.schedulingType === schedulingType ? initialValue.current.hosts : [],
+    });
+  }, [schedulingType, resetField, getValues, submitCount]);
```
zomars, line 162: Can we prevent the need for this useEffect?

zomars: Id rather wait for the initial data in a parent component

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -35,93 +35,195 @@ const sortByLabel = (a: ReturnType<typeof mapUserToValue>, b: ReturnType<typeof
...
+  const { t } = useLocale();
 
   const schedulingTypeOptions: {
     value: SchedulingType;
     label: string;
     // description: string;
   }[] = [
     {
-      value: SchedulingType.COLLECTIVE,
+      value: "COLLECTIVE",
```
zomars, line 208: Was this causing bundle issues?

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -285,8 +275,6 @@ const EventTypePage = (props: EventTypeSetupProps) => {
             locations,
             metadata,
             customInputs,
-            hosts: hostsInput,
-            hostsFixed,
             // We don't need to send send these values to the backend
             // eslint-disable-next-line @typescript-eslint/no-unused-vars
             seatsPerTimeSlotEnabled,
```
Comment (zomars) [inline apps/web/pages/event-types/[type]/index.tsx:280]: no used?

Comment (emrysal) [reply apps/web/pages/event-types/[type]/index.tsx:280]: Not anymore, hosts no longer needs any changing after load so it comes in through ...input later on - a bit simpler than before

apps/web/components/eventtype/EventTeamTab.tsx
emrysal: Was wondering if it made a meaningful difference in bundle size

zomars (commented):
Need to test locally. Also left some useEffect concerns.

CarinaWolli:
Tested locally and works 👍 Can be approved from my side once @zomar's comments are addressed

zomars (approved):

PR #7171, opened 2023-02-17
Embed: Remove ugly windows scrollbar even on unscrollable content

hariombalhara:
See the ugliness here https://share.getcloudapp.com/mXubw6L7 as reported by a user

After Fix:
Screenshot 2023-02-17 at 4 02 39 PM


sean-brydon (approved):
The worst thing about windows is its scroll bar :) 

Ship it

Issue #6714, opened 2023-01-26
[CAL-929] Remove _metadata.ts and InstallAppButtonComponent.tsx files


hariombalhara:
All apps having config.json is not using _metadata.ts files. So, they are the safest to start with.

Apps not having config.json(those that were created before app-store cli came up), first need to create config.json

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-929](https://linear.app/calcom/issue/CAL-929/remove-metadatats-files)</sub>

PR #7174, opened 2023-02-17
feat(app-store): add facetime app to app-store

### LINKED ISSUE #6965: App: Apple FaceTime
I am writing to express my interest in having FaceTime integration added to Caldav app. As a user of your app, I believe that this feature would greatly enhance the app’s functionality and overall user experience.

I want to be able to use FaceTime as an option as a user. So, my clients will have an option to choose “FaceTime” location at cal.com in my schedule.

Regards, Oleksandr T.

Mythie:
## What does this PR do?

Add the FaceTime app to the app-store allowing for bookings locations using FaceTime web links and FaceTime app link schemes.

Fixes #6965

[Loom Video](https://www.loom.com/share/8edb6e61a7f14e41bbe2df3d5f9c97b1)

## Type of change

## How should this be tested?

- [ ] Add the FaceTime app
- [ ] Use FaceTime as a booking location
- [ ] Verify you are able to initial a FaceTime call using the booking link


PeerRich:
thank you so much! feel free to change author to youe github username and not Cal.com Inc

packages/app-store/facetime/config.json
```diff
@@ -0,0 +1,26 @@
...
+  "name": "Facetime",
+  "title": "Facetime",
+  "slug": "facetime",
+  "type": "facetime_video",
+  "imageSrc": "/api/app-store/facetime/icon.svg",
+  "logo": "/api/app-store/facetime/icon.svg",
+  "url": "https://cal.com/apps/facetime",
+  "variant": "conferencing",
+  "categories": ["video"],
+  "publisher": "Cal.com, Inc",
```
PeerRich, line 12:
```suggestion
  "publisher": "Mythe",
```

packages/app-store/facetime/components/InstallAppButton.tsx
```diff
@@ -0,0 +1,18 @@
+import type { InstallAppButtonProps } from "@calcom/app-store/types";
```
hariombalhara, line 1: We don't need to add this Component separately now. It's automatically handled. Can you try deleting this file and see if it works still.

packages/app-store/facetime/config.json
```diff
@@ -0,0 +1,26 @@
...
+  "publisher": "Mythie",
+  "email": "help@cal.com",
+  "description": "Facetime makes it super simple for collaborating teams to jump on a video call.",
+  "__createdUsingCli": true,
+  "appData": {
+    "location": {
+      "linkType": "static",
+      "type": "integrations:facetime_video",
+      "label": "Facetime",
+      "organizerInputPlaceholder": "https://facetime.apple.com/join... or facetime:<contact>",
```
hariombalhara, line 21: facetime:<contact> would only work if the facetime app is installed. Otherwise this link would just not work.

So, I think we shouldn't suggest that. If facetime works through browser that's the right thing to do. 
I hope facetime website would automatically open facetime app if it's installed.

Mythie: It does open the app when visiting the website, however that still requires that a user generate a FaceTime link. If we’re happy for that to be required I can make the update in a few hours once it’s morning again.

packages/app-store/facetime/components/InstallAppButton.tsx
hariombalhara: Removed this file and _metadata.ts. Those files are not required now. New apps created using cli don't have them.
There is an issue in which we need to cleanup old apps https://github.com/calcom/cal.com/issues/6714

packages/app-store/facetime/config.json
hariombalhara: I think that would be a better approach because the organizer can generate the link once. If we allow **facetime:** scheme, bookers won't be able to join meeting by clicking on the link they receive.

Mythie: Latest commit updates the config to follow the new decision on the FaceTime links, the placeholder is also updated correspondingly:

image

hariombalhara (approved):

PR #7180, opened 2023-02-17
Node version was going to 17 on Heroku & Digital Ocean Build Packs

panbanda:
## What does this PR do?

When deploying on heroku and digital ocean I was receiving a node mismatch.  Because the engine is specified to be greater than 16 it was loading in 17 on build.

This fix adds a cap to the node engine so it stays within 16 per `.nvmrc`

## Type of change

## How should this be tested?

- [ ] Deploy to Heroku
- [ ] Deploy to DO

## Checklist


emrysal:
@panbanda is there a reason this is needed for Heroku? Node 17 should work (although we don't support it as it's a short term support version, much like 19) - we do however support Node 18 - can't approve this due to some of our users running Cal.com on node 18 but any feedback you can give on Heroku is welcome so we can find an alternative solution.

PR #7188, opened 2023-02-18
Techdebt: Remove _metadata.ts files for apps that have config.json

### LINKED ISSUE #6714: [CAL-929] Remove _metadata.ts and InstallAppButtonComponent.tsx files

hariombalhara:
Partially fixes #6714

The PR removes `_metadata.ts` files only for those apps that already have `config.json`. This is a pretty safe change because where config.json is present _metadata.ts isn't used

This is important to do cleanup because contributors still copy the code from old apps and end up copying _metadata.ts

Followup
- For apps that don't have config.json, convert _metdata.ts to config.json
- Delete all components/InstallAppButton.tsx files unless there is a reason for that to exist.


packages/app-store/stripepayment/components/index.ts
```diff
@@ -1 +0,0 @@
-export { default as InstallAppButton } from "./InstallAppButton";
```
hariombalhara, line 1: InstallAppButton file doesn't exist. It was there unnecessarily, which means Typechecks are not working on app-store. Need to add that,

zomars (approved):

PR #7194, opened 2023-02-18
Hotfix: Fix location select not rendering

hariombalhara:
Unexpected side effect of `useTheme`.


apps/web/components/ui/form/Select.tsx
```diff
@@ -74,9 +74,9 @@ function Select<
   };
 
   // Till we know in JS the theme is ready, we can't render react-select as it would render with light theme instead
-  if (!isReady) {
-    return <input type="text" className={className} />;
-  }
+  // if (!isReady) {
```
hariombalhara, line 77: There have been changes in Select. Maybe this is not required now. Will check it later.

emrysal (approved):

Issue #7191, opened 2023-02-18
Unable to add new location to event.


heyanurag:
### Issue Summary

Previously I used to see a modal asking me the type of location I want to add ([something like this](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=1739%3A78022&t=FpoXhi0rABWhdMj2-0)) but now I only see the edit location modal with no other option.

### Steps to Reproduce

1. Go to Edit Event
2. Click on Add location

Recording - https://www.loom.com/share/c7a063dc50f042fd9c54df56435243fe

PeerRich:
something is wrong with lazyloading @emrysal @zomars

hariombalhara:
This is fixed now.

harshsinghatz:
@hariombalhara This seem to be caused by a hook added yesterday which gets the mount state and it is referenced in useTheme. 
The RCA for this was that change in ref was not causing re-render due to which even though the ref's current was getting changed, isReady was still holding it's inital value (unchanged) i.e false

PR #7203, opened 2023-02-18
Fix: mounting state use theme

harshsinghatz:
## What does this PR do?
- This PR fixes the rendering issue that was causing the referenced issue below

Getting the unmount state seems redundant after cleanup seems redundant as the main use of this is to show a text only till the isReady is true

Fixes # (issue)
Reference: https://github.com/calcom/cal.com/issues/7191

## Type of change


harshsinghatz:
@hariombalhara

packages/lib/hooks/useMountedState.ts
```diff
@@ -9,13 +9,12 @@ import { useCallback, useEffect, useRef } from "react";
 export default function useMountedState(): () => boolean {
   const mountedRef = useRef<boolean>(false);
```
hariombalhara, line 10: Do we really need this ref now?

packages/lib/hooks/useMountedState.ts
```diff
@@ -9,13 +9,12 @@ import { useCallback, useEffect, useRef } from "react";
...
   const get = useCallback(() => mountedRef.current, []);
+  const [mount, setMount] = useState(false);
```
hariombalhara, line 12: Better name woud be mounted and setMounted

packages/lib/hooks/useMountedState.ts
harshsinghatz: No, removed it!

packages/lib/hooks/useMountedState.ts
harshsinghatz: makes sense, changes made

G3root (changes requested):
`IsMounted` value returned by useTheme hook is only used in the above select component, adding a hook with a setState call can cause unnecessary re-renders in other components which doesn't need the `isMounted` value. It's safe to use the `isMounted` only inside the select component and remove it from the useTheme hook.

harshsinghatz:
> `IsMounted` value returned by useTheme hook is only used in the above select component, adding a hook with a setState call can cause unnecessary re-renders in other components. It's safe to use the `isMounted` only inside the select component and remove it from the useTheme hook.

I don't think the re-renders would be an issue, the only time it would re-render the component is when component is actually mounted to replace the placeholder component. 
but if we don't want it i have removed the hook

harshsinghatz:
@G3root ?

G3root:
> > `IsMounted` value returned by useTheme hook is only used in the above select component, adding a hook with a setState call can cause unnecessary re-renders in other components. It's safe to use the `isMounted` only inside the select component and remove it from the useTheme hook.
> 
> I don't think the re-renders would be an issue, the only time it would re-render the component is when component is actually mounted to replace the placeholder component. but if we don't want it i have removed the hook @G3root

updated my previous comment to be more clear.

G3root (approved):
LGTM :)  Awesome work ❤️

hariombalhara (approved):

PR #7210, opened 2023-02-19
feat(lib): add more tests to lib package

Mythie:
## What does this PR do?

Add more tests to the lib package to make it more robust overall. Additionally, tidy any methods that can be modified without changing behaviour and tighten types where possible.

## Type of change

## How should this be tested?

- [ ] Run jest and verify that new tests pass


PeerRich:
thank you! we'll review once this leaves draft

packages/lib/timeFormat.ts
```diff
@@ -18,11 +18,19 @@ export const setIs24hClockInLocalStorage = (is24h: boolean) =>
 
 export const getIs24hClockFromLocalStorage = () => {
   const is24hFromLocalstorage = localStorage.getItem(is24hLocalstorageKey);
-  if (is24hFromLocalstorage === null) return null;
```
zomars, line 21: Single line returns are preferred for simple comparisons 🙏

zomars: Also we try to avoid git diffs for stylistic choices

Mythie: Oh sorry! It's a habit to remove them as I've found they are often responsible for bugs that go unnoticed since it's harder to parse/spot. I can revert that shortly.

Mythie:
Apologies, got a bit caught up the past week. Will update tomorrow and prep it for PR, I'll have to come back and add more tests later 😄

zomars:
No need to apologize. Thanks again for you contribution. 🙏🏽

zomars (approved):
LGTM 

Thank you for your contribution 🙏 @Mythie

PR #7178, opened 2023-02-17
Improved typings of window.Cal

mattpocock:
## What does this PR do?

Improves the typings of window.Cal by adding a `declare global` and removing a ton of `as CalWindow`

**Environment**: Staging(main branch) / Production

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

Complete

PR #7177, opened 2023-02-17
Improved types in ModalBox.ts

mattpocock:
## What does this PR do?

Refactors some types in `ModalBox.ts` to make the logic more robust.

**Environment**: Staging(main branch) / Production

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

Complete

PR #7212, opened 2023-02-19
improve typing by removing as html element

oscard0m:
## What does this PR do?
- refactor(embeds): use generic HTMLIFrameElement instead of 'as HTMLIFrameElement'
- refactor(embeds): improve types for floatingButton logic

Inspired by:
* https://github.com/calcom/cal.com/pull/7177
* https://github.com/calcom/cal.com/pull/7178
* https://twitter.com/housecor/status/1579837649363558401

## Type of change

## Checklist

✅

### DIFF
--- a/packages/embeds/embed-core/playwright/lib/testUtils.ts
+++ b/packages/embeds/embed-core/playwright/lib/testUtils.ts
@@ -1,4 +1,5 @@
-import { Page, Frame, test, expect } from "@playwright/test";
+import type { Page, Frame } from "@playwright/test";
+import { test, expect } from "@playwright/test";
 
 import prisma from "@calcom/prisma";
 
@@ -38,7 +39,7 @@ export const getEmbedIframe = async ({ page, pathname }: { page: Page; pathname:
   const iframeReady = await page.evaluate(() => {
     return new Promise((resolve) => {
       const interval = setInterval(() => {
-        const iframe = document.querySelector(".cal-embed") as HTMLIFrameElement | null;
+        const iframe = document.querySelector<HTMLIFrameElement>(".cal-embed");
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         if (iframe && iframe.contentWindow && window.iframeReady) {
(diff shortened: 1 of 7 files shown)


packages/embeds/embed-core/playwright/lib/testUtils.ts
```diff
@@ -38,7 +39,7 @@ export const getEmbedIframe = async ({ page, pathname }: { page: Page; pathname:
   const iframeReady = await page.evaluate(() => {
     return new Promise((resolve) => {
       const interval = setInterval(() => {
-        const iframe = document.querySelector(".cal-embed") as HTMLIFrameElement | null;
+        const iframe = document.querySelector<HTMLIFrameElement>(".cal-embed");
```
hariombalhara, line 42: Nice! Thanks !!

packages/embeds/embed-core/src/embed.ts
```diff
@@ -305,27 +305,34 @@ export class Cal {
...
+
     if (attributes?.id) {
-      attributesString += ` id="${attributes.id}"`;
       existingEl = document.getElementById(attributes.id);
     }
-    let el = existingEl;
+    let el: HTMLElement;
     if (!existingEl) {
       const template = document.createElement("template");
-      template.innerHTML = `<cal-floating-button ${attributesString}  data-cal-namespace="${this.namespace}" data-cal-link="${calLink}"></cal-floating-button>`;
```
hariombalhara, line 317: There was a reason to use the `innerHTML` method to create the custom element cal-floating-button. Using the createElement method doesn't actually make it work due to some reason. I don't remember the reason now :D. Can you test it once, it will break certainly. Floating button popup won't come

hariombalhara: Nevermind. I tested it and it was failing. I have pushed a fix for it.

hariombalhara: Also, e2e were passing. It means that we don't have floatingButton popup tests. Need to add those.

emrysal (approved):
Thanks! LGTM 👍

PR #6991, opened 2023-02-09
Fixing issues around workflows

CarinaWolli:
## What does this PR do?

Fixes issue that workflow email reminders of canceled bookings were still sent. 
Fixes issue that bookings that requested rescheduling still sent SMS and email reminders. 
Also, fixes the issue that when editing a workflow the data gets lost when clicking outside of the workflow. 

Fixes #6973
Fixes #6905
Fixes #6956

**Environment**: Staging(main branch) / Production

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #7232, opened 2023-02-20
Fixes that workflow reminders of cancelled and rescheduled bookings are still sent

### LINKED ISSUE #7225: [CAL-1121] reminder emails being sent even after cancellation

### Issue Summary

reminder emails being sent even after cancellation

### Steps to Reproduce

1. set up a workflow with emails
2. appoint a meeting with email workflow
3. cancel the meeting
4. reminder emails still being received

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead? expected to not receive any reminder

<sub>[CAL-1121](https://linear.app/calcom/issue/CAL-1121/reminder-emails-being-sent-even-after-cancellation)</sub>

CarinaWolli:
## What does this PR do?

Fixes that workflow emails are still sent for cancelled and rescheduled bookings. The fix of PR #6991 was wrong and didn't work as expected. 

The SendGrid endpoint `DELETE /v3/user/scheduled_sends/${referenceId}` was not used correctly. This endpoint deletes a cancelled scheduled email from `scheduled_sends` by removing the 'cancel' status (so email will again be scheduled). 

The reason why cancelling our scheduled emails stopped working is that all cancerlled emails are saved in `scheduled_sends` until the scheduled date but the max. of pending cancellations is 100. Once we reached 100 pending cancellations, new cancellations failed and emails were still sent out. 

The solution implemented in this PR: 
- A new field `cancelled` is added to the `workflowReminder` model
- Once an already scheduled email is cancelled, the field is set to true (nothing else is done at this point in time)
- The CRON job `scheduleEmailReminders` is responsible for the final cancellation of the scheduled email (runs every 15 mins)
  - cancels all scheduled emails that are cancelled and that have their scheduled date within the next hour

Fixes #7225

## Type of change


packages/features/ee/workflows/components/WorkflowStepContainer.tsx
```diff
@@ -387,81 +387,75 @@ export default function WorkflowStepContainer(props: WorkflowStepProps) {
                   }}
                 />
               </div>
-              {(isPhoneNumberNeeded || isSenderIdNeeded) && (
+              {isPhoneNumberNeeded && (
```
CarinaWolli, line 390: UI fix

Before: 
Screenshot 2023-02-20 at 11 36 08

After: 
Screenshot 2023-02-20 at 11 35 28

packages/features/ee/workflows/lib/reminders/emailReminderManager.ts
```diff
@@ -194,20 +194,41 @@ export const scheduleEmailReminder = async (
...
   try {
-    await client.request({
-      url: "/v3/user/scheduled_sends",
-      method: "POST",
-      body: {
-        batch_id: referenceId,
-        status: "cancel",
-      },
-    });
+    if (!referenceId) {
```
CarinaWolli, line 203: means that email is not yet scheduled with SendGrid

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -39,6 +40,42 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+        lte: dayjs().add(1, "hour").toISOString(),
+      },
+    },
+  });
+
+  try {
+    const workflowRemindersToDelete: Prisma.Prisma__WorkflowReminderClient<WorkflowReminder, never>[] = [];
+
+    for (const reminder of remindersToCancel) {
+      await client.request({
```
emrysal, line 57: NIT, what happens when this call fails? Perhaps an idea to catch the request error and throw

emrysal (approved):
Besides some considerations of redundancy (calls are too easy to fail without recovery) this is a nice fix, nice diagnosis & lgtm

mfeuerstein (commented):
## PR Review — approved

Reviewed 10 files. 0 high-severity issues found. Verdict: approved.

### `packages/features/bookings/lib/handleNewBooking.ts` (low)

- Reviewed packages/features/bookings/lib/handleNewBooking.ts — looks good

### `packages/prisma/migrations/20230217230604_add_cancelled_to_workflow_reminder/migration.sql` (low)

- Reviewed packages/prisma/migrations/20230217230604_add_cancelled_to_workflow_reminder/migration.sql — looks good

### `packages/features/ee/workflows/lib/reminders/smsReminderManager.ts` (low)

- Reviewed packages/features/ee/workflows/lib/reminders/smsReminderManager.ts — looks good

### `packages/features/ee/workflows/api/scheduleEmailReminders.ts` (low)

- Reviewed packages/features/ee/workflows/api/scheduleEmailReminders.ts — looks good

### `packages/features/ee/workflows/lib/reminders/emailReminderManager.ts` (low)

- Reviewed packages/features/ee/workflows/lib/reminders/emailReminderManager.ts — looks good

### `packages/features/bookings/lib/handleCancelBooking.ts` (low)

- Reviewed packages/features/bookings/lib/handleCancelBooking.ts — looks good

### `packages/prisma/schema.prisma` (low)

- Reviewed packages/prisma/schema.prisma — looks good

### `packages/features/ee/workflows/components/WorkflowStepContainer.tsx` (low)

- Reviewed packages/features/ee/workflows/components/WorkflowStepContainer.tsx — looks good

### `packages/trpc/server/routers/viewer/workflows.tsx` (low)

- Reviewed packages/trpc/server/routers/viewer/workflows.tsx — looks good

### `packages/trpc/server/routers/viewer/bookings.tsx` (low)

- Reviewed packages/trpc/server/routers/viewer/bookings.tsx — looks good

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -483,29 +481,18 @@ async function handler(req: NextApiRequest & { userId?: number }) {
...
   updatedBookings.forEach((booking) => {
     booking.workflowReminders.forEach((reminder) => {
-      if (reminder.scheduled && reminder.referenceId) {
-        if (reminder.method === WorkflowMethods.EMAIL) {
-          deleteScheduledEmailReminder(reminder.referenceId);
-        } else if (reminder.method === WorkflowMethods.SMS) {
-          deleteScheduledSMSReminder(reminder.referenceId);
-        }
+      if (reminder.method === WorkflowMethods.EMAIL) {
+        deleteScheduledEmailReminder(reminder.id, reminder.referenceId);
```
ron-x5labs, line 488: **Non-blocking — fire-and-forget cancellation write.** `deleteScheduledEmailReminder`/`deleteScheduledSMSReminder` are async, but this `forEach` neither `await`s them nor collects the returned promises into `prismaPromises` (line 495 only holds `attendeeDeletes`/`bookingReferenceDeletes`). The deferred email path's whole job here is `prisma.workflowReminder.update({ cancelled: true })`; if that promise doesn't settle before the serverless handler returns (Vercel can freeze the instance after the response), the flag is lost and the CRON never cancels the email — the bug this PR fixes recurs. Previously the row `deleteMany` *was* awaited via `prismaPromises`/`Promise.all`, so this is a reliability regression for the DB write. The same unawaited-`forEach` pattern recurs at `bookings.tsx:490` and `handleNewBooking.ts:966`. Collect the promises and `await Promise.all(...)` (or `await` each).

packages/features/ee/workflows/lib/reminders/emailReminderManager.ts
```diff
@@ -194,20 +194,41 @@ export const scheduleEmailReminder = async (
...
+        },
+      });
+
+      return;
+    }
 
-    await client.request({
-      url: `/v3/user/scheduled_sends/${referenceId}`,
-      method: "DELETE",
+    if (immediateDelete) {
```
ron-x5labs, line 213: **Non-blocking — `immediateDelete` has no DB write and no retry.** This branch POSTs the SendGrid cancel and `return`s (line 222) without deleting the row or setting `cancelled: true`. If the SendGrid request fails, the `catch` at line 233 swallows the error and returns normally — no `cancelled` flag is ever set, so the CRON never retries, and the original SendGrid scheduled send stays active → the email still sends, silently. The deferred path exists precisely to give the CRON a retry signal; `immediateDelete` bypasses it. Consider setting `cancelled: true` (or deleting the row) *regardless* of the SendGrid outcome so a transient SendGrid failure is recoverable, rather than failing closed.

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -39,6 +40,42 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+      scheduledDate: {
+        lte: dayjs().add(1, "hour").toISOString(),
+      },
+    },
+  });
+
+  try {
+    const workflowRemindersToDelete: Prisma.Prisma__WorkflowReminderClient<WorkflowReminder, never>[] = [];
+
+    for (const reminder of remindersToCancel) {
```
ron-x5labs, line 56: **Non-blocking — cancel loop aborts on first error; successful cancels' deletes are lost.** The entire `for` loop is one `try/catch`. `client.request` is `await`ed per reminder and the `delete` is only pushed *after* a successful cancel. If any `client.request` throws, the loop jumps to `catch` and `Promise.all(workflowRemindersToDelete)` (line 74) is never reached, so the `delete` promises for reminders already cancelled at SendGrid never execute — those rows stay `cancelled: true`. On the next run they're re-fetched and SendGrid is asked to cancel an already-cancelled batch (often a 400), which can abort the loop again and stall cleanup of *all* due reminders, not just the failing one. Wrap each iteration in its own `try/catch` (and delete on success) so one bad batch doesn't block the rest.

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -39,6 +40,42 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     },
   });
 
+  //cancel reminders for cancelled/rescheduled bookings that are scheduled within the next hour
+  const remindersToCancel = await prisma.workflowReminder.findMany({
```
ron-x5labs, line 44: **Non-blocking — cancel query missing `method`/`referenceId` guards.** `remindersToCancel` filters only `cancelled: true` + `scheduledDate <= now+1h`, but the loop feeds `reminder.referenceId` straight to SendGrid as `batch_id` (line 61). There is no `method: WorkflowMethods.EMAIL` and no `referenceId: { not: null }` filter, and no DB constraint tying `cancelled = true` to `(method = EMAIL AND referenceId IS NOT NULL)`. This is safe today only because no current path sets `cancelled: true` on an SMS row or a null-referenceId row — an invariant held by accident, not by schema/query. A future caller that violates it produces a SendGrid 400 that (per the finding above) aborts the whole cancel loop. Add the filters (matching the `unscheduledReminders` query below, which already filters `method: EMAIL`).

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -39,6 +40,42 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     },
   });
 
+  //cancel reminders for cancelled/rescheduled bookings that are scheduled within the next hour
```
ron-x5labs, line 43: **Non-blocking — top-of-run `deleteMany` (line 34) ignores `cancelled`.** The `deleteMany` at line 34 deletes all past-due EMAIL reminders regardless of the `cancelled` flag, *before* the cancel loop below. A deferred-cancelled reminder whose `scheduledDate` just passed is removed from the DB without SendGrid ever being told to cancel — and since SendGrid can't cancel within ~10 min of `send_at`, the email may already be locked/sent. This is the inherent limitation of a 15-min CRON + SendGrid's cancel cutoff, but it's unlogged: a cancelled email that slipped through is silently swept. Either cancel `cancelled` rows first, or exclude `cancelled: true` from that sweep (and log when a cancelled row is dropped past due).

packages/features/ee/workflows/lib/reminders/smsReminderManager.ts
```diff
@@ -174,9 +174,16 @@ export const scheduleSMSReminder = async (
   }
 };
 
-export const deleteScheduledSMSReminder = async (referenceId: string) => {
+export const deleteScheduledSMSReminder = async (reminderId: number, referenceId: string | null) => {
   try {
-    await twilio.cancelSMS(referenceId);
+    if (referenceId) {
+      await twilio.cancelSMS(referenceId);
```
ron-x5labs, line 180: **Non-blocking — SMS cancellation failure is unretried.** `twilio.cancelSMS` and the `prisma.workflowReminder.delete` share one `try`; if `cancelSMS` throws, the `catch` swallows it and the `delete` is skipped, leaving the row `scheduled: true` with no `cancelled` flag — so unlike email, the CRON never retries and the SMS still sends. The deferred-retry design this PR introduces for email wasn't extended to SMS. Split the try (delete the row even if the Twilio cancel fails) or add an equivalent deferred signal.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -950,6 +961,19 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
   let videoCallUrl;
 
   if (originalRescheduledBooking?.uid) {
+    try {
+      // cancel workflow reminders from previous rescheduled booking
+      originalRescheduledBooking.workflowReminders.forEach((reminder) => {
+        if (reminder.method === WorkflowMethods.EMAIL) {
+          deleteScheduledEmailReminder(reminder.id, reminder.referenceId, true);
```
ron-x5labs, line 968: **Suggestion — document the immediate-vs-deferred reschedule choice.** This reschedule path uses `immediateDelete: true` (synchronous SendGrid cancel, no DB write), while the sibling tRPC reschedule in `bookings.tsx:490` uses the deferred `cancelled: true` path. The split is actually *correct*: `EventManager.reschedule` deletes the original booking, and `WorkflowReminder.booking` is `onDelete: Cascade`, so a deferred flag would be cascade-removed before the 15-min CRON could act — only an immediate cancel survives. But nothing here says so, and the two reschedule paths read as inconsistent. A one-line comment would prevent a future maintainer from "fixing" the inconsistency by removing `immediateDelete`.

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -923,63 +901,63 @@ export const workflowsRouter = router({
     if (isSMSAction(step.action) /*|| step.action === WorkflowActions.EMAIL_ADDRESS*/ /*) {
```
ron-x5labs, line 901: **Suggestion — revert unrelated re-indentation of a commented-out block.** This re-indentation (~60 lines, stripping leading whitespace) is entirely inside the commented-out `testAction` procedure (`/*` opens ~L848, `*/` closes ~L1010), so it has zero functional effect but adds substantial diff noise that obscures the real behavioral changes in this file. Revert it to keep the PR diff focused on the cancellation fix.

packages/prisma/migrations/20230217230604_add_cancelled_to_workflow_reminder/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "WorkflowReminder" ADD COLUMN     "cancelled" BOOLEAN;
```
ron-x5labs, line 2: **Suggestion — make `cancelled` non-null with a default.** The column is nullable with no default/backfill, so existing rows become `NULL` and new rows created by `scheduleEmailReminder`/`scheduleSMSReminder` (which never set `cancelled`) stay `NULL`. The queries work today (`cancelled: true` excludes `NULL`), but the nullable-without-default design invites null-vs-false ambiguity in any future code reading `reminder.cancelled`, and the unscheduled-reminders query relies on `scheduled: false` rather than an explicit `cancelled` check. `cancelled Boolean @default(false)` with a backfill (`UPDATE "WorkflowReminder" SET cancelled = false WHERE cancelled IS NULL`) would make the semantics explicit and self-documenting.

ron-x5labs (commented):
## Code Review: Fixes that workflow reminders of cancelled and rescheduled bookings are still sent

### Problem
Cancelled/rescheduled bookings were still sending workflow reminder emails. The previous fix (#6991) misused SendGrid's `DELETE /v3/user/scheduled_sends/{id}` (that endpoint expects a scheduled_send id, not a batch_id), and cancellations accumulated in `scheduled_sends` until SendGrid's 100-pending-cancellation cap was hit, after which new cancellations silently failed.

### Solution Reviewed
Adds a nullable `cancelled` flag to `WorkflowReminder`. Cancelling an already-scheduled email reminder now sets `cancelled: true` (deferred) instead of calling SendGrid immediately; the 15-min `scheduleEmailReminders` CRON cancels flagged reminders due within the next hour via `POST /v3/user/scheduled_sends {batch_id, status: "cancel"}` then deletes the rows. SMS reminders keep eager cancel+delete. The incorrect SendGrid DELETE is removed (correct — POST-cancel is the documented method; referenceId is only ever sent in the POST body, never in a URL).

### Summary
The diagnosis is right and the SendGrid API misuse is correctly fixed, and I confirmed no authorization/IDOR regression (ownership scoping retained in the disable-reminders `findMany` via `booking.userId`, and workflow-ownership checks remain). However, the reliability linchpin — the `cancelled: true` flag write — is fire-and-forget at every callsite, and the CRON cancel loop has no per-iteration isolation or query guards. Transient SendGrid/DB failures can therefore silently leave a reminder uncancellable — the exact class of bug this PR fixes. Note: this PR is already merged, so these are retrospective/advisory findings.

### Files Reviewed
- `handleCancelBooking.ts`, `handleNewBooking.ts`, `scheduleEmailReminders.ts`, `emailReminderManager.ts`, `smsReminderManager.ts`, `bookings.tsx`, `workflows.tsx` — deeply reviewed
- `schema.prisma`, `migration.sql` — lightly reviewed (1-line nullable field + migration)
- `WorkflowStepContainer.tsx` — lightly reviewed (UI refactor gates the verification block on `isPhoneNumberNeeded` only; no behavior concern)

### Verification
- Type-check/lint/tests — skipped: the PR head is a 2023 commit checked out in a fresh worktree with no installed dependencies; running the current toolchain against this snapshot would not reflect the PR and would be unreliable. Findings are from code inspection of the checked-out PR head.

### Issues Found
See inline comments. In brief:
- Non-blocking (6): fire-and-forget `cancelled` flag writes across callsites; `immediateDelete` path has no DB write / no retry on SendGrid failure; CRON cancel loop aborts on first SendGrid error (deletes for already-cancelled rows lost); CRON cancel query missing `method: EMAIL` + `referenceId NOT NULL` guards; CRON top `deleteMany` ignores `cancelled`; SMS cancellation failure swallowed with no retry.
- Suggestions (3): document the immediate-vs-deferred reschedule choice; revert the unrelated re-indentation of the commented-out `testAction` block; use `cancelled Boolean @default(false)` + backfill.

### Verdict
Recommend follow-up hardening — the fix is directionally correct, but deferred-cancellation reliability depends on an *awaited* flag write and per-iteration CRON isolation that are currently missing; transient failures can silently re-introduce the bug.

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -483,29 +481,18 @@ async function handler(req: NextApiRequest & { userId?: number }) {
     cancelScheduledJobs(booking);
   });
 
-  //Workflows - delete all reminders for bookings
-  const remindersToDelete: PrismaPromise<Prisma.BatchPayload>[] = [];
+  //Workflows - cancel all reminders for cancelled bookings
   updatedBookings.forEach((booking) => {
```
ron-x5labs, line 485: 🔴 **Blocking — `cancelled=true` may never be persisted.** This `forEach` calls `deleteScheduledEmailReminder` (now `async`) without awaiting it. The deferred email path's only effect is `prisma.workflowReminder.update({ data: { cancelled: true } })`; since the promise is dropped, `await Promise.all(prismaPromises.concat(apiDeletes))` on line 497 does *not* wait for it. The handler returns `"Booking successfully cancelled"` before the `cancelled` flag is committed, and on Vercel the serverless lambda can freeze right after the response — so the flag is never written and the CRON never cancels the SendGrid batch. **The reminder email still sends**, which is exactly the bug this PR exists to fix. The previous code awaited the DB delete via the collected `remindersToDelete` + `Promise.all`, so this is a regression.

The same un-awaited `forEach` pattern recurs at `bookings.tsx:487`, `workflows.tsx:378/519/574`, and `handleNewBooking.ts:965`.

Suggested fix — collect the promises and await them before returning:
```ts
const reminderCancelPromises: Promise<unknown>[] = [];
updatedBookings.forEach((booking) => {
  booking.workflowReminders.forEach((reminder) => {
    if (reminder.method === WorkflowMethods.EMAIL) {
      reminderCancelPromises.push(deleteScheduledEmailReminder(reminder.id, reminder.referenceId));
    } else if (reminder.method === WorkflowMethods.SMS) {
      reminderCancelPromises.push(deleteScheduledSMSReminder(reminder.id, reminder.referenceId));
    }
  });
});
await Promise.all(reminderCancelPromises);
```

packages/features/ee/workflows/api/scheduleEmailReminders.ts
ron-x5labs, line 44: 🟡 **Missing `method: EMAIL` filter.** Every other `workflowReminder.findMany`/`deleteMany` in this file filters `method: WorkflowMethods.EMAIL`, but `remindersToCancel` does not. Today the SMS path never sets `cancelled=true`, so this is latent — but if SMS ever uses the flag, this loop will pass a Twilio SID into `reminder.referenceId` as a SendGrid `batch_id`, producing confusing 400s and skipped cancellations. Add `method: WorkflowMethods.EMAIL` (and consider `NOT: { referenceId: null }`) to match the sibling queries and keep the contract explicit.

packages/features/ee/workflows/api/scheduleEmailReminders.ts
ron-x5labs, line 56: 🟡 **Single try/catch aborts the whole batch on the first failure.** The `for` loop and the trailing `await Promise.all(workflowRemindersToDelete)` are both inside one `try`. A transient SendGrid 5xx on reminder #2 throws, skipping every later cancel *and* every `prisma.workflowReminder.delete` in that run (the delete query objects are lazy and `Promise.all` is never reached). On the next 15-min run those rows are re-fetched and retried, so already-cancelled reminders get redundant SendGrid cancel calls. Consider per-reminder try/catch so one failure doesn't poison the batch, and delete the row inside its own try so a SendGrid hiccup doesn't leak the DB row.

packages/features/ee/workflows/lib/reminders/emailReminderManager.ts
ron-x5labs, line 213: 🟡 **`immediateDelete=true` orphans the DB row and has no retry on SendGrid failure.** After the SendGrid `POST …/scheduled_sends` cancel this branch `return;`s without deleting or updating the row, so the reminder stays `scheduled=true` with `cancelled=null`. The `catch` swallows the error, so if the SendGrid cancel itself fails there is no `cancelled=true` for the CRON to pick up — the email still sends with no retry. This is asymmetric with both the `!referenceId` branch (which deletes the row) and the SMS counterpart. Either delete the row after a successful immediate cancel, or set `cancelled=true` *before* attempting SendGrid so the CRON acts as the retry safety net:
```ts
if (immediateDelete) {
  await prisma.workflowReminder.update({ where: { id: reminderId }, data: { cancelled: true } });
  await client.request({ url: "/v3/user/scheduled_sends", method: "POST", body: { batch_id: referenceId, status: "cancel" } });
}
```

packages/features/ee/workflows/lib/reminders/smsReminderManager.ts
```diff
@@ -174,9 +174,16 @@ export const scheduleSMSReminder = async (
...
+    }
+    await prisma.workflowReminder.delete({
```
ron-x5labs, line 182: 🟡 **DB delete inside `try` after `twilio.cancelSMS` — Twilio failure lingers the row and SMS still sends.** If `twilio.cancelSMS` throws, the `catch` logs and the `prisma.workflowReminder.delete` never runs, so the reminder row remains scheduled and the scheduled SMS is never re-attempted (there is no SMS equivalent of the email `cancelled=true` + CRON retry). The old caller-driven `deleteMany` at least removed the row independently. Consider cancelling-then-deleting with independent error handling, or adding a deferred-retry path for SMS parity with the new email flow.

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -923,63 +901,63 @@ export const workflowsRouter = router({
...
 const hasTeamPlan = (await ctx.prisma.membership.count({ where: { userId: user.id } })) > 0;
```
ron-x5labs, line 902: 💡 **Accidental indentation regression.** The whole `sendTestReminder`-adjacent block here was de-indented (lost its leading whitespace) relative to the rest of the file — looks like an unintended formatter/editor artifact rather than a deliberate change. Run `yarn biome check --write` on this file before pushing so the diff doesn't carry a large unrelated whitespace churn.

ron-x5labs (changes requested):
## Code Review: Fix workflow emails sent for cancelled/rescheduled bookings

### Problem
PR #6991 misused SendGrid's `DELETE /v3/user/scheduled_sends/{id}` (that endpoint *removes* a cancel, re-scheduling the email). Once 100 pending cancels piled up in `scheduled_sends`, new cancels failed and emails kept firing. This PR adds a `cancelled` flag to `WorkflowReminder` and defers the actual SendGrid cancel to the `scheduleEmailReminders` CRON.

### Solution Reviewed
New nullable `cancelled` column. On cancel/reschedule, email reminders are flagged `cancelled=true` instead of cancelling at SendGrid inline; the 15-min CRON cancels flagged reminders whose `scheduledDate` is within the next hour, then deletes the row. SMS still cancels via Twilio inline. Signature changes to `deleteScheduledEmailReminder/deleteScheduledSMSReminder(reminderId, referenceId, …)` ripple through 7 call sites.

### Summary
The core idea (defer cancel to CRON, avoid the 100-pending-cancel cap) is sound, but the most important path regressed: the call sites that set `cancelled=true` now do so via an un-awaited `forEach`, so the handler returns before the flag is persisted — on Vercel the lambda can freeze and the email still sends, which is exactly the bug this PR exists to fix. Needs that await restored before merge.

### Files Reviewed
- `handleCancelBooking.ts` — deep
- `handleNewBooking.ts` — deep
- `scheduleEmailReminders.ts` — deep
- `emailReminderManager.ts` — deep
- `smsReminderManager.ts` — deep
- `viewer/bookings.tsx` — deep
- `viewer/workflows.tsx` — deep
- `schema.prisma` + migration — light (consistent)
- `WorkflowStepContainer.tsx` — light (UI guard change, intentional-looking)

### Verification
- `yarn biome` / `tsc` — skipped: no `node_modules`/`yarn`/`corepack` in this review environment. Findings are source-grounded from a local worktree checkout of the PR.

### Issues Found
- 🔴 Blocking: `handleCancelBooking.ts:485` — un-awaited `forEach` of async `delete*`; `cancelled=true` may never persist.
- 🟡 Non-blocking: `scheduleEmailReminders.ts:44` — `remindersToCancel` query missing `method: EMAIL` filter.
- 🟡 Non-blocking: `scheduleEmailReminders.ts:56` — one try/catch wraps the whole cancel loop + `Promise.all`; first SendGrid error aborts all remaining cancels and all DB deletes.
- 🟡 Non-blocking: `emailReminderManager.ts:213` — `immediateDelete=true` returns after SendGrid cancel without touching the DB row; on SendGrid failure the row stays `scheduled=true` with no retry.
- 🟡 Non-blocking: `smsReminderManager.ts:182` — DB delete moved inside `try` after `twilio.cancelSMS`; a Twilio failure lingers the row and SMS still sends, with no CRON retry.
- 💡 Suggestion: `workflows.tsx:902` — large de-indented block is an accidental formatting regression; run Biome.

### Verdict
Request changes — restore `await` on the cancellation calls (or collect+`Promise.all`) so `cancelled=true` is persisted before the handler returns; the other issues are robustness gaps worth addressing in the same pass.

packages/features/ee/workflows/api/scheduleEmailReminders.ts
ron-x5labs, line 56: 🟡 Non-blocking: A single `try`/`catch` wraps the whole loop, so if `client.request` for one reminder throws, the loop aborts and the remaining reminders in `remindersToCancel` are skipped this run (only retried ~15 min later on the next CRON). Worse, the already-queued `prisma.workflowReminder.delete` promises for successfully-cancelled reminders are never awaited (`Promise.all` is never reached — Prisma promises are lazy), so those rows linger with `cancelled=true` and get re-cancelled at SendGrid on every subsequent run. Also `reminder.referenceId` is `string | null`; a null value would be sent as `batch_id: null`, 400 from SendGrid and abort the whole batch. Consider a per-iteration guard + try/catch and deleting inline on success:

```ts
for (const reminder of remindersToCancel) {
  if (!reminder.referenceId) continue;
  try {
    await client.request({
      url: "/v3/user/scheduled_sends",
      method: "POST",
      body: { batch_id: reminder.referenceId, status: "cancel" },
    });
    await prisma.workflowReminder.delete({ where: { id: reminder.id } });
  } catch (error) {
    console.log(`Error cancelling reminder ${reminder.id}: ${error}`);
  }
}
```

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -950,6 +961,19 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
   let videoCallUrl;
 
   if (originalRescheduledBooking?.uid) {
+    try {
```
ron-x5labs, line 964: 🟡 Non-blocking: This `try`/`catch` (and the `log.error` below) is unreachable. Both `deleteScheduledEmailReminder` and `deleteScheduledSMSReminder` wrap their entire body in `try { ... } catch (error) { console.log(...) }` and never rethrow, so the promise they return always resolves — the outer `catch` can never fire. That means any failure to cancel the previous booking's reminders during a reschedule is completely silent (only an unstructured `console.log` deep in the manager). Either let the managers propagate (rethrow) so this handler's `log.error` actually runs, or drop the dead outer `try`/`catch` and rely on a structured logger inside the managers instead of `console.log`.

ron-x5labs (commented):
## Code Review: Fixes that workflow reminders of cancelled and rescheduled bookings are still sent

### Problem
Workflow email/SMS reminders were still being sent for cancelled and rescheduled bookings. The previous fix (#6991) misused SendGrid's `DELETE /v3/user/scheduled_sends/{id}` endpoint (which *removes* a cancel, re-enabling the send) and hit SendGrid's 100-pending-cancellation cap, after which new cancellations silently failed.

### Solution Reviewed
Introduces a nullable `cancelled` flag on `WorkflowReminder`. On cancel/reschedule, scheduled email reminders are no longer cancelled at SendGrid inline — instead `cancelled` is set to `true` and a 15-minute CRON (`scheduleEmailReminders`) performs the actual SendGrid cancellation for reminders whose `scheduledDate` falls within the next hour, then deletes the row. SMS reminders are still cancelled+deleted inline via Twilio. The `deleteScheduledEmailReminder` / `deleteScheduledSMSReminder` signatures were widened to take `(reminderId, referenceId, immediateDelete?)` and now own the DB row lifecycle.

### Summary
The new two-phase (flag-then-CRON) approach is sound and correctly addresses the SendGrid cancellation-cap root cause. The design is reasonable and the migration is safe. A couple of robustness/observability gaps in the CRON cancel loop and the reschedule error handling are worth tightening before relying on this path in production.

### Files Reviewed
- `packages/features/bookings/lib/handleCancelBooking.ts` — deeply reviewed
- `packages/features/bookings/lib/handleNewBooking.ts` — deeply reviewed
- `packages/features/ee/workflows/api/scheduleEmailReminders.ts` — deeply reviewed (highest risk: CRON cancel loop)
- `packages/features/ee/workflows/lib/reminders/emailReminderManager.ts` — deeply reviewed
- `packages/features/ee/workflows/lib/reminders/smsReminderManager.ts` — deeply reviewed
- `packages/trpc/server/routers/viewer/bookings.tsx` — deeply reviewed
- `packages/trpc/server/routers/viewer/workflows.tsx` — deeply reviewed (the `testAction` block diff is inside a `/* */` comment — no runtime impact)
- `packages/features/ee/workflows/components/WorkflowStepContainer.tsx` — lightly reviewed (splitting the phone/sender blocks is correct; `isSenderIdNeeded` UI is unaffected)
- `packages/prisma/schema.prisma` + migration — lightly reviewed (nullable `BOOLEAN`, no default, NULL-safe)

### Verification
- `yarn type-check:ci` / tests: skipped — this is a 2023-era branch on a now far-ahead `main`; a fresh dependency install + type-check would not resolve cleanly and is disproportionate for a ~190-line diff. Findings below are grounded in the actual checked-out source at the PR head.

### Verdict
Recommend changes before relying on the CRON path in production — the core fix is correct, but the cancel loop aborts the batch on a single SendGrid failure and the reschedule error handling is unreachable.

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -7,6 +7,7 @@ import type { NextApiRequest, NextApiResponse } from "next";
 import dayjs from "@calcom/dayjs";
 import { defaultHandler } from "@calcom/lib/server";
 import prisma from "@calcom/prisma";
+import { Prisma, WorkflowReminder } from "@calcom/prisma/client";
```
ron-x5labs, line 10: 💡 Suggestion: `Prisma` and `WorkflowReminder` are both used only as types here (`Prisma.Prisma__WorkflowReminderClient<WorkflowReminder, never>`), so per the repo convention in `AGENTS.md` this should be `import type { Prisma, WorkflowReminder } from "@calcom/prisma/client";`.

ron-x5labs:
### Test coverage gap

This PR changes the core reminder-cancellation behavior for cancelled/rescheduled bookings — the exact bug being fixed — yet none of the 9 changed files include tests. Specifically untested:

- **CRON `cancelled=true` → SendGrid-cancel flow** (`scheduleEmailReminders.ts:43-77`): the new two-phase cancel loop, including the batch-abort-on-first-failure edge case flagged in the inline review above. A test here would have caught that a single failing `client.request` skips all subsequent reminders and orphans their DB rows.
- **Reschedule `immediateDelete` path** (`handleNewBooking.ts:963-975`): the new `deleteScheduledEmailReminder(reminder.id, reminder.referenceId, true)` path that cancels previous-booking reminders inline during reschedule — no test verifies the previous booking's reminders are actually cancelled.
- **Cancel-booking reminder cleanup** (`handleCancelBooking.ts:484-493`): the shift from `deleteMany` + inline SendGrid cancel to the new `deleteScheduledEmailReminder`/`deleteScheduledSMSReminder(reminder.id, referenceId)` signatures — no test verifies cancelled bookings no longer send reminders.

Given that the previous fix (#6991) was itself wrong and went undetected, regression tests on these three paths would directly guard against reintroducing the original bug. A focused unit/integration test around the CRON cancel loop (mocking the SendGrid client) is the highest-value addition.

PR #7247, opened 2023-02-21
fix: placeholder in custom event name

### LINKED ISSUE #6591: [CAL-844] /event-type/single - The default for what it says in the "Event name" input does not reflect what's actually happening
When I create a new event type, it says that this is what the Event name is:

When someone actually books the event, the name is:

Which to me looks like

`{event_type_title} with {organsiser_full_name} and {schedulder_full_name}`

This is a good default. What we show in the event type screen input should match what we're actually doing.

Also if `event_type_title` is something a user can use, it should be in this modal:

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-844](https://linear.app/calcom/issue/CAL-844/event-typesingle-the-default-for-what-it-says-in-the-event-name-input)</sub>

G3root:
## What does this PR do?

Fixes #6591 

after:
Screenshot 2023-02-21 at 15-10-29 30min Event Type Cal com

Screenshot 2023-02-21 15:09:38

## Type of change

### DIFF
--- a/apps/web/components/eventtype/EventAdvancedTab.tsx
+++ b/apps/web/components/eventtype/EventAdvancedTab.tsx
@@ -101,6 +101,11 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps, 
     }
   }, [eventType.customInputs]);
 
+  const eventNamePlaceholder = t("meeting_with_user")
+    .replace("{Event type title}", eventNameObject.eventType)
+    .replace("{Scheduler}", eventNameObject.attendeeName)
+    .replace("{Organiser}", eventNameObject.host);
+
   return (
     <div className="flex flex-col space-y-8">
       {/**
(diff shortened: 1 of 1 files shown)


apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -101,6 +101,11 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps,
     }
   }, [eventType.customInputs]);
 
+  const eventNamePlaceholder = t("meeting_with_user")
+    .replace("{Event type title}", eventNameObject.eventType)
```
hariombalhara, line 105: This is a duplicate of the code above at https://github.com/calcom/cal.com/pull/7247/files#diff-e6807e04030a7e08ae8814936a8b27eba0efbb12277eba305e2bda7e25a42396R88. 
Let's extract it out and reuse everywhere.

hariombalhara (changes requested):

apps/web/components/eventtype/EventAdvancedTab.tsx
G3root: pushed those changes

hariombalhara (approved):

PR #7249, opened 2023-02-21
feat: JOIN-27736 custom theme

MichalBaran94:
[[JOIN-27736]](https://joinsolutionsag.atlassian.net/browse/JOIN-27736)


sean-brydon:
Hi @MichalBaran94 were you meant to push this to our repo? Looks like youre trying overriding our theme in your own instance. 

If you run into any issues themeing please let me know

PeerRich:
but this PR is actually a great insight of what files we should be able to change on a database level.

i think stuff as favicon etc. we should pull from a server vs from file, that way it wont cause a merge conflict when upgrading

PR #7274, opened 2023-02-22
Checking availability with recurring count/unit

### LINKED ISSUE #7108: [CAL-1065] Recurring events “no users found”issue
Provide a solution for where there is no availability for one particular instance in a recurring event. 

plan is to book it but call out the event with no availability 

[https://threads.com/34440949007?share=959f7da6-6b5b-4cad-b776-1d4a2a16fd36](https://threads.com/34440949007?share=959f7da6-6b5b-4cad-b776-1d4a2a16fd36)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1065](https://linear.app/calcom/issue/CAL-1065/recurring-events-no-users-foundissue)</sub>

leog:
## What does this PR do?

Nowadays we are not checking different months when a recurring event wants to be booked, resulting in numerous errors when it comes to actually booking. This fix intends to take into account the period of time a recurring event needs to see in the future to get availability.

Fixes #7108 

## Type of change


emrysal:
I believe we tried this before but the performance turned dreadful if I remember rightly 😢

leog:
> I believe we tried this before but the performance turned dreadful if I remember rightly 😢

@sean-brydon is this the way you tried to include recurring dates?

We will need to actually add more logic to this PR to check multiple date/time instead of just the one selected by the user. With this change at least we have availability from calendars to check against.

leog:
Closing as there is no proof this approach would result in any measurable improvement of the problem is trying to solve.

PR #7275, opened 2023-02-22
Make the email sender the event organizer instead of the app name

### LINKED ISSUE #7228: Email sender should be the event organizer instead of the app name

When our customers receive e-mails about an event, the sender is the current app name "Awesome app" instead of the event organizer "John Doe's Hair Design" (these are just example names 🙂). This is confusing for the customer because they did not make an appointment with "Awesome app".

It would be better if the e-mails were sent in the organizer's name. In this example they would be sent from "John Doe's Hair Design". Some e-mail templates already use the organizer name (like the `attendee-location-change-email` template), but this is not consistent in all the templates.

--

### Additional context

--

verstratenbram:
## What does this PR do?

Fixes #7228

## Type of change

## Checklist


PeerRich:
oh sweet. reviewing now. if this works then this is a cool change. just to confirm `<${this.getMailerOptions().from}>` is the users email right? or is this hello@cal.com

PeerRich (changes requested):
i dont think this is wise because that way a lot of email clients will save:

John Doe, hello@cal.com

Anna Doe, hello@cal.com

etc. etc.

I don't think we should merge this

PeerRich:
thank you but this would cause more issues than it solves :(

PR #7284, opened 2023-02-22
Remove event type dupes, assign ownership based on _user_eventtype.'B'

emrysal:
## What does this PR do?

Retroactively set `userId` - and remove all duplicates.


zomars:
How can we test this? Should we run this on dev? PR should we let the deploy do it?

emrysal:
I used database-2 to tests @zomars - you can restore the db2 snapshot and try running it - also you can turn the delete into a select and do some random tests to ensure it's getting the right IDs.

Deploy should handle, the way it is right now should work on the fly without conflict errors.

zomars (approved):
You have my trust

PR #3139, opened 2022-06-21
Turborepo improvements

zomars:
## What does this PR do?

Splits common tasks into individual ones. This will allow us to take advantage of caching and remove redundant task runs.

refs #3136

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [ ] `yarn dx`
- [ ] `yarn dev`
- [ ] `yarn test-e2e`
- [ ] `yarn build`
- [ ] `yarn start`
- [ ] @hariombalhara please check if `yarn turbo run embed-tests-update-snapshots:ci --scope=@calcom/embed-react --concurrency=1` works as expected

PR #7296, opened 2023-02-23
Add embed versions

hariombalhara:
It is important to add embed versions so that packages are very clear on what their dependency's version is.

**embed-react** may not be compatible with all previous versions of **embed-core**

The dependency version was changed to '*' in [this PR](https://github.com/calcom/cal.com/pull/3139), so I hope if build is passing it should be okay to merge.

### DIFF
--- a/packages/embeds/embed-react/package.json
+++ b/packages/embeds/embed-react/package.json
@@ -50,7 +50,7 @@
     "vite": "^4.1.2"
   },
   "dependencies": {
-    "@calcom/embed-core": "*",
-    "@calcom/embed-snippet": "*"
+    "@calcom/embed-core": "^1.1.2",
+    "@calcom/embed-snippet": "^1.0.4"
   }
 }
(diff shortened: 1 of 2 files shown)


PeerRich:
can i merge this or does this need to be tested?

hariombalhara:
I need confirmation from @zomars on this

zomars (changes requested):
By doing this we're opting out of monorepo linking and we're actually installing directly from NPM. I don't think this is desired behavior.

hariombalhara:
So, I get the reasoning to use * now. But embed packages are published to NPM as well and for that, we must have versions as dependencies.
How do you suggest we solve this? @zomars

zomars:
> How do you suggest we solve this? @zomars

Not using versions in the source of truth (this monorepo). NPM versions should be used externally. Not internally. Will close this PR until we figure something out.

PR #7310, opened 2023-02-23
chore: bump node version to v18

anonrig:
This pull request updates node version to 18.


anonrig:
Linter is failing due to old Node.js version. I think that's a self-hosted runner that needs to be updated to Node 18. I can quickly fix this by re-adding Node 16 to supported Node.js versions, but it's always recommended to use LTS versions.

PeerRich:
@joeauyeung can you look into upgradingt buildjet's node?

package.json
```diff
@@ -101,7 +101,7 @@
     ]
   },
   "engines": {
-    "node": ">=16.x",
+    "node": ">=18.x",
```
emrysal, line 104: This should stay at >=16.x to keep NodeJS v16 support, rest of the changes are making sense, will proceed with this PR.

apps/web/package.json
```diff
@@ -19,11 +19,11 @@
     "check-changed-files": "ts-node scripts/ts-check-changed-files.ts"
   },
   "engines": {
-    "node": "16.x || 18.x",
+    "node": "16.x || 18.x || 19.x",
```
emrysal, line 22: Also thinking supporting 19.x requires testing our dependencies for 19.x support and I'm predicting that not all libraries do..

emrysal: 🤷 

```bash
error @boxyhq/saml-jackson@1.8.1: The engine "node" is incompatible with this module. Expected version ">=14.18.1 <=18.x". Got "19.7.0"
```

joeauyeung:
> @joeauyeung can you look into upgrading buildjet's node?

We changed all actions back to `ubuntu-latest` when GH actions were refactored

package.json
emrysal: Decided we're dropping node 16.x support

emrysal (approved):

zomars (approved):

anonrig:
🚀 Amazing work!

PeerRich:
man @anonrig thank you so much! for testing and bringing this up.

@zomars also found a way to make data proxies work for us without forcing it to the self-hosters. [https://app.cal.dev/pro](https://app.cal.dev/pro/30min?duration=30) now runs on data proxies

PR #7323, opened 2023-02-24
Fix Google Calendar sharing guest information logic

joeauyeung:
## What does this PR do?

This PR fixes the logic around sharing attendee information in the calendar event.  The `guestsCanSeeOtherGuests` toggles the "See guest list" option

image

Fixes # (issue)

## Type of change

## How should this be tested?

- See guest list should be enabled when
    - Creating a normal booking
    - Creating a seats event with show attendee information selected
- Guest list should be hidden when
    - Creating a seats event with show attendee information not selected

## Checklist


packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -100,7 +100,13 @@ export default class GoogleCalendarService implements Calendar {
...
           useDefault: true,
         },
-        guestsCanSeeOtherGuests: calEventRaw.seatsShowAttendees,
+        // If there are no seats then share attendee information
+        guestsCanSeeOtherGuests: !calEventRaw.seatsPerTimeSlot
+          ? true
+          : // If there are seats and show attendess is enabled then share attendee information
+          calEventRaw.seatsPerTimeSlot && calEventRaw.seatsShowAttendees
+          ? true
+          : false,
```
alannnc, line 109:
```suggestion
        guestsCanSeeOtherGuests: !!calEventRaw.seatsPerTimeSlot ? calEventRaw.seatsShowAttendees : true
```
We can use a falsy and just use one ternary.
Also the second condition returns a boolean so ternary it's not required there.
` calEventRaw.seatsPerTimeSlot && calEventRaw.seatsShowAttendees`

keithwillcode (approved):
Looks solid. Went ahead and fixed the linting issues as well.

keithwillcode:
@joeauyeung Do we also need the same change [here](https://github.com/calcom/cal.com/pull/7323/files#diff-6bfbaf63c841dc9343dba0c7487e30a546ffe2a63c9d9418379d0dbb88644d27R196)?

joeauyeung:
> @joeauyeung Do we also need the same change [here](https://github.com/calcom/cal.com/pull/7323/files#diff-6bfbaf63c841dc9343dba0c7487e30a546ffe2a63c9d9418379d0dbb88644d27R196)?

Good catch, just added that logic

keithwillcode (approved):
Idea: Could be worth consolidating this logic into 1 function so that if it ever needs to change in the future, it's changed for both the `create` and `update` functions and the risk of missing one or the other is eliminated. Not necessary change though if we like the idea of them being calculated separately for the distinct functions.

PR #7328, opened 2023-02-24
fix: use /avatar url

Udit-takkar:
Fixes https://threads.com/34451412442?s=uayNUHouxZhsWcNbjzUkh6

Screenshot 2023-02-24 at 1 59 38 PM

## Type of change


roae (approved):

apps/web/pages/[user].tsx
```diff
@@ -77,7 +78,7 @@ export default function User(props: inferSSRProps<typeof getServerSideProps> & E
                 size="sm"
                 items={props.users.map((user) => ({
                   alt: user.name || "",
-                  image: user.avatar || "",
+                  image: `${WEBAPP_URL}/${user.username}/avatar.png`,
```
Comment (leog) [inline apps/web/pages/[user].tsx:81]: I think user.avatar can be a base64 image, does the avatar endpoint takes care or that to be able to use it as the only way to access an avatar? @Udit-takkar

Comment (Udit-takkar) [reply apps/web/pages/[user].tsx:81]: @leog yeah this endpoint takes cares of  everything https://github.com/calcom/cal.com/blob/main/apps/web/pages/api/user/avatar.ts#L65

Comment (leog) [reply apps/web/pages/[user].tsx:81]: Gotcha

leog (approved):
LGTM

apps/web/pages/[user].tsx
Comment (zomars) [reply apps/web/pages/[user].tsx:81]: I have some observations here. Can we make it so `user.avatar` returns the full formed `${WEBAPP_URL}/${user.username}/avatar.png` instead? WE might benefit by not sending the whole base64 URI in these responses.

zomars (dismissed):

apps/web/pages/[user].tsx
Comment (Udit-takkar) [reply apps/web/pages/[user].tsx:80]: @zomars what do you mean by returning the full formed ${WEBAPP_URL}/${user.username}/avatar.png ?

currently when we make request to the endpoint /api/user/avatar we fetch the user data from database and user.avatar could be base64 image which is then decoded and sent back.

emrysal (approved):

apps/web/pages/[user].tsx
Comment (zomars) [reply apps/web/pages/[user].tsx]: Let's just avoid sending the full base64 image in the responses

PR #7377, opened 2023-02-24
Fixed href's to go to the right place

emrysal:
## What does this PR do?

The mobile settings page broke as it was dependent on the server side redirect which didn't work anymore (as we  have removed the getServerSideProps from the Profile page)

### DIFF
--- a/apps/web/components/SettingsShell.tsx
+++ b/apps/web/components/SettingsShell.tsx
@@ -10,7 +10,7 @@ import NavTabs from "./NavTabs";
 const tabs = [
   {
     name: "profile",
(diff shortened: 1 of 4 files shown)


PeerRich (approved):

Udit-takkar (approved):
https://user-images.githubusercontent.com/53316345/221262239-76a04389-325c-4564-81b0-cdff28a09da0.mov

LGTM. Just one thing the back button seem to be not working it always redirect to /event-types even though i was on /teams

KATT:
You might want https://github.com/tatethurston/nextjs-routes

PR #7378, opened 2023-02-24
make it throw a http error if organizer is unnamed

vanxh:
## What does this PR do?

Make it throw a http error if organizer is unnamed while handling new booking

Fixes # 7359

## Type of change


zomars:
Thank you for your contribution 🙏 can you please untrack your `yarn.lock` changes?

vanxh:
> Thank you for your contribution 🙏 can you please untrack your `yarn.lock` changes?

Sure, done :)

zomars (changes requested):
Taking a closer look. We DO allow users to just remove their name. I don't think throwing an error here is the proper solution. Will close for now but feel free to discuss this further on a new issue to see why this could be a problem a how could we solve it.

PR #7389, opened 2023-02-24
fix for showAttendees default value

### LINKED ISSUE #7262: [CAL-1129] showAttendees prop should be true in default.
A customer said he is having trouble since this feature was introduced, we need to validate that this props comes with a default value as true and add a test to make sure it works as expected.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1129](https://linear.app/calcom/issue/CAL-1129/showattendees-prop-should-be-true-in-default)</sub>

alannnc:
## What does this PR do?

- Defaults showAttendees to true if seatsPerBooking its null

Fixes #7262 

## Type of change

## How should this be tested?

- [ ] Create a booking and install google calendar.
- [ ] Make a booking to it and calendar event shouldn't have its attendees hidden if seats aren't enabled for that eventType.

## Checklist

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -582,7 +582,8 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
     hideCalendarNotes: eventType.hideCalendarNotes,
     requiresConfirmation: requiresConfirmation ?? false,
     eventTypeId: eventType.id,
-    seatsShowAttendees: !!eventType.seatsShowAttendees,
+    // if seats are not enabled we should default true
+    seatsShowAttendees: !!eventType.seatsPerTimeSlot ? eventType.seatsShowAttendees : true,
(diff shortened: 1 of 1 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -582,7 +582,8 @@ async function handler(req: NextApiRequest & { userId?: number | undefined }) {
     hideCalendarNotes: eventType.hideCalendarNotes,
     requiresConfirmation: requiresConfirmation ?? false,
     eventTypeId: eventType.id,
-    seatsShowAttendees: !!eventType.seatsShowAttendees,
+    // if seats are not enabled we should default true
+    seatsShowAttendees: !!eventType.seatsPerTimeSlot ? eventType.seatsShowAttendees : true,
```
alannnc, line 586: As eventType.seatsPerTimeSlot was default null, it was hiding every attendee at every booking.

zomars (approved):

PR #7411, opened 2023-02-26
Playing around with an i18n country select

emrysal:
## What does this PR do?

WIP Country select to help power self hosted phone number selection. Uses Intl and libphonenumber for translations & getting the country list.


Udit-takkar (commented):
The code is working fine but still we won't be able to guess the region of the booker for the the phone number booking question. falling back to organiser's country is fine but still won't be accurate 100% of the time.

emrysal:
Closing for now, may still want this but we can always revisit.

PR #7420, opened 2023-02-27
adds fix-availability-slot-addition

shafkathullah:
## What does this PR do?

This PR fixes `getNextRange` function which till now calculates next range without converting to `utc`, therefore the bug ie, clicking plus button failing to add new entry will come to a quick halt in regions with larger UTC offset countries like for eg: in India with an offset of +05:30 rather than say CET with an offset of +01:00.

Fixes # (issue)

Fixes `getNextRange` function to calculate next range accordingly, with a halt at range end coming at `11:59pm`.
- Before https://www.loom.com/share/8e4502e963a04624a5cf384334962132
- After https://www.loom.com/share/164a0a5c8a79408ca90fd6569f0d3949

**Environment**: Production

## Type of change

## How should this be tested?

- Test best by changing your system timezone to IST +05:30, you will be able to experience above mentioned issue and could test the fix as well.
- The fix doesn't change the return data type in any way as I tested that particularly.


zomars (approved):

Code LGTM will test on the preview deploy before merging.

PR #7422, opened 2023-02-27
refactor: cleanup use of rawInput and pipe mdwr in trpc router

juliusmarminge:
## What does this PR do?

Fixes # (issue)

Cleans up some "low-hanging" fruit after the v10 migration. We now have multiple input parsers so you don't have to fiddle around with parsing the rawInput on your own, and also the ability to chain middlewares to reduce DRY

## Type of change

## How should this be tested?

Just refactored some pieces - existing test suite should be sufficient

- [ ] Test A
- [ ] Test B


packages/trpc/server/routers/viewer/webhook.tsx
```diff
@@ -19,60 +19,55 @@ const webhookIdAndEventTypeIdSchema = z.object({
...
-  const webhookIdAndEventTypeId = webhookIdAndEventTypeIdSchema.safeParse(rawInput);
-  if (!webhookIdAndEventTypeId.success) {
-    throw new TRPCError({ code: "PARSE_ERROR" });
-  }
-  const { eventTypeId, id } = webhookIdAndEventTypeId.data;
+const webhookProcedure = authedProcedure
+  .input(webhookIdAndEventTypeIdSchema.optional())
+  .use(async ({ ctx, input, next }) => {
+    // Endpoints that just read the logged in user's data - like 'list' don't necessary have any input
+    if (!input) return next();
```
juliusmarminge, line 26: here we also use multiple input parsers instead of `rawInput`.

Makes sure that if an input is passed, it follows the expected shape

packages/trpc/server/routers/viewer/webhook.tsx
```diff
@@ -19,60 +19,55 @@ const webhookIdAndEventTypeIdSchema = z.object({
...
+    const { eventTypeId, id } = input;
```
juliusmarminge, line 27: user passed an input, it is then validated automatically

packages/trpc/server/routers/viewer/webhook.tsx
```diff
@@ -19,60 +19,55 @@ const webhookIdAndEventTypeIdSchema = z.object({
...
-});
+    return next();
+  });
 
 export const webhookRouter = router({
   list: webhookProcedure
     .input(
       z
         .object({
-          eventTypeId: z.number().optional(),
```
juliusmarminge, line 75: this prop is now included from the `webhookProcedure`'s input parser

packages/trpc/server/createContext.ts
```diff
@@ -143,5 +143,3 @@ export const createContext = async ({ req, res }: CreateContextOptions, sessionG
     res,
   };
 };
-
-export type Context = trpc.inferAsyncReturnType<typeof createContextInner>;
```
juliusmarminge, line 147: you were only using this type in the `initTRPC` call, which now accepts just passing the function directly so no need to keep this extra type laying around - personal preference perhaps..

packages/trpc/server/trpc.ts
```diff
@@ -32,18 +31,15 @@ const isAuthedMiddleware = t.middleware(({ ctx, next }) => {
   });
 });
 
-const isAdminMiddleware = t.middleware(({ ctx, next }) => {
-  if (!ctx.user || !ctx.session || ctx.user.role !== "ADMIN") {
+const isAdminMiddleware = isAuthedMiddleware.unstable_pipe(({ ctx, next }) => {
```
juliusmarminge, line 34: we can pipe middlewares so that you don't have to repeat the same logic for multiple middlewares.

juliusmarminge: marked unstable cause api could change in the future but very well tested

packages/trpc/server/routers/viewer/eventTypes.ts
```diff
@@ -764,9 +771,6 @@ export const eventTypesRouter = router({
         webhooks: _webhooks,
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
         schedule: _schedule,
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore - not typed correctly as its set on SSR
-        descriptionAsSafeHTML: _descriptionAsSafeHTML,
```
zomars, line 769: I think we still need this

packages/trpc/server/trpc.ts
zomars: This is great!

zomars:
Loving the improvements so far @juliusmarminge I'm just stumbling with a type error locally: 
image

zomars (approved):
Ship it

juliusmarminge:
> Loving the improvements so far @juliusmarminge I'm just stumbling with a type error locally: image

Whoops. Just noticed you inferred the context type from the `createContextInner` function, and then i put the `createContext` function into `initTRPC`. Will fix it real quick!

PR #7449, opened 2023-03-01
fix: use inner context type

### LINKED ISSUE #7441: The signature expects req and res. Which we don't have in an SSG context.
https://github.com/calcom/cal.com/blob/e4b766b9d60bd0bb5f06e14ebecd05622c452cf1/apps/web/server/lib/ssg.ts#L29

```typescript

  const ssg = createProxySSGHelpers({
    router: appRouter,
    transformer: superjson,
    // @ts-expect-error FIXME The signature expects req and res. Which we don't have in an SSG context.
    ctx: {
      prisma,
      session: null,

```

juliusmarminge:
Closes #7441

Ref: https://github.com/calcom/cal.com/pull/7422#issuecomment-1449940507


juliusmarminge:
CleanShot 2023-03-01 at 12 31 01@2x

You we're inferring the context type from the inner function before so this makes sure the type is the same as it was before and not require req/res for ssg

juliusmarminge:
Although I wonder if this is actually 100% accurate typing 🤔

emrysal (approved):
Thank you 🙏

PR #7458, opened 2023-03-01
app: cancel politely (tentatively)

PeerRich:
inspired by https://twitter.com/andrew__reed/status/1630639129146048512

a new "cancellation" method that tentatively cancels an event but only fully cancels it, if the other person also wants to tentatively cancel it.

If only one person tentatively cancels, then the event stays active.

demo: 

https://www.loom.com/share/91115751c31548f0b523898a7215beef


PeerRich:
not a priority for now

PR #7459, opened 2023-03-02
Adds basic global feature flags

### LINKED ISSUE #7304: [CAL-1143] RFC: Feature Flags
right now to disable a certain part of the app you need to comment the code out, which causes merge conflicts.

it would be great to have an area inside [console.cal.com](http://console.cal.com) where you can activate and de-activate certain features for your self-hosted platform product

* every link in navigation (i.e. disable "App Store" or "Workflows" link)
* send emails
* ...

what else?

EDIT: by @zomars

## Types of feature flags

Basically we have 4 kinds of feature flags (taken from this [nice article](https://www.martinfowler.com/articles/feature-toggles.html)):

* Release Flags
* Experiment Flags
* Operation Flags
* Permission Flags

### Release flags

These are used to turn certain features on or off. This helps us work on new features without breaking anything that's already working. Product managers can also use Release Flags to hide new features from users until they're ready. Release Flags are temporary and usually only last for a short time, and they're easy to change if needed. Using Release Flags is a way to make sure that new things are added to software without causing problems for users.

### Experiment flags

These are used to test different versions of a feature to see which one works better. We can put different people into different groups and send them down different paths to see what happens. This helps us figure out what people like and what they don't like, and we can use this information to make things better. Experiment Toggles need to stay in place for long enough to get lots of information, but not too long that the information becomes outdated. They're always changing because different people use them all the time.

### Operation Flags

These can turn parts of the system on and off, just in case there's a problem. For example, if we add a new feature to our system and we're not sure if it will work properly, we can use an Ops Toggle to quickly turn it off if we need to. Most Ops Toggles are only used for a short time, but some are kept for a long time, like a "Kill Switch," which helps us turn off less important parts of the system when there's too much work to do. It's important to be able to change these buttons quickly in case something goes wrong, so we don't have to wait too long to fix things.

### Permission Flags

These control which users can access certain features or product experiences. For example, we may have special features that only paying customers can use, or we may have features that are only available to our internal team or a specific group of beta testers. This technique of turning on new features for a set of internal or beta users is called a Champagne Brunch. It's similar to a Canary Release, which exposes new features to a randomly selected group of users. Permissioning Toggles may be used for a long time, especially when managing features that are only accessible to premium users. This is because permissions are user-specific, so the toggle decision needs to be made for each user request.

## Possible usage examples (taking inspiration from happykit)

On client:

```ts
// pages/index.js
import { useFlags } from "@calcom/features/flags/client";

export default function IndexPage(props) {
  const { flags } = useFlags();

  return flags.new-booking === "true"
    ? "You get the new booking"
    : "You get the old booking";
}
```

On server:

```ts
import { GetServerSideProps } from "next";
import { getFlags } from "@calcom/features/flags/server";

export const getServerSideProps = async (
  context
) => {
  const { flags, data } = await getFlags({ context });
  return { props: { flags, data } };
};
```

On middleware:

```ts
import { NextRequest, NextResponse } from "next/server";
import { getEdgeFlags } from "@calcom/features/flags/edge";

export const config = { matcher: "/demo/middleware" };

export async function middleware(request: NextRequest) {
  const { flags, cookie } = await getEdgeFlags({ request });

  const nextUrl = request.nextUrl.clone();
  nextUrl.pathname = `/demo/middleware/${flags?.checkout || "full"}`;
  const response = NextResponse.rewrite(nextUrl);

  if (cookie) response.cookies.set(...cookie.args);

  return response;
}
```

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1143](https://linear.app/calcom/issue/CAL-1143/rfc-feature-flags)</sub>

zomars:
## What does this PR do?

- Fixes #7304 
- Adds basic global feature flags that allow us to
  - Disable sending emails (kill switch)
  - Enable Workflows (operational switch)
  - Enable Teams (operational switch)
  - Enable Webhooks (operational switch)

So far operational switches only hides the navigation bar. We still need to add more checks in the core logic.

TODO:
- Allow checking flags per user (override globals/permissions)
- Find more flags to add

## Type of change

## How should this be tested?

- [ ] As an admin user, go to `/settings/admin/flags`
- [ ] Enable email kill switch, emails should not be sent anymore
- [ ] Disable workflows, they should disappear from nav bar (TODO: Prevent accessing directly via URL)
- [ ] Disable teams, they should disappear from nav bar (TODO: Prevent accessing directly via URL)
- [ ] Disable webhooks,  they should disappear from settings nav bar (TODO: Prevent accessing directly via URL)

## Checklist


packages/emails/templates/_base-email.ts
```diff
@@ -28,7 +30,13 @@ export default class BaseEmail {
...
     return {};
   }
-  public sendEmail() {
+  public async sendEmail() {
+    const featureFlags = await getFeatureFlagMap(prisma);
+    /** If email kill switch exists and is active, we prevent emails being sent. */
+    if (featureFlags.emails) {
+      console.warn("Skipped Sending Email due to active Kill Switch");
+      return new Promise((r) => r("Skipped Sending Email due to active Kill Switch"));
+    }
```
zomars, line 39: Adds basic email kill switch as example and in case it's needed

packages/features/flags/components/FlagAdminList.tsx
```diff
@@ -0,0 +1,49 @@
+import { trpc } from "@calcom/trpc/react";
+import type { RouterOutputs } from "@calcom/trpc/react";
+import { Badge, List, ListItem, ListItemText, ListItemTitle, Switch } from "@calcom/ui";
+
+export const FlagAdminList = () => {
+  const [data] = trpc.viewer.features.list.useSuspenseQuery();
```
zomars, line 6: Using new suspense queries

packages/features/flags/components/FlagAdminList.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <div className="flex flex-1 flex-col">
+            <ListItemTitle component="h3">
+              {flag.slug}
+              &nbsp;&nbsp;
+              <Badge variant="green">{flag.type}</Badge>
+            </ListItemTitle>
+            <ListItemText component="p">{flag.description}</ListItemText>
+          </div>
+          <div className="flex py-2">
+            <FlagToggle flag={flag} />
```
zomars, line 20: Self contained toggle to enable/disable features

packages/features/flags/config.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Right now we only support boolean flags.
+ * Maybe later on we can add string variants or numeric ones
+ **/
+export type AppFlags = {
```
zomars, line 5: Since we cannot type DB slugs, we need to keep a record here for type safety.

packages/features/flags/context/provider.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Generic Feature Flags
+ *
+ * Entries consist of the feature flag name as the key and the resolved variant's value as the value.
+ */
+export type Flags = AppFlags;
+
+/**
+ * Allows you to access the flags from context
+ */
+const FeatureContext = React.createContext<Flags | null>(null);
```
zomars, line 15: So we can fetch once, use everywhere they're needed.

packages/features/flags/hooks/index.ts
```diff
@@ -0,0 +1,17 @@
+import { trpc } from "@calcom/trpc/react";
+
+// TODO: Make the slugs type safe
+export function useFlagSuspense(slug: string) {
```
zomars: To be used mostly on `packages/features/flags`. Elsewhere we should use the context on client or the server helpers.

packages/features/flags/pages/flag-listing-view.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { FlagAdminList } from "../components/FlagAdminList";
+
+export const FlagListingView = () => {
+  return (
+    <>
+      <Meta title="Feature Flags" description="Here you can toggle your Cal.com instance features." />
+      <Suspense fallback={<FiLoader />}>
+        <FlagAdminList />
+      </Suspense>
```
zomars, line 14: Suspense!

packages/features/flags/server/router.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { getFeatureFlagMap } from "./utils";
+
+export const featureFlagRouter = router({
+  list: publicProcedure.query(async ({ ctx }) => {
+    const { prisma } = ctx;
+    return prisma.feature.findMany({
+      orderBy: { slug: "asc" },
+    });
+  }),
```
zomars, line 13: For admin listing

packages/features/flags/server/router.ts
```diff
@@ -0,0 +1,36 @@
...
+  map: publicProcedure.query(async ({ ctx }) => {
+    const { prisma } = ctx;
+    return getFeatureFlagMap(prisma);
+  }),
```
zomars, line 17: To use in context

packages/features/flags/server/utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { PrismaClient } from "@prisma/client";
+
+import type { AppFlags } from "../config";
+
+export async function getFeatureFlagMap(prisma: PrismaClient) {
```
zomars, line 5: To be reused on server/trpc calls.

packages/features/tailwind.config.js
```diff
@@ -0,0 +1,6 @@
+const base = require("@calcom/config/tailwind-preset");
```
zomars, line 1: Se we can use tailwind auto-complete in features.

packages/lib/isKeyInObject.ts
```diff
@@ -0,0 +1 @@
+export const isKeyInObject = <T extends object>(k: PropertyKey, o: T): k is keyof T => k in o;
```
zomars, line 1: So we can ensure a key exists in a object on both runtime and type land.

packages/prisma/migrations/20230303195432_add_feature_flag_default_values/migration.sql
```diff
@@ -0,0 +1,29 @@
+-- Insert initial feature flags with their default values
```
zomars, line 1: Since we don't have UI to add new keys and we need to code them anyways. We should add them via migrations.

packages/trpc/react/trpc.ts
```diff
@@ -15,7 +16,7 @@ import type { AppRouter } from "../server/routers/_app";
  * A set of strongly-typed React hooks from your `AppRouter` type signature with `createTRPCReact`.
  * @link https://trpc.io/docs/v10/react#2-create-trpc-hooks
  */
-export const trpc = createTRPCNext<AppRouter>({
+export const trpc = createTRPCNext<AppRouter, NextPageContext, "ExperimentalSuspense">({
```
zomars, line 19: So we can use clean up our queries using suspense

zomars (commented):
Self review done!

packages/features/flags/server/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+  const flags = await prisma.feature.findMany({
+    orderBy: { slug: "asc" },
+  });
+  return flags.reduce<AppFlags>((fs, f) => {
+    fs[f.slug as keyof AppFlags] = f.enabled;
+    return fs;
+  }, {} as unknown as AppFlags);
```
sean-brydon, line 12: Do we need this `unknown` seems to work fine for me 

```suggestion
  return flags.reduce<AppFlags>((acc, flag) => {
    acc[flag.slug as keyof AppFlags] = flag.enabled;
    return acc;
  }, {} as AppFlags);
```

roae:
@zomars what is the reason for this increase in page size?

imagen

zomars:
> @zomars what is the reason for this increase in page size?
> 
> imagen

That's a good question. I can't seem to find a specific reason. Will analyze locally.

zomars:
> > @zomars what is the reason for this increase in page size?
> > imagen
> 
> That's a good question. I can't seem to find a specific reason. Will analyze locally.

@roae seems like the bundle size normalized itself 🤔

roae (approved):
It works for me, It's a solid start!

packages/features/flags/pages/flag-listing-view.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { Meta } from "@calcom/ui";
+import { FiLoader } from "@calcom/ui/components/icon";
+
+import { FlagAdminList } from "../components/FlagAdminList";
+
+export const FlagListingView = () => {
+  return (
+    <>
+      <Meta title="Feature Flags" description="Here you can toggle your Cal.com instance features." />
```
alannnc, line 11: NIT: translation

packages/features/flags/config.ts
```diff
@@ -0,0 +1,11 @@
...
+  emails: boolean;
+  teams: boolean;
+  webhooks: boolean;
+  workflows: boolean;
+  "booking-page-v2": boolean;
```
alannnc, line 10: can we improve this to be v2-booking-page? As mentioned in a example down below v2-workflows

alannnc (approved):
Code looks clean! Excited about this.
Worked for me

PR #7505, opened 2023-03-03
Fixes formatting issues with lists in event type description and bio

CarinaWolli:
## What does this PR do?

Lists were disabled in the editor for the bio as we don't have a good way to display it, because we have the text centered like that: 
Screenshot 2023-03-03 at 17 33 51

Ordered and underordered list now for event type description: 
Screenshot 2023-03-03 at 17 35 19

Screenshot 2023-03-03 at 17 36 43

Screenshot 2023-03-03 at 17 36 59

## Type of change

### DIFF
--- a/packages/lib/markdownIt.ts
+++ b/packages/lib/markdownIt.ts
@@ -1,0 +1,12 @@
+import MarkdownIt from "markdown-it";
+
+export const md = new MarkdownIt("default", { html: true, breaks: true, linkify: true });
+
+export function addListFormatting(html: string) {
+  return html
+    .replaceAll("<ul>", "<ul style='list-style-type: disc; list-style-position: inside; margin-left: 12px'>")
+    .replaceAll(
+      "<ol>",
+      "<ol style='list-style-type: decimal; list-style-position: inside; margin-left: 12px'>"
+    );
+}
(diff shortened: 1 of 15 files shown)


packages/lib/markdownIt.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const md = new MarkdownIt("default", { html: true, breaks: true, linkify: true });
+
+export function addListFormatting(html: string) {
+  return html
+    .replaceAll("<ul>", "<ul style='list-style-type: disc; list-style-position: inside; margin-left: 12px'>")
+    .replaceAll(
+      "<ol>",
+      "<ol style='list-style-type: decimal; list-style-position: inside; margin-left: 12px'>"
+    );
```
PeerRich, line 11: smart

PeerRich (approved):
LGTM

zomars (approved):

packages/lib/markdownIt.ts
```diff
@@ -0,0 +1,12 @@
+import MarkdownIt from "markdown-it";
+
+export const md = new MarkdownIt("default", { html: true, breaks: true, linkify: true });
+
+export function addListFormatting(html: string) {
+  return html
+    .replaceAll("<ul>", "<ul style='list-style-type: disc; list-style-position: inside; margin-left: 12px'>")
+    .replaceAll(
+      "<ol>",
+      "<ol style='list-style-type: decimal; list-style-position: inside; margin-left: 12px'>"
```
roae, line 10: why didn't you use tailwind here?

zomars: Tailwind won't detect usage in this context as far as I know.

roae (approved):
looks great

packages/lib/markdownIt.ts
roae: isn't that path in the "content" config of tailwind?

PR #7540, opened 2023-03-06
Fix: Filter Timezones by cities

gitstart-calcom:
The isProblematicTimezone function has been added, which ensures that none of the timezones that break react-timezone-select are included in the timezones prop. With that, the problem of not finding certain cities like “New York” should be solved

## Demo
https://www.loom.com/share/041fcd6982ae4968b201a3e4bc88de06


emrysal (approved):
Looks good! Thanks for this

PR #7544, opened 2023-03-06
Fix/incorrect timezone on booking success

roae:
## What does this PR do?

On the booking confirmation page the browser timezone is displayed and should show the timezone selected by the user.

Captura de pantalla 2023-03-06 a la(s) 1 41 56 p m

Result

Captura de pantalla 2023-03-06 a la(s) 2 37 39 p m


emrysal (approved):
Looks good!

roae:
> @roae Should we maybe show the logged-in user's timezone if the user is coming from /bookings to see the booking details?

I think it should be displayed in the
Time zone of the event

PR #7549, opened 2023-03-07
fixes #7548 - Scroll bar because auto height adjustment doesn't take <0.5 values into account

### LINKED ISSUE #7548: Scroll bar because auto height adjustment doesn't take <0.5 values into account
### Issue Summary

Scroll bar because auto height adjustment doesn't take <0.5 values into account

**Here you can see that on the right the CSS height is `424.297px` but the height of the Iframe only got updated to `424px`**

image

**Result for the bug is this ugly scrollbar**

image

## Solution sugestion
`(i.style.height = o.iframeHeight + n),` would need to up round the `o.iframeHeight`

image

LarsArtmann:
Fixes # (issue)
Fix #7548 - Scroll bar because auto height adjustment doesn't take <0.5 values into account

## Type of change

## How should this be tested?

- [ ] Check If **height** & width updates in iframe **inline** still works fine

## Checklist


packages/embeds/embed-core/src/embed-iframe.ts
```diff
@@ -341,14 +341,8 @@ function keepParentInformedAboutDimensionChanges() {
     }
 
     const mainElementStyles = getComputedStyle(mainElement);
-    const contentHeight =
-      mainElement.offsetHeight +
-      parseInt(mainElementStyles.marginTop) +
-      parseInt(mainElementStyles.marginBottom);
```
hariombalhara, line 347: Oh, I should have used parseFloat.
@LartyHD Thanks. But it I don't think .height and .width consider margin.

hariombalhara (changes requested):

LarsArtmann (commented):
Added marginTop & marginBottom back in the contentHeight calculation. This time with parseFloat instead of parseInt

packages/embeds/embed-core/src/embed-iframe.ts
hariombalhara: @LartyHD Thanks. Also, I couldn't replicate the scroll scenario you screenshotted. Is there a page where I can see it in action?

hariombalhara: What is the motivation behind using .height instead of .offsetHeight. I think offsetHeight includes padding and border as well which we need.

LartyHD: You can check it out here: [https://nobletary-dqm7oxxac-larsartmann.vercel.app/en/appointment](https://nobletary-dqm7oxxac-larsartmann.vercel.app/en/appointment)
(The current version overflows to the footer, but that can be ignored)

LartyHD: I hope that clarifies things

LartyHD: Adding an `overflow-y: hidden` could possibly also be an alternative solution.

LartyHD: The other problem is that it only happens sometimes.
I recorded a video, at second 14 I manage to make the sidebar appear.
https://user-images.githubusercontent.com/23587853/224332473-f7b33baa-5149-4393-9cfb-cb89a86825c0.mov

hariombalhara: @LartyHD Thanks. I could replicate the issue now 🙏

My concern was that .height won't include padding and border. But that happens in the default behaviour of box-sizing(which is content-box). Tailwind explicitly set `box-sizing` to `border-box` in which case it includes padding and border.

So, in this case .height is better than offsetHeight in giving the correct value of height.

packages/embeds/embed-core/src/embed-iframe.ts
```diff
@@ -341,15 +341,15 @@ function keepParentInformedAboutDimensionChanges() {
...
-    const contentWidth =
-      mainElement.offsetWidth +
-      parseInt(mainElementStyles.marginLeft) +
-      parseInt(mainElementStyles.marginRight);
-
+    const contentHeight = Math.ceil(
```
hariombalhara, line 344:
```suggestion
    // Use, .height as that gives more accurate value in floating point. Also, do a ceil on the total sum so that whatever happens there is enough iframe size to avoid scroll.
    const contentHeight = Math.ceil(
```

hariombalhara (approved):

PR #7553, opened 2023-03-07
fix: overflow issue in event type form builder and booking page

### LINKED ISSUE #7551: Option to scroll is missing
There is no option to scroll at the booking questions anymore. Firstly in the dashboard when creating longer lists of radio buttons, checklists, dropdowns, etc. On the other hand, the appointment booking pages themselves are also affected.

Bildschirm­foto 2023-03-07 um 03 42 03

Bildschirm­foto 2023-03-07 um 03 48 53

G3root:
## What does this PR do?

Fixes #7551 

after:

Screenshot 2023-03-07 12:48:17

Screenshot 2023-03-07 12:48:37

## Type of change

### DIFF
--- a/packages/features/form-builder/FormBuilder.tsx
+++ b/packages/features/form-builder/FormBuilder.tsx
@@ -368,7 +368,7 @@ export const FormBuilder = function FormBuilder({
             fieldIndex: -1,
           })
         }>
-        <DialogContent>
+        <DialogContent enableOverflow>
           <DialogHeader title={t("add_a_booking_question")} subtitle={t("form_builder_field_add_subtitle")} />
           <div>
             <Form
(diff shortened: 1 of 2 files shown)


packages/features/form-builder/FormBuilder.tsx
```diff
@@ -368,7 +368,7 @@ export const FormBuilder = function FormBuilder({
             fieldIndex: -1,
           })
         }>
-        <DialogContent>
+        <DialogContent enableOverflow>
```
hariombalhara, line 371: @G3root Why would we not want this to be enabled always?

G3root: i think enabling this by default broke some modals 🤔 
cc: @Udit-takkar

hariombalhara: Even if that is the case, I think we should disable it if needed. But this behaviour is something that should be enabled by default.

hariombalhara (approved):

PR #6900, opened 2023-02-06
Some date overrides ignored in availability check

mischarouleaux:
## What does this PR do?

This is a fix for the following issue: https://github.com/calcom/cal.com/issues/6622

When an user has date overrides, these will be taken into consideration when someone books a meeting.

Fixes # (6622)

Result without the code fix:
Screenshot 2023-02-06 at 15 12 19

Result with the code fix:
Screenshot 2023-02-06 at 15 12 36

**Environment**: Main branch

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [ ] Test A: Create a date override, and place a new booking within that timespan.
- [ ] Test B: Do the same, but now a timespan before the date override.
- [ ] Test C: Do the same, but now a timespan after the date override.
- [ ] Test D: No days available and no date overrides, user shouldn't be able to book the meeting
- [ ] Test E: Book a normal meeting with no date overrides.

## Checklist
- I haven't checked if my PR needs changes to the documentation

/claim #6622

PR #7559, opened 2023-03-07
Fix "User Added Question" Label in Email and Calendar Invite

### LINKED ISSUE #7493: Email shows question name instead of question label for a User added question
For a user added question, email shows the question `name` as label. It should be the question label only.
Screenshot 2023-03-03 at 5 41 05 PM

Email - Wrong
Screenshot 2023-03-03 at 5 40 47 PM

Success Page(booking/[uid]) - Correct
Screenshot 2023-03-03 at 5 40 55 PM

The behaviour is like this because we send same payload to Webhook and Email. Webhook needs to have name as the object key.

**The issue is only in main right now.**

hariombalhara:
Fixes #7493

## What does this PR do?

Fixes # (issue)

**Environment**:  Production

## Type of change

## How should this be tested?

- [x] Simply add user field and check that label is shown in Email and Calendar Invite. Also check that webhook has both label and value

## Checklist

### DIFF
--- a/packages/emails/src/components/UserFieldsResponses.tsx
+++ b/packages/emails/src/components/UserFieldsResponses.tsx
@@ -4,13 +4,20 @@ import { Info } from "./Info";
 
 export function UserFieldsResponses(props: { calEvent: CalendarEvent }) {
   const { customInputs, userFieldsResponses } = props.calEvent;
-  const responses = userFieldsResponses || customInputs;
-  if (!responses) return null;
+  let labelValueMap: Record<string, string | string[]> = {};
+  if (userFieldsResponses) {
+    for (const [, value] of Object.entries(userFieldsResponses)) {
+      labelValueMap[value.label] = value.value;
+    }
+  } else {
+    labelValueMap = customInputs as Record<string, string | string[]>;
+  }
+  if (!labelValueMap) return null;
   return (
     <>
-      {Object.keys(responses).map((key) =>
-        responses[key] !== "" ? (
-          <Info key={key} label={key} description={`${responses[key]}`} withSpacer />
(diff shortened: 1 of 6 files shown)


packages/types/Calendar.d.ts
```diff
@@ -165,10 +165,22 @@ export interface CalendarEvent {
   seatsPerTimeSlot?: number | null;
 
   // It has responses to all the fields(system + user)
-  responses?: Prisma.JsonObject | null;
+  responses?: Record<
```
hariombalhara, line 168: Because CalendarEvent payload is common for Email, Calendar and Webhook, send it in this format. It would allow webhook payload to uniquely identify any field's response even if label changes. It also allows, Email and Calendars to show the current value of label instead of the name for the field

packages/emails/src/components/UserFieldsResponses.tsx
```diff
@@ -4,13 +4,20 @@ import { Info } from "./Info";
 
 export function UserFieldsResponses(props: { calEvent: CalendarEvent }) {
   const { customInputs, userFieldsResponses } = props.calEvent;
-  const responses = userFieldsResponses || customInputs;
-  if (!responses) return null;
+  let labelValueMap: Record<string, string | string[]> = {};
```
emrysal, line 7: This should be within a function & reused. Easier typing too

emrysal (changes requested):

emrysal (approved):
⚡ LGTM

zomars:
We need to update E2E tests. [It is broken on main](https://github.com/calcom/cal.com/pull/6900/files#annotation_9489476712)

PR #7567, opened 2023-03-07
Only show smsReminderNumber when workflow has ATTENDEE_NUMBER as action

### LINKED ISSUE #7640: Workflow creates mandatory SMS field
I do have an event type with two workflows attached. Those two workflows do send E-Mails to the customer if the event date changes and a reminder 24 hours before the event starts.

You can see the workflow setting here:
2023-03-09 16_40_48-24h Reminder _ Cal com – Mozilla Firefox

Although I have no option to send SMS the system creates fields to enter phone numbers for SMS notifications in my event like that

2023-03-09 16_40_18-Ninja Warrior Einzeltraining _ Termintyp _ Cal com – Mozilla Firefox

2023-03-09 16_44_20-Bestätigen Sie Ninja Warrior Einzeltraining mit Parkour Austria _ Cal com – Mozi

I can't toggle those fields off or delete them, as long as the event type is connected with this E-Mail notification workflow.

Those Telephone Number Fields should not automatically be attached to event-types, when not in use.

thanks in advance for looking into it

CarinaWolli:
## What does this PR do?

This PR makes sure that the booking question _Phone number (SMS notifications)_ is only show when the event type is active on an workflow that has the action _send SMS to attendee_

Fixes: #7640

## Type of change

## How should this be tested?
Try the following things and check if _Phone number (SMS notifications)_ is correctly shown or not shown as a booking question
- Adding, editing and deleting steps 
- Setting and removing active event type 
- Active/deactivate workflows in `event-types/[id]?tabName=workflows` settings
- Deleting a workflow


packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -413,6 +416,19 @@ export const workflowsRouter = router({
         }
       });
 
+      for (const activeRef of workflowToDelete.activeOn) {
```
hariombalhara, line 419: Oh, I missed this. I thought I had handled deletion.
But we should use `removeSmsReminderFieldForBooking` as that handles these hardcoded values.

CarinaWolli: Thank you, fixed 👍

packages/trpc/server/routers/viewer/workflows.tsx
```diff
@@ -907,6 +930,20 @@ export const workflowsRouter = router({
...
+                workflowId: id,
+                isSmsReminderNumberRequired: steps.some(
+                  (s) => s.action === WorkflowActions.SMS_ATTENDEE && s.numberRequired
+                ),
+                eventTypeId,
+              });
+            } else {
+              await removeSmsReminderFieldForBooking({ workflowId: id, eventTypeId });
+            }
+          }
```
hariombalhara, line 946: @CarinaWolli I just noticed that this is too much tracking that we have to do, due to this requirement of adding/updating/removing field in bookingFields.
One simplification I just noticed is that we can move this code at the very last of the fn(say line 1116). This would allow us to remove newly added code at 773, 732. This is possible because `activeOn` is the absolute truth we are going to save in DB.
Apart from that if something get's removed from activeOn, that's already handled at line 589.

packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -282,6 +274,18 @@ export const ensureBookingInputsHaveSystemFields = ({
 
   bookingFields = missingSystemBeforeFields.concat(bookingFields);
 
+  // Backward Compatibility for SMS Reminder Number
+  // Note: We still need workflows in `getBookingFields` due to Backward Compatibility. If we do a one time entry for all event-types, we can remove workflows from `getBookingFields`
+  // Also, note that even if Workflows don't explicity add smsReminderNumber field to bookingFields, it would be added as a side effect of this backward compatibility logic
```
hariombalhara, line 279: @CarinaWolli Also wanted to bring to your notice that, even if we remove all that CRUD logic for `smsReminderNumber` field from workflows router, the field will still be shown due to this logic.

This logic is supposed to be temporary. The intention is to remove the requirement of workflows in bookingFields logic so that it can stay fast.

But I can't deny the fact that it keeps things simple.

packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -282,6 +274,18 @@ export const ensureBookingInputsHaveSystemFields = ({
...
+  if (smsNumberSources.length && !bookingFields.find((f) => f.name !== SMS_REMINDER_NUMBER_FIELD)) {
+    const indexForLocation = bookingFields.findIndex((f) => f.name === "location");
+    // Add the SMS Reminder Number field after `location` field always
```
hariombalhara, line 282: It fixes the positioning of smsReminderField

packages/trpc/server/routers/viewer/workflows.tsx
CarinaWolli: Good point, I made some changes to simplify it. @hariombalhara please take another look 🙏

hariombalhara (approved):

PR #7574, opened 2023-03-08
Generate getSchedule response using the invitee timeZone

roae:
## What does this PR do?

fixed a bug when the organizer is in utc- and the guest is in utc+, the response of the getSchedule endpoint was not being 
formed correctly, when grouping by day the slots were being put in the wrong day/group. 

**This fix should correct many issues related to the differences with the connected calendars.**

Example of the current response:

imagen

## Type of change

## How should this be tested?

Create an event with utc-1 time zone e.g. America/Los_Angeles and on the booking page select an utc+ time zone e.g. Europe/Amsterdam. check that the times are displayed correctly.


zomars (approved):
Updated to reduce git diffs. Nice find! @roae 🙏🏽

Issue #7255, opened 2023-02-21
[CAL-1127] [CAL-192] "&" symbol and apostrophes don't get rendered properly in the confirmation email title


milospuac:
This is the name of the event type

![Screenshot 2022-10-19 at 17.27.54.png](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/1927ea3f-e004-4c5c-bf9d-e8b396d5445b/0bd61915-b4c1-4c5c-9bfa-2fe03b1aa978)

And this is how it appears in the email

![Screenshot 2022-10-19 at 17.29.45.png](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/b824eac5-1ca0-4692-8845-fa0603f443ef/c625a4ba-4e5e-4000-8480-31c0d621dc12)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-192](https://linear.app/calcom/issue/CAL-192/and-symbol-and-apostrophes-dont-get-rendered-properly-in-the)</sub>

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-192](https://linear.app/calcom/issue/CAL-192/and-symbol-and-apostrophes-dont-get-rendered-properly-in-the)</sub>

milospuac:
![Screenshot 2022-11-03 at 1.33.47 PM.png](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/9517a12c-a5e3-4e4b-84ff-458c2538290b/007bf7a0-3340-4813-9277-3de12118cb64)

okay so this bug is only on production ( app.cal.com) and not in development. will have to investigate more

milospuac:
Can you try chrome or safari?

milospuac:
brave browser

milospuac:
@Udit-takkar do you use chrome?

milospuac:
![perplexed-mystery.gif](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/3d36ec81-5709-4ee4-b0fc-0e1b13f75585/9dd0f916-9146-4115-9991-87127944c090)

milospuac:
[...]

PR #7591, opened 2023-03-08
Fix save booking title unescaped

G3root:
## What does this PR do?
the encoding of special characters in booking title bug which was closed by  #7490 still appears to be persist. this PR aims to fix that

## Type of change


packages/core/event.ts
```diff
@@ -17,6 +17,9 @@ export function getEventName(eventNameObj: EventNameObjectType, forAttendeeView
       eventName: eventNameObj.eventType,
       host: eventNameObj.host,
       attendeeName: eventNameObj.attendeeName,
+      interpolation: {
```
G3root, line 20: as per https://github.com/calcom/cal.com/issues/7255#issuecomment-1442027857 . shout out to @CarinaWolli

packages/emails/templates/_base-email.ts
```diff
@@ -40,7 +41,7 @@ export default class BaseEmail {
     const parseSubject = z.string().safeParse(payload?.subject);
     const payloadWithUnEscapedSubject = {
       ...payload,
-      ...(parseSubject.success && { subject: decodeURIComponent(parseSubject.data) }),
+      ...(parseSubject.success && { subject: decodeHTML(parseSubject.data) }),
```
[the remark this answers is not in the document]
keithwillcode: Are we sure we can remove using decodeURIComponent completely? Using the playground you linked here, if I add the text `http%3A%2F%2Fwww.testing.com%2F` to the string to be decoded, decodeHTML doesn't work whereas decodeURIComponent does.

Perhaps the solution is to use both? (Note that I'm currently unfamiliar with the expected data to be passed to this function but simply an observation I made while reviewing.)

Screen Shot 2023-03-12 at 8 50 08 PM

keithwillcode (commented):
See comment re: changing out function completely to use `decodeHTML`

G3root:
@keithwillcode pushed some changes

keithwillcode (approved):

G3root:
@keithwillcode some new changes. removed `decodeURIComponent` again because it throws a `malformed URI sequence`  exception if part of the string contains some `surrogate pairs` .  eg: `\uD800`   source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI.

in the future we won't be affected by this problem after this change https://github.com/calcom/cal.com/issues/7255#issuecomment-1442027857 and the subject is decoded as a fallback for booking titles, which has been already saved with encoded titles . i think giving up on decoding url is to good to go for now , whats your thought ?

apps/web/components/booking/BookingListItem.tsx
```diff
@@ -224,7 +224,7 @@ function BookingListItem(booking: BookingItemProps) {
   const showRecordingsButtons =
     (booking.location === "integrations:daily" || booking?.location?.trim() === "") && isPast && isConfirmed;
 
-  const title = decodeURIComponent(booking.title);
+  const title = booking.title;
```
G3root, line 227: introduced in https://github.com/calcom/cal.com/pull/7490 to decode encoded booking title, throw this error if the title has malformed URI. so removed

Screenshot 2023-03-13 at 21-16-42 Buchungen Cal com

keithwillcode:
> @keithwillcode some new changes. removed `decodeURIComponent` again because it throws a `malformed URI sequence` exception if part of the string contains some `surrogate pairs` . eg: `\uD800` source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI.
> 
> in the future we won't be affected by this problem after this change [#7255 (comment)](https://github.com/calcom/cal.com/issues/7255#issuecomment-1442027857) and the subject is decoded as a fallback for booking titles, which has been already saved with encoded titles . i think giving up on decoding url is to good to go for now , whats your thought ?

Cool as long as we aren't losing some existing functionality people rely on today, sounds good to me.

apps/web/package.json
```diff
@@ -69,6 +69,7 @@
     "bcryptjs": "^2.4.3",
     "classnames": "^2.3.1",
     "dotenv-cli": "^6.0.0",
+    "entities": "^4.4.0",
```
zomars, line 72: @G3root @keithwillcode This new dependency seems like it's not being used. Let's revert this change please 🙏🏽

PR #7597, opened 2023-03-08
Fixing multiple duration not selected correctly

leog:
## What does this PR do?

Fixes the autoselection of a multiple duration on booking page when a valid duration comes from the query.

Fixes: issue reported in feedback

## Type of change

## How should this be tested?

Go to a multiple duration event type, select a duration and reload the page, it should stay in the selected multiple duration.


emrysal (approved):
Should probably extract this into it's own hook, good for now 🙏

PR #7609, opened 2023-03-08
new error for no available user

### LINKED ISSUE #7608: [CAL-1205] better issue for "no available user"
image

instead say:

"This time is not available anymore. Can you go back and chose a new time?"

<sub>[CAL-1205](https://linear.app/calcom/issue/CAL-1205/better-issue-for-no-available-user)</sub>

PeerRich:
a better error message until we finally fixed the bug. added "go back" button.

needs testing, cant reproduce the issue locally


apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -615,17 +615,28 @@ export default BookingPage;
...
             {rescheduleUid ? t("reschedule_fail") : t("booking_fail")}{" "}
-            {error instanceof HttpError || error instanceof Error ? t(error.message) : "Unknown error"}
+            {error instanceof HttpError || error instanceof Error ? (
+              <>
+                {t("can_you_try_again")}
+                <span className="underline" onClick={() => router.back()}>
+                  {t("go_back")}
+                </span>
+                .
+              </> /* t(error.message) */
```
emrysal, line 636: "No available users found" is not the only possible error message, we definitely need a way to differentiate the error thrown (Like error code?)

PeerRich: the question is, do we want to show the error there. i dont think so. we need to log the error yes, but ideally we should show a generic error that is helpful (go back) vs a very technical error

emrysal: We'll handle it on the backend.

emrysal (approved):
Tested, made some amends, works.

apps/web/components/booking/pages/BookingPage.tsx
zomars: Can we use error codes? Like E1001 = "The user was not available at this slot" and so on...

PR #7618, opened 2023-03-09
Added serverside translations to the forgot password page.

JeroenReumkens:
## What does this PR do?

Adds server side translations to the forgot password page. Looks better in google and generates proper titles in OG image 😁

## Type of change

## How should this be tested?

- [ ] Visit the forgot password page and check that you don't see a flash in the page title where the translation key is shown
- [ ] Check the page's OG image and ensure that it now has the right text instead of translation keys.

### DIFF
--- a/apps/web/pages/auth/forgot-password/[id].tsx
+++ b/apps/web/pages/auth/forgot-password/[id].tsx
@@ -186,7 +186,7 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
         },
         id,
         csrfToken: await getCsrfToken({ req: context.req }),
-        ...(await serverSideTranslations(context.locale || "en", ["common", "home"])),
+        ...(await serverSideTranslations(context.locale || "en", ["common"])),
       },
     };
   } catch (reason) {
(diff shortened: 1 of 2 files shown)


apps/web/pages/auth/forgot-password/[id].tsx
```diff
@@ -185,6 +186,7 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
         },
         id,
         csrfToken: await getCsrfToken({ req: context.req }),
+        ...(await serverSideTranslations(context.locale || "en", ["common", "home"])),
```
Comment (hariombalhara) [inline apps/web/pages/auth/forgot-password/[id].tsx:189]: What is the purpose of home here?
I don't see home.json files.

Also, I have mostly seen and used this approach for Server side translations. Not sure which is the correct or best appraoch
Screenshot 2023-03-09 at 2 05 17 PM

Comment (JeroenReumkens) [reply apps/web/pages/auth/forgot-password/[id].tsx:189]: Damn, you're right. Home shouldn't be included, I copy pasted 🥲 You got me there. 
As for the the ssr hydrate — yes that's the most common approach. However, the forgot password page is a static page, meaning it doesn't have access to other server stuff (like prisma), which is why the `ssrInit()` can't be used. 

Gonna push a fix for the translation now.

PeerRich (approved):

PR #7626, opened 2023-03-09
Fixes/Embed Snippet Generator

hariombalhara:
Fixes #7625

### DIFF
--- a/apps/web/components/Embed.tsx
+++ b/apps/web/components/Embed.tsx
@@ -669,8 +669,11 @@ const EmbedTypeCodeAndPreviewDialogContent = ({
   embedUrl: string;
 }) => {
   const { t } = useLocale();
+
   const router = useRouter();
   const iframeRef = useRef<HTMLIFrameElement>(null);
+  const dialogContentRef = useRef<HTMLDivElement>(null);
+
   const s = (href: string) => {
     const searchParams = new URLSearchParams(router.asPath.split("?")[1] || "");
     const [a, b] = href.split("=");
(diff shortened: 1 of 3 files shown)


packages/lib/hooks/useTheme.tsx
```diff
@@ -6,12 +6,10 @@ import type { Maybe } from "@calcom/trpc/server";
 
 // makes sure the ui doesn't flash
 export default function useTheme(theme?: Maybe<string>) {
-  let currentTheme: Maybe<string> = theme || "system";
-
   const { resolvedTheme, setTheme, forcedTheme, theme: activeTheme } = useNextTheme();
   const embedTheme = useEmbedTheme();
   // Embed UI configuration takes more precedence over App Configuration
-  currentTheme = embedTheme || theme;
```
hariombalhara, line 14: It was overriding the previously set variable and thus default of "system" was never getting used.

packages/ui/components/form/color-picker/colorpicker.tsx
```diff
@@ -27,8 +29,8 @@ const ColorPicker = (props: ColorPickerProps) => {
             />
           </Popover.Trigger>
         </div>
-        <Popover.Portal>
-          <Popover.Content align="center" sideOffset={10}>
+        <Popover.Portal container={props.container}>
```
hariombalhara, line 32: Container on portal allows Popover to be descendant of the container. 
We need to make popover descendant of the Dialog Content so that it can be focussed.
Here is a related issue https://github.com/radix-ui/primitives/issues/922

apps/web/components/Embed.tsx
```diff
@@ -979,6 +984,8 @@ const EmbedTypeCodeAndPreviewDialogContent = ({
                     <div>Text Color</div>
                     <div className="w-full">
                       <ColorPicker
+                        popoverAlign="start"
```
hariombalhara, line 987: It was needed because when it's shown in Dialog, a certain portion from left was cut-off

zomars (approved):

PR #7641, opened 2023-03-09
Fix jackson build error

keithwillcode:
## What does this PR do?

Fixes build error when prettier runs against the jackson.js file. There was a conflict of variable names between local and global scope.

## Type of change

## How should this be tested?

- [x] Ensure SSO functionality has not been affected


emrysal (approved):
It looks good to me - IMO we don't even need to alias `g` to `global` (and pollute the global NS with it) - so that's one further tweak I'd suggest.

keithwillcode:
> It looks good to me - IMO we don't even need to alias `g` to `global` (and pollute the global NS with it) - so that's one further tweak I'd suggest.

Yeah I was thinking something similar. You mean by using `globalThis` or were you thinking of something else?

emrysal:
@keithwillcode Global this will add an additional layer of safety - good thought

leog (approved):
Thank you for this, it was annoying!

PR #7644, opened 2023-03-09
Show proper default destination calendar

joeauyeung:
## What does this PR do?

This PR fixes a bug where the incorrect calendar was being displayed as the default destination calendar. 

If we have Outlook set as the destination calendar
**Before**
image

**After**
image

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


packages/core/CalendarManager.ts
```diff
@@ -80,6 +80,7 @@ export const getConnectedCalendars = async (
         if (destinationCalendar) {
           destinationCalendar.primaryEmail = primary.email;
           destinationCalendar.integrationTitle = integration.title;
+          destinationCalendar = Object.freeze(destinationCalendar);
```
joeauyeung, line 83: This makes the destination calendar object immutable and if we do try changing the object it fails silently

zomars: Do we know where the mutation is happening? How could we know?

joeauyeung: I can't comment on the unchanged lines of code. This is a part of `getConnectedCalendars()` that loops through all calendar credentials and all calendars inside of that credential. The destination calendar object is declared outside of this loop. 

The bug happened when we had more than one connected account. For example, if we have three credentials and the destination calendar is in the 2nd credential. The if statement on line 80 would trigger again and change the destination calendar object.

keithwillcode: I feel like we should leave a `TODO` (to go back at some point and find the root cause) or `HACK` comment here if we go with this solution. It's not immediately obvious to another reader of this code why the object would be frozen here nor is silently failing an ideal solution.

joeauyeung: @keithwillcode updated with a link to my above comment. I also added a check to see if `destinationCalendar` is frozen.

emrysal (approved):

PR #7647, opened 2023-03-09
Delay workflow email cancellation in sendgrid for all case

CarinaWolli:
## What does this PR do?

Sendgrid's limit for entries in` /scheduled_sends `for cancelled scheduled emails got reached again. After making some first improvements in #7232, we need to make sure to keep the number of cancelled `scheduled_sends` lower (as low as possible).

With this PR `WorkflowStep` and `Booking` will now be optional fields in `WorkflowReminder`, so we can get rid of the immediate delete that was done when workflow or workflow step was deleted and also when an event was rescheduled.

### DIFF
--- a/packages/features/ee/workflows/api/scheduleSMSReminders.ts
+++ b/packages/features/ee/workflows/api/scheduleSMSReminders.ts
@@ -9,7 +9,8 @@ import { bookingMetadataSchema } from "@calcom/prisma/zod-utils";
 
 import { getSenderId } from "../lib/alphanumericSenderIdSupport";
 import * as twilio from "../lib/reminders/smsProviders/twilioProvider";
-import customTemplate, { VariablesType } from "../lib/reminders/templates/customTemplate";
+import type { VariablesType } from "../lib/reminders/templates/customTemplate";
+import customTemplate from "../lib/reminders/templates/customTemplate";
 import smsReminderTemplate from "../lib/reminders/templates/smsReminderTemplate";
 
 async function handler(req: NextApiRequest, res: NextApiResponse) {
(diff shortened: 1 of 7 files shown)


packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -53,83 +54,85 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
   if (!unscheduledReminders.length) res.json({ ok: true });
 
   for (const reminder of unscheduledReminders) {
-    try {
-      const sendTo =
-        reminder.workflowStep.action === WorkflowActions.SMS_NUMBER
-          ? reminder.workflowStep.sendTo
-          : reminder.booking?.smsReminderNumber;
+    if (reminder.workflowStep && reminder.booking) {
```
emrysal, line 57: Invert and continue; - you're going to deep 🙏

CarinaWolli: Fixed, and in scheduleEmailReminder as well

packages/prisma/migrations/20230309203435_make_booking_and_workflow_step_optional_for_workflow_reminder/migration.sql
```diff
@@ -0,0 +1,15 @@
+-- DropForeignKey
+ALTER TABLE "WorkflowReminder" DROP CONSTRAINT "WorkflowReminder_bookingUid_fkey";
+
+-- DropForeignKey
+ALTER TABLE "WorkflowReminder" DROP CONSTRAINT "WorkflowReminder_workflowStepId_fkey";
+
+-- AlterTable
+ALTER TABLE "WorkflowReminder" ALTER COLUMN "bookingUid" DROP NOT NULL,
+ALTER COLUMN "workflowStepId" DROP NOT NULL;
```
emrysal, line 9: Checked there are no NULL variables on production for `bookingUid` and `workflowStepId`

emrysal (approved):
LGTM 👍

PR #7652, opened 2023-03-10
Make sure that a hidden field/question is never required

### LINKED ISSUE #7653: [CAL-1218] Can't book an event with hidden but required field.
Can't go through with a booking which has an hidden field which is marked required.

Thanks @Udit-takkar for reporting the issue.

<sub>[CAL-1218](https://linear.app/calcom/issue/CAL-1218/cant-book-an-event-with-hidden-but-required-field)</sub>

hariombalhara:
## What does this PR do?

Fixes #7653

**Environment**:  Production

## Type of change

## How should this be tested?

- [x] Create a field. By default, it would be required. Now toggle it to hidden. Note that you would just see **Hidden** badge now. If you would go and edit the field you would still see that it's marked required, but it won't actually be required during booking. Go and make a booking without prefilling the hidden field value.
## Checklist

### DIFF
--- a/packages/features/form-builder/FormBuilder.tsx
+++ b/packages/features/form-builder/FormBuilder.tsx
@@ -307,9 +307,9 @@ export const FormBuilder = function FormBuilder({
                     <div className="flex items-center space-x-2">
                       {field.hidden ? (
                         // Hidden field can't be required, so we don't need to show the Optional badge
-                        <Badge variant="gray">Hidden</Badge>
+                        <Badge variant="gray">{t("hidden")}</Badge>
                       ) : (
-                        <Badge variant="gray">{isRequired ? "Required" : "Optional"}</Badge>
+                        <Badge variant="gray">{isRequired ? t("required") : t("optional")}</Badge>
                       )}
                       {Object.entries(groupedBySourceLabel).map(([sourceLabel, sources], key) => (
                         // We don't know how to pluralize `sourceLabel` because it can be anything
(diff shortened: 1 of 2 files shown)


packages/features/form-builder/FormBuilder.tsx
```diff
@@ -266,8 +266,7 @@ export const FormBuilder = function FormBuilder({
           {fields.map((field, index) => {
             const fieldType = FieldTypesMap[field.type];
 
-            // Hidden fields can't be required
-            const isRequired = field.required && !field.hidden;
+            const isRequired = field.required;
```
hariombalhara, line 269: No need to show the field as optional now. As mentioned by @Jaibles we would not be showing 'Optional' badge if field is hidden. There would just be Hidden badge.

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -306,8 +305,12 @@ export const FormBuilder = function FormBuilder({
                       {field.label || t(field.defaultLabel || "")}
                     </div>
                     <div className="flex items-center space-x-2">
-                      <Badge variant="gray">{isRequired ? "Required" : "Optional"}</Badge>
-                      {field.hidden ? <Badge variant="gray">Hidden</Badge> : null}
+                      {field.hidden ? (
+                        // Hidden field can't be required, so we don't need to show the Optional badge
+                        <Badge variant="gray">Hidden</Badge>
```
CarinaWolli, line 310: Can you add translation for hidden and required?

CarinaWolli (approved):
LGTM!

packages/features/form-builder/FormBuilder.tsx
hariombalhara: Thanks for pointing out @CarinaWolli. Fixed these and Label and Placeholder as well.

emrysal (approved):

PR #7206, opened 2023-02-18
empty screen for unpublished teams

PeerRich:
before: 

we throw a generic 404 because right now we turn the slug to "null" when unpublished

CleanShot 2023-02-19 at 14 52 20@2x

after:

CleanShot 2023-02-19 at 14 49 59@2x

- [x] i18n
- [x] show team name in headline (replace ACME)
- [x] ~~link button: mailto:teamowner@company.com~~ 
- [x] check for something like team.published from endpoint

PR #7662, opened 2023-03-10
Fixed linting issues in Playwright config

keithwillcode:
## What does this PR do?

Fixes linting issues in the Playwright config where we were using the `Function` type declaration, which is not ideal. This was causing downstream issues in #7206.

## Type of change

## How should this be tested?

- [ ] Ensure the playwright tests pass

### DIFF
--- a/packages/embeds/embed-core/playwright/config/playwright.config.ts
+++ b/packages/embeds/embed-core/playwright/config/playwright.config.ts
@@ -2,7 +2,7 @@ import type { PlaywrightTestConfig, Frame } from "@playwright/test";
 import { devices, expect } from "@playwright/test";
 import * as path from "path";
 
-// eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


packages/embeds/embed-core/playwright/config/playwright.config.ts
```diff
@@ -79,7 +80,8 @@ declare global {
     interface Matchers<R> {
       toBeEmbedCalLink(
         calNamespace: string,
-        getActionFiredDetails: Function,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
keithwillcode, line 83: Don't love the idea of adding more `eslint-disable` rules but sometimes in these very generic cases (like in test suites) it's necessary. I see we do disable this in some other places where the code is generic. Open to suggestions here.

The change from not using `Function` was due to an eslint error.

hariombalhara (approved):

PR #7678, opened 2023-03-12
Fixes weird background color on selects

sean-brydon:
I have no clue why we did this. All it seems to do is break stuff

Before

CleanShot 2023-03-12 at 23 30 46@2x

After:
CleanShot 2023-03-12 at 23 30 30@2x


emrysal (approved):

PeerRich:
nice that it was merged but even that hover is incorrect, see figma 

dont we wanna use the new dropdown component?

CleanShot 2023-03-12 at 16 58 32@2x

PR #7698, opened 2023-03-13
fix: use zod in getting started and settings

Udit-takkar:
## What does this PR do?

Fixes 
 - add zod validation in settings page and onboarding

## Type of change


emrysal:
Why are we adding a maximum length for the username? 🤔

hariombalhara (changes requested):
This might fail the schema for existing usernames which are crossing the max length limit. This can end up crashing the profile page and maybe other pages for such users.

Such usernames can be in our hosted cal.com or for self-hosted instances.

Maybe we should enforce this limit only on Hosted cal.com(after verifying that all usernames are under that limit) and we can give an opt-in for self hosted users.

EDIT: It came to my mind that the solution it can be as simple as keeping the limit very high by default and reduce it only for cal.com.

Udit-takkar:
@hariombalhara sorry changed the name of the variable It was for Full name not username. The limit is required because of this https://linear.app/calcom/issue/CAL-1190/very-long-names-issue and No this won't crash the page for user who have set their full name very long earlier because the limit  is on the zod schema (UI and API) and not on postgres DB.

 Now users won't be able to update to a new full name greater than the limit or if they want to update profile in settings/my-account/profile they would also see the message to keep their username less than 50 chars.

hariombalhara:
@Udit-takkar Makes sense then 👍

hariombalhara (approved):

PR #7699, opened 2023-03-13
Fix dynamic bookings overwriting users

### LINKED ISSUE #7648: Success page of a group booking is not correctly displaying attendees

I made a group booking and it only showed my information and the host's information was duplicated.

34454665462_image2048

joeauyeung:
## What does this PR do?

This PR moves some logic in `handleNewBooking` that caused some users to be overwritten.

Fixes #7648

## Type of change

## How should this be tested?

- Create a dynamic booking with "free+pro"
- Both users should be saved in the booking

## Checklist

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -700,10 +700,8 @@ async function handler(
       : [];
 
   const teamMembers = await Promise.all(teamMemberPromises);
-  console.log("🚀 ~ file: handleNewBooking.ts:706 ~ teamMembers:", teamMembers);
 
   const attendeesList = [...invitee, ...guests];
-  console.log("🚀 ~ file: handleNewBooking.ts:709 ~ attendeesList:", attendeesList);
 
   const responses = "responses" in reqBody ? reqBody.responses : null;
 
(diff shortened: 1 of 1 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -668,20 +682,6 @@ async function handler(
...
   const uid = translator.fromUUID(uuidv5(seed, uuidv5.URL));
 
-  let locationBodyString = location;
-  let defaultLocationUrl = undefined;
-  if (dynamicUserList.length > 1) {
-    users = users.sort((a, b) => {
-      const aIndex = (a.username && dynamicUserList.indexOf(a.username)) || 0;
-      const bIndex = (b.username && dynamicUserList.indexOf(b.username)) || 0;
-      return aIndex - bIndex;
-    });
```
joeauyeung, line 678: On line 639, we assigned the first entry in the `users` array as the organizer. Here we rearrange the `users` array so there's no guarantee that the first entry in the array will stay the same. I moved these lines of code earlier in the function so when we assign the organizer, the `users` array will not change anymore.

zomars (approved):

PR #7700, opened 2023-03-13
feat: add rate limiting to reset password endpoint

G3root:
## What does this PR do?

Fixes CAL-1189

## Type of change


apps/web/pages/api/auth/forgot-password.ts
```diff
@@ -17,7 +34,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   try {
     const maybeUser = await prisma.user.findUnique({
       where: {
-        email: req.body?.email?.toLowerCase(),
+        email: email,
```
zomars, line 37:
```suggestion
        email,
```

apps/web/pages/api/auth/forgot-password.ts
```diff
@@ -1,14 +1,31 @@
...
+import { z } from "zod";
 
 import dayjs from "@calcom/dayjs";
 import { sendPasswordResetEmail } from "@calcom/emails";
 import { PASSWORD_RESET_EXPIRY_HOURS } from "@calcom/emails/templates/forgot-password-email";
+import rateLimit from "@calcom/lib/rateLimit";
 import { getTranslation } from "@calcom/lib/server/i18n";
 import prisma from "@calcom/prisma";
 
+const limiter = rateLimit({
```
zomars, line 12: We need to come up with a better strategy. As discussed with @roae `memory-cache` doesn't play nice with server less environments.

emrysal: Doesn't have to be perfect though for rate limiting, whether this needs fixing needs to be shown by real world tests, perhaps it isn't exactly 10 times but if it ratelimits after 50 in different serverless runners that's still achieving the necessary limiting.

zomars (approved):

apps/web/pages/api/auth/forgot-password.ts
```diff
@@ -1,14 +1,31 @@
...
+
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
   const t = await getTranslation(req.body.language ?? "en", "common");
+  const email = z
+    .string()
+    .email()
+    .transform((val) => val.toLowerCase())
+    .parse(req.body?.email);
+
+  const { isRateLimited } = limiter.check(10, email); // 10 requests per minute
```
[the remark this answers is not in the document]
G3root: i'll open a PR for that

PR #7706, opened 2023-03-13
Adds prisma data proxy

### LINKED ISSUE #7556: [CAL-1191] speed up public booking page (prisma data proxies?)
look into prisma data proxies

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1191](https://linear.app/calcom/issue/CAL-1191/speed-up-public-booking-page-prisma-data-proxies)</sub>

zomars:
## What does this PR do?

Closes #7556

Adds missing config to enable Prisma Data Proxy. It should be opt-in by default.

~~It introduces a breaking change as it requires everyone to set the new `DIRECT_DATABASE_URL` env variable as there's no easy way to add this programmatically and conditionally. User who don't want to use Prisma Data Proxy can just set `DIRECT_DATABASE_URL` the same as `DATABASE_URL` and leave `PRISMA_GENERATE_DATAPROXY` as false.~~ 

Removes migrations from build/start scripts. We will automate that in a follow up for our use case. We also need to update the docs to make this clear. Each self-hosted user should be responsible on how and when to run migrations (cc @alishaz-polymath) We should offer a default way to doing so. Either via a script or a GitHub action, etc.

We don't need to change imports yet. It is only the case if we're planning to use prisma in the edge runtime (middleware/edge endpoints,etc.) [Quoting the docs](https://www.prisma.io/docs/data-platform/data-proxy/use-data-proxy#steps-1):

> To use the Data Proxy with Node.js, instantiate `@prisma/client`
> To use the Data Proxy in Edge runtimes, such as Cloudflare Workers or Vercel Edge Functions, instantiate `@prisma/client/edge`.

## Type of change

## How should this be tested?

- [ ] Set `PRISMA_GENERATE_DATAPROXY=true`
- [ ] Set `DATABASE_URL` to the prisma connection string: `prisma://...` 
- [ ] yarn
- [ ] yarn build
- [ ] yarn start


PeerRich:
can we not check if env is null and then use database_url? @keithwillcode

zomars:
> can we not check if env is null and then use database_url? @keithwillcode 

We could create a script that modifies the prisma schema before build (and run). AFAIK we cannot put conditions inside the prisma schema.

zomars:
That or we can leave DATABASE_URL As it is and remove all prisma migrate deploy calls from our build/start scripts. @emrysal and myself were discussing this before.

Cons:
- No more auto DB migrate from deploys
- Figure out when and how to run migrations

Pros:
- Way less conflicts on previews
- More streamlined deploys, possibly faster

PeerRich:
ideally we reduce breaking changes when it comes to .env's

emrysal:
I agree @zomars - we can always automate in a different way (I still think we can get prisma migrate deploy to run automatically)

zomars:
Redrafting then. Will do:
- Remove the need for `DIRECT_DATABASE_URL`
- Remove auto migrations from building process
- Update GitHub Actions accordingly

PeerRich:
nice! very excited about this

.env.example
```diff
@@ -22,6 +22,8 @@ CALCOM_LICENSE_KEY=
 
 # - DATABASE ************************************************************************************************
 DATABASE_URL="postgresql://postgres:@localhost:5450/calendso"
+# So we can use Prisma Data Proxy (https://www.prisma.io/docs/data-platform/data-proxy/use-data-proxy)
```
emrysal, line 25: Perhaps a better description (what can a self hoster use this for if they want to?, can they, should they?)

zomars: The link should be enough? Or do you mean specific benefits like cold-boots?

zomars: Added a better explainer

zomars:
Ready for re-review

roae:
@zomars when will migrations be run?

zomars:
> @zomars when will migrations be run?

Not in build that's for sure. We will follow up to decide in another PR.

emrysal (approved):
Let's go 🚀

zomars:
> fyi this broke signup and api

Also adding to this. It was not the code changes that broke it. It was the lack of simultaneous deployment on submodules that created a mismatch between webapp, website and api. Just to clarify and add to our learnings.

Issue #7705, opened 2023-03-13
[CAL-1243] Public booking form: Remove placeholders


Jaibles:
![A7DCF1AA-20AB-493D-8D83-C1D96872A880.jpeg](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/09c6c22d-f9c7-4c66-9530-729f5acf6aae/093b17a3-d7f0-4d58-ac03-8d4173d27d6c)

We should just remove these. People know what the expected values are here

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1243](https://linear.app/calcom/issue/CAL-1243/public-booking-form-remove-placeholders)</sub>

HiMoritz:
I was hoping to work on this as my first open-source contribution, but I missed that this issue was already assigned. I have a solution working locally, should I still proceed and open a PR for review?

PeerRich:
@HiMoritz yes please

PR #7708, opened 2023-03-13
Removed email & name placeholder from booking page 

HiMoritz:
## What does this PR do?
This change removes the placeholder values for the **name** and **email** input fields on the booking page. 

refs #7705

Booking Page Screenshot

## Type of change
- [ x] Bug fix (non-breaking change which fixes an issue) / UI improvement 

## How should this be tested?
I tested the changes manually and visually on a local development instance

- [ ] Check the booking page for any type of meeting


G3root (commented):
Awesome Work ❤️  . LGTM :)

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -162,7 +162,10 @@ const BookingFields = ({
         }
 
         const label = noLabel ? "" : field.label || t(field.defaultLabel || "");
-        const placeholder = field.placeholder || t(field.defaultPlaceholder || "");
+        const placeholder =
```
G3root, line 165: NIT: prefered `includes` over `indexOf`, it should be consistent throughout the codebase whenever comparing values

https://handbook.cal.com/engineering/best-practices/avoid-comparing-multiple-values-with-includes

G3root (commented):
Awesome Work ❤️  . LGTM :)

apps/web/components/booking/pages/BookingPage.tsx
hariombalhara: This isn't the right place to do it. You should remove the defaultPlaceholder here https://github.com/calcom/cal.com/blob/7271bae91ade4055829e31cf7e1e5f1d6c4a777f/packages/features/bookings/lib/getBookingFields.ts#L166

hariombalhara (changes requested):

PeerRich (dismissed):
i know @Jaibles  requested this but giving it a closer look i think we should keep placeholders for a11y

Jaibles:
> i know @Jaibles  requested this but giving it a closer look i think we should keep placeholders for a11y

I don't see any evidence for that. 
"placeholder text is not broadly supported across assistive technologies"
- https://www.w3.org/WAI/tutorials/forms/instructions/
- https://www.a11yproject.com/posts/placeholder-input-elements/

PeerRich:
you are right

however lets only remove it here and not on the routing forms inputs

@hariombalhara's comment is still required to be addressed

PeerRich:
or should we remove placeholders app wide?

Jaibles:
> or should we remove placeholders app wide?

I think we can just remove them on the booking page for now and asses other places individually. From what I understand, you should not use them for anything that needs to be read and therefore not for accessibility. I'm not 100% sure that they can't still be used in certain places.

apps/web/components/booking/pages/BookingPage.tsx
HiMoritz: @hariombalhara If I read that correctly, these `systemBeforeFields`are used in other instances beyond the book page. Should the placeholders then be removed in all of those instances or specifically on the book page?

hariombalhara: Yeah, that is the single source of Truth for Name and Email fields. This would also allow the placeholder to go away from Manage Booking Questions - UI as well.
Screenshot 2023-03-14 at 7 26 13 PM

hariombalhara (approved):

PR #7720, opened 2023-03-14
fix: use title instead of eventtype

### LINKED ISSUE #7046: [CAL-1247] Empty variables: {{eventType}} + {{name}}  in Email Subject
I noticed when a event-type is on confirmation only mode, that the email to the attendee has empty variables in the header. 

`Booking created: {{eventType}} with {{name}} at... Email Subject has empty variables in the Header`

<sub>[CAL-1247](https://linear.app/calcom/issue/CAL-1247/empty-variables-eventtype-name-in-email-subject)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/7046

## Type of change

### DIFF
--- a/apps/web/public/static/locales/fr/common.json
+++ b/apps/web/public/static/locales/fr/common.json
@@ -68,7 +68,7 @@
   "password_reset_instructions": "Si vous n'en avez pas fait la demande, vous pouvez ignorer cet e-mail en toute sécurité et votre mot de passe ne sera pas modifié.",
   "event_awaiting_approval_subject": "En attente de validation : {{title}} à {{date}}",
   "event_still_awaiting_approval": "Un événement est toujours en attente de votre validation",
-  "booking_submitted_subject": "Réservation confirmée : {{eventType}} avec {{name}} le {{date}}",
+  "booking_submitted_subject": "Réservation confirmée :  {{title}} le {{date}}",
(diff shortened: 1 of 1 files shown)


apps/web/public/static/locales/fr/common.json
```diff
@@ -68,7 +68,7 @@
   "password_reset_instructions": "Si vous n'en avez pas fait la demande, vous pouvez ignorer cet e-mail en toute sécurité et votre mot de passe ne sera pas modifié.",
   "event_awaiting_approval_subject": "En attente de validation : {{title}} à {{date}}",
   "event_still_awaiting_approval": "Un événement est toujours en attente de votre validation",
-  "booking_submitted_subject": "Réservation confirmée : {{eventType}} avec {{name}} le {{date}}",
```
hariombalhara, line 71: I think we should be able to write a utility that can ensure that a string has the same variables in all languages. I have seen this kind of language-specific issue many times. It's difficult to identify during testing and only the users who use the language end up facing the issue.

hariombalhara (approved):

PeerRich:
unfortunately I think we need to fix this in cal.crowdin.com, otherwise it'll be overwritten in the next sync

PeerRich:
merging now, lets monitor if this is the case

PR #7732, opened 2023-03-14
Fix: hydration error /booking/[uid]

G3root:
## What does this PR do?

changes done 

- [x] fixed hydration error
- [x] remove console logs
- [x] use callback ref for scroll into view rather than the `window.scrollTo `

before:
Screenshot 2023-03-15 01:23:28

## Type of change


zomars (approved):
Noice

PR #7736, opened 2023-03-14
Adds new responses to zapier payload

zomars:
## What does this PR do?

- Adds new responses to Zapier webhooks payload
- https://twitter.com/kevinyun/status/1635659185647128576?s=61&t=Tixb9sNDRIxDEA2AkQw9ag

## Type of change

## How should this be tested?

- [ ] Check payload from Zapier
- [ ] It should include both customInputs and responses

### DIFF
--- a/packages/features/webhooks/lib/sendPayload.ts
+++ b/packages/features/webhooks/lib/sendPayload.ts
@@ -1,4 +1,4 @@
-import { Webhook } from "@prisma/client";
+import type { Webhook } from "@prisma/client";
 import { createHmac } from "crypto";
 import { compile } from "handlebars";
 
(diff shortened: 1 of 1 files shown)


packages/features/webhooks/lib/sendPayload.ts
```diff
@@ -45,6 +45,7 @@ function getZapierPayload(data: CalendarEvent & EventTypeInfo & { status?: strin
     title: data.title,
     description: data.description,
     customInputs: data.customInputs,
+    responses: data.responses,
     startTime: data.startTime,
```
hariombalhara, line 49: Looks good. But I would really like to have the same payload as our webhook for zapier hooks as well. What's the reason it's different cc @CarinaWolli

hariombalhara (approved):

packages/features/webhooks/lib/sendPayload.ts
CarinaWolli: I will change that: https://github.com/calcom/cal.com/issues/7797

PR #7738, opened 2023-03-15
Upgrades yarn v1 to v3. Improves CI times by 2x

zomars:
## What does this PR do?

- [Migrates yarn classic v1 to modern v3](https://yarnpkg.com/getting-started/migration)
- Moves `eslint` to root package json
- Moves `typescript` to root package json 
- ~~Reintroduces setup step so subsequent jobs have a warm cache~~ 
- Ensures `@calcom/embed-core` is always linked locally instead of from our published package in NPM
- Prevents generated embed scripts to be linted/formatted in CI
- Splits embed-react tests to single file
- Upgrades playwright
- Removes all embed `playwright` config files but the one in root. (can you help with this @hariombalhara )

### Before

image

### After

image

## Type of change

## How should this be tested?

In theory all scripts should work as expected without any changes. Please let me know if you find any "command not found" or similar issues.

- [ ] yarn
- [ ] yarn dx
- [ ] yarn dev
- [ ] yarn build
- [ ] yarn e2e


.github/actions/cache-build/action.yml
```diff
@@ -20,6 +20,7 @@ runs:
       with:
         path: |
           ${{ github.workspace }}/apps/web/.next
+          ${{ github.workspace }}/apps/web/public/embed
```
zomars, line 23: Caches generated embed scripts as they were missing

.github/workflows/pr.yml
```diff
@@ -40,24 +40,32 @@ jobs:
     uses: ./.github/workflows/env-create-file.yml
     secrets: inherit
 
+  setup:
+    name: Yarn install & cache
+    uses: ./.github/workflows/yarn-install.yml
+    secrets: inherit
+
```
zomars, line 47: With yarn v3 this is way faster. I still need to make some test if skipping this makes significant speed improvements.

.yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
zomars, line 1: This should be committed. Allows us to use `yarn upgrade-interactive` with v3.

.yarn/releases/yarn-3.4.1.cjs
zomars, line 1: This file should be committed as well according to the docs.

apps/web/package.json
```diff
@@ -28,9 +28,9 @@
     "@calcom/app-store-cli": "*",
     "@calcom/core": "*",
     "@calcom/dayjs": "*",
-    "@calcom/embed-core": "*",
-    "@calcom/embed-react": "*",
-    "@calcom/embed-snippet": "*",
+    "@calcom/embed-core": "workspace:*",
+    "@calcom/embed-react": "workspace:*",
+    "@calcom/embed-snippet": "workspace:*",
```
zomars, line 33: This will force linking locally instead of installing from NPM

packages/app-store/exchange2016calendar/package.json
```diff
@@ -9,8 +9,8 @@
     "@calcom/lib": "*",
     "@calcom/prisma": "*",
     "@calcom/ui": "*",
-    "react-hook-form": "^7.43.3",
-    "ews-javascript-api": "^0.11.0"
+    "ews-javascript-api": "^0.11.0",
+    "react-hook-form": "^7.43.3"
```
zomars, line 13: These got autosorted

yarn.lock
zomars, line 1: Big big changes. Please test locally first.

zomars (commented):
Self review done!

zomars:
Here's the latest run: https://github.com/zomars/cal.com/actions/runs/4422261627

Looking into fixing embed tests

packages/app-store/routing-forms/playwright/tests/basic.e2e.ts
```diff
@@ -49,7 +49,7 @@ test.describe("Routing Forms", () => {
 
       await page.reload();
 
-      expect(await page.inputValue(`[data-testid="description"]`)).toMatch(description);
+      expect(await page.inputValue(`[data-testid="description"]`)).toBe(description);
```
zomars, line 52: toMatch expected a regex and we got a string.

.github/workflows/pr.yml
```diff
@@ -57,7 +61,7 @@ jobs:
 
   build:
     name: Production build
-    needs: env
+    needs: [env]
```
zomars, line 64:
```suggestion
    needs: env
```

zomars:
Embeds test will be fixed in a follow up. CI improvements are good enough to merge. cc @emrysal @hariombalhara

emrysal (approved):
Works on my local 🙏 - let's go! 🚀

PR #7750, opened 2023-03-15
Fix minimum booking notice set to 0

joeauyeung:
## What does this PR do?

This PR, solves the "Booking out of bounds" error when minimum notice was set to 0

Fixes # (issue)

## Type of change

## How should this be tested?

- In an event type, set the minimum booking notice to 0 under the "Limits" tab
- Go to the availability page. It should load

## Checklist


apps/web/components/booking/SlotPicker.tsx
```diff
@@ -133,10 +133,7 @@ export const SlotPicker = ({
     eventTypeId: eventType.id,
     eventTypeSlug: eventType.slug,
     usernameList: users,
-    startTime:
-      browsingDate === undefined || browsingDate.get("month") === dayjs.tz(undefined, timeZone).get("month")
-        ? dayjs.tz(undefined, timeZone).subtract(2, "days").startOf("day")
```
joeauyeung, line 138: This was the cause of the "bookings can't be made in the past" error. For example, if it is March 15th 3 pm, when looking for slots the start date would be March 13th 3 pm. This fails the `isOutOfBounds` check. Why was there a reason we started slots 2 days in the past?

apps/web/components/booking/SlotPicker.tsx
```diff
@@ -133,10 +133,7 @@ export const SlotPicker = ({
...
-        : browsingDate?.startOf("month"),
+    startTime: dayjs().startOf("day"),
```
joeauyeung, line 136: We can just use the current day and determine the availability from that. It wouldn't fail the `isOutofBounds` check since it compares it to `new Date()`

apps/web/components/eventtype/EventLimitsTab.tsx
```diff
@@ -88,6 +88,7 @@ const MinimumBookingNoticeInput = React.forwardRef<
           label={t("minimum_booking_notice")}
           type="number"
           placeholder="0"
+          min={0}
```
joeauyeung, line 91: For minimum bookings, they should never be below zero. Ex -2

apps/web/components/booking/SlotPicker.tsx
zomars: @roae may have more context of this

apps/web/components/eventtype/EventLimitsTab.tsx
zomars: Can we validate this on server as well?

apps/web/components/booking/SlotPicker.tsx
roae: I'm really not sure why that is done that way, I kept it because of that, I guess, it subtracts that time for utc- time zones.

joeauyeung: @zomars with @roae's comment are we good to go?

apps/web/components/eventtype/EventLimitsTab.tsx
zomars: But that's not enough if we're not actually using the schema in de endpoint. Are we?

zomars (approved):
Tested and worked 🙏🏽

PR #7756, opened 2023-03-15
Revert - create new bookings - writing team members to DB

joeauyeung:
## What does this PR do?

Fixes # (issue)

## Type of change


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1454,7 +1455,30 @@ async function handler(
...
+              //perhaps make email fetch request to see if his locale is stored, else
+              const retObj = {
+                name: attendee.name,
+                email: attendee.email,
+                timeZone: attendee.timeZone,
+                locale: attendee.language.locale,
+              };
+              return retObj;
+            }),
+            // Have this for now until we change the relationship between bookings & team members
```
emrysal, line 1470: Is this bit needed? also could you add it to the attendeesData? https://github.com/calcom/cal.com/pull/7756/files#diff-a7c21727d9a0d4d1aaff9d24fca9293bc28468861bc8022dfa03116908b6fae7R1432

emrysal (changes requested):

packages/features/bookings/lib/handleNewBooking.ts
zomars: Can we create a helper for this?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1454,7 +1455,30 @@ async function handler(
...
+            ...(evt.team?.members
+              ? evt.team.members.map((member) => {
+                  return {
+                    email: member.email,
+                    name: member.name,
+                    timeZone: member.timeZone,
+                    locale: member.language.locale,
+                  };
+                })
+              : []),
```
zomars, line 1480: Do we want ALL team members as attendees? Does this takes into account round-robin?

zomars:
@joeauyeung What is this fixing exactly?

emrysal (approved):
Push team members into attendeesData array if present.

packages/features/bookings/lib/handleNewBooking.ts
emrysal: The flow isn't nice, but it goes from users -> availableUsers -> teamMemberPromises -> teamMembers -> team.members = teamMembers.

PR #7763, opened 2023-03-16
Hotfix/account provider should be lowercase

emrysal:
## What does this PR do?

* `Account.provider` is 'google' whereas iDP was 'GOOGLE', - `account.provider` has the right lowercase value.
* Updates the identityProviderId to the account.providerAccountId instead of the userId
* Create a userland solution for fixing this retroactively.

### DIFF
--- a/apps/web/pages/auth/logout.tsx
+++ b/apps/web/pages/auth/logout.tsx
@@ -43,7 +43,7 @@ export default function Logout(props: Props) {
           </div>
         </div>
       </div>
-      <Button href="/auth/login" passHref className="flex w-full justify-center">
+      <Button href="/auth/login" className="flex w-full justify-center">
         {t("go_back_login")}
       </Button>
     </AuthContainer>
--- a/packages/features/auth/lib/next-auth-options.ts
+++ b/packages/features/auth/lib/next-auth-options.ts
@@ -444,7 +444,7 @@ export const AUTH_OPTIONS: AuthOptions = {
                 identityProvider: idP,
               },
               {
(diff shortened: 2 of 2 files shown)


emrysal:
Hi @deepakprabhakara Setting the identityProviderId to the userId seems very deliberate, but the ID we get from Google for example is very, very different. How was this intended to work and are you happy with this change?

apps/web/pages/auth/logout.tsx
```diff
@@ -43,7 +43,7 @@ export default function Logout(props: Props) {
           </div>
         </div>
       </div>
-      <Button href="/auth/login" passHref className="flex w-full justify-center">
```
emrysal, line 46: This is an unrelated fix for an error in the console.

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -537,6 +537,33 @@ export const AUTH_OPTIONS: AuthOptions = {
           },
         });
 
+        /* --- START FIX LEGACY ISSUE WHERE 'identityProviderId' was accidentally set to userId --- */
```
keithwillcode, line 540: @emrysal Should we write a migration to fix this for all users in 1 go?

keithwillcode (approved):

packages/features/auth/lib/next-auth-options.ts
emrysal: Opted against it as I'm not sure we have the identityProviderId for all Google accounts, looks like the account system is newer - we should revisit in a ~2 months and drop this code.

PR #7764, opened 2023-03-16
Prevent refresh_token nulling if the returned refresh token is empty

emrysal:
## What does this PR do?

* Fixes nulling the refresh_token if it is missing from the response.
* Adds zod for parsing the refresh token response schema.


joeauyeung (approved):
LGTM

PR #7766, opened 2023-03-16
fix: badge component line-height according to the DS

G3root:
## What does this PR do?

fixes badge components line-height to be exact to the Design System

before:

Screenshot 2023-03-16 at 10-05-44 Event Types Cal com

after:
Screenshot 2023-03-16 at 10-07-27 Event Types Cal com

## Type of change


PeerRich:
according to DS it should be text-xs

CleanShot 2023-03-16 at 10 21 39@2x

PeerRich (approved):

PR #7772, opened 2023-03-16
allow users to set a new password if they have signed up with Google or SAML provider

devkiran:
## What does this PR do?

This PR add the support for setting a new password for users who have signed up with Google or SAML provider. This allow the same user to login with email/password in the future if they want to. The users can set their new password from `/settings/security/password` page. 

We should hide the `Old Password` field for these users for the first time. If the user has already set a password, we show the `Old Password` field (PENDING)

Since the `user` object doesn't have the password field, at the moment frontend have no idea if the user has already set a password or not. Either we should create a new API or update the `user` to return a flag to indicate if the user has already set a password or not.

## Type of change

## How should this be tested?

- Sign up with Google or SAML provider
- Go to `/settings/security/password` page
- You should see the `New Password` field and `Old Password` field should be hidden
- Set a new password
- Logout and login with email/password
- It should work


devkiran:
As requested created the PR @deepakprabhakara. cc @PeerRich

PeerRich:
@devkiran can we undraft this?

devkiran:
@PeerRich Sure.

joeauyeung (commented):
The new password is written into the DB, but we should change the `identityProvider` field back to `CAL` and add some messaging that the user has to use a password to login now.

devkiran:
@joeauyeung I have updated the PR. It will now update the identityProvider field to CAL. Should we also reset the password field to null when the user login via Google?

alannnc:
@joeauyeung @devkiran Is this PR still relevant? just updated it with main latest.

PR #7784, opened 2023-03-16
fix/collective-event-handle-multiple-calendars-credentials

### LINKED ISSUE #7754: [CAL-1267] Collective Events have only one team attendee on the google event
Looks like when booking a collective event, in the google calendar event it only displays the current user.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1267](https://linear.app/calcom/issue/CAL-1267/collective-events-have-only-one-team-attendee-on-the-google-event)</sub>

alannnc:
## What does this PR do?

When collective event is created, EventManager.createCalendars just uses organiser calendar credentials for creating the event. With these changes we can use team destination calendars and create each calendars for each team member with right list of attendees.

Fixes #7754

## Type of change

## How should this be tested?

- [ ] Create a Collective Team Event
- [ ] Add Teampro and Team free as hosters
- [ ] Add to each user a different google calendar
- [ ] Book using a random user
- [ ] Finally you should be able to see calendar in google with right attendees and for each user host calendar.

## Checklist

### DIFF
--- a/packages/core/EventManager.ts
+++ b/packages/core/EventManager.ts
@@ -311,7 +311,38 @@ export default class EventManager {
 
     let createdEvents: EventResult<NewCalendarEventType>[] = [];
     if (event.destinationCalendar) {
-      if (event.destinationCalendar.credentialId) {
+      if (
+        event.schedulingType === "COLLECTIVE" &&
+        event.team &&
+        event.team.destinationCalendar.length > 0 &&
+        event.team.members.length > 0
+      ) {
+        // Loop through each member of the event's team and create events for their destination calendars
+        for (const destinationCalendar of event.team.destinationCalendar) {
+          if (destinationCalendar.credentialId) {
+            const dbCredential = await prisma.credential.findFirst({
+              where: {
+                id: destinationCalendar.credentialId,
+              },
+            });
+            if (!dbCredential) {
+              console.error("Credential not found");
+              continue;
+            }
+            const credential: CredentialWithAppName = {
+              ...dbCredential,
+              appName: "", // Can appName be more easy to get?
+            };
+            if (credential) {
+              const memberEvent = { ...event, destinationCalendar };
+              const createdEvent = await createEvent(credential, memberEvent);
+              if (createdEvent) {
+                createdEvents.push(createdEvent);
+              }
+            }
+          }
+        }
+      } else if (event.destinationCalendar.credentialId) {
         const credential = this.calendarCredentials.find(
           (c) => c.id === event.destinationCalendar?.credentialId
         );
--- a/packages/core/EventManager.ts
+++ b/packages/core/EventManager.ts
@@ -311,15 +311,9 @@ export default class EventManager {
 
     let createdEvents: EventResult<NewCalendarEventType>[] = [];
     if (event.destinationCalendar) {
-      if (
-        event.schedulingType === "COLLECTIVE" &&
-        event.team &&
-        event.team.destinationCalendar &&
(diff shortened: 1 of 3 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -557,25 +557,46 @@ async function handler(
...
+
+        return users.length ? users : eventType.users;
+      }
+    } catch (error) {
+      if (error instanceof HttpError || error instanceof Prisma.PrismaClientKnownRequestError) {
+        throw new HttpError({ statusCode: 400, message: error.message });
+      }
+      throw new HttpError({ statusCode: 500, message: "Unable to load users" });
+    }
+  };
```
alannnc, line 598: Some clean up for loadUser functions, should be the same except it is easy to read now and has type safe.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -723,25 +744,38 @@ async function handler(
...
-              locale: user.locale ?? "en",
-            },
-          };
-        })
-      : [];
+
+  const teamDestinationCalendar: DestinationCalendar[] = [];
+
+  // Organizer or user owner of this event type it's not listed as a team member.
+  const teamMemberPromises = users.slice(1).map(async (user) => {
```
alannnc, line 751: This was working good just made some little improvements, but teamMemberPromises shouldn't add organiser here.

alannnc: Probably we can rename this const better, so it indicates that all team members should be here except host[0].

packages/core/EventManager.ts
```diff
@@ -311,7 +311,31 @@ export default class EventManager {
...
+                id: destinationCalendar.credentialId,
+              },
+            });
+            if (!dbCredential) {
+              console.error("Credential not found for ID:", destinationCalendar.credentialId);
+              continue;
+            }
+            const credential: CredentialWithAppName = {
+              ...dbCredential,
+              appName: "", // Can appName be more easy to get?
```
alannnc, line 329: Do we have a right way to obtain appName? I think that value should be on DB? @zomars

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -723,25 +744,34 @@ async function handler(
...
-            timeZone: user.timeZone,
-            language: {
-              translate: await getTranslation(user.locale ?? "en", "common"),
-              locale: user.locale ?? "en",
-            },
-          };
-        })
-      : [];
+
+  const teamDestinationCalendar: DestinationCalendar[] = [];
```
joeauyeung, line 748: NIT: I think this should be `teamDestinationCalendars`. It'll make things easier to read and understand

joeauyeung (changes requested):
Great work @alannnc I was going to jump into this but @emrysal shared this with me. 
Some thoughts I had:
- Why restrict this to collective events? Especially if we're checking for `isTeamEvent` wouldn't we want to update everyone's calendars?
- When creating a booking there are two things I noticed
    - If I have a collective event, one of the team members are being adding twice to the booking in the db
    - If we are updating all calendars we should save those calendars as `bookingReferences` which builds into my next point
- In the `eventManager` I think we should also handle rescheduling and canceling of bookings for all calendars.

Other than that great work!

packages/core/EventManager.ts
zomars: We could save it in the DB. Let's open a ticket for this.

PR #7792, opened 2023-03-17
Amend logic to calc slots time already in DST

### LINKED ISSUE #7687: [CAL-1251] bookable timeslots calculated wrong
I have two event types, one is correct and the second one does not count my availability correct.

My availability times are:

2023-03-13 11_22_07-Ninja Warrior Einzeltraining _ Termintyp _ Cal com – Mozilla Firefox

so until 8pm with an 90 minutes event

2023-03-13 11_24_29-Ninja Warrior Einzeltraining _ Parkour Austria _ Cal com – Mozilla Firefox

But the system calculates my last bookable time at 5:30pm => it should be 6:30pm

I've also double and triple checked my time limits and browser cache.
I've also created a whole new availability "work time" but it still does not calculate it correctly, but only for my second event type.. the first one calculates correct.

<sub>[CAL-1251](https://linear.app/calcom/issue/CAL-1251/bookable-timeslots-calculated-wrong)</sub>

roae:
## What does this PR do?

Calculates the DST difference based on the current time and the time of each slot of the organizer and booker

Fixes #7687

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B


CarinaWolli (approved):
Tested and works 👍 Code looks good to me as well

emrysal (approved):
I think this is it 🙏

PR #7811, opened 2023-03-17
Add the new Team->useHostDefault option

emrysal:
## What does this PR do?

Adds a new "Use organiser default conferencing app" intended at falling through and using the organizerUser default conferencing app. Only shows for team events.

image


Jaibles:
@emrysal can we change the label to "Organizer's default app" and have it under the mini title of "Conferencing" instead of "Team"?

We don't use the word "Use" for anything else and in terms of location it's a conferencing option, the fact that it's for a team is handled by our logic, not sure we need a title for it.

emrysal:
Implemented suggestions @Jaibles 🙏

apps/web/components/booking/AvailableEventLocations.tsx
```diff
@@ -8,54 +8,61 @@ import { FiLink } from "@calcom/ui/components/icon";
 
 import type { Props } from "./pages/AvailabilityPage";
 
+const excludeNullValues = (value: unknown) => !!value;
+
 export function AvailableEventLocations({ locations }: { locations: Props["eventType"]["locations"] }) {
```
CarinaWolli, line 13: _Organizer's default app_ is not shown as a location in the booking description. I have _Google Meet_ and _Organizer's default app_ set as locations: 
Screenshot 2023-03-20 at 11 34 02

Everything else looks good to me 👍

emrysal: I do this intentionally - Wouldn't make much sense to a booker to talk about organizers on a round robin event, e.g. Ideally we'd already know who you will be booked with at the booking stage but that's a future improvement 🙏

apps/web/components/booking/AvailableEventLocations.tsx
```diff
@@ -8,54 +8,61 @@ import { FiLink } from "@calcom/ui/components/icon";
...
+    const eventLocationType = getEventLocationType(location.type);
+    if (!eventLocationType) {
+      // It's possible that the location app got uninstalled
+      return null;
+    }
+    if (eventLocationType.variable === "hostDefault") {
+      return null;
+    }
+
+    const translateAbleKeys = [
```
sean-brydon, line 26: NIT 

```suggestion
    const translatableKeys = [
```

sean-brydon (approved):
NIT but LGTM

PR #7812, opened 2023-03-17
Refactor app store & fix admin apps list bugs

joeauyeung:
## What does this PR do?

This PR does the following
- Removes install app button in favour of the default one
- Adds `dirName` to app meta data. This is because in the `App` table `dirName` is a column with a unique constraint. So if apps are left blank then it would cause problems with enable apps
- Fixes bug where as an admin, enabling an app wouldn't happen on the first click
- Fixes bug where as an admin and saving app keys, the schema would not properly be parsed

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


packages/trpc/server/routers/viewer/apps.tsx
```diff
@@ -111,6 +111,7 @@ export const appsRouter = router({
       })
     )
     .mutation(async ({ ctx, input }) => {
+      console.log("🚀 ~ file: apps.tsx:114 ~ .mutation ~ input:", input);
```
zomars, line 114:
```suggestion
```

zomars (commented):
Loving the cleanup so far. Nice job @joeauyeung. Testing locally before approval. 🙏🏽

zomars (changes requested):
Can we prevent the toggle be turned on if the required keys fail to validate?

https://www.loom.com/share/db98ee7bf4a849568350729a161e3109

hariombalhara:
@joeauyeung Thank you so much for this PR ❤️ Been wanting to do it for a long time.

hariombalhara:
> Adds dirName to app meta data. This is because in the App table dirName is a column with a unique constraint. So if apps are left blank then it would cause problems with enable apps

I don't think we need to add dirName to every config. In AppStore Cli, the slug given is used to create the directory, so slug and dirName would be same. Some legacy apps, have different dirName though. 
So, we should simply do `dirName = dirName || slug`

packages/app-store-cli/src/core.ts
```diff
@@ -112,6 +112,7 @@ export const BaseAppFork = {
       isTemplate,
       // Store the template used to create an app
       __template: template,
+      dirName: slug,
```
hariombalhara, line 115: No need to set it in config.json. I want to deprecate that.
seed-app-store.config.json already has it and seeding uses seed-app-store.config.json only. So, it gets the dirName https://github.com/calcom/cal.com/blob/5a94228df61e34d0948f067e7b5ac1fd6060c7ab/packages/prisma/seed-app-store.ts#L308 https://github.com/calcom/cal.com/blob/5a94228df61e34d0948f067e7b5ac1fd6060c7ab/packages/app-store-cli/src/core.ts#L164

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -73,7 +73,7 @@ function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMuta
 
     if (!isOmniInstall) {
       gotoUrl(json.url, json.newTab);
-      return;
+      return { setupPending: externalUrl || json.url.endsWith("/setup") };
```
hariombalhara, line 76: @joeauyeung Can you let me know what exact bug it fixes, I would test that flow.

hariombalhara:
I see 2 more InstallAppButton components that can be removed.
Screenshot 2023-03-22 at 10 38 45 AM

hariombalhara (commented):
Looks good to me. There is one thing mainly
- dirName changes need to be removed
After that I would test on local

packages/app-store-cli/src/core.ts
joeauyeung: I noticed that `dirName` wasn't being added when enabling apps from the admin apps panel. Do we need the `dirName` column if it's just the app slug?

hariombalhara: Certain Legacy apps have different dirname than slug, so yes for now.

joeauyeung:
> Can we prevent the toggle be turned on if the required keys fail to validate?
> 
> https://www.loom.com/share/db98ee7bf4a849568350729a161e3109

https://loom.com/share/35b5cc94d6ae4c97a7eeb9f6d6a92994

packages/app-store/_utils/useAddAppMutation.ts
joeauyeung: When going to the setup pages for certain apps, ex. Apple Cal a toast would appear saying the app was successfully installed even though it wasn't yet. Same thing with apps that went to Oauth pages

packages/app-store/amie/config.json
```diff
@@ -12,5 +12,6 @@
   "email": "support@cal.com",
   "description": "The joyful productivity app\r\r",
   "__createdUsingCli": true,
-  "dependencies": ["google-calendar"]
+  "dependencies": ["google-calendar"],
+  "dirName": "amie"
```
hariombalhara, line 16: I am not sure if you are working on it, but `dirName` should be removed from everywhere.

zomars (approved):

PR #7815, opened 2023-03-18
fix: updating password ui bug #7775

### LINKED ISSUE #7775: [CAL-1272] updating password ui bug

### Issue Summary

1) go to http://localhost:3000/settings/security/password enter old password and new password
2) if you clicked update when both old password and new password you would see error message as expected but even if you change the new password . you are not able to update password.

https://user-images.githubusercontent.com/53316345/225638285-e4d57b84-741f-41dc-98fc-3b9d03a093de.mov

<sub>[CAL-1272](https://linear.app/calcom/issue/CAL-1272/updating-password-ui-bug)</sub>

joe-shajan:
## What does this PR do?
Fixes #7775

Currently, when the old password and new password are the same or the old password is incorrect, the server returns an error. We use the `setError` method from the `useForm` hook to set an `apiError`. However, since this `apiError` is not linked to any of the input fields, it will not trigger validation again when the user types. As a result, the handleSubmit function will not be triggered by React Hook Form because the error is set to true.

To solve this issue, we can check if  `apiError` is true when we submit the form. If it's true, we can clear it before submitting it. This approach is working well.

### Recording: [Link to loom video of the screen recording](https://www.loom.com/share/868ca58d447c454c826ff6ee354893a1)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B


packages/ui/components/form/inputs/Input.tsx
```diff
@@ -328,7 +328,7 @@ type FormProps<T extends object> = { form: UseFormReturn<T>; handleSubmit: Submi
   "onSubmit"
 >;
 
-const PlainForm = <T extends FieldValues>(props: FormProps<T>, ref: Ref<HTMLFormElement>) => {
+const PlainForm = (props: FormProps<{ apiError: string }>, ref: Ref<HTMLFormElement>) => {
```
joeshajan, line 331: I updated this `FormProps<T>`  to `FormProps<{ apiError: string }>`  if I don't do this the typescript will show a warning here `form.clearErrors("apiError");` in line 343. So my doubt is will this cause any other issues?

emrysal: Yes it will, this isn't the right fix unfortunately - we can look at creating a "globalError" utility in the future but by not forwarding the T extends FieldValues it will disable/cause type errors elsewhere. This should be handled in the change password form. 🙏

emrysal (changes requested):

joeshajan:
By changing `form.clearErrors("apiError")` to `form.clearErrors("root")`, the issue was resolved. The `form.clearErrors("root")` function clears all errors prior to form submission. Therefore, I checked for the presence of `apiError` before proceeding.

joeshajan:
Hey @emrysal what do you think about this solution?

emrysal:
A lot better! I wonder what happens when; on successful submission, the reset() function is used. Thinking about it that behaviour may feel very natural in all forms

```javascript
form.reset();
```

joeshajan:
> A lot better! I wonder what happens when; on successful submission, the reset() function is used. Thinking about it that behaviour may feel very natural in all forms
> 
> ```js
> form.reset();
> ```

Really sorry @emrysal I don't get what you meant. Are you asking me to change something?

emrysal:
@joeshajan Well it was more a thought from my end, if you remove the if clause and then add just a reset() - I wonder if that works "better" or similar behaviour.

joeshajan:
> @joeshajan Well it was more a thought from my end, if you remove the if clause and then add just a reset() - I wonder if that works "better" or similar behaviour.

Ok got it. but if we reset the form before submitting I assume that values inside the fields will be removed and then there is nothing to submit. 

So I think `form.reset();` function will remove all the values inside the text field along with errors right?

emrysal (approved):
Yeah you're right, probably not the right fix for now. Merging this 🙇 Thank you!

joeshajan:
Thanks @emrysal 🙂

PR #7820, opened 2023-03-18
Bugfix/format without time zone

emrysal:
## What does this PR do?

* Removes the UTC suffix from the formatted timeZone in Office 365
* Ensures that Zoom start_time is given in UTC

## Zoom
The start time should be provided in ISO 8601 format with the "Z" character at the end to indicate that it is in UTC format. For example, the start time for a meeting on March 24th, 2023 at 4:00 PM Pacific Time (which is 7:00 PM Eastern Time) would be:

```json
"start_time": "2023-03-25T00:00:00Z"
```

Note that in this example, the start time is actually midnight UTC on March 25th, because that is the equivalent of 4:00 PM Pacific Time on March 24th.

If you need to specify a different time zone for the meeting, you can include the "timezone" parameter in the request payload to indicate the desired time zone. The Zoom API will then convert the date-time values to the specified time zone.

### DIFF
--- a/packages/app-store/office365calendar/lib/CalendarService.ts
+++ b/packages/app-store/office365calendar/lib/CalendarService.ts
@@ -253,7 +253,6 @@ export default class Office365CalendarService implements Calendar {
   };
 
   private translateEvent = (event: CalendarEvent) => {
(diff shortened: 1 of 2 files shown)


zomars (approved):
Awesome find @emrysal 🙏

PR #7823, opened 2023-03-19
Fixes vercel url usage on the client

zomars:
## What does this PR do?

- Using CAL_URL in staging always fallback to localhost:3000 since `NEXT_PUBLIC_WEBAPP_URL` isn't defined and `process.env.VERCEL_URL` isn't available on the client 

## Type of change

## How should this be tested?

- [ ] On previews try to click on the copy link or preview event type buttons
- [ ] It should NOT take you to localhost:3000


zomars (commented):
Ready for review

emrysal (approved):
looks good 👍

PR #7824, opened 2023-03-19
Upgrade react-select 

sean-brydon:
Adds classNames styles to classNames prop in select - also adds unstyled. this allows us to not have to important everything under the sun to get the select working

TODO: (before merge) swap out icon components with our own 
TODO: (Followup PR) fix timezone select to use new styles and get rid of `getReactSelectProps` 

https://www.loom.com/share/2ae58016b6c64c518fae47c69f220d79

Booking page dark - timezone:
CleanShot 2023-03-20 at 16 36 16@2x

booking page light - timezone: 
CleanShot 2023-03-20 at 16 38 24@2x

booking page - question select:
CleanShot 2023-03-20 at 16 40 08@2x

Storybook: 
CleanShot 2023-03-20 at 16 40 51@2x


packages/ui/components/form/select/selectTheme.ts
sean-brydon, line 1: Added a new file here - as hopefully soon we will be able to create a generator function to reuse the styles for the select + timezone select.

Atm its not possible as the state changes dont seem to bubble down.

packages/ui/components/form/select/Select.tsx
```diff
@@ -1,109 +1,96 @@
...
   const reactSelectProps = React.useMemo(() => {
     return getReactSelectProps<Option, IsMulti, Group>({
-      className,
       components: components || {},
+      menuPlacement,
     });
-  }, [className, components]);
+  }, [components, menuPlacement]);
 
+  // Annoyingly if we update styles here we have to update timezone select too
```
JeroenReumkens, line 34: Could you explain a bit more in this PR why that's not possible? Would be really great if we could indeed use some sort of generator function so we can't forget to update the timezone select 😁

JeroenReumkens:
Amazing work @sean-brydon! This is SO SO much better!! 👏

PeerRich (approved):
great job, finally!

PR #7827, opened 2023-03-20
fix: remove cal.com from title when branding is disabled

### LINKED ISSUE #7818: Disable branding doesn't remove Cal.com from page title (regression)
good point! we'll see how we can fit that into the roadmap 🙏

_Originally posted by @PeerRich in https://github.com/calcom/cal.com/issues/1290#issuecomment-993644795_

This bug appears to have regressed #1290

G3root:
## What does this PR do?

Fixes #7818 

## Type of change


sean-brydon (approved):
Nice one!

PR #7830, opened 2023-03-20
Fix: Send responses in confirm booking flow

### LINKED ISSUE #7810: [CAL-1287] Differing webhook responses when using "requires confirmation"
Hey guys not sure if you're aware of this issue with the booking creation webhook payload. Everything works great when a normal booking is completed, all custom input responses come through as expected. However, when I change the booking to "Required Confirmation" I don't get the same payload data and am missing some important custom input responses.

Please see the images below and you'll notice "responses" is completely missing when "Require Confirmation" is activated.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1287](https://linear.app/calcom/issue/CAL-1287/differing-webhook-responses-when-using-requires-confirmation)</sub>

hariombalhara:
## What does this PR do?

Fixes #7810

**Environment**: Production

## Type of change

## How should this be tested?

- [x] Make a booking for an Event Type which requires confirmation. Accept the booking and notice that webhook payload has the `responses` and `userFieldsResponses` props.
- [x] Make a booking like the earlier test and before accepting the booking, delete the Event Type. Notice that `responses` and `userFieldsResponses` props are still there.

## Checklist

### DIFF
--- a/packages/features/bookings/lib/getBookingResponsesSchema.ts
+++ b/packages/features/bookings/lib/getBookingResponsesSchema.ts
@@ -9,6 +9,20 @@ type EventType = Parameters<typeof preprocess>[0]["eventType"];
 // eslint-disable-next-line @typescript-eslint/ban-types
 type View = ALL_VIEWS | (string & {});
 
+export const bookingResponse = z.union([
+  z.string(),
+  z.boolean(),
+  z.string().array(),
+  z.object({
+    optionValue: z.string(),
+    value: z.string(),
+  }),
+]);
+
+export const bookingResponsesDbSchema = z.record(bookingResponse);
+
+const catchAllSchema = bookingResponsesDbSchema;
+
 export const getBookingResponsesPartialSchema = ({
   eventType,
   view,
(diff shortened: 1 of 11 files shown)


packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -9,14 +9,28 @@ type EventType = Parameters<typeof preprocess>[0]["eventType"];
...
+  z.string(),
+  z.boolean(),
+  z.string().array(),
+  z.object({
+    optionValue: z.string(),
+    value: z.string(),
+  }),
+]);
+
+export const bookingResponsesDbSchema = z.record(bookingResponse);
```
hariombalhara, line 22: It allows disconnecting current event type data from the responses collected. It thus fixes scenarios like
- Earlier, a  field was of type text and has value "Test" and when we changed the type to "checkbox" booking started showing the value as `false`
- Earlier, a field was there and was recorded in a booking but after removing that field/question, it stopped showing in existing bookings as well.

CarinaWolli (approved):
Nice! Works and code looks good as well 🙏

Issue #5980, opened 2022-12-12
[CAL-1511] Host with base path


Quentin-Guillemin:
### Issue Summary

I want to host cal.com with base path /cal-com. So I needed to edit files (listed in steps to reproduce).
The setup is not working (basePath/auth/setup) on the first install cause the POST request doesn't take care of my added base path. It call `/api/auth/setup` and not `basePah/api/auth/setup`.

Were can I rewrite this ?

### Steps to Reproduce

1. Setup as it is describe on https://github.com/calcom/cal.com#setup.
    .env sould be like :
    ```env
    NEXT_PUBLIC_WEBAPP_URL='https://YOUR_URL/cal-com'
    NEXT_PUBLIC_WEBSITE_URL=''
    NEXT_PUBLIC_CONSOLE_URL=''
    NEXT_PUBLIC_EMBED_LIB_URL='https://YOUR_URL/cal-com/embed/embed.js'
    NEXTAUTH_URL=''
    ```

1. Add `basePath: "/cal-com",` in `apps/web/next.config.js` on line 82 to obtain :
    ```js
    const nextConfig = {
        basePath: "/cal-com",
        ...
    }
    ```

3. Edit `apps/web/lib/app-providers.tsx` on line 75 to obtain :
    ```tsx
    <SessionProvider session={session || undefined} basePath={`${process.env.NEXT_PUBLIC_WEBAPP_URL}/api/auth`}>
    ```

4. Edit value of const `url` in `apps/web/pages/_app.tsx` to obtain :
    ```tsx
    const url = `${process.env.NEXT_PUBLIC_WEBAPP_URL}/api/trpc`;
    ```

### Technical details

- Browser version : Firefox 107.0.1
- Node.js version : 18.12.1
- POST request :
```json
{
  "log": {
    "version": "1.2",
    "creator": {
      "name": "Firefox",
      "version": "107.0.1"
    },
    "browser": {
      "name": "Firefox",
      "version": "107.0.1"
    },
    "pages": [
      {
        "startedDateTime": "2022-12-12T09:27:23.000+01:00",
        "id": "page_1",
        "title": "https://YOUR_URL/cal-com/auth/setup",
        "pageTimings": {
          "onContentLoad": 301,
          "onLoad": 1807
        }
      }
    ],
    "entries": [
      {
        "pageref": "page_1",
        "startedDateTime": "2022-12-12T09:27:23.000+01:00",
        "request": {
          "bodySize": 116,
          "method": "POST",
[...]

PR #7838, opened 2023-03-20
feat: Add uri on self-hosting

Quentin-Guillemin:
## What does this PR do?

Add `NEXT_PUBLIC_WEBAPP_URI` env variable to be able to host cal.com on a URI.

Fixes # (issue)

https://github.com/calcom/cal.com/issues/5980

## Type of change

## How should this be tested?

- [ ] Build with URI and test each route

## Checklist


packages/lib/constants.ts
```diff
@@ -2,13 +2,14 @@ const VERCEL_URL = process.env.NEXT_PUBLIC_VERCEL_URL ? `https://${process.env.N
 const RAILWAY_STATIC_URL = process.env.RAILWAY_STATIC_URL ? `https://${process.env.RAILWAY_STATIC_URL}` : "";
 const HEROKU_URL = process.env.HEROKU_APP_NAME ? `https://${process.env.HEROKU_APP_NAME}.herokuapp.com` : "";
 const RENDER_URL = process.env.RENDER_EXTERNAL_URL ? `https://${process.env.RENDER_EXTERNAL_URL}` : "";
+export const WEBAPP_URI = process.env.NEXT_PUBLIC_WEBAPP_URI || "";
```
hariombalhara, line 5: I think we should call it something like WEBAPP_PREFIX_PATH

hariombalhara: Also, it might be a nice to have a validation for it that it must start with /

apps/web/components/apps/App.tsx
```diff
@@ -127,7 +127,7 @@ const Component = ({
         <div className="mb-8 flex pt-4">
           <header>
             <div className="mb-4 flex items-center">
-              <img className="min-h-16 min-w-16 h-16 w-16" src={logo} alt={name} />
+              <img className="min-h-16 min-w-16 h-16 w-16" src={WEBAPP_URI + logo} alt={name} />
```
hariombalhara, line 130: Can we not remove all these usages of WEBAPP_URI by setting base href. My concern is that developers will certainly forget adding that WEBAPP_URI

hariombalhara (commented):
Thanks @Quentin-Guillemin for the contribution 🙏. I have left some comments

zomars (changes requested):
Agreed that we should support hosting on a different base path. Although I think we can achieve this without too much destructive changes. Also we should make this opt-in so people who are not using it aren't affected by these changes.

Blocking until these concerns and @hariombalhara comments are addressed.

Thanks again for your contribution 🙏🏽

hariombalhara:
@Quentin-Guillemin I was talking about using base HTML tag. Found an article that recommends that https://skryvets.com/blog/2018/09/20/an-elegant-solution-of-deploying-react-app-into-a-subdirectory/

But there is a problem with this approach as well, we would have to convert all paths to relative paths, which is also not straightforward.

One approach might be to just prefix every image src with WEBAPP_URL and have a linting rule that ensures it is always done.

I would have to test more to understand what approach will work best.

Quentin-Guillemin:
Hi @hariombalhara,

Have you chosen a solution ?

hariombalhara:
@Quentin-Guillemin I am afraid, I couldn't prioritize it yet.

keithwillcode:
@hariombalhara @zomars Are we good to come back and review this once more?

hariombalhara:
@keithwillcode There are just too many changes with the current approach. Unless we figure out a way to solve it in a trivial manner, we should avoid the implementation as it isn't a very common requirement.

PeerRich:
> @keithwillcode There are just too many changes with the current approach. Unless we figure out a way to solve it in a trivial manner, we should avoid the implementation as it isn't a very common requirement.

I agree with @hariombalhara here. So much goin on here with little known potential side-effects. Very niche request.

@Quentin-Guillemin this seems like it's solving a problem for yourself, I think its smart to keep this fix in your own personal fork for now! Thank you for the effort of raising this PR tho!

closing

PR #5105, opened 2022-10-19
Abstracting booking cancellation

leog:
## What does this PR do?

Abstracted cancellation logic to be able to unify logic for API.

API PR https://github.com/calcom/api/pull/191 depends on this.

**Environment**: Staging(main branch) / Production

## Type of change

- [X] Chore (refactoring code, technical debt, workflow improvements)

PR #7839, opened 2023-03-20
No delete of booking on seated events, fix id->uid

emrysal:
## What does this PR do?

* Fixes orphaned cancelled events in the /bookings list.

@leog If you still remember, could you provide the reasoning behind the uid -> id change? I miss that context. https://github.com/calcom/cal.com/pull/5105/files#diff-b56ec7c13bd93d1e15d6398bf474cc86185a6178f6043f6d99071e8189c8f738

@alannnc Fixup to seated events to not delete the original event (just cancel)


alishaz-polymath (approved):
Looks good to me but waiting to hear from @alannnc and @leog as they might have kor context why we used both id and UID in the first place.

PR #7207, opened 2023-02-19
Fix - add team members to emails

joeauyeung:
## What does this PR do?

This PR fixes a bug where team members were not included in the email and calendar bodies. This PR allows sending team emails to the member's set language.

image

image

Fixes #6456

**Environment**: Staging(main branch) / Production

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- Create both a collective and a round-robin event
    - For the round-robin event, set a fixed host
    - For one of the team members to a different language
- For both event types do the following actions and ensure that the emails contain team members
    - Request booking
    - Create booking
    - Reschedule
    - Delete 

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #7841, opened 2023-03-20
Hotfix: Adds missing team members to gCal event

zomars:
## What does this PR do?

- Re-adds team members to calendar invite that were removed in #7207

## Type of change

## How should this be tested?

- [ ] Create a team collective event and add team members to it
- [ ] Connect a google calendar
- [ ] Create an event, all team members should appear as part of the calendar event invitees

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works (Will do in a follow up)

### DIFF
--- a/packages/app-store/googlecalendar/lib/CalendarService.ts
+++ b/packages/app-store/googlecalendar/lib/CalendarService.ts
@@ -75,6 +75,13 @@ export default class GoogleCalendarService implements Calendar {
       ...rest,
       responseStatus: "accepted",
     }));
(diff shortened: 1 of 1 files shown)


zomars (commented):
Self-review done

emrysal (approved):
Approved for now; still I think we should keep the calendar event dumber, so eventually move this logic back out.

zomars:
> Approved for now; still I think we should keep the calendar event dumber, so eventually move this logic back out.

Agreed. Hence the "hot fix".

PR #7851, opened 2023-03-21
Fix rescheduling updating calendars & duplicate bookings

joeauyeung:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- Reschedule a booking with a destination calendar

## Checklist


packages/core/EventManager.ts
```diff
@@ -439,7 +461,7 @@ export default class EventManager {
         });
       }
 
-      if (newBooking) {
+      if (newBooking?.references.length) {
```
joeauyeung, line 464: Only look through the new booking, if references are created

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1650,7 +1650,7 @@ async function handler(
 
     // Use EventManager to conditionally use all needed integrations.
 
-    const updateManager = await eventManager.reschedule(evt, originalRescheduledBooking.uid, booking?.id);
```
joeauyeung, line 1653: The third parameter `newBookingId` is only used when rescheduling seated bookings. This was causing seated logic on normal bookings

packages/core/EventManager.ts
```diff
@@ -279,6 +279,27 @@ export default class EventManager {
...
+        bookingId: booking.id,
+      },
+    });
+    const attendeeDeletes = prisma.attendee.deleteMany({
+      where: {
+        bookingId: booking.id,
+      },
+    });
+
+    const bookingDeletes = prisma.booking.delete({
```
emrysal, line 294: Only this delete is needed, attendee & bookingReference are in Cascade and will be auto-deleted 👍

emrysal (dismissed):
Do we want to delete the original booking on reschedule? Why doesn't a reschedule just update the same booking at a different time 😕 .

packages/core/EventManager.ts
joeauyeung: I realize that isn't the right thing to do as well. Just pushed the change

packages/core/EventManager.ts
```diff
@@ -423,6 +423,7 @@ export default class EventManager {
     booking: PartialBooking,
     newBookingId?: number
   ): Promise<Array<EventResult<NewCalendarEventType>>> {
+    newBookingId = undefined;
```
emrysal, line 426: I don't know; are you still working on this? Setting this to undefined makes all kinds of dead code - please draft if so 👍

joeauyeung: Oh that's just me experimenting. Thanks for catching

packages/core/EventManager.ts
```diff
@@ -439,7 +439,7 @@ export default class EventManager {
         });
       }
 
-      if (newBooking) {
+      if (newBooking?.references.length) {
         calendarReference = newBooking.references.find((reference) => reference.type.includes("_calendar"));
       } else {
         // Bookings should only have one calendar reference
```
alishaz-polymath, line 445:
```suggestion
       calendarReference = newBooking?.references.length
         ? newBooking.references.find((reference) => reference.type.includes("_calendar"))
         : booking.references.find((reference) => reference.type.includes("_calendar"));
```
Minor NIT suggestion

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1650,7 +1650,7 @@ async function handler(
...
+    const updateManager = await eventManager.reschedule(evt, originalRescheduledBooking.uid);
```
alishaz-polymath, line 1653: How does this change handle rescheduling for seated bookings? Or is that being handled elsewhere?

alishaz-polymath (approved):
Apart from NITs, it looks good. Tested locally, works

packages/features/bookings/lib/handleNewBooking.ts
joeauyeung: It's being handled in the reschedule method, having that third parameter triggers that logic to run.

emrysal (approved):
LGTM :shipit:

PR #7870, opened 2023-03-21
Theme fixes

### LINKED ISSUE #7858: [CAL-1304] bug: theme is switching
https://app.birdeatsbug.com/sessions/c9UBYrqRCDzD3m3ns8RrJa8joHtx03nnQ-X6GSFMKtU4

not sure why

<sub>[CAL-1304](https://linear.app/calcom/issue/CAL-1304/bug-theme-is-switching)</sub>

hariombalhara:
## What does this PR do?

Fixes #7858

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


apps/web/components/booking/TimeOptions.tsx
```diff
@@ -15,8 +13,7 @@ type Props = {
 
 const TimeOptions: FC<Props> = ({ onSelectTimeZone }) => {
   const [selectedTimeZone, setSelectedTimeZone] = useState("");
-  const query = useMeQuery();
-  const userTheme = useTheme(query?.data?.theme).resolvedTheme;
+  const userTheme = useTheme().resolvedTheme;
```
hariombalhara, line 16: TimeOptions is used on booking page and booking page theming is based on the organizer being booked and not the logged in user.

apps/web/lib/app-providers.tsx
```diff
@@ -67,12 +67,15 @@ const AppProviders = (props: AppPropsWithChildren) => {
...
+  // @ts-ignore
+  const bookingPageSlug = props.pageProps.profile?.username || "";
   // Use namespace of embed to ensure same namespaced embed are displayed with same theme. This allows different embeds on the same website to be themed differently
   // One such example is our Embeds Demo and Testing page at http://localhost:3100
   // Having `getEmbedNamespace` defined on window before react initializes the app, ensures that embedNamespace is available on the first mount and can be used as part of storageKey
   const embedNamespace = typeof window !== "undefined" ? window.getEmbedNamespace() : null;
-  const storageKey = typeof embedNamespace === "string" ? `embed-theme-${embedNamespace}` : "theme";
-
+  const storageKey =
+    typeof embedNamespace === "string" ? `embed-theme-${embedNamespace}` : `theme-${bookingPageSlug}`;
```
hariombalhara, line 78: next-themes persist in localStorage which is shared for the domain. 
Due to this we can have 2 pages opened simultaneously with 2 different themes but they are trying to use the same localStorage item. If the 2 booking pages have conflicting themes configured they would forever try to change each other's value.

To avoid this, I plan to use profile.username as a differentiator. For team pages, it can be team name.

@emrysal @zomars What do you think of this approach?

packages/lib/hooks/useTheme.tsx
sean-brydon, line 1: Will it be easier to note that we will be using system theme i believe shortly for most of the authed app? Not sure if this has any effect here or not

hariombalhara: From the comment on threads:

Theming is very complex because of the following reasons:
- App can't use themes. It has to be light only
- Booking Pages have to have different themes for different users
- Next-theme uses localStorage to persist the preference(to avoid Flash of unthemed content).
   - But localStorage is shared across the domain (specifically origin) and thus one theme persisted by Next-Theme impacts the theme of other booking page.
- Then there are embeds which can have different theme than configured in app because there is an API to switch the theme for embed only
- There can be multiple embeds having different themes configured and they have to work with those themes

That would fix just the first point and would also eliminate the need of setting isThemeSupported for booking pages, as isThemeSupported would always be true. But it would certainly simplify the code and avoid possible bugs in future as we would be designing all components with theme support in mind.

sean-brydon: All components in future will have automatic theme support with tokens :) The way it is currently in my branch 

`bg-subtle text-emphasis` handles all themes needed

PeerRich:
@hariombalhara whats needed here to leave draft?

hariombalhara:
Waiting on @sean-brydon's dark mode to merge now.

hariombalhara:
Closing in favour of #8108 which handles everything except the case where different booking pages can have different themes and that makes them switch.

PR #7873, opened 2023-03-22
fix: metadata get lost when rescheduling

### LINKED ISSUE #7867: [CAL-1312] Booking metadata gets lost on reschedule
We want to store some metadata in the 'Booking' object. Here are some steps:
We append metadata query parameter to the booking link on our side before sending it to the user;
User opens the link and creates a meeting -> 'booking' is created with metadata object;
User clicks on the "Reschedule" button;
It redirects them to the initial page where they update date of the meeting;
After saving changes metadata disappears from the Booking object.
It happened because the user clicked on the "Reschedule" button they were redirected without including metadata to the link. I see two possible options:
Include metadata to the booking link when rescheduling the event;
Do patch for metadata when updating 'Booking', meaning to merge the previous metadata value with a new one.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1312](https://linear.app/calcom/issue/CAL-1312/booking-metadata-gets-lost-on-reschedule)</sub>

G3root:
## What does this PR do?

Fixes #7867

## Type of change

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -1485,6 +1485,9 @@ async function handler(
       newBookingData.recurringEventId = reqBody.recurringEventId;
     }
     if (originalRescheduledBooking) {
+      newBookingData.metadata = {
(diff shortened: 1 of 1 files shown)


CarinaWolli (approved):
LGTM!

PR #7871, opened 2023-03-21
Handle empty label from Custom inputs

hariombalhara:
Handle empty label during migration

PR #7874, opened 2023-03-22
fix: Custom Input Migration - Make sure that label is also set

hariombalhara:
## What does this PR do?

Fixes # (issue)

Follow-up of #7871. Even though zod allows empty label, we don't allow empty label(see https://github.com/calcom/cal.com/blob/351b4228d4d23dac052f3ccf79c4af4e7e3a5063/packages/features/bookings/lib/handleNewBooking.ts#L396) during booking because label is shown in email. So, generate a proper label as well/

**Environment**:  Production

## Type of change


emrysal (approved):
Hmm, 👍 lgtm

PR #7875, opened 2023-03-22
Fix timezone select - create two variants

sean-brydon:
CleanShot 2023-03-22 at 14 30 27@2x
CleanShot 2023-03-22 at 14 46 35@2x
CleanShot 2023-03-22 at 14 46 45@2x

--- 
Fixed: changes 
CleanShot 2023-03-25 at 13 52 19@2x
CleanShot 2023-03-25 at 13 53 55@2x


packages/ui/form/TimezoneSelect.tsx
sean-brydon, line 1: Duplicate

Udit-takkar (dismissed):
Screenshot 2023-03-24 at 11 36 19 PM

Few minor issues that needs to be fixed

1) In light mode the text color of timezone select is darker than the above text( Cal Video, 30 minutes)
2) can you increase the width of the dropdown component?

roae (approved):
LGTM! 🚀

emrysal (approved):
Approved :)

PR #7888, opened 2023-03-22
fix: remove "Nameless" from  title in booking cancel page

G3root:
## What does this PR do?
fixes the issue as mentioned in https://threads.com/34457815439?u=34440454662

## Type of change

### DIFF
--- a/apps/web/pages/booking/[uid].tsx
+++ b/apps/web/pages/booking/[uid].tsx
@@ -485,7 +485,7 @@ export default function Success(props: SuccessProps) {
                       </>
                     )}
                     <div className="font-medium">{t("what")}</div>
-                    <div className="col-span-2 mb-6 last:mb-0">{eventName}</div>
+                    <div className="col-span-2 mb-6 last:mb-0">{props.bookingInfo.title}</div>
                     <div className="font-medium">{t("when")}</div>
                     <div className="col-span-2 mb-6 last:mb-0">
                       {reschedule && !!formerTime && (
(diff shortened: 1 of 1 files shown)


apps/web/pages/booking/[uid].tsx
```diff
@@ -527,7 +527,7 @@ export default function Success(props: SuccessProps) {
                             </div>
                           )}
                           {bookingInfo?.attendees.map((attendee) => (
-                            <div key={attendee.name} className="mb-3 last:mb-0">
+                            <div key={attendee.name + attendee.email} className="mb-3 last:mb-0">
```
Comment (G3root) [inline apps/web/pages/booking/[uid].tsx:530]: fixes the same key warning when additional guests are present

apps/web/pages/booking/[uid].tsx
```diff
@@ -485,7 +485,7 @@ export default function Success(props: SuccessProps) {
                       </>
                     )}
                     <div className="font-medium">{t("what")}</div>
-                    <div className="col-span-2 mb-6 last:mb-0">{eventName}</div>
+                    <div className="col-span-2 mb-6 last:mb-0">{props.bookingInfo.title}</div>
```
Comment (hariombalhara) [inline apps/web/pages/booking/[uid].tsx:488]: @G3root can you comment why nameless was coming and how switching from eventName to props.bookingInfo.title fixed it?

Also, is it something that recently regressed or like this always.

hariombalhara (changes requested):
Need clarity on change

apps/web/pages/booking/[uid].tsx
Comment (G3root) [reply apps/web/pages/booking/[uid].tsx:488]: previously event name has been derived from [here](https://github.com/calcom/cal.com/blob/1bf50e09cdeb4eb06683db080e9bd61abe991671/apps/web/pages/booking/%5Buid%5D.tsx#LL217-L220C20), in the event of cancelled booking , attendees are empty so it fallbacks to `Nameless`. using the title of the booking seems reliable here.

i think it is recently regressed.

CarinaWolli (approved):
Looks good to me 👍

hariombalhara (approved):

PR #7890, opened 2023-03-22
fix: add belongsToActiveTeam

### LINKED ISSUE #7884: [CAL-1315] belongs to Active team is not set for google sign in

### Issue Summary

[https://threads.com/34454702096?s=efVDCvYtY35GbWZmashtof](https://threads.com/34454702096?s=efVDCvYtY35GbWZmashtof)

Due to this bug some users are getting 403 error while downloading recordings

### Technical details

* Browser version, screen recording, console logs, network requests: You can make a recording with [Bird Eats Bug](https://birdeatsbug.com/).
* Node.js version
* Anything else that you think could be an issue.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1315](https://linear.app/calcom/issue/CAL-1315/belongs-to-active-team-is-not-set-for-google-sign-in)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/7884

Screenshot 2023-03-23 at 2 26 50 AM

for google sign-in the belongsToActiveTeams property was not set

## Type of change

### DIFF
--- a/packages/features/auth/lib/next-auth-options.ts
+++ b/packages/features/auth/lib/next-auth-options.ts
@@ -391,6 +391,11 @@ export const AUTH_OPTIONS: AuthOptions = {
             name: true,
             email: true,
             role: true,
+            teams: {
+              include: {
+                team: true,
+              },
+            },
           },
         });
 
@@ -398,9 +403,23 @@ export const AUTH_OPTIONS: AuthOptions = {
           return token;
         }
 
+        // Check if the existingUser has any active teams
+        const belongsToActiveTeam =
+          existingUser.teams.filter((m: { team: { metadata: unknown } }) => {
+            if (!IS_TEAM_BILLING_ENABLED) return true;
+            const metadata = teamMetadataSchema.safeParse(m.team.metadata);
+            if (metadata.success && metadata.data?.subscriptionId) return true;
+            return false;
+          }).length > 0;
+
         return {
-          ...existingUser,
+          id: existingUser.id,
+          username: existingUser.username,
+          name: existingUser.name,
+          email: existingUser.email,
+          role: existingUser.role,
           ...token,
+          belongsToActiveTeam,
         };
       };
       if (!user) {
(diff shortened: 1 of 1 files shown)


packages/features/auth/lib/next-auth-options.ts
```diff
@@ -391,16 +391,35 @@ export const AUTH_OPTIONS: AuthOptions = {
...
+            },
           },
         });
 
         if (!existingUser) {
           return token;
         }
 
+        // Check if the existingUser has any active teams
+        const belongsToActiveTeam =
```
hariombalhara, line 407: Let's abstract out the code from line 160 into a fn in this module and reuse that here.

hariombalhara (dismissed):
Left a comment

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -391,16 +391,35 @@ export const AUTH_OPTIONS: AuthOptions = {
...
+          existingUser.teams.filter((m: { team: { metadata: unknown } }) => {
+            if (!IS_TEAM_BILLING_ENABLED) return true;
+            const metadata = teamMetadataSchema.safeParse(m.team.metadata);
+            if (metadata.success && metadata.data?.subscriptionId) return true;
+            return false;
+          }).length > 0;
+
         return {
-          ...existingUser,
+          id: existingUser.id,
```
hariombalhara, line 416: cc @emrysal @zomars This is a critical place to change, looks good to me, but you should check too.

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -1,5 +1,6 @@
 import type { UserPermissionRole } from "@prisma/client";
```
emrysal, line 1:
```suggestion
import type { UserPermissionRole, Membership, Team } from "@prisma/client";
```
NIT

packages/features/auth/lib/next-auth-options.ts
emrysal: I'm not sure why we shouldn't continue to just ...existingUser, belongsToActiveTeam will append. 💭

emrysal (changes requested):
It works, but please keep this logic as consistent to before as possible.

Udit-takkar:
@emrysal don't want to pass the teams field.

emrysal:
> @emrysal don't want to pass the teams field with existingUser

Ah makes sense, use deconstructing 👍 

`const { teams, ...existingUser } = ...`

optionally then cleanup using
`delete teams;`

emrysal (approved):
LGTM, looking good 👍

PR #7894, opened 2023-03-23
Techdebt/bundle size experiments

emrysal:
## What does this PR do?

* Removed locales from @calcom/dayjs (~2.3% package size reduce)

* Lodash does not tree shake correctly when you do `import { debounce } from "lodash"` - but it does when (which we did inconsistently throughout) using  `import merge from "lodash/merge"`. This is where dynamicImports shines, so we now (will) have the best of both worlds. 

  Results in a meaningful ~ -7% drop in the size of affected page entries.

* /settings/my-account/profile included `crypto` into the client side. Removing this reduced this entry 35%.


apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -1,6 +1,5 @@
 import { zodResolver } from "@hookform/resolvers/zod";
 import { IdentityProvider } from "@prisma/client";
-import crypto from "crypto";
```
emrysal, line 3: ~ -35% on the profile.tsx page 👍

packages/core/CalendarManager.ts
```diff
@@ -54,8 +54,8 @@ export const getConnectedCalendars = async (
           };
         }
         const cals = await calendar.listCalendars();
-        const calendars = _(cals)
```
emrysal, line 57: This one was a bad one, but serverside only 🤔

zomars (approved):
Noice

PR #7900, opened 2023-03-23
Added starting concept of auto discovery

copoer:
I had an issue connecting cal.com to AWS WorkMail so created this as a example of a potential solution using auto discovery. **This is not finished and is only and example**


copoer:
This is a example for the following ticket:
https://github.com/calcom/cal.com/issues/7901

copoer:
I found an alternative solution

PowerAppsDarren:
What was your solution? I having a hard time connecting my M365.

PR #6511, opened 2023-01-16
/insights

PeerRich:
- [x] test https://www.tremor.so/
- [x] add upgrade screen
- [x] componentise 
- [x] add data @alannnc @sean-brydon 

CleanShot 2023-01-16 at 16 40 07@2x

Update:
image

PR #7902, opened 2023-03-23
Removes the getUser logic from createContext, instead use isAuthed

emrysal:
## What does this PR do?

Intended to speed up unprotected routes when a session is available.

### DIFF
--- a/packages/trpc/server/trpc.ts
+++ b/packages/trpc/server/trpc.ts
@@ -105,13 +105,14 @@ const perfMiddleware = t.middleware(async ({ path, type, next }) => {
 
 const isAuthed = t.middleware(async ({ ctx: { session, locale, ...ctx }, next }) => {
   const user = await getUserFromSession({ session });
-  if (!user) {
+  if (!user || !session) {
     throw new TRPCError({ code: "UNAUTHORIZED" });
   }
   const i18n =
     user.locale && user.locale !== locale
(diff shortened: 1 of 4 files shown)


zomars (approved):
Very nice improvements 🙏🏽

packages/trpc/server/trpc.ts
```diff
@@ -40,41 +135,10 @@ const isAdminMiddleware = isAuthedMiddleware.unstable_pipe(({ ctx, next }) => {
...
-      },
-    });
-  });
-
 export const router = t.router;
 export const mergeRouters = t.mergeRouters;
 export const middleware = t.middleware;
 export const publicProcedure = t.procedure.use(perfMiddleware);
-export const authedProcedure = t.procedure.use(perfMiddleware).use(isAuthedMiddleware);
-export const authedRateLimitedProcedure = ({ intervalInMs, limit }: IRateLimitOptions) =>
```
zomars, line 74: @emrysal why was this removed?

emrysal: Looking in the code it's unused and also we can (and should) handle this on the middleware level of Vercel 👍  - just too low to be effective at rate limiting at this point.

zomars: It is used in the upcoming #6511 although we should come up with a proper replacement first.

PR #7903, opened 2023-03-23
[CAL-1222] Timezone dropdown shows london several times

### LINKED ISSUE #7660: [CAL-1222] Timezone dropdown shows london several times
### Issue Summary

When you search 'London' in the timezone dropdown it shows 3 times 'London GMT' but two of them are wrong. Two of them are timezone -5 and when clicking them it shows America/Detroit 

Screenshot 2023-03-10 at 10 16 13

<sub>[CAL-1222](https://linear.app/calcom/issue/CAL-1222/timezone-dropdown-shows-london-several-times)</sub>

gitstart-calcom:
## What does this PR do?

Fixes #7660 

https://www.loom.com/share/d8166e2baf1241f3aa6fb3a856e1d769


packages/trpc/server/routers/viewer.tsx
```diff
@@ -158,7 +158,21 @@ const publicViewerRouter = router({
...
+    const maxPops: { [key: string]: { city: string; timezone: string; pop: number } } = {};
+    cityMapping.forEach((city) => {
+      const cityPop = city.pop;
+      if (maxPops[city.city]?.pop === undefined || cityPop > maxPops[city.city].pop) {
+        maxPops[city.city] = { city: city.city, timezone: city.timezone, pop: city.pop };
+      }
+    });
+    const uniqueCities = Object.values(maxPops);
+    uniqueCities.forEach((city) => {
+      if (city.city === "London") city.timezone = "Europe/London";
```
PeerRich, line 171: if we find more cities, can we add them here?

PeerRich (approved):
LGTM but waiting for @zomars

packages/trpc/server/routers/viewer.tsx
gitstart-calcom: Yes

packages/trpc/server/routers/viewer.tsx
```diff
@@ -159,7 +159,21 @@ const publicViewerRouter = router({
...
   // REVIEW: This router is part of both the public and private viewer router?
   slots: slotsRouter,
-  cityTimezones: publicProcedure.query(() => cityMapping),
+  cityTimezones: publicProcedure.query(() => {
+    const maxPops: { [key: string]: { city: string; timezone: string; pop: number } } = {};
+    cityMapping.forEach((city) => {
+      const cityPop = city.pop;
+      if (maxPops[city.city]?.pop === undefined || cityPop > maxPops[city.city].pop) {
+        maxPops[city.city] = { city: city.city, timezone: city.timezone, pop: city.pop };
+      }
```
zomars, line 168: Can we add some explainers on what's happening here? It is kinda difficult to tell what's going on in the code.

gitstart-calcom: We are filtering out all cities that have the same "city" key and only using the one with the highest population. This way we return a new array of cities without running the risk of having more than one city with the same name on the dropdown and prevent users from mistaking the time zone of the desired city

zomars: Can we put that in the code comments?

packages/trpc/server/routers/viewer.tsx
```diff
@@ -159,7 +159,21 @@ const publicViewerRouter = router({
     }),
   // REVIEW: This router is part of both the public and private viewer router?
   slots: slotsRouter,
-  cityTimezones: publicProcedure.query(() => cityMapping),
+  cityTimezones: publicProcedure.query(() => {
+    const maxPops: { [key: string]: { city: string; timezone: string; pop: number } } = {};
+    cityMapping.forEach((city) => {
+      const cityPop = city.pop;
```
zomars, line 165: Let's try being more explicit here:

```suggestion
      const cityPopulationCount = city.pop;
```

zomars (approved):
Lazy loaded `city-timezones` (thanks @roae for the tip), added comments and improved legibility a bit.

PR #7907, opened 2023-03-23
Add detailed error logs to handleNewBooking

roae:
## What does this PR do?

The objective is to have more details when the error "No available users found." occurs and to know if it is by time zone and daylight saving time.


emrysal (approved):
Pending minor nit lgtm

Issue #7901, opened 2023-03-23
Microsoft Exchange: AWS WorkMail


copoer:
### Issue Summary

I am unable to connect cal.com to my AWS WorkMail Calendar. 

I have a tried all three Exchange Plugins:
- Microsoft Exchange
- Microsoft Exchange 2013 Calendar
- Microsoft Exchange 2016 Calendar

I have tried the following URLs
1. https://ews.mail.us-east-1.awsapps.com/EWS/Exchange.asmx
2. https://outlook.mail.us-east-1.awsapps.com

I get one of the following errors:

Microsoft Exchange;Basic Auth;URL 1:
The specified server version is invalid.

Microsoft Exchange;NTML;URL 1:
Invalid message signature: Basic realm="EWS"

Microsoft Exchange;Basic Auth;URL 2:
Could not add this exchange account

Microsoft Exchange;NTML;URL 2:
www-authenticate not found on response of second request

Microsoft Exchange 2013 Calendar;URL 1:
Could not add this exchange account

Microsoft Exchange 2013 Calendar;URL 2:
Could not add this exchange account

Microsoft Exchange 2016 Calendar;URL 1:
Could not add this exchange account

Microsoft Exchange 2016 Calendar;URL 2:
Could not add this exchange account

Documentation of AWS Workmail states that manually connecting is possible, but it seems like it results in an error for all the current options

https://docs.aws.amazon.com/workmail/latest/userguide/outlook-client.html

My guess is you may need to implement exchange auto discovery? The underlying libaray cal.com uses does allow for it. Here is an example PR I made:
https://github.com/calcom/cal.com/pull/7900

Has anyone else had any success connecting cal.com to AWS WorkMail?

### Technical details

- Browser version Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36

copoer:
This is most likely related:
https://github.com/calcom/cal.com/issues/5312

copoer:
I found a solution to this issue. The exchange version needs to be set to 2010 for it to function. I am making a PR now.

Issue #5312, opened 2022-11-01
[CAL-643] Adding Exchange app fails - calendly works :/ 


dkgrieshammer:
Adding exchange connector app fails with 401 on cal.com (free web version);
Also tried outlook connector and legacy exchange apps multiple times, don’t work either. 
Adding the same exchange/credentials to Apple calendar, fantastical, (calendly just to test connector :) is working fine. 

I realize that only cal.com requires full path of echange server asmx file, others just use owa.tld.com (guess that means they have auto-discovery?)

University Admin desk couldn’t help either since it’s working fine with all other services it seems to be a problem in cal.com‘s connector? Any help or hint much appreciated 🙏😚

<sub>[CAL-643](https://linear.app/calcom/issue/CAL-643/adding-exchange-app-fails-calendly-works)</sub>

milospuac:
Hi there, thanks for flagging this. Our team will investigate the issue

kopfpolster:
Hi! I have the same issue. calendly uses owa.tld.com and it works. Im Using Exchange 2016 Server.
Thank you for investigating.

rimkov:
hi! i have the same issue with our local exchange server 2019. calendso (3.15.2) receive a 401 unathorized error.
thank you for your support.

RiRe:
Same problem here

saschaschworm:
Hi. I authored this integration and it works fine with Microsoft Exchange Server 2016 (15.1.2507.16) using Cal.com 2.3.5 on-premises. Exchange on-premises servers (like the one used by the university I work for) often have basic authentication disabled and require NTLM. Have you tried this? Either with gzip compression enabled or disabled?

dkgrieshammer:
Hi Sascha, yes I tried all options, also on multiple days and cross checked with calendly to be sure it’s not a hiccup on our servers side.Thanks for your work on this btw 👏

saschaschworm:
I guess I overfitted this integration to work fine with the Exchange on-premises server I'm currently using but failed to generalize it to other possible scenarios. Hard to debug though because configurations vary across servers and I can only access one specifically. However, I'll refactor this integration to make use of the Autodiscover Service and to match the on-boarding process to the one users are more familiar with. Want to assign me to this issue, @PeerRich?

PeerRich:
sure go for it!

dkgrieshammer:
[...]

PR #7908, opened 2023-03-23
feat: Added option to select the exchange server version

copoer:
## What does this PR do?

Currently, the exchange addon is hard coded to one version of exchange. This prevents users from connecting to older versions of the exchange servers. This feature enables users to select their own version of exchange.  

Fixes 
https://github.com/calcom/cal.com/issues/7901
Potentially fixes some of these issue
https://github.com/calcom/cal.com/issues/5312

## Type of change

## How should this be tested?

Attempt to connect an account to an exchange server and select a specific version. I tested it using AWS WorkMail. In order for it to work on AWS WorkMail I had to select version 2010 SP 2 or older


packages/app-store/exchangecalendar/lib/CalendarService.ts
```diff
@@ -198,7 +198,7 @@ export default class ExchangeCalendarService implements Calendar {
   }
 
   private getExchangeService(): ExchangeService {
-    const service: ExchangeService = new ExchangeService();
+    const service: ExchangeService = new ExchangeService(this.payload.exchangeVersion);
```
joeauyeung, line 201:
```suggestion
    const service: ExchangeService = new ExchangeService(ExchangeVersion[this.payload.exchangeVersion]);
```

@copoer thank you so much for your contribution. When initiating the calendar service I noticed that the number was being passed and not the Exchange version.

copoer: The ExchangeService constructor is expecting a enum not a string. This does not need to be added.
https://github.com/gautamsi/ews-javascript-api/blob/9488cdf389f16a7daa8a17299ab195a487b9b30f/typings/ExchangeWebService.d.ts#L241

joeauyeung (approved):
Thanks for looking into this. LGTM, thanks again @copoer

PR #7910, opened 2023-03-23
Delete rescheduled booking

CarinaWolli:
## What does this PR do?

Fixes that rescheduled bookings are not deleted anymore for non seat events. 

Also fixes the number of available seats shown on BookingPage. It was taking the number of available seats of the booking that was rescheduled and not of the new slot the users wants to book. 

## Type of change

## How should this be tested?

- Book event 
- reschedule event 
- check and make sure rescheduled event is not in the list of upcoming bookings anymore


emrysal (approved):
Works for me 🥳

emrysal (changes requested):
But I defer to the tests 🙏  - some fixes needed

apps/web/playwright/reschedule.e2e.ts
```diff
@@ -97,9 +97,6 @@ test.describe("Reschedule Tests", async () => {
 
     await expect(page.locator("[data-testid=success-page]")).toBeVisible();
 
-    // NOTE: remove if old booking should not be deleted
-    expect(await (await booking.self()).status).toBe(BookingStatus.CANCELLED);
```
joeauyeung, line 101: Since we are removing the old booking we don't need this line

apps/web/playwright/reschedule.e2e.ts
```diff
@@ -148,7 +145,6 @@ test.describe("Reschedule Tests", async () => {
     });
 
     await expect(page).toHaveURL(/.*payment/);
-    await payment.delete();
```
joeauyeung, line 151: Since the old booking is deleted, the old payment is also deleted. A new payment record is created with the new booking

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -584,23 +585,27 @@ const BookingPage = ({
                     <div className="text-bookinghighlight flex items-start text-sm">
                       <FiUser
                         className={`ml-[2px] mt-[2px] inline-block h-4 w-4 ltr:mr-[10px] rtl:ml-[10px] ${
-                          booking && booking.attendees.length / eventType.seatsPerTimeSlot >= 0.5
+                          currentSlotBooking &&
```
joeauyeung, line 588: Why are we introducing `currentSlotBooking`? Wouldn't we have the attendee information already from the `booking` prop?

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -584,23 +585,27 @@ const BookingPage = ({
...
+                          currentSlotBooking.attendees.length / eventType.seatsPerTimeSlot >= 0.5
```
joeauyeung, line 589: Why are we introducing `currentSlotBooking`? Wouldn't we have the attendee information already from the `booking` prop?

joeauyeung (commented):
@CarinaWolli, thanks for finding the missing code. I adjusted the tests to the new behaviour and just have a question.

apps/web/components/booking/pages/BookingPage.tsx
CarinaWolli: In the case of rescheduling a booking with seats, the `booking` prop is the booking that should be rescheduled. But we also need the booking that might already exist for the new slot the user want to book (to show how many open slots this new booking has)

In [this line](https://github.com/calcom/cal.com/pull/7910/files#diff-201b6fc0fd34f1a79a852b4f445f1bae9b7ba9503e01b8826411e2e0bdeb7a5cR243) you see, if we have a reschedule uid the booking is the rescheduled booking. If it is not a reschedule then the `booking` prop will be the booking that already exists in the picked slot.

apps/web/playwright/reschedule.e2e.ts
```diff
@@ -97,11 +97,12 @@ test.describe("Reschedule Tests", async () => {
...
     await expect(page.locator("[data-testid=success-page]")).toBeVisible();
 
-    // NOTE: remove if old booking should not be deleted
-    expect(await (await booking.self()).status).toBe(BookingStatus.CANCELLED);
-
     const newBooking = await prisma.booking.findFirst({ where: { fromReschedule: booking.uid } });
+    const rescheduledBooking = await prisma.booking.findFirst({ where: { uid: booking.uid } });
+
     expect(newBooking).not.toBeNull();
+    expect(rescheduledBooking).toBeNull();
```
CarinaWolli, line 104: @emrysal Test for my fix. Fails before and succeeds now

apps/web/playwright/onboarding.e2e.ts
```diff
@@ -26,7 +26,11 @@ test.describe("Onboarding", () => {
         await page.locator("input[name=username]").fill("new user onboarding");
         await page.locator("input[name=name]").fill("new user 2");
         await page.locator("input[role=combobox]").click();
-        await page.locator("text=Europe/London GMT +0:00").click();
```
CarinaWolli, line 29: London is now in timezone GMT +1 because of dst

emrysal (approved):
Approved.

I also added a follow-up ticket on the back of this to (maybe) change the delete behaviour.

https://github.com/calcom/cal.com/issues/7946

PR #7915, opened 2023-03-23
fix: use the correct timezone of the user in booking detail page

G3root:
## What does this PR do?

context:

i'm in `Asia/Colombo` timezone. when i view the detail of my upcoming booking, it shows the time in the time zone of the organizer. this PR aims to fix that
Screenshot 2023-03-24 at 01-56-06 Your recurring meeting has been confirmed Cal com(1)

## Type of change


apps/web/pages/booking/[uid].tsx
```diff
@@ -1017,6 +1017,14 @@ const handleSeatsEventTypeOnBooking = (
 
 export async function getServerSideProps(context: GetServerSidePropsContext) {
   const ssr = await ssrInit(context);
+  const session = await getServerSession(context);
```
Comment (G3root) [inline apps/web/pages/booking/[uid].tsx:1020]: if a user is authenticated return his timezone

apps/web/pages/booking/[uid].tsx
```diff
@@ -197,12 +199,10 @@ export default function Success(props: SuccessProps) {
...
     email,
     seatReferenceUid,
+    tz: timeZoneQuery,
   } = querySchema.parse(router.query);
 
-  const tz =
-    (isSuccessBookingPage
-      ? props.bookingInfo.attendees.find((attendee) => attendee.email === email)?.timeZone
-      : props.bookingInfo.eventType?.timeZone || props.bookingInfo.user?.timeZone) || timeZone();
+  const tz = timeZoneQuery ? timeZoneQuery : props.tz ? props.tz : timeZone();
```
Comment (G3root) [inline apps/web/pages/booking/[uid].tsx:205]: if `tz` query param present use that or if user is authenticated user that or fallback to guessing his timezone

packages/app-store/stripepayment/lib/client/createPaymentLink.ts
```diff
@@ -9,11 +9,12 @@ export function createPaymentLink(opts: {
...
   email?: Maybe<string>;
+  tz?: Maybe<string>;
   absolute?: boolean;
 }): string {
-  const { paymentUid, name, email, date, absolute = true } = opts;
+  const { paymentUid, name, email, date, tz, absolute = true } = opts;
   let link = "";
   if (absolute) link = WEBSITE_URL;
-  const query = stringify({ date, name, email });
+  const query = stringify({ date, name, email, tz });
```
CarinaWolli, line 18: why do we need tz here?

G3root: false assumption here. pushed the change removing this

apps/web/pages/booking/[uid].tsx
```diff
@@ -197,12 +199,10 @@ export default function Success(props: SuccessProps) {
     formerTime,
     email,
     seatReferenceUid,
+    tz: timeZoneQuery,
   } = querySchema.parse(router.query);
 
-  const tz =
-    (isSuccessBookingPage
-      ? props.bookingInfo.attendees.find((attendee) => attendee.email === email)?.timeZone
```
Comment (CarinaWolli) [inline apps/web/pages/booking/[uid].tsx:204]: Can't we leave it like that to show the attendee's timezone when `isSuccessBookingPage` is true, instead of passing it through the query?

If `isSuccessBookingPage` is false then we show the tz of the logged in user (as you did it)

Comment (G3root) [reply apps/web/pages/booking/[uid].tsx:204]: added tz to the query for a edges case , if a authenticated user booked in a different timezone rather than his default one

emrysal (approved):
We should look at storing more details into the session so we can do less /me calls.

PR #7922, opened 2023-03-23
insights follow up

alannnc:
- [x] headline styling
- [x] dateRange styling 
- [x] longer input for team name

TODO: 
- [ ] skeleton when loading
- [ ] ability to see "My insights" (not only team insights) that are the personal links, not team links


Jaibles:
Great work @alannnc!

Small UI issues (I'll keep adding to this list)...
- [ ] Make the border around each chart 200 not 300. (We use 300 for inputs, buttons & interactive things, not content containers.)
- [ ] Make the page title just "Insights" not "Insights for...". The latter kinda implies you can't change it.
- [ ] We need to improve the clear filter UI and the fact that it jumps left. I'll provide designs for this as there's too many things we need to fix to list. This isn't Alan's doing, the end result has different filters than design so just need to account for that.
- [ ] Put an `mb-8` below the title to separate the charts instead of `mb-4`
- [ ] There's an `mt-4` on the chart separating it from the title. We can remove it..

If anything is not clear, just let me know and I'll share a loom.

packages/features/insights/components/Card.tsx
```diff
@@ -0,0 +1,18 @@
+import { Card } from "@tremor/react";
+
+interface ICardProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
+const CardInsights = (props: ICardProps) => {
```
zomars, line 8:
```suggestion
export const CardInsights = (props: ICardProps) => {
```

packages/features/insights/components/Card.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const { children, className = "", ...rest } = props;
+
+  return (
+    <Card className={`shadow-none ring-gray-300 ${className}`} {...rest}>
+      {children}
+    </Card>
+  );
+};
+
+export { CardInsights };
```
zomars, line 18:
```suggestion
```

zomars (changes requested):

zomars (approved):

packages/features/insights/filters/DateSelect.tsx
```diff
@@ -15,44 +16,49 @@ export const DateSelect = () => {
...
-    />
+          // If range has value and it's of type RangeType
+        }}
+        options={undefined}
+        enableDropdown={true}
+        placeholder={t("select_date_range")}
+        enableYearPagination={true}
+        minDate={currentDate.subtract(2, "year").toDate()}
+        maxDate={currentDate.toDate()}
+        color="black"
```
zomars, line 59:
```
Type '"black"' is not assignable to type '"slate" | "gray" | "zinc" | "neutral" | "stone" | "red" | "orange" | "amber" | "yellow" | "lime" | "green" | "emerald" | "teal" | "cyan" | "sky" | "blue" | "indigo" | "violet" | "purple" | "fuchsia" | "pink" | "rose" | undefined'.ts(2322)
```

PR #7928, opened 2023-03-24
Fixes formatted description in email + sanitize html everywhere

CarinaWolli:
## What does this PR do?

- Adds formatting to event-type description in email:
- Sanitize all description and bio set with `dangerouslySetInnerHTML` with [sanitize-html]( https://github.com/apostrophecms/sanitize-html)


packages/prisma/index.ts
```diff
@@ -21,7 +22,6 @@ if (process.env.NODE_ENV !== "production") {
 }
 // If any changed on middleware server restart is required
 bookingReferenceMiddleware(prisma);
-eventTypeDescriptionParseAndSanitizeMiddleware(prisma);
```
CarinaWolli, line 24: @zomars & @emrysal I removed that middleware now as `descriptionAsSafeHTML` is coming from all parts of my code already. Any objections to this?

CarinaWolli: @zomars what do you think about removing the middleware for sanitizing the event type description?

emrysal (approved):
Looks good to me, not a change we can't come back from and it works.

PR #7929, opened 2023-03-24
fix(booking-page): prefetch slots query to avoid waterfalls

### LINKED ISSUE #4836: [CAL-1660] event booking page loads availability slowly without any spinner
# Issue Summary

Loading a public availability page shows all the days greyed out for several seconds:

image

After waiting a few seconds, it loads properly:

image

I don't remember ever seeing this until recently, so it feels like there has been some kind of regression here.

# Proposed solutions

IMHO there are three aspects to this problem, each of which can be tackled separately:

## 1. Reduce the loading time as much as possible

I don't know if there are architectural limitations causing this delay, but if there is room for optimisation then obviously that would be a very positive step forwards.

## 2. Show a spinner icon or some other visual indication that the page is loading

Regardless of whether the delay is somewhat unavoidable or can be optimised away, there should certainly be something in the UI which makes it clear to the user that the page isn't broken but it's still loading and that they should wait patiently for the availability to appear.  Currently there is no such indication, so users are liable to think it's hung or the site's crashed/broken or similar, and in the worst case they will just give up, close their browser tab, and go away without a successful booking.

## 3. Show calendar days on first load

Before the availability has loaded, it should immediately display the numbers of the days of the month rather than blank grey squares.  At least then while the user is waiting for it to load, they can potentially still use those dates and that calendar month layout to start mentally figuring out their own availability.

# Steps to Reproduce

1. Go to https://cal.com/aspiers/30-mins or probably any other booking page
2. Note that it initially appears like the above screenshot.
3. Wait a few seconds
4. Note that 

# Technical details

Saw this on both Chrome mobile on Android 11, and desktop Chrome on Linux, which are effectively totally different browsers.  This seems to strongly indicate that the problem is not specific to browser versions and more likely server-side.

<sub>[CAL-1660](https://linear.app/calcom/issue/CAL-1660/event-booking-page-loads-availability-slowly-without-any-spinner)</sub>

juliusmarminge:
## What does this PR do?

Fixes #4836

The above view is for a single booking page - where prefetch has been applied. The bottom one is the dynamic booking thing (+ in the url to select multiple users) which still isn't prefetched

https://user-images.githubusercontent.com/51714798/227522241-4f6278b6-6fd7-42d5-8e5e-fea7cc8b5b2f.mp4

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/packages/prisma/schema.prisma
+++ b/packages/prisma/schema.prisma
@@ -204,6 +204,8 @@ model User {
   verifiedNumbers      VerifiedNumber[]
   hosts                Host[]
 
+  @@index([username])
+  @@index([allowDynamicBooking])
   @@map(name: "users")
 }
 
@@ -329,6 +331,9 @@ model Booking {
   seatsReferences       BookingSeat[]
   /// @zod.custom(imports.bookingMetadataSchema)
   metadata              Json?
+
+  @@index([startTime])
+  @@index([endTime])
 }
 
 model Schedule {
@@ -358,6 +363,8 @@ model Availability {
 
   @@index([eventTypeId])
   @@index([scheduleId])
+  @@index([startTime])
+  @@index([endTime])
 }
 
 model SelectedCalendar {
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -136,6 +136,7 @@
     "@calcom/types": "*",
     "@microsoft/microsoft-graph-types-beta": "0.15.0-preview",
     "@playwright/test": "^1.31.2",
+    "@tanstack/react-query-devtools": "^4.28.0",
     "@testing-library/react": "^13.3.0",
     "@types/accept-language-parser": "1.5.2",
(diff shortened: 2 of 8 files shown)


packages/prisma/schema.prisma
```diff
@@ -204,6 +204,8 @@ model User {
   verifiedNumbers      VerifiedNumber[]
   hosts                Host[]
 
+  @@index([username])
+  @@index([allowDynamicBooking])
```
KATT, line 208: is this needed?

juliusmarminge: Some queries are filtering based on these fields so indexes should speed that up? I'll be honest though I'm not very familiar with db optimizations with indexes though

emrysal: Let's leave these out, we should add indexes but we should also evaluate our query optimisation before we do so.

PR #7931, opened 2023-03-24
Fix Zod errors in Outlook auth function

joeauyeung:
## What does this PR do?

This PR fixes Zod parsing errors in the Outlook auth function. By using safeParse we will not throw an error when the response from MS is not as expected. We also added increased logging of MS error responses to better handle future errors.

Fixes # (issue)

https://www.loom.com/share/3ca4e240e5864151854c50ed88431cc3

## Type of change

## How should this be tested?

- Connect an Outlook account
- In the DB change the expiry date to force a refresh
- Refresh the listed calendars page
    - There should be a new token
- In the DB remove a property so that the zod parse will fail
- Refresh the listed calendars page
    - There should be a new token
- Create a fake response in the Outlook auth function. The error log should contain the zod error & the response

## Checklist

### DIFF
--- a/packages/app-store/office365calendar/lib/CalendarService.ts
+++ b/packages/app-store/office365calendar/lib/CalendarService.ts
@@ -219,7 +219,7 @@ export default class Office365CalendarService implements Calendar {
     };
     const o365AuthCredentials = credential.key as O365AuthCredentials;
 
-    const refreshAccessToken = async (refreshToken: string) => {
+    const refreshAccessToken = async (o365AuthCredentials: O365AuthCredentials) => {
       const { client_id, client_secret } = await getOfficeAppKeys();
       const response = await fetch("https://login.microsoftonline.com/common/oauth2/v2.0/token", {
         method: "POST",
@@ -227,12 +227,17 @@ export default class Office365CalendarService implements Calendar {
         body: new URLSearchParams({
           scope: "User.Read Calendars.Read Calendars.ReadWrite",
           client_id,
-          refresh_token: refreshToken,
+          refresh_token: o365AuthCredentials.refresh_token,
           grant_type: "refresh_token",
           client_secret,
         }),
       });
-      const o365AuthCredentials = refreshTokenResponseSchema.parse(await handleErrorsJson(response));
+      const responseJson = await handleErrorsJson(response);
+      const tokenResponse = refreshTokenResponseSchema.safeParse(responseJson);
+      o365AuthCredentials = { ...o365AuthCredentials, ...(tokenResponse.success && tokenResponse.data) };
+      if (!tokenResponse.success) {
+        console.error("zodError:", tokenResponse.error, "MS response:", responseJson);
+      }
       await prisma.credential.update({
         where: {
           id: credential.id,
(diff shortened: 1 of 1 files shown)


packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -219,20 +219,25 @@ export default class Office365CalendarService implements Calendar {
...
           client_id,
-          refresh_token: refreshToken,
+          refresh_token: o365AuthCredentials.refresh_token,
           grant_type: "refresh_token",
           client_secret,
         }),
       });
-      const o365AuthCredentials = refreshTokenResponseSchema.parse(await handleErrorsJson(response));
+      const responseJson = await handleErrorsJson(response);
+      const tokenResponse = refreshTokenResponseSchema.safeParse(responseJson);
```
joeauyeung, line 237: Change to safeParse so we don't throw an error if MS sends a response we aren't expecting

packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -219,20 +219,25 @@ export default class Office365CalendarService implements Calendar {
...
+      o365AuthCredentials = { ...o365AuthCredentials, ...(tokenResponse.success && tokenResponse.data) };
```
joeauyeung, line 238: If the responses is what we were expecting, update the new keys. In the spread operator, if keys are present in both objects, the second object's keys will be written

packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -219,20 +219,25 @@ export default class Office365CalendarService implements Calendar {
...
+      if (!tokenResponse.success) {
+        console.error("zodError:", tokenResponse.error, "MS response:", responseJson);
+      }
```
joeauyeung: Capture the response on Axiom so we can better diagnose the problem

packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -246,9 +257,10 @@ export default class Office365CalendarService implements Calendar {
 
     return {
       getToken: () =>
+        refreshTokenResponseSchema.safeParse(o365AuthCredentials).success &&
         !isExpired(o365AuthCredentials.expires_in)
```
joeauyeung, line 261: Refresh the tokens if the keys object is not what we expected or the access token is expired

packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -219,20 +219,31 @@ export default class Office365CalendarService implements Calendar {
...
           client_id,
-          refresh_token: refreshToken,
+          refresh_token: o365AuthCredentials.refresh_token,
           grant_type: "refresh_token",
           client_secret,
         }),
       });
-      const o365AuthCredentials = refreshTokenResponseSchema.parse(await handleErrorsJson(response));
+      // const responseJson = await handleErrorsJson(response);
+      const responseJson = { error: "This is an error" };
```
emrysal, line 236: I don't think I'm ready to approve this just yet 😄

emrysal (changes requested):
Other than that looking good 👍

packages/app-store/office365calendar/lib/CalendarService.ts
joeauyeung: Haha good catch

emrysal (approved):

PR #7933, opened 2023-03-24
CALCOM-7895 - [CAL-1320] remove "application_fee_amount" for all regions

### LINKED ISSUE #7895: [CAL-1320] remove "application_fee_amount" for some regions (India, Mexico, Brazil, ..)
Stripe connect can't charge an application fee for:

Brazil, India, Mexico, Malaysia, Singapore, Thailand

we need to exclude those regions as it makes the Stripe App unusable for customers from those regions.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1320](https://linear.app/calcom/issue/CAL-1320/remove-application-fee-amount-for-some-regions-india-mexico-brazil)</sub>

gitstart-calcom:
Fixes #7895


packages/app-store/stripepayment/lib/PaymentService.ts
```diff
@@ -61,7 +61,6 @@ export class PaymentService implements IAbstractPaymentService {
         amount: payment.amount,
         currency: this.credentials.default_currency,
         payment_method_types: ["card"],
-        application_fee_amount: paymentFee,
       };
```
hariombalhara, line 64: Verified that collected fee doesn't show up for me with this branch. It is collected in main
Screenshot 2023-04-11 at 9 21 12 PM

hariombalhara: Screenshot 2023-04-11 at 9 24 04 PM
Should we update it here as well ? cc @PeerRich I am not sure how long we would remain without fee, so it should reflect that ideally.

hariombalhara (changes requested):
Waiting on confirmation from @PeerRich. Otherwise it's good to merge. Blocking till then

packages/app-store/stripepayment/lib/PaymentService.ts
PeerRich: worth keeping, we will readd it later

PeerRich (approved):

hariombalhara (approved):

PR #7934, opened 2023-03-24
Fix for Apple Calendar failing entire availability check because of Shared Calendars

### LINKED ISSUE #7201: Conflict checks break with Apple Calendar

### Issue Summary

Adding an Apple Calendar app causes all conflict checks to break.

### Steps to Reproduce

1. Set up a Google calendar
2. Check that existing events block available slots on cal.com
3. Add an Apple calendar
4. Check again that all slots become available on cal.com

I have a few all-day events in my Apple account, but none of the calendars with all-day events are selected for conflict checks.

#### Browser

Chrome Version 110.0.5481.100 (Official Build) (x86_64)

#### Screen recording

https://user-images.githubusercontent.com/42545742/219881388-06a33866-5158-4a38-a6ff-e855f9c1fd97.mov

alishaz-polymath:
## What does this PR do?

Removes bad Broad Spectrum ReplaceAll call from the ICal body sanitizer function that resulted in poor EOL replacement in very specific situations on a shared Apple Calendar (and I imagine a few other calendars).

The Sanitizer function requires a nice proper testing and potentially  revamp, but this should do as a hotfix for now without any major issues.

Fixes: #7201 

## Type of change

### DIFF
--- a/packages/lib/sanitizeCalendarObject.ts
+++ b/packages/lib/sanitizeCalendarObject.ts
@@ -4,7 +4,6 @@ export const sanitizeCalendarObject = (obj: DAVObject) => {
   return obj.data
     .replaceAll("\r\n", "\r")
     .replaceAll("\r", "\r\n")
-    .replaceAll(/(: \r\n|:\r\n|\r\n:|\r\n :)/gm, ":")
     .replaceAll(/(; \r\n|;\r\n|\r\n;|\r\n ;)/gm, ";")
     .replaceAll(/(= \r\n|=\r\n|\r\n=|\r\n =)/gm, "=");
 };


packages/lib/sanitizeCalendarObject.ts
```diff
@@ -4,7 +4,6 @@ export const sanitizeCalendarObject = (obj: DAVObject) => {
   return obj.data
     .replaceAll("\r\n", "\r")
     .replaceAll("\r", "\r\n")
-    .replaceAll(/(: \r\n|:\r\n|\r\n:|\r\n :)/gm, ":")
```
alishaz-polymath, line 7: This caused 
`URL;VALUE=URI:\r\nEND:VEVENT`
to turn into 
`URL;VALUE=URI:END:VEVENT`

This doesn't seem like much but it basically destroys the END of VEVENT by pulling it up instead of in a new line which is essential for the ICAL parser to know that that's the end of the VEVENT. 🤯 

So, 
```
URL;VALUE=URI:END:VEVENT
```
is what it becomes whereas it should have been 
```
URL;VALUE=URI:
END:VEVENT
```

alishaz-polymath (commented):
Self Review

zomars (approved):
You found a needle in the haystack @alishaz-polymath great work! 🙏🏽 

Agreed that we should follow up with unit testing.

PR #7952, opened 2023-03-27
feat: Google Tag Manager (GTM) App

StefanFlorea00:
## What does this PR do?

There is an existing app for Google Analytics 4. We created an app that works in the same way but with Google Tag Manger integration.

## Type of change

## How should this be tested?

-  Install the app
-  Add the app to an event, fill in the Tracking ID with a GTM container ID
-  Ensure that the GTM script is visible on the booking page (in inspect elements)
-  Preview and test if GTM events are fired successfully using the tagmanager

## Checklist

## TODO:
- [ ] copy changes and screenshot review @Jaibles @PeerRich


packages/app-store/google-tag-manager/README.md
hariombalhara, line 1: I am not sure why this file is there. We can remove it. If it came from AppStore Cli, will fix that.

packages/app-store/google-tag-manager/static/1.jpeg
hariombalhara, line 1: We need the correct image here that demoes that the App is working. cc @PeerRich

yarn.lock
hariombalhara, line 1: I am not sure why this file is there. There are no dependencies update. So, we should just uncommit this file.

StefanFlorea00: Removed

hariombalhara (approved):

PR #7970, opened 2023-03-28
feat: soft delete rescheduled booking

### LINKED ISSUE #7946: [CAL-1341] Change reschedule booking flow handling of old event

Currently we are deleting a rescheduled event, this causes the event to 404 when a link is opened that has since been rescheduled. Instead of a hard delete we should set the event as cancelled so a "This event has been cancelled" message can be shown. We may even want to add a new booking status to identity rescheduled events.

<sub>[CAL-1341](https://linear.app/calcom/issue/CAL-1341/change-reschedule-booking-flow-handling-of-old-event)</sub>

G3root:
## What does this PR do?

Fixes #7946. rather than deleting a rescheduled this PR aims to soft delete the rescheduled bookings.

## Type of change


apps/web/public/static/locales/en/common.json
```diff
@@ -262,6 +262,7 @@
   "user_needs_to_confirm_or_reject_booking": "{{user}} still needs to confirm or reject the booking.",
   "user_needs_to_confirm_or_reject_booking_recurring": "{{user}} still needs to confirm or reject each booking of the recurring meeting.",
   "meeting_is_scheduled": "This meeting is scheduled",
+  "meeting_is_rescheduled": "This meeting is rescheduled",
```
emrysal, line 265: Similarly, we don't really need a special message for rescheduled bookings, just show cancel message

CarinaWolli (changes requested):
e2e test is failing: 
Screenshot 2023-03-29 at 07 26 23
let's remove that line and check if the status of the rescheduled booking is CANCELLED

CarinaWolli (approved):
LGTM!

PR #7975, opened 2023-03-28
Fix: AppStore CLI Create command not working after upgrading to yarn 3

hariombalhara:
## What does this PR do?

Fixes
- Failing `yarn create-app` command. Reported [here](https://calendso.slack.com/archives/C04MXT89BNF/p1679917991428659) image
- Also removes stale README.md from template

**Environment**: Production

## Type of change

## How should this be tested?

- [x] Run `yarn create-app` and notice that it was failing earlier and now it passes.

### DIFF
--- a/packages/app-store-cli/src/core.ts
+++ b/packages/app-store-cli/src/core.ts
@@ -1,7 +1,7 @@
 import fs from "fs";
 import path from "path";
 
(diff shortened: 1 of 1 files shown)


emrysal (approved):
👍

PR #7988, opened 2023-03-28
Fix missing edge case where team is paid but needs to be updated

zomars:
## Type of change

## How should this be tested?

- [ ] Make new team with x amount of members
- [ ] Pay it / publish it
- [ ] Try to add/remove members
- [ ] Subscription should update accordingly


emrysal (approved):
Oops.

PR #7999, opened 2023-03-29
Added serverside translations to teams page.

JeroenReumkens:
## What does this PR do?

Adds serverside translations to teams page.

## Type of change


apps/web/pages/teams/index.tsx
```diff
@@ -25,6 +28,14 @@ function Teams() {
...
 }
 
+export const getServerSideProps = async (context: GetServerSidePropsContext) => {
+  return {
+    props: {
+      ...(await serverSideTranslations(context.locale || "en", ["common"])),
+    },
+  };
+};
+
```
PeerRich, line 38: i removed those a while ago to speed things up

packages/app-store/apps.browser.generated.tsx
```diff
@@ -3,6 +3,7 @@
     Don't modify this file manually.
 **/
 import dynamic from "next/dynamic";
+
```
JeroenReumkens, line 6: This is due to new prettier config according to hariom. So a change that should only occur once, then it's on main and shouldn't happen anymore.

apps/web/pages/teams/index.tsx
```diff
@@ -28,10 +27,10 @@ function Teams() {
   );
 }
 
-export const getServerSideProps = async (context: GetServerSidePropsContext) => {
+export const getStaticProps = async () => {
   return {
     props: {
-      ...(await serverSideTranslations(context.locale || "en", ["common"])),
+      ...(await serverSideTranslations("en", ["common"])),
```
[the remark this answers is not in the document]
JeroenReumkens: @Udit-takkar  sorry, missed this comment! This page is static instead of serverside, Peer didn't want to make it serverside because of performance. If I'm correct in that case you can not get the locale from the context, right? Since we don't have the locale as part of our url. So we default to English here, which was okay for Peer.

hariombalhara (approved):

PR #7406, opened 2023-02-25
fix: set organiser email to external id

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/5588, https://github.com/calcom/cal.com/issues/7329

Do suggest if there are better approach to fix these issues

BEFORE:-
Screenshot 2023-02-25 at 3 18 54 PM

AFTER:-
Screenshot 2023-02-25 at 3 19 16 PM

**Environment**: Staging(main branch) / Production

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #8010, opened 2023-03-29
Set GCal & Outlook `iCalUID` in .ics file

joeauyeung:
## What does this PR do?

This PR fixes Gmail parsing our .ics file and adding double events in GCal by adding the iCalUID when creating events to the .ics file we generate.

Fixes # (issue)

## Type of change

## How should this be tested?

- Connect GCal
- Create an event
- From the email get the .ics file
- Send the .ics to your inbox. When clicking "View on Google Calendar" it should open the event that we created
- Reschedule that booking
- The new .ics file should still point to the event that we created

## Checklist


packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -73,6 +73,10 @@ export default class Office365CalendarService implements Calendar {
         body: JSON.stringify(this.translateEvent(event)),
       });
 
+      const responseJson = await handleErrorsJson(response);
+
+      return { ...responseJson, iCalUID: responseJson.iCalUId };
```
joeauyeung, line 78: Notice in Outlook the property has a lower case "d". Through out the call I think we should standardize it to UID as that's the property name in the .ics file

Udit-takkar (changes requested):
This does not work when i have selected secondary calendar as destination.

First i scheduled the event 

Screenshot 2023-03-30 at 6 23 33 PM

then i rescheduled the event to 30th april which created event in both primary and secondary calendar

Screenshot 2023-03-30 at 6 28 47 PM

Udit-takkar:
@joeauyeung i fixed something similar before here https://github.com/calcom/cal.com/pull/7406 . maybe we should do the same thing in updateEvent method in googlecalendar/lib/CalendarService

packages/emails/templates/attendee-scheduled-email.ts
```diff
@@ -37,21 +37,21 @@ export default class AttendeeScheduledEmail extends BaseEmail {
       // ics appends "RRULE:" already, so removing it from RRule generated string
       recurrenceRule = new RRule(this.calEvent.recurringEvent).toString().replace("RRULE:", "");
     }
-    const partstat: ParticipationStatus = "NEEDS-ACTION";
+    const partstat: ParticipationStatus = "ACCEPTED";
     const role: ParticipationRole = "REQ-PARTICIPANT";
     const icsEvent = createEvent({
+      uid: this.calEvent.iCalUID || this.calEvent.uid!,
```
joeauyeung, line 43: We set non-null for `uid` here since every booking should have a `uid` associated with it on Cal

Udit-takkar (approved):
LGTM. working fine

emrysal (approved):
Reverted the serverConfig change, but besides that it looks good to me 🙏 🎉

PR #8021, opened 2023-03-30
fix: disable 2FA switch for non `CAL` auth identity providers

G3root:
## What does this PR do?

after:

Screenshot 2023-03-30 08:10:33

## Type of change


sean-brydon (approved):

apps/web/pages/settings/security/two-factor-auth.tsx
```diff
@@ -32,11 +32,19 @@ const TwoFactorAuthView = () => {
...
   if (isLoading) return <SkeletonLoader />;
 
+  const isCalProvider = user?.identityProvider === "CAL";
   return (
     <>
       <Meta title={t("2fa")} description={t("2fa_description")} />
+      {!isCalProvider && (
+        <Alert
+          severity="neutral"
+          message="Two-Factor authentication can be enabled only for email and password authentication."
```
sean-brydon, line 42: approved but we should really translate this :)

G3root: done :)

apps/web/public/static/locales/en/common.json
```diff
@@ -404,6 +404,7 @@
   "add_time_availability": "Add new time slot",
   "add_an_extra_layer_of_security": "Add an extra layer of security to your account in case your password is stolen.",
   "2fa": "Two-Factor Authentication",
+  "2fa_disabled":"Two-Factor authentication can be enabled only for email and password authentication",
```
PeerRich, line 407:
```suggestion
  "2fa_disabled":"Two-Factor authentication can only be enabled for email and password authentication",
```

apps/web/pages/settings/security/two-factor-auth.tsx
```diff
@@ -32,11 +32,14 @@ const TwoFactorAuthView = () => {
 
   if (isLoading) return <SkeletonLoader />;
 
+  const isCalProvider = user?.identityProvider === "CAL";
```
PeerRich, line 35: will we need this in the future? maybe worth moving into lib

PeerRich (approved):

PR #8026, opened 2023-03-30
Delete canceled workflow emails from scheduled_sends (SendGrid endpoint) 

CarinaWolli:
## What does this PR do?

When a scheduled workflow email is canceled in SendGrid the batch_id is still stored in the SendGrid endpoint `/v3/user/scheduled_sends`. I wrongly assumed that batch_ids were removed automatically when the scheduled date is in the past (I thought I read that somewhere, but I guess I was wrong 🤔) . As there is a limit of entries (which now again was full) we need to delete them ourselves once the scheduled date is in the past.

It also fixes that already verified numbers are shown as verified and don't need to be verified again: 
Screenshot 2023-03-30 at 11 36 29


emrysal (approved):
lgtm :), looks good

Issue #7570, opened 2023-03-07
CalDav NextCloud integration has stopped working


Frikster:
### Issue Summary

CalDav integration has been working fine for over a month. This week I noticed that clients could book on dates where I already had events on my calendar - strangely only events that were "recurring" though they were marked as "busy." Another strange thing I noticed when I went to settings is that the CalDav calendar had disconnected and the Google calendar I had also connected had been set to the default calendar.

I have a phone and desktop calendar app and both sync fine with my CalDav Nextcloud server, so I concluded there was something weird up with the cal.com calendar sync and removed both calendars only to now find I cannot reconnect to my CalDav calendar at all...

 Is this a known issue that is being worked on or handled elsewhere? I might investigate it myself if nobody is already dealing with it.

### Steps to Reproduce

I have _also_ tested these steps with a completely different nextcloud provider and a brand new cal.com account, so it doesn't seem to be an issue specific to either the provider or my account.

1. Connect to a NextCloud CalDav server
2. Get error message below

On the Linux desktop application I see:

![image](https://user-images.githubusercontent.com/10172142/223526263-28a78e0a-4b42-4e97-994c-7d1acd2302ae.png)

On the web browser I see:

![image](https://user-images.githubusercontent.com/10172142/223526344-b20353be-60d6-43f5-b22e-91fd89ebd80b.png)

### Technical details

Whether using [Brave Browser](https://brave.com/) [Version 1.48.158 Chromium: 110.0.5481.77 (Official Build) (64-bit)](https://brave.com/latest/)
Or [LibreWolf](https://librewolf.net/) 109.0.1-2 (Firefox-based) I get the same issue.

Console logs I see in case useful:
![image](https://user-images.githubusercontent.com/10172142/223526658-7c181918-6f55-4cf4-81a8-9c415ba91299.png)

Disconcertingly I also get error messages when I try to remove NextCloud CalDav calendars that failed to get added:
![image](https://user-images.githubusercontent.com/10172142/223535416-6d81d21d-369e-48ac-83a8-fb63706246c0.png)

(also tested and I get this error on a brand new cal.com account using a brand new Nextcloud provider account)

Let me know if any additional info might be useful

alishaz-polymath:
A similar error was fixed in #7557 
Can you confirm if that sorts it out once the release is out?

Frikster:
@alishaz-polymath Will do. Just ping me once it's in prod

PeerRich:
can you try again @Frikster ?

[...]

Issue #5604, opened 2022-11-21
[CAL-417] “e is undefined” error when adding CalDAV


xplosionmind:
### Issue Summary

I see the following error when trying to add a Nextcloud CalDAV url to my [cal.com](http://cal.com) account. **Note**: I already added another CalDAV Nextcloud calendar and everything went smoothly.

![](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/2efd6216-ba6f-42f5-8048-e37a8783cf94/35a8dc6a-e4f9-46ff-a26e-6a7f7a9f1dc8)

![](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/0dd4c500-9009-42e6-a94a-56df0d347928/f663aa4b-ab58-4b0b-8e3b-33ea76575a38)

### Steps to Reproduce

1. Click on “add CalDAV” calendar in `https://app.cal.com/apps/installed/calendar?hl=caldav-calendar`
2. Trying to add two different URLs: `https://nuvola.scambi.org/apps/calendar/p/zogAXYXHjYHwksDp` (public calendar link) and `https://nuvola.scambi.org/remote.php/dav/principals/users/tommi/`
3. The error I shared pops up

---

Related to #3457

> From [SyncLinear.com](https://synclinear.com)

alishaz-polymath:
Thanks for the report. I'll add it to the list 🙏

larseggert:
Same issue here against a https://sabre.io/baikal/ server

alishaz-polymath:
Hey @larseggert could you please provide me with steps to reproduce the error?
Thanks

alishaz-polymath:
Hey @xplosionmind 
Could you please create a test account and share credentials with me so that I can test it locally?
Thanks 🙏

handkerchief333:
I have the same problem with my Nextcloud calendar.

My steps to reproduce:
1. Install a nextcloud instance with the setup-nextcloud.php from the community: 
https://nextcloud.com/install/
DOWNLOAD Server --> COMMUNITY PROJECTS --> Web installer
2. On your installed Nextcloud platform, go to the Calendar: /index.php/apps/calendar/dayGridMonth/now
3. On the bottom left, click on "calendar settings"
4. Click on "copy CalDAV address"
5. This gives you something like this: https://yourownnextcloudurl/remote.php/dav
6. Switch to your cal.com account: /apps/caldav-calendar
7. Click on install app
[...]

Issue #5238, opened 2022-10-26
[CAL-262] Zimbra connection keeps getting revoked


milospuac:
![Screenshot 2022-10-26 at 19.21.10.png](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/4a01dd3b-ac77-4dcb-8388-99c1ab17ed8d/d7826a49-5a41-4fba-801e-b1f0134179e0)

It connects then a few seconds later connection fails claiming revoked permissions.

> From [Linear-GitHub Sync](https://synclinear.com)

PR #8031, opened 2023-03-30
Caldav/refactor: Mutiple Fixes and General code improvement

### LINKED ISSUE #5570: Meetings in Apple calendar are offset when calculating available time slots

The calendar I use is iCloud and I am located in continental Europe (CET). My Calendar is in this timezone on my device. I have confirmed on [icloud.com](http://icloud.com/) that it is the same there. When an event is scheduled via [cal.com](http://cal.com/), it is correctly scheduled at the right time. However, the events in my calendar that are checked for conflicts are not taken into account correctly. They are offset by 1 hour. It's as if an event set in the calendar outside of [cal.com](http://cal.com/) for 10:00 CET is in fact considered to be at 11:00 CET, thus defeating the purpose of the app, since double bookings are possible. Events which are schedules within [cal.com](http://cal.com/), however, are correctly taken into account.

### Steps to Reproduce

1. I opened a new account
2. I added two Apple calendars, with preexisting meetings, and both set to Central European Time (CET)
3. I created an event type limited to 3 days in the future (for an event) and with an availability schedule for 10:00-19:00 every day
4. Then I went to the booking page
5. The unavailable time slots (due to a preexisting meeting) are offset by one hour on the booking page.
6. I can book a meeting during the preexisting meeting, resulting in a collision
7. That meeting appears correctly in the Apple calendar
8. That meeting also results in correct unavailable time slots

Thus, almost everything seems fine, apart from the fact that the preexisting events are incorrectly scheduled when calculating the available or unavailable time slots. The offset is 1 hour, which is the current time difference between my time zone and UTC - that could be the issue, perhaps.

### Technical details

OS : Mac OS and iOS (tried on 2 devices)
Browsers : Chrome 107 and Firefox 107 (tried both)
Calendar timezone : CET

### LINKED ISSUE #7796: Conflict checks stopped working
### Issue Summary

I am using CalDAV with mailbox.org and cal.com is no longer recognizing recurring events from mailbox.org as blocked. So now I need to reschedule a lot of things :/

This worked before!

### Steps to Reproduce

1. Go to https://www.heylogin.com/call/dominik
2. You'll see that you can book Fridays in my calendar, but Fridays are fully blocked with recurring events
3. Individual events (for example 17-19th April) are recognized correctly

image

alishaz-polymath:
### Refactoring the CalDAV and Apple Calendar code to improve overall stability and support for various caldav services

- Handle failure better so that it doesn't stop the conflict checks for other working integrations
- Improves the Sanitization function to  match the colons, semicolons, and equal signs only when they are followed by whitespace and then immediately followed by another colon, semicolon, or equal sign on the next line. This is an improvement on the current sanitizer function making it broader spectrum and avoiding errors that caused Shared Apple Calendar (other's ownership) and a few other CalDAV service providers to fail. In general, this should allow support for more CalDAV service providers.
- We handle the expanded recurring events now, which should block off all the occurrences of the expanded recurring event in the given timeframe
- Some CalDAV service providers fails silently when we use `EXPAND` property in the fetching of the Calendar Objects (eg. Zoho) and they do this in the back-end. Since there is no way to know for sure if `EXPAND` property is supported or not, we first send the request **with** the `EXPAND` property and then if that returns empty (silent failure) we make a call without the `EXPAND` property.
- We assume that any ICAL object returned without any extension is a `.ics` file and treats it as such. This adds support to the Service Providers which send valid ICAL objects without an extension.

Fixes #7796 #7570 #5604 #5238 #8098 #8166 

### To follow up:

- Improve the timezone registration when timezone is missing
- DST compatibility
- Tests when adding a CalDAV integration

Follow ups issues to fix: #5570 

## Type of change

## How should this be tested?

- [x] Connect with a shared Apple Calendar with someone else's ownership. It should now work.
- [x] Connect a failed integration. It shouldn't stop working calendar integrations from conflict checking anymore.
- [x] Tested Support for various CalDAV integrations from Vates (Bluemind) to Qloud MSP (NextCloud) and Zoho. Although, reschedule in Zoho isn't working but that's on their end, and the issue is being checked by their team. 

## Checklist


packages/lib/CalendarService.ts
```diff
@@ -300,133 +303,127 @@ export default abstract class BaseCalendarService implements Calendar {
...
-        standard.addPropertyWithValue("tzoffsetfrom", tzoffsetfrom);
-        standard.addPropertyWithValue("tzoffsetto", tzoffsetto);
-        // provide a standard dtstart
-        standard.addPropertyWithValue("dtstart", "1601-01-01T00:00:00");
-        timezoneComp.addSubcomponent(standard);
-        vcalendar.addSubcomponent(timezoneComp);
+      let vcalendar: ICAL.Component;
+      try {
+        const jcalData = ICAL.parse(sanitizeCalendarObject(object));
+        vcalendar = new ICAL.Component(jcalData);
```
alishaz-polymath, line 319: We place ICAL.parse in a `try-catch` as .parse function throws error when the object is not of iCalendar 2.0 standard, and it ends up breaking the working calendars as well. The `try-catch` fails it without breaking the flow.

packages/lib/CalendarService.ts
```diff
@@ -300,133 +303,127 @@ export default abstract class BaseCalendarService implements Calendar {
...
-      const vtimezone = vcalendar.getFirstSubcomponent("vtimezone");
-
-      if (event.isRecurring()) {
-        let maxIterations = 365;
-        if (["HOURLY", "SECONDLY", "MINUTELY"].includes(event.getRecurrenceTypes())) {
-          console.error(`Won't handle [${event.getRecurrenceTypes()}] recurrence`);
-          return;
+      // const vevent = vcalendar.getFirstSubcomponent("vevent");
+      const vevents = vcalendar.getAllSubcomponents("vevent");
+      vevents.forEach((vevent) => {
```
alishaz-polymath, line 326: A single Calendar Object might contain multiple events when we're talking about expanded recurring events in particular. Placing them in a forEach ensures we don't miss the next events. This fixes the earlier issue where only the first occurrence of a recurring event was being considered for the freebusy check.

alishaz-polymath (commented):
Self Review added

leog (approved):
Code looks good, approving.

PR #8033, opened 2023-03-30
Fix showing seat attendees in emails

joeauyeung:
## What does this PR do?

This PR fixes attendee data being shared on seats events when sending the following emails
- Location change
- Rescheduling
- Canceling

Fixes # (issue)

## Type of change

## How should this be tested?

- Create a seats event
- Booking multiple attendees
- For each of the steps, check that the organizer email gets all the attendees and that each attendee email is only receiving the that specific attendee
    - Reschedule the booking
    - Change the location of the booking
    - Cancel the booking

## Checklist

### DIFF
--- a/packages/emails/email-manager.ts
+++ b/packages/emails/email-manager.ts
@@ -88,7 +88,8 @@ export const sendRescheduledEmails = async (calEvent: CalendarEvent) => {
 
   emailsToSend.push(
     ...calEvent.attendees.map((attendee) => {
-      return sendEmail(() => new AttendeeRescheduledEmail(calEvent, attendee));
+      const clonedCalEvent = cloneDeep(calEvent);
+      return sendEmail(() => new AttendeeRescheduledEmail(clonedCalEvent, attendee));
     })
   );
 
@@ -177,7 +178,8 @@ export const sendCancelledEmails = async (calEvent: CalendarEvent) => {
 
   emailsToSend.push(
     ...calEvent.attendees.map((attendee) => {
-      return sendEmail(() => new AttendeeCancelledEmail(calEvent, attendee));
+      const clonedCalEvent = cloneDeep(calEvent);
+      return sendEmail(() => new AttendeeCancelledEmail(clonedCalEvent, attendee));
     })
   );
 
@@ -254,7 +256,8 @@ export const sendLocationChangeEmails = async (calEvent: CalendarEvent) => {
 
   emailsToSend.push(
     ...calEvent.attendees.map((attendee) => {
-      return sendEmail(() => new AttendeeLocationChangeEmail(calEvent, attendee));
+      const clonedCalEvent = cloneDeep(calEvent);
+      return sendEmail(() => new AttendeeLocationChangeEmail(clonedCalEvent, attendee));
     })
   );
 
--- a/packages/emails/email-manager.ts
+++ b/packages/emails/email-manager.ts
@@ -88,8 +88,7 @@ export const sendRescheduledEmails = async (calEvent: CalendarEvent) => {
 
   emailsToSend.push(
     ...calEvent.attendees.map((attendee) => {
-      const clonedCalEvent = cloneDeep(calEvent);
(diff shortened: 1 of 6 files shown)


packages/emails/email-manager.ts
```diff
@@ -88,7 +88,8 @@ export const sendRescheduledEmails = async (calEvent: CalendarEvent) => {
 
   emailsToSend.push(
     ...calEvent.attendees.map((attendee) => {
-      return sendEmail(() => new AttendeeRescheduledEmail(calEvent, attendee));
+      const clonedCalEvent = cloneDeep(calEvent);
+      return sendEmail(() => new AttendeeRescheduledEmail(clonedCalEvent, attendee));
```
joeauyeung, line 92: Most attendee emails are extended from `AttendeeScheduledEmail` which has logic to filter out other attendee information if the event is a seats event & show attendees is false. Since it makes edits to the event we have to create a new cloned event for every attendee.

What was happening before was for the first attendee, the event attendees would be filtered. When moving on the to second attendee it would receive the same event as the first attendee with the filtered attendees' array.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1010,6 +1010,16 @@ async function handler(
...
+        // Moving forward in this block is the owner making changes to the booking. All attendees should be affected
+        evt.attendees = originalRescheduledBooking.attendees.map((attendee) => {
+          return {
+            name: attendee.name,
+            email: attendee.email,
+            timeZone: attendee.timeZone,
+            language: { translate: tAttendees, locale: attendee.locale ?? "en" },
+          };
+        });
+
```
joeauyeung, line 1022: This is in the block when the organizer reschedules the event. Before this `evt.attendees` only contained a single attendee. This is due to the limitation of our booking page where only one attendee can be entered at a time.

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -672,6 +672,8 @@ export const bookingsRouter = router({
           recurringEvent: parseRecurringEvent(booking.eventType?.recurringEvent),
           location,
           destinationCalendar: booking?.destinationCalendar || booking?.user?.destinationCalendar,
+          seatsPerTimeSlot: booking.eventType?.seatsPerTimeSlot,
+          seatsShowAttendees: booking.eventType?.seatsShowAttendees,
```
joeauyeung, line 676: When changing location, we pass in a `CalendarEvent`. It was missing seats data which would trigger hiding other attendees from each other on seats events.

packages/emails/templates/attendee-scheduled-email.ts
```diff
@@ -16,20 +16,18 @@ export default class AttendeeScheduledEmail extends BaseEmail {
...
-  attendees: Person[];
 
   constructor(calEvent: CalendarEvent, attendee: Person, showAttendees?: boolean | undefined) {
     super();
+    if (!showAttendees && calEvent.seatsPerTimeSlot) {
+      this.calEvent = cloneDeep(calEvent);
+      this.calEvent.attendees = [attendee];
+    } else {
+      this.calEvent = calEvent;
+    }
```
joeauyeung, line 27: @emrysal I implemented your feedback here.

alannnc: I was thinking the same about cloning inside the required function. This looks great now.

alannnc: It's the BaseEmail doing some kind of attendee filtering?

emrysal (approved):
This looks good 🚀

packages/emails/templates/attendee-scheduled-email.ts
joeauyeung: If you're talking about `BaseScheduledEmail` we don't do any filtering on that level because the calendar event being passed into it is already filtered here.

PR #8034, opened 2023-03-30
Fix workflows on seated events

alannnc:
## What does this PR do?

Add workflows to be executed when a booking with seats is executed.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Add a booking with seats, setup any workflow
- [ ] Book that event type
- [ ] Expect workflows to be executed

## Checklist


packages/features/ee/workflows/lib/reminders/reminderScheduler.ts
```diff
@@ -32,7 +42,7 @@ export const scheduleWorkflowReminders = async (
         (workflow.trigger === WorkflowTriggerEvents.RESCHEDULE_EVENT && isRescheduleEvent) ||
         workflow.trigger === WorkflowTriggerEvents.AFTER_EVENT
       ) {
-        workflow.steps.forEach(async (step) => {
+        for (const step of workflow.steps) {
```
alannnc, line 45: Prefer for of instead of forEach async.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1229,24 +1232,26 @@ async function handler(
...
-        await prisma.bookingSeat.update({
-          where: {
-            id: bookingSeat.id,
-          },
-          data: {
-            bookingId: newTimeSlotBooking.id,
-          },
-        }),
-      ]);
+      if (seatAttendee?.id && bookingSeat?.id) {
```
alannnc, line 1235: This has wrong type from main, had to do some conditionals to don't run if seatAttendee is null.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -906,15 +906,15 @@ async function handler(
...
-  > => {
+  const handleSeats = async () => {
+    let resultBooking:
+      | (Partial<Booking> & {
+          appsStatus?: AppsStatus[];
+          seatReferenceUid?: string;
+          paymentUid?: string;
+          message?: string;
+        })
+      | null = null;
```
alannnc, line 917: Changes here are: Add some logic so we only have one return so we can run workflows and everything else in every case.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1856,7 +1877,16 @@ async function handler(
...
   }
 
-  const metadata = videoCallUrl ? { videoCallUrl: getVideoCallUrl(evt) || videoCallUrl } : undefined;
+  const metadata = videoCallUrl
+    ? {
+        videoCallUrl: getVideoCallUrlFromEvent({
+          additionalInformation: evt.additionalInformation,
+          location: evt?.location ?? undefined,
+          // updatedEvent: evt.updatedEvent,
+          videoCallData: evt.videoCallData,
```
alannnc, line 1886: passing down only required props

CarinaWolli (changes requested):

CarinaWolli (approved):
LGTM! Thank you 🙏

PR #8041, opened 2023-03-31
perf: split up routers to separate lambdas

juliusmarminge:
## What does this PR do?

Splits up the all tRPC routers into their own API routes. (Edited by @keithwillcode)
### New bundle sizes

The slots router is massive compared to all the rest endpoints together...

image

### Old bundle size

Seems like the analyzer doesn't really compare apples to apples here, as it states the previous bundle size is tiny. Not sure what's going on there, does it not take into account the size of the appRouter since it's just joining in others?

CleanShot 2023-04-01 at 15 26 09@2x

But, given that the entire appRouter is imported which comes in at 334kB, that should give an idea of the actual size:

CleanShot 2023-04-01 at 15 24 40@2x

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test the entire app basically to ensure no tRPC routes have been broken.


emrysal:
Latest main changes to reduce the appStore size look to have been succesful:

image

packages/trpc/server/routers/viewer.tsx
```diff
@@ -158,7 +158,7 @@ const publicViewerRouter = router({
       }
     }),
   // REVIEW: This router is part of both the public and private viewer router?
-  slots: slotsRouter,
+  // slots: slotsRouter,
   cityTimezones: publicProcedure.query(async () => {
     /**
      * Lazy loads third party dependency to avoid loading 1.5Mb for ALL tRPC procedures.
```
zomars, line 164: Would it make sense to split this specific endpoint to another router? Since it seems like dynamic importing doesn't actually reduce bundle size [according to these findings](https://calcominc.slack.com/archives/C04H3D79LAK/p1681235764477849?thread_ts=1681232021.113919&cid=C04H3D79LAK).

zomars: image

zomars (approved):
Codewise LGTM and testing I couldn't find any issues so far. Left a comment about large dependencies and how should be handle those cases.

Also adding to this, how could we keep this as maintainable as possible and how often should we split trpc routers?

zomars (dismissed):
Blocking to check E2E failures

apps/web/pages/api/trpc/slots/[trpc].ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * This file contains tRPC's HTTP response handler
+ */
+import { z } from "zod";
+
+import { getServerSession } from "@calcom/features/auth/lib/getServerSession";
```
Comment (juliusmarminge) [inline apps/web/pages/api/trpc/slots/[trpc].ts:6]: We can also test if it helps to remove this import. It's quite big from Lukas's findings and the slots router is all publicProcedures, so just a `() => null` would work as a session getter for this endpoint/router.

keithwillcode:
I've gotten this branch up to date with main and have brought over my changes from an experimental branch to separate all tRPC routers into their own API routes.

keithwillcode:
Documenting here for clarity. There was a separate test done in #8551 that verified when accessing the route /api/trpc/viewer.public.i18n under a different API route of /api/trpc/public/i18n, the time difference in cold start load was down to 2 seconds from 15+. When I realized the full path forward was to separate all routers into their own API routes so we avoid a compounding effect of loading lots of routers into 1 core router, I brought those changes back over here to build upon the work @juliusmarminge had started.

The effect this will have is if you look at the network traffic for cold starts on pages in the app, you will now see multiple tRPC calls take upwards of 3 seconds to load but since they run concurrently, the effect is not compounded, which was what resulted in such long load times.

Udit-takkar (approved):
LGTM. tested some core features locally

roae (approved):
Great job!! 🚀 

Left a comment about how we can make sure this runs on separate lambdas. we can do it in  other PR or add it here before we merge if you agree, your choice @keithwillcode

keithwillcode:
The way it’s written now, since the api routes are separated into different
folders, Vercel already creates separate functions for each one. We don’t
have to include this config to ensure those are created distinctly. We can
if we want to change config of course though.

PR #8038, opened 2023-03-31
fix: flash of grey when navigating to workflow page

G3root:
## What does this PR do?

before:

[Screencast 2023-03-31 14:56:04.webm](https://user-images.githubusercontent.com/84864519/229082237-ec852344-88a2-45f1-bea6-0873ee4c0165.webm)

after:

[Screencast 2023-03-31 14:54:34.webm](https://user-images.githubusercontent.com/84864519/229082263-2b1eeb01-f99a-4893-a88c-50e6229a7956.webm)

## Type of change

- Bug fix (non-breaking change which fixes an issue)

PR #8058, opened 2023-04-03
Fixes duplicate New button in /workflows

CarinaWolli:
## What does this PR do?

Fixes duplicate new button and not needed filter in /workflows. It was introduced by #8038

Before: 
Screenshot 2023-04-03 at 13 30 32

After: 
Screenshot 2023-04-03 at 13 29 51


emrysal (approved):
🚀

PR #8078, opened 2023-04-04
fix: show upgrade tip for trial users in insights

G3root:
## What does this PR do?

fixes some unauthorized rpc calls

Screenshot 2023-04-04 15:14:12

## Type of change


packages/features/tips/UpgradeTip.tsx
```diff
@@ -32,11 +33,14 @@ export function UpgradeTip({
 }) {
   const { t } = useLocale();
   const { isLoading, hasTeamPlan } = useHasTeamPlan();
+  const { data, isLoading: upgradeAbleLoading } = trpc.viewer.teams.getUpgradeable.useQuery();
 
-  if (hasTeamPlan) return children;
+  const loading = isParentLoading || isLoading || upgradeAbleLoading;
+  const shouldRenderChildren = hasTeamPlan && data && !data.length;
```
hariombalhara, line 39: This is a change that would affect upgrade tip everywhere and not just insights.
It's not clear to me what exact case we are trying to fix here.

When exactly does `getUpgradeable` return data with 0 items. I can see that it checks for the presence of subscriptionId. If we don't have subscriptionId set there, can we assume that it's a trial only case? or it can happen in some other scenario as well. Also, I thought we don't have free trial now.

hariombalhara (changes requested):
Need some clarification, blocking merge til then

PR #8079, opened 2023-04-04
Wrap useSuspense with NoSSR

emrysal:
## What does this PR do?

Implements https://github.com/maapteh/react-no-ssr/blob/master/src/index.tsx


packages/trpc/react/trpc.ts
```diff
@@ -23,7 +23,7 @@ export const trpc = createTRPCNext<AppRouter, NextPageContext, "ExperimentalSusp
         ? "/api/trpc"
         : process.env.VERCEL_URL
         ? `https://${process.env.VERCEL_URL}/api/trpc`
-        : `http://${process.env.NEXT_PUBLIC_WEBAPP_URL}/api/trpc`;
+        : `${process.env.NEXT_PUBLIC_WEBAPP_URL}/api/trpc`;
```
zomars, line 26: Is this still needed?

emrysal: Good question, it was broken (always) but we never noticed it, makes me wonder if we actually ever use trpc serverside.

keithwillcode (approved):

PR #8080, opened 2023-04-04
Bug/multiple duration reschedule

sean-brydon:
Fixes the bug where if you are rescheduling an event with multiple durations that time slot isnt included in the getBusyTime results brought back 

Preventing you from changing your time within any of the overlapping gap. (Hope that makes sense)

More info: https://threads.com/34459850903?s=u6bnwb7f5F6SkLbVgSeXVC

### DIFF
--- a/packages/core/getBusyTimes.ts
+++ b/packages/core/getBusyTimes.ts
@@ -62,7 +62,7 @@ export async function getBusyTimes(params: {
    *  logic within getSchedule.test.ts:addBookings
    */
   performance.mark("prismaBookingGetStart");
-  console.log({ rescheduleBookingId });
+
   const sharedQuery = {
     startTime: { gte: new Date(startTime) },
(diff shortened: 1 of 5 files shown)


packages/core/getBusyTimes.ts
```diff
@@ -67,6 +69,9 @@ export async function getBusyTimes(params: {
     status: {
       in: [BookingStatus.ACCEPTED],
     },
+    uid: {
+      not: rescheduleBookingId, // We dont want to fetch the booking that is being rescheduled - This allows us to reschedule for the timeslot that the existing booking takes up
```
CarinaWolli, line 73: I am getting an error when I am trying to book the same slot again, `rescheduleBookingId` is undefined in this case
Screenshot 2023-04-07 at 09 47 42

CarinaWolli (changes requested):

PR #8092, opened 2023-04-05
Prevent updating app to enabled on seed app store

joeauyeung:
## What does this PR do?

This PR prevents enabling apps on the seed event if they are disabled

Fixes # (issue)

## Type of change

## How should this be tested?

- As an admin disable some apps
- Either run `yarn dx` or seed the app store. The apps should stay disabled

## Checklist


hariombalhara (approved):
I was thinking today itself why do we let the seed automatically enable an app 😅

PR #8108, opened 2023-04-05
Fix/Keep themes in peace across embed and booking pages and App

### LINKED ISSUE #8068: [CAL-1456] Booking area flickring after embedding
The booking area is flickering badly after embedding I tested it on multiple websites and the result is the same.

Video link: https://www.loom.com/share/68a7b6ad672344228df18134d2b31a03

<sub>[CAL-1456](https://linear.app/calcom/issue/CAL-1456/booking-area-flickring-after-embedding)</sub>

hariombalhara:
## What does this PR do?

Fixes #8068 
- Ensures, different embeds with different namespace can have different themes simultaneously.
- Ensure, App can be in one theme for an organizer and  his Booking Pages can be in another theme. This is required because App uses System Theme only and Booking Pages use theme configured in Appearance.
- Cleans up unnecessary use of useTheme and documents it better so that it's users would know when it should be used.

**Environment**:  Production

## Type of change

## How should this be tested?

- [x] [When someone goes from the login page to the dashboard without a page refresh, the theme should change automatically.](https://www.loom.com/share/b04080b601f2418d8e9355868b23cd18)
- [x] [System Preference is set to Dark and the booking page has **Appearance** set to **Light** and the user switches tab b/w booking page and App.](https://www.loom.com/share/26b90dd30b6b4d1bb1590dc7c32a43a5)
- [x] [System Preference is set to Dark and the booking page has **Appearance** set to **Dark** and the system theme is toggled](https://www.loom.com/share/75b9cccbc7ec469b806cad65d6e65448)
- [x] [System Preference is set to Dark and the booking page has **Appearance** set to **System Theme** and the user toggles system theme](https://www.loom.com/share/406b8da0b2cc4711975366e26ca39efe)
- [x] [Embeds can simultaneously work with Light, Dark and Auto(System Preference) theme](https://www.loom.com/share/bf7d95797b1646e0965df812bbe5452d)
## Checklist


apps/web/lib/app-providers.tsx
```diff
@@ -61,36 +62,58 @@ const CustomI18nextProvider = (props: AppPropsWithChildren) => {
...
     }
-    return props.isThemeSupported;
+
+    return props.isBookingPage;
   })();
 
-  const forcedTheme = !isThemeSupported ? "light" : undefined;
+  const themeSupport = isBookingPage
+    ? ThemeSupport.Booking
+    : props.isThemeSupported === false
```
hariombalhara, line 92: if isThemeSupport is explicitly false, then theme would be none otherwise we consider it App Theme

apps/web/lib/app-providers.tsx
```diff
@@ -61,36 +62,58 @@ const CustomI18nextProvider = (props: AppPropsWithChildren) => {
...
+    ? ThemeSupport.None
+    : ThemeSupport.App;
+
+  const forcedTheme = themeSupport === ThemeSupport.None ? "light" : undefined;
```
hariombalhara, line 96: If theme is none, we force light mode regardless of what's stored in storage.

apps/web/lib/app-providers.tsx
```diff
@@ -61,36 +62,58 @@ const CustomI18nextProvider = (props: AppPropsWithChildren) => {
...
+  const storageKey =
+    typeof embedNamespace === "string"
+      ? `embed-theme-${embedNamespace}`
+      : themeSupport === ThemeSupport.App
+      ? "app-theme"
+      : themeSupport === ThemeSupport.Booking
+      ? "booking-theme"
+      : undefined;
+
+  console.log("storageKey", storageKey);
```
hariombalhara, line 110:
```suggestion
```

hariombalhara (commented):
Self review done

apps/web/components/booking/CancelBooking.tsx
```diff
@@ -34,7 +33,6 @@ export default function CancelBooking(props: Props) {
   const [loading, setLoading] = useState(false);
   const telemetry = useTelemetry();
   const [error, setError] = useState<string | null>(booking ? null : t("booking_already_cancelled"));
-  useTheme(props.theme);
```
hariombalhara, line 37: No need to use `useTheme` here. It's used by booking/[uid] and that has `useTheme` call.

apps/web/pages/[user].tsx
```diff
@@ -196,6 +196,8 @@ export default function User(props: inferSSRProps<typeof getServerSideProps> & E
   );
 }
 
+User.isBookingPage = true;
```
Comment (hariombalhara) [inline apps/web/pages/[user].tsx:199]: This has to be set on all booking pages(including routing forms)

apps/web/styles/globals.css
```diff
@@ -44,8 +44,6 @@
   --cal-brand-text: white;
 }
 .dark {
-  color-scheme: dark;
```
hariombalhara, line 47: @sean-brydon I removed this in the PR. Though it should go live as a separate hotfix once you confirm

sean-brydon (approved):
LGTM - One question will this change anything like login/onboarding/password reset where we always force light? 

Guessing not since its controlled by isThemeSupported still?

hariombalhara:
> isThemeSupported

Yeah isThemeSupported is explicitly false there. So, themeSupport would be none
Screenshot 2023-04-17 at 4 57 08 PM

PR #8135, opened 2023-04-07
fix: limit avatar upload size

### LINKED ISSUE #8063: [CAL-1397] avatar upload: if file too big, it fails silently without updating
instead throw a toast maybe?

<sub>[CAL-1397](https://linear.app/calcom/issue/CAL-1397/avatar-upload-if-file-too-big-it-fails-silently-without-updating)</sub>

G3root:
## What does this PR do?

Fixes #8063

after:

[Screencast 2023-04-07 12:20:25.webm](https://user-images.githubusercontent.com/84864519/230557369-6e792a4a-902b-4d66-b9cc-2a8c9e9b0514.webm)

 

## Type of change


Udit-takkar (dismissed):
Code works fine. I think we should also add this condition to updateProfile route also because you can still make a request to that route with avatar bigger than 5MB and it won't throw any error.

G3root:
> Code works fine. I think we should also add this condition to updateProfile route also because you can still make a request to that route with avatar bigger than 5MB and it won't throw any error.

it's hard to validate in the server as we are uploading the image as base64 . i think it's good to go as we are only exposing this api through tRPC

hariombalhara (approved):

hariombalhara:
@G3root Needs conflict resolution

PeerRich (approved):

zomars:
A nice follow up could be to generate a compressed version before uploading.

https://img.ly/blog/how-to-compress-an-image-before-uploading-it-in-javascript/

PR #8169, opened 2023-04-10
Feat: Support setting custom css variables through embed API.

hariombalhara:
## What does this PR do?

[As requested on slack](https://calendso.slack.com/archives/C050EQFTMAS/p1680870044820539) supports configuring border for embed.

[Demo Embed API](https://www.loom.com/share/5b5836a67d2a441183c683e388bb0615)
```javascript
Cal('ui', {
	cssVarsPerTheme:{
		light: {
			'cal-border-booker':'green', 
			'cal-border-booker-width':'20px'
			//...any other css variable defined in tailwind-preset.js
		},
		dark: {
			'cal-border-booker':'red', 
			'cal-border-booker-width':'5px'
			//...any other css variable defined in tailwind-preset.js
		}
	}
})
```

**Environment**:  Production

## Type of change

## How should this be tested?

- [x] See Demo Loom for the tests.

## Checklist

## Followup
- [  ] Make codesandbox examples for few ways css vars can be used to customize embed
- [  ] Document all supported css variables. Once those variables are documented and people start using them, we would have to consider backward compatibility before renaming/removing them cc @sean-brydon


apps/web/pages/_document.tsx
```diff
@@ -43,6 +43,19 @@ class MyDocument extends Document<Props> {
           <meta name="msapplication-TileColor" content="#ff0000" />
           <meta name="theme-color" content="var(--cal-bg)" />
           <Script src="/embed-init-iframe.js" strategy="beforeInteractive" />
+          <style
+            dangerouslySetInnerHTML={{
```
hariombalhara, line 47: Only for testing, to be supported through embed API.

packages/config/tailwind-preset.js
```diff
@@ -58,8 +58,9 @@ module.exports = {
       borderColor: {
         emphasis: "var(--cal-border-emphasis, #9CA3AF)",
         default: "var(--cal-border, #D1D5DB)",
-        subtle: "var(--cal-border-subtle, #E5E7EB)",
+        subtle: `var(--cal-border-subtle, ${subtleColor})`,
         muted: "var(--cal-border-muted, #F3F4F6)",
+        booker: `var(--cal-border-booker, ${subtleColor})`,
```
hariombalhara, line 63: booker was using subtle so use the same colour here so that by default the border color remains same.

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -521,8 +521,8 @@ const BookingPage = ({
         <div
           className={classNames(
             "main",
-            isBackgroundTransparent ? "" : "dark:bg-darkgray-100 bg-default dark:border",
```
hariombalhara, line 524: dark mode classed needn't be applied after design token changes. It was pending to be fixed in this component.

apps/web/styles/globals.css
```diff
@@ -64,6 +64,7 @@
   --cal-border-emphasis: #575757;
   --cal-border: #444444;
   --cal-border-subtle: #2b2b2b;
+  --cal-border-booker: #2b2b2b;
```
hariombalhara, line 67: @sean-brydon I added this only for dark mode. I believe light mode variables are automatically generated by tailwind and they aren't required?

Also, I found a similar file colors.css which isn't being used probably and has the same variables defined there. I didn't make a change intentionally there.

packages/config/tailwind-preset.js
```diff
@@ -106,6 +107,9 @@ module.exports = {
       boxShadow: {
         dropdown: "0px 2px 6px -1px rgba(0, 0, 0, 0.08)",
       },
+      borderWidth: {
+        "booker-width": "var(--cal-border-booker-width, 1px)",
```
hariombalhara, line 111: Named it booker-width so that it doesn't conflict with border-booker which is a color.

packages/embeds/embed-core/src/embed-iframe.ts
```diff
@@ -279,6 +282,12 @@ const methods = {
     log("Method: ui called", uiConfig);
     const stylesConfig = uiConfig.styles;
 
+    if (stylesConfig) {
```
hariombalhara, line 285: As discussed in thread, I find the CSS variables approach the best approach. On requirement basis, we can keep creating the css variables and use them. It ensures that any supported UI customization is explicitly handled in the code and thus avoids regression.

hariombalhara: https://threads.com/34462876572?s=c3wvA7ob5mRkwkg6je4CTk

packages/embeds/embed-core/playground.ts
```diff
@@ -4,14 +4,7 @@ type A = GlobalCalWithoutNs;
...
   const detail = e.detail;
-  const namespace = detail.namespace || "default";
-
-  if (detail.type === "linkReady") {
-    document.querySelector(`#cal-booking-place-${namespace} .loader`).remove();
-  } else if (detail.type === "linkFailed") {
-    document.querySelector(`#cal-booking-place-${namespace} .loader`).remove();
-  }
-  document.querySelector(`#cal-booking-place-${namespace} .last-action`).innerHTML = JSON.stringify(e.detail);
+  console.log("Event: ", e.type, detail);
```
zomars, line 7: Is this console log intended?

zomars (commented):
Code LGTM will delegate approval to @sean-brydon

packages/embeds/embed-core/playground.ts
hariombalhara: Yeah, it's for the embed playground

apps/web/pages/_app.tsx
```diff
@@ -27,6 +28,18 @@ const calFont = localFont({
...
+  const cssVarsStyle = [];
+  if (cssVarsPerTheme) {
+    for (const [themeName, cssVars] of Object.entries(cssVarsPerTheme)) {
+      cssVarsStyle.push(`.${themeName} {`);
+      for (const [cssVarName, value] of Object.entries(cssVars)) {
+        cssVarsStyle.push(`--${cssVarName}: ${value};`);
+      }
+      cssVarsStyle.push(`}`);
+    }
+  }
```
sean-brydon, line 41: Can we reuse useCalTheme here in anyway ?

packages/ui/styles/useCalcomTheme.tsx

apps/web/pages/booking/[uid].tsx
```diff
@@ -325,7 +325,9 @@ export default function Success(props: SuccessProps) {
               <div
                 className={classNames(
                   "main inline-block transform overflow-hidden rounded-lg border sm:my-8 sm:max-w-xl",
-                  isBackgroundTransparent ? "" : " bg-default dark:bg-muted border-subtle",
+                  isBackgroundTransparent
+                    ? ""
+                    : " bg-default dark:bg-muted border-booker border-booker-width",
```
Comment (sean-brydon) [inline apps/web/pages/booking/[uid].tsx:330]:
```suggestion
                  !isBackgroundTransparent && " bg-default dark:bg-muted border-booker border-booker-width",
```

apps/web/styles/globals.css
sean-brydon: Youre right - however if possible please define them in globals.css too  - Makes them easier to manage i believe.

sean-brydon (changes requested):
Nice! Few NITS that we should figure out before merging i believe

apps/web/pages/_app.tsx
hariombalhara: Nice will check that

hariombalhara: useCalComTheme is doing that in useEffect and needs document to work. I am currently generating style string which works on server as well. So, that won't be usable here.

sean-brydon (approved):
Nice one!

PR #8170, opened 2023-04-10
bugfix/Reading timezone of undefined on getUserAvailability function

roae:
## What does this PR do?

Avoid get error (Cannot read properties of undefined (reading 'timeZone')) when a user has no schedule defined in getUserAvailability function. 

## Type of change


roae (commented):
Self reviewed

zomars (approved):

PR #8189, opened 2023-04-11
Add Plausible (Custom) App

hybes:
## What does this PR do?

Add's support self-hosted instances of Plausible, where the URL of the script.js file may be configured by the user.
Adjusted configuration of BookingPageTagManager to also val.replace {PLAUSIBLE_URL} with the configured value.
Tested locally and the script is set correctly in the final site HTML.

Fixes # (issue)

No issue related

## Type of change

## How should this be tested?

Add new app "Plausible (Custom)"

- [ ] Configure Plausible URL
- [ ] Configure Domain URL

## Checklist


hariombalhara:
Thanks for the PR @Hybes 🙏
I am not very sure but have you considered the possibility of using the same Plausible app that can allow specifying the PLAUSIBLE_URL. We can make some changes so that the configuration shows up only for self hosters.

Hybes:
> Thanks for the PR @Hybes 🙏 I am not very sure if it can work but have you considered the possibility of using the same Plausible app that can allow specifying the PLAUSIBLE_URL. We can make some changes so that the configuration shows up only for self hosters.
> 
> If same app can work, I think that would be the best approach

Yeah it's certainly possible, i'm not certain how to make that value of PLAUSIBLE_URL have a default, basically all it needs to do, is work the exact same way as mine, but if the user doesn't specify - it should just use the URL "plausible.io"

hariombalhara:
That should be trivial. Maybe you can make the required changes to work with existing app first and then it can be done.

hariombalhara (changes requested):
Let's make the changes in existing app.

Hybes:
let me know if you need anything else done to it :)

Hybes:
@hariombalhara @PeerRich this sat inactive for a reason? do you need me to do any more changes to this?

packages/app-store/BookingPageTagManager.tsx
```diff
@@ -17,13 +17,22 @@ export default function BookingPageTagManager({
...
+        }
         const trackingId = getEventTypeAppData(eventType, appId as keyof typeof appDataSchemas)?.trackingId;
         if (!trackingId) {
           return null;
         }
         const parseValue = <T extends string | undefined>(val: T): T =>
           //TODO: Support more template variables.
-          val ? (val.replace(/\{TRACKING_ID\}/g, trackingId) as T) : val;
+          val
+            ? (val.replace(/\{TRACKING_ID\}/g, trackingId).replace(/\{PLAUSIBLE_URL\}/g, plausibleUrl) as T)
```
hariombalhara, line 34: BookingPageTagManager is a generic fn and it shouldn't be aware of an app specific template variable. I am fixing it.

packages/app-store/BookingPageTagManager.tsx
```diff
@@ -17,13 +17,27 @@ export default function BookingPageTagManager({
...
+
+        // We assume that the first variable(which is trackingId) is a required field. Every analytics app would have atleast one variable
+        if (!appData?.variable1) {
           return null;
         }
-        const parseValue = <T extends string | undefined>(val: T): T =>
-          //TODO: Support more template variables.
-          val ? (val.replace(/\{TRACKING_ID\}/g, trackingId) as T) : val;
+
+        const parseValue = <T extends string | undefined>(val: T): T => {
```
hariombalhara, line 28: @Hybes Supports any random variable now

packages/app-store/plausible/zod.ts
```diff
@@ -4,6 +4,7 @@ import { eventTypeAppCardZod } from "../eventTypeAppCardZod";
 
 export const appDataSchema = eventTypeAppCardZod.merge(
   z.object({
+    PLAUSIBLE_URL: z.string().optional().default("https://plausible.io/js/script.js").or(z.undefined()),
```
hariombalhara, line 7: Whe can it be undefined if default is already a string.

packages/app-store/BookingPageTagManager.tsx
Hybes: great stuff!

hariombalhara (approved):

hariombalhara:
@Hybes can you do a final round of testing. Will merge after that.

packages/app-store/plausible/components/EventTypeAppCardInterface.tsx
```diff
@@ -24,8 +25,18 @@ const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({
         }
       }}
       switchChecked={enabled}>
+      <TextField
+        name="Plausible URL"
+        defaultValue="https://plausible.io/js/script.js"
+        placeholder="https://plausible.io/js/script.js"
```
hariombalhara, line 31: Added this placeholder so that if someone removes the entire input text, he knows what kind of inputs goes in it.

hariombalhara:
Merging it. Thanks for your contribution @Hybes !!

PR #8208, opened 2023-04-11
Remove Intl.Provider

joeauyeung:
## What does this PR do?

This PR removes the `Intl.Provider` package to format prices and instead users i18n.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


apps/web/public/static/locales/en/common.json
```diff
@@ -1717,5 +1717,9 @@
   "seats_and_no_show_fee_error": "Currently cannot enable seats and charge a no-show fee",
   "complete_your_booking": "Complete your booking",
   "complete_your_booking_subject": "Complete your booking: {{title}} on {{date}}",
+  "currency_string": "{{amount, currency}}",
+  "charge_card_dialog_body": "You are about to charge the attendee {{amount, currency}}. Are you sure you want to continue?",
+  "charge_attendee": "Charge attendee {{amount, currency}}",
+  "payment_app_commission": "Require payment (0.5% + {{fee, currency}} commission per transaction)",
```
emrysal, line 1723: 0.5% should be PAYMENT_FEE_PERCENTAGE

apps/web/components/apps/App.tsx
```diff
@@ -247,7 +247,7 @@ const Component = ({
             t("free_to_use_apps")
           ) : (
             <>
-              {Intl.NumberFormat("en-US", {
+              {Intl.NumberFormat(i18n.language, {
```
joeauyeung, line 250: @emrysal I was going to leave this field undefined so it could detect the browser locale. The problem was if my locale was "en-CA" and the currency was CAD then it wouldn't display the currency code. The `i18n.language` is just the first the language so it will always show the currency code.

packages/features/eventtypes/components/EventTypeDescription.tsx
```diff
@@ -32,15 +30,13 @@ export const EventTypeDescription = ({
...
 }: EventTypeDescriptionProps) => {
-  const { t } = useLocale();
+  const { t, i18n } = useLocale();
 
   const recurringEvent = useMemo(
     () => parseRecurringEvent(eventType.recurringEvent),
     [eventType.recurringEvent]
   );
 
-  const stripeAppData = getPaymentAppData(eventType);
```
zomars, line 42: Is this deprecated? cc @hariombalhara

emrysal (approved):
Looks good to me 🚀

PR #8218, opened 2023-04-12
fix: Change Detroit to New York as a default for EST time zone

### LINKED ISSUE #8119: [CAL-1419] Change Detroit to New York as a default for EST time zone
CleanShot 2023-04-06 at 11 54 49@2x

Detroit as default for EST is weird.

- [ ] Instead do `America/New York`
- [ ] Also: lets replace "_" with " ": 

CleanShot 2023-04-06 at 11 55 34@2x

Los_Angeles => Los Angeles

Created via Threads. See full discussion: [https://threads.com/34461922510](https://threads.com/34461922510)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1419](https://linear.app/calcom/issue/CAL-1419/change-detroit-to-new-york-as-a-default-for-est-time-zone)</sub>

gitstart-calcom:
Fixes #8119 

Loom video: https://www.loom.com/share/441a4a207e5845c1bf9a786b83334dd9


alannnc (approved):
Looks like a fix that can work on a short term. For a long term we can actually fork react-timezone-select and modify with our custom data.

packages/ui/components/form/timezone-select/TimezoneSelect.tsx
```diff
@@ -33,20 +33,29 @@ export function TimezoneSelect({
     });
   }, [components]);
 
+  // We use modifiedTimezones in place of the allTimezones object replacing any underscores in the curly braces
+  // with spaces and removing the America/Detroit timezone, adding the America/New_York timezone instead.
+  const modifiedTimezones = useMemo(() => {
+    const { "America/Detroit": _, ...rest } = allTimezones;
```
hariombalhara, line 39: I think a better approach would be to fork react-timezone-select and change it there. It's a simple  https://github.com/ndom91/react-timezone-select/blob/main/src/timezone-list.ts#L15 file. Also, maybe worth opening a PR there if they want to incorporate this. cc @zomars 

If someone changes Detroit to something else in the package, it would stop working. So, it isn't a stable change to do. Good as a hack only.

hariombalhara (dismissed):
Left a comment suggesting a change in the approach. But would wait for @zomars @emrysal input.

packages/ui/components/form/timezone-select/TimezoneSelect.tsx
zomars: Is worth generate a [patch-package](https://github.com/ds300/patch-package) for this?

PeerRich: that's a good idea! I'm not a fan of forking npm packages

hariombalhara: @zomars Wow!! Never heard about patch-package. This is what we need here.

zomars (changes requested):
Blocking to go with the patch approach instead. My rationale is that this is too hacky and unsustainable in the long term.

gitstart-calcom:
> Blocking to go with the patch approach instead. My rationale is that this is too hacky and unsustainable in the long term.

We are working on the patch approach and we have made some progress. We will share an update soon.

gitstart-calcom:
Hi @zomars @PeerRich, it seems like the patch package might not solve the issue. 

Using the path-package lib to make the changes directly on the node_modules file, we faced some problems because the react-timezone-select lib uses a lib called spacetime to make the integration with the timezones, and these timezones should have the keys with "_" (like "America/Sao_Paulo). So these changes can be done only on the frontend (like we did as the initial idea), in this case, we will change the timezone names only visually.

As an example here, we need to pass the timezone as "America/Sao_Paulo", if we change this we have the following error: 

image (3)

The same goes for "America/Detroit".

zomars:
> Hi @zomars @PeerRich, it seems like the patch package might not solve the issue.
> 
> Using the path-package lib to make the changes directly on the node_modules file, we faced some problems because the react-timezone-select lib uses a lib called spacetime to make the integration with the timezones, and these timezones should have the keys with "_" (like "America/Sao_Paulo). So these changes can be done only on the frontend (like we did as the initial idea), in this case, we will change the timezone names only visually.
> 
> As an example here, we need to pass the timezone as "America/Sao_Paulo", if we change this we have the following error:
> 
> image (3)
> 
> The same goes for "America/Detroit".

Let's revisit the patch solution later on. I think we're good for now.

zomars (approved):
Ship it

PR #8229, opened 2023-04-12
Hotfix: Revert color-scheme for embed

hariombalhara:
## What does this PR do?

Fixes # (issue)
It would just disable the dark mode with native browser elements.

Avoids the following bug
image

**Environment**: Production


apps/web/lib/app-providers.tsx
```diff
@@ -101,6 +101,17 @@ const CalcomThemeProvider = (
       forcedTheme={forcedTheme}
       storageKey={storageKey}
       attribute="class">
+      {/* Embed Mode can be detected reliably only on client side here as there can be static generated pages as well which can't determine if it's embed mode at backend */}
+      {/* color-scheme makes background:transparent not work in iframe which is required by embed. */}
+      {typeof window !== "undefined" && !isEmbedMode && (
+        <style jsx global>
```
hariombalhara, line 107: At server embed can be false(even if it is embed) and thus this style will be added always. So check for presence of window as well.

sean-brydon (approved):
Happy with this - weird that color-scheme breaks transparent

PR #8235, opened 2023-04-12
Fix payment app credential search

alannnc:
## What does this PR do?

Changes stripe type credential find with actual appId and paymenAppCredential.appId.
This was old code but since using payment gateway it should rely on using appId instead of type.

## Type of change

## Checklist


zomars (approved):

hariombalhara (approved):
Code LGTM

PR #6728, opened 2023-01-26
feat: add sylaps into the app-store in video category

hthetiot:
## What does this PR do?

Add Sylaps (aka https://Sylaps.com) video conference service to the Apps.

Screenshot 2023-01-26 at 14 20 02

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How should this be tested?

- [x] Visit Apps
- [x] Install Sylaps
- [x] Create a event
- [x] Set location to Sylaps
- [x] Test the generated sylaps link

Changes by @hariombalhara 
- imageSrc was still being used for Location select images. Replaced that with logo.
- Ensured that appMetaData.ts module itself derives `dirName` from `slug`. 
- Ensured that converting logo to absolute URL happens at a single place which is `appMetaDeta.ts`
- Dynamic Video apps need not have video in their slug name now. 
- Removed _metadata.ts and added config.json. _metadata.ts is deprecated

PR #8256, opened 2023-04-14
Fix: Sylaps app shown uninstalled after installation

hariombalhara:
As reported here https://github.com/calcom/cal.com/pull/6728#issuecomment-1507771582

The reason for the issue was changing slug from `sylaps` to `sylapsvideo`. A few places were left.
I have removed hardcoded value and instead using the value from config directly.

A Follow-up would be to create a template for dynamic video link apps, so that all new apps being created won't make the same mistakes.

### DIFF
--- a/packages/app-store/sylapsvideo/api/add.ts
+++ b/packages/app-store/sylapsvideo/api/add.ts
@@ -3,6 +3,7 @@ import type { NextApiRequest, NextApiResponse } from "next";
 import prisma from "@calcom/prisma";
 
 import getInstalledAppPath from "../../_utils/getInstalledAppPath";
+import config from "../config.json";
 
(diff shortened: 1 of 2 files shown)


packages/app-store/sylapsvideo/api/add.ts
```diff
@@ -29,7 +30,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         type: appType,
         key: {},
         userId: req.session.user.id,
-        appId: "sylaps",
+        appId: config.slug,
```
hariombalhara, line 33: `slug` is `appId` in the credential table

PeerRich (approved):

PR #8263, opened 2023-04-14
feat: add user hash for intercom

G3root:
Make Intercom more secure

Created via Threads. See full discussion: https://threads.com/34464425134


hariombalhara:
@G3root What exactly is it trying to fix? need a PR description

apps/web/pages/api/intercom-hash.ts
```diff
@@ -0,0 +1,28 @@
...
+    case "GET":
+      if (id.success) {
+        // eslint-disable-next-line turbo/no-undeclared-env-vars
+        const hmac = crypto.createHmac("sha256", process.env.INTERCOM_SECRET ?? "");
+        hmac.update(id.data);
+        const hash = hmac.digest("hex");
+
+        res.status(200).json({ hash });
+      } else {
+        res.status(404).json({ message: "user id is required" });
```
hariombalhara, line 20:
```suggestion
        res.status(400).json({ message: "user id is required" });
```

apps/web/pages/api/intercom-hash.ts
```diff
@@ -0,0 +1,28 @@
...
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { query, method } = req;
+
+  const id = z.string().safeParse(query?.id);
+
+  switch (method) {
+    case "GET":
+      if (id.success) {
+        // eslint-disable-next-line turbo/no-undeclared-env-vars
+        const hmac = crypto.createHmac("sha256", process.env.INTERCOM_SECRET ?? "");
```
hariombalhara, line 14: It looks like a new env variable. It should be added in .env.example with comment on when/why it should be set.

apps/web/pages/api/intercom-hash.ts
```diff
@@ -0,0 +1,28 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import crypto from "node:crypto";
+import { z } from "zod";
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { query, method } = req;
+
+  const id = z.string().safeParse(query?.id);
+
+  switch (method) {
```
hariombalhara, line 10: We should use `defaultHandler` that automatically handles this switch case

packages/features/ee/support/lib/intercom/useIntercom.ts
```diff
@@ -25,12 +25,23 @@ export const useIntercom = () => {
   const { hasPaidPlan } = useHasPaidPlan();
   const { hasTeamPlan } = useHasTeamPlan();
 
-  const open = () => {
+  const open = async () => {
+    let userHash;
+
+    if (data && data.id) {
```
hariombalhara, line 31:
```suggestion
    if (data?.id) {
```

apps/web/pages/api/intercom-hash.ts
```diff
@@ -0,0 +1,28 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import crypto from "node:crypto";
+import { z } from "zod";
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { query, method } = req;
```
hariombalhara, line 6: This endpoint seems open and anyone can send any user ID. Shouldn't we restrict it to userId of the session? In that case, `userId` can be derived from session itself.

hariombalhara (changes requested):
Left some comments

hariombalhara (approved):

turbo.json
```diff
@@ -277,6 +277,7 @@
     "ZOHOCRM_CLIENT_ID",
     "ZOHOCRM_CLIENT_SECRET",
     "ZOOM_CLIENT_ID",
-    "ZOOM_CLIENT_SECRET"
+    "ZOOM_CLIENT_SECRET",
+    "INTERCOM_SECRET"
```
zomars, line 281: Can we sort this alphabetically?

PeerRich: ah shit, didnt see this. maybe in a follow up PR?

zomars: Nah, it's ok

PR #8264, opened 2023-04-14
Feat/subdomain logochanges

sean-brydon:
## What does this PR do?
Allows logos to be configured on a per subdomain basis

Fixes # (issue)
Loom: See threads LFR 

## Type of change

## How should this be tested?
Please add a domain to your hosts files
change package.json 
`"dev": "next dev --hostname 0.0.0.0",`
add a image URL to appLogo on team schema  (we will move this over to a file upload when we have a more permanent org solution) 

start dev server
visit domain configured in /etc/hosts

You may also need to change nextauth url on local host due to CSRF token


packages/features/orgs/SubdomainProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const hostnameParts = url.hostname.split(".");
+
+      if (hostnameParts.length >= 3) {
+        const subdomain = hostnameParts.slice(0, -2).join(".");
+        const domain = hostnameParts.slice(-2).join(".");
+        if (domain === "cal.com" || domain === "cal.dev") {
+          if (subdomain !== "app" && subdomain !== "console") {
+            setIsSubdomain(true);
+            setSubdomain(subdomain);
+            console.log(`Subdomain found: ${subdomain}.${domain}`);
```
sean-brydon, line 24:
```suggestion
```

packages/features/orgs/server/router.ts
```diff
@@ -0,0 +1,31 @@
...
+      const foundOrg = Object.entries(orgs).find(([key, value]) => key === subdomain);
+
+      if (!foundOrg) return LOGO;
+      const [_, entry] = foundOrg;
+      const parsedDomainMap = domainMapSchema.safeParse(entry);
+      return parsedDomainMap.success ? `${WEBSITE_URL}/logos/${parsedDomainMap.data.logo}` : LOGO;
+    }
+    return LOGO;
+  }),
+});
```
sean-brydon, line 31: This will likely be scrapped once orgs is implemented properly -> We can get a lot of this from DB instead of .env

packages/ui/components/logo/Logo.tsx
```diff
@@ -1,14 +1,31 @@
...
+          <img className={small ? "h-4 w-auto" : "h-5 w-auto"} alt="Cal" title="Cal" src={logoSrc ?? LOGO} />
         )}
+        {!isSubdomain &&
+          (icon ? (
+            <img className="mx-auto w-9" alt="Cal" title="Cal" src={LOGO_ICON} />
+          ) : (
+            <img className={small ? "h-4 w-auto" : "h-5 w-auto"} alt="Cal" title="Cal" src={LOGO} />
+          ))}
       </strong>
     </h3>
```
sean-brydon, line 30: When orgs is a feature we should probably refactor this to be our logo  OR orgLogo if defined

apps/web/components/ui/AuthContainer.tsx
```diff
@@ -18,10 +17,7 @@ export default function AuthContainer(props: React.PropsWithChildren<Props>) {
   return (
     <div className="flex min-h-screen flex-col justify-center bg-[#f3f4f6] py-12 sm:px-6 lg:px-8">
       <HeadSeo title={props.title} description={props.description} />
-      {props.showLogo && (
-        // eslint-disable-next-line @next/next/no-img-element
-        <img className="mb-auto h-4" src={LOGO} alt={`${APP_NAME} Logo`} />
-      )}
+      {props.showLogo && <Logo small inline={false} className="mx-auto mb-auto" />}
```
sean-brydon, line 20: Ensure shell uses our logo component

packages/features/orgs/SubdomainProvider.tsx
```diff
@@ -0,0 +1,64 @@
...
+            setIsSubdomain(true);
+            setSubdomain(subdomain);
+          }
+        }
+      }
+    };
+    if (router.isReady) {
+      checkSubdomain();
+    }
+  }, [router]);
```
sean-brydon, line 31: Please test - this shouldnt cause a re-render in the while tree as react is kinda smart and realises the values of the provider dont change. 

We should only get one render per navigation still.

packages/features/orgs/server/router.ts
sean-brydon: We'd likely fetch logo from S3 in the future too so we can provide a direct link without having to put them on website logo 

cc @keithwillcode  i know we briefly talked about logo hosting

packages/features/orgs/SubdomainProvider.tsx
zomars: Can we get this from the server side?

PeerRich (dismissed):
@zomars mentioned an API approach would be better here

packages/features/orgs/SubdomainProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return useContext(SubdomainContext);
+};
+
+interface SubdomainProviderProps {
+  children: ReactNode;
+}
+
+export const SubdomainProvider = ({ children }: SubdomainProviderProps) => {
+  const [enabled, setEnabled] = useState(true);
+  const { data } = trpc.viewer.orgs.getLogo.useQuery(undefined, { enabled });
```
sean-brydon, line 24: We have to disable this query from running after its already been set in the provider. Not need to keep fetching logo URL from endpoint after the inital launch. 

Any subdomain changes would hard refresh the page anyway so we dont need to force this.

packages/features/orgs/server/router.ts
```diff
@@ -0,0 +1,49 @@
...
+      // Remove port from domain if it exists (e.g. cal.com:3000) -> cal.com
+      const domainParts = domain.split(":");
+      if (domainParts.length > 1) {
+        domain = domainParts[0];
+      }
+      if (domain === "cal.com" || domain === "cal.dev") {
+        if (subdomain !== "app" && subdomain !== "console") {
+          appSubdomain = subdomain;
+        }
+      }
```
sean-brydon, line 28: Currently we only support this on our domain

packages/features/orgs/server/router.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+      select: {
+        appLogo: true,
+      },
+    });
+
+    if (!team) return DEFAULT_LOGO;
+
+    return {
+      logo: team?.appLogo || LOGO,
```
sean-brydon, line 44: If the team doesnt have an app-logo and is on a subdomain return our logo

packages/ui/components/logo/Logo.tsx
```diff
@@ -1,20 +1,41 @@
...
-          <img className="mx-auto w-9 dark:invert" alt="Cal" title="Cal" src={LOGO_ICON} />
-        ) : (
+        {isSubdomain && (
           <img
             className={classNames(small ? "h-4 w-auto" : "h-5 w-auto", "dark:invert")}
-            alt="Cal"
-            title="Cal"
-            src={LOGO}
+            alt={subdomain || "Cal"}
+            title={subdomain || "Cal"}
```
sean-brydon, line 24: We can use the subdomain as the alt text here to improve a11y.

"ikea.cal.com" will have the alt "ikea"

packages/features/orgs/SubdomainProvider.tsx
zomars, line 24: We can get the same behavior by just doing this. No need to use useState here.

```suggestion
  const { data } = trpc.viewer.orgs.getLogo.useQuery(undefined, {
    staleTime: Infinity,
  });
```

packages/features/orgs/SubdomainProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  // We only want this to run once so we disable the query after the first run
+  useEffect(() => {
+    if (!data?.isSubdomain) setEnabled(false); // If we're not on a subdomain, disable the query
+    if (data?.subdomain && enabled) {
+      setEnabled(false);
+    }
+  }, [data, enabled]);
```
zomars, line 32: See comment above

```suggestion
```

packages/trpc/server/routers/viewer.tsx
```diff
@@ -1338,6 +1339,7 @@ export const viewerRouter = mergeRouters(
     appRoutingForms: app_RoutingForms,
     eth: ethRouter,
     features: featureFlagRouter,
+    orgs: orgsRouter,
```
zomars, line 1342:
```suggestion
    organizations: orgsRouter,
```

apps/web/pages/api/logo.ts
zomars, line 1: Added this example @sean-brydon. This is what I've meant that keep it all in the server.

This endpoint can detect if we're serving from a subdomain and serve the logo dynamically.

That way we only need to call `/api/logo` from the frontend and it will always respond with the correct logo.

We could either add variants like square icon (for tablet) and event dark and light mode if we get fancy.

packages/ui/components/logo/Logo.tsx
```diff
@@ -1,18 +1,33 @@
...
+            // TODO: Implement LOGO_ICON on endpoint
+            src="/api/logo?icon=1"
+          />
         ) : (
           <img
             className={classNames(small ? "h-4 w-auto" : "h-5 w-auto", "dark:invert")}
             alt="Cal"
             title="Cal"
-            src={LOGO}
+            src="/api/logo"
```
zomars, line 30: Updated to use the previously made example. With this we don't even need a SubdomainProvider at all

apps/web/pages/api/logo.ts
```diff
@@ -0,0 +1,106 @@
...
+import { z } from "zod";
+
+import { LOGO, LOGO_ICON } from "@calcom/lib/constants";
+import logger from "@calcom/lib/logger";
+
+const log = logger.getChildLogger({ prefix: [`[api/logo]`] });
+
+function removePort(url: string) {
+  return url.replace(/:\d+$/, "");
+}
```
sean-brydon, line 14: Get rid of the port if testing on local host

apps/web/pages/api/logo.ts
```diff
@@ -0,0 +1,106 @@
...
+
+function extractSubdomainAndDomain(url: string): [string, string] | null {
+  try {
+    const parsedUrl = new URL(url);
+    const hostParts = parsedUrl.href.split(".");
```
sean-brydon, line 19: @zomars you had this as `parsedUrl.hostname` This was always empty for me so we'd always get the Cal.com logo no matter what 🤔

apps/web/pages/api/logo.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+  } catch (error) {
+    return null;
+  }
+}
+
+const logoApiSchema = z.object({
+  icon: z.coerce.boolean().optional(),
+});
+
```
sean-brydon, line 39: Support `api/logo?icon=1`

apps/web/pages/api/logo.ts
```diff
@@ -0,0 +1,106 @@
...
+      },
+      select: {
+        appLogo: true,
+        appIconLogo: true,
+      },
+    });
+
+    const filteredLogo = parsedQuery?.icon ? team?.appIconLogo : team?.appLogo;
+
+    if (!filteredLogo) throw new Error("No team appLogo");
```
sean-brydon, line 94: Use Icon logo on a team level too

sean-brydon:
Thanks for the help there @zomars happy with how this looks now. Let me know if you have any questions. 

i did think about the additional fields to add darkmode logo + darkmode icon too. However, not 100% sure if we should add them to the team field or go with a metadata approach @leog we have discussed this in detail before

zomars (approved):
So clean! Nice job @sean-brydon 

Ship it

PR #8265, opened 2023-04-14
Hotfix: Label selects the option of different Radio Group

hariombalhara:
## What does this PR do?

Fixes bug where Option-1 label for Radio-2 selects Option-1 for Radio-1.  [See this loom](https://www.loom.com/share/10a07bfb97bc416783abda9c35cc0cec)

You can replicate the bug here https://cal.com/hariom/book?type=264603&duration=15&date=2023-04-18T09%3A00%3A00%2B05%3A30&slug=2-radios&timeFormat=HH%3Amm&count=

### DIFF
--- a/packages/features/form-builder/Components.tsx
+++ b/packages/features/form-builder/Components.tsx
@@ -259,7 +259,7 @@ export const Components: Record<BookingFieldType, Component> = {
   },
   radio: {
     propsType: "select",
-    factory: ({ setValue, value, options }) => {
+    factory: ({ setValue, name, value, options }) => {
       return (
         <Group
           value={value}
@@ -272,7 +272,7 @@ export const Components: Record<BookingFieldType, Component> = {
                 label={option.label}
                 key={`option.${i}.radio`}
                 value={option.label}
-                id={`option.${i}.radio`}
+                id={`${name}.option.${i}.radio`}
               />
             ))}
           </>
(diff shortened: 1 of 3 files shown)


packages/features/form-builder/Components.tsx
```diff
@@ -272,7 +272,7 @@ export const Components: Record<BookingFieldType, Component> = {
                 label={option.label}
                 key={`option.${i}.radio`}
                 value={option.label}
-                id={`option.${i}.radio`}
+                id={`${name}.option.${i}.radio`}
```
hariombalhara, line 275: It won't be required if we move radio input inside label in https://github.com/calcom/cal.com/blob/2752061263fafa93cbaa756ec1f77e2dc5b8534b/packages/ui/form/radio-area/Radio.tsx#L56

@sean-brydon What do you think about it?

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -640,9 +640,9 @@ export const ComponentForField = ({
       <WithLabel field={field} readOnly={readOnly}>
         <componentConfig.factory
           placeholder={field.placeholder}
+          name={field.name}
```
hariombalhara, line 643: Send name everywhere just in case others also need it.

PeerRich (approved):

PR #8270, opened 2023-04-14
Move some dependencies into async

emrysal:
## What does this PR do?

Creates a new sendVerificationRequest function which independently imports handlebars and the transporter, hoping to keep these out of the global context.


zomars (approved):
LGTM

PR #8271, opened 2023-04-14
feat: Added more query params to successRedirect page

### LINKED ISSUE #7802: Pass meeting details in query or body to redirection URL
If post-reservation URL redirection is enabled, it would be VERY important to OPTIONALLY pass all the meeting details to the URL, in the query or body.

Currently, the only parameters that are passed in the query are the following:

```
isSuccessBookingPage:true
email:example@email.com
eventTypeSlug:secret
```

This should include also meeting date, question responses, etc. (The same way as calendly does).

Additionally, it would be great to CONFIGURE the DELAY of the redirection. Currently the redirection is donde after 10 or more seconds after the confirmation page is shown. This is way TOO much, before the redirection occurs probably the user has closed the window. 
Ideally the seconds of the redirection should be configurable from admin settings.

titanventura:
## What does this PR do?

Pass extra booking params to redirect URL
Fixes #7802

#### Edit by @hariombalhara 
___
Right now there is an option to enable sending these query params in this PR.
Screenshot 2023-05-02 at 1 25 51 PM
A sample redirect URL generated using the avove configuration `https://www.google.com/?isSuccessBookingPage=true&email=d%40e.com&eventTypeSlug=abc&title=ABC+between+Team+Pro+Example+and+d&description=sd&startTime=2023-05-12T03%3A45%3A00.000Z&endTime=2023-05-12T04%3A00%3A00.000Z&location=integrations%3Awhereby_video`

Params Sent to Custom Success Redirect Page
- **booker_email** - Email of the person booking the meeting.
- **title** - Booking Title
- **description** - Additional Notes of the booking
- **startTime** - Booking Start Time as ISO String (Sample `2023-05-12T03:45:00.000Z`)
- **endTime** - Booking End Time as ISO String(Sample `2023-05-12T04:00:00.000Z`)
- **location** - It has many possible values. For Video Apps, the value will be in format` integrations:{APPNAME}_video` e.g. integrations:whereby_video. See more possible values for it [here](https://cal.com/docs/core-features/bookings/prefill-fields#pre-selecting-organizer-address-in-location) 

I am of the opinion that we should do it by default and that checkbox isn't required. There is a very edge case where sending a lot of query params can break the URL size limit but we can ignore that for now.
cc @PeerRich @Jaibles 
___

## Type of change

## Checklist


apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -268,6 +269,30 @@ const BookingPage = ({
...
+    const redirectQueryParamKeys = [
+      "title",
+      "description",
+      "startTime",
+      "endTime",
+      "location",
+      ...objectQueryParamKeys,
+    ];
+
+    const params = _.pick(booking, redirectQueryParamKeys);
```
sean-brydon, line 287: Lets try and not import lodash - especially in our booking page. 

I havent tested but something like this should replicate the functionality 

```suggestion
    const params = Object.keys(booking).filter(key => redirectQueryParamKeys.includes(key)).reduce((obj, key) => Object.assign(obj, { [key]: booking[key] }), {});
```

sean-brydon (dismissed):
Thanks for the contribution :) Havent tested but lets not implement lodash - especially in our booking page 

Its huge

apps/web/components/booking/pages/BookingPage.tsx
titanventura: thank you for the suggestion. added this snippet to my branch and removed lodash import from the top of the file

titanventura:
Thanks for the support ! I would like to thank @adarsh-a-tw - [link](https://github.com/adarsh-a-tw) who paired with me and helped me raise this PR !

titanventura:
@sean-brydon Please review. Thanks !

titanventura:
Please review !

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -268,6 +268,32 @@ const BookingPage = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  const getBookingRedirectExtraParams = (booking: any) => {
+    if (!eventType.metadata?.passExtraParamsToRedirectUrl) {
+      return {};
+    }
+
+    const objectQueryParamKeys = ["customInputs", "responses", "metadata", "user", "attendees"];
```
hariombalhara, line 276: `customInputs` is deprecated(it is there only for old Event Types and is migrated to responses format when that even-type is saved again). So, we can remove it
```suggestion
    const objectQueryParamKeys = [ "responses", "metadata", "user", "attendees"];
```

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -290,6 +316,7 @@ const BookingPage = ({
         eventTypeSlug: eventType.slug,
         seatReferenceUid: "seatReferenceUid" in responseData ? responseData.seatReferenceUid : null,
         ...(rescheduleUid && booking?.startTime && { formerTime: booking.startTime.toString() }),
+        ...getBookingRedirectExtraParams(responseData),
```
hariombalhara, line 319: This should be done as part of `bookingSuccessRedirect` and that fn is called at multiple places as redirections happens from multiple places.

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -268,6 +268,32 @@ const BookingPage = ({
...
+    }
+
+    const objectQueryParamKeys = ["customInputs", "responses", "metadata", "user", "attendees"];
+    const redirectQueryParamKeys = [
+      "title",
+      "description",
+      "startTime",
+      "endTime",
+      "location",
+      ...objectQueryParamKeys,
```
hariombalhara, line 283: Instead of differentiating manually b/w object and non-object params, we should use the typeof operator to identify the value. This would allow us to maintain a single list.

Also, I would like to keep the list very strict because whatever we expose has to be considered for backward compatibility as websites would rely on those params.

Stable Params - Good to be passed as is
- title
- description
- startTime
- endTime
- location


`user` and `attendees` might have different props depending on what we queried. So, we need to pick keys specifically from here. So, it might be better to remove these keys and wait for a requirement from users as to what they use it for.

hariombalhara (changes requested):
Left some comments.

titanventura:
thanks for the suggestions @hariombalhara . Have made the changes. Please review !

titanventura:
Any updates on this ?

packages/lib/bookingSuccessRedirect.ts
```diff
@@ -1,21 +1,42 @@
...
+    "title",
+    "description",
+    "startTime",
+    "endTime",
+    "location",
+  ];
+
+  return (Object.keys(booking) as BookingResponseKey[])
+    .filter((key) => redirectQueryParamKeys.includes(key))
+    .reduce((obj, key) => Object.assign(obj, { [key]: booking[key] }), {});
```
hariombalhara, line 19:
```suggestion
    .reduce((obj, key) => ({ ...obj, [key]: booking[key] }), {});
```

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -13,17 +13,13 @@ import { CAL_URL } from "@calcom/lib/constants";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { Button, Checkbox } from "@calcom/ui";
 
-import type { EventType } from ".prisma/client";
-
 type Props = {
   payment: Omit<Payment, "id" | "fee" | "success" | "refunded" | "externalId" | "data"> & {
     data: StripePaymentData | StripeSetupIntentData;
   };
-  eventType: { id: number; successRedirectUrl: EventType["successRedirectUrl"] };
```
hariombalhara, line 22: We must use correct type here.

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -53,21 +49,21 @@ const PaymentForm = (props: Props) => {
...
-          return_url: props.eventType.successRedirectUrl || `${CAL_URL}/booking/${props.bookingUid}`,
+          return_url: props.eventType.successRedirectUrl || `${CAL_URL}/booking/${props.booking.uid}`,
         },
       });
     } else if (paymentOption === "ON_BOOKING") {
       payload = await stripe.confirmPayment({
         elements,
         confirmParams: {
-          return_url: props.eventType.successRedirectUrl || `${CAL_URL}/booking/${props.bookingUid}`,
+          return_url: props.eventType.successRedirectUrl || `${CAL_URL}/booking/${props.booking.uid}`,
```
hariombalhara, line 66: It also needs to be passed those query params

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -13,17 +13,13 @@ import { CAL_URL } from "@calcom/lib/constants";
...
   payment: Omit<Payment, "id" | "fee" | "success" | "refunded" | "externalId" | "data"> & {
     data: StripePaymentData | StripeSetupIntentData;
   };
-  eventType: { id: number; successRedirectUrl: EventType["successRedirectUrl"] };
+  eventType: any;
   user: { username: string | null };
-  location?: string | null;
-  bookingId: number;
-  bookingUid: string;
+  booking: any;
```
hariombalhara, line 22: Here too

hariombalhara (changes requested):
Thanks, @titanventura for the patience 🙏. 
I left some comments that require changes in the code.

Also waiting on confirmation from @PeerRich  about removing that checkbox and enable it by default.

hariombalhara:
@titanventura Please remove the checkbox now. I confirmed this from @PeerRich

packages/features/ee/payments/components/Payment.tsx
titanventura: 👍

packages/features/ee/payments/components/Payment.tsx
titanventura: 👍

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -270,9 +271,7 @@ const BookingPage = ({
   }, []);
 
   const mutation = useMutation(createBooking, {
-    onSuccess: async (responseData) => {
-      const { uid } = responseData;
-
+    onSuccess: async (responseData: BookingResponse) => {
```
zomars, line 274: [Prefer inferred types over explicit ones](https://handbook.cal.com/engineering/best-practices/prefer-inferred-types-over-explicit-ones)

```suggestion
    onSuccess: async (responseData) => {
```

apps/web/components/booking/pages/BookingPage.tsx
```diff
@@ -49,6 +49,7 @@ import { AlertTriangle, Calendar, RefreshCw, User } from "@calcom/ui/components/
 
 import { timeZone } from "@lib/clock";
 import useRouterQuery from "@lib/hooks/useRouterQuery";
+import type { BookingResponse } from "@lib/types/booking";
```
zomars, line 52:
```suggestion
```

apps/web/test/lib/handleChildrenEventTypes.test.ts
```diff
@@ -96,11 +96,11 @@ describe("handleChildrenEventTypes", () => {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
       // eslint-disable-next-line
-      const { schedulingType, id, teamId, timeZone, users, ...evType } = mockFindFirstEventType({
-        id: 123,
-        metadata: { managedEventConfig: {} },
-        locations: [],
-      });
+      const {schedulingType, id, teamId, timeZone, users, scheduleId, offsetStart, ...evType} = mockFindFirstEventType({
```
zomars, line 99: Prettier won't be happy

```suggestion
      const { schedulingType, id, teamId, timeZone, users, scheduleId, offsetStart, ...evType } = mockFindFirstEventType({
```

apps/web/test/lib/handleChildrenEventTypes.test.ts
```diff
@@ -217,11 +218,11 @@ describe("handleChildrenEventTypes", () => {
...
```
zomars, line 221: Let's avoid formatting changes

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -53,21 +51,35 @@ const PaymentForm = (props: Props) => {
 
     let payload;
     const params: { [k: string]: any } = {
-      uid: props.bookingUid,
+      uid: props.booking.uid,
       email: router.query.email,
     };
+
+    const returnUrl = new URL(
```
zomars, line 58: Can we encapsulate the return URL logic in a function?

```suggestion
    const return_url = getReturnUrl(props);
```

yarn.lock
zomars, line 1: Will undo lock file changes here.

zomars (dismissed):
Left some more comments 🙏🏽

packages/emails/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "dx": "docker compose up -d"
+    "dx": "docker compose up -d || docker-compose up -d"
```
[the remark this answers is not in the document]
titanventura: `yarn dx threw some errors for me. The issue is that, `compose` subcommand is not available for the version of docker installed in my system. Also saw that `|| docker-compose up -d` was included in other packages under calcom. Seems like a relevant change. Do let me know if this interpretation is wrong. Thanks !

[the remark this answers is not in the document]
zomars: I see. In that case we rather introduce this change in another PR. Our rationale is that this change has nothing to do with "Adding option to pass extra booking query params to redirect url"

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -53,21 +65,24 @@ const PaymentForm = (props: Props) => {
 
     let payload;
     const params: { [k: string]: any } = {
-      uid: props.bookingUid,
+      uid: props.booking.uid,
       email: router.query.email,
     };
+
+    const returnUrl = getReturnUrl(props);
```
zomars, line 72: `returnUrl` is already a string. We can just do

```suggestion
    const return_url = getReturnUrl(props);
```

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -53,21 +65,24 @@ const PaymentForm = (props: Props) => {
...
+
     if (paymentOption === "HOLD" && "setupIntent" in props.payment.data) {
       payload = await stripe.confirmSetup({
         elements,
         confirmParams: {
-          return_url: props.eventType.successRedirectUrl || `${CAL_URL}/booking/${props.bookingUid}`,
+          return_url: returnUrl.toString(),
```
zomars, line 78: And use shorthand notation here.
```suggestion
          return_url,
```

packages/features/ee/payments/components/Payment.tsx
```diff
@@ -53,21 +65,24 @@ const PaymentForm = (props: Props) => {
...
         },
       });
     } else if (paymentOption === "ON_BOOKING") {
       payload = await stripe.confirmPayment({
         elements,
         confirmParams: {
-          return_url: props.eventType.successRedirectUrl || `${CAL_URL}/booking/${props.bookingUid}`,
+          return_url: returnUrl.toString(),
```
zomars, line 85: And here

```suggestion
          return_url,
```

packages/trpc/server/routers/viewer/teams/inviteMember.handler.ts
```diff
@@ -135,7 +135,6 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
         inviteTeamOptions.isCalcomMember = false;
       }
 
-
```
zomars, line 138: Unnecessary diff here.

zomars (commented):
Left some more comments. Almost there! 🙏🏽

packages/emails/package.json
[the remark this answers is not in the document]
titanventura: Sure, seems valid. I'll revert this, create a new PR. Apart from this change, are there any other changes related to this PR ?

packages/trpc/server/routers/viewer/teams/inviteMember.handler.ts
titanventura: I think this got introduced when i ran `yarn lint -- --fix`

Jaibles:
I agree that I don't think we need the checkbox. I also don't think we need the text in red in the screenshot.

hariombalhara:
> Im not sure about the checkbox option. The average user is not going to know what that sentence means so we either need to just do it by default and hide the checkbox or explain it better. Perhaps a little info ℹ️ tooltip like we have in other places.
> 
> I don't think we need the text in red in the screenshot though.

@Jaibles Yeah we agreed on removing the checkbox.

hariombalhara:
@titanventura Are you there to review the changes I pushed?

I have made the changes to send email as new param `booker_email` so that it's clear whose email it is. Also documented, the params that are are sending.
I have also pulled in latest main and resolved some conflicts.

packages/prisma/zod-utils.ts
```diff
@@ -552,7 +552,6 @@ export const allManagedEventTypeProps: { [k in keyof Omit<Prisma.EventTypeSelect
   price: true,
   slug: true,
   length: true,
-  offsetStart: true,
```
hariombalhara, line 555: Why are these changes made in managedEventTypes ?

PeerRich:
/bonus 20 @titanventura can you take another look at this PR and fix the merge conflicts? <3

hariombalhara (approved):
LGTM now !! Thanks @titanventura

PR #8274, opened 2023-04-14
refactor: remove unnecessary or operation

anonrig:
Unless I'm not seeing something obvious, this OR statement shouldn't be required.

### DIFF
--- a/packages/trpc/server/routers/viewer/bookings.tsx
+++ b/packages/trpc/server/routers/viewer/bookings.tsx
@@ -145,15 +145,7 @@ export const bookingsRouter = router({
         },
         unconfirmed: {
           endTime: { gte: new Date() },
-          OR: [
-            {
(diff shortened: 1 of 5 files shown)


anonrig:
It seems like the automatic git hook fixes a listing error while I commit...

roae (approved):
Nice catch, thank you @anonrig! :pray:

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -145,15 +145,7 @@ export const bookingsRouter = router({
         },
         unconfirmed: {
           endTime: { gte: new Date() },
-          OR: [
-            {
-              recurringEventId: { not: null },
```
zomars, line 150: Recurring events are displayed in another tab, so this condition is needed.

PR #8275, opened 2023-04-14
perf: parallelize database calls to improve speed

anonrig:
No need to sequentially make multiple database calls, since they're not bound to each other.


roae (approved):
LGTM! :rocket:

zomars (approved):
LGTM

PR #8276, opened 2023-04-14
perf: reduce array iteration on bookings router

anonrig:
No need to filter and later reduce the array. This is an o(2n) operation. This is improved by removing filter. The other change is calling `has` directly, which removes the force casting to boolean.


roae (approved):
Thanks @anonrig

zomars (approved):

PR #8282, opened 2023-04-14
fix: /[user]/[type]: redirect to /[user] if invalid `type`

### LINKED ISSUE #4687: Redirection of users landing on your 404
## Problem

For example I create

cal.com/calcom/30m

and later decide to only provide a shorter time and update it to

cal.com/calcom/20m

users that'll try the old url will land on an 404 page, it looks nice but will decrease the conversion extremely

Bildschirm­foto 2022-09-24 um 01 25 44

#### Importance

This could happen unknowingly and cause a big loss of conversions

## Idea

Introduction of an redirection option.

For the start maybe toggle to redirect every 404 to start root (cal.com/calcom/ in example)

Later even select an event to redirect to, for better conversion.

mihirgupta0900:
## What does this PR do?

Redirects all 404 of users back to the profile. For example, if someone lands on `/exampleuser/nonexistingevent`, they will be redirected to `/exampleuser`

Fixes #4687
/claim #4687

## Type of change

## How should this be tested?

- [ ] Visit a non-existent event of a user, and you should land on the user's profile page instead of a 404

## Checklist


mihirgupta0900:
Very minimal changes for the issue, would love to know if:
- there are other places that might need a similar change
- any tests that I need to create/edit

hariombalhara (changes requested):
The approach looks good. But we should do the same change for team events team/[slug]/[type]

Also, might make sense to do it for d/[link]/[slug] (private links) as well

hariombalhara:
> Very minimal changes for the issue, would love to know if:
> 
> * there are other places that might need a similar change
> * any tests that I need to create/edit

I don't think we need tests for this

ameeetgaikwad:
Hey if this is not going forward, can i take the issue?

ologbonowiwi:
@hariombalhara can you review #9325? addressed your changes there

I didn't redirect to private links because this would go against the idea of having a link protecting the user/slug.

PR #8294, opened 2023-04-15
Improvement: Improve App-Store Seeding

hariombalhara:
## What does this PR do?
- Make app-store seeding handle `slug`/`dirName` change well
- One App shouldn't be able to break other App or the entire cal.com build. Ignore if an App fails to seed

**Environment**: Production

**Motivation**:
`sylaps` app during development had `sylaps` slug and somehow during development itself, author probably decided to change the slug to `sylapsvideo` but the change wasn't done everywhere. The PR got merged and then Sylaps app wasn't working because the slug in DB was still going as `sylaps`.  In [the PR](https://github.com/calcom/cal.com/pull/8256/files#diff-1837d0b47f980bd1f8d6833fadea05b258edc5a40703fc4f92bad3ff8397eec0), slug was fixed but this actually broke the seeder because it actually tried to create new App with same dirName.
See https://calendso.slack.com/archives/C02QP4Z3G4E/p1681534380620769 where the bug was reported initially.
It didn't happen in the production deployment because the app seeder doesn't run there now. It only runs locally(AFAIK)
https://calendso.slack.com/archives/C02QP4Z3G4E/p1681534380620769

## Type of change

## How should this be tested?

- [x] Try changing slug/dirName one by one and see that already seeded App is able to get that update.


packages/prisma/seed-app-store.config.json
```diff
@@ -232,5 +232,12 @@
     "slug": "discord",
     "type": "discord_video",
     "isTemplate": false
+  },
+  {
+    "dirName": "sylapsvideo",
```
hariombalhara, line 237: Add sylaps video through new AppStore seeding flow.

packages/prisma/seed-app-store.ts
```diff
@@ -157,16 +157,60 @@ async function createApp(
...
+          {
+            slug,
+          },
+          {
+            dirName,
+          },
+        ],
+      },
+    });
+    const data = { slug, dirName, categories, keys };
```
hariombalhara, line 183: Removed enabled:true. This is to be consistent with https://github.com/calcom/cal.com/pull/8092

emrysal (approved):
Looks good to me, thanks for the clarifying comment

PR #8296, opened 2023-04-15
Fix: cssVars API not working correctly on Booker Timeslot selection page

### LINKED ISSUE #8321: [CAL-1505] [CAL-1504] CSS Vars changes don't work on timeslot selection page of Booker
Screenshot 2023-04-17 at 12 30 29 PM

<sub>[CAL-1504](https://linear.app/calcom/issue/CAL-1504/css-vars-changes-dont-work-on-timeslot-selection-page-of-booker)</sub>

hariombalhara:
## What does this PR do?
- The reason seems to be that useEmbedUIConfig is being used in _app.tsx and currently Embed flow isn't correctly able to trigger renders for that for SSG pages.
- Because setting CSS Vars doesn't actually need React. It would be more performant to simply use Javascript to create a style tag with those CSS Vars. This would allow us to not spend time fixing that React rendering issue(mentioned above). We can't do SSR for the style as it works through Embed API that works on client.

Fixes #8321

**Environment**: Production

## Type of change

## How should this be tested?

- [x] Open rick/get-rick-rolled directly in embed(Browsing through profile to the page would work already). You will see that css vars changes aren't applied. It is fixed by this PR.

## Checklist


apps/web/public/embed-init-iframe.js
```diff
@@ -24,3 +24,23 @@ window.getEmbedNamespace = () => {
   const namespace = url.searchParams.get("embed");
   return namespace;
 };
+window.CalEmbed = window.CalEmbed || {};
+window.CalEmbed.applyCssVars = (cssVarsPerTheme) => {
```
hariombalhara, line 28: Verify that Next.js doesn't revert this style being added.

apps/web/pages/_app.tsx
```diff
@@ -28,17 +27,6 @@ const calFont = localFont({
 function MyApp(props: AppProps) {
   const { Component, pageProps, err, router } = props;
   let pageStatus = "200";
-  const { cssVarsPerTheme } = useEmbedUiConfig();
```
hariombalhara, line 31: Handled by JS out of React context

apps/web/public/embed-init-iframe.js
hariombalhara, line 1: Moved as a ts file in embed-core.

hariombalhara (commented):
Self review done

packages/embeds/embed-react/inline.tsx
```diff
@@ -20,6 +20,19 @@ function App() {
...
       });
+
+      api("ui", {
+        cssVarsPerTheme: {
+          light: {
+            "cal-border-booker": "red",
+            "cal-border-booker-width": "20px",
+          },
+          dark: {
+            "cal-border-booker": "red",
```
sean-brydon, line 31: Do we default these to red ? Or was this just for testing?

hariombalhara: This is actually a test page for inline embed. 
Not clear from the name though. Will rename it in a followup

sean-brydon: Gotcha :) Other than that LGTM

sean-brydon (approved):
LGTM - nice improvement :)

PR #8301, opened 2023-04-15
perf: improve application list execution

anonrig:
This pull request includes the following changes:

- Removes unnecessary string to string comparisons. For example: `val === undefined` is always faster than `typeof val === 'undefined'`
- Removes unnecessary memory allocations. Creating an array and later pushing it will always be slower than calling `flatMap`.
- Removed unnecessary branches (if statements). The most significant performance boost in js is done through reducing the if statements


roae (approved):
Looks good, thanks 🙏

zomars (approved):
Tested and LGTM! 

Ship it

packages/trpc/server/routers/viewer/apps.tsx
```diff
@@ -74,34 +62,35 @@ export const appsRouter = router({
...
-            type: app.type,
-            title: app.title,
-            description: app.description,
-            enabled: dbData?.enabled || false,
-            dirName: app.dirName || app.slug,
-            keys: Object.keys(keys).length === 0 ? null : keys,
-          });
+        // `typeof val === 'undefined'` is always slower than !== undefined comparison
+        // it is important to avoid string to string comparisons as much as we can
+        if (keysSchema !== undefined) {
```
zomars, line 74: I think this is causing a type error:

image

zomars: Pushed a fix

PR #8302, opened 2023-04-15
perf: improve application toggle flow

anonrig:
This pull request includes some caching on translation logic. I've also added some TODOs for future reference.


packages/trpc/server/routers/viewer/apps.tsx
```diff
@@ -142,6 +142,8 @@ export const appsRouter = router({
 
       // If disabling an app then we need to alert users based on the app type
       if (input.enabled) {
```
roae, line 144: disabling or enabling?

packages/trpc/server/routers/viewer/apps.tsx
```diff
@@ -142,6 +142,8 @@ export const appsRouter = router({
...
+        const translations = new Map(["en", await getTranslation("en", "common")]);
```
roae, line 145: English will always be loaded, even when the user does not use it?

packages/trpc/server/routers/viewer/apps.tsx
```diff
@@ -158,18 +160,28 @@ export const appsRouter = router({
             },
           });
 
+          // TODO: This should be done async probably using a queue.
           Promise.all(
             appCredentials.map(async (credential) => {
```
roae, line 165: Here we could use `for ..of`, right?

roae (changes requested):

packages/trpc/server/routers/viewer/apps.tsx
anonrig: For of will block the execution. Cal needs something like a queue that handles this operation async and unrelated from the execution of this current function.

packages/trpc/server/routers/viewer/apps.tsx
anonrig: Updated with the changes.

zomars (approved):
Pushed some fixes, tested and LGTM to ship.

LGTM

packages/trpc/server/routers/viewer/apps.tsx
zomars: pushed a fix for this

roae (approved):
LGTM!

PR #8303, opened 2023-04-15
perf: replace `notIn` with `equals` on booking

anonrig:
This pull request replaces `notIn` with `equals` due to performance constraints.


roae (approved):
🚀

PR #8304, opened 2023-04-15
perf: improve booking confirm authorization

anonrig:
Previous implementation was wrong and unperformant. This pull request fixes the following bug, and reduces application layer data and moves some comparison to the database layer.

```
eventType.users.find((user) => user.id === user.id)
```

(It was always returning truthy)


roae (approved):
Looks good! 🙏

zomars (approved):

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -781,30 +781,22 @@ export const bookingsRouter = router({
...
 
     const authorized = async () => {
-      // if the organizer
-      if (booking.userId === user.id) {
-        return true;
-      }
       const eventType = await prisma.eventType.findUnique({
         where: {
           id: booking.eventTypeId || undefined,
+          schedulingType: SchedulingType.COLLECTIVE,
```
zomars, line 787: We have a type error here:

image

zomars: Pushed a fix

PR #8308, opened 2023-04-16
perf: use getSlimServerSession for session retrieval

Mythie:
## What does this PR do?

Creates a new `getSlimServerSession` method and replaces all calls to `getServerSession`.

The new method is much faster than `getServerSession` with it not requiring the NextAuth options object which includes a number of additional packages.

Additionally introduces a primitive in-memory cache for slim sessions. This cache may need to be reworked prior to this PR being accepted since it could affect self-hosters due to lack of TTL and such, if we feel strongly about that we could use something like `lru-cache` instead which would automatically clear out less frequently accessed entries.

I've documented some caveats to this new method within the file so please have a read through.

## Type of change

- Performance Improvement

## How should this be tested?

- Verify that you are still able to access all endpoints as per normal.
- Ideally verify that the speed is faster but that's harder to do without tracing.


PeerRich:
do we need the old getServerSession? if not, lets overwrite it no?

Mythie:
> do we need the old getServerSession? if not, lets overwrite it no?

That's an option, I'd kept it for posterity since `getSlimServerSession` may not always be what you want due to caveats like not refreshing tokens.

If we think we're not too stressed on that front since our tokens are issued for 30 days at a time then I'm happy to just replace.

PeerRich:
> > do we need the old getServerSession? if not, lets overwrite it no?
> 
> That's an option, I'd kept it for posterity since `getSlimServerSession` may not always be what you want due to caveats like not refreshing tokens.
> 
> If we think we're not too stressed on that front since our tokens are issued for 30 days at a time then I'm happy to just replace.

will defer here to @roae and @zomars

zomars:
> > do we need the old getServerSession? if not, lets overwrite it no?
> 
> That's an option, I'd kept it for posterity since `getSlimServerSession` may not always be what you want due to caveats like not refreshing tokens.
> 
> If we think we're not too stressed on that front since our tokens are issued for 30 days at a time then I'm happy to just replace.

I'm ok replacing the old getServerSession altogether.

Mythie:
Am I missing something or are the unit tests and such not running for this PR?

zomars (approved):
- Reverted code changes to minimize git diffs (and preventing conflicts with other PRs)
- Moved slim logic to original getServerSession
- Updated getSlimServerSession TS signature to match the old getServerSession

Ship it

PR #8311, opened 2023-04-16
perf: Generate calendar cache in background on cold starts

roae:
## What does this PR do?

The calendar cache is also affected by cold starts which means that the getSchedule endpoint will have to wait for this cache to be generated on a cold start increasing the response time.

This PR fetches calendar events without using the cache on cold start, while the cache is generated in the background for upcoming requests.

## Type of change


keithwillcode (approved):

zomars (approved):
LGTM

PR #8316, opened 2023-04-16
more fixes to dark mode booker

PeerRich:
before:

CleanShot 2023-04-16 at 23 50 48@2x

after:

CleanShot 2023-04-16 at 23 50 27@2x


PeerRich:
@sean-brydon can you talk to @Jaibles about some dark mode things? we may close this PR in favor of a general dark mode PR with higher contrasts

sean-brydon:
> @sean-brydon can you talk to @Jaibles about some dark mode things? we may close this PR in favor of a general dark mode PR with higher contrasts
@Jaibles a few people on twitter and on the team have complained about contrast across the app - Could we take a look into this? 

I personally dont see the issue that much - maybe it can be solved with a very subtle bg gradient on the booking page?

PR #8335, opened 2023-04-17
fix: set `referrerpolicy` document specifically for bot requests

p6l-richard:
## What does this PR do?

The summary can be found here: https://github.com/calcom/website/pull/464#issue-1645716509

Fixes # (issue)
image

see full [details here](https://github.com/calcom/website/pull/464#issue:~:text=our%20crawler%20).

**Note:** I'm not sure if this will set the referrer header for bot requests as they do not use a browser.

## Type of change

## How should this be tested?

Once deployed to staging, issue a request to fetch `cal.dev/apps/amie` from a crawler that does _not_ use the browser.
- [ ] There should only be one `_next/static/chunk` resrouce that fails (404), all others should return 200


apps/web/pages/_document.tsx
```diff
@@ -1,13 +1,25 @@
...
 import { csp } from "@lib/csp";
 
 type Props = Record<string, unknown> & DocumentProps;
+// So that bots without referrer header have the referrerpolicy set explicitly by the script tag
+class CustomHead extends Head {
+  getScripts(files: Parameters<Head["getScripts"]>[0]) {
+    const originalScripts = super.getScripts(files);
+    return originalScripts.map((script) => {
+      return cloneElement(script, {
+        referrerpolicy: "strict-origin-when-cross-origin",
```
anonrig, line 18: You can update this on the `next.config.js` as well. You don't need to create a new head components for this. 

An example next.config.mjs

```js
export default {
 headers() {
  return [{ source: '/(.*)', headers: [  {
    key: 'Referrer-Policy',
    value: 'origin-when-cross-origin',
  }]
 }
}

Comment (zomars) [reply apps/web/pages/_document.tsx:18]: Agreed!

Comment (p6l-richard) [reply apps/web/pages/_document.tsx:18]: Oh, that's great actually. Much better, thank you @anonrig.

I also saw that it's possible to set the `referrerpolicy` on a document basis by a single `meta` element. Not if next rewriting the headers or adding this tag to the document make any difference. I will start with the header rewrite just as you proposed and if that doesn't work, I can try out the `meta` element.


```diff
// _document.tsx
 render() {
     //...
     return (
       <Html lang={locale} dir={dir}>
         <Head nonce={nonce}>
           {/* So that our rewrite logic depending on the `referrer` header is always set: */}
+         <meta name="referrer" content="strict-origin-when-cross-origin" />
           <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
```

p6l-richard:
@zomars & @anonrig 

Do you think it will be a security concern if I add the `Referer` header in case it's missing for our `website` repo?
```ts
  async headers() {
    return [
      // So that our asset proxy works also for bot requests without a referer header
      {
        source: "/_next/:data*",
        missing: {
          type: "header",
          key: "Referer",
        },
        headers: [
          {
            key: "Referer",
            value: process.env.NEXT_PUBLIC_WEBSITE_URL + "/_next/:data*",
          },
        ],
      },
    ];
  },
```

MDN states that the referrer could be [a security concern](https://developer.mozilla.org/en-US/docs/Web/Security/Referer_header:_privacy_and_security_concerns#the_referrer_problem) but since I'm setting the `Referer` header only to the current URL, I don't think we're exposing any new information (the request URL was used to fetch the resource after all).

As you can see here, the assets chunks mostly return a 404 if the header isn't set. (note: they sometimes return a 200 as you can see at the end of the first video; that's unexpected I have no idea why, they should return a 404 from my understanding)

- I've captured my screen while i made multiple requests in a row: https://share.cleanshot.com/6WR60T0n
- When I set the referrer to the current URL, it always responds with a 200: https://share.cleanshot.com/FcVfZzX4

If you think this is not a concern, I will close this PR and implement the changes in website's `next.config.js`.

apps/web/pages/_document.tsx
p6l-richard: changed the approach to adding the `Referer` on the website repo instead: [see comment](https://github.com/calcom/cal.com/pull/8335#issuecomment-1513277529).

zomars:
Did changes got lost?

p6l-richard:
> Did changes got lost?

@zomars, yes they did because the proposed change is already present, so I removed the changes.

Instead, my [new proposal](https://github.com/calcom/cal.com/pull/8335#issuecomment-1513277529) will be implemented on the `website` repo but I wanted to get some feedback regarding security concerns before closing this PR.

zomars:
> If you think this is not a concern, I will close this PR and implement the changes in website's next.config.js.

Go for it!

emrysal:
I just looked over the proposed header add and it looks to me that it is completely safe to add said header, given it's only added when the header is missing on an asset page. 🦺

PR #8339, opened 2023-04-18
perf: use LRU cache for getServerSession

G3root:
## What does this PR do?

currently the Map object is used as a in memory-store for storing session tokens which might cause memory leaks while used on a long running servers. this PR aims to fix that

## Type of change


zomars (approved):
Nice catch thinking about non-serverless environments.

Ship it

PR #8344, opened 2023-04-18
perf: Adds PageWrapper to pages

### LINKED ISSUE #8340: [CAL-1513] Create PageWrapper
Dependencies in \_app.tsx are loaded in the API endpoints even when they are not needed. The goal of the AppWrapper is to avoid this by ensuring that these dependencies are only loaded by NextJs pages.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1513](https://linear.app/calcom/issue/CAL-1513/create-appwrapper)</sub>

roae:
## What does this PR do?

Dependencies in _app.tsx are loaded in the API endpoints even when they are not needed. The goal of the PageWrapper is to avoid this by ensuring that these dependencies are only loaded by NextJs pages.

Fixes #8340 

## Type of change


apps/web/pages/_app.tsx
```diff
@@ -1,101 +1,13 @@
...
-          </LicenseRequired>
-        ) : (
-          <Component {...pageProps} err={err} />
-        ),
-        router
-      )}
-    </AppProviders>
-  );
+  const { Component, pageProps } = props;
+  if (Component.PageWrapper !== undefined) return Component.PageWrapper(props);
```
sean-brydon, line 9: Great PR :) 

Would it not be better to always use PageWrapper unless explicitly told not to? 

This would be a very close to the approach we do for isThemeSupported now -> Always supported unless set to false

sean-brydon: Edit: I see why we may have went with this approach - due to it being on api pages. Maybe disregard this comment :P

keithwillcode: Yeah I've seen this wrapper technique used before which is why we decided to test this out. Even though Vercel is adding a feature flag to disable preloads of app and document, this helps in general to unbloat the API routes.

From a maintenance/remembering to put the PageWrapper on new pages, @sean-brydon, do you see any issues with this approach?

hariombalhara: What should I expect if I forget to add PageWrapper to a page route component?

I would have said to add a linter rule to ensure it's added. But if Next.js is adding this flag then it's a temporary change it seems and after that we can remove PageWrapper?

hariombalhara:
Something seems to be increasing the bundle size a lot.
Screenshot 2023-04-18 at 6 16 23 PM

keithwillcode:
> Something seems to be increasing the bundle size a lot. Screenshot 2023-04-18 at 6 16 23 PM

Likely due to global bundle size decreasing
Screen Shot 2023-04-18 at 10 17 53 AM

hariombalhara:
Yeah, that makes sense. So, it should be _app.tsx imports only that were global earlier and now loaded per route 👍

apps/web/pages/_app.tsx
hariombalhara: Screenshot 2023-04-18 at 7 10 17 PM
Okay, so forgetting to add it is a crash for the route. It was missing for single event-type page. I have added it.

hariombalhara (approved):
Tested. Looks good to me 🙏 Only found one route that was missing PageWrapper, fixed that.

keithwillcode:
Interesting thing to note is that the bundle size for API routes in Vercel didn't change. Still 31MB

apps/web/pages/_app.tsx
keithwillcode: Nice find. Yeah basically any pages without this will crash hard.

roae:
> Interesting thing to note is that the bundle size for API routes in Vercel didn't change. Still 31MB

@keithwillcode yeah, the bundle remains the same, to reduce it we would have to eliminate code or remove dependencies, the important thing is to reduce the amount of code that is loaded in each endpoint or page, in the experiments we saw that each import costs a lot in the cold starts. Right now we are making unnecessary imports as in this case the app dependencies are imported into the api endpoints. 

The important thing in this PR is firstLoad of the API endpoints is reduced. before every API Endpoint has 192kb in first load, this PR reduces it by almost 50%


imagen

PR #8355, opened 2023-04-18
feat: Add copy invite link

### LINKED ISSUE #6581: [CAL-837] add "Copy invite link" to our invite dialog (like slack)
[d92a9455-9484-4193-8b51-38de5260cdd5](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/0bbf5b65-c90d-4154-b0d6-0497c4e8a018/d92a9455-9484-4193-8b51-38de5260cdd5)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-837](https://linear.app/calcom/issue/CAL-837/add-copy-invite-link-to-our-invite-dialog-like-slack)</sub>

ShaneMaglangit:
## What does this PR do?

- [x] Adds an option to create a shared invite link. Limited to Member role for security.
- [x] Automatically send team invite to the user who accesses the invite link.

Fixes #6581
/claim #6581

Code Creation: https://www.loom.com/share/98d349beeeac4e16bb4dbe4583cff874
Open Invite Link: https://www.loom.com/share/18e33d78ba73436f9ce17ff49df1ad74

## Type of change

## How should this be tested?

- [x] On an account with team, create an invite link.
- [x] On another account, access the generated link.


packages/prisma/schema.prisma
```diff
@@ -233,6 +233,15 @@ model Team {
   brandColor          String           @default("#292929")
   darkBrandColor      String           @default("#fafafa")
   verifiedNumbers     VerifiedNumber[]
+  invite              Invite?
+}
+
+model Invite {
```
zomars, line 239: Let's be more explicit here:

```suggestion
model TeamInvite {
```

packages/trpc/server/routers/viewer/teams.tsx
```diff
@@ -735,4 +739,100 @@ export const viewerTeamsRouter = router({
...
+    .input(
+      z.object({
+        code: z.string(),
+      })
+    )
+    .mutation(async ({ ctx, input }) => {
+      const { code } = input;
+      await ctx.prisma.invite.delete({ where: { code } });
+    }),
+  addMemberByLink: authedProcedure
```
zomars, line 791: Why can't we reuse the same procedure (`inviteMember`) to handle both email and code invites? As far as I know both generate a code to accept the invitation. Why do we need another procedure?

zomars (changes requested):
Thank you for your contribution 🙏 I have some observations.

- The current code allows any logged in user to createInvite/setInviteExpiration/deactivateInvite/addMemberByLink for ANY team. That's a huge security concern.
- Can we reuse the `inviteMember` procedure as much as possible? We already ask the team admin if he wants to send the invite via email. We could add the option here to just get the link here.
- If we absolutely need to add a new Invite model to the DB then go for it (just rename it to TeamInvite). BUT I would like to reuse `VerificationToken` as much as possible which can be modified to serve the same purpose.

packages/trpc/server/routers/viewer/teams.tsx
ShaneMaglangit: @zomars If I am not mistaken:

- `inviteMembers` assumes we know who the invitee is. It preemptively creates an membership request for the specified email prior to notifying the user. The link it generates is nothing more than a link to the teams page.

- `addMemberByLink` doesn't have the same assumption. A membership request is created only during the time where the page is accessed given the code.

ShaneMaglangit: @zomars Would you prefer if we instead seperate `addMemberByLink` logic into two queries? Such that we would have `getInvite` that finds an invite/team based on the code and use `inviteMembers` if any is found?

We would be able to reuse inviteMembers but we would do two queries instead.

ShaneMaglangit:
@zomars can you give me a bit more context on what is `VerificationToken` being used for? Been looking at `/signup` page and it seems like its used for registration instead.

zomars:
> @zomars can you give me a bit more context on what is `VerificationToken` being used for? Been looking at `/signup` page and it seems like it's only used for registration instead.

It is also used for team invites.

https://github.com/calcom/cal.com/blob/d3f1f8b906a8588c0f5a7998b831ee44a70dafeb/packages/trpc/server/routers/viewer/teams.tsx#L339-L345

> If we're going to reuse it, we would need to add a new metadata for the expiration (or modify it atleast). VerificationToken currently have expiresAt as a non-nullable field. Team invites must have a "never expire" option. We would have to make it nullable or add a new field (expireInDays). Wdyt?

I'm ok with updating it to fit both cases.

PeerRich:
should we rename "Deactivate" to "Delete"? basically you delete the link, no? we already have "delete" as a string, too

PeerRich:
just to confirm: this also works if the person has no account already

ShaneMaglangit:
> just to confirm: this also works if the person has no account already

Yup, my fundamental assumption (which I tested prior), is that when accessing a protected route with unauthenticated. You will be redirected to the signup page with a corresponding callback link (invite link).

PeerRich:
> > just to confirm: this also works if the person has no account already
> 
> Yup, my fundamental assumption (which I tested prior), is that when accessing a protected route while unauthenticated. You will be redirected to the signup page with a corresponding callback link (invite link).
> 
> If we somehow lose the invite link with code because of some auth flow process, the user can always revisit the link.
> 
> That, I think, should sufficiently handle this case, no?

awesome, yes!

PeerRich:
@ShaneMaglangit has the blocking feedback been addressed?
https://github.com/calcom/cal.com/pull/8355#pullrequestreview-1391087410

ShaneMaglangit:
> @ShaneMaglangit has the blocking feedback been addressed? [#8355 (review)](https://github.com/calcom/cal.com/pull/8355#pullrequestreview-1391087410)

I believe so

> The current code allows any logged in user to createInvite/setInviteExpiration/deactivateInvite/addMemberByLink for ANY team. That's a huge security concern.

- [x] Added check to see if current user is admin/owner [Commit #1](https://github.com/calcom/cal.com/pull/8355/commits/cec198df68d36a3bab03d485220341e577a0f501), [Commit #2](https://github.com/calcom/cal.com/pull/8355/commits/8d27db9f6bc2ed2652907b34511d6d5b7b37db3a)

> Can we reuse the inviteMember procedure as much as possible? We already ask the team admin if he wants to send the invite via email. We could add the option here to just get the link here.

- [x] inviteMember has a slightly different logic hence we can't repurpose it for this. [Explanation](https://github.com/calcom/cal.com/pull/8355#discussion_r1170700223)

> If we absolutely need to add a new Invite model to the DB then go for it (just rename it to TeamInvite). BUT I would like to reuse VerificationToken as much as possible which can be modified to serve the same purpose.

- [x] Modified and reused VerificationToken for invite token [Commit](https://github.com/calcom/cal.com/pull/8355/commits/f40eacdefd352e54e8ded731a9adf0aaba7babda)

ShaneMaglangit:
@PeerRich Oh PR got stale, do we still have issues to work on for this?

zomars (changes requested):
Blocking to update the new trpc structure

zomars (approved):
Updates tRPC procedures to the new format. 

LGTM

PeerRich:
hey @ShaneMaglangit would you be down to fix the recent merge conflicts? should be easy.

if yes, lemme add $100 to the bounty

ShaneMaglangit:
> hey @ShaneMaglangit would you be down to fix the recent merge conflicts? should be easy.
> 
> if yes, lemme add $100 to the bounty

Oh hey :wave: Sure thing, I'll have a look at this on the weekend. Just a bit busy with work rn.

PeerRich:
thank you!

zomars (changes requested):
Fixed faulty translations that broke e2e tests. But there's now another failure with App Store tests. Blocking to check.

zomars (approved):
Tested and seems like it's good to go! Thank you for your contribution 🙏 

Ship it

PR #8362, opened 2023-04-18
perf: improve bookings performance

anonrig:
Couple of improvements to avoid calling and creating unnecessary async contexts.


zomars (approved):
LGTM

PR #8363, opened 2023-04-18
perf: replace `cloneDeep` with `structuredClone`

anonrig:
`structuredClone` is available on both browser and server for faster cloning.

Ref: https://caniuse.com/?search=structuredClone


zomars (approved):
Nice! Thank you for your contribution 🙏

hariombalhara (approved):

hariombalhara (dismissed):
This is breaking for Node.js 16(which I have been using and tests also use it seems as they are failing). Checked package.json and we seem to be requiring Node.js > =18 but that doesn't look like a hard requirement as it's been working for me with Node.js 16
Screenshot 2023-04-19 at 12 13 17 PM

Others might also be still using old Node.js version?


Blocking merge till I get the clarification from @zomars maybe.

zomars:
> This is breaking for Node.js 16(which I have been using and tests also use it seems as they are failing). Checked package.json and we seem to be requiring Node.js > =18 but that doesn't look like a hard requirement as it's been working for me with Node.js 16 Screenshot 2023-04-19 at 12 13 17 PM
> 
> Others might also be still using old Node.js version?
> 
> Blocking merge till I get the clarification from @zomars maybe.

We dropped v16 support quite a while ago. There might be some remnants still, but we don't officially support it anymore.

zomars:
For some reason this change is breaking booking pages:

https://github.com/calcom/cal.com/actions/runs/4747262818/jobs/8432465703?pr=8363

🤔

hariombalhara:
I believe tests are still on Node16 or something.

zomars:
> I believe tests are still on Node16 or something.

I don't think that's it: https://github.com/calcom/cal.com/blob/986f9dca8dff9c900a1ae1572243492b52296eaf/.github/actions/yarn-install/action.yml#L17

joeauyeung:
> > > I believe tests are still on Node16 or something.
> > 
> > 
> > I don't think that's it:
> > https://github.com/calcom/cal.com/blob/986f9dca8dff9c900a1ae1572243492b52296eaf/.github/actions/yarn-install/action.yml#L17
> 
> There is a function in evt object it seems that `structuredClone` doesn't support functions and throws error Screenshot 2023-04-21 at 12 39 41 PM Screenshot 2023-04-21 at 12 39 52 PM Screenshot 2023-04-21 at 12 41 34 PM

If this is the case, should we close this PR for now?

PR #8379, opened 2023-04-19
Fixes editor not saving new value

CarinaWolli:
## What does this PR do?

Same fixes as #8368 just for everywhere else where editor is used


CarinaWolli:
Maybe we can remove the need for the firstRender state but for now, we should merge that fix

hariombalhara (approved):

PR #8382, opened 2023-04-19
Shaves off an easy 1.4kb win from the global package

emrysal:
## What does this PR do?

Allows tree shaking of React in the feature flags provider 🙏


keithwillcode (approved):
Nice work.

PR #8385, opened 2023-04-19
Add event type redirect URL to Stripe callback

joeauyeung:
## What does this PR do?

This PR adds the event type redirect URL if enabled to the Stripe `confirmParams`

Fixes # (issue)

## Type of change

## How should this be tested?

- Create an event type with Stripe enabled and a redirect URL
- Create a booking, it should redirect you to the redirect URL
    - Do this for both charge on booking & charging no-show fee
- Go through the same booking flow again but this time disable redirect URL

## Checklist


joeauyeung (commented):
@hariombalhara do you think we still need `bookingSuccessRedirect` if Stripe forces a redirect after a success?

zomars (approved):
LGTM

PR #8396, opened 2023-04-20
Fixing default event location URL for dynamic events

sean-brydon:
## What does this PR do?

https://www.loom.com/share/ef27b7edb2554fbf934ec80aa01d8df9

## Type of change

## How should this be tested?
- [ ] Set a default location other than cal.video (Link apps)
- [ ] Create a booking

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -686,7 +686,6 @@ async function handler(
     const app = getAppFromSlug(firstUsersMetadata?.defaultConferencingApp?.appSlug);
     locationBodyString = app?.appData?.location?.type || locationBodyString;
     defaultLocationUrl = firstUsersMetadata?.defaultConferencingApp?.appLink;
-    console.log({ dynamicUserList, locationBodyString, defaultLocationUrl });
   }
 
   if (eventType && eventType.hasOwnProperty("bookingLimits") && eventType?.bookingLimits) {
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -755,6 +755,15 @@ async function handler(
 
   const [organizerUser] = users;
   const tOrganizer = await getTranslation(organizerUser?.locale ?? "en", "common");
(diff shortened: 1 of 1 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -794,7 +794,11 @@ async function handler(
   const seed = `${organizerUser.username}:${dayjs(reqBody.start).utc().format()}:${new Date().getTime()}`;
   const uid = translator.fromUUID(uuidv5(seed, uuidv5.URL));
 
-  const bookingLocation = getLocationValueForDB(locationBodyString, eventType.locations);
+  let bookingLocation = getLocationValueForDB(locationBodyString, eventType.locations);
```
sean-brydon, line 797: getLocationValueForDB returns locationBodyString if no location is found in the event type

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -794,7 +794,11 @@ async function handler(
...
+
+  if (dynamicUserList.length > 1 && defaultLocationUrl) {
+    bookingLocation = defaultLocationUrl;
+  }
```
sean-brydon, line 801: So if there dynamic users -> re set the bookingLocation to the defaultLocationUrl which is fetched above

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara, line 797: I think we would be able to co-locate location calculation related code like this
Screenshot 2023-04-20 at 4 09 10 PM

sean-brydon: I see! Good shout! Just having lunch will implement this 🙏

hariombalhara: Also, at line 782 in my screenshot, why don't we directly set locationBodyString equal to the defaultLocationUrl. THen we won't need the below new code.

sean-brydon: Yeah just testing and that was the only issue it seems

hariombalhara (approved):

PR #8417, opened 2023-04-20
Show team invites regardless of user plan

zomars:
## What does this PR do?

Customer got confused on why round robin event  wasn't working and it turned out that some member didn't accept the team invites yet.

This fix shows team invites regardless if the user already have a paid feature or not.

## Type of change

## How should this be tested?

- [ ] Receive an invite as a free user
- [ ] Go to /teams and the invite should be visible

image


emrysal (approved):
LGTM

PR #8419, opened 2023-04-20
perf: tRPC procedures and middleware refactor

### LINKED ISSUE #8342: [CAL-1514] Set the user session only on the endpoints where it is needed.
Currently the session is set in the default context of tRPC, this adds the session dependencies to all endpoints. Perhaps, the getServerSession function should be run only in AuthedProcedure, or add SessionProcedure that is in charge of setting the session in the context, in the endpoints where it is necessary.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1514](https://linear.app/calcom/issue/CAL-1514/set-the-user-session-only-on-the-endpoints-where-it-is-needed)</sub>

roae:
## What does this PR do?

Remove the session and the i18n object from the default tRPC context, to add it in the procedures that are needed, and prevent endpoints that don't need to access the user session from loading unwanted dependencies.

The procedures and middleware were put in separate files, so  they are imported only where needed.

Right now this will not have a big impact on performance, but when tRPC is split it will benefit from this.

Fixes #8342 

## Type of change


apps/web/middleware.ts
hariombalhara, line 1: Middleware is still needed.

apps/web/pages/_document.tsx
```diff
@@ -17,12 +17,17 @@ function setHeader(ctx: NextPageContext, name: string, value: string) {
...
-      setHeader(ctx, "x-csp", "initialPropsOnly");
+    try {
+      if (!process.env.CSP_POLICY) {
+        setHeader(ctx, "x-csp", "not-opted-in");
+      } else if (!ctx.res?.getHeader("x-csp")) {
+        // If x-csp not set by gSSP, then it's initialPropsOnly
+        setHeader(ctx, "x-csp", "initialPropsOnly");
+      }
+    } catch (e) {
+      console.log((e as Error).message);
```
hariombalhara, line 28: I don't think we need this try-catch. `res.setHeader` can throw error which is already try catched by `setHeader` function. getHeader shouldn't throw any error

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -57,7 +57,7 @@ const GeneralQueryView = () => {
   }
   return (
     <WithQuery
-      success={({ data }) => <GeneralView user={user} localeProp={data.locale} />}
+      success={({ data }) => <GeneralView user={user} localeProp={`${data.locale}`} />}
```
hariombalhara, line 60: I don't see any difference.

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -591,13 +591,13 @@ async function handler(req: CustomRequest) {
...
     // Posible to refactor TODO:
-    const paymentApp = await appStore[paymentAppCredential?.app?.dirName as keyof typeof appStore];
+    const paymentApp = await appStore[paymentAppCredential?.app?.dirName as keyof typeof appStore]();
     if (!(paymentApp && "lib" in paymentApp && "PaymentService" in paymentApp.lib)) {
       console.warn(`payment App service of type ${paymentApp} is not implemented`);
       return null;
     }
 
-    const PaymentService = paymentApp.lib.PaymentService;
+    const PaymentService = paymentApp.lib.PaymentService as any;
```
hariombalhara, line 600: Do we really need this type assertion?

hariombalhara (dismissed):
Left a few comments. Partially done with the review. Will continue later.

roae:
> Left a few comments. Partially done with the review. Will continue later.

Thanks @hariombalhara, by mistake I made commit with changes that should not be uploaded, I will clean up the branch 🙏🏼

packages/features/bookings/lib/handleCancelBooking.ts
roae: Yep, I think we must create a generic PaymentService type at this point  the PaymentService attributes is unknown, thoughts?

packages/app-store/routing-forms/trpc/_router.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // Unreachable code but required for type safety
+    if (!UNSTABLE_HANDLER_CACHE.deleteForm) {
+      throw new Error("Failed to load handler");
+    }
+
+    return UNSTABLE_HANDLER_CACHE.deleteForm({ ctx, input });
+  }),
+
+  report: authedProcedure.input(ZReportInputSchema).query(async ({ ctx, input }) => {
```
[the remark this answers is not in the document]
roae: What do you mean?

apps/web/next.config.js
```diff
@@ -133,7 +134,10 @@ const nextConfig = {
   images: {
     unoptimized: true,
   },
-  webpack: (config, { webpack, buildId }) => {
+  webpack: (config, { webpack, buildId, isServer }) => {
+    if (isServer) {
```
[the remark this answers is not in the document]
zomars: Agreed. Let's put this behind an env variable. Maybe `NEXT_PUBLIC_DEBUG=1`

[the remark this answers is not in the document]
roae: No, I deleted it, it wasn't supposed to get here, yet.

packages/app-store/routing-forms/trpc/_router.ts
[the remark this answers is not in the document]
hariombalhara: I mean to avoid repeating cache reading and setting logic. I think something like following could work

```javascript
const cachedResponse = (key, importer)=>{
    if (!UNSTABLE_HANDLER_CACHE[key]) {
       UNSTABLE_HANDLER_CACHE[key] = importer();
    }
    if (!UNSTABLE_HANDLER_CACHE[key]) {
       throw new Error("Failed to load handler");
    }
    return UNSTABLE_HANDLER_CACHE[key]({ ctx, input });
}

const appRoutingForms = {
   response: publicProcedure.input(ZResponseInputSchema).mutation(async ({ ctx, input }) => {
       return cachedImport('response', ()=>await import("./response.handler")).responseHandler)
   })
}
```

packages/features/bookings/lib/handleCancelBooking.ts
hariombalhara: Yeah, I think that would be the right approach. We do something similar for  Calendar.
Screenshot 2023-04-27 at 3 14 46 PM

packages/app-store/routing-forms/trpc/_router.ts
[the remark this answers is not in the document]
roae: Ohh yes good idea @hariombalhara, it could work, but I think we can do it in another pr.

packages/features/bookings/lib/handleCancelBooking.ts
hariombalhara: @roae I don't understand why TypeScript isn't able to infer types for PaymentService, it is still able to do it for CalendarService. For now, I have asserted the type on the instance of PaymentService so that we can call it's methods safely.

packages/app-store/index.ts
```diff
@@ -1,33 +1,35 @@
...
-  facetime: import("./facetime"),
-  sylapsvideo: import("./sylapsvideo"),
+  // example: (): Promise<typeof import("./example")> => import("./example"),
+  applecalendar: (): Promise<typeof import("./applecalendar")> => import("./applecalendar"),
+  caldavcalendar: (): Promise<typeof import("./caldavcalendar")> => import("./caldavcalendar"),
+  closecom: (): Promise<typeof import("./closecom")> => import("./closecom"),
+  dailyvideo: (): Promise<typeof import("./dailyvideo")> => import("./dailyvideo"),
+  googlecalendar: (): Promise<typeof import("./googlecalendar")> => import("./googlecalendar"),
+  googlevideo: (): Promise<typeof import("./googlevideo")> => import("./googlevideo"),
+  hubspot: (): Promise<typeof import("./hubspot")> => import("./hubspot"),
```
emrysal, line 9: This can also be
```suggestion
  hubspot: () => import("./hubspot"),
```

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -37,6 +38,7 @@ export default trpcNext.createNextApiHandler({
    * @link https://trpc.io/docs/caching#api-response-caching
    */
   responseMeta({ ctx, paths, type, errors }) {
+    console.log("responseMeta method");
```
Comment (emrysal) [inline apps/web/pages/api/trpc/[trpc].ts:41]: Some debugs remain

emrysal (dismissed):

packages/app-store/index.ts
roae: > This can also be
yes, but the types are not being inferred, I tried that but got the same result, so yes.

keithwillcode (commented):
This is an awesome refactor and the speed improvements are big. Great work!

One thing I noticed: The `midlewares` folder is missing a `d`. It should be `middlewares`.

keithwillcode (approved):

PR #8420, opened 2023-04-20
perf: Add log messages on cold/hot requests

roae:
## What does this PR do?

We want to measure the number of cold versus hot requests made, so add a console.log to the middleware to identify these requests and report on them in Axiom.


apps/web/middleware.ts
```diff
@@ -7,9 +7,16 @@ import { CONSOLE_URL, WEBAPP_URL, WEBSITE_URL } from "@calcom/lib/constants";
...
 
+let cold = true;
+
 const middleware: NextMiddleware = async (req) => {
   const url = req.nextUrl;
+  const requestHeaders = new Headers(req.headers);
 
+  // This console.log is required to create a report in axios for hot and cold requests
+  console.log(cold ? "Cold Start" : "Hot Start");
+  requestHeaders.set("x-cal-cold-start", cold ? "true" : "false");
```
roae, line 18: This helps us to identify if a request was executed cold.

apps/web/middleware.ts
```diff
@@ -7,9 +7,16 @@ import { CONSOLE_URL, WEBAPP_URL, WEBSITE_URL } from "@calcom/lib/constants";
...
 import { extendEventData, nextCollectBasicSettings } from "@calcom/lib/telemetry";
 
+let cold = true;
+
 const middleware: NextMiddleware = async (req) => {
   const url = req.nextUrl;
+  const requestHeaders = new Headers(req.headers);
 
+  // This console.log is required to create a report in axios for hot and cold requests
+  console.log(cold ? "Cold Start" : "Hot Start");
```
keithwillcode, line 17: Instead of writing another new console line to axiom on every request (which will be a lot), do we have a mechanism to store log variables in some sort of dictionary and then write them all at once into a single message? cc @zomars @emrysal

keithwillcode (approved):
Let's get merged and see logs in Axiom and then decide if we want to merge them.

PR #8435, opened 2023-04-21
Fix workflow not firing for event types without video app as location

CarinaWolli:
## What does this PR do?

Fixes that workflows without a videoCallUrl didn't trigger

## Type of change

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -1503,7 +1503,7 @@ async function handler(
     // Here we should handle every after action that needs to be done after booking creation
 
     // Obtain event metadata that includes videoCallUrl
-    const metadata = { videoCallUrl: evt.videoCallData?.url };
+    const metadata = evt.videoCallData?.url ? { videoCallUrl: evt.videoCallData?.url } : undefined;
     try {
       await scheduleWorkflowReminders({
         workflows: eventType.workflows,
@@ -2119,7 +2119,7 @@ async function handler(
     log.error("Error while creating booking references", error);
   }
 
-  const metadataFromEvent = { videoCallUrl };
+  const metadataFromEvent = videoCallUrl ? { videoCallUrl } : undefined;
 
   try {
     await scheduleWorkflowReminders({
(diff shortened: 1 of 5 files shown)


CarinaWolli:
fixing type errors

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -187,7 +186,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
           location: reminder.booking.location || "",
           additionalNotes: reminder.booking.description,
           responses: responses,
-          meetingUrl: bookingMetadataSchema.parse(reminder.booking.metadata || {})?.videoCallUrl,
```
hariombalhara, line 190: How does it fix the error specifically?
Also, can we not simply make it`bookingMetadataSchema.parse(reminder.booking.metadata || {})?.videoCallUrl || ""`

hariombalhara: Screenshot 2023-04-21 at 3 09 29 PM
Also, this might make the default as empty string everywhere.

CarinaWolli: > Screenshot 2023-04-21 at 3 09 29 PM
> 
> Also, this might make the default as empty string everywhere.

I tried that, but it's still throwing zod error "Expected string, received null"

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1503,7 +1503,7 @@ async function handler(
     // Here we should handle every after action that needs to be done after booking creation
 
     // Obtain event metadata that includes videoCallUrl
-    const metadata = { videoCallUrl: evt.videoCallData?.url };
+    const metadata = evt.videoCallData?.url ? { videoCallUrl: evt.videoCallData?.url } : undefined;
```
hariombalhara, line 1506:
```suggestion
    const metadata = evt.videoCallData?.url ? { videoCallUrl: evt.videoCallData.url } : undefined;
```

hariombalhara (approved):

PR #8440, opened 2023-04-21
Hotfix: Prevent looping b/w dark and light

hariombalhara:
Found the root cause of the theme flickering issue.

`booking-theme` is the storage key for Booking Page-1 and Booking Page-2. Infact it's same for all booking pages.
Scenario:
Booking Page-1 decides that it needs dark theme and tries to update `booking-theme` in localStorage to dark. 
If Booking Page-2 decides that it needs light theme, it would try to update `booking-theme` in localStorage to light.

There is `storage` event listener that can detect on Booking-Page-1 that someone outside(here Booking Page-2) has changed the `booking-theme` in localStorage and it updates it's theme to that. But, our logic in useTheme doesn't agree with it and tries to revert it.

### DIFF
--- a/packages/lib/hooks/useTheme.tsx
+++ b/packages/lib/hooks/useTheme.tsx
@@ -15,16 +15,17 @@ export default function useTheme(themeToSet?: Maybe<string>) {
 
   useEffect(() => {
     // If themeToSet is not provided the purpose is to just return the current the current values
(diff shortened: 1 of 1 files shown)


sean-brydon (approved):
Tested on preview and no longer getting the flash 

Good fix :)

PR #8450, opened 2023-04-21
Use avatar.png instead of base64.

emrysal:
## What does this PR do?

This fixes a booking page crash for big teams as it includes the avatars as base64 from different team members.

### DIFF
--- a/packages/lib/getEventTypeById.ts
+++ b/packages/lib/getEventTypeById.ts
@@ -37,7 +37,6 @@ export default async function getEventTypeById({
     name: true,
     username: true,
     id: true,
-    avatar: true,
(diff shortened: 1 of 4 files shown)


packages/lib/getEventTypeById.ts
```diff
@@ -37,7 +37,6 @@ export default async function getEventTypeById({
     name: true,
     username: true,
     id: true,
-    avatar: true,
```
emrysal, line 40: For team members this was patched already, but it still returned the avatar in the background, likely slowing down Prisma and data calls.

roae (approved):
Looks good! 🚀

PR #8451, opened 2023-04-21
Fix booking page when toggling payment app settings

joeauyeung:
## What does this PR do?

Replaces getting `eventType.price` with data from `getPaymentAppData` in: 
- `BookingListItem`
- `EventTypeDescription`

Bookings with different payment options all under the same event type
image

Fixes # (issue)

## Type of change

## How should this be tested?

- Create an event type and install Stripe but do not enable
    - Create a booking
- Now enable Stripe and collect the booking on payment
    - Create another booking
- Now change to collect no-show fee
    - Create a booking
- Go to the  booking page
    - All bookings should appear as expected
- Disable Stripe on that event type and go back to the event type page. The price badge should be gone now

## Checklist


apps/web/components/booking/BookingListItem.tsx
```diff
@@ -413,7 +417,7 @@ function BookingListItem(booking: BookingItemProps) {
               {title}
               <span> </span>
 
-              {!!booking?.eventType?.price && !booking.paid && (
+              {paymentAppData.enabled && !booking.paid && booking.payment.length && (
```
joeauyeung, line 420: Only show the awaiting payment badge if payments are enabled and there was a payment associated with that booking

packages/trpc/server/routers/viewer/bookings.tsx
```diff
@@ -365,6 +367,7 @@ export const bookingsRouter = router({
           eventType: {
             ...booking.eventType,
             recurringEvent: parseRecurringEvent(booking.eventType?.recurringEvent),
+            metadata: EventTypeMetaDataSchema.parse(booking.eventType?.metadata || {}),
```
joeauyeung, line 370: When querying for bookings, we need something in the metadata field of the booking for `getPaymentAppData`

packages/trpc/server/routers/viewer/eventTypes.ts
```diff
@@ -647,27 +646,6 @@ export const eventTypesRouter = router({
...
-          key: true,
-        },
-      });
-
-      if (paymentCredential?.type === "stripe_payment") {
-        const { default_currency } = stripeDataSchema.parse(paymentCredential.key);
-        data.currency = default_currency;
-      }
-    }
-
```
joeauyeung, line 670: Since we rely on payment app data, I don't think we need this as a part of the event type update function. 

Payment apps aren't working on team event types (yet) and the only place we use `eventType.price` is when converting legacy event types.

zomars (approved):
Nice job @joeauyeung LGTM 🙏🏽

zomars (approved):

PR #8453, opened 2023-04-21
Hotfix - CalDAV/Wraps timezone addition in try catch

alishaz-polymath:
## What does this PR do?

Sometimes ICAL contains TZID in non-standard format which isn't supported by iCalendar 2.0 in which case there's a crash of getAvailability. This PR wraps the timezone creation in `try-catch` as a hotfix for that issue. I'll look into a potential follow up to see if it's possible to somehow convert those non-standard formats into IANA time zone as expected by iCalendar 2.0

## Type of change

## How should this be tested?

- The availability shouldn't be failing due to one bad TZID value in ICAL


packages/lib/CalendarService.ts
```diff
@@ -323,20 +323,25 @@ export default abstract class BaseCalendarService implements Calendar {
...
+            const tzoffsetto = dayjs(event.endDate.toJSDate()).tz(tzid).format("Z");
+
+            // set timezone offset
+            standard.addPropertyWithValue("tzoffsetfrom", tzoffsetfrom);
+            standard.addPropertyWithValue("tzoffsetto", tzoffsetto);
+            // provide a standard dtstart
+            standard.addPropertyWithValue("dtstart", "1601-01-01T00:00:00");
+            timezoneComp.addSubcomponent(standard);
+            vcalendar.addSubcomponent(timezoneComp);
+          } catch (e) {
```
emrysal, line 342:
```suggestion
          } catch (e) {
            // Adds try-catch to ensure the code proceeds when Apple Calendar provides non-standard TZIDs 
```

emrysal (approved):

PR #8455, opened 2023-04-22
Possible fix for iOS redirection

### LINKED ISSUE #8452: Redirect doesn't work on mobile
Redirect to external URL is not working on Safari on iOS 16.4.1. Instead, the Cal.com success page shows up.

To reproduce:
1. Visit https://themindoasis.in/book-a-session/ on your phone
2. Click "Book a session" and then select any session type
3. Select a random date and fill in meeting details
4. Hit "Confirm"

Expected behaviour:
It should redirect to https://themindoasis.in/thank-you/ (only works on Desktop)

Current behaviour:
It lands on the Cal.com success page.

hariombalhara:
## What does this PR do?

Fixes #8452
On iOS it seems like location.href is ignored if it is followed by router.push which is probably doing a soft navigation.
When I put a breakpoint to debug it works. So, this is a potential fix for the issue.

**Environment**:  Production

## Type of change

## How should this be tested?

- [x] When you book a meeting with "Redirect on booking" enabled on iOS the redirect doesn't happen.

## Checklist

### DIFF
--- a/packages/lib/bookingSuccessRedirect.ts
+++ b/packages/lib/bookingSuccessRedirect.ts
@@ -24,6 +24,7 @@ export const bookingSuccessRedirect = async ({
 
     // Using parent ensures, Embed iframe would redirect outside of the iframe.
     window.parent.location.href = url.toString();
+    return;
   }
   return router.push({
     pathname: `/booking/${bookingUid}`,


packages/lib/bookingSuccessRedirect.ts
```diff
@@ -24,6 +24,7 @@ export const bookingSuccessRedirect = async ({
 
     // Using parent ensures, Embed iframe would redirect outside of the iframe.
     window.parent.location.href = url.toString();
+    return;
```
hariombalhara, line 27: It was a missing return statement.
location.href doesn't immediately abort the control flow and it is followed by router.push. It seems to work on android and mac os but location.href get's ignored in iOS.

emrysal (approved):

PR #8469, opened 2023-04-23
fix: 7040 fix - team profile description

### LINKED ISSUE #7040: Different Page formatting on Profile and Team page
I noticed that there is a formatting difference on Profile compare to Team page view, which is  not really nice. 

On the user profile page the description is full-width which looks nice as its the same width as the event-types under it:

image

On the team profile page the description is in a block and about 50% the width of the event-types, which looks quite ugly:

image

Please make both profile and team description on profile, the same width or 90% of the event-types.

Hallidayo:
## What does this PR do?

Fixes #7040

You will see that the user profile page and team profile page description width does not match. This is the current styling:

Screenshot 2023-04-23 at 09 10 34

Changed to:

Screenshot 2023-04-23 at 09 13 24

 I noticed that both the user and team profile page styling does not match e.g. top margin, image size etc, Do these need to be changed as well so both styling on the pages are the same? If so I get get these changed as well if needed?

Also some duplicate tailwind classes have been removed. 

## Type of change

### DIFF
--- a/apps/web/pages/team/[slug].tsx
+++ b/apps/web/pages/team/[slug].tsx
@@ -56,12 +56,12 @@ function TeamPage({ team, isUnpublished }: TeamPageProps) {
   }
 
   const EventTypes = () => (
-    <ul className="border-subtle border-subtle rounded-md border">
+    <ul className="border-subtle rounded-md border">
       {team.eventTypes.map((type, index) => (
         <li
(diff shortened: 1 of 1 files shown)


hariombalhara (approved):
LGTM !! Thanks @Hallidayo 🙏

PR #8487, opened 2023-04-24
Routing-Forms/External Redirect URL improvement

hariombalhara:
## What does this PR do?
- Ensure that a valid URL is entered for external redirect option.
Screenshot 2023-04-24 at 7 10 47 PM

- Redirect is done outside of embed now. It was the case with success page redirect already but not with Routing Form redirect.

**Environment**: Production

## Type of change

## How should this be tested?

- [x] Add an external redirect with invalid URL and see that an error is shown.
- [x] Check that when a valid external redirect is added, redirect takes the user outside of the embed.

## Checklist


leog (approved):
LGTM

PR #8508, opened 2023-04-25
Bookings with no-show fee to set proper location & prevent broken payments from blocking payment page

joeauyeung:
## What does this PR do?

This PR fixes a bug in which locations were not respected when bookings enabled no-show fees. Also if a payment record was malformed, it would block the bookings page from loading

Fixes # (issue)

## Type of change

## How should this be tested?

- Create an event type with a location other than Cal video & Stripe enabled set to charge a no-show fee
- Create a booking
- After entering your payment information the confirmed booking emails should contain the set location link
- In the DB delete the payment record for that booking
- Go to the bookings page. It should still load with an error message
    
image

## Checklist


apps/web/components/booking/BookingListItem.tsx
```diff
@@ -280,7 +280,7 @@ function BookingListItem(booking: BookingItemProps) {
         isOpenDialog={isOpenSetLocationDialog}
         setShowLocationModal={setIsOpenLocationDialog}
       />
-      {booking.paid && (
+      {booking.paid && booking?.payment[0] && (
```
hariombalhara, line 283:
```suggestion
      {booking.paid && booking.payment[0] && (
```

hariombalhara: We can make the similar change at other places.

packages/features/ee/payments/api/webhook.ts
```diff
@@ -335,7 +337,14 @@ const handleSetupSuccess = async (event: Stripe.Event) => {
   // If the card information was already captured in the same customer. Delete the previous payment method
 
   if (!eventTypeRaw?.requiresConfirmation) {
```
hariombalhara, line 339: Does it work correctly when Requires Confirmation is conditional? i.e. when it returns true but it actually doesn't require confirmation.


I see this logic in handleNewBooking which isn't here.
Screenshot 2023-04-25 at 1 43 12 PM

packages/features/ee/payments/api/webhook.ts
```diff
@@ -315,6 +315,8 @@ const handleSetupSuccess = async (event: Stripe.Event) => {
     bookingData.status = BookingStatus.ACCEPTED;
   }
 
+  if (booking.location) evt.location = booking.location;
```
hariombalhara, line 318: I think this could be moved to `getBooking` itself which creates evt ? Also as a follow-up maybe we should think about making the `location` a required property in `CalendarEvent` type

apps/web/components/booking/BookingListItem.tsx
hariombalhara: Shouldn't it be added with this condition?
`booking.paid && booking?.payment[0]?.paymentOption === "HOLD"
as per `https://github.com/calcom/cal.com/blob/d88a2bba0df1f08f6e2fa8c01a3b22a8cd24fd4d/apps/web/components/booking/BookingListItem.tsx#L465

Though I feel that we can simply remove this condition(which makes things complex), as Dialog won't be shown anyway unless the CTA is there to be clicked.

hariombalhara (changes requested):
Left some comments.

joeauyeung (commented):
@hariombalhara removed optional chaining when not needed and addressed other feedback

hariombalhara:
Screenshot 2023-04-26 at 4 04 46 PM
Earlier there used to be a confirm button even for bookings that were pending payment. I am not sure if that was kind of a feature(e.g. if in case someone pays through other mediums and now the organizer wants to confirm the meeting).

hariombalhara:
It is possible that an attendee would pay for the booking(that requires confirmation) and then the organizer had to reject the booking. Such bookings can maybe be shown in its dedicated place? The same is the case when it's a no-show from the organizer. The organizer may want to refund or just reschedule the meeting.

hariombalhara (approved):
Tested following scenarios:
- Enabled confirmation with time threshold of 2 hours and confirmation related email were sent.
- On successful booking **paid** badge was shown correctly
- On successful booking with 'No Show Fee', **hold** badge was shown correctly.
- A payment pending email goes in all scenarios.
- "Payment Pending badge" is shown correctly when payment is pending

Couldn't test failed payment scenario e2e but I verified the UI by modifying the DB.


Left 2 comments that might not be directly related to the PR but makes sense in general with paid bookings. Can be worked upon in some followup.

joeauyeung:
> Screenshot 2023-04-26 at 4 04 46 PM
> 
> Earlier there used to be a confirm button even for bookings that were pending payment. I am not sure if that was kind of a feature(e.g. if in case someone pays through other mediums and now the organizer wants to confirm the meeting).

I purposely removed that as I thought it was a bug. If it was a part of the design I can add that back but I would have to do some work to ensure that no-show fee bookings w/o payment methods can't collect on the no-show fee.

joeauyeung:
> It is possible that an attendee would pay for the booking(that requires confirmation) and then the organizer had to reject the booking. Such bookings can maybe be shown in its dedicated place? The same is the case when it's a no-show from the organizer. The organizer may want to refund or just reschedule the meeting.

For no-show fee bookings, no payment is made on booking so if the organizer reschedules/cancels then the attendee won't be charged. For bookings where payment is required to confirm the booking, then there is logic already in place to refund the attendee if the organizer cancels.

PR #8510, opened 2023-04-25
Don't use alphanumeric sender id for sending sms to Australian numbers

CarinaWolli:
## What does this PR do?

Unregistered Alphanumeric Sender IDs aren’t allowed in Australia anymore. It is added to the list of countries codes that don't support alphanumeric sender IDs (country code +61)

See Australia: https://support.twilio.com/hc/en-us/articles/223133767-International-support-for-Alphanumeric-Sender-ID


packages/features/ee/workflows/lib/alphanumericSenderIdSupport.ts
```diff
@@ -84,4 +84,5 @@ const noAlphanumericSenderIdSupport = [
   "+58",
   "+84",
   "+260",
+  "+61",
```
hariombalhara, line 87: Isn't there an API from Twilio that we can use and cache the response for?
It might be overkill if it's a rare change though. Also, how does it work if someone uses that alphanumeric ID and then it longer stops being supported. What happens to all such workflows.

hariombalhara (approved):

packages/app-store/apps.browser.generated.tsx
```diff
@@ -11,11 +11,11 @@ export const InstallAppButtonMap = {
   vital: dynamic(() => import("./vital/components/InstallAppButton")),
 };
 export const AppSettingsComponentsMap = {
-  "general-app-settings": dynamic(
-    () => import("./templates/general-app-settings/components/AppSettingsInterface")
+  "general-app-settings": dynamic(() =>
```
emrysal, line 14: @hariombalhara this causes a linting error?

packages/features/ee/workflows/lib/alphanumericSenderIdSupport.ts
CarinaWolli: It should then just fall back to our normal twilio number

emrysal (approved):

PR #8515, opened 2023-04-25
New booker fixes for layout toggling

JeroenReumkens:
## What does this PR do?

Before this PR when switching in the new booker back from mobile to desktop, it would stay in the mobile layout. Now this will reset to the `small_calendar` view. Next to this resizing to desktop also triggered lots of rerenders because the `onValueChange` function was a new instance every time. Added memo around that now. 

## Type of change


sean-brydon (approved):
Nice!

PR #8537, opened 2023-04-26
Pull latest main in every submodule

hariombalhara:
- It would ensure that one can quickly update all submodules to the latest so that yarn.lock is correct
- It also avoids unnecessary unstaging of changes.

### DIFF
--- a/git-setup.sh
+++ b/git-setup.sh
@@ -23,6 +23,10 @@ for module in "$@"; do
     git submodule add --force $project "apps/$module"
     # Set the default branch to main
     git config -f .gitmodules --add "submodule.apps/$module.branch" main
+
+    # Update to the latest from main in that submodule
+    cd apps/$module && git pull origin main && cd ../..
+
     # Adding the subdmoule ignores the `.gitignore` so a reset is needed
     git reset
   else
--- a/git-setup.sh
+++ b/git-setup.sh
@@ -23,12 +23,12 @@ for module in "$@"; do
     git submodule add --force $project "apps/$module"
     # Set the default branch to main
     git config -f .gitmodules --add "submodule.apps/$module.branch" main
-
+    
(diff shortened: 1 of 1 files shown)


git-setup.sh
```diff
@@ -23,6 +23,10 @@ for module in "$@"; do
     git submodule add --force $project "apps/$module"
     # Set the default branch to main
     git config -f .gitmodules --add "submodule.apps/$module.branch" main
+
+    # Update to the latest from main in that submodule
+    cd apps/$module && git pull origin main && cd ../..
```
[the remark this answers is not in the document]
JeroenReumkens: What would happen if you're on a different git branch for a given submodule? I think it will simply error, right? Which might be okay. Mainly hoping it wouldn't discard any open changes here.

git-setup.sh
```diff
@@ -23,8 +23,12 @@ for module in "$@"; do
...
     # Set the default branch to main
     git config -f .gitmodules --add "submodule.apps/$module.branch" main
-    # Adding the subdmoule ignores the `.gitignore` so a reset is needed
-    git reset
+    
+    # Update to the latest from main in that submodule
+    cd apps/$module && git pull origin main && cd ../..
+
+    # We forcefully added the subdmoule which was in .gitignore, so unstage it.
+    git restore --staged apps/$module
```
hariombalhara, line 31: `git reset` unstages all staged changes. This just undoes what we did.

git-setup.sh
```diff
@@ -35,3 +39,4 @@ for module in "$@"; do
     fi
   fi
 done
+git restore --staged .gitmodules
```
hariombalhara, line 42: Adding a git submodule also adds this file which is in .gitignore, unstage this as well.

zomars (approved):
Let's try this for sure!

PR #8538, opened 2023-04-26
Allow booking date overrides outside of working hours

### LINKED ISSUE #8258: [CAL-1489] Date override doesn't work for a day that isn't originally available.
Date override doesn't work for a day that isn't originally available.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1489](https://linear.app/calcom/issue/CAL-1489/message-from-sarah-katy)</sub>

CarinaWolli:
## What does this PR do?

Currently, date overrides outside of working hours are correctly shown as available but when submitting the booking it is throwing `Error: No available users found`: 

Screenshot 2023-04-26 at 13 45 33

Fixes #8258

## Type of change

## How should this be tested?

- Create a date override that is outside of your working hours
- Try to book that slot
- See that it works

## Checklist

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -64,7 +64,7 @@ import {
 import type { BufferedBusyTime } from "@calcom/types/BufferedBusyTime";
 import type { AdditionalInformation, AppsStatus, CalendarEvent, Person } from "@calcom/types/Calendar";
 import type { EventResult, PartialReference } from "@calcom/types/EventManager";
-import type { WorkingHours } from "@calcom/types/schedule";
+import type { WorkingHours, TimeRange as DateOverride } from "@calcom/types/schedule";
 
 import type { EventTypeInfo } from "../../webhooks/lib/sendPayload";
 import sendPayload from "../../webhooks/lib/sendPayload";
@@ -117,10 +117,14 @@ const isWithinAvailableHours = (
   timeSlot: { start: ConfigType; end: ConfigType },
   {
     workingHours,
+    dateOverrides,
     organizerTimeZone,
+    inviteeTimeZone,
   }: {
     workingHours: WorkingHours[];
+    dateOverrides: DateOverride[];
     organizerTimeZone: string;
+    inviteeTimeZone: string;
   }
 ) => {
   const timeSlotStart = dayjs(timeSlot.start).utc();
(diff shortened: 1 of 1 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -144,6 +148,22 @@ const isWithinAvailableHours = (
       return true;
     }
   }
+
+  // check if it is a date override
+  for (const dateOverride of dateOverrides) {
```
CarinaWolli, line 153: we only checked if the slot is within working hours but it can also be a date override that is outside of the working hours

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -144,6 +148,22 @@ const isWithinAvailableHours = (
...
+    const utcOffSet = dayjs(dateOverride.start).tz(inviteeTimeZone).utcOffset();
```
hariombalhara, line 154: Screenshot 2023-04-26 at 7 50 20 PM
As per my understanding, dateOverride.start doesn't matter. The utcOffset depends just on inviteeTimeZone.
So, we can make it and then move outside the loop and compute just once.
```javascript
    const utcOffSet = dayjs().tz(inviteeTimeZone).utcOffset();
```

hariombalhara (changes requested):
Left a comment that can possibly be a bug.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -144,6 +148,22 @@ const isWithinAvailableHours = (
...
+
+    const slotStart = dayjs(timeSlotStart).add(utcOffSet, "minute");
+    const slotEnd = dayjs(timeSlotEnd).add(utcOffSet, "minute");
+
+    if (
+      slotStart.isBetween(dateOverride.start, dateOverride.end) &&
```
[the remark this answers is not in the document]
CarinaWolli: We are having a bigger known issue with date overrides which is that date overrides currently are not adjusted to the timezone. 

So if the schedule has a date override from 1PM - 3PM, then no matter what timezone the booking person has it will always show the 1PM - 3PM slots (in their timezone) open. This is something we need to fix separately (it's quite complex)

For now, when you log the date override and the slot start it should both be the time of how the slots are shown for the booking person. 

@hariombalhara let me know if that makes sense to you and if you have any other concerns with that implementation 🙏

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara: Thanks that makes sense 🙏

hariombalhara (approved):

PR #8547, opened 2023-04-27
fix: remove duplicate value i18n

### LINKED ISSUE #8546: [CAL-1600] Duplicate values in en.json
### Issue Summary

There are many duplicate values in the locales/en/common.json file. for example key booking_submitted 

<sub>[CAL-1600](https://linear.app/calcom/issue/CAL-1600/duplicate-values-in-enjson)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/8546

I just used this simple script to find all duplicate values

```

const valueSet = new Set();

for (const key in common) {
    if (valueSet.has(common[key])) {
      console.log(`Duplicate value found in: ${key}`);
    } else {
      valueSet.add(common[key]);
    }
}
```

Edit: I have added duplicate check in next.config.js as suggested by hariom

Screenshot 2023-06-07 at 5 22 41 PM

## Type of change


Udit-takkar:
NEXT: remove all these keys from other languages JSON file

apps/web/pages/settings/security/two-factor-auth.tsx
```diff
@@ -36,7 +36,7 @@ const TwoFactorAuthView = () => {
   const isCalProvider = user?.identityProvider === "CAL";
   return (
     <>
-      <Meta title={t("2fa")} description={t("2fa_description")} />
+      <Meta title={t("2fa")} description={t("password_description")} />
```
hariombalhara, line 39: We actually need to fix the text for `2fa_description` here IMHO. This is 2 factor page and not password page.

apps/web/public/static/locales/en/common.json
```diff
@@ -485,15 +478,11 @@
...
   "in_person_meeting": "In-person meeting",
-  "attendee_in_person": "In Person (Attendee Address)",
   "in_person": "In Person (Organizer Address)",
   "link_meeting": "Link meeting",
-  "phone_call": "Attendee Phone Number",
-  "your_number": "Your phone number",
   "phone_number": "Phone Number",
   "attendee_phone_number": "Attendee Phone Number",
   "organizer_phone_number": "Organizer Phone Number",
-  "host_phone_number": "Your Phone Number",
```
hariombalhara, line 496: Is it possible that the text is stored in DB and we won't find it in code search so we consider it unused?

Though if that is being done, we should fix that. But just wanted to put this question out there.

apps/web/pages/settings/security/two-factor-auth.tsx
Udit-takkar: what text should i use here?

hariombalhara (changes requested):
It's an awesome PR !!
Left a comment or two.
Apart from that, I think we should add the check in next.config.js itself which can throw a warning(may be an error) when it detects duplicate values in en/common.json.

It would avoid doing this work again.

apps/web/pages/settings/security/two-factor-auth.tsx
PeerRich: "Set up your Two-factor authentication"

PeerRich:
@Udit-takkar can you reveisit this one? or is it too many conflicts?

Udit-takkar:
@PeerRich fixing this PR

apps/web/next.config.js
```diff
@@ -57,6 +57,20 @@ if (process.env.GOOGLE_API_CREDENTIALS && !validJson(process.env.GOOGLE_API_CRED
   );
 }
 
+const checkDuplicateValue = () => {
```
hariombalhara, line 60:
```suggestion
const informAboutDuplicateTranslations = () => {
```

apps/web/next.config.js
```diff
@@ -57,6 +57,20 @@ if (process.env.GOOGLE_API_CREDENTIALS && !validJson(process.env.GOOGLE_API_CRED
...
+  for (const key in englishTranslation) {
+    if (valueSet.has(englishTranslation[key])) {
+      console.warn("\x1b[33mDuplicate value found in:", "\x1b[0m", key);
+    } else {
+      valueSet.add(englishTranslation[key]);
+    }
+  }
+};
+
+checkDuplicateValue();
```
hariombalhara, line 72:
```suggestion
informAboutDuplicateTranslations();
```

hariombalhara (approved):

PR #8548, opened 2023-04-27
add an enum generator, stop importing from @prisma/client

juliusmarminge:
## What does this PR do?

Ref: https://calcominc.slack.com/archives/C04H3D79LAK/p1682597485408419

Moves runtime imports away from @prisma/client as possible and instead pulls enums out to it's own entrypoints via a generator


packages/config/eslint-preset.js
```diff
@@ -19,6 +19,18 @@ module.exports = {
...
+      {
+        paths: [
+          {
+            name: "@prisma/client",
+            message: "Import from @calcom/prisma (or @calcom/prisma/enums) instead",
+            allowTypeImports: true,
+          },
+        ],
+      },
+    ],
```
juliusmarminge, line 33: added this just to make sure there we're no more anomalies left - but might be good to keep it so that you don't import runtime stuff from here in the future

emrysal (approved):

PR #8569, opened 2023-04-29
Fix: Full name is not refreshing on user update

### LINKED ISSUE #8567: Full Name not refreshing after Profile Update
### Issue Summary

Left Navigation Profile Name not refreshed after Full Name update under My Profile on this route: `/settings/my-account/profile`

### Steps to Reproduce

1. Went to Settings -> MyAccount -> Profile page
2. Modified Full Name and Clicked on Update
3. Full Name didn't update/refresh immediately on the Left Navigation.

Refer to the screenshot below:

Screenshot 2023-04-29 at 2 41 22 PM

### Technical details

- Google Chrome version: 112.0.5615.137
- Node.js version: v16.19.0

AshishBarvaliya:
## What does this PR do?

Fixes #8567 

## Type of change
The session may retain previous user information even after it has been updated. Therefore, to ensure that we are using the most up-to-date user information, we are retrieving the user details directly from the `user` table rather than relying solely on the session details.

### DIFF
--- a/packages/features/settings/layouts/SettingsLayout.tsx
+++ b/packages/features/settings/layouts/SettingsLayout.tsx
@@ -105,14 +105,15 @@ const adminRequiredKeys = ["admin"];
 
 const useTabs = () => {
   const session = useSession();
+  const { data: user } = trpc.viewer.me.useQuery();
 
   const isAdmin = session.data?.user.role === UserPermissionRole.ADMIN;
 
   tabs.map((tab) => {
-    if (tab.name === "my_account") {
-      tab.name = session.data?.user?.name || "my_account";
+    if (tab.href === "/settings/my-account") {
+      tab.name = user?.name || "my_account";
       tab.icon = undefined;
-      tab.avatar = WEBAPP_URL + "/" + session.data?.user?.username + "/avatar.png";
+      tab.avatar = WEBAPP_URL + "/" + user?.username + "/avatar.png";
     }
     return tab;
   });


AshishBarvaliya:
@zomars 
If my proposed fix is appropriate, then we may also need to update this part of the code(workflow - filters) since it is also currently retrieving user information from the session, rather than the updated user information.

`packages/features/ee/workflows/pages/index.tsx`

Screenshot 2023-04-29 at 6 30 38 PM

Screenshot 2023-04-29 at 6 34 46 PM

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -105,14 +105,15 @@ const adminRequiredKeys = ["admin"];
...
 const useTabs = () => {
   const session = useSession();
+  const { data: user } = trpc.viewer.me.useQuery();
 
   const isAdmin = session.data?.user.role === UserPermissionRole.ADMIN;
 
   tabs.map((tab) => {
-    if (tab.name === "my_account") {
-      tab.name = session.data?.user?.name || "my_account";
+    if (tab.href === "/settings/my-account") {
```
hariombalhara, line 113: Why are we using href. The name seems to be my_account only.
Screenshot 2023-05-06 at 6 39 40 PM

hariombalhara (changes requested):
The way I understand `session` provided by next-auth isn't aware of this change. But I just read that there is[`update`](https://next-auth.js.org/getting-started/client#updating-the-session) method available that if called will take a refresh from the server.

Maybe we can try calling the update method and see if it works. If it does, then wherever session is used the problem would be automatically solved.

alannnc:
Used some changes from here to mix with #8570

PR #8571, opened 2023-04-29
Updated the CSS classes for the user name 

### LINKED ISSUE #8563: Bottom part of some lowercase characters are hidden on user profile section
### Issue Summary

Characters such as j,q,y etc have their bottom section hidden on the user profile section

image

### Technical details

-Tested on chrome and firefox

udayfakie:
This PR fixes the issue with lowercase characters such as j,q,y being cut off at the bottom in the user profile section. The bottom section of these characters were previously hidden due to a styling issue.

Fixes #8563

To fix this issue, I removed the block class and added the pb-1 class to the affected element in the code.

To test this fix, go to the user profile section and check if the bottom part of lowercase characters such as j,q,y are no longer hidden.


sean-brydon (approved):
Sure this works :)

packages/features/shell/Shell.tsx
```diff
@@ -308,10 +308,10 @@ function UserDropdown({ small }: { small?: boolean }) {
             {!small && (
               <span className="flex flex-grow items-center truncate">
                 <span className="flex-grow truncate text-sm">
-                  <span className="text-emphasis mb-1 block truncate font-medium leading-none">
+                  <span className="text-emphasis mb-1 pb-1 block truncate font-medium leading-none">
                     {user.name || "Nameless User"}
                   </span>
-                  <span className="text-default block truncate font-normal leading-none">
+                  <span className="text-default pb-1 truncate font-normal leading-none">
```
hariombalhara, line 314: This change itself fixes the issue for me. Line 311 change doesn't seem required.

In both the cases, I am not sure if this is the right solution, adding padding to solve a text not visible issue seems hackish.

PR #8593, opened 2023-05-01
perf: cache i18n response 

G3root:
## What does this PR do?

this PR aims to cache the i18n response rather than re-returning the same object using `http-cache`

## Type of change


apps/web/package.json
```diff
@@ -96,6 +96,7 @@
     "next-themes": "^0.2.0",
     "nodemailer": "^6.7.8",
     "otplib": "^12.0.1",
+    "pretty-cache-header": "^1.0.0",
```
G3root, line 99: a utility to write cache control headers.

apps/web/pages/api/trpc/[trpc].ts
```diff
@@ -67,7 +68,7 @@ export default trpcNext.createNextApiHandler({
     // we want to create a map that can match potential paths with their desired cache value
     const cacheRules = {
       "viewer.public.session": `no-cache`,
-      "viewer.public.i18n": `no-cache`,
+      "viewer.public.i18n": cacheHeader({ public: true, maxAge: "1d", staleWhileRevalidate: "1d" }),
```
Comment (G3root) [inline apps/web/pages/api/trpc/[trpc].ts:71]: cache i18n response

packages/trpc/server/routers/publicViewer/_router.tsx
```diff
@@ -43,9 +45,20 @@ export const publicViewerRouter = router({
       throw new Error("Failed to load handler");
     }
 
-    return UNSTABLE_HANDLER_CACHE.i18n({
-      ctx,
-    });
+    return UNSTABLE_HANDLER_CACHE.i18n(input.locale);
+  }),
+
+  locale: publicProcedure.query(async ({ ctx }) => {
```
G3root, line 51: a procedure to fetch the users locale

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -45,42 +45,51 @@ interface GeneralViewProps {
...
-    onSuccess: async () => {
-      // Invalidate our previous i18n cache
-      await utils.viewer.public.i18n.invalidate();
+    onSuccess: async (value) => {
+      const previousLocale = utils.viewer.public.locale.getData();
+
+      if (value?.locale) {
+        const locale = value?.locale;
+
+        await utils.viewer.public.i18n.fetch({ locale });
```
G3root, line 74: fetch the new locale

Udit-takkar:
i think @hariombalhara mentioned before that he faced some issue while trying to cache i18n

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -5,16 +5,26 @@ import { useEffect } from "react";
...
     staleTime: Infinity,
-    /**
-     * i18n should never be clubbed with other queries, so that it's caching can be managed independently.
-     * We intend to not cache i18n query
-     **/
     trpc: {
       context: { skipBatch: true },
     },
   });
+  return trpc.viewer.public.i18n.useQuery(
```
hariombalhara, line 14: This is a waterfall now and thus would be slow to fetch translations. Can we avoid this?
Just a suggestion, maybe we can store the locale in session(and update it there when user changes the locale). 
The benefit being it's readily available along with the session.

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -45,42 +45,51 @@ interface GeneralViewProps {
...
-const WithQuery = withQuery(trpc.viewer.public.i18n, undefined, { trpc: { context: { skipBatch: true } } });
-
 const GeneralQueryView = () => {
   const { t } = useLocale();
 
+  const { data: locale, isLoading: localeLoading } = trpc.viewer.public.locale.useQuery();
   const { data: user, isLoading } = trpc.viewer.me.useQuery();
-  if (isLoading) return <SkeletonLoader title={t("general")} description={t("general_description")} />;
+  if (isLoading || localeLoading)
+    return <SkeletonLoader title={t("general")} description={t("general_description")} />;
```
hariombalhara, line 54: Earlier SkeletonLoader was only waiting for user data. Now it's also waiting for the locale. Is this intentional?

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -45,42 +45,51 @@ interface GeneralViewProps {
...
   if (!user) {
     throw new Error(t("something_went_wrong"));
   }
-  return (
-    <WithQuery
-      success={({ data }) => <GeneralView user={user} localeProp={data.locale} />}
-      customLoader={<SkeletonLoader title={t("general")} description={t("general_description")} />}
-    />
-  );
+  return <GeneralView user={user} localeProp={locale?.locale ?? "en"} />;
```
hariombalhara, line 58: I think we should show the error here when we can't get the locale.Earlier that error was tackled WithQuery automatically but now we are just defaulting to en.

apps/web/pages/settings/my-account/general.tsx
hariombalhara: Shouldn't we simply be able to invalidate both i18n and locale? Even though they have infinite staleTime, manually invalidating should work still.

apps/web/server/lib/ssg.ts
```diff
@@ -21,21 +20,19 @@ export async function ssgInit<TParams extends { locale?: string }>(opts: GetStat
...
       session: null,
       locale,
-      i18n: _i18n,
     },
   });
 
   // always preload i18n
-  await ssg.viewer.public.i18n.fetch();
+
+  await ssg.viewer.public.i18n.prefetch({ locale });
```
hariombalhara, line 35: What's the difference?

apps/web/server/lib/ssr.ts
```diff
@@ -26,7 +26,10 @@ export async function ssrInit(context: GetServerSidePropsContext) {
   });
 
   // always preload "viewer.public.i18n"
-  await ssr.viewer.public.i18n.fetch();
+
+  const { locale } = await ssr.viewer.public.locale.fetch();
+
+  await ssr.viewer.public.i18n.prefetch({ locale });
```
hariombalhara, line 32: Does it make sense to introduce an i18nWithLocale kind of thing that takes care of both in a single call?

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -137,4 +137,6 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
       .then(() => console.info("Booking pages revalidated"))
       .catch((e) => console.error(e));
   }
+
+  return { ...updatedUser, ...(input?.locale && { locale: input.locale }) };
```
hariombalhara, line 141: Why do we need this change?

hariombalhara (changes requested):
Left some comments. 
@Udit-takkar the caching bug that occured earlier was that i18n(specific to locale) was cached along with locale at the server(I think we were using the s-maxage header) and thus whenever one user changed its locale, the translations got updated for all users because they were served from server cache.

Now that locale and i18n are different queries, they can be cached separately. So, that should solve the problem.

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -5,16 +5,26 @@ import { useEffect } from "react";
 import { trpc } from "@calcom/trpc/react";
 
 export function useViewerI18n() {
-  return trpc.viewer.public.i18n.useQuery(undefined, {
+  const { data } = trpc.viewer.public.locale.useQuery(undefined, {
```
hariombalhara, line 8: `locale` needs session, so it shouldn't be under public namespace now.

zomars (changes requested):
I appreciate the effort @G3root but it seems like this PR isn't achieving the goal of improving perf and DX. Let's take a step back and discuss further strategies for this. We already had some other ideas cc @emrysal @roae where we could cache locales more aggressively by incorporating the locale into the requested URL itself.

zomars:
Closing for now

PR #8597, opened 2023-05-01
fix: Use correct values on append DayRange

harrydigos:
## What does this PR do?

This PR solves the issue of appending the previous value of time select if the select has changed in the availability schedule.

Fixes # (issue)

### Before
Working Hours _Availability_Cal com_Before

### After
WorkingHours _Availability_ Cal com_After

## Type of change

## Checklist


Udit-takkar (approved):
LGTM

PR #8616, opened 2023-05-02
use named imports instead to avoid polluting the global namespace. in…

tony83033:
## What does this PR do?

in -> packages/emails/templates/_base-email.ts
The code imports nodemailer packages using default imports. It's recommended to use named imports instead to avoid polluting the global namespace.

## Type of change


packages/emails/templates/_base-email.ts
```diff
@@ -1,5 +1,5 @@
 import { decodeHTML } from "entities";
-import nodemailer from "nodemailer";
+import {createTransport} from "nodemailer";
```
Udit-takkar, line 2: @tony83033  are there any benefits of importing only createTransport ?

yarn.lock
hariombalhara, line 1: Unrelated change in yarn.lock. It needs to be reverted.

packages/emails/templates/_base-email.ts
hariombalhara: I think tree shaking works better this way.

hariombalhara (dismissed):
There is an unrelated change in yarn.lock, otherwise it's good.

packages/emails/templates/_base-email.ts
emrysal, line 2:
```suggestion
import { createTransport } from "nodemailer";
```

emrysal (approved):

yarn.lock
tony83033: Thanks, @hariombalhara  sir for pointing this out, actually I am new to the world of open source, could you please explain to me how can I revert it?

hariombalhara: @emrysal took care of that.  I was talking about git revert and you can easily find articles on it 🙏

tony83033: ohh got it Thanks

PR #8621, opened 2023-05-02
Disable react-hooks/exhaustive-deps, autofix is very risky

emrysal:


packages/config/eslint-preset.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
     "jsx-a11y/role-supports-aria-props": "off", // @see https://github.com/vercel/next.js/issues/27989#issuecomment-897638654
     "react/jsx-curly-brace-presence": ["error", { props: "never", children: "never" }],
     "react/self-closing-comp": ["error", { component: true, html: true }],
+    "react-hooks/exhaustive-deps": "off",
```
hariombalhara, line 27: I think we should disable it for the specific useEffect call with the reasoning.

keithwillcode: @hariombalhara Can you elaborate on why?

hariombalhara: As per my understanding, this rule if enabled tells the developer about a missing dependency in useEffect. In general, one would want that when a variable changes its value, `useEffect` logic should run again. He might want to avoid that if he knows that it will trigger unnecessary or infinite renders.
e.g. https://github.com/calcom/cal.com/blob/81b643ad3679adcf9bcd423b391327e2beadf379/packages/lib/hooks/useTheme.tsx#L26 there is a comment I added here(though I forgot to disable the rule there I think).

Here we have disabled the rule, https://github.com/calcom/cal.com/blob/81b643ad3679adcf9bcd423b391327e2beadf379/apps/web/components/auth/TwoFactor.tsx#L22 but without any reasoning.

Disabling the rule might create a scenario where useEffect isn't called and we won't be informed about this by linter.

keithwillcode: Ah yeah good points. I tend to agree that, at least for now, it should be on a case-by-case basis instead of fully disabled.

zomars: Agreed with @keithwillcode and @hariombalhara will close for now.

PR #8629, opened 2023-05-02
App store

### LINKED ISSUE #8617: `app-store create` doesn't add import to the `appStore` object
### Issue Summary

All app store packages are exported from [`/packages/app-store/index.ts`](https://github.com/calcom/cal.com/blob/ea676907b21966597021dd45621701f478e19be9/packages/app-store/index.ts). When creating a new app using the CLI (`app-store create`), the resulting package is not automatically added to this exports list.

### Steps to Reproduce

1. `yarn app-store create` and make an app called `example`
2. open `/packages/app-store/index.ts` -> `example: import("./example")` not there

Maybe this is intentional, as most apps might not need their stuff exported in this way. But this tripped me up when I tried developing an app. It's not clear that the package needs to be manually added to this object.

elhadis:
This PR fixes the issue -Create app-store create ,

Fixes https://github.com/calcom/cal.com/issues/8617

to fix this issues I use yarn app-store create.

import examlpe from " /packages/app-store/index.ts "


hariombalhara (changes requested):
Thank you @elhadis 🙏 But the PR at the moment has many unrelated changes.

Also, https://github.com/calcom/cal.com/issues/8617#issuecomment-1532438209 the implementation needs discussion first.

PR #8633, opened 2023-05-03
fix: location select logo in dark mode

G3root:
## What does this PR do?

prevent inverting app logos in dark mode.

before:
Screenshot 2023-05-03 at 09-13-08 30min Event Type Cal com

after:

Screenshot 2023-05-03 at 09-07-48 30min Event Type Cal com

## Type of change


apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -285,7 +285,7 @@ export const EventSetupTab = (
                           "h-4 w-4",
                           // invert all the icons except app icons
                           eventLocationType.iconUrl &&
-                            !eventLocationType.iconUrl.startsWith("/api") &&
```
hariombalhara, line 288: There should be a better way to do this. The path check isn't the right way.
Don't we just want to invert SVGs? or we want to invert non svg images as well ?

hariombalhara (changes requested):
Left a comment that should make the code future proof and more logical.

apps/web/components/eventtype/EventSetupTab.tsx
hariombalhara: Thanks for fixing it 🙏 I believe it's a regression from my App Store images cleanup PR.

G3root: the idea is to revert only generic icons ex: `location`,`phone` etc... which are SVG's. agreed on not a better a way to this, i think we are limited to this hack for now 🤔 . any suggestions ?

hariombalhara: Then we can just change the logic to check for `.svg` and invert.

G3root: we don't want to invert app icons which also `.svg`

hariombalhara: I am approving it for now because it's broken in production. 
But can you give me examples of 2 URLs, one where we want to invert and the other where we don't want to?

My only concern is that it will break in future too

hariombalhara (approved):
LGTM

Issue #6262, opened 2023-01-04
Option to add self-hosted Jitsi


gokceozan:
### Is your proposal related to a problem?

It is wonderful to have Jitsi as an option for a meeting. But it is frustrating to not be able to use a self-hosted (or collective-hosted) instance. I am sure privacy-focused individuals would prefer a self-hosted instance as an option.

### Describe the solution you'd like

Have the option on the app to connect to a self-hosted (or collective-hosted) Jitsi instance as the provider. Failing to provide that would result in public Jitsi servers. Similar to the CalDav connection.

### Describe alternatives you've considered

N/A

### Additional context

N/A

PeerRich:
we would need community contributions for this. happy to accept incoming PRs

sopel:
This duplicates https://github.com/calcom/cal.com/issues/1847.

* Note that the conversation lists some attempts/workarounds to use a self-hsoted Jitsi instance, so depending on your requirements that might already work for you.

chief-nerd:
There is an open pull request from me that would add this feature: https://github.com/calcom/cal.com/pull/8636

PR #8636, opened 2023-05-03
Make Jitsi Host configurable

### LINKED ISSUE #1847: Allow using self-hosted Jitsi

Currently, you can only use Jiti (https://meet.jit.si/) servers with Jitsi meetings.

I'd like the option to hosting meetings on my own Jitsi server on my own domain.

### Additional context
Jisti is the largest (only?) FOSS video calling solution, and it would be a great complement to the only FOSS scheduling solution.

chief-nerd:
## What does this PR do?

Fixes #1847 and #6262  - Allow configuring Jitsi Host from settings/admin/apps. See https://www.loom.com/share/270199a955d14959aa1f0315beac9060

## Type of change

## How should this be tested?
- [ ] Configure jitsi_host in settings/admin/apps interface.
- [ ] Schedule a meeting with Location: Jitsi
- [ ] URL should be the self-hosted instance

## Checklist


Udit-takkar:
@chief-nerd i have re triggered them

chief-nerd:
If you are looking for a quick fix (assume running the Docker version):
`docker compose exec -it calcom /calcom/scripts/replace-placeholder.sh meet.jit.si  meet.mydomain.com`

agustif:
Hey @chief-nerd I think the issue is just some linting issues around using tabs instead of spaces for the modified lines (probably some defaults from your IDE)

Also was missing to add in turbo.json the jitsu env var 

Deployed a PR with the fixes here in case it helps https://github.com/calcom/cal.com/pull/8681

packages/app-store/jitsivideo/zod.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from "zod";
+
+export const appKeysSchema = z.object({
+  jitsi_host: z.string(),
```
hariombalhara, line 4: @chief-nerd This should show the jitsi_host variable in settings/admin/apps UI.

hariombalhara (changes requested):
Thanks for the PR @chief-nerd 🙏
I have pushed a change that would show jitsi_host in UI and save it.

Can you please take care of reading the host from keys. That should avoid adding the new env variable.

chief-nerd (commented):
Thanks for the update

chief-nerd (commented):
Ok

chief-nerd (commented):
Great. Thanks. @hariombalhara I think I accepted your contribution.... hope so. And this can be merged now.

packages/app-store/jitsivideo/lib/VideoApiAdapter.ts
```diff
@@ -31,4 +43,16 @@ const JitsiVideoApiAdapter = (): VideoApiAdapter => {
   };
 };
 
+function generateEventSlug(pattern: string, event: CalendarEvent): string {
+  pattern
```
hariombalhara, line 47: We are not using the return value here. So, this fn doesn't do it's job.

Also, I would like to discuss more about what variables need to be supported here and why we even need this?

The way I understand the Jitsi URL is automatically generated. Why would anyone care if it follows a certain format?

@chief-nerd Could you remove the slug pattern implementation from this PR for now?

packages/app-store/jitsivideo/zod.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from "zod";
+
+export const appKeysSchema = z.object({
+  jitsi_host: z.string().default("https://meet.jit.si"),
```
hariombalhara, line 4:
```suggestion
  jitsiHost: z.string().default("https://meet.jit.si"),
```

hariombalhara, line 4:
```suggestion
  jitsi_host: z.string().default("https://meet.jit.si/cal/"),
```

The default is this. Also, / is very important at the end, otherwise the link would be broken.

packages/app-store/jitsivideo/lib/VideoApiAdapter.ts
chief-nerd: @hariombalhara make the changes.

> The way I understand the Jitsi URL is automatically generated. Why would anyone care if it follows a certain format?

Cause it looks nice :) - especially on my locally hosted version
https://meet.mydomain.com/e8d513ef-b385-48de-8c06-6a483e78e224
vs
https://meet.mydomain.com/Meet-and-Greet-Max-Musterman

packages/app-store/jitsivideo/zod.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from "zod";
+
+export const appKeysSchema = z.object({
+  jitsiHost: z.string().optional(),
```
hariombalhara, line 4: Removed default from here as that doesn't work currently in the way this schema is used.

hariombalhara (approved):
LGTM now !! Thanks @chief-nerd 

Tests:
- Tested that existing installation of Jitsi would keep working
- Tested that configuring Jitsi with new Host, uses the new URL.

chief-nerd:
🎉 -- so looking forward to finally (officially) using this

PR #8638, opened 2023-05-03
[URGENT] Hotfix: re-adds the password and username set check in signup

alishaz-polymath:
## What does this PR do?

Check for existing password and username to ensure it isn't an invite. 
If we send a team invite, the email is added to the DB and the existing check would render their sign-up process to fail as their email is in the DB.
~Reverting this to a previous state (prior to commit-1ee3783db30fe62dcf7fe1eef746759275525ecc) to check for a set password and a set username which ensures that the user is already registered and it isn't just an invite.~
Added a check to allow the previous state (if a password and username is also set), along with a check based on the fact that email is verified, which is true for SAML/SSO.

Fixes # (issue)

## Type of change

### DIFF
--- a/apps/web/pages/api/auth/signup.ts
+++ b/apps/web/pages/api/auth/signup.ts
@@ -36,8 +36,9 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     return;
   }
 
-  // There is actually an existingUser if username matches
-  // OR if email matches and both username and password are set
+  // There is an existingUser if the username matches
+  // OR if the email matches AND either the email is verified
+  // or both username and password are set
   const existingUser = await prisma.user.findFirst({
     where: {
       OR: [
--- a/apps/web/pages/api/auth/signup.ts
+++ b/apps/web/pages/api/auth/signup.ts
@@ -43,7 +43,17 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       OR: [
         { username },
         {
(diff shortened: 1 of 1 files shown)


hariombalhara (approved):

keithwillcode (approved):

keithwillcode (commented):
Did you mean to push up `CalendarDebug.tsx`?

alishaz-polymath:
> Did you mean to push up `CalendarDebug.tsx`?

Definitely didn't. Wrong use of tab to auto add on my end. 😢 
Thanks for the catch!

apps/web/pages/api/auth/signup.ts
```diff
@@ -36,14 +36,25 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
     where: {
       OR: [
         { username },
         {
-          AND: [{ email: userEmail }],
+          AND: [
+            { email: userEmail },
+            {
+              OR: [
+                { emailVerified: { not: null } },
```
keithwillcode, line 51: Is this better to be a boolean check for true?

keithwillcode: Cool.

keithwillcode (approved):

keithwillcode:
Merging with failing tests because we are having issues in multiple branches/PR with E2E tests failing for reasons unassociated with code changes

PR #8639, opened 2023-05-03
chore: add canonical to the document via `PageWrapper`

p6l-richard:
## What does this PR do?
Fixes `Duplicate without user-selected canonical` issue in [google search console](https://search.google.com/search-console/index/drilldown?resource_id=sc-domain%3Acal.com&item_key=CAMYDyAC):
CleanShot 2023-05-03 at 15 21 13
CleanShot 2023-05-03 at 15 22 26

## Demo
CleanShot 2023-05-03 at 15 23 28

**Environment**: Production

## Type of change

## How should this be tested?

- [ ] Check the dev tools network tab and see if the document includes the `canonical` link


Udit-takkar (approved):
LGTM

packages/ui/components/head-seo/HeadSeo.tsx
```diff
@@ -73,16 +73,17 @@ export const HeadSeo = (props: HeadSeoProps): JSX.Element => {
...
   // Get the router's path
   const path = useRouter().asPath;
-  // Build the canonical URL using the router's path, without query parameters. Note: on homepage it omits the trailing slash
-  const calcomCanonical = `https://cal.com${path === "/" ? "" : path}`.split("?")[0];
+  const selfHostedOrigin = origin || "https://cal.com";
   // Set the default URL to either the current URL (if self-hosted) or https://cal.com canonical URL
-  const defaultUrl = isCalcom ? calcomCanonical : url;
+  const defaultUrl = isCalcom
+    ? buildCanonical({ path, origin: "https://cal.com" })
+    : buildCanonical({ path, origin: selfHostedOrigin });
```
roae, line 87: AFAIK this must be generated by the server, to have an effect on the seo, so, I think we should not get the origin from the browser, because this info should be available on build time, SSR or ISR. 

CC. @PeerRich @zomars @keithwillcode @emrysal

roae (changes requested):
Need clarification.

packages/ui/components/head-seo/HeadSeo.tsx
p6l-richard: I agree with this. This is also why the buildCanonical doesn't use this. 

The 'isCalcom' should read the URL ideally on something available on the server.

roae: I think we can use `WEBAPP_URL` constant 

`import { WEBAPP_URL } from "@calcom/lib/constants";`

PeerRich: @p6l-richard how is this looking?

PeerRich:
@roae can you adjust your review?

packages/ui/components/head-seo/HeadSeo.tsx
roae: I use the `WEBSITE_URL` instead of the `origin` from getBrowserInfo() to avoid an empty canonical value on SSR

roae (approved):
Thanks @p6l-richard for the contribution! 🙏

PR #8640, opened 2023-05-03
Stabilize Tests

hariombalhara:
Earlier waitForNavigation was behaving in a very flaky manner and it was replaced by waitForURL everywhere. But waitForURL has a difference that it can consider the same page in assertion without actually waiting for navigation(vs waitForNavigation which would require an actual navigation to even match the current page URL)

There are fixes in this PR for the above scenario.

### DIFF
--- a/apps/web/playwright/app-store.e2e.ts
+++ b/apps/web/playwright/app-store.e2e.ts
@@ -12,6 +12,7 @@ test.describe("App Store - Authed", () => {
     await pro.login();
     await page.goto("/apps/categories/calendar");
     await page.click('[data-testid="app-store-app-card-apple-calendar"]');
+    await page.waitForURL("/apps/apple-calendar");
     await page.click('[data-testid="install-app-button"]');
     await expect(page.locator(`text=Connect to Apple Server`)).toBeVisible();
   });
--- a/apps/web/playwright/booking-pages.e2e.ts
+++ b/apps/web/playwright/booking-pages.e2e.ts
@@ -94,7 +94,12 @@ testBothBookers.describe("pro user", () => {
     });
   });
 
-  test("Can cancel the recently created booking and rebook the same timeslot", async ({ page, users }) => {
+  test("Can cancel the recently created booking and rebook the same timeslot", async ({
+    page,
+    users,
+  }, testInfo) => {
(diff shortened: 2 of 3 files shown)


apps/web/playwright/app-store.e2e.ts
```diff
@@ -12,6 +12,7 @@ test.describe("App Store - Authed", () => {
     await pro.login();
     await page.goto("/apps/categories/calendar");
     await page.click('[data-testid="app-store-app-card-apple-calendar"]');
+    await page.waitForURL("/apps/apple-calendar");
```
hariombalhara, line 15: Well the install-app-button is on this page only. So, wait for it.

apps/web/playwright/booking-pages.e2e.ts
```diff
@@ -103,7 +108,7 @@ testBothBookers.describe("pro user", () => {
     await page.goto("/bookings/upcoming");
     await page.locator('[data-testid="cancel"]').first().click();
     await page.waitForURL((url) => {
-      return url.pathname.startsWith("/booking");
+      return url.pathname.startsWith("/booking/");
```
hariombalhara, line 111: `/booking` also matches `/bookings/upcoming` but we wanted to match /booking/something

packages/embeds/embed-core/playwright/lib/testUtils.ts
```diff
@@ -91,10 +91,11 @@ async function selectFirstAvailableTimeSlotNextMonth(frame: Frame, page: Page) {
 }
 
 export async function bookFirstEvent(username: string, frame: Frame, page: Page) {
-  // Click first event type
+  // Click first event type on Profile Page
   await frame.click('[data-testid="event-type-link"]');
   await frame.waitForURL((url) => {
-    return !!url.pathname.match(new RegExp(`/${username}/.*$`));
+    // Wait for reaching the event page
+    return !!url.pathname.match(new RegExp(`/${username}/.+$`));
```
hariombalhara, line 98: Profile page also has the format /${username}/ but we want to wait for /${username}/{slug}

keithwillcode (approved):
Nice work

PR #8660, opened 2023-05-04
CAL-1558: Fixed responsiveness of booker for small calendar layout

JeroenReumkens:
## What does this PR do?

Improves responsive behaviour for non fullscreen layouts. Issue only occurred in this view when also having the timeslots open. 

### Previous behaviour: 

CleanShot 2023-05-04 at 11 07 39@2x

### After this PR: 
CleanShot 2023-05-04 at 11 08 07@2x

Fixes CAL-1558

## Type of change

## How should this be tested?

- [ ] Ensure that the view now behaves correctly responsive
- [ ] View shouldn't be broken on desktop now
- [ ] Other two views should still be responsive too
- [ ] When going to mobile we always force the small calendar view

### DIFF
--- a/packages/features/bookings/Booker/config.ts
+++ b/packages/features/bookings/Booker/config.ts
@@ -54,18 +54,21 @@ export const resizeAnimationConfig: ResizeAnimationConfig = {
   small_calendar: {
     default: {
       width: "calc(var(--booker-meta-width) + var(--booker-main-width))",
+      // maxWidth: "100%",
       gridTemplateAreas: `"meta main"`,
       gridTemplateColumns: "var(--booker-meta-width) var(--booker-main-width)",
     },
     selecting_time: {
-      width: "calc(var(--booker-meta-width) + var(--booker-main-width) + var(--booker-timeslots-width))",
+      width: "100%",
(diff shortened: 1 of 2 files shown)


packages/features/bookings/Booker/config.ts
```diff
@@ -54,25 +54,29 @@ export const resizeAnimationConfig: ResizeAnimationConfig = {
...
       gridTemplateAreas: `"meta main"`,
       gridTemplateColumns: "var(--booker-meta-width) var(--booker-main-width)",
     },
     selecting_time: {
-      width: "calc(var(--booker-meta-width) + var(--booker-main-width) + var(--booker-timeslots-width))",
+      width: "100%",
+      maxWidth: "calc(var(--booker-meta-width) + var(--booker-main-width) + var(--booker-timeslots-width))",
       gridTemplateAreas: `"meta main timeslots"`,
-      gridTemplateColumns: "var(--booker-meta-width) var(--booker-main-width) var(--booker-timeslots-width)",
+      gridTemplateColumns: "var(--booker-meta-width) 1fr var(--booker-timeslots-width)",
```
JeroenReumkens, line 64: This actually is the real improvement here. The 1fr instead of a fixed width allows it to scale down and fit in the remaining space.

sean-brydon (approved):
Tested - works much better

PR #8662, opened 2023-05-04
Fixing availability for MET

leog:
## What does this PR do?

 Fixes:
 - Availability for MET
 - Delete dialog in event type list was not showing text correctly
 - Event type list actions for wide and narrow screens


apps/web/components/eventtype/EventAvailabilityTab.tsx
```diff
@@ -205,6 +205,8 @@ const EventTypeSchedule = ({ eventType }: { eventType: EventTypeSetup }) => {
       const value = options.find((option) =>
         scheduleId
           ? option.value === scheduleId
+          : isManagedEventType
+          ? option.value === 0
```
leog, line 209: We want to show "members' default schedule" text if no specific availability schedule was assigned to the MET

apps/web/pages/event-types/index.tsx
```diff
@@ -352,6 +352,9 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
 
   const firstItem = types[0];
   const lastItem = types[types.length - 1];
+  const isManagedEventPrefix = () => {
+    return deleteDialogTypeSchedulingType === SchedulingType.MANAGED ? "_managed" : "";
+  };
```
leog, line 357: Part of the fix for delete dialog in the event type list

CarinaWolli (approved):
I think embed should be added to the dropdown for children managed even types, as we also have embed as an option within the event type: 
Screenshot 2023-05-08 at 14 54 30

Can be done in a follow-up as well

Other than that, it looks good to me 👍

PR #8670, opened 2023-05-04
CAL-1589: Added utm link to powered by in booker. 

JeroenReumkens:
## What does this PR do?

Added utm link to powered by in booker. Also added new colors to powered by component, so they're also used in current booker form now on.

Also added the option to only have the logo instead of adding the 'powered by' text, but this is for now only used in the new booker. 

@hariombalhara  FYI this slightly changes the logo colors in the current booker (and thus embed) as well. But since this will be the new colors that will be used soon I don't think it matters. FYI nevertheless. 

CleanShot 2023-05-04 at 13 59 23@2x

Fixes CAL-1589

## Type of change


Udit-takkar (approved):
LGTM

Screenshot 2023-05-04 at 10 48 52 PM

PR #8672, opened 2023-05-04
feat: Choosing the email you want to be displayed as the organizer

### LINKED ISSUE #5774: [CAL-499] Choosing the email you want to be displayed as the organizer
**Problem**

* I have a personal [cal.com](http://cal.com) account for (eg [myname@gmail.com](mailto:myname@gmail.com)).
* I also want to use [cal.com](http://cal.com) for my work purposes. To do that I attached my corporate Gmail calendar (eg [myname@mycompany.com](mailto:myname@mycompany.com)) to the existing personal [cal.com](http://cal.com) account. As a result now I get scheduled events in the right calendar ([myname@mycompany.com](mailto:myname@mycompany.com)) but as a participant I am displayed there with my personal email ([myname@gmail.com](mailto:myname@gmail.com))

**Desired outcome**

* Let the user choose on a per event basis which email address for them to appear as the organiser for.
* The list of choices should be pulled from their list of connected calendars.
* We would likely place this alongside the "Add to calendar" UI in the event type settings

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-499](https://linear.app/calcom/issue/CAL-499/choosing-the-email-you-want-to-be-displayed-as-the-organizer)</sub>

gitstart-calcom:
## What does this PR do?

Fixes #5774 

## Demo
https://www.loom.com/share/06ee5751444c4c53a8ff43bcc2b1cd8c

image


agustif:
>  If you could please direct us to the right component it would save us some time. Thank you! cc: Omar López Ciarán Hanrahan


Hey there, not sure about if this is the correct answer here but did you check on the email templates?
https://github.com/calcom/cal.com/tree/main/packages/emails/templates

Also checkout CalendarManager where the primary email is passed.

https://github.com/calcom/cal.com/blob/05b42c001f997604cc55e4789ccaaf052cfca434/packages/core/CalendarManager.ts#LL84C1-L84C1

joeauyeung:
@gitstart-calcom the video in your PR description isn't working. Can you reupload it or maybe use Loom?

gitstart-calcom:
> @gitstart-calcom the video in your PR description isn't working. Can you reupload it or maybe use Loom?

Sure, we will add a new demo video

gitstart-calcom:
@joeauyeung The demo video has been updated now
https://www.loom.com/share/06ee5751444c4c53a8ff43bcc2b1cd8c

joeauyeung (commented):
Some feedback
- The user should be able to choose which email is displayed on the confirmation page. I noticed in the PR it comes from the calendar. I would review the desired outcomes of #5774 
- On the success page, we should show the email immediately. Could we store the email the user wants to display in the event type DB record?

gitstart-calcom:
Apologies for the delay on the change request. We were trying to tackle some edge cases and got to the following conclusion:

We have tried 2 approaches trying to avoid modifying the db:
- For the booking success page, we tried fetching the destination calendar alongside the booking data. But to retrieve the corresponding calendar email would take a lot of computation. Taking how it is done in CalendarManager.ts in getCalendarCredentials and getConnectedCalendars functions.
- Another option was to make trpc calls within the React function to get the connected calendars and then retrieve the primary (organizer) email from it. An issue with this is that the organizer email won't show on the page load, so we would have to show a loader first (we got [feedback](https://github.com/calcom/cal.com/pull/8672#pullrequestreview-1423532947) to remove it). Another issue is that we would still make this API call in all the files where we need the correct organizer email.

So, storing the email the user wants to show in the eventType db record may be a better approach. We would like to confirm this approach before we go ahead.

joeauyeung:
> Apologies for the delay on the change request. We were trying to tackle some edge cases and got to the following conclusion:
> 
> We have tried 2 approaches trying to avoid modifying the db:
> 
>     * For the booking success page, we tried fetching the destination calendar alongside the booking data. But to retrieve the corresponding calendar email would take a lot of computation. Taking how it is done in CalendarManager.ts in getCalendarCredentials and getConnectedCalendars functions.
> 
>     * Another option was to make trpc calls within the React function to get the connected calendars and then retrieve the primary (organizer) email from it. An issue with this is that the organizer email won't show on the page load, so we would have to show a loader first (we got [feedback](https://github.com/calcom/cal.com/pull/8672#pullrequestreview-1423532947) to remove it). Another issue is that we would still make this API call in all the files where we need the correct organizer email.
> 
> 
> So, storing the email the user wants to show in the eventType db record may be a better approach. We would like to confirm this approach before we go ahead.

Thanks for looking into it. I think we can store which email the user wants to be displayed under the event type metadata. Although if a user does not choose to change the displayed email for that event type then default to the email on their profile.

gitstart-calcom:
> Thanks for looking into it. I think we can store which email the user wants to be displayed under the event type metadata. Although if a user does not choose to change the displayed email for that event type then default to the email on their profile.

Thank you, we will make the changes and get back to you soon

gitstart-calcom:
Hi @joeauyeung, we have updated the PR. Please take a look.

PeerRich:
image

why do we have a skeleton here?

CarinaWolli:
> Hi @joeauyeung, we have updated the PR. Please take a look.

@joeauyeung could you take a look at this one again?

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -481,6 +483,13 @@ const EventTypePage = (props: EventTypeSetupProps) => {
               }
             }
             const { availability, ...rest } = input;
+
+            if (connectedCalendarsQuery.data?.connectedCalendars.length) {
+              defaultCalendar = connectedCalendarsQuery.data?.connectedCalendars
+                .map((connected) => connected.primary)
+                .find((cal) => cal?.externalId === input.destinationCalendar.externalId);
```
Comment (joeauyeung) [inline apps/web/pages/event-types/[type]/index.tsx:490]: When no destination calendar is selected and we use the default calendar `input.destinationCalendar` is set to null. We should fall back to the destination calendar email.

joeauyeung (changes requested):
Great job saving the emails to the event type metadata. Two things I think should be addressed:

- See my comment about saving an event type with the default destination calendar
- If a user changes their default destination calendar, we need to handle if we have to change the email of any event types that are using the default

gitstart-calcom:
Hi @joeauyeung, we moved this PR for this other [PR](https://github.com/calcom/cal.com/pull/10348), please take a look, we made the requested changes 😄

CarinaWolli:
This can be closed as PR #10348 has the most recent changes

PR #8677, opened 2023-05-04
Bug/checkbox calendar sync state 

### LINKED ISSUE #8438: [CAL-1570] 20 Second Delay on Button Color Showing on Selecting Calendar in Calendar Installation
eventually button color loaded

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1570](https://linear.app/calcom/issue/CAL-1570/20-second-delay-on-button-color-showing-on-selecting-calendar-in)</sub>

sean-brydon:
Fixes: https://github.com/calcom/cal.com/issues/8438

Move calendar switch to local state instead of waiting on async call to change value. 

Before
CleanShot 2023-05-04 at 10 08 24

After
CleanShot 2023-05-04 at 10 35 23


apps/web/components/apps/CalendarListContainer.tsx
```diff
@@ -31,83 +29,6 @@ type Props = {
   destinationCalendarId?: string;
 };
 
-function CalendarSwitch(props: {
```
sean-brydon, line 34: Move to resuable comp

packages/features/calendars/CalendarSwitch.tsx
```diff
@@ -69,15 +76,23 @@ const CalendarSwitch = (props: ICalendarSwitchProps) => {
...
       <label className="ml-3 text-sm font-medium leading-5" htmlFor={externalId}>
         {name}
       </label>
+      {!!props.destination && (
+        <span className="bg-subtle text-default ml-8 inline-flex items-center gap-1 rounded-md px-2 py-1 text-sm font-normal sm:ml-4">
+          <ArrowLeft className="h-4 w-4" />
+          {t("adding_events_to")}
+        </span>
+      )}
+      {mutation.isLoading && <RotateCcw className="text-muted h-4 w-4 animate-spin ltr:ml-1 rtl:mr-1" />}
```
sean-brydon, line 95: Add a spinner to let the user know the switch is doing something - without it you can accidentally navigate away when the request is being fired.

Udit-takkar:
@sean-brydon this looks much better. should i close this PR https://github.com/calcom/cal.com/pull/8595 then ?

packages/features/calendars/CalendarSwitch.tsx
```diff
@@ -60,6 +66,7 @@ const CalendarSwitch = (props: ICalendarSwitchProps) => {
         await utils.viewer.integrations.invalidate();
```
Udit-takkar, line 66: i think you would also have to add await utils.viewer.connectedCalendars.invalidate(); here

sean-brydon:
> @sean-brydon this looks much better. should i close this PR #8595 then ?

Yeah i believe this is a better approach

packages/features/calendars/CalendarSwitch.tsx
sean-brydon: Good spot - testing

sean-brydon: Commited

vaibhav135:
I think @hariombalhara recommended to not use local states for this? check #8595

sean-brydon:
@hariombalhara here is the example below without local state.  (I dont like it) 
CleanShot 2023-05-05 at 15 25 58

I personally much prefer having the switch be responsive as soon as i click it but open to be convinced. This api call doesnt use trpc FYI 

@Jaibles design feedback would be welcome here... You expect the switch to work as soon as you click it right

hariombalhara:
@sean-brydon My reasoning for it is the same as there for the `Save` button(except that it has a possibly larger payload and can take more time). When you click on the Save button, there is a loader that runs because it is updating something. We wouldn't want to save to be considered done when we know it can fail.
There are 2 ways we are using toggles in the app. 
- Here toggling it doesn't actually update anything unless you click the Save button later.
 Screenshot 2023-05-06 at 12 33 26 PM
- Current case that we are discussing, where toggling it actually updates something. [This case is similar to the Save button]
You mentioned that it's not using tRPC, but it's using react-query already so we can do an optimistic update to further improve the experience. That would then match the local state implementation in terms of responsiveness.

sean-brydon:
> @sean-brydon My reasoning for it is the same as there for the `Save` button(except that it has a possibly larger payload and can take more time). When you click on the Save button, there is a loader that runs because it is updating something. We wouldn't want `save` to be considered done when we know it can fail. There are 2 ways we are using toggles in the app.
> 
> * Here toggling it doesn't actually update anything unless you click the Save button later.
> 
> Screenshot 2023-05-06 at 12 33 26 PM
> 
> * Current case that we are discussing, where toggling it actually updates something. [This case is similar to the Save button]
>   You mentioned that it's not using tRPC, but it's using react-query already so we can do an optimistic update to further improve the experience. That would then match the local state implementation in terms of responsiveness.

Sure i agree with that :) We can do this in a follow up - This is affecting customers now and would like to get this merged so we can get it in for the next retreat

hariombalhara:
Sure @sean-brydon. Totally makes sense. It's a miss on my part that I didn't notice it's an important fix.

sean-brydon:
> Sure @sean-brydon. Totally makes sense. It's a miss on my part that I didn't notice it's an important fix.

No worries! The delay between api calls is (with no loading state also) is causing users to think the switch just doesnt work currently. Especially on slower connections.

Udit-takkar (changes requested):
Screenshot 2023-05-08 at 11 55 25 PM

there is some issue of spacing here /settings/my-account/calendars

Udit-takkar (approved):
Screenshot 2023-05-09 at 3 03 09 PM

LGTM. I pushed some changes in the PR. spacing was 16px in design.

Issue #8610, opened 2023-05-01
Profile picture does not update on the left sidebar


oneknucklehead:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

The profile picture does not update on the leftside navbar even though the profile picture has been updated already and the page has been refreshed continuously.

![Screenshot 2023-05-02 012913](https://user-images.githubusercontent.com/75200954/235521299-a3ad2570-54da-49d0-8381-7037dc070425.png)

### Steps to Reproduce

1. Go to https://app.cal.com/settings/my-account/profile
2. try updating your profile picture

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

### Technical details

- Browser version, screen recording, console logs, network requests: You can make a recording with [Bird Eats Bug](https://birdeatsbug.com/).
- Node.js version
- Anything else that you think could be an issue.

loki1205:
Hi, I know how to solve this issue. What's happening here is that the image data from the navbar is set to a source that gets data on the user login. Hence, when you sign out and again sign in, fresh data is stored and now the sidebar will have new data (image). Kindly assign this issue to me, and I can help you solve it.

Also, there were some UI changes. Like, in the profile settings section, there must be an edit button, which users can click on to edit their profile. Until the edit button is clicked, the profile data must be read-only. Also, when the "Update" button is clicked message 'Profile updated successfully' should be displayed and the profile should go back to read-only. Do you this is a good idea?

ujjwal-45:
@oneknucklehead  I have reproduced the bug that you have mentioned above , but the profile picture is updating on the sidebar too. Here is the scree
![Screenshot 2023-05-02 at 12-08-08 Profile Cal com](https://user-images.githubusercontent.com/106880198/235597204-6eacad42-9d4e-463f-8621-e74263d1c2f1.png)
nshot ..

loki1205:
> @oneknucklehead I have reproduced the bug that you have mentioned above , but the profile picture is updating on the sidebar too. Here is the scree ![Screenshot 2023-05-02 at 12-08-08 Profile Cal com](https://user-images.githubusercontent.com/106880198/235597204-6eacad42-9d4e-463f-8621-e74263d1c2f1.png) nshot ..

Sorry, I didn't understand. Is this on localhost? Or on the server, https://app.cal.com/settings/my-account/profile. Do you mean, you have fixed the bug or that the bug is not showing up for you?

ujjwal-45:
[...]

PR #8694, opened 2023-05-05
fix: avatar loading issue

### LINKED ISSUE #8448: [CAL-1574] Avatar not Loading after New Account Setup
This is a dumby account, but the avatar isn't loading after it's been set in settings (granted username has not been paid for)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1574](https://linear.app/calcom/issue/CAL-1574/avatar-not-loading-after-new-account-setup)</sub>

Keyurkd:
## What does this PR do?

Fixes #8448 

## Type of change

- Bug fix:  Fixes an avatar not updating based on user profile update.

### DIFF
--- a/apps/web/pages/event-types/index.tsx
+++ b/apps/web/pages/event-types/index.tsx
@@ -20,6 +20,8 @@ import { HttpError } from "@calcom/lib/http-error";
 import { SchedulingType } from "@calcom/prisma/enums";
 import type { RouterOutputs } from "@calcom/trpc/react";
 import { trpc, TRPCClientError } from "@calcom/trpc/react";
+import useAvatarQuery from "@calcom/trpc/react/hooks/useAvatarQuery";
+import useMeQuery from "@calcom/trpc/react/hooks/useMeQuery";
 import {
   Avatar,
   AvatarGroup,
@@ -660,6 +662,8 @@ const EventTypeListHeading = ({
 }: EventTypeListHeadingProps): JSX.Element => {
   const { t } = useLocale();
   const router = useRouter();
+  const { data: loggedInUser } = useMeQuery();
+  const { data: avatar } = useAvatarQuery();
 
   const publishTeamMutation = trpc.viewer.teams.publish.useMutation({
     onSuccess(data) {
@@ -675,7 +679,11 @@ const EventTypeListHeading = ({
       <Avatar
         alt={profile?.name || ""}
         href={teamId ? `/settings/teams/${teamId}/profile` : "/settings/my-account/profile"}
-        imageSrc={`${WEBAPP_URL}/${profile.slug}/avatar.png` || undefined}
+        imageSrc={
+          loggedInUser?.username === profile.slug
+            ? avatar?.avatar
+            : `${WEBAPP_URL}/${profile.slug}/avatar.png` || undefined
+        }
         size="md"
         className="mt-1 inline-flex justify-center"
       />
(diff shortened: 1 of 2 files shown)


Keyurkd:
@hariombalhara @Udit-takkar ,
Please review this.

apps/web/pages/event-types/index.tsx
```diff
@@ -675,7 +679,11 @@ const EventTypeListHeading = ({
       <Avatar
         alt={profile?.name || ""}
         href={teamId ? `/settings/teams/${teamId}/profile` : "/settings/my-account/profile"}
-        imageSrc={`${WEBAPP_URL}/${profile.slug}/avatar.png` || undefined}
+        imageSrc={
+          loggedInUser?.username === profile.slug
```
hariombalhara, line 683: I don't understand the reasoning behind this change. `${WEBAPP_URL}/${profile.slug}/avatar.png` is supposed to serve the correct avatar. Why is it not working?

hariombalhara (changes requested):
Need clarification

apps/web/pages/event-types/index.tsx
Keyurkd: @hariombalhara Yes `${WEBAPP_URL}/${profile.slug}/avatar.png` should have returned the correct avatar. It does return the correct avatar till the time user doesn't change the avatar in the current session. Due to session caching in the browser, it's not updating the image that the logged-in user has changed in the profile settings of a currently active session.

Keyurkd:
@hariombalhara You can follow these https://github.com/calcom/cal.com/issues/8567 https://github.com/calcom/cal.com/issues/8610 issues which have been raised on the profile update for more information.

apps/web/pages/event-types/index.tsx
Udit-takkar: @Keyurkd is there any way we can invalidate the avatar in the session cache when user updates it

hariombalhara: I don't think it is about the session. The problem is that the endpoint avatar.png is actually a [dynamic endpoint](https://github.com/calcom/cal.com/blob/5362cd66b4ab23392f1ad5ce5d5eca84aad0ab2a/apps/web/pages/api/user/avatar.ts#L68) that decides what's the profile pic for a user.  It currently has no cache-control headers and thus browsers are free to cache it. 

Making it completely non-cacheable will fix this issue but it would unnecessarily make function invocations for something that rarely changes.

This needs an RFC, but one quicker approach in our context can be that change the URL of the image every time the avatar changes. e.g. we compute and md5 of the image being uploaded and use that as query param in front of URL. Now, even if it is fully cached, changing the query param would be considered a different resources in terms of cache and will be fetched and cached again. We would have to store that md5 per user.

Also, if the issue is about the session not updating as well then there is a an update method available that might help https://next-auth.js.org/getting-started/client#updating-the-session

zomars:
While this solution works fine. Can't we a find a way to invalidate `/[usename]/avatar.png` from the client instead? That way we could force refresh everywhere without minor code changes.

Keyurkd:
@hariombalhara 

The `trigger` and `newSession` properties don't exist on NextAuth Options (`AUTH_OPTIONS`). Please refer to the below screenshots.

Screenshot 2023-05-12 at 4 00 56 PM


Screenshot 2023-05-12 at 4 03 27 PM


The `update` method doesn't exist in the useSession().data.

Screenshot 2023-05-12 at 4 07 07 PM

PeerRich:
tldr: i think the avatar endpoint is the culprit. turning into draft

PR #8696, opened 2023-05-05
Fix switch label hover

sean-brydon:
Fixes hover being unreadable on darkmode 

Makes it impossible to tell if the switch was actually there when the state was off

Before:
CleanShot 2023-05-05 at 12 11 53@2x

After:
CleanShot 2023-05-05 at 12 06 35@2x
CleanShot 2023-05-05 at 12 06 45@2x


leog (approved):
LGTM

hariombalhara (approved):

PR #8697, opened 2023-05-05
feat: add meta and seo tags for new booker component

### LINKED ISSUE #8522: [CAL-1598] new booker has no title (and likely no meta data?)
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1598](https://linear.app/calcom/issue/CAL-1598/new-booker-has-no-title-and-likely-no-meta-data)</sub>

G3root:
## What does this PR do?

Fixes #8522 and CAL-1598

## Type of change


apps/web/pages/new-booker/[user]/[type].tsx
```diff
@@ -89,13 +92,15 @@ async function getUserPageProps(context: GetServerSidePropsContext) {
     booking = await getBookingByUidOrRescheduleUid(`${rescheduleUid}`);
   }
 
+  await ssr.viewer.public.event.prefetch({ username, eventSlug: slug });
```
Comment (G3root) [inline apps/web/pages/new-booker/[user]/[type].tsx:95]: prefetch to prevent flash of title in the page

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -37,15 +37,19 @@ const useBrandColors = ({ brandColor, darkBrandColor }: { brandColor?: string; d
   useCalcomTheme(brandTheme);
 };
 
-const BookerComponent = ({ username, eventSlug, month, rescheduleBooking }: BookerProps) => {
+const BookerComponent = ({ username, eventSlug, month, rescheduleBooking, hideBranding }: BookerProps) => {
   const { t } = useLocale();
+  const { data: event, isLoading } = trpc.viewer.public.event.useQuery(
```
G3root, line 42: had to use the vanilla trpc query rather than the `useEvent` hook because prefetched data in the `serverSideProps` isn't reused here. maybe because of some race conditions with zustand , `useInitializeBookerStore` is called after the `useEvent` hook and both depends on each other

JeroenReumkens: The reason the new booker uses the `useEvent` hook is because it should do everything client side rather than server side. Soon the booker component will get published to NPM, and should then be able to run fully standalone, so without any serverside props passed in. Therefore we actually have to revert this change.

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -79,24 +83,47 @@ const BookerComponent = ({ username, eventSlug, month, rescheduleBooking }: Book
...
+  const profileImage = event?.profile?.image;
+  const title = event?.title ?? "";
   return (
     <>
       {/*
         If we would render this on mobile, it would unset the mobile variant,
         since that's not a valid option, so it would set the layout to null.
       */}
+
+      <HeadSeo
```
JeroenReumkens, line 108: Perhaps it makes most sense to do a separate query on the server (that only queries the SEO data?) and isn't reused on the client. That way we don't have a flash, but are also able to render the SEO data without querying unnecessary data on the server that then isn't used.

JeroenReumkens (changes requested):
Thanks a lot for your changes @G3root! I think we need to change a few things up though, since the component shouldn't depend on SSR data, but be able to run it's own queries. See my other review comments.

JeroenReumkens:
Hey @G3root 

I was looking at the changes, and I actually came up with something different that might make it even easier. 

* We move the HeadSEO to the pages (so `new-booker/[user]` and `new-booker/team/..`). That would be needed anyway because the booker component should not include any SEO meta data (since this is next specific, and the booker should be a standalone component that can be run anywhere). 
* Once we move it there, we can use the query directly in the page (trpc.usequery that is). And because of the prefetching that data is immediately set. 
* Then we can also revert the changes in the `useEvent` hook 
* Using `trpc.useQuery` in the page isn't a problem at all, since the only thing that page does is "enhance" the booker component with things like SEO data and for example fetching the booking data SSR in case of a reschedule. These aren't functionalities the booker component itself has, and thus not functionalities people can expect from the booker component when using it as a standalone component. So therefore it makes sense to use the trpc query in the page (but not in the booker component). 

Let me know if this makes sense 😁

G3root:
@JeroenReumkens  that makes sense and clean ❤️ . pushed some changes implementing requested changes.

JeroenReumkens (approved):
This is great @G3root  Looks super clean now indeed! The abstraction into a reusable component is also 👨‍🍳💋

LGTM!

PR #8713, opened 2023-05-05
issue#8601: fix

### LINKED ISSUE #8601: [CAL-1610] Every CalendarService must send externalCalendarId
This was causing reschedule to not update the Bigin Event. This might be a problem with other CRM apps as well.

_Originally posted by @hariombalhara in https://github.com/calcom/cal.com/pull/8158#discussion_r1181552060_

If a CalendarService doesn't set externalCalendarId, it's updateEvent never get's called.
This check exists here and I think we should do away with this requirement in `updateAllCalendarEvents`
https://github.com/calcom/cal.com/blob/810a4e6b7f829183dee8a798e08346f082cfb8f9/packages/core/EventManager.ts#L459 

cc @zomars

<sub>[CAL-1610](https://linear.app/calcom/issue/CAL-1610/every-calendarservice-must-send-externalcalendarid)</sub>

utkarsh2310:
## What does this PR do?

Removed throw error. Previously updateEvent is not getting called when externalId is undefined. Now it is fixed.

Fixes #8601

## Type of change

## Checklist


zomars (approved):

PR #8715, opened 2023-05-05
add try catch block in packages/email/templates/_base-email.ts getRec…

tony83033:
## What does this PR do?

Lack of error handling in the  /packages/email/templates/_base-email.ts -> getRecipientTime method: The dayjs library can throw errors if it is passed an invalid date string, but these errors are not being handled in the getRecipientTime method.

## Type of change


emrysal:
Ideally we'd let this error bubble up to the main error handler and handle there; not handling in the function itself; hope that makes sense 🙏

PR #8719, opened 2023-05-06
Hotfix: Can't use settings/admin/apps

hariombalhara:
Fixes unconnected appsRouter due to new per lambda routes related change

Bug can be seen in this screenshot. There is error fetching apps list due to which the apps list is empty.
Screenshot 2023-05-06 at 11 40 53 AM

Also, I think with the new changes, TypeScript safety isn't there with routes as this route was defined as per the TypeScript but it wasn't actually working.


sean-brydon (approved):
LGTM - Tested on preview and code works

Thanks @hariombalhara <3

PR #8725, opened 2023-05-06
chore: resolveEndpoint function is made readable

### LINKED ISSUE #8709: Update our trpc routes so they are more clear.
This function parses paths like the following and maps them

to the correct API endpoints.

- viewer.me - 2 segment paths like this are for logged in requests

- viewer.public.i18n - 3 segments paths can be public or authed

eslint-disable-next-line @typescript-eslint/no-explicit-any

https://api.github.com/calcom/cal.com/blob/cdba1920fce52deefeb314696f34049b7614cf7f/packages/trpc/react/trpc.ts#L45

```typescript

import type { inferRouterInputs, inferRouterOutputs, Maybe } from "../server";
import type { AppRouter } from "../server/routers/_app";

/**
 * We deploy our tRPC router on multiple lambdas to keep number of imports as small as possible
 * TODO: Make this dynamic based on folders in trpc server?
 */
const ENDPOINTS = [
  "apiKeys",
  "appRoutingForms",
  "apps",
  "auth",
  "availability",
  "bookings",
  "deploymentSetup",
  "eth",
  "eventTypes",
  "features",
  "insights",
  "payments",
  "public",
  "saml",
  "slots",
  "teams",
  "users",
  "viewer",
  "webhook",
  "workflows",
] as const;
export type Endpoint = (typeof ENDPOINTS)[number];

// eslint-disable-next-line @typescript-eslint/no-explicit-any
const resolveEndpoint = (links: any) => {
  // TODO: Update our trpc routes so they are more clear.
  // This function parses paths like the following and maps them
  // to the correct API endpoints.
  // - viewer.me - 2 segment paths like this are for logged in requests
  // - viewer.public.i18n - 3 segments paths can be public or authed
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  return (ctx: any) => {
    const parts = ctx.op.path.split(".");
    let endpoint;
    let path = '';
    if (parts.length == 2) {
      endpoint = parts[0] as keyof typeof links;
      path = parts[1];
    } else {
      endpoint = parts[1] as keyof typeof links;
      path = parts.splice(2, parts.length - 2).join('.');
    }

    return links[endpoint]({ ...ctx, op: { ...ctx.op, path } });
  };
};

/**
 * A set of strongly-typed React hooks from your `AppRouter` type signature with `createTRPCReact`.
 * @link https://trpc.io/docs/v10/react#2-create-trpc-hooks

```

ManVTish:
## What does this PR do?

This PR is an attempt to refactor the resolveEndpoint function which maps the path to correct endpoints, and make it more readable.

## Type of change

- Chore (refactoring code,)


packages/trpc/react/trpc.ts
```diff
@@ -43,22 +43,27 @@ export type Endpoint = (typeof ENDPOINTS)[number];
...
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const resolveEndpoint = (links: any) => {
-  // TODO: Update our trpc routes so they are more clear.
   // This function parses paths like the following and maps them
   // to the correct API endpoints.
   // - viewer.me - 2 segment paths like this are for logged in requests
   // - viewer.public.i18n - 3 segments paths can be public or authed
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   return (ctx: any) => {
+    // split the segmented path into an array
```
hariombalhara, line 52: I don't think this comment is needed. `split` makes it obvious that we are converting into an array.

packages/trpc/react/trpc.ts
```diff
@@ -43,22 +43,27 @@ export type Endpoint = (typeof ENDPOINTS)[number];
...
     const parts = ctx.op.path.split(".");
-    let endpoint;
+
+    // type-safe for ensuring the keys are form ENDPOINTS
+    let endpoint: keyof typeof links;
```
hariombalhara, line 56: links is of type `any`, so I don't think this makes a difference.

packages/trpc/react/trpc.ts
```diff
@@ -43,22 +43,27 @@ export type Endpoint = (typeof ENDPOINTS)[number];
...
   // to the correct API endpoints.
   // - viewer.me - 2 segment paths like this are for logged in requests
   // - viewer.public.i18n - 3 segments paths can be public or authed
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   return (ctx: any) => {
+    // split the segmented path into an array
     const parts = ctx.op.path.split(".");
-    let endpoint;
+
+    // type-safe for ensuring the keys are form ENDPOINTS
```
hariombalhara, line 55: I don't think it works because links is of type any

hariombalhara (changes requested):
Let's remove unnecessary comments.

Also, I don't think it fixes the issue #8709. That is something else.

packages/trpc/react/trpc.ts
ManVTish: hmm...I guess I wasn't focusing on that part.

packages/trpc/react/trpc.ts
```diff
@@ -43,22 +43,27 @@ export type Endpoint = (typeof ENDPOINTS)[number];
...
-      endpoint = parts[0] as keyof typeof links;
-      path = parts[1];
+      // assign first segment to endpoint & second segment to path
+      [endpoint, path] = parts;
     } else {
-      endpoint = parts[1] as keyof typeof links;
-      path = parts.splice(2, parts.length - 2).join(".");
+      // assign second segment to endpoint & the rest to path
+      endpoint = parts[1];
+      path = parts.slice(2).join(".");
```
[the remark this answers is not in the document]
ManVTish: Thanks.

ManVTish:
I understand the ENDPOINT needs to be made dynamic, but that's opened as different issue. However, could you give me some hint on what needs to be addressed?
Thank you @hariombalhara for reviewing my PR. Really appreciate your detailed feedback.

keithwillcode:
@ManVTish When cold starts were fixed as part of #8041, a larger refactor to remove "viewer" from all of the UI code was not done. Hence, this endpoint parsing code was introduced. The ultimate fix is to actually update all of the UI components that call tRPC and point them directly to the new tRPC routes. However, that's a much larger change. Your refactor here is still helpful but we need to leave the TODO in place.

keithwillcode:
This PR is being closed due to inactivity. Please reopen if work is intended to be continued.

PR #8729, opened 2023-05-06
implemented async using queue data structure

### LINKED ISSUE #8345: This should be done async probably using a queue.
https://api.github.com/calcom/cal.com/blob/ae22246f05ad1917da6bc6786d29a3ac50018453/packages/trpc/server/routers/viewer/apps.tsx#L151

```javascript

            },
          });

          // TODO: This should be done async probably using a queue.
          Promise.all(
            appCredentials.map(async (credential) => {
              // No need to continue if credential does not have a user
              if (!credential.user || !credential.user.email) return;

              const locale = credential.user.locale ?? "en";
              let t = translations.get(locale);

              if (!t) {
                t = await getTranslation(locale, "common");
                translations.set(locale, t);
              }

              await sendDisabledAppEmail({
                email: credential.user.email,
                appName: appMetadata?.name || app.slug,
                appType: app.categories,
                t,
              });
            })
          );
        } else {

```

ManVTish:
## What does this PR do?

Fixes #8345

This PR implements a queue data structure to sequentially process each credential rather than using Promise.all. However the logic is kept same for sendDisabledAppEmail process.

## Type of change

### DIFF
--- a/packages/trpc/server/routers/viewer/apps/toggle.handler.ts
+++ b/packages/trpc/server/routers/viewer/apps/toggle.handler.ts
@@ -69,11 +69,31 @@ export const toggleHandler = async ({ input, ctx }: ToggleOptions) => {
         },
       });
 
-      // TODO: This should be done async probably using a queue.
-      Promise.all(
-        appCredentials.map(async (credential) => {
+      class Queue<T> {
+        private items: T[] = [];
+
+        enqueue(item: T) {
+          this.items.push(item);
+        }
+
+        dequeue(): T | undefined {
+          return this.items.shift();
+        }
+
+        isEmpty(): boolean {
+          return this.items.length === 0;
+        }
+      }
+
+      const processQueue = async (queue: Queue<(typeof appCredentials)[0]>) => {
+        while (!queue.isEmpty()) {
+          const credential = queue.dequeue();
+
+          // no need to continue if credential is not present
+          if (!credential) continue;
+
           // No need to continue if credential does not have a user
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/apps/toggle.handler.ts
```diff
@@ -69,11 +69,31 @@ export const toggleHandler = async ({ input, ctx }: ToggleOptions) => {
...
+        }
+
+        dequeue(): T | undefined {
+          return this.items.shift();
+        }
+
+        isEmpty(): boolean {
+          return this.items.length === 0;
+        }
+      }
```
sean-brydon, line 86: Should we extract Queue so we can use it elsewhere in the code if required? Feels a bit weird just hardcoding this class here imho

ManVTish: Thanks @sean-brydon for reviewing my PR. Yeah, I definitely agree with you. Will make that change, np. Thanks.

ManVTish: Hey @sean-brydon would like to suggest where to extract the Queue class, or should I go ahead raising a PR and get it reviewed?

emrysal:
Thanks for this @ManVTish - however the comment was a bit vague on what should be made async. the async bit is talking about queueing emails rather than the call to the email. Sorry about the confusion, this logic should remain promise.all but the email system in entirety should be async.

PR #8739, opened 2023-05-07
Updated The Slugify to accept username as emoji

### LINKED ISSUE #7718: When using EMOJI as username Avatar breaks 💀
### Issue Summary

When the username is set to an emoji, the user's avatar fails to load.
Looking at the network calls, it seems that the avatar is downloaded twice, with the version from Gravtar failing to load.

This could be due to the encoding of the emoji to gravtar url

### Steps to Reproduce

1. Set username to an emoji
2. Avatar in account/public page is no longer loading

Only tested with single character (emoji) username https://cal.com/🍕
Has not been tested with usernames with mixed alphanumeric + emoji (e.g. McPizza🍕)

### Technical details

N/A - applies across browsers

_`P.S. Very low priority`_

pranjalg13:
## What does this PR do?
Added changes to accept emoji as username
- Previously it updates the emoji as (--) instead of the actual emoji
- Added a regex in which the u flag enables Unicode matching, and the g flag enables global matching (i.e., all matches are replaced, not just the first one).

Fixes #7718 

## Type of change

## How should this be tested?

- [ ] Before code changes the emoji username where updated with -- value
- [ ] After this change it is updated according to the emoji provided and /avatar/emoji path is also working fine

## Changes Screenshot
Screenshot 2023-05-07 at 7 55 01 PM
Screenshot 2023-05-07 at 7 58 09 PM


pranjalg13:
@PeerRich @zomars please review this once
Thanks

sean-brydon:
Thanks for your contribution. I'm gonna close this in favour of https://github.com/calcom/cal.com/pull/8740 as also improves functionality of our slugify function

pranjalg13:
Cool. no issues @sean-brydon

PR #8740, opened 2023-05-07
Extend slugify function

### LINKED ISSUE #7718: When using EMOJI as username Avatar breaks 💀

AyushMainali123:
## What does this PR do?

Closes #7718 

The slugify function appears to have limited functionality, which led to issue #7718. Additionally, I observed that the function does not trim multiple spaces in the slug.

For example:
Previously, `ayush      mainali` was updated to `ayush------mainali`.
`test test test 👩🏽👨🏽 漢`  would be converted to `test-test-test-----------`, which is quite confusing.

Now,
`ayush      mainali` will be converted to  `ayush-mainali`,
`test test test 👩🏽👨🏽 漢` will be converted to `test-test-test-👩🏽👨🏽-漢`

I didn't use something like:
`/[^a-zA-Z0-9-\p{Emoji}]/gu` because, it couldn't properly slugify combined emoji like 👪 . It would return something like `👨-👩-👦` when slugified.

Fixes # (7718 )
https://user-images.githubusercontent.com/49358949/236681331-41a7cd2d-7334-4291-a5ff-30ebc8bc8936.mp4

## Type of change

## How should this be tested?

- [x]   Register a user and add a username with a string including multiple spaces and emojis. The output will be a properly slugified username.
- [x]   Update username, the text will be properly slugified here too.

## Checklist

I have checked test suites using yarn run test (yarn test-e2e was testing against 0 test suites in my local) . 

Test
` Weekday tests › fn: weekdayNames › should return the weekday names for a given locale and format` was failing in the production branch, so it isn't addressed in my PR.

Except for that test case, all other test case has passed in the PR I submitted.


sean-brydon (commented):
Tested - LGTM

Will wait for a few others to test tho as this is core code. 

@zomars @keithwillcode <3

sean-brydon:
@PeerRich @baileypumfleet are we actually wanting to support emojis in usernames 🤔

zomars (approved):
Definitively an improvement. Thank you for your contribution 🙏

zomars:
>  we actually wanting to support emojis in usernames

We could discuss this in a follow up

PR #8747, opened 2023-05-08
disable standard confirmation emails if workflow exists

### LINKED ISSUE #3680: [CAL-1424] Disable standard confirmation email in even type settings

CarinaWolli:
## What does this PR do?

Allow disabling standard confirmation emails for host and attendees if event type is active on a workflow that will send these emails instead. 

The trigger EMAIL_ATTENDEE now sends emails to all attendees not just the first one. 

Switches to disable emails can be found in advanced event type settings:
Screenshot 2023-05-08 at 13 13 14

Fixes #3680

Loom: https://www.loom.com/share/f569f130746748ffaa6481a9536ba2e3


ciaranha (commented):
Looks great 👏

Should the copy read the other way around in terms of what gets applied to what? ie workflow applied to event type rather than event type applied to workflow? I know technically in the workflow we do select the event type, but feels to me like a workflow to event type relationship.

CleanShot 2023-05-08 at 16 45 23@2x

CarinaWolli:
Updated text:
Screenshot 2023-05-09 at 08 55 51

Udit-takkar (approved):
LGTM. tested locally

PeerRich:
@CarinaWolli i mentioned this before, should the checkbox not be visible inside the workflows area?

PeerRich:
ok merging when ready

PR #8761, opened 2023-05-08
Disable reserve slot on seated events

joeauyeung:
## What does this PR do?

This PR fixes a bug where the reserve slots function was blocking some users booking on seated events. The reserve slots logic should only run on the last seat available.

https://www.loom.com/share/ec793dff3da8434fabf783bac6158706

Fixes # (issue)

## Type of change

## How should this be tested?

- Create a seated event
- In one window, choose a time slot
- In a second window, choose the same timeslot
    - You should notice that the same slot doesn't say it has any seats taken
    - Book that time slot, it should succeed
- Now book seats on that time slot until there is one seat left
- When choosing to book that slot, on another window it should say the booking is full

## Checklist

### DIFF
--- a/packages/trpc/server/routers/viewer/slots/reserveSlot.handler.ts
+++ b/packages/trpc/server/routers/viewer/slots/reserveSlot.handler.ts
@@ -52,7 +52,6 @@ export const reserveSlotHandler = async ({ ctx, input }: ReserveSlotOptions) => 
 
   if (eventType && shouldReserveSlot) {
     try {
-      console.log("This triggers");
       await Promise.all(
         eventType.users.map((user) =>
           prisma.selectedSlots.upsert({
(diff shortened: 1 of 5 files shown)


packages/trpc/server/routers/viewer/slots/reserveSlot.handler.ts
```diff
@@ -19,43 +19,68 @@ interface ReserveSlotOptions {
   input: TReserveSlotInputSchema;
 }
 export const reserveSlotHandler = async ({ ctx, input }: ReserveSlotOptions) => {
+  console.log("🚀 ~ file: reserveSlot.handler.ts:22 ~ reserveSlotHandler ~ input:", input);
```
alannnc, line 22:
```suggestion
```

alannnc (approved):

packages/trpc/server/routers/viewer/slots/reserveSlot.handler.ts
```diff
@@ -21,41 +21,65 @@ interface ReserveSlotOptions {
...
+  if (!eventType) {
     throw new TRPCError({
       message: "Event type not found",
       code: "NOT_FOUND",
     });
   }
+
+  let shouldReserveSlot = true;
+
+  // If this is a seated event then don't reserve a slot
```
alannnc, line 40:
```suggestion
  // If this is a seated event figure out if this is the last seat being offered
```

PR #8769, opened 2023-05-09
fix: strip markdown from meta description tag in users page

### LINKED ISSUE #8745: [CAL-1627] Social share previews don't render ampersands
Created via Threads. See full discussion: [https://threads.com/34470827863](https://threads.com/34470827863)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1627](https://linear.app/calcom/issue/CAL-1627/social-share-previews-dont-render-ampersands)</sub>

G3root:
## What does this PR do?

remove markdown from description meta tags in user page and also closes  #8745

Fixes 

## How should this be tested?

1. add a markdown content to your bio
2. go to your public profile page
3. check the description html meta tag, markdown should be stripped


apps/web/pages/api/social/og/image.tsx
```diff
@@ -75,12 +74,10 @@ export default async function handler(req: NextApiRequest) {
         imageType,
       });
 
-      const title_ = md.render(title).replace(/(<([^>]+)>)/gi, "");
```
G3root, line 78: we don't need this as we are stripping the markdown content in the user's public profile page

JeroenReumkens:
Hey @G3root, 

Using remark for this seems like a very big dependency. Even though this runs on the server, it will still load this big dependency on the server and slow things down unnecessary. Are there any other options that could work as well without adding a big dependency? Perhaps https://github.com/stiang/remove-markdown will work already? Or maybe a simple regex would be sufficient since there's only a few markdown chars we can add? 

On top of that I'd suggest adding this as a util to the lib package, similar to `markdownToSafeHTML`, this way we can reuse the `markdownToPlainText` utility easily and prevent other people from making their own solution. 

Thanks 🫶

JeroenReumkens (changes requested):
See other comment.

G3root:
@JeroenReumkens pushed some changes replacing remark

JeroenReumkens:
Looks great to me @G3root 
Could you also please merge main in and add this to the new booker's meta when #8697 is auto-merged? After that I'm ready to approve 🙏

G3root:
@JeroenReumkens new bookers meta won't need this right ?

JeroenReumkens (approved):
LGTM, see other comments <3

PR #8787, opened 2023-05-09
fix: use time format

### LINKED ISSUE #8746: Wrong time format in emails and SMS
I noticed that the time format (and partly the German date format) is no longer correct for the emails and SMS. I have set the 24 hour format in my settings, but still the 12 hour format is displayed.

**So far I have been able to locate the problem in these places:**.
1. E-mails confirming the rescheduling of an appointment (in the subject of the mail) - the German date format does not work here either (in the subject)!
2. E-mails confirming the cancellation of an appointment (in the subject, as well as in the content of the mail) - the German date format does not work here either (in the subject)!
3. Workflow e-mails (everywhere where the variables for the start time or end time of the appointment are used (in the subject, as well as in the content of the mail))
4. Workflow SMS (everywhere where the variables for the start time or end time of the appointment are used)

**Screenshot of e-mail confirming the rescheduling of an appointment:** IMG_9604

**Screenshot of email confirming the cancellation of an appointment:** IMG_9605

**Screenshot Workflow E-Mail:** IMG_9602

**Screenshot Workflow SMS:** IMG_ABF3C896ECF3-1

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/8746 ( 1 and 2)

Screenshot 2023-05-09 at 11 09 22 PM

## Type of change


packages/emails/templates/attendee-daily-video-download-recording-email.ts
```diff
@@ -43,16 +44,32 @@ export default class AttendeeDailyVideoDownloadRecordingEmail extends BaseEmail
...
+  protected getLocale(): string {
+    return this.attendee.language.locale;
+  }
+
   protected getInviteeStart(format: string) {
-    return this.getRecipientTime(this.calEvent.startTime, format);
+    return this.getFormattedRecipientTime({
+      time: this.calEvent.startTime,
+      format,
+      locale: this.getLocale(),
```
hariombalhara, line 55: If locale and timezone are on the instance then can we avoid passing these and calculate in `getFormattedRecipientTime` itself?

packages/emails/templates/organizer-scheduled-email.ts
```diff
@@ -120,17 +121,34 @@ ${callToAction}
...
+    return this.getFormattedRecipientTime({
+      time: this.calEvent.endTime,
+      format,
+      locale: this.getLocale(),
+      timeZone: this.getTimezone(),
+    });
   }
 
   protected getFormattedDate() {
-    return `${this.getOrganizerStart("h:mma")} - ${this.getOrganizerEnd("h:mma")}, ${this.t(
```
hariombalhara, line 132: getFormattedDate seems to be using the same logic in all classes. Maybe we can extract out the logi in a `getFormattedDateRange(start, end)` kind of thing.

Suitable for a followup I guess.

hariombalhara (dismissed):
Left a comment. Will test after that. This is a critical flow. So, testing is super important.

packages/emails/templates/organizer-scheduled-email.ts
LunaticProgrammer: also this function maybe be moved to **packages/lib/timeFormat.ts**

packages/emails/templates/attendee-daily-video-download-recording-email.ts
Udit-takkar: we can't calculate this in getFormattedRecipientTime because locale and timezone is different in both organiser and attendee . For eg. this.attendee.language.locale and this.calEvent.organizer.language.locale.

packages/emails/templates/organizer-scheduled-email.ts
Udit-takkar: okay will do this in a follow up PR

shivamklr (dismissed):
Tested locally, works as expected.

Udit-takkar:
Fixed type error

keithwillcode:
@shivamklr Could you test this one on a preview build before we merge to main?

shivamklr (approved):
Tested fix. Feel free to review the code.

PeerRich:
merged accidentially, @shivamklr @hariombalhara wanna test again once this hits app.cal.dev?

PR #8790, opened 2023-05-09
fix: available time slot picker

### LINKED ISSUE #8403: Slot picker, disable button after selection is made
Could we get the slot picker buttons to be disabled once an option is clicked? 

Right now, it’s possible to click multiple selections and the last one selected is chosen, not ideal if there is a bit of lag to process the selection:

https://user-images.githubusercontent.com/126418022/233367338-04b93ca5-b8cd-4581-93e2-ccba070711d9.mov

ExplorerAadi:
## What does this PR do?

Fixes #8403 

A little explaination of my changes:
I used a local state instead of the state from reservationSlotMutation because there is still some delay between the mutation completes and the links takes the user to the next page and during that delay, they can still select a different time slot by mistake.

 Loom Video: https://www.loom.com/share/2075383e6995474aab7310717c99ada5

## Type of change

## How should this be tested?

- [ ] Try booking an event

## Checklist

### DIFF
--- a/apps/web/components/booking/AvailableTimes.tsx
+++ b/apps/web/components/booking/AvailableTimes.tsx
@@ -53,6 +53,9 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
   const { rescheduleUid } = router.query;
 
   const [brand, setBrand] = useState("#292929");
+  const [isSlotReservationPending, setIsSlotReservationPending] = useState(false);
+
+  const didSlotReservationFail = reserveSlotMutation.isError;
 
   useEffect(() => {
     setBrand(getComputedStyle(document.documentElement).getPropertyValue("--brand-color").trim());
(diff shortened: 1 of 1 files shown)


apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -161,7 +173,8 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
                         prefetch={false}
                         className={classNames(
                           " bg-default dark:bg-muted border-default hover:bg-subtle hover:border-brand-default text-emphasis mb-2 block rounded-md border py-2 text-sm font-medium",
-                          brand === "#fff" || brand === "#ffffff" ? "" : ""
+                          brand === "#fff" || brand === "#ffffff" ? "" : "",
+                          isSlotReservationPending ? "pointer-events-none opacity-50" : ""
```
roae, line 177: According to Cal's design system this should be the styles of the disable state of the buttons.

```suggestion
                          reserveSlotMutation.isLoading ? "cursor-not-allowed opacity-40" : ""
```

the disabled attribute must be added.

`disabled={reserveSlotMutation.isLoading}`

roae (changes requested):
We could simplify this, leave a few comments, thanks @ExplorerAadi 🙏

roae:
@JeroenReumkens This will be useful for the new booking page?

apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -53,6 +53,9 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
   const { rescheduleUid } = router.query;
 
   const [brand, setBrand] = useState("#292929");
+  const [isSlotReservationPending, setIsSlotReservationPending] = useState(false);
+
+  const didSlotReservationFail = reserveSlotMutation.isError;
```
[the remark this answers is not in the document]
ExplorerAadi: Hey @roae , Thanks for the input and yes I tried that initially but it didn't work as I explained in the description of the PR at the top:
"I used a local state instead of the state from reservationSlotMutation because there is still some delay between the mutation completes and the links takes the user to the next page and during that delay, they can still select a different time slot by mistake."

apps/web/components/booking/AvailableTimes.tsx
ExplorerAadi: Yes, I tried this but `next/link` doesn't take disabled prop that is why I had to use pointer events to prevent user from redirecting to a different link while reservation mutation is happening.

JeroenReumkens:
@roae In the new booker we combined both views (slot picker + form) into a single view / route. So there's no delay between showing these two views. So we don't need this actually 😁

PeerRich:
the old booker has been deprecated. thank you for your contribution but we focus engineering now on the new booking page!

PR #8798, opened 2023-05-10
Fix booking flow crash when zoho-bigin app is enabled.

hariombalhara:
Add missing zoho-bigin import which was crashing the booking flow when the app is enabled.

Regression from https://github.com/calcom/cal.com/pull/8419/files#diff-8d0187f3e8930f70c4a3baae2905502d659e5bbadcf136e011d792c4ea7487e9L18

### DIFF
--- a/packages/app-store/_utils/getCalendar.ts
+++ b/packages/app-store/_utils/getCalendar.ts
@@ -12,7 +12,14 @@ export const getCalendar = async (credential: CredentialPayload | null): Promise
   if (calendarType?.endsWith("_other_calendar")) {
     calendarType = calendarType.split("_other_calendar")[0];
   }
-  const calendarApp = await appStore[calendarType.split("_").join("") as keyof typeof appStore]();
+  const calendarAppImportFn = appStore[calendarType.split("_").join("") as keyof typeof appStore];
+
+  if (!calendarAppImportFn) {
+    log.warn(`calendar of type ${calendarType} is not implemented`);
+    return null;
+  }
+
+  const calendarApp = await calendarAppImportFn();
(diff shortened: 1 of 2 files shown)


packages/app-store/index.ts
```diff
@@ -28,6 +28,7 @@ const appStore = {
   exchangecalendar: () => import("./exchangecalendar"),
   facetime: () => import("./facetime"),
   sylapsvideo: () => import("./sylapsvideo"),
+  "zoho-bigin": () => import("./zoho-bigin"),
```
hariombalhara, line 31: It probably got removed during merge conflict resolution. https://github.com/calcom/cal.com/pull/8419/files#diff-8d0187f3e8930f70c4a3baae2905502d659e5bbadcf136e011d792c4ea7487e9L18

packages/app-store/_utils/getCalendar.ts
```diff
@@ -12,7 +12,14 @@ export const getCalendar = async (credential: CredentialPayload | null): Promise
   if (calendarType?.endsWith("_other_calendar")) {
     calendarType = calendarType.split("_other_calendar")[0];
   }
-  const calendarApp = await appStore[calendarType.split("_").join("") as keyof typeof appStore]();
+  const calendarAppImportFn = appStore[calendarType.split("_").join("") as keyof typeof appStore];
+
+  if (!calendarAppImportFn) {
```
hariombalhara, line 17: This ensures that a missing import for an app doesn't crash the booking flow.
No app should be important enough to crash the core booking flow.

roae (approved):
🚀

PR #8812, opened 2023-05-10
feat: add infinite query for team members search

### LINKED ISSUE #8750: [CAL-1629] Fuzzy matching while searching

we have been implementing search functionality here https://github.com/calcom/cal.com/pull/8742 and https://github.com/calcom/cal.com/pull/8659/files. 

It would be much better experience if we support fuzzy searching which would give us similar results related to search query so that we see relevant result even when we misspell something

Might be relevant https://github.com/prisma/prisma/issues/7986

or we can also look into Full text seach feature in prisma https://www.prisma.io/dataguide/managing-databases/intro-to-full-text-search

G3root:
## What does this PR do?

Fixes #8750. this PR adds infinite query for team members list and server side search for team members. previously the search was handled in client side.

## Type of change


zomars (changes requested):
@G3root lets add more context in the PR description. also let's keep the "How this PR should be tested" section. 🙏🏽

PeerRich:
@zomars @G3root where are we here? Some merge conflicts occured

packages/trpc/server/routers/viewer/teams/search.handler.ts
```diff
@@ -0,0 +1,93 @@
...
+          bio: true,
+        },
+      },
+      teamId: true,
+      userId: true,
+    },
+    where: {
+      teamId,
+      ...(search &&
+        search !== "" && {
```
hussamkhatib, line 41: isn't `search !== ""` redundant ?
`""` is a falsy value
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

PR #8827, opened 2023-05-10
Fix/ Routing Form Routes

hariombalhara:
Routing Form Routes were split here #8419 but they weren't used in _router.ts


joeauyeung (commented):
I noticed the endpoints do not follow the same format as the other `_router` files. Is this optional for routing forms?

image

hariombalhara:
Thanks @joeauyeung. Added caching logic as well. I didn't add that initially because that's a lot of repetitive code. Have abstracted out the caching logic now.
cc @roae to verify if it's doing what we expect. Also, while making the changes, I was thinking that we are probably trying to save import cost here but isn't it something that Next.js should be doing already?

roae:
> Thanks @joeauyeung. Added caching logic as well. I didn't add that initially because that's a lot of repetitive code. Have abstracted out the caching logic now.
> 
> cc @roae to verify if it's doing what we expect. Also, while making the changes, I was thinking that we are probably trying to save import cost here but isn't it something that Next.js should be doing already?

I'm not sure if Next does anything to handle this, but what I am sure of some dependencies took hundreds of ms even on a hot lambda.

Maybe the cache should be applied only in some handlers with large dependencies, but for the moment we are applying it to all handlers to remove doubts about the
Performance

Maybe @Mythie can tell us more about this.

hariombalhara:
@roae @joeauyeung  Changes are done wrt caching as well

roae (approved):
LGTM!

PR #8828, opened 2023-05-10
fix: not able to reschedule an event/meeting when new time slot overlaps with old selected time slot

### LINKED ISSUE #2668: [CAL-1685] "Edit booking" doesn't allow a new time which overlaps old time
### Issue Summary

Someone booked a 60 minute slot via my cal.com event link, but then pointed out they wanted to move it 15 minutes later; however the "edit booking" feature doesn't allow this.

### Steps to Reproduce

0. Confirm a 60 minute booking using an event type which has `Time-slot intervals` set to 15 minutes.  (In my case, the booking was at 11:30am.)
1. Go to https://app.cal.com/bookings/upcoming
2. Click on "Reschedule"
3. Click on "Edit booking"
4. Select the same day as the original booking time
5. Observe that the times available for rebooking don't overlap with the original time, so I can't shift the original 60 minute booking at 11:30am earlier or later by 15 or 30 or 45 minutes:

image

### Expected behaviour

When rescheduling, the availability algorithm should consider the existing booking as (provisionally) cancelled just for the purposes of editing its time, in order to avoid the old time incorrectly overlapping with potential new times.

### Technical details

Google Chrome Version 101.0.4951.41 (Official Build) (64-bit)

This is clearly a server-side issue, but happy to provide more client details if someone needs them.

<sub>[CAL-1685](https://linear.app/calcom/issue/CAL-1685/edit-booking-doesnt-allow-a-new-time-which-overlaps-old-time)</sub>

asadath1395:
## What does this PR do?

Fix not able to reschedule an event/meeting when new time slot overlaps with old selected time slot

## Type of change

## How should this be tested?

- Create a new event type of length 60 minutes
- Use `Time-slot intervals` as 15 minutes and save the event
- Now open the event link and schedule a booking
- Now try to re-schedule the event and verify the time slot near to the previously selected time slot is available

## Checklist


joeauyeung (commented):
When rescheduling, available slots overlapping the original booking still do not show up. For example, I have a booking at 11:30 so I should expect to see 11:45, 12, etc

image

asadath1395:
@joeauyeung I tested this again and i can see it working as expected in my local. Mind sharing the event details you created? Also i tested this from 2 places, one using the link as an attendee and the other as a creator. Is there any other place where we can modify this?

CarinaWolli (changes requested):
It is throwing an error when I am trying to book the same slot when rescheduling: 

Screenshot 2023-05-11 at 08 29 54

Screenshot 2023-05-11 at 08 30 30

joeauyeung:
> It is throwing an error when I am trying to book the same slot when rescheduling:
> Screenshot 2023-05-11 at 08 29 54 Screenshot 2023-05-11 at 08 30 30

@asadath1395 I was able to get it working. In terms of @CarinaWolli's comment, we should hide the slot of the booking we're rescheduling in order to reduce any confusion.

asadath1395:
@CarinaWolli @joeauyeung I have added the fix to remove current schedule's time slot from the list. There is another problem which i noticed, my fix works fine if the user doesn't have any calendar linked but fails if he has any calendar linked. It happens because when any user books a slot, immediately a booking is marked in organiser's calendar and i couldn't find a way to distinguish whether the booking came from cal or the user had the slot marked in some other way. Let me know if i am not clear

joeauyeung:
> @CarinaWolli @joeauyeung I have added the fix to remove current schedule's time slot from the list. There is another problem which i noticed, my fix works fine if the user doesn't have any calendar linked but fails if he has any calendar linked. It happens because when any user books a slot, immediately a booking is marked in organiser's calendar and i couldn't find a way to distinguish whether the booking came from cal or the user had the slot marked in some other way. Let me know if i am not clear

Good job testing this out @asadath1395. My first thought would be to grab the external calendar event id from the Cal booking reference. Then when checking busy times of external calendars, ignore the calendar event that matches the id from the Cal booking reference.

I hope that makes sense.

PeerRich:
hey gents, whats missing here to get merged?

CarinaWolli:
@PeerRich It's not yet working when there is a calendar connected 

@asadath1395 Did you look into @joeauyeung's suggestion?

CarinaWolli (changes requested):
Blocking until it's also working with connected calendars

PR #8832, opened 2023-05-11
CAL-1624: Timezone select layout improvements when typing

JeroenReumkens:
## What does this PR do?

* Removes the layout shift when starting to type in the timezone select
* Gives timezone select a minimum width to have even less movement (I don't really like to min width, see video below)
* Gave no results option some more breathing room, it was a very narrow row before. 

Fixes CAL-1624

https://github.com/calcom/cal.com/assets/2969573/64a5e1c1-8fe4-4a5c-884b-f19d46da6131

@PeerRich  The min-width is the only thing I have doubts about. I tried this before and wasn't really happy with the results. See the video. 
I could also make it full width always, but that also looks a bit awkward. Now I gave it a reasonable min width that still grows if needed. Mainly the arrow is bothering me, that this moves along. But when I move this to the right, it's always on the right side, also if you have selected an option. Not sure if that's better, see the second video below. **note:** This full width layout is NOT included in this PR. I could easily add it if we like it more. 

cc @Jaibles  

https://github.com/calcom/cal.com/assets/2969573/cfbbdb06-ef7b-4cef-8ef4-eb2444fde9d4

## Type of change

## How should this be tested?

- [ ] Test layout of the timezone select

### DIFF
--- a/packages/features/bookings/Booker/components/EventMeta.tsx
+++ b/packages/features/bookings/Booker/components/EventMeta.tsx
@@ -66,18 +66,20 @@ export const EventMeta = () => {
             <EventDetails event={event} />
             <EventMetaBlock
               className="cursor-pointer [&_.current-timezone:before]:focus-within:opacity-100 [&_.current-timezone:before]:hover:opacity-100"
-              contentClassName="relative"
+              contentClassName="relative max-w-[90%]"
               icon={Globe}>
               {bookerState === "booking" ? (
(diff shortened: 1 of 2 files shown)


sean-brydon (approved):
Happy with this <3 

Thank you!

sean-brydon (approved):
Happy with this <3 

Thank you!

PR #8834, opened 2023-05-11
CAL-1553: Show dates for multi occurence event

JeroenReumkens:
## What does this PR do?

Show dates for multi occurrence event instead of input to change occurrences, during booking step. 
Also makes form less wide and meta a bit wider during booker step (in small view only), so form looks better and dates fit on a single line for the multiple occurrences. When split over multiple lines they would otherwise take up a lot of unnecessary space.

Fixes CAL-1553 CAL-1562

https://github.com/calcom/cal.com/assets/2969573/fc8c600a-c79c-43fd-9329-3d3f77a2b145

## Type of change

## How should this be tested?

- [ ] Ensure all dates are shown for the multi occurrence event
- [ ] Ensure there's not empty occurrence list shown for a single occurrence event

### DIFF
--- a/packages/features/bookings/Booker/Booker.tsx
+++ b/packages/features/bookings/Booker/Booker.tsx
@@ -19,7 +19,7 @@ import { EventMeta } from "./components/EventMeta";
 import { LargeCalendar } from "./components/LargeCalendar";
 import { LargeViewHeader } from "./components/LargeViewHeader";
 import { BookerSection } from "./components/Section";
-import { fadeInLeft, useBookerResizeAnimation } from "./config";
+import { fadeInLeft, getBookerSizeClassNames, useBookerResizeAnimation } from "./config";
 import { useBookerStore, useInitializeBookerStore } from "./store";
 import type { BookerLayout, BookerProps } from "./types";
 import { useEvent } from "./utils/event";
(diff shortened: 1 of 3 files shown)


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -128,11 +128,8 @@ const BookerComponent = ({ username, eventSlug, month, rescheduleBooking }: Book
         <div
           ref={animationScope}
           className={classNames(
-            // Size settings are abstracted on their own lines purely for readbility.
-            "[--booker-main-width:480px] [--booker-timeslots-width:240px] lg:[--booker-timeslots-width:280px]",
-            layout === "small_calendar" && "[--booker-meta-width:240px] lg:[--booker-meta-width:280px]",
-            layout !== "small_calendar" && "[--booker-meta-width:340px] lg:[--booker-meta-width:424px]",
-            // Other styles
+            // Sets booker size css variables for the size of all the columns.
+            ...getBookerSizeClassNames(layout, bookerState),
```
JeroenReumkens, line 132: Decided to move this to the config file so the booker component itself becomes a bit cleaner. 
Also that way we can use some more comments in the config file, since they won't bother anyone there. See next file.

packages/features/bookings/components/event-meta/Occurences.tsx
```diff
@@ -22,6 +28,32 @@ export const EventOccurences = ({ event }: { event: PublicEvent }) => {
 
   if (!event.recurringEvent) return null;
 
+  if (bookerState === "booking" && recurringEventCount && selectedTimeslot) {
```
JeroenReumkens, line 31: These changes ensure that we show the selected recurring timeslots instead of the input field when the user is in the "booking" state.

sean-brydon (approved):

PR #8671, opened 2023-05-04
feat: firstAndLastName variant for name field

hariombalhara:
## What does this PR do?

Fixes #5784

**Other things that were implemented as part of the PR to make it a clean and stable implementation**
- Capability to define variants for a field type in a declarative way.
- Adds 2 variants to **name** type of field. 
	- `fullName` -> Default for all existing Event Types and any new one that is being created.
	- `firstAndLastName` -> Adds the fields in the same row on the booking page.
- Cleans up Form Builder with a small refactor.
	- This ticks off some of the tasks to reuse FormBuilder for Routing Forms
	- Should make it easier to implement `cancellationReason` using Form Builder which is a different form than Booking Form.
- Components required by Booker are separated away from other components required by the App. This gave an improvement in bundle size.
Screenshot 2023-05-05 at 2 53 45 PM

## Demo
https://www.loom.com/share/319379202da54e46a4f232d012807952

## Changes in Prefill Params
- `name` param can still be used to prefill `fullName` variant and thus there would be no change required to prefill all existing event-types' and any new event-types' booking pages by default. So the format would remain as `name=John Doe`
	- You can also prefill `fullName` variant  as`name={"fullName": "John Doe"}`
- If `firstAndLastName` variant is chosen then the two fields can be prefilled as `name={"firstName":"John", "lastName": "Doe"}`. 
	- 	It can also be prefilled as name=John Johny Janardan where **John** would become **firstName** and **John Janardan** would be **lastName**

## Changes in Webhook data
All listeners on webhooks for an existing event type would not have any change in data as those event-types would be using `fullName` variant which is the default.

For`firstAndLastName` variant the format would be`{"label":"your_name","value":{"firstName":"ABC","lastName":"DEF"}`

## Changes in Emails & Attendee Table
- As expected there would be no changes in the case of `fullName` variant in terms of how and what is shown for User name.
- For the `firstAndLastName` variant, the `attendee.name` is stored as `${firstName} ${lastName}` in Attendee table and thus email would also mention the name as the concatenation of two parts of the name.

## More Behaviours
- If a booking has been done for an event-type with `firstAndLastName` variant and later the event-type switches to use the `fullName` variant, the booking would not crash and will be able to handle the conversion of collected response from `firstAndLastName` format to `fullName` format. Vice versa is also true

## TODO
- [x] Test that enabling seats disable guests toggle.
[...]

PR #8840, opened 2023-05-11
feat: Make cancellationReason configurable using FormBuilder

### LINKED ISSUE #7780: [CAL-1290] Make it necessary to enter a reason for canceling the appointment
Your new update makes it possible to require the reason to reschedule the event. Can you do the same for the reason of cancellation of an event? So far, this is still completely optional.

Bildschirm­foto 2023-03-16 um 17 24 20

<sub>[CAL-1290](https://linear.app/calcom/issue/CAL-1290/make-it-necessary-to-enter-a-reason-for-canceling-the-appointment)</sub>

hariombalhara:
Fixes  #7780

TODO:
- [x] Merge #8671 first
- [x]  Figure out UI. 
	- I don't see a usecase of allowing to create user fields for cancellation form. So, we can show `cancellationReason` as part of a separate form in the Booking Questions, that won't have user fields support.


hariombalhara:
Closing it for now as it's low priority

PR #8865, opened 2023-05-12
chore: removed duplicate value in className @ Tooltip.tsx

MendyLanda:
## What does this PR do?
Removed a duplicate value from the tailwind classes at the Tooltip component.
Specifically "relative".


zomars (approved):

PR #8869, opened 2023-05-12
fix: add exp in /meeting-tokens

### LINKED ISSUE #8802: [CAL-1642] Generate new cal video link after rescheduling

### Issue Summary

 we need to close the old cal video and generate a new cal video because the old cal video link can get expired.

<sub>[CAL-1642](https://linear.app/calcom/issue/CAL-1642/generate-new-cal-video-link-after-rescheduling)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/8802

The issue was that the Cal video meeting URL would expire if someone rescheduled the meeting for a later time than the previously selected one.

exp denotes expiry timestamp https://docs.daily.co/reference/rest-api/rooms/set-room-config#exp.

## Type of change


apps/web/pages/_document.tsx
```diff
@@ -46,7 +46,7 @@ class MyDocument extends Document<Props> {
           <link rel="manifest" href="/site.webmanifest" />
           <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#000000" />
           <meta name="msapplication-TileColor" content="#ff0000" />
-          <meta name="theme-color" media="(prefers-color-scheme: light)" content="white" />
+          <meta name="theme-color" media="(prefers-color-scheme: light)" content="#f9fafb" />
```
PeerRich, line 49: @Udit-takkar hijacking your PR for this NIT 🙏

leog:
Changes look good :+1: although while testing, when rescheduling I get a new video link, in which case do we want to not expire old video link? @Udit-takkar just to understand the context

PeerRich (approved):

leog:
> @leog earlier when i tested on friday. it generated the same url when meeting was rescheduled and it's expiry time was updated and same url was valid for that much time instead of getting expired.
> 
> But now that i test it again i also get a new meeting url and both of the urls ( earlier and new one) are valid till expiry time( i don't think we require old meeting url if we are getting a new one but both of them works till the new endTime). Although this still solves the bug as i get new meeting url which is valid till the new end Time and the url is also updated in the calendar integration.


Ok, good to know, and agree with you @PeerRich. The changes are still a good improvement so it's good to merge from my POV

PR #8879, opened 2023-05-13
[WIP] feat: File Upload (Amazon S3) app

### LINKED ISSUE #6939: [CAL-1011] App: File Upload (Starting with Amazon S3)
starting with S3, organisers can ask guests for a file upload (i.e. a CV or similar)

for security reasons, organisers need to provide their own S3 bucket since [Cal.com](http://Cal.com) does not want to host your files

this app would be of type "EventTypeApp"

in event-types:

booking page:

questions:

should this be always optional? or always required? or ability to decide?

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1011](https://linear.app/calcom/issue/CAL-1011/app-file-upload-starting-with-amazon-s3)</sub>

miguelnietoa:
## What does this PR do?

Maybe this is not the best implementation but I decided to give it a try since I didn't see any progress 😄 

Fixes #6939
/claim #6939

https://www.loom.com/share/b8a0d309ab404d1d990086bc53cbef58

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


rjackson:
I’ve been similarly interested in this app due to the lack of progress on the issue 😄

One major issue that I noticed with the idea for the app, that I think your implementation also suffers from, is that anything stored in ‘appData’ (or the ‘metadata’ column in the database, if I remember correctly) is stored in plaintext and exposed to the browser.

For most apps I don’t think that’s a problem, for this app - where we’re storing third party Amazon API keys - it’s a big one. Any bookee could exfiltrate these API keys via “View Source” 😬

‘appKeys’ might be a better place to store these keys, but then they’ll be shared across all instances of the file upload app; not varied per event type, and not as easy to manage from the UI.

I’ve been thinking about wider changes to make a public/private split of ‘appData’, where private entries are encrypted at rest and not exposed to browsers at all; they can only be used server-side (app settings card being the only, authenticated, exception of course).

Thought I’d best flag this in case you have any thoughts, or perhaps you’ve already solved this and I just haven’t spotted how. Happy to keep exploring my thoughts on this and work together on a solution if that might be helpful?

miguelnietoa:
Hey @rjackson 👋🏻 

I totally agree with you. The appData being stored in plaintext is currently a significant security issue.

I just uploaded the PR to ask for feedback, and of course, we could work together!

I think I have a solution, which is to use `symmetricEncrypt` and `symmetricDecrypt`. 
The API Keys are encrypted/decrypted using the `CALENDSO_ENCRYPTION_KEY` env var, and that's it.

This way the API Keys are stored directly in the eventType metadata but encrypted. What do you think about this idea?

hariombalhara:
Just thinking out loud, haven't gone into much detail but saw the comment and thought to jump in.
There are credentials per user that are stored in the `Credential` table. In this case, we want to store Credential per EventType, maybe we can update the Credential schema to accommodate that.

There is one drawback with that the size of Credential was earlier increasing by per user, now it will increase by per user X eventType.

We can also think about creating a different Credential table just for EventType credentials. That can be useful because the EventTypeCredentials table would be pretty lightweight and the queries would be faster on it which we might need during the booking flow(keeping in mind that we want to keep that flow blazingly fast).

cc @keithwillcode @emrysal @zomars

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 16.17.0
```
PeerRich, line 1: what is this?

PR #8903, opened 2023-05-15
remove todo to issue action

PeerRich:


emrysal (approved):

zomars:
I'm copying my comments from Threads:

> I'm against removing but also I'm in favor for providing enough context in these TODOs so the generated issues can stand on their own.

PR #8905, opened 2023-05-15
CAL-1673: Use dark/light colors from user profile settings in booker

JeroenReumkens:
## What does this PR do?

New booker didn't use the theme setting in appearance to make the booker dark / light. With this PR that is now the case. 

Fixes CAL-1673

## Type of change

## How should this be tested?

- [ ] Ensure that when changing the appearance in the app settings, the booker is either force light/dark or automatic based on your system preferences.


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -23,20 +22,13 @@ import { fadeInLeft, getBookerSizeClassNames, useBookerResizeAnimation } from ".
...
 import type { BookerLayout, BookerProps } from "./types";
 import { useEvent } from "./utils/event";
+import { useBrandColors } from "./utils/use-brand-colors";
 
 const PoweredBy = dynamic(() => import("@calcom/ee/components/PoweredBy"));
 const DatePicker = dynamic(() => import("./components/DatePicker").then((mod) => mod.DatePicker), {
   ssr: false,
 });
 
-const useBrandColors = ({ brandColor, darkBrandColor }: { brandColor?: string; darkBrandColor?: string }) => {
```
JeroenReumkens, line 32: Moved it to a separate file so booker component stays more readable <3

leog (approved):
LGTM, great stuff!

packages/features/bookings/Booker/utils/use-brand-colors.tsx
```diff
@@ -0,0 +1,26 @@
...
+    lightVal: brandColor,
+    darkVal: darkBrandColor,
+  });
+  useCalcomTheme(brandTheme);
+  const { setTheme } = useTheme(theme);
+
+  useEffect(() => {
+    if (theme) setTheme(theme);
+  }, [setTheme, theme]);
+};
```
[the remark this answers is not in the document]
JeroenReumkens: Yeah great question. I took the logic from the current booker, and actually not really sure if it's used anywhere else (and the current booker will be gone soon). 

@sean-brydon  what do you think about this? It pretty much combines most of your hooks into one hook – so perhaps it indeed makes sense to reuse it. On top of that I also added an effect that triggers the `setTheme` again, because in our case the `theme` is undefined during load, which would cause to set the default and not update it anymore later.

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -113,6 +115,7 @@ export const getPublicEvent = async (username: string, eventSlug: string, prisma
         image: `${WEBAPP_URL}/${users[0].username}/avatar.png`,
         brandColor: users[0].brandColor,
         darkBrandColor: users[0].darkBrandColor,
+        theme: users[0].theme,
```
hariombalhara, line 118: It doesn't seem to handle the dynamic group booking separately. I am not sure what's the right approach, but currently dynamic group bookings always use system theme(by setting theme to `null`)

hariombalhara: Also, I don't see handling for `team.theme`. That got added not too long ago.

hariombalhara (changes requested):
Looks good. Works good 🙏

Blocking merge just to get clarity on the dynamic booking link theme first.

packages/features/eventtypes/lib/getPublicEvent.ts
JeroenReumkens: Good point! I'll change dynamic theme to system theme indeed. Makes a lot of sense! Thanks for spotting this!

JeroenReumkens:
@hariombalhara  Thanks a lot for checking this! Just now updated the dynamic booking to indeed use the system theme. Makes a lot more sense. Ready to approve? 😁

hariombalhara (approved):
All good !!

PR #8926, opened 2023-05-16
Fixes locked workflow toggle in event type settings

CarinaWolli:
## What does this PR do?

Fixes that the toggle to disable a workflow on a user event type is locked

This was the bug: 
Screenshot 2023-05-16 at 11 50 09

## Type of change

## How should this be tested?

- Create a user workflow
- Set it active on a normal user event type 
- Got to the settings of that event type and click the workflows tab 
- See that the toggle of the workflow is not locked and workflow can be disabled

### DIFF
--- a/packages/features/ee/workflows/components/EventWorkflowsTab.tsx
+++ b/packages/features/ee/workflows/components/EventWorkflowsTab.tsx
@@ -197,7 +197,7 @@ function EventWorkflowsTab(props: Props) {
         const dataWf = data.workflows.find((wf) => wf.id === workflowOnEventType.id);
         return {
           ...workflowOnEventType,
-          readOnly: isChildrenManagedEventType && dataWf?.teamId ? true : dataWf?.readOnly ?? true,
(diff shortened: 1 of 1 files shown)


packages/features/ee/workflows/components/EventWorkflowsTab.tsx
```diff
@@ -197,7 +197,7 @@ function EventWorkflowsTab(props: Props) {
         const dataWf = data.workflows.find((wf) => wf.id === workflowOnEventType.id);
         return {
           ...workflowOnEventType,
-          readOnly: isChildrenManagedEventType && dataWf?.teamId ? true : dataWf?.readOnly ?? true,
+          readOnly: isChildrenManagedEventType && dataWf?.teamId ? true : dataWf?.readOnly ?? false,
```
CarinaWolli, line 200: if readOnly is not defined, it shouldn't be read-only

hariombalhara (approved):

PR #8937, opened 2023-05-17
feat: GTM, Accept trackingId with and without GTM-

### LINKED ISSUE #8261: [CAL-1492] GTM app not connecting
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1492](https://linear.app/calcom/issue/CAL-1492/gtm-app-not-connecting)</sub>

hariombalhara:
## What does this PR do?

Used zod to transform without GTM- values to start with GTM-
Fixes #8261 
Fixes CAL-1492

[Loom showing that existing values in DB with GTM- or without GTM- would both work.](https://www.loom.com/share/8ba39f31b7d249b7a81338cf19d85368)

## Type of change

## How should this be tested?

- [x] See that an existing value of tracking ID without GTM- would keep on working
- [x] See that an existing value of tracking ID with GTM- would also start working now
- [x] Any new value of trackingId added without GTM- would automatically add "GTM-" to it in DB.

## Checklist


joeauyeung (commented):
Does this mean the onus is on the user to add the GTM- prefix to the tag? If so should we add a hint to the input field to include GTM-. If not, can we add the GTM- suffix to the input field for the tag?

I am assuming this video is still up to date on how users would grab the tag.
https://www.youtube.com/watch?v=P4suvDuj0kI&t=174s

hariombalhara:
Ah no, the title was wrong. Fixed that. Both will work now, with and without GTM-

zomars (approved):

Issue #8893, opened 2023-05-15
[CAL-1677] Wrong language for months in confirmation email of a recurring booking.


hariombalhara:
### Issue Summary

Scheduled a recurring event and received a confirmation email with months in German which I think is of the organizer in this case.

Expected: I should have received it in English

![](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/dcf48b55-0c18-4936-82f5-445663efeb12/fbcac480-7a89-419c-a394-bb6000c627da)

Another related bug:
When I added the event through invite.ics file to my calendar. It shows what, when and timezone in German.

![](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/0bb6c068-d567-453c-94eb-e3f436e7fe09/072b4db5-ccc9-4b47-987d-6fa361a93adf)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1677](https://linear.app/calcom/issue/CAL-1677/wrong-language-for-months-in-confirmation-email-of-a-recurring-booking)</sub>

vedant-jain03:
@hariombalhara I would like to work on it!

hariombalhara:
Sure @vedant-jain03. I am assigning it to you.

I think @Udit-takkar might be able to help if you need any. There was similar PR he was working on, not sure if it's the same issue.

vedant-jain03:
<img width="563" alt="Screenshot 2023-05-16 at 9 13 29 AM" src="https://github.com/calcom/cal.com/assets/76901313/4c5c6a51-9ed5-46d0-986e-de9dab44836e">

@hariombalhara I try to reproduce the issue, but it was showing correct language. I made the event recurring. Let me know any particular step to reproduce the bug

G3root:
fixed by #8941

PR #8945, opened 2023-05-17
fix: wrong locale in attendee email ics

G3root:
## What does this PR do?

fixes the issue as mentioned in #8893 

## Type of change


CarinaWolli (commented):
@G3root Can you fix the type error? Besides that, it looks good 👍

packages/emails/templates/attendee-was-requested-to-reschedule-email.ts
```diff
@@ -91,7 +91,7 @@ ${this.t("request_reschedule_subtitle", {
 })},
 ${this.getWhen()}
 ${this.t("need_to_reschedule_or_cancel")}
-${getManageLink(this.calEvent)}
+${getManageLink(this.calEvent, this.calEvent.organizer.language.translate)}
```
CarinaWolli, line 94: Shouldn't this be the language of the attendee here? 

Also, I tried to test it, and the whole request to reschedule email for the attendee was in the organizer's language. But this doesn't need to be fixed in this PR

CarinaWolli: Let's use `this.t` as in the lines above. And then we still need a follow-up as this email is sent to the attendee but I think the organizer language is used

packages/emails/templates/attendee-scheduled-email.ts
```diff
@@ -112,7 +112,7 @@ ${this.t(
 )}
 ${this.t(subtitle)}
 
-${getRichDescription(this.calEvent)}
+${getRichDescription(this.calEvent, this.attendee.language.translate)}
```
zomars, line 115:
```suggestion
${getRichDescription(this.calEvent, this.t)}
```

zomars (commented):
Left some comments

G3root:
@zomars @CarinaWolli pushed some changes to address the requested changes

zomars (approved):

zomars (approved):

packages/lib/CalEventParser.ts
```diff
@@ -176,38 +177,40 @@ export const getRescheduleLink = (calEvent: CalendarEvent): string => {
   return `${WEBAPP_URL}/reschedule/${seatUid ? seatUid : Uid}`;
 };
 
-export const getRichDescription = (calEvent: CalendarEvent /*, attendee?: Person*/) => {
+export const getRichDescription = (calEvent: CalendarEvent, t_?: TFunction /*, attendee?: Person*/) => {
+  const t = t_ ?? calEvent.organizer.language.translate;
```
hariombalhara, line 181: I would recommend making t_ required here so that from everywhere we are passing the language explicitly and it would not use wrong language.

hariombalhara (approved):
Looks good. I have just one suggestion

PR #8960, opened 2023-05-17
fix focus ring on number of seats [CAL-1699]

### LINKED ISSUE #8928: Focus ring for nr of seats input is too big
### Issue Summary

The focus ring needs to be fixed:
Screenshot 2023-05-16 at 14 38 46

Can be found in: `/event-types/[id]?tabName=advanced`

alexpgates:
## What does this PR do?

This PR addresses #8928 not by shortening the width of the focus ring but by adjusting the width of the field to be consistent with other with similar settings with addOn suffixes. 

Screenshot from original issue:
238645877-614dfe8d-ecc0-49e6-9ca0-018397a65633

Previously, this text field had a `w-24`class that shortened up the width of the field. However, when compared to other settings (for example, the Offset Start Timers setting on the "Limits" page) it seemed inconsistent to use the `w-24`class. 

So instead, for consistency, I just removed the `w-24`class which corrects the issue with the focus ring and displays the field like this:

Screenshot 2023-05-17 at 17 22 31

I did consider an approach to shorten the focus ring similar to how it's done here:

Screenshot 2023-05-17 at 17 22 20

However I was unable to get a good width class that aligned. 
See:

Screenshot 2023-05-17 at 17 00 20

Too wide.

But I do maintain that the focus ring on the shortened fields look kind of bad anyway. I'd go with the full width field! 

Fixes  #8928

Loom Video: https://www.loom.com/share/347251d8eaea4b108984e36e1883a772

Note that the loom video demos the corrected field first (the number of seats) but then shows the "Offset Start Timers" field on the "Advanced" page to show how the updated field is consistent with other similar fields

**Environment**: main production (?)

## Type of change

## How should this be tested?

I only tested by viewing the page and observing the field with and focus ring behavior

## Checklist

### DIFF
--- a/apps/web/public/static/locales/ar/common.json
+++ b/apps/web/public/static/locales/ar/common.json
@@ -1572,7 +1572,7 @@
   "sender_name": "اسم المرسل",
   "already_invited": "تمت دعوة هذا الحاضر بالفعل",
   "no_recordings_found": "لم يتم العثور على تسجيلات",
-  "new_workflow_subtitle": "سير عمل جديد من أجل...",
+  "new_workflow_subtitle": "سير عمل جديد لـ...",
   "reporting": "تقديم التقارير",
   "reporting_feature": "عرض جميع البيانات الواردة وتنزيلها كملف CSV",
   "teams_plan_required": "خطة Teams مطلوبة",


yarn.lock
joeauyeung, line 1: We shouldn't change the `yarn.lock` file in this PR. Can you checkout this file from `main` to remove the changes.

joeauyeung (changes requested):
Thanks for your contribution and the detailed explanation. We only need to make one change to the `yarn.lock` file and then we're good to go.

PR #8968, opened 2023-05-18
Implement `UpgradeTip` in App installation flow for apps that need teams plan

### LINKED ISSUE #7973: [CAL-1707] Routing Forms misinformation
The [Routing forms App](https://app.cal.com/apps/routing-forms) is stated as free but is not available after installation without a team account.

<sub>[CAL-1707](https://linear.app/calcom/issue/CAL-1707/routing-forms-misinformation)</sub>

hariombalhara:
## What does this PR do?

Fixes #7973

## Demo
https://www.loom.com/share/eae39ce07e824e1385be4e39ee82b5d7

## Type of change

## How should this be tested?
- [x] Try installing Routing Forms App when for an account that doesn't have teams plan and you would notice that you are shown Upgrade Tip.
- [x] Try the same for an account that has teams plan and it won't show Upgrade Tip.

## Checklist

### DIFF
--- a/apps/web/components/apps/App.tsx
+++ b/apps/web/components/apps/App.tsx
@@ -36,6 +36,7 @@ const Component = ({
   tos,
   privacy,
   isProOnly,
+  teamsPlanRequired,
   descriptionItems,
   isTemplate,
   dependencies,
@@ -152,6 +153,7 @@ const Component = ({
                   type={type}
                   isProOnly={isProOnly}
                   disableInstall={disableInstall}
+                  teamsPlanRequired={teamsPlanRequired}
                   render={({ useDefaultComponent, ...props }) => {
                     if (useDefaultComponent) {
                       props = {
@@ -192,6 +194,7 @@ const Component = ({
               type={type}
               isProOnly={isProOnly}
               disableInstall={disableInstall}
+              teamsPlanRequired={teamsPlanRequired}
               render={({ useDefaultComponent, ...props }) => {
                 if (useDefaultComponent) {
                   props = {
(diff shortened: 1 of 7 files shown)


joeauyeung (commented):
I watched the Loom and being redirected to the teams page seems jarring without any explanation seems jarring. Could we reuse the same UI when apps requires a prerequisite app to be installed?

Ex. Google Meet requires GCal to be installed

From the apps page, the install button is disabled which should lead the user to the details page
image

Then on the details page, we could have a banner explaining what is needed and a link to create the team.
image

hariombalhara:
Yeah I agree that it's not a good UX. But I think (as I mentioned in TODO) a dialog is more appropriate here. We have no implementation of Dialog for Upgrade right now. 
With the dialog in place, I think we won't need to add more messaging as you suggest. 

@PeerRich What do you think should be thr approach here?

sean-brydon (approved):
Nice job! I agree with the dialog approach but that can be handled in a follow up

sean-brydon:
> Yeah I agree that it's not a good UX. But I think (as I mentioned in TODO) a dialog is more appropriate here. We have no implementation of Dialog for Upgrade right now.
> 
> With the dialog in place and Pricing in there already, I think we won't need to add more messaging as you suggest. Also, maybe we can highlight the pricing a bit better.
> 
> Screenshot 2023-05-19 at 11 45 05 AM
> @PeerRich What do you think should be the approach here? I am inclined towards adding the flow quickly as it's misleading pricing right now. A followup with the correct approach can be done later.

@Jaibles if you can design a dialog update tip i'll make it for @hariombalhara so he can use it in a follow up :)

Jaibles:
Proposed designs here: 
https://www.figma.com/file/ZJNAFHCAq6rnuGRgUFxjzQ/4.-Orgs-%2F-Teams.---New-Features?type=design&node-id=4786%3A212846&t=174oVDVZxCkNZJpU-1
Comment with any feedback!

hariombalhara:
Merging this one. @sean-brydon or I would pick up the design as a followup.

PR #8978, opened 2023-05-19
fix: user full name not wrapping on navbar

### LINKED ISSUE #8977: User full name overflowing nav bar width on settings page
### Issue Summary

Because nav bar has a fixed width, when the user has a long name, it overflows the nav bar width.

image

### Steps to Reproduce

1. Go to settings page
2. If you have a long full name like me you'll see the name overflowing the nav bar width

The text is expected to wrap and show the user's full name

### Technical details

- Arc Version 0.103.0 (38794) | Chromium Engine Version 113.0.5672.92

ouwargui:
## What does this PR do?
Wraps user full name if it exceeds the full width of the nav bar. This also makes sure that if someday a new section with a long name is added to the navigation bar, the text will wrap automatically.

Fixes #8977

Loom Video: https://www.loom.com/share/0c3a41ba624443cdb84bff7eceb01edc

**Environment**: Production

## Type of change

## How should this be tested?

- [ ] Go to settings page
- [ ] Make sure your user has a very long full name, like mine.
- [ ] See the text wrapping instead of overflowing


CarinaWolli (commented):
Longer usernames that need 3 lines are cut off. There is a max. of 50 characters for usernames, can you make sure that also these longer ones are shown properly? 

Screenshot 2023-05-19 at 08 40 43

ouwargui:
@CarinaWolli I set the height to auto, it does fix the problem, but it can grow as much as it needs. However, since the full name has a 50 characters max, it shouldn't be a problem.
image

PeerRich:
should we truncate with ellipsis instead? 

Peer Richels...

Jaibles:
> should we truncate with ellipsis instead?
> 
> Peer Richels...

This is a good idea. The purpose of this is to clarify which account you're in so you don't need to see every character.

ouwargui:
@PeerRich @Jaibles Would we like to truncate the rest of the section titles as well? This is not a problem today, but if a new section is added with a long name, I think wrapping the text would be a better UX.

alannnc (dismissed):
As comments say we can instead truncate if more than X characters.

ouwargui:
Sorry for the delay, was busy with some work stuff. I truncated the user full name but left the rest of the sections wrapping the title. Is that ok or do we want to truncate for all sections?
imageimage

Jaibles:
> Sorry for the delay, was busy with some work stuff. I truncated the user full name but left the rest of the sections wrapping the title. Is that ok or do we want to truncate for all sections?
> 
> imageimage
> 
> 

Makes sense to truncate all

Jaibles:
> What do we think? imageimage

Looks good to me!

CarinaWolli (approved):
LGTM! Thank you 🙏

PeerRich (approved):

PR #8981, opened 2023-05-19
fix: Text beyond the X icon fix

### LINKED ISSUE #8976: Text beyond the X icon

### Issue Summary

The text on this input is overwriting the X icon space

### Steps to Reproduce

1. Edit some event-type
2. Go to "Advance"
3. Add a question and select "MultiSelect" type
4.Add more than two options

Loom video: https://www.loom.com/share/259ab60b675f497abb986d712e745fd9

VinoV1999:
## What does this PR do?

prevents text to not go behind the cross sysmbol if cross symbol visible.

Fixes #8976

## Type of change

- passed classname attribute

## ScreehShot:-

Screenshot 2023-05-19 at 8 49 19 AM


packages/features/form-builder/FormBuilder.tsx
```diff
@@ -195,6 +195,7 @@ export const FormBuilder = function FormBuilder({
                 <div className="flex items-center">
                   <Input
                     required
+                    className={value.length > 2?"pr-6":""}
```
hariombalhara, line 198: Why is there a value.length check in here?

hariombalhara (changes requested):
Left a comment.

packages/features/form-builder/FormBuilder.tsx
alannnc:
```suggestion
                    className={"pr-6"}
```
2 as char length is probably nothing I think we are better with "pr-6" always.

VinoV1999: Cross sysmbol appears only when value.length is greater than 2. if we directly add padding right then it seems some unwanted extra padding added on right. That looks not good i feel

with two options

so to avoid that padding right shld be added only if cross symbol occurs  

## FYI 

 **This is the output if length is lesser than or equal to 2**

without padding

 **This is the output if length is greater than 2**
 
with padding

hariombalhara: I don't like the idea of input length deciding classes. @sean-brydon Need your expertise here.

sean-brydon: Yeah im honestly not too much of a fan of this either. I wonder if we should just make this whole box a div and the input and X lives in side of it. 

Might be an easier approach

sean-brydon (changes requested):
Lets rethink the approach for handling the X button here - We could look at using TextField component that already supports addonleading 

cal/packages/ui/components/form/inputs/Input.tsx

VasuDevrani:
> Lets rethink the approach for handling the X button here - We could look at using TextField component that already supports addonleading
> 
> cal/packages/ui/components/form/inputs/Input.tsx

How about these two kinds of changes, I'll create the pr if they are fine

1.
image

</br>
2.  </br>
image

sean-brydon:
Thanks for the contribution @VinoV1999 

closing in favour of https://github.com/calcom/cal.com/pull/9198?no-redirect=1 as it was a cleaner approach!

PR #8983, opened 2023-05-19
Fixes SMS workflow actions for premium users

CarinaWolli:
## What does this PR do?

Fixes that workflows with SMS actions could not be saved when user was not on Teams plan but had premium username. 

## Type of change

## How should this be tested?

- User should not be part of a team and should have premium user name (`metadata: {"isPremium":true}`)
- Create a workflow 
- Add a SMS action 
- Save workflow
- See that no error is thrown
- Go back to the workflow and see that the SMS action is saved


emrysal (approved):
This can be a separate function, in case this is needed anywhere else (likely) - approved as is for now 👍 .

PR #8986, opened 2023-05-19
Add config support in `floatingButton` instruction

### LINKED ISSUE #7716: [CAL-1244] Missing prefill support with Floating Button popup Embed
Config is what actually allows certain form values to be prefilled and it's missing in floating button popup

Also, `config` should have `prefill` property to clearly signify where prefill parameters can be specified.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1244](https://linear.app/calcom/issue/CAL-1244/missing-prefill-support-with-floating-button-popup)</sub>

hariombalhara:
## What does this PR do?

Allows providing config object to `floatingButton` too same as `inline` instruction.

Demo:
https://www.loom.com/share/a69425c8df364d4689c2614ef0439d0b

Fixes #7716

## Type of change

## How should this be tested?

- [x] Go to [playground](http://localhost:3100/embed/?only=ns:floatingButton#cal-booking-place-floatingButton-iframe) and click Floating Button at the bottom. You would now see prefilled booking form which happens because the config is provided for it.

## Checklist


packages/embeds/embed-core/src/embed.ts
```diff
@@ -466,6 +468,10 @@ class CalApi {
       el.dataset.calLink = calLink;
       el.dataset.calNamespace = this.cal.namespace;
       el.dataset.calOrigin = calOrigin ?? "";
+      if (config) {
+        el.dataset.calConfig = JSON.stringify(config);
```
hariombalhara, line 472: Because floating button popup uses Element Click popup logic only, adding this data attribute is enough.

alannnc (approved):
Code looks good, just unsure why so many files changed in size in bundle analysis.

PR #8992, opened 2023-05-19
Fixes cron job to schedule all workflow emails

CarinaWolli:
## What does this PR do?

Fixes that the Cron job `scheduleEmailReminders` didn't schedule workflow emails correctly. 

## Type of change

### DIFF
--- a/packages/features/ee/workflows/api/scheduleEmailReminders.ts
+++ b/packages/features/ee/workflows/api/scheduleEmailReminders.ts
@@ -102,10 +102,10 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     where: {
       method: WorkflowMethods.EMAIL,
       scheduled: false,
-      cancelled: false,
       scheduledDate: {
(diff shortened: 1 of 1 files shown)


packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -102,10 +102,10 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     where: {
       method: WorkflowMethods.EMAIL,
       scheduled: false,
-      cancelled: false,
       scheduledDate: {
         lte: dayjs().add(72, "hour").toISOString(),
       },
+      OR: [{ cancelled: false }, { cancelled: null }],
```
CarinaWolli, line 108: `cancelled` is an optional field, before we we only checked for `cancelled: false` which didn't include `null` values. That's why some workflowReminders weren't scheduled

emrysal (approved):

PR #8993, opened 2023-05-19
feat: Organizations 

### LINKED ISSUE #7142: [CAL-1088] RFC: Create Organisations layer above teams
We want to create organisations within [Cal.com](http://Cal.com) to enable people to easily and effectively manage multiple teams. An organisation will live above the current teams layer.

**An organisation will have**

* **People:** These people will be added as members to any number of teams within the organisation.
* **Teams:** will operate very much in the same way as they do now, except that certain things will be able to be shared across teams.
* it's own org-name that will compete with usernames: [cal.com/acme](http://cal.com/acme) = company, \[[cal.com/acme/anna](http://cal.com/acme/anna)\]
* when joining an org, you get a new org-username ([cal.com/peer](http://cal.com/peer) and [cal.com/cal/peer](http://cal.com/cal/peer) could be different people)
* subteams will compete with org-usernames for namespace:
  * [cal.com/acme/marketing](http://cal.com/acme/marketing) = marketing team
  * ([http://cal.com/acme/anna](http://cal.com/acme/anna)) = CEO of acme
  * [cal.com/acme/anna/30min](http://cal.com/acme/anna/30min) = 30 min with CEO anna
* org-usernames dont require premium and only compete with other names in the org

**Roles within an org**

* Owner (can edit all org & team settings)
* Admin (can edit all org & team settings)
* Analyst (see everything, read only)
* Team Admin (edit all settings team level)
* Member (view only within their team, can’t edit)

**Open questions**

* What else can be shared across teams via the organisation? ie workflows? apps? routing forms? etc.
* Does an org need it's own event types? Assumption for now is no.
* Can a member see all teams within an org?
* Can a member see all people within an org or just the people in the shared teams.

**(out of date) WIP chart - subdomain is now subpath**

**Current state of designs**

[Organisations - Cal.com Teams (Figma)](https://www.figma.com/file/ZJNAFHCAq6rnuGRgUFxjzQ/Cal.com-Teams?node-id=217%3A53765&t=nxZiKwVYEAFvYa2Q-1)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1088](https://linear.app/calcom/issue/CAL-1088/rfc-create-organisations-layer-above-teams)</sub>

leog:
## What does this PR do?

This PR concentrates all the implementations towards having organizations in the app.

Every PR merged to this PR will have its own review so unless requested, no screenshot will be added to this PR.

Implements #7142 

## Type of change

## Checklist


.env.example
```diff
@@ -32,11 +33,13 @@ PRISMA_GENERATE_DATAPROXY=
...
 
 # - SHARED **************************************************************************************************
+# Set this to http://app.cal.local:3000 if you want to enable organizations, and
+# check variable ORGANIZATIONS_ENABLED at the bottom of this file
 NEXT_PUBLIC_WEBAPP_URL='http://localhost:3000'
 # Change to 'http://localhost:3001' if running the website simultaneously
 NEXT_PUBLIC_WEBSITE_URL='http://localhost:3000'
 NEXT_PUBLIC_CONSOLE_URL='http://localhost:3004'
-NEXT_PUBLIC_EMBED_LIB_URL='http://localhost:3000/embed/embed.js'
+NEXT_PUBLIC_EMBED_LIB_URL='http://app.cal.local:3000/embed/embed.js'
```
hariombalhara, line 42: This should correspond to `NEXT_PUBLIC_WEBAPP_URL` only

.env.example
```diff
@@ -58,7 +61,7 @@ PGSSLMODE=
 # @see: https://github.com/calendso/calendso/issues/263
 # @see: https://next-auth.js.org/configuration/options#nextauth_url
 # Required for Vercel hosting - set NEXTAUTH_URL to equal your NEXT_PUBLIC_WEBAPP_URL
-NEXTAUTH_URL='http://localhost:3000'
+NEXTAUTH_URL='http://app.cal.local:3000'
```
hariombalhara, line 64: Same here I guess?

.env.example
```diff
@@ -183,3 +186,17 @@ CSP_POLICY=
...
 NEXT_PUBLIC_MINUTES_TO_BOOK=5 # Minutes
+
+# - ORGANIZATIONS *******************************************************************************************
+# Enable Organizations non-prod domain setup, works in combination with organizations feature flag
+# This is mainly needed locally, because for orgs to work a full domain name needs to point
+# to the app, i.e. app.cal.local instead of using localhost, which is very disruptive
+#
+# This variable should only be set to 1 or true if you are in a non-prod environment and you want to
+# use organizations
+ORGANIZATIONS_ENABLED=
```
hariombalhara, line 197: Maybe we can enable it automatically on dev by checking if NEXT_PUBLIC_WEBAPP_URL is a organisation supported domain or not(which is basically not localhost)

apps/web/middleware.ts
```diff
@@ -79,6 +108,8 @@ const middleware: NextMiddleware = async (req) => {
 
 export const config = {
   matcher: [
+    "/",
+    "/:path*",
```
hariombalhara, line 112: That would significantly increase the invocations of serverless functions(as it's matching all the requests coming in). Can we achieve the same effect using next.config.js rewrites ?

apps/web/pages/api/logo.ts
```diff
@@ -118,7 +118,7 @@ async function getTeamLogos(subdomain: string) {
     }
     // load from DB
     const { default: prisma } = await import("@calcom/prisma");
-    const team = await prisma.team.findUnique({
+    const team = await prisma.team.findFirst({
       where: {
         slug: subdomain,
       },
```
sean-brydon, line 124: Cant we get an overlap of slug being passed in here? 'Marketing' can exists in two or more places

apps/web/pages/api/user/avatar.ts
```diff
@@ -18,7 +18,7 @@ async function getIdentityData(req: NextApiRequest) {
   const { username, teamname } = querySchema.parse(req.query);
 
   if (username) {
-    const user = await prisma.user.findUnique({
+    const user = await prisma.user.findFirst({
       where: { username },
```
sean-brydon, line 22: Same as above - if i create a user called "peer" inside an org i get peers gravitar logo/the first peer in the DB

apps/web/pages/api/user/avatar.ts
```diff
@@ -29,7 +29,7 @@ async function getIdentityData(req: NextApiRequest) {
     };
   }
   if (teamname) {
-    const team = await prisma.team.findUnique({
+    const team = await prisma.team.findFirst({
       where: { slug: teamname },
       select: { logo: true },
```
sean-brydon, line 34: Same as above - we should probably tie this i with orgId

apps/web/pages/new-booker/d/[link]/[slug].tsx
```diff
@@ -68,7 +68,7 @@ async function getUserPageProps(context: GetServerSidePropsContext) {
     };
   }
 
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
     where: {
       username,
     },
```
Comment (sean-brydon) [inline apps/web/pages/new-booker/d/[link]/[slug].tsx:74]: This will need org context too

apps/web/pages/new-booker/[user]/[type].tsx
```diff
@@ -94,7 +94,7 @@ async function getUserPageProps(context: GetServerSidePropsContext) {
 
   const { ssrInit } = await import("@server/lib/ssr");
   const ssr = await ssrInit(context);
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
     where: {
       username,
     },
```
Comment (sean-brydon) [inline apps/web/pages/new-booker/[user]/[type].tsx:100]: Will also need org context

packages/core/getUserAvailability.ts
```diff
@@ -126,7 +126,7 @@ export async function getUserAvailability(
     throw new HttpError({ statusCode: 400, message: "Invalid time range given." });
   }
 
-  const where: Prisma.UserWhereUniqueInput = {};
+  const where: Prisma.UserWhereInput = {};
   if (username) where.username = username;
```
sean-brydon, line 130: if username we will need to take into account org also

packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+  DialogHeader,
+  Label,
+  Input,
+} from "@calcom/ui";
+import { ArrowRight, Info } from "@calcom/ui/components/icon";
+
+function extractDomainFromEmail(email: string) {
+  let out = "";
+  try {
+    const match = email.match(/^(?:.*?:\/\/)?.*?(?<root>[\w\-]*(?:\.\w{2,}|\.\w{2,}\.\w{2}))(?:[\/?#:]|$)/);
```
sean-brydon, line 31: NIT: since this is an email isnt this always just the same as `email.split("@")[1]` 

Or am i not understanding this regex correctly?

sean-brydon: user@example.com
regex output: example.com
split output: example.com

https://www.example.co.uk/path/to/page
regex output: example.co.uk 
split output: example.co.uk/path/to/page (however emails are never in this format?)

packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+                inputMode="decimal"
+                {...digits[0]}
+                autoFocus
+                autoComplete="one-time-code"
+              />
+              <Input className={digitClassName} name="2fa2" inputMode="decimal" {...digits[1]} />
+              <Input className={digitClassName} name="2fa3" inputMode="decimal" {...digits[2]} />
+              <Input className={digitClassName} name="2fa4" inputMode="decimal" {...digits[3]} />
+              <Input className={digitClassName} name="2fa5" inputMode="decimal" {...digits[4]} />
+              <Input className={digitClassName} name="2fa6" inputMode="decimal" {...digits[5]} />
```
sean-brydon, line 102: NIT: Lets make this a component as its used in a few places 

Doesnt need to be picked up here - created a issue

packages/lib/constants.ts
```diff
@@ -8,7 +8,7 @@ export const WEBAPP_URL =
   RAILWAY_STATIC_URL ||
   HEROKU_URL ||
   RENDER_URL ||
-  "http://localhost:3000";
+  "http://app.cal.local:3000";
```
sean-brydon, line 11: should this be changed?

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -0,0 +1,132 @@
...
+    user: NonNullable<TrpcSessionUser>;
+  };
+  input: TCreateInputSchema;
+};
+
+const vercelCreateDomain = async (domain: string) => {
+  const response = await fetch(
+    `https://api.vercel.com/v8/projects/${process.env.PROJECT_ID_VERCEL}/domains?teamId=${process.env.TEAM_ID_VERCEL}`,
+    {
+      body: `{\n  "name": "${domain}.${subdomainSuffix()}"\n}`,
```
sean-brydon, line 28: Can we not just `JSON.stringify({name:${domain}.${subdomainSuffix()}})`

.env.example
leog: Reverting to localhost, forgot to do so when introducing the new orgs env var. Would it be possible to use `NEXT_PUBLIC_WEBAPP_URL` from where `NEXT_PUBLIC_EMBED_LIB_URL` is used and have it only point to `/embed/embed.js` which is the location of the lib? so it's not repetitive and completely dependant on `NEXT_PUBLIC_WEBAPP_URL` from the core.

.env.example
leog: Reverting back to localhost as well, forgot to do that when introduced the new orgs env var. Is there a reason we have this splitted in so many env vars? could we merge them into one?

apps/web/middleware.ts
leog: Managed to remove matcher `"/"` as the one with `"/:path*"` covers it, but unfortunately, we can't do anything for that one left, because we need to be able to distinguish a user from a team as `/username` and `/teamslug` are both accessible equally. I would have preferred to leave it as we had `/username` and `/team/teamslug` but the decision was to be able to support that at all costs if I remember correctly.

apps/web/pages/api/user/avatar.ts
leog: Making this API subdomain-aware, so asking `acme.cal.com/api/user/avatar?teamname=engineering` will return the avatar for a team name engineering in acme org.

apps/web/pages/new-booker/d/[link]/[slug].tsx
Comment (leog) [reply apps/web/pages/new-booker/d/[link]/[slug].tsx:74]: Added, thanks.

packages/core/getUserAvailability.ts
leog: Done, thanks.

packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
leog: This was meaning to extract the correct domain from subdomain names as well, such as `leo@engineering.cal.com` to be `cal.com`

packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
leog: Good catch!

packages/lib/constants.ts
leog: Reverted :+1:

.env.example
leog: Yeah, functionality wise the feature flag can be turned off. Not entirely sure what to do here.

.vscode/launch.json
```diff
@@ -5,11 +5,12 @@
       "type": "node",
       "request": "launch",
       "name": "Next.js Node Debug",
-      "runtimeExecutable": "${workspaceFolder}/node_modules/next/dist/bin/next",
+      "runtimeExecutable": "${userHome}/.yarn/bin/yarn",
```
zomars, line 8: Why do we need this? What it is solving?

apps/web/components/getting-started/steps-views/UserSettings.tsx
```diff
@@ -69,7 +70,7 @@ const UserSettings = (props: IUserSettingsProps) => {
     <form onSubmit={onSubmit}>
       <div className="space-y-6">
         {/* Username textfield */}
-        <UsernameAvailabilityField user={user} />
+        <UsernameAvailabilityField user={user} organization={organization} />
```
zomars, line 73: My two cents here. Let's use trpc inside to avoid prop drilling. We can take advantage of suspense here as well.

```ts
const [user] = trpc.viewer.me.useSuspenseQuery();
```

apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
```diff
@@ -116,9 +131,7 @@ const UsernameTextfield = (props: ICustomUsernameProps) => {
             ref={usernameRef}
             name="username"
             value={inputUsernameValue}
-            addOnLeading={
-              <>{process.env.NEXT_PUBLIC_WEBSITE_URL.replace("https://", "").replace("http://", "")}/</>
-            }
+            addOnLeading={<>{usernamePrefix}/</>}
```
zomars, line 134: Can't we just pass `addOnLeading` in the parent? So `UsernameTextfield` doesn't care about org-specific logic.

apps/web/components/ui/UsernameAvailability/index.tsx
```diff
@@ -18,11 +20,13 @@ interface UsernameAvailabilityFieldProps {
   onSuccessMutation?: () => void;
   onErrorMutation?: (error: TRPCClientErrorLike<AppRouter>) => void;
   user: Pick<User, "username" | "metadata">;
+  organization: { slug?: string | null | undefined; metadata: z.infer<typeof teamMetadataSchema> } | null;
```
zomars, line 23: Let's suspense-query on child instead prop-drilling.

zomars: [Avoid prop-drilling](https://handbook.cal.com/engineering/best-practices/avoid-prop-drilling)

apps/web/middleware.ts
zomars: I know this is necessary for orgs to function but expanding on @hariombalhara comment. This WILL increase our server less function invocations by a lot. This will be a big hit to self-hosted users billing in Vercel. We're fortunate enough to be sponsored but can't say the same for Self-hosted users. cc @keithwillcode @PeerRich and @baileypumfleet for visibility.

apps/web/components/ui/UsernameAvailability/index.tsx
```diff
@@ -51,7 +65,7 @@ export const UsernameAvailabilityField = ({
             setInputUsernameValue={onChange}
             onSuccessMutation={onSuccessMutation}
             onErrorMutation={onErrorMutation}
-            user={user}
+            addOnLeading={usernamePrefix}
```
zomars, line 68: Refactored. Noticed we pass the full user just to determine the `addOnLeading`. Let's just pass the `addOnLeading` from here instead.

apps/web/components/ui/UsernameAvailability/index.tsx
```diff
@@ -17,14 +19,24 @@ export const UsernameAvailability = IS_SELF_HOSTED ? UsernameTextfield : Premium
...
+
+function useUserNamePrefix(organization: RouterOutputs["viewer"]["me"]["organization"]): string {
+  return organization
+    ? organization.slug
+      ? `${organization.slug}.${subdomainSuffix()}`
+      : organization.metadata && organization.metadata.requestedSlug
+      ? `${organization.metadata.requestedSlug}.${subdomainSuffix()}`
+      : process.env.NEXT_PUBLIC_WEBSITE_URL.replace("https://", "").replace("http://", "")
+    : process.env.NEXT_PUBLIC_WEBSITE_URL.replace("https://", "").replace("http://", "");
+}
```
zomars, line 32: Abstracted to keep the main component legible.

apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
```diff
@@ -116,9 +117,6 @@ const UsernameTextfield = (props: ICustomUsernameProps) => {
             ref={usernameRef}
             name="username"
             value={inputUsernameValue}
-            addOnLeading={
-              <>{process.env.NEXT_PUBLIC_WEBSITE_URL.replace("https://", "").replace("http://", "")}/</>
-            }
```
zomars, line 121: Explained I'm my other comment.

apps/web/pages/[user]/calendar-cache/[month].tsx
```diff
@@ -16,7 +16,7 @@ export const getStaticProps: GetStaticProps<
   { user: string }
 > = async (context) => {
   const { user: username, month } = paramsSchema.parse(context.params);
-  const userWithCredentials = await prisma.user.findUnique({
+  const userWithCredentials = await prisma.user.findFirst({
```
Comment (zomars) [inline apps/web/pages/[user]/calendar-cache/[month].tsx:19]: Should we add org is null to the query? @leog

apps/web/pages/api/auth/forgot-password.ts
```diff
@@ -44,7 +44,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
   }
 
   try {
-    const maybeUser = await prisma.user.findUnique({
+    const maybeUser = await prisma.user.findFirst({
```
zomars, line 47: Why can't we use findUnique anymore? Trying to understand.

turbo.json
```diff
@@ -276,6 +276,10 @@
     "ZOHOCRM_CLIENT_SECRET",
     "ZOOM_CLIENT_ID",
     "ZOOM_CLIENT_SECRET",
-    "INTERCOM_SECRET"
+    "INTERCOM_SECRET",
+    "PROJECT_ID_VERCEL",
+    "TEAM_ID_VERCEL",
+    "AUTH_BEARER_TOKEN_VERCEL",
+    "ORGANIZATIONS_ENABLED"
```
zomars, line 283: Nit: sort this alphabetically.

zomars (commented):
Partial review done. Made some type fixes and refactoring. Great job so far @leog 🙏🏽 Will continue reviewing and testing.

apps/web/components/getting-started/steps-views/UserProfile.tsx
```diff
@@ -8,21 +8,15 @@ import { md } from "@calcom/lib/markdownIt";
...
   bio: string;
 };
-interface IUserProfileProps {
-  user: IOnboardingPageProps["user"];
-}
 
-const UserProfile = (props: IUserProfileProps) => {
-  const { user } = props;
+const UserProfile = () => {
+  const [user] = trpc.viewer.me.useSuspenseQuery();
```
zomars, line 19: Refactored to [Avoid prop-drilling](https://handbook.cal.com/engineering/best-practices/avoid-prop-drilling)

.env.example
hariombalhara: Marking it resolved. Let's leave it as is for now.

packages/features/ee/organizations/lib/orgDomains.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+  // Define which is the current domain/subdomain
+  return hostname.replace(`.${currentHostname}` ?? "", "");
+}
+
+export function orgDomainConfig(hostname: string) {
+  const currentOrgDomain = getOrgDomain(hostname);
+  return {
+    currentOrgDomain,
+    isValidOrgDomain: currentOrgDomain !== "app" && !appHostnames.includes(currentOrgDomain),
```
hariombalhara, line 29: To me it looks like `!appHostnames.includes(currentOrgDomain)` would always be true.
What's the case where it can be false?

apps/web/pages/[user]/[type].tsx
```diff
@@ -19,6 +21,12 @@ export type AvailabilityPageProps = inferSSRProps<typeof getStaticProps> & Embed
 
 export default function Type(props: AvailabilityPageProps) {
   const { t } = useLocale();
+  const [isValidOrgDomain, setIsValidOrgDomain] = useState(false);
+
+  useEffect(() => {
+    const { isValidOrgDomain } = orgDomainConfig(window.location.host ?? "");
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:27]: Trying to avoid the useEffect.

Can we use `WEBAPP_URL` from constants to instead derive if the domain is valid org domain or not?

We anyway expect the location.host to be same as in WEBAPP_URL.
Screenshot 2023-06-13 at 10 58 34 AM

This can make `isValidOrgDomain` a derived constant itself

apps/web/pages/[user]/calendar-cache/[month].tsx
Comment (hariombalhara) [reply apps/web/pages/[user]/calendar-cache/[month].tsx:19]: This reminds me that wherever we are querying by username, we need to add that check, otherwise an organisation username might override the settings.

apps/web/pages/api/auth/forgot-password.ts
hariombalhara: I think it's because email is no longer unique in DB. Same goes for username.

apps/api/pages/api/availability/_get.ts
```diff
@@ -119,8 +120,10 @@ const availabilitySchema = z
 async function handler(req: NextApiRequest) {
   const { prisma, isAdmin, userId: reqUserId } = req;
   const { username, userId, eventTypeId, dateTo, dateFrom, teamId } = availabilitySchema.parse(req.query);
+  const { currentOrgDomain, isValidOrgDomain } = orgDomainConfig(req.headers.host ?? "");
```
hariombalhara, line 123: We should be able to simplify by making sure `currentOrgDomain` is set only if `isValidOrgDomain` is true

apps/api/pages/api/availability/_get.ts
```diff
@@ -119,8 +120,10 @@ const availabilitySchema = z
...
   if (!teamId)
     return getUserAvailability({
+      orgSlug: isValidOrgDomain ? currentOrgDomain : undefined,
```
hariombalhara, line 126: We should be able to simplify by making sure currentOrgDomain is set only if isValidOrgDomain is true

apps/web/pages/[user].tsx
```diff
@@ -284,7 +287,7 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
     avatar: `${WEBAPP_URL}/${user.username}/avatar.png`,
   }));
 
-  if (!users.length) {
+  if (!users.length || (!isValidOrgDomain && !users.some((user) => user.organizationId === null))) {
```
Comment (hariombalhara) [inline apps/web/pages/[user].tsx:290]: We are just ensuring here that in non-org context, org users aren't entertained.
But we still accept user of second org in an org.

I had two users `hariom` in 2 different orgs(org1, org2) and when I went to org1.cal.local/hariom it considered it a dynamic booking link, because it got 2 results.

apps/web/pages/event-types/index.tsx
```diff
@@ -362,7 +371,9 @@ export const EventTypeList = ({ group, groupIndex, readOnly, types }: EventTypeL
       <ul ref={parent} className="divide-subtle !static w-full divide-y" data-testid="event-types">
         {types.map((type, index) => {
           const embedLink = `${group.profile.slug}/${type.slug}`;
-          const calLink = `${CAL_URL}/${embedLink}`;
+          const calLink = `${
+            orgBranding ? `${orgBranding.slug}.${subdomainSuffix()}` : CAL_URL
```
hariombalhara, line 375: **[Blocker-Org]**
In case of orgBranding being true, the result link doesn't have a URL scheme(https://) and thus this link doesn't work when we click on preview

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,63 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
-        .map(mapEventType)
-        .filter((evType) => evType.userId === null || evType.userId === ctx.user.id)
-        .filter((evType) =>
-          membership.role === MembershipRole.MEMBER ? evType.schedulingType !== SchedulingType.MANAGED : true
-        ),
-    }))
+    user.teams
+      .filter((mmship) => {
+        const metadata = teamMetadataSchema.parse(mmship.team.metadata);
+        return !metadata?.isOrganization;
```
hariombalhara, line 212: How is it different from checking `mmship.team.children.length` ?

apps/web/pages/api/auth/forgot-password.ts
hariombalhara: It can fetch wrong user because we are getting the first one and not the context specific.
e.g. Acme's John or Deel's John

apps/web/pages/api/auth/reset-password.ts
```diff
@@ -26,7 +26,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       return res.status(400).json({ message: "Couldn't find an account for this email" });
     }
 
-    const maybeUser = await prisma.user.findUnique({
+    const maybeUser = await prisma.user.findFirst({
```
hariombalhara, line 29: Similarly here, it can fetch wrong user.

.vscode/launch.json
leog: Oh I just found it more effective using yarn than the script that dev is calling, but I can revert, not a needed change for the PR

apps/web/components/booking/BookingDescription.tsx
```diff
@@ -46,7 +46,10 @@ const BookingDescription: FC<Props> = (props) => {
   const { profile, eventType, isBookingPage = false, children } = props;
   const { date: bookingDate } = useRouterQuery("date");
   const { t } = useLocale();
-  const { duration, setQuery: setDuration } = useRouterQuery("duration");
+  const { duration, setQuery: setDuration } = useRouterQuery("duration", {
+    // Only set duration query parameter when event type has multiple durations
+    disabled: !eventType.metadata?.multipleDuration,
```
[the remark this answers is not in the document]
leog: Reverting, wanted to remove the unneeded "duration" query param for non-multiple-duration event-types and also make sure we respected the pathname when we rewrite the URL when on an org, but it's not worth breaking anything like this of course.

apps/web/pages/[user].tsx
Comment (leog) [reply apps/web/pages/[user].tsx:290]: Gotcha, good catch @hariombalhara. I added the org slug when available to narrow the user findings.

apps/web/pages/[user]/calendar-cache/[month].tsx
Comment (leog) [reply apps/web/pages/[user]/calendar-cache/[month].tsx:19]: Yeah, we need to narrow down the queries now because of orgs. The thing is that we cannot get hostname from GetStaticProps unless we add org as a route parameter as we have [user] for this. Any suggestion @roae?

apps/web/pages/api/auth/forgot-password.ts
leog: Email is still unique, reverting this change as `findUnique` will still work here: 
image

apps/web/pages/event-types/index.tsx
leog: Good catch, fixed.

packages/features/ee/organizations/lib/orgDomains.ts
leog: When `getOrgDomain` returns empty string it would be false I think.

apps/api/pages/api/availability/_get.ts
leog: We could, we can tackle it as an improvement later on if you agree.

apps/web/pages/[user]/[type].tsx
Comment (zomars) [reply apps/web/pages/[user]/[type].tsx:27]: Can we get the hostname from getServerSideProps?

apps/web/pages/api/auth/forgot-password.ts
hariombalhara: Screenshot 2023-06-14 at 11 07 44 AM
So email is no longer unique at DB level. Do we want to make it unique or not?

apps/api/pages/api/availability/_get.ts
hariombalhara: Absolutely.

apps/web/middleware.ts
hariombalhara: Being discussed separately. So resolving this.

apps/web/pages/[user]/[type].tsx
```diff
@@ -70,7 +93,7 @@ async function getUserPageProps(context: GetStaticPropsContext) {
   const { type: slug, user: username } = paramsSchema.parse(context.params);
   const ssg = await ssgInit(context);
 
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
     where: {
       /** TODO: We should standarize this */
       username: username.toLowerCase().replace(/( |%20)/g, "+"),
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:99]: **[Blocker - Org]**
The org context is missing here. I tried adding it but the problem is that this page is SSG and we don't have access to context.req. I think we would have to make it a regular SSR route to achieve this.
This is what happens https://www.loom.com/share/e9aa0b84ea9141e1a24d3940f567f1d1

apps/web/pages/[user]/calendar-cache/[month].tsx
Comment (hariombalhara) [reply apps/web/pages/[user]/calendar-cache/[month].tsx:19]: I have faced a similar issue here https://github.com/calcom/cal.com/pull/8993/files#r1229091228

apps/web/pages/api/logo.ts
```diff
@@ -118,7 +118,7 @@ async function getTeamLogos(subdomain: string) {
     }
     // load from DB
     const { default: prisma } = await import("@calcom/prisma");
-    const team = await prisma.team.findUnique({
+    const team = await prisma.team.findFirst({
       where: {
         slug: subdomain,
```
hariombalhara, line 123: Should we check for metadata.isOrganization here?

apps/web/pages/api/user/avatar.ts
```diff
@@ -16,10 +17,18 @@ const querySchema = z
 
 async function getIdentityData(req: NextApiRequest) {
   const { username, teamname } = querySchema.parse(req.query);
+  const { currentOrgDomain, isValidOrgDomain } = orgDomainConfig(req.headers.host ?? "");
```
hariombalhara, line 20: I think as a followup, we can cleanup this recomputation of orgDomain at all places. We are already computing it in middleware, so we can set it on req directly as a header maybe?

apps/web/pages/api/user/avatar.ts
```diff
@@ -16,10 +17,18 @@ const querySchema = z
...
 
   if (username) {
-    const user = await prisma.user.findUnique({
-      where: { username },
+    const user = await prisma.user.findFirst({
```
hariombalhara, line 23: Follow: Also we can create a wrapper to get user/team that would take care of this org context all by itself.

packages/features/auth/lib/getServerSession.ts
```diff
@@ -46,7 +46,7 @@ export async function getServerSession(options: {
     return cachedSession;
   }
 
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
     where: {
       email: token.email.toLowerCase(),
```
hariombalhara, line 51: **[Blocker - Non-Org]**
Email being non-unique, I think it needs org context.

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -69,7 +69,7 @@ const providers: Provider[] = [
         throw new Error(ErrorCode.InternalServerError);
       }
 
-      const user = await prisma.user.findUnique({
+      const user = await prisma.user.findFirst({
```
hariombalhara, line 72: This one too needs the context.

packages/features/ee/organizations/api/subteams.ts
```diff
@@ -0,0 +1,36 @@
...
+
+async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const parsedQuery = querySchema.safeParse(req.query);
+
+  if (!parsedQuery.success) throw new HttpError({ statusCode: 400, message: parsedQuery.error.message });
+
+  const {
+    data: { org: slug },
+  } = parsedQuery;
+  if (!slug) if (!slug) return res.status(400).json({ message: "Org is needed" });
```
hariombalhara, line 21:
```suggestion
  if (!slug) return res.status(400).json({ message: "Org is needed" });
```

packages/features/ee/sso/lib/saml.ts
```diff
@@ -31,7 +31,7 @@ export const isSAMLAdmin = (email: string) => {
 };
 
 export const samlTenantProduct = async (prisma: PrismaClient, email: string) => {
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
     where: {
       email,
```
hariombalhara, line 36: I am not sure how the org context would reach here but there has to be one here I guess.

packages/features/eventtypes/components/OrganizationEventTypeFilter.tsx
```diff
@@ -0,0 +1,118 @@
+import { useSession } from "next-auth/react";
```
hariombalhara, line 1: followup: It can be replaced with [TeamsFilter ](https://github.com/calcom/cal.com/pull/9417/files#diff-08a37fc124270dd05eb90e0e85aa8d77cff174f1bfcde23e84c371ced895570b)component

packages/lib/constants.ts
```diff
@@ -8,7 +8,7 @@ export const WEBAPP_URL =
   RAILWAY_STATIC_URL ||
   HEROKU_URL ||
   RENDER_URL ||
-  "http://localhost:3000";
+  "http:/localhost:3000";
```
hariombalhara, line 11:
```suggestion
  "http://localhost:3000";
```

packages/lib/server/checkRegularUsername.ts
```diff
@@ -5,7 +5,7 @@ export async function checkRegularUsername(_username: string) {
   const username = slugify(_username);
   const premium = !!process.env.NEXT_PUBLIC_IS_E2E && username.length < 5;
 
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
     where: { username },
```
hariombalhara, line 9: context should be here as well

packages/trpc/server/routers/loggedInViewer/deleteMe.handler.ts
```diff
@@ -20,7 +20,7 @@ type DeleteMeOptions = {
 
 export const deleteMeHandler = async ({ ctx, input }: DeleteMeOptions) => {
   // Check if input.password is correct
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 23:
```suggestion
  const user = await prisma.user.findUnique({
```

packages/trpc/server/routers/loggedInViewer/deleteMeWithoutPassword.handler.ts
```diff
@@ -12,7 +12,7 @@ type DeleteMeWithoutPasswordOptions = {
 };
 
 export const deleteMeWithoutPasswordHandler = async ({ ctx }: DeleteMeWithoutPasswordOptions) => {
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 15:
```suggestion
  const user = await prisma.user.findUnique({
```

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -0,0 +1,132 @@
...
+  if (data.error?.code === "domain_taken")
+    throw new TRPCError({ code: "CONFLICT", message: "domain_taken_project" });
+
+  return true;
+};
+
+export const createHandler = async ({ input }: CreateOptions) => {
+  const { slug, name, adminEmail, adminUsername, check } = input;
+
+  const userCollisions = await prisma.user.findFirst({
```
hariombalhara, line 53:
```suggestion
  const userCollisions = await prisma.user.findUnique({
```

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -0,0 +1,132 @@
...
+    where: {
+      email: adminEmail,
+    },
+  });
+
+  const slugCollisions = await prisma.team.findFirst({
```
hariombalhara, line 59: Can anyone reserve an organization slug? What happens if someone not from acme(doesn't even have acme verified email), uses acme slug. Do we want to prevent it?

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -0,0 +1,132 @@
...
+    .update(`${adminEmail}${process.env.CALENDSO_ENCRYPTION_KEY}`)
+    .digest("hex");
+  const hashedPassword = await hashPassword(password);
+
+  if (check === false) {
+    const createOwnerOrg = await prisma.user.create({
+      data: {
+        username: adminUsername,
+        email: adminEmail,
+        emailVerified: new Date(),
```
hariombalhara, line 82: We are not actually verifying email right now. Assuming, it's' planned for 3.1, it needs a ticket.

packages/trpc/server/routers/viewer/organizations/setPassword.handler.ts
```diff
@@ -0,0 +1,57 @@
...
+import type { TSetPasswordSchema } from "./setPassword.schema";
+
+type UpdateOptions = {
+  ctx: {
+    user: NonNullable<TrpcSessionUser>;
+  };
+  input: TSetPasswordSchema;
+};
+
+export const setPasswordHandler = async ({ ctx, input }: UpdateOptions) => {
```
hariombalhara, line 19: It being a user password change only, I think we should be able to reuse auth/changePassword

packages/features/ee/sso/lib/saml.ts
```diff
@@ -31,7 +31,7 @@ export const isSAMLAdmin = (email: string) => {
 };
 
 export const samlTenantProduct = async (prisma: PrismaClient, email: string) => {
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 34:
```suggestion
  const user = await prisma.user.findUnique({
```

hariombalhara (commented):
Gone through all the files and left some comments. Once those are addressed, I am happy to merge. 
Awesome work @leog !!

packages/trpc/server/routers/viewer/organizations/create.handler.ts
leog: Anyone can as this will be part of the Enterprise tier and handled by us mainly first, not self-served. But even if it was public, we don't mind anyone squatting org slugs, probably not going to happen either way.

packages/trpc/server/routers/viewer/organizations/create.handler.ts
leog: Yes we do, the org creation wizard requests a code sent to the entered email before actually creating the org in the DB.

packages/emails/src/templates/TeamInviteEmail.tsx
```diff
@@ -121,7 +133,12 @@ export const TeamInviteEmail = (
             marginTop: "32px",
             lineHeightStep: "24px",
           }}>
-          <>{props.language("email_no_user_signoff", { appName: APP_NAME })}</>
+          <>
+            {props.language("email_no_user_signoff", {
+              appName: APP_NAME,
+              entity: props.language(props.isOrg ? "organization" : "team").toLowerCase(),
```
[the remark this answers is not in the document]
leog: Forgot to remove it from here, done. This string is actually from "Cal.com team" nothing we want to convert to org 😅

apps/web/pages/[user]/calendar-cache/[month].tsx
Comment (leog) [reply apps/web/pages/[user]/calendar-cache/[month].tsx:19]: This file is addressed by this other PR: https://github.com/calcom/cal.com/pull/9498

apps/web/pages/[user]/[type].tsx
Comment (leog) [reply apps/web/pages/[user]/[type].tsx:27]: Made changes to use getServerSideProps for this file now.

apps/web/pages/[user]/[type].tsx
Comment (leog) [reply apps/web/pages/[user]/[type].tsx:98]: Made changes to use getServerSideProps for this file now.

apps/web/pages/api/logo.ts
leog: Done, thanks

packages/lib/server/checkRegularUsername.ts
leog: Done, thanks!

zomars (approved):
LFG!!!!

LFG

packages/features/shell/Shell.tsx
```diff
@@ -348,8 +359,8 @@ function UserDropdown({ small }: { small?: boolean }) {
                   <span className="text-default truncate pb-1 font-normal">
                     {user.username
                       ? process.env.NEXT_PUBLIC_WEBSITE_URL === "https://cal.com"
-                        ? `cal.com/${user.username}`
-                        : `/${user.username}`
+                        ? `${orgBranding && orgBranding.slug}cal.com/${user.username}`
+                        : `${orgBranding && orgBranding.slug}/${user.username}`
```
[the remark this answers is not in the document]
leog: This will be covered by the sidebar PR from @roae 🙏

apps/web/pages/[user]/[type].tsx
Comment (leog) [reply apps/web/pages/[user]/[type].tsx]: I see, thanks for stepping in!

packages/emails/src/templates/OrganizationAccountVerifyEmail.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <div style={{ lineHeight: "6px" }}>
+        <p style={{ fontWeight: 400, lineHeight: "24px" }}>
+          <>
+            {props.language("happy_scheduling")} <br />
+            <a
+              href={`mailto:${SUPPORT_MAIL_ADDRESS}`}
+              style={{ color: "#3E3E3E" }}
+              target="_blank"
+              rel="noreferrer">
+              <>{props.language("the_calcom_team")}</>
```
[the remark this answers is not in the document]
leog: Done

PR #8997, opened 2023-05-19
fix/replacing bookingAttendees for bookingUid

alannnc:
## What does this PR do?

- Replace bookingAttendees length for bookingUid, if required we fetch booking attendees on server.

## Type of change

## Checklist

### DIFF
--- a/apps/web/components/booking/AvailableTimes.tsx
+++ b/apps/web/components/booking/AvailableTimes.tsx
@@ -72,7 +72,7 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
       slotUtcStartDate: slot.time,
       eventTypeId,
       slotUtcEndDate: dayjs(slot.time).utc().add(duration, "minutes").format(),
-      bookingAttendees: bookingAttendees || undefined,
+      bookingUid: slot.bookingUid,
     });
   };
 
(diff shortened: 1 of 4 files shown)


apps/web/components/booking/AvailableTimes.tsx
```diff
@@ -72,7 +72,7 @@ const AvailableTimes: FC<AvailableTimesProps> = ({
       slotUtcStartDate: slot.time,
       eventTypeId,
       slotUtcEndDate: dayjs(slot.time).utc().add(duration, "minutes").format(),
-      bookingAttendees: bookingAttendees || undefined,
+      bookingUid: slot.bookingUid,
```
alannnc, line 75: Instead of sending total bookingAttendees we send currentSlotBooking UID

packages/trpc/server/routers/viewer/slots/reserveSlot.handler.ts
```diff
@@ -40,8 +40,13 @@ export const reserveSlotHandler = async ({ ctx, input }: ReserveSlotOptions) =>
   // If this is a seated event then don't reserve a slot
   if (eventType.seatsPerTimeSlot) {
     // Check to see if this is the last attendee
-    if (bookingAttendees) {
-      const seatsLeft = eventType.seatsPerTimeSlot - bookingAttendees;
+    const bookingWithAttendees = await prisma.booking.findFirst({
```
alannnc, line 43: Fetch current attendees on server

apps/web/components/booking/AvailableTimes.tsx
alannnc, line 75: Instead of sending total bookingAttendees we send currentSlotBooking UID

JeroenReumkens (approved):
LGTM! Thanks a lot @alannnc! 

@roae The main reason for this change is because I asked it. Before this change this reserve mutation expected to have the data of the bookingattendees client side. That means that the booker needs to do an additional query, only to then send it to the backend and not use it themselves. So imo it made more sense to query this data directly in the backend, which is faster as well. That way I didn't need to add an additional query to the new booker.

PR #9013, opened 2023-05-21
chore: Add the or operator to the emails package.json as well

tachyonicbytes:
## What does this PR do?
Adds `docker-compose up -d` to the emails package.json as well, same as it is in prisma.

`yarn dx` did not work for me before.

## Type of change

## How should this be tested?

- [ ] run `yarn dx`.


leog (approved):
Looks good to me, good catch @TachyonicBytes

packages/emails/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "dx": "docker compose up -d"
+    "dx": "docker compose up -d || docker-compose up -d"
```
zomars, line 7: Do we actually use docker in the calcom/emails package? If not, should we remove it instead?

zomars (dismissed):

PR #9028, opened 2023-05-22
fix: /getting-started/connected-calendar Toggle the first calendar on

### LINKED ISSUE #4384: [CAL-115] /onboarding /connected-calendar - Toggle the first calendar on
- [ ] toggle the first calendar on by default

CleanShot 2022-09-12 at 11 12 25@2x

asadath1395:
## What does this PR do?

[CAL-115] /getting-started/connected-calendar Toggle the first calendar on

Fixes #4384 

https://github.com/calcom/cal.com/assets/37188392/aa9a0179-a355-464d-84c6-15fde1b07762

## Type of change

## How should this be tested?

1. Create a new account
2. Follow the steps and connect any calendar
3. You should see the first calendar selected by default

## Checklist


packages/trpc/server/routers/loggedInViewer/connectedCalendars.handler.ts
```diff
@@ -48,6 +48,15 @@ export const connectedCalendarsHandler = async ({ ctx }: ConnectedCalendarsOptio
       So create a default destination calendar with the first primary connected calendar
       */
     const { integration = "", externalId = "", credentialId } = connectedCalendars[0].primary ?? {};
+    // Select the first calendar matching the primary by default since that will also be the destination calendar
+    if (externalId) {
+      const calendarIndex = connectedCalendars[0].calendars.findIndex(
```
CarinaWolli, line 53: `connectedCalendars[0].calendars` can be undefined. This causes a type error (also in all other places were `connectedCalendars[0].calendars` is used)

CarinaWolli (changes requested):
I am getting the error "Something went wrong when toggeling ..." Here is a loom to see what I tried to do: https://www.loom.com/share/0c4c4360ba5e4bbd8bea6ca9171a86ce

packages/trpc/server/routers/loggedInViewer/connectedCalendars.handler.ts
asadath1395: If `externalId ` is present, then it means there should be atleast one calendar as per my understanding. Ref: https://github.com/calcom/cal.com/blob/main/packages/core/CalendarManager.ts#L72. I could be wrong though.

CarinaWolli: We still getting type errors though: https://github.com/calcom/cal.com/actions/runs/5043948903/jobs/9046435316?pr=9028

```suggestion
      const calendarIndex = connectedCalendars[0].calendars?.findIndex(
```

CarinaWolli (changes requested):
I would suggest making the changes directly to `steps-views/ConnectedCalendars` by calling the `/api/availability/calendar` endpoint for the first calendar

packages/trpc/server/routers/loggedInViewer/connectedCalendars.handler.ts
```diff
@@ -48,6 +54,19 @@ export const connectedCalendarsHandler = async ({ ctx }: ConnectedCalendarsOptio
       So create a default destination calendar with the first primary connected calendar
       */
     const { integration = "", externalId = "", credentialId } = connectedCalendars[0].primary ?? {};
+    // Select the first calendar matching the primary by default since that will also be the destination calendar
```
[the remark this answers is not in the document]
asadath1395: @CarinaWolli We can't do that in the client side because the `isSelected` value comes from the db, so i have added a new optional query parameter to just do this in the onboarding screen. There is one more bug https://github.com/calcom/cal.com/blob/main/apps/web/pages/settings/my-account/calendars.tsx#L190 because of which the first item is always enabled in `apps/installed/calendar`.

asadath1395:
@CarinaWolli I have addressed your concern, please re-review. Thanks

CarinaWolli (approved):
Looks good to me now and works great! Thank you 🙏

PR #9030, opened 2023-05-22
Fix: Embed - Auto theme configured using embed `config` doesn't work when there is non-system theme configured in appearance. 

### LINKED ISSUE #9027: [CAL-1732] Embed Auto theme configuration is ignored.
If embed theme is configured to be auto, it doesn't honour that. Only dark and light works. It is because auto is stored as null which is falsy and thus it falls back to configured appearance for booking page.

Also, theme in config isn't typed. So, provide the definition in TypeScript as well.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1732](https://linear.app/calcom/issue/CAL-1732/embed-auto-theme-configuration-is-ignored)</sub>

hariombalhara:
## What does this PR do?

Fixes #9027 

Now, we differentiate b/w explicitly set auto theme and non-configured theme.

## Type of change

## How should this be tested?

See [loom ](https://www.loom.com/share/4cb5c0e34e374678b978e430cf8b7077)

## Checklist

### DIFF
--- a/packages/embeds/embed-core/playground.ts
+++ b/packages/embeds/embed-core/playground.ts
@@ -55,6 +55,7 @@ if (only === "all" || only === "ns:second") {
         iframeAttrs: {
           id: "cal-booking-place-second-iframe",
         },
+        theme: "auto",
       },
     },
   ]);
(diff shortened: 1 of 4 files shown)


sean-brydon (approved):
LGTM - Havent tested but loom + code looks good

PR #9040, opened 2023-05-22
fix: Route Builder rules should be case insensitive 

### LINKED ISSUE #5439: [CAL-336] Route Builder rules should be case insensitive
Right now Route builder rules are case-sensitive.

To give more context on this, the text operators should be case insensitive

e.g.

[hariom@cal.com](mailto:hariom@cal.com)  [HARIOM@CAL.COM](mailto:HARIOM@CAL.COM) should be considered same in context of rules evaluation

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-335](https://linear.app/calcom/issue/CAL-335/[cal-336]-route-builder-rules-should-be-case-insensitive)</sub>

jemiluv8:
## What does this PR do?
Override jsonLogic operators on string operands to allow for case insensitive comparisons

Fixes #5439

[Case insensitive "equals" comparison](https://www.loom.com/share/c03892743934450aba0e76d5a5a97a4d)
[Case insensitive "not equals" comparison](https://www.loom.com/share/9a15da4043324d6aa974fdaa8267d975)

[Case insensitive "contains" comparison](https://www.loom.com/share/61e636b6e11b4350831e67027fea5e3c)
[Case insensitive "not contains" comparison](https://www.loom.com/share/c8f4402e6cc74875a0d6ecd07bd922d7)

## Type of change

## How should this be tested?

- [ ] Install the "Routing Forms" app
- [ ] Create a new form
- [ ] Create form fields under the "Form" tab
- [ ] Go to the routing tab and create a rule for any of the string operators ("contains", "not contains", "equals", "not equals")
- [ ] Use the test preview on the left sidebar to fill out an example form
- [ ] Use the "Test Routing" button to confirm the case insensitive comparison of the form values for relevant rules

Edit by @hariombalhara 
- We haven't implemented the same for Reporting because [Prisma as of this moment doesn't support case insensitive JSON querying](https://github.com/prisma/prisma/issues/7390).

## Checklist

/claim #5439


packages/app-store/routing-forms/lib/jsonLogicOverrides.ts
```diff
@@ -0,0 +1,32 @@
...
+});
+
+jsonLogic.add_operation("!=", function (a: any, b: any) {
+  return normalize(a) != normalize(b);
+});
+
+jsonLogic.add_operation("in", function (a: any, b: any) {
+  const first = normalize(a);
+  const second = normalize(b);
+  if (!second || typeof second.indexOf === "undefined") return false;
```
hariombalhara, line 28: What would be the case when second.indexOf is undefined. As per my testing second is either an array of string or string. I couldn't find such a case.

packages/app-store/routing-forms/lib/jsonLogicOverrides.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import jsonLogic from "json-logic-js";
+
+// converts input to lowercase if string
+function normalize<T extends string | string[]>(input: T): T {
+  if (typeof input === "string") {
+    return input.toLowerCase() as T;
+  }
+  if (input instanceof Array) {
```
hariombalhara, line 9: A failing test caught this bug where input can sometimes be an array as well(for MultiSelect it is)

hariombalhara (changes requested):
Looks good so far except 1 bug where MultiSelect equals stopped working.

That was because 'in' operator got second argument as an array of string which wasn't handled in normalize. So, one operand got lowercased but not the other one causing in operator to always fail. I have fixed that case.

@jemiluv8 I would suggest you to do one more round of testing with all the operators and field types combination.

Also, there is an open comment.

packages/app-store/routing-forms/lib/jsonLogicOverrides.ts
jemiluv8: The "in" function was copied over from the jsonLogic github repository. 
My intention was to not change anything besides normalizing the input.

I assumed it was to cater for a corner case but will review and decide whether
to modify it with the expectation of second being string or string[] as you found
out from your tests.

packages/app-store/routing-forms/lib/jsonLogicOverrides.ts
jemiluv8: Got it.

jemiluv8:
@hariombalhara, I just completed another round of testing and all looks fine.
I admit I didn't pay much attention to select/multiselect field types and their
respective operations

**About the redundant check in our override for the "in" operator**
I decided to remove the redundant check for the ```typeof second.indexOf === "undefined"```.

Like I mentioned in the comment above I tried to copy over the exact implementation of
various jsonLogic operators from the repo. For the "in" operator, the library assumed the 
second operand could be of any type and was therefore guarding against situations where
it may not be a string or an array.

Our current useage of the jsonLogic library ensures that the "in" operator will always get either a
string or string[]. That makes the extra check redundant.

playwright.config.ts
```diff
@@ -11,12 +11,12 @@ const outputDir = path.join(__dirname, "test-results");
 // Dev Server on local can be slow to start up and process requests. So, keep timeouts really high on local, so that tests run reliably locally
 
 // So, if not in CI, keep the timers high, if the test is stuck somewhere and there is unnecessary wait developer can see in browser that it's stuck
-const DEFAULT_NAVIGATION_TIMEOUT = process.env.CI ? 15000 : 50000;
-const DEFAULT_EXPECT_TIMEOUT = process.env.CI ? 15000 : 50000;
+const DEFAULT_NAVIGATION_TIMEOUT = process.env.CI ? 15000 : 120000;
+const DEFAULT_EXPECT_TIMEOUT = process.env.CI ? 15000 : 120000;
```
hariombalhara, line 15: Added this change here because sometimes dev server can be too slow and it times out unncessarily.

hariombalhara (approved):

PR #9050, opened 2023-05-22
Delete .infisical.json

PeerRich:
I dont think we're planning on adding it anymore, right? @zomars


zomars (approved):
Not for now. Will merge

pumfleet (approved):

PR #8917, opened 2023-05-16
Team webhooks

CarinaWolli:
## What does this PR do?

- Adds webhooks for teams. Only admins and owners can create, edit and delete webhooks:

Admin/Owner: 
Screenshot 2023-05-16 at 14 19 47

Member: 
Screenshot 2023-05-16 at 14 20 06

- Also adds event type webhooks for user event types. Until now only team event types could have even type webhooks

Fixes #8815

## Type of change

- New feature (non-breaking change which adds functionality)
- Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- Create team, event type and user webhooks 
- Test editing and deleting webhooks
- Test with users that have different membership roles 
- Test if webhooks trigger as expected

PR #9052, opened 2023-05-23
Follow-up team webhooks

CarinaWolli:
## What does this PR do?

Follow-up for https://github.com/calcom/cal.com/pull/8917

- Fixes 'Webhook subscriber url is already defined' error when editing event type webhook. 
- Removes the 'Form Submitted' trigger from team webhooks and from event type webhooks

## Type of change


hariombalhara (approved):
LGTM !

PR #8203, opened 2023-04-11
fix: cancelled booking doesn't show the booker

G3root:
## What does this PR do?

Fixes #8185 

changes made:
- [x] prevent attendee deletion
- [x] added a e2e test  

## Type of change

- Bug fix (non-breaking change which fixes an issue)

PR #9055, opened 2023-05-23
fix: use the correct locale for event name in booking detail page

G3root:
## What does this PR do?

previously event name saved in the db is used for booking detail page as event title, which is in the locale of the organizer. this change was introduced in #7888 to prevent "Nameless" replacing attendee while cancelling a booking. as we are now not deleting the attendees #8203 , above mentioned issue won't happen. this PR aims to fallback to `eventName` for event title, which also fixes the wrong attendee locale issue.

before:

Screenshot 2023-05-23 at 12-35-45 Your booking has been confirmed Cal com

after:

Screenshot 2023-05-23 at 12-35-14 Your booking has been confirmed Cal com

## Type of change


apps/web/pages/booking/[uid].tsx
```diff
@@ -400,7 +400,7 @@ export default function Success(props: SuccessProps) {
                       </>
                     )}
                     <div className="font-medium">{t("what")}</div>
-                    <div className="col-span-2 mb-6 last:mb-0">{props.bookingInfo.title}</div>
+                    <div className="col-span-2 mb-6 last:mb-0">{eventName}</div>
```
Comment (alannnc) [inline apps/web/pages/booking/[uid].tsx:403]: I'm not seeing a fallback behaviour, but is this the correct value to display instead, right?

Comment (G3root) [reply apps/web/pages/booking/[uid].tsx:403]: yeah

PeerRich (approved):

PR #9058, opened 2023-05-23
Add core team to UI ownership

hariombalhara:
Update  CODEOWNER for UI


sean-brydon (approved):
Agreed - I'll make a quick thread about this

PR #9059, opened 2023-05-23
Fixes timezone of {EVENT_DATE}

CarinaWolli:
## What does this PR do?

Fixes timezone of {EVENT_DATE} variable. It was shown in UTC and not the attendee/organizer timezone. 

## Type of change

## How should this be tested?

- Organizer should be in Los Angeles time zone
- Create workflow to send email/sms to organizer when new event is booked with the {EVENT_DATE} variable 
- Set an event type active and make a test booking on this event type 
   - Book any time that would be the next day already in UTC (so for example 6pm)
   - Check the workflow notification and see that it shows the correct day (before it showed the day next day)


emrysal (approved):
Looks OK - lgtm

Issue #9047, opened 2023-05-22
[CAL-1740] Implement app theme toggle in settings appearance.


sean-brydon:
Implement theme toggle in settings that controls our APP theme for the user and not their booking theme. -> Force lightmode even if system theme is dark. 

@Jaibles can we  have a figma design for this 

I made but we could probably do with having different images for the selectors/move booking appearance elsewhere?

 https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/2.-Cal---Live?type=design&node-id=26-60131&t=XQRa4jpyHwcLCv3S-4

<sub>[CAL-1740](https://linear.app/calcom/issue/CAL-1740/implement-app-theme-toggle-in-settings-appearance)</sub>

Subhojit-Dey1234:
I have created the pull request for the issue.

rkreddy99:
https://www.loom.com/share/cdc94fe097d54cd6bf2ce09dcffff403 
@Jaibles I have worked on this issue, is this what were you expecting

PS: the labelling part is not done yet, i have just implemented the feature

Jaibles:
@rkreddy99 functionally that's it exactly. We just need to do new illustrations to have ones that reflect the app so we have some visual differentiation between changing appearance for the app vs the booking page. I'll work on them today/tomorrow.

Jaibles:
<img width="1435" alt="CleanShot 2023-05-23 at 15 33 19@2x" src="https://github.com/calcom/cal.com/assets/4536123/2e8acd64-a1fe-457a-9fad-be4bb881958f">
These work cc @sean-brydon ?
https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/2.-Cal---Live?type=design&node-id=26%3A60253&t=p748TaivFZz0ii2W-1

Jaibles:
https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/Live?type=design&node-id=26%3A60119&mode=design&t=r8BbEEtmgvnfnFTK-1

sean-brydon:
Already implemented

PR #8975, opened 2023-05-18
fix: dark icons

saurabbot:
## What does this PR do?
Fixes #8966 
in the /settings/admin/apps/conferencing the icons werent visible cuz of the dark mode.

Fixes # (issue)

Screenshot 2023-05-18 at 7 13 54 PM

Screenshot 2023-05-19 at 1 07 38 PM
Screenshot 2023-05-19 at 1 12 24 PM
Screenshot 2023-05-19 at 1 12 35 PM

ran the project after changes works fine

PR #8925, opened 2023-05-16
fix: Skeleton Loader

saurabbot:
## What does this PR do?
fixes #8922 

there seems to be a loading check already in the upgradeTip component so the loader decision is taken care of in the parent component
Fixes # (issue)
Adds a loader to the routing forms page

**Environment**: Staging(main branch) / Production

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Chore (refactoring code, technical debt, workflow improvements)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

## How should this be tested?

- [x] Test A (for user who has a paid plan)
- [x] Test B (not a paid plan user)

PR #9076, opened 2023-05-23
feat: toggleTheme

saurabbot:
## What does this PR do?
This pr adds a feature which allows the user the app theme.
Implements #9047 

https://www.loom.com/share/c3f222b633954baeb6305ec61e3f133d

## How should this be tested?

- [x] Test A->changing the themes and checking the change in db
- [ ] Test B


apps/web/pages/settings/my-account/appearance.tsx
```diff
@@ -85,45 +87,82 @@ const AppearanceView = () => {
   if (!user) return null;
 
   const isDisabled = isSubmitting || !isDirty;
+  console.log(user.appTheme, formMethods, "user");
 
   return (
     <Form
       form={formMethods}
       handleSubmit={(values) => {
+        console.log(values, "updatemutate");
```
leog, line 96: This should be removed

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -26,6 +26,7 @@ type UpdateProfileOptions = {
 
 export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions) => {
   const { user } = ctx;
+  console.log(input, "updateinput");
```
leog, line 29: This should be removed

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -338,7 +338,7 @@ type FormProps<T extends object> = { form: UseFormReturn<T>; handleSubmit: Submi
 
 const PlainForm = <T extends FieldValues>(props: FormProps<T>, ref: Ref<HTMLFormElement>) => {
   const { form, handleSubmit, ...passThrough } = props;
-
+  console.log(form, "ww");
```
leog, line 341: This should be removed

leog (changes requested):
Thanks for the contribution @firfox341. It does look good! Left some comments regarding code to be removed.

Also, what would be the rationale behind using a new column in the database versus the existing `theme` column?

leog:
Makes sense @firfox341. I would like to invoke the presence of @Jaibles for this one. I'm not sure we want to maintain both variables at the same time; if you choose dark mode why would you like your booking page to be light mode or automatic? Just in case, his blessing is always a good thing to have 😄

Jaibles:
@leog  @saurabbot is the question: "Should we allow people to have different dark/light mode settings for app.cal.com vs their public booking page"? If so, the answer is definitely yes. 

I think a perfectly likely scenario is people who prefer dark mode wanting that in the app, but wanting to go with system settings for the public page so that their bookers get to see what they prefer.

Jaibles:
Settings - My Account - Appearance

We do need to use the new svgs though. I haven't had a chance to optimise them yet in terms of filesize.

packages/prisma/.env
zomars, line 1: Why was this file deleted? Let's put it back:

`git checkout origin/main packages/prisma/.env`

packages/ui/components/form/inputs/Input.tsx
```diff
@@ -338,7 +338,6 @@ type FormProps<T extends object> = { form: UseFormReturn<T>; handleSubmit: Submi
 
 const PlainForm = <T extends FieldValues>(props: FormProps<T>, ref: Ref<HTMLFormElement>) => {
   const { form, handleSubmit, ...passThrough } = props;
-
```
zomars, line 341: Unnecessary diff here. Let's revert it.

apps/web/pages/settings/my-account/appearance.tsx
```diff
@@ -85,6 +87,7 @@ const AppearanceView = () => {
   if (!user) return null;
 
   const isDisabled = isSubmitting || !isDirty;
+  console.log(user.appTheme, formMethods, "user");
```
zomars, line 90:
```suggestion
```

packages/features/settings/ThemeLabel.tsx
```diff
@@ -4,20 +4,20 @@ interface ThemeLabelProps {
   label: string;
   defaultChecked?: boolean;
   register: any;
+  page: string;
 }
 
 export default function ThemeLabel(props: ThemeLabelProps) {
   const { variant, label, value, defaultChecked, register } = props;
-
```
zomars, line 11: Missing padding here

zomars (changes requested):
Thank you for your contribution 🙏 Left some comments to be addressed before merging

Jaibles:
> gtg right ? is the yarn.lock change fine ? could you guys merge these prs also #8975 #8925

@saurabbot I'm assuming the new svgs I made aren't in yet so that should be blocking. I'll prepare them now.

Udit-takkar:
@saurabbot any update on this PR?

leog:
> Hey @Udit-takkar this was done long back and ready from my side, I was waiting for you guys to test the feature.

Hey @saurabbot. Actually, we were waiting for you to resolve the comments we did in the code (addressed already) and request a re-review so it returns to us to review again to be able to approve it. Take that into account for future contributions, so we can merge great contributions like yours faster 👍

leog:
> @leog gtg i guess..

@firefox341 did you tested the latest? There is some problem with loading the svgs and when saving, the toast message does not look good. Please test it locally again and tackle those issues.

leog:
@saurabbot are you able to continue working on this PR? we might need to close it if there is no more activity in it 😓

PR #9081, opened 2023-05-24
feat: email verification

sean-brydon:
## What does this PR do?
Implements email verification controlled by a feature flag

Loom: https://www.loom.com/share/0b0b9302a0e8499fad1d6444b013337c

## Type of change

## How should this be tested?

- [x] Login as ADMIN
- [x] Enable Feature flag
- [x] Verify your account
- [x] create a new account
- [x] verify flow works as expected


apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { token } = verifySchema.parse(req.query);
+
+  const foundToken = await prisma.verificationToken.findFirst({
+    where: {
+      token,
+      type: "VERIFY_ACCOUNT",
+    },
+  });
```
sean-brydon, line 19: Find the token that is of type verify account

apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,37 @@
...
+      email: foundToken?.identifier,
+    },
+    data: {
+      emailVerified: new Date(),
+    },
+  });
+
+  const hasCompletedOnboarding = user.completedOnboarding;
+
+  res.redirect(`${WEBAPP_URL}/${hasCompletedOnboarding ? "/event-types" : "/getting-started"}`);
```
sean-brydon, line 36: If you have already completed onboarding we redirect to event types

Makes the flow much nicer as you dont get a flash of wrong content

apps/web/pages/auth/verify-email.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const router = useRouter();
+  const { t } = useLocale();
+  const mutation = trpc.viewer.auth.resendVerifyEmail.useMutation();
+
+  useEffect(() => {
+    if (data?.isVerified) {
+      router.replace("/getting-started");
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [data?.isVerified]);
```
sean-brydon, line 26: If you navigate back to this tab after verifying your account we redirect to straight to getting started

packages/emails/templates/account-verify-email.ts
```diff
@@ -0,0 +1,51 @@
...
+export type EmailVerifyLink = {
+  language: TFunction;
+  user: {
+    name?: string | null;
+    email: string;
+  };
+  verificationEmailLink: string;
+};
+
+export const PASSWORD_RESET_EXPIRY_HOURS = 6;
```
sean-brydon, line 17:
```suggestion
```

apps/web/pages/signup.tsx
```diff
@@ -161,11 +166,19 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   const ssr = await ssrInit(ctx);
   const token = z.string().optional().parse(ctx.query.token);
 
+  const isEmailVerifyEnabled = await prisma.feature.findFirst({
+    where: {
+      slug: "email-verification",
+      enabled: true,
+    },
+  });
```
sean-brydon, line 174: Putting this here was a nicer approach than waiting for it to load on the client - it didnt always seem to load in time and disabling the button was weird if you didnt actually have it enabled 

@zomars

packages/prisma/migrations/20230523101834_email_verification_feature_flag/migration.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO
+  "Feature" (slug, enabled, description, "type")
+VALUES
+  (
+    'email-verification',
+    true,
+    'Enable email verification for new users',
+    'OPERATIONAL'
+  ) ON CONFLICT (slug) DO NOTHING;
```
sean-brydon, line 9: Enable by default

apps/web/pages/api/auth/signup.ts
```diff
@@ -17,7 +18,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   }
 
   const data = req.body;
-  const { email, password } = data;
+  const { email, password, language } = data;
```
hariombalhara, line 21: nit: Might be a good idea to introduce zod schema for this.

sean-brydon: I've actually done this in my signup migration branch - i can bring it over for this PR if youd like? :P

packages/features/auth/lib/verifyEmail.ts
```diff
@@ -0,0 +1,65 @@
...
+});
+
+interface VerifyEmailType {
+  username?: string;
+  email: string;
+  language?: string;
+}
+
+export const sendEmailVerification = async ({ email, language, username }: VerifyEmailType) => {
+  const token: string = randomBytes(32).toString("hex");
```
hariombalhara, line 21:
```suggestion
  const token = randomBytes(32).toString("hex");
```

packages/features/auth/lib/verifyEmail.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const sendEmailVerificationEnabled = await prisma.feature.findFirst({
+    where: {
+      slug: "email-verification",
+      enabled: true,
+    },
+  });
+
+  if (!sendEmailVerificationEnabled) {
+    console.log("Email verification is disabled - Skipping");
```
hariombalhara, line 32: nit: Better to use logger here

packages/features/auth/lib/verifyEmail.ts
```diff
@@ -0,0 +1,65 @@
...
+    return { ok: true, skipped: true };
+  }
+
+  await prisma.verificationToken.create({
+    data: {
+      identifier: email,
+      token,
+      type: TokenType.VERIFY_ACCOUNT,
+      expires: new Date(new Date().setHours(24)), // +1 day
```
hariombalhara, line 41: setHours doesn't seem to have 24 as a valid value. [The range is from 0-23.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours#:~:text=An%20integer%20between%200%20and%2023%20representing%20the%20hours.)

apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,37 @@
...
+  token: z.string(),
+});
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { token } = verifySchema.parse(req.query);
+
+  const foundToken = await prisma.verificationToken.findFirst({
+    where: {
+      token,
+      type: "VERIFY_ACCOUNT",
```
hariombalhara, line 17: Why do we need to use token type here? Isn't token unique ?

apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+  });
+
+  if (!foundToken) {
+    res.status(401).json({ message: "No token found" });
+  }
+
+  const user = await prisma.user.update({
```
hariombalhara, line 25: What if the token has expired?

apps/web/pages/api/auth/verify-email.ts
sean-brydon: Good point tho - it is unique on db level

apps/web/pages/signup.tsx
```diff
@@ -161,11 +166,19 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   const ssr = await ssrInit(ctx);
   const token = z.string().optional().parse(ctx.query.token);
 
+  const isEmailVerifyEnabled = await prisma.feature.findFirst({
+    where: {
+      slug: "email-verification",
```
hariombalhara, line 171: **Unrelated to this PR** but I had an observation.

With this approach there is no type safety with the slugs. What if we move the features list to a file? It would be faster as well no matter we do it on client or server.

Also, we might be able to tree shake the code as well as at compile time we know the status of the feature flag.

To allow the features list to be extended by self hosters, we can have 2 configs.
 features.base.json -> Commited in git and managed by us
 features.extended.json -> Not committed in git by us but if it exists we merge it over features.base.json
 
 cc @zomars

sean-brydon: We're actually meant to user the useFlagsMap hook - however, this for some reason was always returning null on the signup page (works everywhere else tho)

apps/web/pages/signup.tsx
```diff
@@ -161,11 +166,19 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
...
+    },
+  });
+
   const props = {
     isGoogleLoginEnabled: IS_GOOGLE_LOGIN_ENABLED,
     isSAMLLoginEnabled,
     trpcState: ssr.dehydrate(),
     prepopulateFormValues: undefined,
+    isEmailVerifyEnabled: !!isEmailVerifyEnabled,
   };
```
hariombalhara, line 182: Screenshot 2023-05-30 at 4 37 11 PM
I see some old verificationToken related code in getServerSideProps. Is that still valid?

sean-brydon: I believe this is for team invites off the top of my head

sean-brydon: Yeah it is ^

packages/features/auth/lib/verifyEmail.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  const params = new URLSearchParams({
+    token,
+  });
+
+  const { isRateLimited } = limiter.check(10, email); // 10 requests per minute
+
+  if (isRateLimited) {
+    throw new Error("Too many requests");
```
hariombalhara, line 52: Might be better to send 429 status code here. This would cause a 500

sean-brydon: Sure happy with that - we do this elsewhere too tho 🤔

apps/web/pages/signup.tsx
hariombalhara: `useFlagsMap` is client only(requires another request). I like this approach better where you can do it on server.
I just noticed that there is typesafety through flags/config.ts, we can introduce a helper method with typesafety later.

apps/web/pages/api/auth/verify-email.ts
sean-brydon: We might honestly not need it - i dont think i realised token was unique

packages/trpc/server/routers/viewer/auth/resendVerifyEmail.handler.ts
```diff
@@ -0,0 +1,19 @@
...
+type ResendEmailOptions = {
+  ctx: {
+    user: NonNullable<TrpcSessionUser>;
+  };
+};
+
+export const resendVerifyEmail = async ({ ctx }: ResendEmailOptions) => {
+  const email = await sendEmailVerification({
+    email: ctx.user.email,
+    username: ctx.user?.username ?? undefined,
```
hariombalhara, line 14:
```suggestion
    username: ctx.user.username ?? undefined,
```
or maybe make ResendEmailOptions["ctx"]["user"] optional

packages/trpc/server/routers/viewer/auth/resendVerifyEmail.handler.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { TrpcSessionUser } from "../../../trpc";
+
+type ResendEmailOptions = {
+  ctx: {
+    user: NonNullable<TrpcSessionUser>;
+  };
+};
+
+export const resendVerifyEmail = async ({ ctx }: ResendEmailOptions) => {
+  const email = await sendEmailVerification({
```
hariombalhara, line 12: nit: Might be a good idea to check if email is already verified or not.

packages/trpc/server/routers/loggedInViewer/shouldVerifyEmail.handler.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const shouldVerifyEmailHandler = async ({ ctx }: ShouldVerifyEmailType) => {
+  const { user } = ctx;
+  const isVerified = !!user.emailVerified;
+  const isCalProvider = user.identityProvider === "CAL"; // We dont need to verify on OAUTH providers as they are already verified by the provider
+
+  const obj = {
+    id: user.id,
+    email: user.email,
+    isVerified: (isVerified && isCalProvider) || !isCalProvider,
```
hariombalhara, line 17:
```suggestion
    isVerified: isVerified || !isCalProvider,
```
can be simplified like this.

apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,48 @@
...
+  const { token } = verifySchema.parse(req.query);
+
+  const foundToken = await prisma.verificationToken.findFirst({
+    where: {
+      token,
+    },
+  });
+
+  if (!foundToken) {
+    res.status(401).json({ message: "No token found" });
```
zomars, line 22:
```suggestion
    return res.status(401).json({ message: "No token found" });
```

apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  if (dayjs(foundToken?.expires).isBefore(dayjs())) {
+    res.status(401).json({ message: "Token expired" });
```
zomars, line 26: Without returns the rest of the code still runs, which may not be a desirable behavior.

```suggestion
    return res.status(401).json({ message: "Token expired" });
```

apps/web/pages/api/auth/verify-email.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // Delete token from DB after it has been used
+  await prisma.verificationToken.delete({
+    where: {
+      id: foundToken?.id,
+    },
+  });
+
+  const hasCompletedOnboarding = user.completedOnboarding;
+
```
zomars, line 46: Can we login the user here? Should we use the token as a One Time Password in case it opens up in another browser (as demoed in your loom)?

apps/web/pages/auth/verify-email.tsx
zomars: Can we just return a new view that says. "Email is already verified"? Maybe add an CTA that takes you back to the dashboard.

apps/web/pages/signup.tsx
zomars: Ill Allow It Spanish GIF

packages/features/ee/common/components/LicenseRequired.tsx
```diff
@@ -21,6 +21,7 @@ const LicenseRequired = ({ children, as = "", ...rest }: LicenseRequiredProps) =
   const Component = as || Fragment;
   const hasValidLicense = session.data ? session.data.hasValidLicense : null;
 
+  return children;
```
zomars, line 24:
```suggestion
```

packages/prisma/schema.prisma
```diff
@@ -278,13 +278,19 @@ model Membership {
...
+  TEAM_INVITE
+  VERIFY_ACCOUNT
+}
+
 model VerificationToken {
-  id         Int      @id @default(autoincrement())
+  id         Int        @id @default(autoincrement())
   identifier String
-  token      String   @unique
+  token      String     @unique
```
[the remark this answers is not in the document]
zomars: We aren't really changing the file here. So we can just restore it

`git checkout origin/main packages/prisma/schema.prisma`

zomars (commented):
So far so good @sean-brydon Still need to tests further before merging. Left some comments as well. 🙏🏽

hariombalhara (dismissed):
LGTM! Great work @sean-brydon 
There is just one comment that's left addressed https://github.com/calcom/cal.com/pull/9081/files#r1210204910

Tests done.
- With feature flag enabled(Default)
	- Created a new account
		- Landing on Email has been sent page.
		- Got an email. 
		- Able to reach onboarding page after clicking that link(In the browser I did signup on)
		- One onboarding is complete, I din't see the banner to verify your email. All good !!
	- Created a new account
		- Landing on Email has been sent page.
		- Retried verification email multiple times. Got multiple emails
		- Used one of the email links and flow worked like above
	- Created a new account
		- Didn't click on verification link email
		- Directly went to localhost:3000 and was taken to onboarding screen
		- After onboarding saw a banner in the app to verify my email
		- Clicked on it and received email again. Clicked on Email and the banner got removed.
	- Tried to reuse a token and it shows a Not Found error.
- Without feature flag enabled(Disabled manually)
	- I was able to signup and then went straight to onboarding after completion of which I didn't see any banner.

Blocking merge because CI is failing. Once it passes it has an approval from my side.

apps/web/pages/auth/verify-email.tsx
sean-brydon: I dont think it was clear what i meant by this.

if you switch back to this tab after you have verified your email. It navigates you to getting-started. Meaning you dont have to log in or hard refresh

apps/web/pages/api/auth/verify-email.ts
sean-brydon: Can this be tackled in a follow up? 

You're probably really unlikely to open the email in a different browser from the one you intend on using to log in with.

zomars: > Can this be tackled in a follow up?

I would insist in this same PR. But feel free to put in an issue so we can follow up.

apps/web/pages/signup.tsx
zomars: I just realized that we can achieve the same effect by just doing:

```
await ssr.viewer.features.map.prefetch();
```

Will test it out and update if it works

apps/web/pages/api/auth/verify-email.ts
zomars: @sean-brydon why was my suggestion ignored? We do need a return here. If there isn't the rest of the code will still run after the response which WILL cause unintended effects.

apps/web/pages/signup.tsx
```diff
@@ -29,10 +30,10 @@ type FormValues = {
 };
 
 export default function Signup({ prepopulateFormValues, token }: inferSSRProps<typeof getServerSideProps>) {
-  const { t } = useLocale();
+  const { t, i18n } = useLocale();
   const router = useRouter();
+  const flags = useFlagMap();
```
zomars, line 35: With #9374 this is now always defined on first render.

apps/web/server/lib/ssr.ts
```diff
@@ -26,6 +26,8 @@ export async function ssrInit(context: GetServerSidePropsContext) {
 
   // always preload "viewer.public.i18n"
   await ssr.viewer.public.i18n.fetch();
+  // So feature flags are available on first render
+  await ssr.viewer.features.map.prefetch();
```
zomars, line 30: Same code as #9374 just need to merge that one first.

packages/features/auth/lib/verifyEmail.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const sendEmailVerification = async ({ email, language, username }: VerifyEmailType) => {
+  const token = randomBytes(32).toString("hex");
+  const translation = await getTranslation(language ?? "en", "common");
+  const flags = await getFeatureFlagMap(prisma);
+
+  if (!flags["email-verification"]) {
+    log.warn("Email verification is disabled - Skipping");
+    return { ok: true, skipped: true };
+  }
```
zomars, line 32: I'd rather use the getFeatureFlagMap on server-side instead of calling individual features.

zomars (approved):
LGTM

shivamklr:
@sean-brydon If it is the wrong token, should we not show a proper page? Right now, it only shows the json body. Let me know your thoughts.

PR #9082, opened 2023-05-24
feat: event settings booker layout toggle

JeroenReumkens:
## What does this PR do?

Adds layout toggle for new booker layout. 

### How this works? (you can also watch the video)

We have 3 layout options for the booker page. You can change that in 2 places. 
First in your account's profile settings. These settings are used as a default for all events that don't have their own layout set. Then on the event-type page, you can override this setting for a specific event. In that case we use these settings instead of the global ones. 

When changing layouts, only those layouts are visible and allowed to toggle between. So if you only enabled a single layout, you also won't see the toggle to change layouts anymore. 

Fixes CAL-1703

Loom Video: 
The date bug shown in the video is fixed! 
https://www.loom.com/share/4614d127613e4965b7170f1ac5567696

## Type of change

## How should this be tested?

- [ ] Set layouts on profile level as well as for an event type, and ensure that the correct layouts are shown.


apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -159,6 +160,10 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps,
           }
         />
       </div>
+      <hr className="border-subtle [&:has(+div:empty)]:hidden" />
```
JeroenReumkens, line 163: This hidden classname is on there for now because the `BookerLayoutSelector` can hide itself based on a feature toggle. In that case it would result in two lines right after each other, which is ugly. So we hide one for now.

apps/web/components/eventtype/EventTypeSingleLayout.tsx
```diff
@@ -365,7 +365,7 @@ function EventTypeSingleLayout({
           <Button
             className="ml-4 lg:ml-0"
             type="submit"
-            loading={formMethods.formState.isSubmitting || isUpdateMutationLoading}
```
JeroenReumkens, line 368: This `isSubmitting` boolean never goes to false when an error throws. That resulted in the user needing to refresh if an error occurs. Using only the mutation boolean seems sufficient here.

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -96,6 +101,7 @@ export type FormValues = {
   hosts: { userId: number; isFixed: boolean }[];
   bookingFields: z.infer<typeof eventTypeBookingFields>;
   availability?: AvailabilityOption;
+  bookerLayouts: BookerLayoutSettings;
```
Comment (JeroenReumkens) [inline apps/web/pages/event-types/[type]/index.tsx:104]: This event type page only does the logic to get and pass the form properties into the mutation. And it does the validation here too. See methods down below.

apps/web/pages/settings/my-account/appearance.tsx
```diff
@@ -74,8 +77,12 @@ const AppearanceView = () => {
       await utils.viewer.me.invalidate();
       showToast(t("settings_updated_successfully"), "success");
     },
-    onError: () => {
-      showToast(t("error_updating_settings"), "error");
+    onError: (error) => {
+      if (error.message) {
+        showToast(error.message, "error");
```
JeroenReumkens, line 82: Added the ability to show a custom error message. Needed here because the messages for the booker layout are quite helpful.

packages/features/bookings/Booker/types.ts
```diff
@@ -42,5 +44,5 @@ export interface BookerProps {
 }
 
 export type BookerState = "loading" | "selecting_date" | "selecting_time" | "booking";
-export type BookerLayout = "small_calendar" | "large_timeslots" | "large_calendar" | "mobile";
+export type BookerLayout = BookerLayouts | "mobile";
```
JeroenReumkens, line 47: Reuse types here. Mobile isn't a valid setting for the end user, but is a layout that the component itself uses. That's why we add that still here.

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -183,6 +195,8 @@ function getProfileFromEvent(event: Event) {
     brandColor: profile.brandColor,
     darkBrandColor: profile.darkBrandColor,
     theme: profile.theme,
+    bookerLayouts:
+      event.bookerLayouts || ("defaultBookerLayouts" in profile ? profile.defaultBookerLayouts : null),
```
JeroenReumkens, line 199: If the event has a layout set, we pass that. If there's no layout set, we default to the profile settings of the user. If that's not set, we use null, which in the booker will be resolved to showing all 3 layouts with the month_view as default.

packages/features/settings/BookerLayoutSelector.tsx
```diff
@@ -0,0 +1,192 @@
+import * as RadioGroup from "@radix-ui/react-radio-group";
+import { Trans } from "next-i18next";
+import Link from "next/link";
+import { useCallback, useState } from "react";
+import { Controller, useFormContext } from "react-hook-form";
+
+import { classNames } from "@calcom/lib";
```
JeroenReumkens, line 7: Abstracted this file into a separate component so we can reuse it both in the event-types as well as in the appearance settings of the profile.

packages/features/settings/BookerLayoutSelector.tsx
```diff
@@ -0,0 +1,192 @@
...
+  fallbackToUserSettings,
+}: BookerLayoutSelectorProps) => {
+  const { control, getValues } = useFormContext();
+  const { t } = useLocale();
+  // Only fallback if event current does not have any settings, and the fallbackToUserSettings boolean is set.
+  const shouldShowUserSettings =
+    (fallbackToUserSettings && getValues(name || "bookerLayouts") === null) || false;
+
+  // @TODO: Enable this before merge, so we don't show the settings before release.
+  // const flags = useFlagMap();
```
JeroenReumkens, line 40: Added a feature toggle here. That way we can decide if we want to disable / enable the layout settings via a toggle :) Disabled it for now, because the migrations haven't run on the preview env yet. 

- [ ]  So todo is to enable this flag before merge.

packages/lib/validateBookerLayouts.ts
```diff
@@ -0,0 +1,29 @@
+import { bookerLayoutOptions, type BookerLayoutSettings } from "@calcom/prisma/zod-utils";
+
+export const validateBookerLayouts = (settings: BookerLayoutSettings) => {
```
JeroenReumkens, line 3: Reusable validation function that's used when submitting the forms.

packages/prisma/migrations/20230523102615_added_booker_layouts/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
```
JeroenReumkens, line 1: Add bookerlayouts to both event type as well as users. I think the json format makes the most sense here, right? 
See zod file a few files down for the format.

packages/prisma/migrations/20230524105015_added_newbooker_feature_flag/migration.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO
```
JeroenReumkens, line 1: Migration for the feature flag. I created this manually, I think that's what we're supposed to do, right?

packages/prisma/zod-utils.ts
```diff
@@ -539,3 +539,21 @@ export const allManagedEventTypeProps: { [k in keyof Omit<Prisma.EventTypeSelect
 export const unlockedManagedEventTypeProps = {
   ...pick(allManagedEventTypeProps, ["locations", "scheduleId", "destinationCalendar"]),
 };
+
+export const bookerLayoutOptions = ["month_view", "week_view", "column_view"] as const;
+// @TODO: HAHA. Let's just look the array above and find out why that's not working..
```
JeroenReumkens, line 544: If anyone knows how to fix this? 😂
`z.union(bookerLayoutOptions.map(layout > z.literal(layout)))` didn't seem to work somehow.

packages/ui/components/form/toggleGroup/ToggleGroup.tsx
```diff
@@ -34,58 +33,29 @@ const OptionalTooltipWrapper = ({
...
         {...props}
-        onValueChange={setValue}
+        onValueChange={onValueChange}
         className={classNames(
           "min-h-9 bg-muted border-default relative inline-flex gap-0.5 rounded-md border p-1",
           props.className,
           isFullWidth && "w-full"
         )}>
-        {/* Active toggle. It's a separate element so we can animate it nicely. */}
-        <span
```
JeroenReumkens, line 56: I decided to remove the animated background (for now?). It doesn't update properly when the current value changes, and it already took me quite some time to try to get it to work (it's implementation already changed a few times in the past weeks). I think it's not worth it to try to keep fixing it, so removed it for now. We can always later try and see if we can add it back in a better way.

apps/web/components/eventtype/EventAdvancedTab.tsx
PeerRich: very clever

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -45,9 +46,9 @@ const BookerComponent = ({ username, eventSlug, month, rescheduleBooking }: Book
     (state) => [state.selectedTimeslot, state.setSelectedTimeslot],
     shallow
   );
-  const extraDays = layout === "large_timeslots" ? (isTablet ? 2 : 4) : 0;
+  const extraDays = layout === "column_view" ? (isTablet ? 2 : 4) : 0;
```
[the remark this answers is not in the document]
PeerRich: good initiative!

packages/prisma/migrations/20230524105015_added_newbooker_feature_flag/migration.sql
sean-brydon: Yeah this seems to be correct.  

run 
`prisma migrate dev --create-only` 
add the migration in to the created file

packages/prisma/migrations/20230523102615_added_booker_layouts/migration.sql
```diff
@@ -0,0 +1,5 @@
...
+ALTER TABLE "EventType" ADD COLUMN     "bookerLayouts" JSONB;
+
+-- AlterTable
+ALTER TABLE "users" ADD COLUMN     "defaultBookerLayouts" JSONB;
```
sean-brydon, line 5: Should we use users `metadata` here instead of adding a new column to our 90k plus users?

packages/prisma/schema.prisma
```diff
@@ -225,6 +227,8 @@ model User {
   routingForms         App_RoutingForms_Form[] @relation("routing-form")
   verifiedNumbers      VerifiedNumber[]
   hosts                Host[]
+  /// @zod.custom(imports.bookerLayouts)
+  defaultBookerLayouts Json?
```
sean-brydon, line 231: Could we add this to users metadata?

apps/web/pages/settings/my-account/appearance.tsx
```diff
@@ -90,6 +97,9 @@ const AppearanceView = () => {
     <Form
       form={formMethods}
       handleSubmit={(values) => {
+        const layoutError = validateBookerLayouts(values.defaultBookerLayouts || null);
+        if (layoutError) throw new Error(t(layoutError));
```
sean-brydon, line 101: This wont hit the above onError in the toast right? 

This will actually hit our error boundary? Is this intended? 

-- 
Can we just do this validation on the backend and throw the error in the toast like we do above?

packages/lib/validateBookerLayouts.ts
sean-brydon: I mentioned this where it was used too - is there any reason we dont do this on the backend?

packages/prisma/migrations/20230523102615_added_booker_layouts/migration.sql
sean-brydon: I just mention this because we already have quiet a few bits of new settings for a specific user that lives in there. e.g. defaultConferncingApp lives in there feels like they are the same sort of thing

apps/web/pages/settings/my-account/appearance.tsx
JeroenReumkens: It get's caught in our default form component, which then shows that as a toast. See: 

CleanShot 2023-05-30 at 15 54 51@2x

I did however find an open issue now that seems to not reset the form `isSubmitting` state to false when an error occurred. And that boolean is used in some forms to put the submit button in a loading/disabled state. So I think it makes sense to remove the client side validation here, to prevent us needing to overhaul the whole default form logic.

packages/lib/validateBookerLayouts.ts
JeroenReumkens: We do it both actually. I think the faster we can give errors, the better?  That's at least why it's a shared utility, so we can easily run it in both places.

packages/prisma/migrations/20230523102615_added_booker_layouts/migration.sql
JeroenReumkens: Yes good question. I'm unsure what's the best approach. That's also why I added the question above. I'm okay with moving it. 

Let me know what you think @zomars or @emrysal

packages/prisma/schema.prisma
JeroenReumkens: Same question as for the events. Let's wait for Omar's / Alex' response, then I'll apply it do both.

packages/features/bookings/Booker/store.ts
```diff
@@ -104,11 +104,11 @@ const checkLayout = (layout: BookerLayout) => {
 export const useBookerStore = create<BookerStore>((set, get) => ({
   state: "loading",
   setState: (state: BookerState) => set({ state }),
-  layout: checkLayout(getQueryParam("layout") as BookerLayout) || "small_calendar",
+  layout: checkLayout(getQueryParam("layout") as BookerLayout) || "month_view",
   setLayout: (layout: BookerLayout) => {
     // If we switch to a large layout and don't have a date selected yet,
     // we selected it here, so week title is rendered properly.
-    if (["large_calendar", "large_timeslots"].includes(layout) && !get().selectedDate) {
+    if (["week_view", "column_view"].includes(layout) && !get().selectedDate) {
```
zomars, line 111: Can we make these string constants?

packages/features/flags/config.ts
```diff
@@ -10,5 +10,6 @@ export type AppFlags = {
   workflows: boolean;
   "v2-booking-page": boolean;
   "managed-event-types": boolean;
+  newbooker: boolean;
```
zomars, line 13: Isn't this the same as `v2-booking-page` ?

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -19,7 +19,6 @@ import {
   BooleanToggleGroupField,
   SelectField,
   InputField,
-  Input,
```
zomars, line 22: Intended?

packages/prisma/migrations/20230523102615_added_booker_layouts/migration.sql
zomars: Agreed with @sean-brydon let's use metadata for this.

packages/prisma/migrations/20230524105015_added_newbooker_feature_flag/migration.sql
zomars: It is correct. Prisma isn't involved here.

packages/prisma/schema.prisma
zomars: Replied

packages/prisma/zod-utils.ts
zomars: Ill Allow It Spanish GIF

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
```diff
@@ -110,6 +113,18 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
+  if (bookerLayoutsError) {
+    // Function returns a translation key, this maps it to an actual English text.
+    const errorText = {
+      bookerlayout_error_min_one_enabled: "At least one layout has to be enabled.",
+      bookerlayout_error_default_not_enabled:
+        "The layout you selected as the default view is not part of the enabled layouts.",
+      bookerlayout_error_unknown_layout: "The layout you selected is not a valid layout.",
+    };
+    throw new TRPCError({ code: "BAD_REQUEST", message: errorText[bookerLayoutsError] });
+  }
```
zomars, line 126:
```
import { getTranslation } from "@calcom/lib/server/i18n";
```


```suggestion
  if (bookerLayoutsError) {
    const t = await getTranslation("en", "common");
    throw new TRPCError({ code: "BAD_REQUEST", message: t(bookerLayoutsError) });
  }
```

Extra point for guessing the actual locale here. But English is good for now.

zomars (commented):
So far so good @JeroenReumkens left some comments to address. Also let's stick with user metadata to save the layout settings for now 🙏🏽

packages/features/flags/config.ts
JeroenReumkens: Oh yes it is, I totally overlooked it. 
Did a quick search now though, and it doesn't seem to be used at all. There's also no migration that adds it. So there only seems to be a type floating around somewhere that isn't used. So I'm gonna delete the v2 one.

packages/features/bookings/Booker/store.ts
JeroenReumkens: Converted it into an enum!

packages/prisma/migrations/20230524105015_added_newbooker_feature_flag/migration.sql
```diff
@@ -0,0 +1,9 @@
...
+  "Feature" (slug, enabled, description, "type")
+VALUES
+  (
+    'newbooker',
```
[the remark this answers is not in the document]
JeroenReumkens: No reason why not. Thanks for the suggestion, just changed it!

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
JeroenReumkens: Nice!! Soooo much better.

packages/features/form-builder/FormBuilder.tsx
JeroenReumkens: Yes, done by eslint. It was an unused import.

PeerRich (approved):
great work

PeerRich (approved):
great work

PeerRich:
@JeroenReumkens one merge conflict, afterwards ready to merge

packages/features/settings/BookerLayoutSelector.tsx
```diff
@@ -0,0 +1,194 @@
...
+  name,
+  fallbackToUserSettings,
+}: BookerLayoutSelectorProps) => {
+  const { control, getValues } = useFormContext();
+  const { t } = useLocale();
+  // Only fallback if event current does not have any settings, and the fallbackToUserSettings boolean is set.
+  const shouldShowUserSettings = (fallbackToUserSettings && !getValues(name || defaultFieldName)) || false;
+
+  const flags = useFlagMap();
+  if (flags["booker-layouts"] !== true) return null;
```
N-NeelPatel, line 42: @JeroenReumkens - when is the flags["booker-layouts"] set to true?

PR #9083, opened 2023-05-24
fix: text beyond X icon issue

AkhilG87:
Issue Fix - Text beyond the X icon
#8976 

Screenshot 2023-05-24 at 4 07 13 PM


kshitiz-goel07:
I think you @AkhilG87  should disable the cross icon when the user isn't typing anything
We can connect in order to solve this issue together :-)

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -194,6 +194,7 @@ export const FormBuilder = function FormBuilder({
               <li key={index}>
                 <div className="flex items-center">
                   <Input
+                    style={{ paddingRight: "2em" }}
```
PeerRich, line 197: please use tailwind and not inline styles

AkhilG87:
..

PR #9101, opened 2023-05-24
fix: check for premium username

CarinaWolli:
## What does this PR do?

Fixes check for premium username


keithwillcode (approved):
Ideally we have 1 function that does this logic instead of 4 places but maybe a refactor for later since we want this fixed now.

zomars (approved):

zomars (approved):

PR #9106, opened 2023-05-25
fix: update button fix

### LINKED ISSUE #9073: Eventhough, After updation of Appearance Theme. Update button is still enabled
### Issue Summary

Eventhough Appearance theme is update. The update button is still enabled and disabled when we try to select the previous theme.

https://github.com/calcom/cal.com/assets/123581387/555ea77b-2814-4e83-8318-00bb92d94aa0

### Steps to Reproduce

1. Went to Settings -> Appearance.
2. Update theme.
3. select the previous theme.

Once we updated, update button should be disable and only enables when changes occurs.

VinoV1999:
## What does this PR do?

Reset the default values once the updation done successfully. 

Fixes #9073

## Type of change

- Reset the defaultValues.
- Return values to be reset from updateProfileHandler.

## ScreehRecord :-

https://github.com/calcom/cal.com/assets/123581387/6f6fb028-d568-4b25-9237-af9eb5a220c6

Since all the values to be reset with the updated values and the input of updateProfileHandler is same. I have get the input as a response and pass it as an argument of reset. 

The reason for returning values from server is to prevent reset when server is down. Also it won't affect app Theme enhancement.


CarinaWolli (approved):
Looks good 👏🏻 Thank you for your contribution!

emrysal (approved):

emrysal:
@VinoV1999 Thanks for your contribution; apologies for the delay we had some issues with the pipeline 🙏

PR #9145, opened 2023-05-26
Tests: Automated TypeScript build check for a sample react App using @calcom/embed-react

### LINKED ISSUE #9091: [CAL-1757] [CI Automation]Make an independent react application that uses @calcom/embed-react and verify that it builds fine.
This is needed because a built package when published to npm might have certain files missing and types are imported from *.d.ts files instead of .ts files directly in that case.

<sub>[CAL-1757](https://linear.app/calcom/issue/CAL-1757/[ci-automation]make-an-independent-react-application-that-uses)</sub>

hariombalhara:
## What does this PR do?

Closes #9091 

It would allow us to solve these problems
- Easily update all code sandboxes.
- Codesandboxes would be reviewable now
- Verify that the TypeScript code in them builds correctly
- We would have list of all working embed examples in one place.

## Type of change

- Test (non-breaking change which fixes an issue)

## Observations
- If the motive is to find out if type definitions are correctly specified in package.json, then a single demo app that imports embed-react should be able to test it.
	- 	For testing the types of supported instructions(like ui, on, off), when we write unit tests in TypeScript, that would automatically be verified.
- If the motive is to just find out if the file specified at packageJson.types exist then we can do a check after build if it exists.
- Codesandbox doesn't allow a particular folder or a package in the monorepo to be linked to a Codesandbox.
	- Connecting the entire repo to codesandbox makes it slow and people don't quickly fork different variations of codesandboxes the way it's possible right now

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist


packages/embeds/codesandboxes/embed-react/CSS Variables API/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "CSS Variables API",
+  "packageManager": "yarn@3.4.1",
+  "scripts": {
+    "type-check": "tsc --pretty --noEmit"
+  },
+  "dependencies": {
+    "@calcom/embed-react": "workspace:^"
```
hariombalhara, line 8: We need workspace: here so that we can test with latest changes from @calcom/embed react that aren't yet published.

But for the synced codesandbox to work it has to be a an npm repository depedency, something like '1.0.1'

packages/embeds/embed-core/src/embed.ts
```diff
@@ -125,7 +125,7 @@ type SingleInstruction = SingleInstructionMap[keyof SingleInstructionMap];
 export type Instruction = SingleInstruction | SingleInstruction[];
 export type InstructionQueue = Instruction[];
 
-type PrefillAndIframeAttrsConfig = Record<string, string | string[] | Record<string, string>> & {
+export type PrefillAndIframeAttrsConfig = Record<string, string | string[] | Record<string, string>> & {
```
hariombalhara, line 128: Add missing types. These are found because the build for codesandbox failed.

PR #9165, opened 2023-05-28
fix: embed hides floating button

### LINKED ISSUE #8851: [CAL-1655] Embed on mobile button stays above booking modal
So I can’t actually click to confirm the booking. This floating button should disappear while modal is open. 

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1655](https://linear.app/calcom/issue/CAL-1655/embed-on-mobile-button-stays-above-booking-modal)</sub>

riteshsp2000:
## What does this PR do?

- Increased modalbox z-index to highest. It would stay on top of button because it is added after the button.
- Also, some code cleanup that was required for hiding floating button popup which isn't required now.

-- Edit by hariombalhara

Fixes #8851

https://github.com/calcom/cal.com/assets/56112399/02362901-8a5b-478b-9278-b08323c49549

## Type of change

## How should this be tested?

- ran the embed tests from the root directory - 4 errors, 5 skips, 1 successful (same as that of in production)
- in terms of manual testing, ensured that the functionality of the floating popup is alright by going through the entire flow (click -> change timezone -> search timezone -> select date -> change time format -> select time -> update details -> confirm)
- ensured that the button is hidden when the Modal is opened and is shown back again when the modal is closed (in any possible way: close button, click away, cancel button)

## Checklist


PeerRich:
can you uncommit yarn.lock please? 🙏

hariombalhara:
@riteshsp2000 Can you add proper tests that you did under 'How should this be tested?'

riteshsp2000:
> @riteshsp2000 Can you add proper tests that you did under 'How should this be tested?'

@hariombalhara haven't added the required unit tests. How about I write a unit test that checks if the button is hidden or not after the modal is opened and vice a versa

hariombalhara:
You can add a list of manual tests there instead of mentioning just Test A, Test B.
There are e2e tests for embed. But there are no tests for floating popup as of now. You can add that in a separate PR if you would like
https://github.com/calcom/cal.com/blob/main/packages/embeds/embed-core/playwright/tests/action-based.e2e.ts#L104

hariombalhara (changes requested):
There is a comment that has to be addressed first.

riteshsp2000:
> You can add a list of manual tests there instead of mentioning just Test A, Test B. There are e2e tests for embed. But there are no tests for floating popup as of now. You can add that in a separate PR if you would like https://github.com/calcom/cal.com/blob/main/packages/embeds/embed-core/playwright/tests/action-based.e2e.ts#L104

@hariombalhara I have been trying to run the embed tests specifically but got an error that `playwright.config.ts` is not found. I checked the test command and the config file mentioned here doesn't actually exist. What should I be doing now?

https://github.com/calcom/cal.com/blob/8ad513e4ece852128a0bff7a36eb862af495ca33/packages/embeds/embed-core/package.json#L29

hariombalhara:
@riteshsp2000 Ah, sorry for the trouble. There was a refactor some time back and the command isn't updated.
You need to run `yarn e2e:embed` from the root of the repo to run embed-core tests.

hariombalhara:
@riteshsp2000 Also, I am okay to merge without automation tests. They can be done in a different PR. 
I just need you to list down the manual testing that you have done in the PR.

riteshsp2000:
> @riteshsp2000 Also, I am okay to merge without automation tests. They can be done in a different PR. I just need you to list down the manual testing that you have done in the PR.

@hariombalhara Sorry for the delay, was traveling for the last few days. Have updated the PR description with all sorts of tests performed by me.

packages/embeds/embed-core/src/embed.ts
```diff
@@ -666,6 +666,19 @@ document.addEventListener("click", (e) => {
...
+
+  /**
+   * on click, checks if the target element is the floating button
+   * if so, hides the button to ensure that it does not cover
+   * any of the modal's content.
+   *
+   * bringing the element back - check out the close() method
+   * of ModalBox element
+   */
+  if (targetEl.dataset.calNamespace === "floatingButton") {
```
hariombalhara, line 678: `calNamespace` can be anything. It is just a name given when there are multiple embeds on the page. Here it's floatingButton because it's been given that name through init call. So, we shouldn't check for namespace.





Also, I don't think we need to check if the modal is opened through floating button popup or element click. When the modal is open we should hide the floating button element so that it would never come in the way of booking.

hariombalhara (changes requested):
We are almost there. Requesting one more change. I think we should just remove the calNamespace check so that floating button element gets hidden whenever modal opens(through floating button or element click)

After that you can test both the modals.

Added via Giphy

hariombalhara:
@riteshsp2000 I guess you are involved somewhere. I would take over this PR as it's tagged for this milestone.

packages/embeds/embed-core/src/ModalBox/ModalBox.ts
```diff
@@ -105,7 +113,7 @@ export class ModalBox extends HTMLElement {
     this.attachShadow({ mode: "open" });
     ModalBox.htmlOverflow = document.body.style.overflow;
     document.body.style.overflow = "hidden";
-
+    this.open();
```
hariombalhara, line 116: Let it fire open event.

hariombalhara (approved):

PR #9212, opened 2023-05-30
chore: Removed extra db-seed call since we use cache-db

keithwillcode:
## What does this PR do?

Removes a duplicate db-seed call (saves 40 seconds per E2E run)


zomars (approved):
Nice catch! @keithwillcode 🙏🏽

PR #9213, opened 2023-05-30
fix: deleteTokenIfUsed

### LINKED ISSUE #9208: [CAL-1823] Delete verificationTokens after they have been used
There is no point keeping them in the DB after they have been successfully used

lets remove them after we have verified that they have been used. This would also prevent them from having more than one use

Thanks @hariombalhara for the suggestion 

<sub>[CAL-1823](https://linear.app/calcom/issue/CAL-1823/delete-verificationtokens-after-they-have-been-used)</sub>

saurabbot:
fixes #9208


apps/web/pages/signup.tsx
```diff
@@ -181,18 +181,27 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
...
     return {
       notFound: true,
     };
   }
 
+  await prisma.verificationToken.deleteMany({
+    where: {
+      token,
+    },
+  });
```
sean-brydon, line 203: I think this can be just delete 

token is unique in our DB so this should never be many

sean-brydon (commented):
Great Job - Sorry for the late reply - I will make the changes that i mention so you dont have to come back to this PR <3 

thank you

sean-brydon (changes requested):
After some testing - we can't actually delete the token here as this is where it is read. 

we need this token on the backend to preform the signup. 

If you actually try and sign in with a team invite - the token never exists cause its always deleted.

keithwillcode:
This PR is being closed due to inactivity. Please reopen if work is intended to be continued.

PR #9220, opened 2023-05-31
fix: Remove cache-db from without database build

hariombalhara:
Remove cache-db from without database build which was causing the step to fail.


hariombalhara:
Because there is a change in workflow file, the result isn't here for that. Here is the passing workflow results for this branch.
https://github.com/calcom/cal.com/actions/runs/5130469314/jobs/9229308042

PeerRich:
@baileypumfleet @ConnorGargano 
force merging because CI doesnt pass for this (its fixing it) and its not touching core code base

zomars:
nice catch @hariombalhara thanks again!

PR #9226, opened 2023-05-31
fix: layout effect hook warning

G3root:
## What does this PR do?

fixes this warning

`
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
`

## Type of change


hariombalhara (approved):
LGTM !!

PR #9023, opened 2023-05-22
feat: new booker embed

JeroenReumkens:
## What does this PR do?

Make booker work with embed:
* Enables routes when cookie is set
* Gives an option in the embed configurator to select the layout you want to see
* No layout toggle in embed possible, we show the layout the user selects in the settings. 

Fixes CAL-1716

**Environment**: Staging(main branch)

## Type of change

- New feature (non-breaking change which adds functionality)

## How should this be tested?

- [ ] View the embed playground in /packages/embed/embed-core by running `yarn dev`. 
- [ ] Test the embed settings on the event types (`<>` icon in top right when editing an event) to test new layout toggle

PR #9227, opened 2023-05-31
feat: CAL-1674 Added new header layout for new booker.

JeroenReumkens:
## What does this PR do?

Added new header design for fullscreen layout in booker. 
This header is not shown in the month view. In that case you only see the layout toggle. 

PS: There's currently an issue that you can't scroll in the fullscreen layouts – this is fixed in #9023 

Fixes CAL-1674

CleanShot 2023-05-31 at 11 31 19@2x

## Type of change

## How should this be tested?
- [ ] Test the header in the fullscreen layouts
- [ ] Ensure it's not broken in month view or mobile view


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -98,35 +95,6 @@ const BookerComponent = ({
 
   return (
     <>
-      {/*
```
JeroenReumkens, line 101: Moved into the header component.

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -0,0 +1,111 @@
...
+          />
+        </ButtonGroup>
+      </div>
+      <div className="ml-auto flex gap-3">
+        <TimeFormatToggle />
+        <div className="fixed top-4 right-4">
+          <LayoutToggle onLayoutToggle={onLayoutToggle} layout={layout} />
+        </div>
+        {/*
+          This second layout toggle is hidden, but needed to reserve the correct spot in the DIV
```
JeroenReumkens, line 63: Couldn't come up with a better solution. Don't think this is too bad, but also not 100% happy with it. For now it works well I think, maybe later we can see if we can come up with a better solution.

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -0,0 +1,111 @@
+import { ChevronLeft, ChevronRight } from "lucide-react";
```
JeroenReumkens, line 1: Whole file looks new, but main part is actually a rename of the `LargeViewHeader` component below. Only added the layout toggle and did a few tweaks.

packages/features/bookings/Booker/components/Header.tsx
sean-brydon: I was actually curious how this was gonna be tackled ... easier than i had it in my head :P

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -0,0 +1,111 @@
...
+  onLayoutToggle,
+  layout,
+}: {
+  onLayoutToggle: (layout: string) => void;
+  layout: string;
+}) => {
+  const { t } = useLocale();
+
+  return (
+    <div className="[&>div]:bg-muted z-10">
```
sean-brydon, line 87:
```suggestion
    <div className="[&>div]:bg-muted z-10">
```

We do this a lot whenever we seem to use ToggleGroup - is it worth just creating a variant that changes the background of the div?

packages/features/bookings/Booker/config.ts
```diff
@@ -50,42 +50,60 @@ export const resizeAnimationConfig: ResizeAnimationConfig = {
       minHeight: "0px",
       gridTemplateAreas: `
           "meta"
+          "header"
           "main"
           "timeslots"
         `,
```
sean-brydon, line 56: So header now only displays the 24h selector on mobile? and not the layout view?

packages/features/bookings/Booker/components/Header.tsx
JeroenReumkens: Hahaha. I tried a lot with framer motion too, but you always saw the jump. So instead of spending days on an animation I think this for now was the best solution, haha.

packages/features/bookings/Booker/components/Header.tsx
JeroenReumkens: I tested this, and actually it isn't needed anymore. `bg-muted` is the default, so the classname didn't do anything anymore here. Removed it.

PeerRich:
NIT:
- the 12/24h switch is differently positioned than the calendar switch
- the next/previous calendar switch is shifting layout (likely requires a min-width)

https://github.com/calcom/cal.com/assets/8019099/fd1eed20-f856-47ba-adf3-111ce06c50b9

PeerRich:
another NIT:


https://github.com/calcom/cal.com/assets/8019099/1205433b-3d9a-44b4-9fac-1c7805aa6648

PeerRich:
NIT: missing padding bottom

CleanShot 2023-06-01 at 12 10 52@2x

PeerRich:
this calendar should be bottom aligned

CleanShot 2023-06-01 at 12 11 56@2x

JeroenReumkens:
@PeerRich  All issues have been fixed. Could you please retest it? 
Only one issue left where the resize animation doesn't work properly. But part of the solution is on the embed PR. So I will take another look at the animation once all branches have been merged. Created a ticket in Linear for it.

sean-brydon (approved):
CleanShot 2023-06-06 at 11 38 53@2x
CleanShot 2023-06-06 at 11 38 35@2x

Found two weird displays at some random breakpoints that would only happen if youre changing the scale by hand. 

Other than that all issues are fixed and LGTM. I'd be happy with these being fixed in a follow up due to the fact they're a bit out of scope of this header PR.

Nice work as always @JeroenReumkens animations are looking 👨‍🍳

JeroenReumkens:
> Nice work as always @JeroenReumkens animations are looking 👨‍🍳

Thanks a lot for the review @sean-brydon. I'm indeed going ahead and merge this, because I can use these changes in the embed PR too. Will make a follow up PR in the end to fix the last things for the layout and animations.

PR #9237, opened 2023-05-31
fix: removed additional notes in email even if its empty

### LINKED ISSUE #9224: [CAL-1830] Why not remove Additional Notes from email? if it's empty
See the following screenshot.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1830](https://linear.app/calcom/issue/CAL-1830/why-not-remove-additional-notes-from-email-if-its-empty)</sub>

sohamdata:
## What does this PR do?
Fixes #9224 

If **additional notes** is empty, it is not included in the email.

## Screenshots:
With additional notes:
yesNotes
Without additional notes:
noNotes

## Type of change
- Bug fix


packages/lib/CalEventParser.ts
```diff
@@ -205,6 +204,12 @@ ${calEvent.paymentInfo.link}
     : ""
 }
   `.trim();
+
+  const additionalNotes = getAdditionalNotes(calEvent, t);
```
hariombalhara, line 208: We shouldn't change the order of the sections.

hariombalhara (dismissed):
Thanks @sohamdata. Left a comment and will wait for @PeerRich's confirmation as well.

packages/lib/CalEventParser.ts
sohamdata: Got it, I think we can convert the return value for getAdditionalNotes to a boolean and render it conditionally in that order.

sohamdata: fixed in commit #[1f6b642](https://github.com/calcom/cal.com/pull/9237/commits/1f6b642afa6a5802dde8555711c6fae28255d9e5)

sohamdata: @hariombalhara fixed in commit #[1f6b642] (https://github.com/calcom/cal.com/pull/9237/commits/1f6b642afa6a5802dde8555711c6fae28255d9e5)

packages/lib/CalEventParser.ts
```diff
@@ -188,7 +191,7 @@ ${getWho(calEvent, t)}
 ${t("where")}:
 ${getLocation(calEvent)}
 ${getDescription(calEvent, t)}
-${getAdditionalNotes(calEvent, t)}
+${hasAdditionalNotes ? additionalNotes : ""}
```
hariombalhara, line 194: I just noticed that, there is already a [handling for empty string inside getAdditionalNotes](https://github.com/calcom/cal.com/blob/c41583175e1c8ac1ad6ff5ff702084402cc85eed/packages/lib/CalEventParser.ts#L68)

@sohamdata Can you confirm that you are able to replicate the issue and this change fixes it?

hariombalhara: I saw the issue, so it's also possible that I entered a space there accidentally and it happens only in that case. We can do a trim if that's the case.

hariombalhara (changes requested):
Left a comment.

packages/lib/CalEventParser.ts
sohamdata: @hariombalhara you are right, it happens when space is entered. However, when I tried checking for `calEvent.additionalNotes.trim()` in the getAdditionalNotes function, the additional notes still show up for some reason, no matter the number of spaces. Can you confirm the same?

sohamdata: @hariombalhara
for debugging I changed to (sorry for the gibberish, it had catch my eye in the terminal)
```
export const getAdditionalNotes = (calEvent: CalendarEvent, t: TFunction) => {
  console.log("HRERHERHERHEREHREHREHREHRERHERH", calEvent.additionalNotes, "HRERERERER");
  calEvent.additionalNotes === calEvent.additionalNotes?.trim();
  console.log("2nd HRERHERHERHEREHREHREHREHRERHERH", calEvent.additionalNotes, "2nd HRERERERER");

  if (!calEvent.additionalNotes) {
    return "";
  }
  return `
${t("additional_notes")}:
${calEvent.additionalNotes}
  `;
};
```

terminal:
image
image

It's not trimming for some reason

Jaibles:
@sohamdata @hariombalhara anything we can do to get this merged?

PR #9250, opened 2023-05-31
fix: API: Only include accepted users, ignore all pending, unregistered or not

### LINKED ISSUE #9246: [CAL-1835] API: If the team contains a pending invite of an unregistered user, the team availability query fails
The team availability query currently fails if there is an unregistered user with a pending invite within the team. 

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1835](https://linear.app/calcom/issue/CAL-1835/api-if-the-team-contains-a-pending-invite-of-an-unregistered-user-the)</sub>

emrysal:
## What does this PR do?

An unregistered user does not have availability yet which causes the crash, but also pending users shouldn't be returned yet as part of the availability call - as they have not yet accepted the invite.

Fixes #9246


alishaz-polymath (approved):
🚀

zomars (approved):
LGTM

PR #9253, opened 2023-05-31
feat: organization event type filter

Udit-takkar:
## What does this PR do?

Fixes https://linear.app/calcom/issue/CAL-1839/org-event-types-filter-ui

https://github.com/calcom/cal.com/assets/53316345/af6e7dff-9e46-4a31-b36b-8ac9bfbd374a

## Type of change


sean-brydon (approved):
CleanShot 2023-06-05 at 10 11 50@2x
LGTM - tested and it seems to work.

Not sure what we would do when we select two. The label seems to be a bit weird maybe we can add a max-width and concat them? @Jaibles

PR #9257, opened 2023-06-01
fix: Made dynamic text for booking with only booker or multiple attendees

### LINKED ISSUE #9239: [CAL-1832] Make 'We emailed you and the other attendees a calendar invitation with all the details.' text dynamic
1. If there is just booker only and no guests.
   1. "We emailed you a calendar invitation with all the details." →
2. If there is just booker and atleast one guest:
   1. "We emailed you and the other attendees a calendar invitation with all the details."

We should see all the occurrences of such texts and fix accordingly.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1832](https://linear.app/calcom/issue/CAL-1832/make-we-emailed-you-and-the-other-attendees-a-calendar-invitation-with)</sub>

Malav20:
## What does this PR do?
Added logic for checking attendees in the booking and showing message based on that

Fixes #9239 

## Type of Change
Bug Fix (non-breaking change which fixes an issue)

## How should this be tested?

- Go to booking
- Create a booking with only 1 member 
- Now check the message
- Now create another booking with multiple members
- Now check the message


apps/web/pages/booking/[uid].tsx
```diff
@@ -268,7 +268,11 @@ export default function Success(props: SuccessProps) {
       }
       return t("needs_to_be_confirmed_or_rejected" + titleSuffix);
     }
-    return t("emailed_you_and_attendees" + titleSuffix);
+    if (bookingInfo.attendees.length > 1) {
+      return t("emailed_you_and_attendees" + titleSuffix);
+    } else {
+      return t("email_all_details" + titleSuffix);
```
Comment (hariombalhara) [inline apps/web/pages/booking/[uid].tsx:274]: I don't see `email_all_details_recurring` to be available in common.json.
Also better key name might be `email_you`.

apps/web/public/static/locales/en/common.json
```diff
@@ -264,6 +264,7 @@
   "add_another_calendar": "Add another calendar",
   "other": "Other",
   "email_sign_in_subject": "Your sign-in link for {{appName}}",
+  "email_all_details": "We emailed you with all the details",
   "emailed_you_and_attendees": "We emailed you and the other attendees a calendar invitation with all the details.",
   "emailed_you_and_attendees_recurring": "We emailed you and the other attendees a calendar invitation for the first of these recurring events.",
   "emailed_you_and_any_other_attendees": "You and any other attendees have been emailed with this information.",
```
hariombalhara, line 270: It also has attendees in it. So, wherever this key is used we need to fix there as well.

hariombalhara (changes requested):
Thanks @Malav20 🙏 Left some comments.

apps/web/pages/booking/[uid].tsx
Comment (Malav20) [reply apps/web/pages/booking/[uid].tsx:274]: For recurring bookings also we need dynamic message? same as single booking?

apps/web/public/static/locales/en/common.json
Malav20: Did not get this. Can you please explain this?

hariombalhara: I mean the key 'emailed_you_and_any_other_attendees' also needs to be changed in a similar way.

apps/web/pages/booking/[uid].tsx
Comment (hariombalhara) [reply apps/web/pages/booking/[uid].tsx]: Yeah, the fix is supposed to handle all cases where we mention "other attendees" when infact there are no other attendees. It can be 
- For Recurring Booking
- For One time Booking
- In this file
- Any other file
- Other keys that I have not mentioned.

Comment (Malav20) [reply apps/web/pages/booking/[uid].tsx]: @hariombalhara I guess I have made the required changes. I also checked that in my local. Can you please go through the changes?

PR #8705, opened 2023-05-05
chore: Rounding, frames and radio buttons of the booking forms not consistent

gitstart-calcom:
Fixes #7508 

## Demo
https://www.loom.com/share/f020b84243ca48928f6d44d8b9df6f92

Steps to reproduce: 
1. Edit some event-type 
2. Click on advanced settings and add all the options
3. Go to settings/appearance and change the brand color
4. Try to schedule some meeting

FIgma with the updated design: https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/%E2%9D%96-Cal-DS?node-id=309-44674&t=XVOdTk8kjKnXp6ti-0

PR #9258, opened 2023-06-01
fix: [Hotfix]No phone number is valid

hariombalhara:
We recently moved to react-phone-input-2 and after that all phone numbers are considered invalid by `isValidPhoneNumber`. 
 https://github.com/calcom/cal.com/pull/8705/files
 
 Tests:
- Verified workflow SMS sending to specific number.
- Booking Form Phone Number Submission

 Demo https://www.loom.com/share/2593647cf1ea442e8ba3b7873c5fa127

### DIFF
--- a/packages/ui/form/PhoneInput.tsx
+++ b/packages/ui/form/PhoneInput.tsx
@@ -65,7 +65,7 @@ const useDefaultCountry = () => {
     retry: false,
     onSuccess: (data) => {
       if (isSupportedCountry(data?.countryCode)) {
-        setDefaultCountry(data.countryCode);
+        setDefaultCountry(data.countryCode.toLowerCase());
       }
     },
   });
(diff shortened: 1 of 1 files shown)


packages/ui/form/PhoneInput.tsx
```diff
@@ -30,7 +30,9 @@ function BasePhoneInput({ name, className = "", onChange, ...rest }: PhoneInputP
         required: rest.required,
         placeholder: rest.placeholder,
       }}
-      onChange={(value) => onChange(value)}
+      onChange={(value) => {
+        onChange("+" + value);
```
hariombalhara, line 34: I noticed that new react-phone-input-2 doesn't have + in the value it gives which is required by isValidPhoneNumber

Udit-takkar (dismissed):
1) I can still see the validation error.  

Screenshot 2023-06-01 at 1 31 47 PM

2) we were able to auto detect country from IP address which is not working now.

Udit-takkar:
Follow up:  we should remove react-phone-number-input if we are not using it anywhere. It has a very big bundle size

hariombalhara:
That phone number was invalid earlier as well.I think it's not a valid number

Udit-takkar (approved):
LGTM

PR #9259, opened 2023-06-01
fix: Improve new booker background colors in dark mode

JeroenReumkens:
## What does this PR do?

Fix for booker colors in dark mode, we can't use bg-default in both places because we use a custom background in dark mode. Otherwise it would be a similar color in dark. So we need to supply a custom dark mode color.

### Before

image

### After

CleanShot 2023-06-01 at 08 50 07@2x

## Type of change

## How should this be tested?

- [ ] Test booker in both light and dark mode, and all layouts and see if the background color looks correct in all variants.


PeerRich (approved):
works great!

sean-brydon (approved):

PR #9260, opened 2023-06-01
fix: make new booker work with private links

JeroenReumkens:
## What does this PR do?

Make private links work for new booker. 

Fixes CAL-1698

## Type of change

## How should this be tested?

- [ ] Enabled the new booker, set a private link for an event, and verify that the private link now shows the new booker, and that you can successfully book the event.


zomars (approved):
Nice catch @JeroenReumkens tested and works like a charm 🙏🏽 

Ship it

PR #9266, opened 2023-06-01
feat: organization settings general and members page

Udit-takkar:
## What does this PR do?

Fixes https://linear.app/calcom/issue/CAL-1847/org-general-and-members-page

Screenshot 2023-06-05 at 1 12 29 PM

Screenshot 2023-06-05 at 1 12 59 PM

## Type of change


packages/trpc/server/routers/viewer/teams/inviteMember.handler.ts
```diff
@@ -41,7 +41,7 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
     ? input.usernameOrEmail
     : [input.usernameOrEmail];
 
-  emailsToInvite.forEach(async (usernameOrEmail) => {
```
Udit-takkar, line 44: we cannot throw error inside forEach so replaced it with for of loop

PeerRich:
weird error on vercel build
image


can you run yarn build locally?

Udit-takkar:
Screenshot 2023-06-05 at 1 32 39 PM

@PeerRich  getting this error in new-booker page

packages/features/ee/organizations/pages/settings/members.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const router = useRouter();
+  const session = useSession();
+  const utils = trpc.useContext();
+  const showDialog = router.query.inviteModal === "true";
+  const [showMemberInvitationModal, setShowMemberInvitationModal] = useState(showDialog);
+  const { data: team, isLoading } = trpc.viewer.organizations.listMembers.useQuery(undefined, {
+    onError: () => {
+      router.push("/settings");
+    },
+  });
```
sean-brydon, line 73: On delete/edit  in MemberListItem we need to invalidate this item so we dont have to manually refresh to see changes

sean-brydon (changes requested):
Left a NIT.

We also need to remember on removal of a member to remove the membership connection between all users-> teams that have a parent ID of this ORG

Nice work! Very close to being mergeable

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -86,6 +86,7 @@ export const createHandler = async ({ input }: CreateOptions) => {
 
     totp.options = { step: 90 };
     const code = totp.generate(secret);
+    console.log("OTP", code);
```
sean-brydon, line 89:
```suggestion
```

packages/features/ee/organizations/pages/settings/general.tsx
```diff
@@ -0,0 +1,195 @@
...
+    onError: () => {
+      router.push("/settings");
+    },
+  });
+  const { data: user } = trpc.viewer.me.useQuery();
+
+  if (isLoading) return <SkeletonLoader title={t("general")} description={t("general_description")} />;
+  if (!currentOrg) {
+    throw new Error(t("something_went_wrong"));
+  }
```
sean-brydon, line 59: Should we really throw an Error here IIRC this would cause a 500 on the client - Should we not redirect (as they shouldnt be in an org) ? We can do a check in the `onSettled` and redirect back to settings?

packages/features/ee/teams/components/MemberListItem.tsx
```diff
@@ -44,6 +44,14 @@ const useCurrentUserId = () => {
   return user?.id;
 };
 
+const checkIsOrg = (team: Props["team"]) => {
+  return typeof team?.metadata === "object" &&
+    !Array.isArray(team?.metadata) &&
+    team?.metadata?.isOrganization
+    ? true
+    : false;
+};
```
sean-brydon, line 53: Lets use zod for this check - the teamMetadataSchema should have this in

sean-brydon (changes requested):
Few things we should address/discuss before we merge! Nice job. Everything seems to work as expected

packages/features/ee/organizations/pages/settings/general.tsx
Udit-takkar: I have already added    
```
onError: () => {
      router.push("/settings");
}, 
```
in trpc.viewer.organizations.listCurrent

sean-brydon:
```
  if (!currentOrg) {
    throw new Error(t("something_went_wrong"));
  }
  ```
  Wont count as a trpc error tho right 🤔

Comment (Udit-takkar) [reply packages/features/ee/organizations/pages/settings/general.tsx:59]: yeah this is just for type checking ensuring that currentOrg exist and then passing to <GeneralView />. If there is an error while fetching or organisation doesn't exist we would be automatically redirected to /settings because of onError.

Comment (Udit-takkar) [reply packages/features/ee/organizations/pages/settings/general.tsx:59]: we would have to do this a lot of currentOrg && currentOrg.timeZone.  if the condition ( if(!currentOrg) ) is not added

Comment (sean-brydon) [reply packages/features/ee/organizations/pages/settings/general.tsx:59]: Not sure i like the idea of throwing an error - maybe we just return an empty component incase something goes wrong? Also let me know if you dont like this idea :D 

 if(!currentOrg) {
return null
}

File: packages/trpc/server/routers/viewer/teams/removeMember.handler.ts [hunk 6]
```diff
@@ -40,6 +41,23 @@ export const removeMemberHandler = async ({ ctx, input }: RemoveMemberOptions) =
...
+        userId: membership.userId,
+      },
+    });
+
+    await prisma.user.update({
+      where: { id: membership.userId },
+      data: { organizationId: null },
+    });
+  }
+
```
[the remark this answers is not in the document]
Udit-takkar: anything to change here ?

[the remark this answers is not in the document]
sean-brydon: No i was just saying well done :) *chefs kiss*

packages/trpc/server/routers/viewer/teams/inviteMember.handler.ts
Udit-takkar: sean already updated the invitmember.handler in his Billing PR so we can use that. not making any change here.

sean-brydon (approved):
Thanks for fixing the issues! Happy with this

PR #9268, opened 2023-06-01
chore: prevent merge conflicts in common.json

PeerRich:
by adding your new strings in the second last row, you'll prevent merge conflicts because of the missing trailing "," in the end


emrysal (approved):
Yes please, it's the , that causes it and just doing this will fix it 🥳

PR #9296, opened 2023-06-02
feat: API: /bookings optional end property - default to event length

emrysal:
## What does this PR do?

Allows not having to manually compute the end of the booking, instead use eventType settings.


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -468,6 +468,13 @@ function getBookingData({
           if (val.responses) {
             const unwantedProps: string[] = [];
             legacyProps.forEach((legacyProp) => {
+              if (typeof val[legacyProp as keyof typeof val] !== "undefined") {
```
emrysal, line 471: Unrelated, gives a warning when passing `location` or `notes` as a falsy value (but still present). Just log as this would be a breaking change.

keithwillcode (approved):
Awesome work.

zomars (approved):
Ship it

PR #9303, opened 2023-06-03
fix: New Booking on Group Event shows other guest name on confirmation

### LINKED ISSUE #9292: New Booking on Group Event shows other guest name on confirmation
### Issue Summary

As a Guest, When making a booking on a Group Event that already has some seats taken;
The confirmation dialogue shows another guests name rather than your own
image

### Steps to Reproduce

1. Create group event with x seats
2. as a guest, book one seat
3. as a different guest, book another seat
4. Check confirmation dialogue

- p.s. when booking as a guest, I am a logged in cal user (auto fill name and email)

rvikunwar:
## What does this PR do?
I have changed the title structure of booking for group event with multiple seats. Now for group event it be like [ event name ] by [ host name ] in place of [ event name ] between [ host name ] and [ attendee ]. This will fix the issue of showing other attendee name while booking a group event by other guest user, earlier the title was created based upon the name of first attendee now it will be a fixed title only for group event else will be same as earlier. 

Fixes #9292 

Before - 
before

After -
after

## Type of change

## How should this be tested?
Create group event booking and book the event as guest, you'll see change in title (what) of the scheduled meeting


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -911,9 +911,16 @@ async function handler(
   const calEventUserFieldsResponses =
     "calEventUserFieldsResponses" in reqBody ? reqBody.calEventUserFieldsResponses : null;
 
+  const createStaticTitle = () => {
+    return eventNameObject.eventType + ' by ' + eventNameObject.host
```
alannnc, line 915: I think the new logic should be inside of our core logic. I think for seated events the second attendee gets booking title from DB, when we should be reading event-type.title and parsing with getEventName()

alannnc (changes requested):
New logic should be inside core getEventName() and seated events should use eventType.title rather than booking.title.

PR #9322, opened 2023-06-04
fix: Availability Name does Not Update

### LINKED ISSUE #9315: [CAL-1863] Availability Name does Not Update
The schedule does pull properly through to the proper event-types, it just doesn't update under the "Availability" tag 

Similar how if the event type is changed from round robin to collective, the wording under "Assignment" changes to Collective

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1863](https://linear.app/calcom/issue/CAL-1863/availability-name-does-not-update)</sub>

rvikunwar:
## What does this PR do?
Earlier, the tag under event availability was only changing for the Managed events or any event with locked features in it but not for Collective events and round robin event.
Now, I have updated code which updates event availabilty tag for other event types.  

Fixes #9315 

Before
before

After
after

## Type of change

## How should this be tested?
1. Navigate to the collective even
2. Click on "Availability"
3. Toggle on "common schedule"
4. Pick a new schedule that isn't the default
5. Now, the schedule under "Availability" on the left would update, earlier it was not.

### DIFF
--- a/apps/web/components/eventtype/EventTypeSingleLayout.tsx
+++ b/apps/web/components/eventtype/EventTypeSingleLayout.tsx
@@ -193,7 +193,7 @@ function EventTypeSingleLayout({
                   : `default_schedule_name`
               }`
             : eventType.scheduleName ?? `default_schedule_name`
-          : `default_schedule_name`,
+          : eventType.scheduleName ?? `default_schedule_name`,
     });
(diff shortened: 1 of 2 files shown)


Udit-takkar (approved):
@rvikunwar LGTM. I have pushed some changes

PR #9325, opened 2023-06-04
feature: redirect to anchor page

### LINKED ISSUE #4687: Redirection of users landing on your 404

wesleymatosdev:
- refactor: redirect to user's page
- feat: redirect to team's slug

## What does this PR do?

/claim #4687

Fixes #4687

### Not found team page before implementation
https://github.com/calcom/cal.com/assets/100464352/93828937-9d99-470b-9af2-d84f92d1ec3f
### Not found team page after implementation
https://github.com/calcom/cal.com/assets/100464352/49672e86-f002-432e-86e4-0e63d6eb518a

I did the same with user event pages 😄 

## Type of change

## How should this be tested?

- [] When hitting an undefined /team/[slug]/[event-type], should be redirected to /team/[slug] instead of not found page
- [] When hitting an undefined /[user]/[event-type], should be redirected to /[user] instead of not found page

## Checklist

### DIFF
--- a/apps/web/pages/[user]/[type].tsx
+++ b/apps/web/pages/[user]/[type].tsx
@@ -125,7 +125,18 @@ async function getUserPageProps(context: GetStaticPropsContext) {
       },
     },
   });
-  if (!user || !user.eventTypes.length) return { notFound: true };
+
+  if (!user) return { notFound: true };
+
+  // If there aren't any event types, redirect to the user's profile page
+  if (!user.eventTypes.length) {
+    return {
+      redirect: {
+        destination: `/${user.username}`,
+        permanent: false,
+      },
+    };
+  }
 
   const [eventType]: ((typeof user.eventTypes)[number] & {
     users: Pick<User, "name" | "username" | "hideBranding" | "timeZone">[];
@@ -143,7 +154,15 @@ async function getUserPageProps(context: GetStaticPropsContext) {
     },
   ];
 
-  if (!eventType) return { notFound: true };
+  // If the event type was not found, redirect to the user's profile page
+  if (!eventType) {
+    return {
+      redirect: {
+        destination: `/${user.username}`,
+        permanent: false,
+      },
+    };
+  }
 
   //TODO: Use zodSchema to verify it instead of using Type Assertion
   const locations = eventType.locations ? (eventType.locations as LocationObject[]) : [];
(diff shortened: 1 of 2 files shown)


PeerRich:
now looking at this, we are having a conversation internally whether we actually should redirect for 404 pages. there are some pro's and cons for doing this:

https://www.quora.com/Is-404-page-redirection-to-home-page-bad-for-SEO

muescha:
My guess:

Since it redirects a custom user URL, it is likely not relevant for SEO. The crawling agent typically looks for each link starting from the homepage, but the custom user URL is not visible from the homepage.

p6l-richard:
First off, this is indeed an SEO relevant change.
- We have a subset of user pages in our sitemap we submit to Google,
- Google indexes a couple of user pages outside of this as well by crawling the web (not only our homepage)

Now, about the redirect, the preferred method of handling this is to show a better 404 page instead of a redirect. Here's what Google's John Muller said on the topic:
> [A redirect] is not critically broken/bad, but additional complexity for no good reason - make a better 404 page instead.
[John Muller](https://twitter.com/JohnMu/status/1082565162597605377?s=20)

While we redirect to the next-best category here instead of the homepage, Google will likely still treat these redirects as so called "soft 404s" and the redirect therefore won't help our SEO.

The reason why the redirect is a confusing UX is that the user expects to navigate to the event page and instead gets redirected to another page without any notice -- the user never learns that the orignal event page is missing/outdated/deleted.

It's therefore a good practice to show a 404 and ideally even customise it. In this case, the customisation could be to show the actually visible events for the profile on the 404 page directly and/or add a link to the profile page for the user to navigate.

ologbonowiwi:
@PeerRich @p6l-richard @muescha we have a design of the new page that needs to be built then?

PeerRich:
ok i think our current solution is still better than redirecting or making the 404 more custom. closing for now. thank you @ologbonowiwi for your work, I would like to tip you for the work!

PR #9327, opened 2023-06-05
chore: stricter self-review guideline

hariombalhara:
Adds stricter self-review guideline


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,8 +8,6 @@ Fixes # (issue)
  Loom Video: https://www.loom.com/
 -->
 
-**Environment**: Staging(main branch) / Production
```
hariombalhara, line 11: Discussed earlier in group that this isn't needed as we do patch releases every other day and main is almost the production.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -27,13 +25,14 @@ Fixes # (issue)
 - [ ] Test A
 - [ ] Test B
 
+## Mandatory Tasks
+- [ ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.
```
hariombalhara, line 29: Moves the self-review guideline to a new place so that it's noticed immediately. Also, added under a stricter headline.

emrysal (approved):

pumfleet (approved):

muescha:
Are there any visible tasks that need to be completed by the PR author?
How can you determine if a PR has not undergone self-review?
Maybe we provide a link that offers a detailed explanation of what it means for a PR to be "self-reviewed"?

hariombalhara:
Yeah that makes total sense. This is my understanding of the task.
- Go through each of the file keeping in mind that the reviewer hasn't worked together with you on that PR but he would have some context of what's going on.
- If it's something that would be useful say 3-4 months down the line, then it should probably be a comment in the code itself(and not the PR). It might be a README change as well.
- Add comments at appropriate place in the code
- Once done, go to Review and write a comment like 'Self review complete.'

PR #9334, opened 2023-06-05
chore: switch over to use variable for app/companyName

hybes:
* enable_apps_description - appName
* recording_from_your_recent_call - appName
* the_calcom_team - companyName
* calcom_is_better_with_team - appName

## What does this PR do?

Uses variables in a few more of the i18n files, to show the App/Company name in more places on self-hosted.

## Type of change

## How should this be tested?

See the new variables in place, when an APP_NAME or COMPANY_NAME is set, otherwise it still falls back to "Cal.com"

## Checklist


PeerRich:
@zomars what happens if people change the language files? will crowdin understand this and pick this up or will they overwrite it?

PeerRich (approved):

p6l-richard:
Did we run a CLI command to upload the translations?

I don't think the GitHub action is configured to upload the translations before downloading from crowdin. We could therefore lose the changes if we rely only on the GitHub action.

Basically, we can either:
- run `crowdin upload translations` every time we edit non-en .json files, or
- update the GitHub action to set `upload_translations: true`:
```yml
name: Crowdin Action
...
jobs:
  synchronize-with-crowdin:
    ...
    steps:
      ...
      - name: crowdin action
        uses: crowdin/github-action@1.5.1
        with:
          upload_sources: true
          upload_translations: true # <- THIS HERE
          ...
        env:
          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
          CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
          CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
```

---
☝️ Note
When dealing with objects or primitives in the .json files, this won't work as crowding automatically transforms them. We look good here, as I couldn't find any references of these in the common.json files on q a quick search (e.g. ": [ ) but should be checked locally first.

PR #9337, opened 2023-06-05
feat: Enable Apps for Teams & Orgs [CAL-1782]

joeauyeung:
## What does this PR do?

This PR does the following:
- Allows users to install apps for either themselves or teams if they have the permission
- Show personal and team apps on the installed page
- Enable apps on the event type apps page for teams
- Allow apps for managed event types

Followup:
- Video Apps support for teams.

https://www.loom.com/share/ab4033942aac43a084cd077551b1b3c9

Fixes # (issue)

## Type of change

## How should this be tested?

- [x] Create a team
- [x] On the app store page, install an app
    - [x] A drop down should appear to install for either the user or a team
    - [x] Install the app for both instances
- [x] Go to a specific app page
    - [x] When install the app the dropdown should appear for the user or a team
    - [x] This time install the app for just a team
- [x] On the installed app pages
    - [x] Installed apps for the user and teams should appear
- [x] Go to a team event type and navigate to the apps page
    - [x] It should show credentials for the user and the team
    - [x] When enabling one it should save and not enable both instances of the app

## Mandatory Tasks

## Checklist

#### Blocking
- [x] Broken UI for Remove App Dialog. It is broken in production as well. [So, created a separate issue.](https://github.com/calcom/cal.com/issues/9846)
- [x] Can't install an App for a team(after i have installed it for user)
- [x] Revert [this](https://github.com/calcom/cal.com/pull/9337/files#r1248819904), [this](https://github.com/calcom/cal.com/pull/9337/files#r1248862915) and [this](https://github.com/calcom/cal.com/pull/9337/files#r1250164219) change 
- [x] [Possible 500](https://github.com/calcom/cal.com/pull/9337/files#r1248873555)
- [x] Clarity on https://github.com/calcom/cal.com/pull/9337/files#r1250136887
- [x] [Permission related issues](https://github.com/calcom/cal.com/pull/9337#issuecomment-1617307264)
- TODO
	- [x] Zoom installation because state is being passed now
	- [x] Stripe booking payment test for Team App installation and user App installation. Refer to this [change](https://github.com/calcom/cal.com/pull/9337/files#r1248874162)
	- [x] Test Managed Event Types.


apps/web/pages/api/integrations/[...args].ts
```diff
@@ -29,21 +31,21 @@ const defaultIntegrationAddHandler = async ({
...
-  await createCredential({ user: user, appType, slug });
+  await createCredential({ user: user, appType, slug, teamId });
 };
 
 const handler = async (req: NextApiRequest, res: NextApiResponse) => {
   // Check that user is authenticated
   req.session = await getServerSession({ req, res });
 
-  const { args } = req.query;
+  const { args, teamId } = req.query;
```
Comment (sean-brydon) [inline apps/web/pages/api/integrations/[...args].ts:48]: Shall we parse this with zod? Makes sure we dont have to cast as number below either (L:71)

packages/app-store/_utils/createAppCredential.ts
```diff
@@ -0,0 +1,57 @@
...
+) => {
+  const userId = req.session?.user.id;
+  // For OAuth flows, see if a teamId was passed through the state
+  const state = decodeOAuthState(req);
+  if (state?.teamId) {
+    // Check that the user belongs to the team
+    const team = await prisma.team.findUnique({
+      where: { id: state.teamId },
+      select: { id: true, members: { select: { userId: true } } },
+    });
```
sean-brydon, line 29: Havent tested but this should allow us to leave the search in the prisma query instead of in the IF below (L:31) 

```suggestion
    const team = await prisma.team.findUnique({
     where:{
      id:state.teamId,
      members:{
        some:{
          userId:ctx.user.id
        }
      }
    },
     select: { id: true, members: { select: { userId: true } } },
    });
```

packages/app-store/_utils/createAppCredential.ts
```diff
@@ -0,0 +1,57 @@
...
+        data: {
+          type: appData.type,
+          key: key || {},
+          teamId: state.teamId,
+          appId: appData.appId,
+        },
+      });
+    } else {
+      throw new Error("User does not belong to the team");
+    }
```
sean-brydon, line 42: If the above logic works we can early return 

if(!team) throw new Error etc

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -57,9 +57,12 @@ function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMuta
             location.search
           ),
       ...(type === "google_calendar" && { installGoogleVideo: options?.installGoogleVideo }),
+      ...(variables && variables.teamId && { teamId: variables.teamId }),
     };
     const stateStr = encodeURIComponent(JSON.stringify(state));
-    const searchParams = `?state=${stateStr}`;
+    const searchParams = `?state=${stateStr}${
+      variables && variables.teamId ? `&teamId=${variables.teamId}` : ""
+    }`;
```
sean-brydon, line 65: More readable and cleaner imho 

```suggestion
const searchParams = new URLSearchParams({
  state: stateStr,
  ...(variables && variables.teamId && { teamId: variables.teamId }),
}).toString();
```

packages/features/ee/teams/lib/getUserAdminTeams.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  return teams;
+};
+
+export default getUserAdminTeams;
+
+export type UserAdminTeams =
+  | (Awaited<Promise<ReturnType<typeof getUserAdminTeams>>>[number] & { isUser?: boolean })[]
+  | [];
```
sean-brydon, line 47: Can be simplified a little bit. 

We dont need | [] from what i can see as even if there isnt any teams the typeof the array would still be 
`teams: {
    id: number;
    name: string;
    logo: string | null;
}[]
`
Just empty 

```suggestion
export type UserAdminTeams = (Awaited<ReturnType<typeof getUserAdminTeams>>[number] & { isUser?: boolean })[];

```

sean-brydon (commented):
This is some amazing work joe. Left a few NITs. Impressed with how quick you got this together

packages/app-store/_utils/createAppCredential.ts
joeauyeung: Good catch, made the change.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -863,7 +863,10 @@ async function handler(
   const seed = `${organizerUser.username}:${dayjs(reqBody.start).utc().format()}:${new Date().getTime()}`;
   const uid = translator.fromUUID(uuidv5(seed, uuidv5.URL));
 
-  const bookingLocation = getLocationValueForDB(locationBodyString, eventType.locations);
+  const { bookingLocation, bookingLocationCredentialId } = getLocationValueForDB(
```
joeauyeung, line 866: If there is a `credentialId` then it will parse that from the event type location JSON

packages/features/ee/managed-event-types/hooks/useLockedFieldsManager.tsx
```diff
@@ -8,7 +8,7 @@ import type { _EventTypeModel } from "@calcom/prisma/zod/eventtype";
 import { Tooltip } from "@calcom/ui";
 import { Lock } from "@calcom/ui/components/icon";
 
-const Indicator = (label: string) => (
+export const LockedIndicator = (label: string) => (
```
joeauyeung, line 11: Export this component and pass to event type app cards

packages/app-store/giphy/api/add.ts
```diff
@@ -10,15 +10,19 @@ import getInstalledAppPath from "../../_utils/getInstalledAppPath";
  * @param res
  */
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  console.log("🚀 ~ file: add.ts:13 ~ handler ~ req:", req);
```
alannnc, line 13:
```suggestion
```

packages/app-store/giphy/lib/giphyManager.ts
```diff
@@ -6,6 +6,7 @@ let api_key = "";
 
 const checkGiphyApiKey = async () => {
   const appKeys = await getAppKeysFromSlug("giphy");
+  console.log("🚀 ~ file: giphyManager.ts:9 ~ checkGiphyApiKey ~ appKeys:", appKeys);
```
alannnc, line 9:
```suggestion
```

packages/app-store/routing-forms/api/add.ts
```diff
@@ -9,12 +9,13 @@ const handler: AppDeclarativeHandler = {
...
   supportsMultipleInstalls: false,
   handlerType: "add",
-  createCredential: async ({ user, appType, slug }) => {
+  createCredential: async ({ user, appType, slug, teamId }) => {
     return await prisma.credential.create({
       data: {
         type: appType,
         key: {},
-        userId: user.id,
+        // userId: user.id,
```
alannnc, line 17:
```suggestion
       userId: user.id,
```

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -27,12 +27,12 @@ type DeleteCredentialOptions = {
...
 
 export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOptions) => {
-  const { id, externalId } = input;
+  const { id, externalId, teamId } = input;
 
   const credential = await prisma.credential.findFirst({
     where: {
       id: id,
-      userId: ctx.user.id,
+      OR: [{ userId: ctx.user.id }, { teamId }],
```
alannnc, line 35: Don't use org here, if either it's provided just use that for the filter.
```suggestion
const whereDelete: Prisma.CredentialWhereUniqueInput  = {
   id,
};
if (teamId) {
whereDelete.teamId = teamId;
} else {
whereDelete.userId = ctx.user.id;
}

...
where: whereDelete,
...

```

packages/trpc/server/routers/loggedInViewer/locationOptions.handler.ts
```diff
@@ -4,22 +4,30 @@ import { getTranslation } from "@calcom/lib/server/i18n";
...
   };
+  input: TLocationOptionsInputSchema;
 };
 
-export const locationOptionsHandler = async ({ ctx }: LocationOptionsOptions) => {
+export const locationOptionsHandler = async ({ ctx, input }: LocationOptionsOptions) => {
   const credentials = await prisma.credential.findMany({
     where: {
-      userId: ctx.user.id,
+      OR: [{ userId: ctx.user.id }, input?.teamId ? { teamId: input.teamId } : {}],
```
alannnc, line 19: Here the same as above, adjust whereConditional if teamId is provided.

packages/app-store/fathom/components/EventTypeAppCardInterface.tsx
```diff
@@ -10,17 +9,17 @@ import type { appDataSchema } from "../zod";
...
     <AppCard
       setAppData={setAppData}
       app={app}
       switchOnClick={(e) => {
         if (!e) {
-          setEnabled(false);
+          updateEnabled(false);
         } else {
-          setEnabled(true);
+          updateEnabled(true);
```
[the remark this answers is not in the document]
joeauyeung: Fixed in multiple app cards and in the app card template

packages/features/ee/teams/lib/getUserAdminTeams.ts
joeauyeung: This wasn't picking up the property `isUser` so I decided to change this. Line 6 of the same file

packages/trpc/server/routers/loggedInViewer/locationOptions.handler.ts
joeauyeung: We're not including locations for V1 but will keep this in mind.

apps/web/components/AppListCard.tsx
```diff
@@ -85,6 +88,21 @@ export default function AppListCard(props: AppListCardProps) {
...
           )}
         </div>
+        {credentialOwner && (
+          <div>
+            <Badge variant="gray">
+              <div className="flex items-center">
+                <Avatar
+                  className="mr-2"
+                  alt={credentialOwner.name}
+                  size="sm"
```
sean-brydon, line 98: NIT: I think 24px x 24px here might be a bit big 
```suggestion
                  size="xs"
```

apps/web/components/eventtype/EventAppsTab.tsx
```diff
@@ -62,11 +66,51 @@ export const EventAppsTab = ({ eventType }: { eventType: EventType }) => {
     t("locked_fields_member_description")
   );
 
+  const appsWithTeamCredentials = eventTypeApps?.items.filter((app) => app.teams.length) || [];
+  const cardsForAppsWithTeams = appsWithTeamCredentials.map((app) => {
+    console.log("🚀 ~ file: EventAppsTab.tsx:104 ~ cardsForAppsWithTeams ~ app:", app);
```
sean-brydon, line 71:
```suggestion
```

apps/web/pages/apps/installed/[category].tsx
```diff
@@ -135,72 +137,106 @@ const IntegrationsList = ({ data, handleDisconnect, variant }: IntegrationsListP
...
+        description={item.description}
+        title={item.name}
+        logo={item.logo}
+        isDefault={appIsDefault}
+        shouldHighlight
+        slug={item.slug}
+        invalidCredential={item?.invalidCredentialIds ? item.invalidCredentialIds.length > 0 : false}
+        credentialOwner={item?.credentialOwner}
+        actions={
+          <div className="flex  justify-end">
```
Comment (sean-brydon) [inline apps/web/pages/apps/installed/[category].tsx:163]:
```suggestion
          <div className="flex justify-end">
```

packages/app-store/_components/AppCard.tsx
sean-brydon, line 1: Nice changes :)

packages/app-store/_utils/useIsAppEnabled.ts
```diff
@@ -0,0 +1,56 @@
+import { useState, useEffect } from "react";
+
+import type { EventTypeAppCardApp } from "../types";
+
+const useIsAppEnabled = (app: EventTypeAppCardApp, getAppData: unknown, setAppData: unknown) => {
```
sean-brydon, line 5: I managed to fix some of the type errors here (Might have knock on effects not sure) not blocking by any means 
```suggestion
import type { SetAppDataGeneric, GetAppDataGeneric } from "../EventTypeAppContext";
function useIsAppEnabled<TAppData extends ZodType>(
  app: EventTypeAppCardApp,
  getAppData: GetAppDataGeneric<TAppData>,
  setAppData: SetAppDataGeneric<TAppData>
){```

**Review by sean-brydon — COMMENTED** [review]
Some minor nits 

Great work here @joeauyeung im amazed how quick you got this done! @hariombalhara It will 100% be worth you taking a look into this

File: packages/app-store/_utils/useIsAppEnabled.ts (hunk 17 above)
Comment (joeauyeung) [reply packages/app-store/_utils/useIsAppEnabled.ts:5]: Where would `TAppData` come from? I'm thinking do we generate a schema file for each event type app (ex. `apps.keys-schemas.generated.ts` and pass it into `TAppData`?

Comment (sean-brydon) [reply packages/app-store/_utils/useIsAppEnabled.ts:5]: As discussed in DMs TAppData is infered from usage on getAppData and setAppData

File: packages/ui/components/apps/AppCard.tsx [hunk 18]
```diff
@@ -170,3 +167,103 @@ export function AppCard({ app, credentials, searchText }: AppCardProps) {
...
+    onError: (error) => {
+      if (error instanceof Error) showToast(error.message || t("app_could_not_be_installed"), "error");
+    },
+  });
+
+  if (
+    !userAdminTeams?.length ||
+    appCategories.some(
+      // We can remove these as we allow more of these installs
+      (category) => category === "calendar" || category === "video"
```
hariombalhara, line 202: As per [this](https://github.com/calcom/cal.com/blob/d051a54aa91d4a34f96680405a18dfba39f04a82/apps/web/components/apps/App.tsx#L101) only `calendar` category allows multiple install.

hariombalhara, line 202:
```suggestion
      (category) => category === "calendar"
```

packages/app-store/_components/AppCard.tsx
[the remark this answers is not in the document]
hariombalhara: Yeah it should certainly be renamed. How about calling it BaseAppCard ? Can't think of a better name, atleast it would resolve the name conflict with ui/components/AppCard

apps/web/components/AppListCard.tsx
```diff
@@ -85,6 +88,21 @@ export default function AppListCard(props: AppListCardProps) {
             </div>
           )}
         </div>
+        {credentialOwner && (
+          <div>
+            <Badge variant="gray">
+              <div className="flex items-center">
+                <Avatar
+                  className="mr-2"
+                  alt={credentialOwner.name || "Nameless"}
```
hariombalhara, line 97: Followup: We have nameless spread across the app. I think I confirmed elsewhere that name is a required field and it shouldn't be unset. But even if that's the case, we should introduce `function getName({name:string})`

apps/web/components/AppListCard.tsx
```diff
@@ -19,6 +20,7 @@ type AppListCardProps = {
   isTemplate?: boolean;
   invalidCredential?: boolean;
   children?: ReactNode;
+  credentialOwner?: CredentialOwner;
```
hariombalhara, line 23: I think we should make it a required field and expect `null` to be passed explicitly if someone doesn't want to pass `credentialOwner`. It helps in catching unexpected misses in prop passing.

I see CalendarListContainer(and other places) not passing the prop? Is that intentional?
Screenshot 2023-06-14 at 6 05 21 PM

packages/ui/components/apps/AppCard.tsx
joeauyeung: Sorry this was an artifact comment. For orgs right now we don't want calendars and video apps to be installed team wide.

apps/web/components/AppListCard.tsx
joeauyeung: The places where we don't want to pass the `credentialOwner` prop are the apps where we don't want to be installed team-wide (calendars & videos for now).

zomars (dismissed):
e2e tests are failing due to new install app dropdown. Let's fix the code or update the test to take the dropdown into account.

joeauyeung:
> e2e tests are failing due to new install app dropdown. Let's fix the code or update the test to take the dropdown into account.

I fixed one test that was affected by this PR. The other tests are touching the bookers which are out of scope of this PR. By chance are other tests failing as well?

packages/app-store/_utils/getCalendar.ts
```diff
@@ -19,8 +25,9 @@ export const getCalendar = async (credential: CredentialPayload | null): Promise
     return null;
   }
 
-  const calendarApp = await calendarAppImportFn();
-  if (!(calendarApp && "lib" in calendarApp && "CalendarService" in calendarApp.lib)) {
+  const calendarApp = (await calendarAppImportFn()) as CalendarApp;
+
+  if (!calendarApp.lib?.CalendarService) {
```
joeauyeung, line 30: I added this new `CalendarApp` type as I got this weird type error.

https://www.loom.com/share/7dd62f03e6af463bb51605408b0a9e8c

packages/trpc/server/routers/loggedInViewer/appCredentialsByType.handler.ts
```diff
@@ -9,7 +11,34 @@ type AppCredentialsByTypeOptions = {
   input: TAppCredentialsByTypeInputSchema;
 };
 
+/** Used for grabbing credentials on specific app pages */
 export const appCredentialsByTypeHandler = async ({ ctx, input }: AppCredentialsByTypeOptions) => {
   const { user } = ctx;
-  return user.credentials.filter((app) => app.type == input.appType).map((credential) => credential.id);
+  const userAdminTeams = await getUserAdminTeams({ userId: ctx.user.id, getUserInfo: true });
+
+  const teamIds = userAdminTeams.reduce((teamIds, team) => {
```
joeauyeung, line 19: I'm getting a weird type error here. The type check thinks that `userAdminTeams` is type never but locally, it's being typed correctly.

image

ericrommel (changes requested):
Labels from Team Event Type are not being updated after the app installations.

image

joeauyeung:
> Labels from Team Event Type are not being updated after the app installations.
> 
> image

@ericrommel this is now fixed

ericrommel (changes requested):
The form for new orgs shows `undefined` as the default value for `Organization URL`. Also, after filling out the `Your organization email address` field, the `Organization URL` doesn't change.

image

ericrommel (dismissed):
Apps already installed are not being shown on the Event Type area, under app installed. E.g., Fathon is already installed, but when adding an Event Type, it's not. However, if you try to install it again, you get a notification saying that it's already installed.
image

image

apps/web/components/eventtype/EventAppsTab.tsx
```diff
@@ -92,15 +125,18 @@ export const EventAppsTab = ({ eventType }: { eventType: EventType }) => {
...
-              setAppData={getAppDataSetter(app.slug as EventTypeAppsList)}
-              key={app.slug}
-              app={app}
-              eventType={eventType}
-            />
-          ))}
+          {cardsForAppsWithTeams.map((apps) => apps.map((cards) => cards))}
+          {installedApps
+            ?.filter((app) => !app.teams.length)
+            .map((app) => (
```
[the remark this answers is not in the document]
zomars: @sean-brydon you're kinda right. But this code would leave us with a mixed array of teams and undefined values.

apps/web/components/eventtype/EventAppsTab.tsx
```diff
@@ -16,12 +16,16 @@ export type EventType = Pick<EventTypeSetupProps, "eventType">["eventType"] &
...
+    teamId: eventType.team?.id || eventType.parent?.teamId,
   });
+
   const methods = useFormContext<FormValues>();
-  const installedApps = eventTypeApps?.filter((app) => app.credentials.length);
-  const notInstalledApps = eventTypeApps?.filter((app) => !app.credentials.length);
+  const installedApps =
+    eventTypeApps?.items.filter((app) => app.credentialIds.length || app.teams.length) || [];
+  const notInstalledApps =
+    eventTypeApps?.items.filter((app) => !app.credentialIds.length && !app.teams.length) || [];
```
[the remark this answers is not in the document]
zomars: Got for it!

packages/app-store/_utils/getCalendar.ts
zomars: we can solve this with a type guard as well

zomars: Type narrowing to be exact: https://www.typescriptlang.org/docs/handbook/2/narrowing.html

apps/web/components/apps/App.tsx
```diff
@@ -385,3 +386,97 @@ export default function App(props: {
...
+          {multiInstall ? t("install_another") : t("install_app")}
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuPortal>
+        <DropdownMenuContent>
+          <DropdownMenuLabel>{t("install_app_on")}</DropdownMenuLabel>
+          {userAdminTeams.map((team) => (
+            <DropdownItem
+              type="button"
+              data-testid={team.isUser ? "install-app-button-personal" : "anything else"}
```
zomars, line 449: Why do we need to do this? I feel that we're introducing unnecessary bloat here. It's hard to understand why do we need this by reading the code itself.

joeauyeung: It's a part of the first iteration for orgs until the account switcher is created. @Jaibles

packages/lib/constants.ts
```diff
@@ -8,7 +8,7 @@ export const WEBAPP_URL =
   RAILWAY_STATIC_URL ||
   HEROKU_URL ||
   RENDER_URL ||
-  "http://localhost:3000";
+  "http:/localhost:3000";
```
hariombalhara, line 11:
```suggestion
  "http://localhost:3000";
```

packages/ui/components/apps/AppCard.tsx
```diff
@@ -170,3 +167,100 @@ export function AppCard({ app, credentials, searchText }: AppCardProps) {
...
+              onClick={() => {
+                mutation.mutate(
+                  team.isUser ? addAppMutationInput : { ...addAppMutationInput, teamId: team.id }
+                );
+              }}>
+              <p>
+                {team.name}{" "}
+                {credentials &&
+                  credentials.some((credential) =>
+                    credential?.teamId ? credential?.teamId === team.id : credential.userId === team.id
```
hariombalhara, line 256: It would be better to extract it out in a variable called `isInstalledForTeamOrUser` as it's used in 2 places.

hariombalhara:
Not able to install an app to multiple teams. After installing it for user, I can't install it for a team.
https://www.loom.com/share/8e86d8fc5dc04b8f9b86c658adbaf441

hariombalhara:
> Not able to install an app to multiple teams. After installing it for user, I can't install it for a team. [loom.com/share/8e86d8fc5dc04b8f9b86c658adbaf441](https://www.loom.com/share/8e86d8fc5dc04b8f9b86c658adbaf441)

Fixed it. It was a simple non-use of teamId which I think started happening after some change. cc @joeauyeung I would be reviewing, testing and fixing things that I can.

hariombalhara:
On EventType - Apps Tab, an app for a Team Event(created for an organization's(VWO) Team(Test Team)) is shown twice.
image

Fathom app is installed for both VWO(organization) and Test Team(a team of VWO organization) and it is shown twice due to it.
Screenshot 2023-06-29 at 4 40 27 PM

packages/app-store/_utils/createOAuthAppCredential.ts
```diff
@@ -0,0 +1,64 @@
...
+  // For OAuth flows, see if a teamId was passed through the state
+  const state = decodeOAuthState(req);
+  if (state?.teamId) {
+    // Check that the user belongs to the team
+    const team = await prisma.team.findFirst({
+      where: {
+        id: state.teamId,
+        members: {
+          some: {
+            userId: req.session?.user.id,
```
hariombalhara, line 31:
```suggestion
            userId: req.session?.user.id,
            accepted: true
```
We are not verifying the accepted status. So, members in pending invitation state might be able to create credentials for team.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -57,9 +57,13 @@ function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMuta
             location.search
           ),
       ...(type === "google_calendar" && { installGoogleVideo: options?.installGoogleVideo }),
+      ...(variables && variables.teamId && { teamId: variables.teamId }),
```
hariombalhara, line 60: nit: Better to save variables.teamId in a variable and reuse that.

packages/app-store/_utils/useIsAppEnabled.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!enabled) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      setAppData("credentialId", undefined);
+    }
+  }, [enabled]);
+
+  const updateEnabled = (newValue: boolean) => {
+    setEnabled(newValue);
+  };
```
hariombalhara, line 57: I feel this logic can be heavily simplified like this. It also avoids useEffect which are always the last resort.
```Javascript
const [enabled, setEnabled] = useState(() => {
    if (!app.credentialOwner) {
      return getAppData("enabled");
    }
    const credentialId = getAppData("credentialId");
    const isAppEnabledForCredential =
      app?.credentialIds && app.credentialIds.some((id) => id === credentialId);
    return isAppEnabledForCredential;
  });

  const updateEnabled = (newValue: boolean) => {
    if (!newValue) {
      setAppData("credentialId", undefined);
    }

    if (newValue && app.credentialIds) {
      setAppData("credentialId", app.credentialIds[0]);
    }
    setEnabled(newValue);
  };
```

packages/app-store/_utils/useIsAppEnabled.ts
```diff
@@ -0,0 +1,62 @@
+import { useState, useEffect } from "react";
+import type { ZodType } from "zod";
+
+import type { SetAppDataGeneric, GetAppDataGeneric } from "../EventTypeAppContext";
+import type { EventTypeAppCardApp } from "../types";
+
+function useIsAppEnabled<TAppData extends ZodType>(
+  app: EventTypeAppCardApp,
+  getAppData: GetAppDataGeneric<TAppData>,
+  setAppData: SetAppDataGeneric<TAppData>
```
hariombalhara, line 10: I think instead of passing these as args we should be able to get them from `useAppContextWithSchema`

packages/app-store/ga4/components/EventTypeAppCardInterface.tsx
```diff
@@ -1,32 +1,36 @@
...
 import { TextField } from "@calcom/ui";
 
 import type { appDataSchema } from "../zod";
 
-const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({ app }) {
+const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({
+  app,
+  eventType,
+  disabled,
+  LockedIcon,
```
hariombalhara, line 13: Can we get these from useAppContextWithSchema itself ?

packages/app-store/utils.ts
```diff
@@ -57,10 +57,15 @@ export function getLocationGroupedOptions(integrations: ReturnType<typeof getApp
   integrations.forEach((app) => {
     if (app.locationOption) {
       // All apps that are labeled as a locationOption are video apps. Extract the secondary category if available
-      let category =
-        app.categories.length >= 2 ? app.categories.find((category) => category !== "video") : app.category;
-      if (!category) category = "video";
-      const option = { ...app.locationOption, icon: app.logo, slug: app.slug };
+      const nonVideoCategories = app.categories.find((category) => category !== "video");
```
hariombalhara, line 60:
```suggestion
      const hasNonVideoCategory = app.categories.find((category) => category !== "video");
```

packages/app-store/zoomvideo/api/add.ts
```diff
@@ -19,11 +20,13 @@ async function handler(req: NextApiRequest) {
   });
 
   const { client_id } = await getZoomAppKeys();
+  const state = encodeOAuthState(req);
```
hariombalhara, line 23: nit: I find it a more apt name `getOAuthStateAsJson`

packages/core/EventManager.ts
```diff
@@ -371,7 +371,7 @@ export default class EventManager {
    * @private
    */
 
-  private getVideoCredential(event: CalendarEvent): CredentialWithAppName | undefined {
+  private async getVideoCredential(event: CalendarEvent): Promise<CredentialWithAppName | undefined> {
```
hariombalhara, line 374: Why is it declared async now without any change in functionality?

packages/features/auth/lib/getServerSession.ts
```diff
@@ -46,7 +46,7 @@ export async function getServerSession(options: {
     return cachedSession;
   }
 
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 49: Why are we doing a findFirst. emails is still unique field. Looks like a change from one of the org branches that was reverted before going to main.

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -69,7 +69,7 @@ const providers: Provider[] = [
         throw new Error(ErrorCode.InternalServerError);
       }
 
-      const user = await prisma.user.findUnique({
+      const user = await prisma.user.findFirst({
```
hariombalhara, line 72: This too should be reverted.

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -593,8 +596,10 @@ async function handler(req: CustomRequest) {
     }
 
     // Posible to refactor TODO:
-    const paymentApp = await appStore[paymentAppCredential?.app?.dirName as keyof typeof appStore]();
-    if (!(paymentApp && "lib" in paymentApp && "PaymentService" in paymentApp.lib)) {
+    const paymentApp = (await appStore[
+      paymentAppCredential?.app?.dirName as keyof typeof appStore
+    ]()) as PaymentApp;
+    if (!paymentApp.lib?.PaymentService) {
```
hariombalhara, line 602: Why are we doing a type assertion? Earlier code was safer as it's possible that paymentApp or paymentApp.lib is nullish

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1559,7 +1559,9 @@ async function handler(
       if (!Number.isNaN(paymentAppData.price) && paymentAppData.price > 0 && !!booking) {
         const credentialPaymentAppCategories = await prisma.credential.findMany({
           where: {
-            userId: organizerUser.id,
+            ...(paymentAppData.credentialId
+              ? { id: paymentAppData.credentialId }
+              : { userId: organizerUser.id }),
```
hariombalhara, line 1564: nit: I think we could just use `credentialId` in both team App and non team app installation case. It would simplify the code.

apps/web/components/apps/App.tsx
```diff
@@ -385,3 +386,97 @@ export default function App(props: {
...
+              onClick={() => {
+                mutation.mutate(
+                  team.isUser ? addAppMutationInput : { ...addAppMutationInput, teamId: team.id }
+                );
+              }}>
+              <p>{team.name}{" "}
+                {credentials &&
+                  credentials.some((credential) =>
+                    credential?.teamId ? credential?.teamId === team.id : credential.userId === team.id
+                  ) &&
```
hariombalhara, line 474: Should be abstracted it into a variable like `isInstalled`.

apps/web/components/ui/form/LocationSelect.tsx
```diff
@@ -40,11 +42,13 @@ export default function LocationSelect(props: Props<LocationOption, false, Group
...
-            <OptionWithIcon icon={props.data.icon} label={props.data.label} />
-          </components.Option>
-        ),
+        Option: (props) => {
+          return (
+            <components.Option {...props}>
+              <OptionWithIcon icon={props.data.icon} label={props.data.label} />
+            </components.Option>
+          );
+        },
```
hariombalhara, line 51: Looks like we are not supporting credentialId in here. We can revert this change.

Also @joeauyeung are we planning to enable Video Apps for teams with this PR?

packages/trpc/server/routers/loggedInViewer/integrations.handler.ts
```diff
@@ -12,21 +14,57 @@ type IntegrationsOptions = {
...
+    });
+    credentials = [...credentials, ...teamAppCredentials];
+  }
 
   const enabledApps = await getEnabledApps(credentials);
   //TODO: Refactor this to pick up only needed fields and prevent more leaking
   let apps = enabledApps.map(
     ({ credentials: _, credential: _1, key: _2 /* don't leak to frontend */, ...app }) => {
-      const credentialIds = credentials.filter((c) => c.type === app.type).map((c) => c.id);
+      const credentialIds = credentials.filter((c) => c.type === app.type && !c.teamId).map((c) => c.id);
```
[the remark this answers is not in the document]
hariombalhara: Either we should not filter out team credentials or we should rename it to `userCredentialIds`. It is confusing as is.

packages/trpc/server/routers/loggedInViewer/deleteMeWithoutPassword.handler.ts
```diff
@@ -12,7 +12,7 @@ type DeleteMeWithoutPasswordOptions = {
 };
 
 export const deleteMeWithoutPasswordHandler = async ({ ctx }: DeleteMeWithoutPasswordOptions) => {
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 15:
```suggestion
  const user = await prisma.user.findUnique({
```

packages/trpc/server/routers/loggedInViewer/integrations.handler.ts
```diff
@@ -44,8 +124,20 @@ export const integrationsHandler = async ({ ctx, input }: IntegrationsOptions) =
...
+    );
+  }
+
+  if (extendsFeature) {
+    apps = apps
+      .filter((app) => app.extendsFeature?.includes(extendsFeature))
+      .map((app) => ({
+        ...app,
+        isInstalled: !!app.credentialIds?.length || !!app.teams?.length || app.isGlobal,
+      }));
```
hariombalhara, line 138: What's the significance of this code? I don't think it's being used.

apps/web/components/eventtype/EventAppsTab.tsx
```diff
@@ -16,12 +16,16 @@ export type EventType = Pick<EventTypeSetupProps, "eventType">["eventType"] &
 
 export const EventAppsTab = ({ eventType }: { eventType: EventType }) => {
   const { t } = useLocale();
-  const { data: eventTypeApps, isLoading } = trpc.viewer.apps.useQuery({
```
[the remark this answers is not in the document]
hariombalhara: This makes viewer.apps a dead code. We can remove that.

packages/trpc/server/routers/loggedInViewer/integrations.handler.ts
```diff
@@ -10,23 +13,100 @@ type IntegrationsOptions = {
...
   const { user } = ctx;
-  const { variant, exclude, onlyInstalled } = input;
-  const { credentials } = user;
+  const { variant, exclude, onlyInstalled, includeTeamInstalledApps, extendsFeature, teamId } = input;
+  let { credentials } = user;
+  let userAdminTeams: TeamQueryWithParent[] = [];
+
+  if (includeTeamInstalledApps || teamId) {
+    // Get app credentials that the user is an admin or owner to
+    userAdminTeams = await prisma.team.findMany({
```
hariombalhara, line 37: nit: Good candidate for getUserAdminTeams

packages/trpc/server/routers/loggedInViewer/integrations.handler.ts
```diff
@@ -10,23 +13,100 @@ type IntegrationsOptions = {
...
+      where: {
+        ...(teamId && { id: teamId }),
+        members: {
+          some: {
+            userId: user.id,
```
hariombalhara, line 42:
```suggestion
            userId: user.id,
            accepted: true
```

hariombalhara:
Non admin/owner team member not able to see the App installed for it's Team or Org.
- Installed Apps View
Screenshot 2023-07-03 at 10 16 39 AM
Screenshot 2023-07-03 at 10 16 50 AM

hariombalhara (changes requested):
So many complexities and changes in this PR. Great work so far @joeauyeung 🙏.
I have left some comments. The feedback that's important to be addressed, I have added in the PR itself so that it doesn't get lost with so many comments.

joeauyeung:
> On EventType - Apps Tab, an app for a Team Event(created for an organization's(VWO) Team(Test Team)) is shown twice. image
> 
> Fathom app is installed for both VWO(organization) and Test Team(a team of VWO organization) and it is shown twice due to it. Screenshot 2023-06-29 at 4 40 27 PM



> On EventType - Apps Tab, an app for a Team Event(created for an organization's(VWO) Team(Test Team)) is shown twice. image
> 
> Fathom app is installed for both VWO(organization) and Test Team(a team of VWO organization) and it is shown twice due to it. Screenshot 2023-06-29 at 4 40 27 PM

This is done on purpose. There could be an instance where there is the same app but an organization scoped credential maybe different from a team scoped credential.

packages/core/EventManager.ts
joeauyeung: I did have locations enabled for this PR, but we realized it will take more design work. There are some remnants of that functionality that I may have missed.

packages/features/bookings/lib/handleCancelBooking.ts
joeauyeung: I was getting a weird type error in this file. Here's a Loom with what I was experiencing. Still the same app.

https://www.loom.com/share/7dd62f03e6af463bb51605408b0a9e8c

packages/features/bookings/lib/handleNewBooking.ts
joeauyeung: This was for backwards compatibility if current `paymentAppData` doesn't contain the `credentialId`

apps/web/components/ui/form/LocationSelect.tsx
joeauyeung: Adding video apps will be in a separate PR as it requires more designs.

packages/app-store/gtm/zod.ts
```diff
@@ -4,12 +4,15 @@ import { eventTypeAppCardZod } from "@calcom/app-store/eventTypeAppCardZod";
...
-    }),
+    trackingId: z
+      .string()
+      .transform((val) => {
+        let trackingId = val.trim();
+        // Ensure that trackingId is transformed if needed to begin with "GTM-" always
+        trackingId = !trackingId.startsWith("GTM-") ? `GTM-${trackingId}` : trackingId;
+        return trackingId;
+      })
+      .optional(),
```
hariombalhara, line 15: What's the need to make it optional?

packages/trpc/server/routers/loggedInViewer/integrations.handler.ts
joeauyeung: Good point. This query included the parent team so I add a `getParentInfo` to `getUserAdminTeams`

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara: Ah makes sense.

packages/trpc/server/routers/loggedInViewer/integrations.handler.ts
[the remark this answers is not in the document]
joeauyeung: @hariombalhara I changed the the variable to `userCredentialIds`. I searched where `integrations.useQuery()` was used to make sure the change didn't affect any functionality. You can see the commit here 08ff16d500f2c54340ac8686a6a30c1b942d5829

packages/app-store/ga4/components/EventTypeAppCardInterface.tsx
joeauyeung: Great idea, moved these into the provider in this commit b2826f5497df8b671f405c8bd0f768b0d8c9a88e

packages/trpc/server/routers/loggedInViewer/deleteMeWithoutPassword.handler.ts
hariombalhara: @joeauyeung This and some other comments are marked resolved but there aren't any changes. Can you add a comment if it's not supposed to change before resolving the comment.
If you have made the suggested change you can just resolve the conversation.

joeauyeung:
@hariombalhara installed apps for teams will now appear for members. These apps will be readonly. Commit 
40e20dcaa196ba94dfb0e5a18ac775af5a73535d

image

packages/trpc/server/routers/loggedInViewer/deleteMeWithoutPassword.handler.ts
joeauyeung: With the comments without any changes, it looks like it synced properly with `main` so there wouldn't be any changes reflected in this PR.

packages/app-store/_utils/createAppCredential.ts
joeauyeung: @hariombalhara change made in `createOAuthAppCredentials`

packages/lib/constants.ts
joeauyeung: Added after merge with `main`

packages/app-store/utils.ts
joeauyeung: Merge conflict with `main` didn't have these lines

packages/features/auth/lib/getServerSession.ts
joeauyeung: Merged with `main`

packages/features/ee/sso/lib/saml.ts
```diff
@@ -31,7 +31,7 @@ export const isSAMLAdmin = (email: string) => {
 };
 
 export const samlTenantProduct = async (prisma: PrismaClient, email: string) => {
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 34:
```suggestion
  const user = await prisma.user.findUnique({
```

packages/trpc/server/routers/loggedInViewer/deleteMe.handler.ts
```diff
@@ -20,7 +20,7 @@ type DeleteMeOptions = {
 
 export const deleteMeHandler = async ({ ctx, input }: DeleteMeOptions) => {
   // Check if input.password is correct
-  const user = await prisma.user.findUnique({
+  const user = await prisma.user.findFirst({
```
hariombalhara, line 23:
```suggestion
  const user = await prisma.user.findUnique({
```

packages/features/bookings/lib/handleCancelBooking.ts
hariombalhara, line 602:
```suggestion
    if (!paymentApp?.lib?.PaymentService) {
```
That type error is seriously weird. Not focussing on it, just making the code runtime error free

hariombalhara (commented):
QR Code App went crazy. I have it installed just once for Organization. Let's fix it in a followup.



**Org Owner View**

hariombalhara (dismissed):
The biggest PR I have ever reviewed 🙏 Thanks @joeauyeung for the quick implementation of feedback. 
I have tested org stuff and non-org stuff and feel confident enough to merge it. There might still be bugs but we can do followup for that.

hariombalhara (dismissed):

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
```diff
@@ -40,8 +40,11 @@ export const createHandler = async ({ ctx, input }: CreateOptions) => {
     locations = [{ type: DailyLocationType }];
   }
 
-  // If its defaulting to daily no point handling compute as its done
-  if (defaultConferencingData && defaultConferencingData.appSlug !== "daily-video") {
+  // If it is a team event default to the organizer's default conferencing app
+  if (teamId) {
+    locations = [{ type: "conferencing" }];
```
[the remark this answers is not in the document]
joeauyeung: It was going to small fix I was putting into the PR but it's going to need more work for managed event types. I'm removing the change and creating a separate PR for this.

packages/app-store/fathom/components/EventTypeAppCardInterface.tsx
```diff
@@ -1,32 +1,31 @@
...
+  const [getAppData, setAppData, LockedIcon, disabled] = useAppContextWithSchema<typeof appDataSchema>();
   const trackingId = getAppData("trackingId");
-  const [enabled, setEnabled] = useState(getAppData("enabled"));
+  const { enabled, updateEnabled } = useIsAppEnabled(app);
 
   return (
     <AppCard
       setAppData={setAppData}
       app={app}
+      disableSwitch={disabled}
```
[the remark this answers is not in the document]
joeauyeung: I found some other improvements we can make, so I'm creating a separate PR.

https://github.com/calcom/cal.com/pull/9950

hariombalhara (approved):

PR #9374, opened 2023-06-07
fix: Prevents undefined features flags on first render

zomars:
## What does this PR do?

This allows feature flags to be prefetched on most pages. You still need to use `getServerSideProps` and `ssrInit` helper in order to benefit from this.

## Type of change

## How should this be tested?

- [x] On any page with `ssrInit` try using `const featureFlags = useFlagMap();`
- [x] It should never return an empty object.

## Mandatory Tasks

### DIFF
--- a/apps/web/server/lib/ssr.ts
+++ b/apps/web/server/lib/ssr.ts
@@ -26,6 +26,8 @@ export async function ssrInit(context: GetServerSidePropsContext) {
 
   // always preload "viewer.public.i18n"
   await ssr.viewer.public.i18n.fetch();
+  // So feature flags are available on first render
+  await ssr.viewer.features.map.prefetch();
 
   return ssr;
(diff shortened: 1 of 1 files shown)


apps/web/server/lib/ssr.ts
```diff
@@ -26,6 +26,8 @@ export async function ssrInit(context: GetServerSidePropsContext) {
 
   // always preload "viewer.public.i18n"
   await ssr.viewer.public.i18n.fetch();
+  // So feature flags are available on first render
```
zomars, line 29: I chose this common helper instead of each individual page because prefetched feature flags would be desirable in most cases.

zomars (commented):
Self review done

roae (approved):
LGTM! 🙏

PR #9390, opened 2023-06-07
refactor: Update test utils

azadnsu:
## What does this PR do?

- One helper function was added for asserting toaster text. None of the tests has toaster text assertion. This function could help in all the specs where toaster text assertion could be included. Added one check-in team's creation test. 
- Also, one extra assertion was added in the team's spec to verify slug remains the same after the team created

## Type of change

- Improvement in test utility library

## How should this be tested?

- Run teams.e2e.ts test multiple times and see test passes always


keithwillcode (changes requested):
@azadnsu Thanks for the contribution! 

Can you please revert the yarn.lock change?

azadnsu:
> @azadnsu Thanks for the contribution!
> 
> Can you please revert the yarn.lock change?

@keithwillcode I deleted yarn.lock from this pull request. Could you please check now? Thanks for the quick check. It was my first PR.

PR #9392, opened 2023-06-07
refactor: TwoFactor.tsx

Riyan-Mo:
## What does this PR do?

Improvements made:
1. Renamed the onChange function to setValue for clarity.
2. Changed the usage of onChange to setValue in useDigitInput and removed the dependency on the value state variable.
3. Replaced the hidden type="hidden" input with the register function from react-hook-form.
4. Used map to render the Input components instead of manually repeating them.
5. Updated the key prop for each Input component to include the index.
6. Added the autoFocus prop to the first Input component for better user experience.
7. Removed the unnecessary eslint-disable-next-line comment.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Mandatory Tasks

## Checklist


Udit-takkar (approved):
LGTM

PR #9393, opened 2023-06-07
fix: for recurring events in new booker

JeroenReumkens:
## What does this PR do?

Recurring event doesn't neccessarily mean an event that requires confirmation. We need to show the recurring option nevertheless. Before recurring events option wasn't shown when event did NOT require confirmation. 

## Type of change


emrysal (approved):
lgtm :)

PR #9394, opened 2023-06-07
fix: Make public event endpoint take less data

JeroenReumkens:
## What does this PR do?

Only take owner data we need in public event api, instead of everything. Saves bandwidth.

## Type of change


emrysal (approved):
lgtm!

PR #9395, opened 2023-06-07
feat: organizations sidebar

leog:
## What does this PR do?

Fixes # (issue)

## Organization member
imagen

## Simple user
imagen

## Type of change


sean-brydon:
CleanShot 2023-06-13 at 17 35 04@2x
When logging in with a new account (none org) we get a flash of this before redirected to getting-started

packages/features/shell/Shell.tsx
```diff
@@ -785,39 +780,92 @@ type SideBarContainerProps = {
...
+  const { status, data } = useSession();
   const router = useRouter();
 
   // Make sure that Sidebar is rendered optimistically so that a refresh of pages when logged in have SideBar from the beginning.
   // This improves the experience of refresh on app store pages(when logged in) which are SSG.
   // Though when logged out, app store pages would temporarily show SideBar until session status is confirmed.
   if (status !== "loading" && status !== "authenticated") return null;
   if (router.route.startsWith("/v2/settings/")) return null;
-  return <SideBar bannersHeight={bannersHeight} />;
+  return <SideBar bannersHeight={bannersHeight} user={data?.user} />;
```
joeauyeung, line 795: If there isn't a user session we shouldn't render the side bar. Example the App store is available publicly but with no user logged in the sidebar still renders.

@sean-brydon side not the verify email banner still shows up on top.

image

sean-brydon (dismissed):
CleanShot 2023-06-13 at 18 01 46@2x
Few design changes that need fixed that we spotted.

[Figma designs](https://www.figma.com/file/mWENDNaJPdRfZMtA75MghZ/Organizations?type=design&node-id=24%3A28821&t=kKgbYhFPXJLUYLYz-1)

packages/features/shell/Shell.tsx
```diff
@@ -343,29 +350,30 @@ function UserDropdown({ small }: { small?: boolean }) {
...
-              {user.away && (
-                <div className="border-muted absolute bottom-0 right-0 h-3 w-3 rounded-full border-2 bg-yellow-500" />
-              )}
+              <span
+                className={classNames(
+                  "border-muted absolute -bottom-1 -right-1  rounded-full border-2 bg-green-500",
+                  user.away ? "bg-yellow-500" : "bg-green-500",
+                  small ? "-bottom-0.5 -right-0.5 h-2.5 w-2.5" : "bottom-0 right-0 h-3 w-3"
+                )}
+              />
```
sean-brydon, line 359: I think the avatar can be switched out with our `Avatar` component

packages/features/shell/Shell.tsx
```diff
@@ -847,14 +907,44 @@ function SideBar({ bannersHeight }: SideBarProps) {
...
-              <UserDropdown small />
-            </span>
-          </div>
+          {bottomNavItems.map(({ icon: Icon, ...item }) => (
+            <Tooltip side="right" content={t(item.name)} className="lg:hidden" key={item.href}>
+              <Link
+                href={item.href}
+                aria-label={t(item.name)}
+                target={item.target}
+                className={classNames(
```
sean-brydon, line 916: To prevent the issue i mentioned above we have the `ButtonOrLink` component that changes the render based on if a href is passed

joeauyeung (dismissed):
Great work so far. Just a few comments, and I think should be blocking.

packages/features/shell/Shell.tsx
roae: I can't see the sidebar when I'm not authenticated, in the line 793 return null if the session status !== "authenticated" `  if (status !== "loading" && status !== "authenticated") return null;`

Jaibles:
CleanShot 2023-06-15 at 00 29 04@2x
@roae Looks like name is still not truncating. Height of it should be 36px not 40px also.

packages/features/shell/Shell.tsx
hariombalhara: I don't see the issue.

hariombalhara (approved):
LGTM!!

Tested 
- KBar
- Links at the bottom of sidebar 
- Away status toggle
- Long username
- Long teamname
- Different screensizes


Non Blocker Issues
- Skeleton Missing - [Bug filed](https://github.com/calcom/cal.com/issues/9535)
image

Fixed one issue:
- Removed hovered effect that was there on `Copy Page Link` without hovering

PR #9398, opened 2023-06-07
fix: Improve booker header for large views

JeroenReumkens:
## What does this PR do?

Show header layout in large views even if theres only one layout enable. Don't show toggle though.

CleanShot 2023-06-07 at 18 57 51@2x

Fixes CAL-1890

## Type of change

## How should this be tested?

- [ ] Enable only column view, header bar with date and prev/next toggle should be visible. There should be no additional space at the top though. Layout toggle should not be visible. 
- [ ] Enable only month view, this layout has no top bar, so that should not be visible. Layout toggle also should not be visible
- [ ] Enable both or all 3 layouts – layout toggle should be visible and in column view the header bar should be visible as well (with date range that's visible).

### DIFF
--- a/packages/features/bookings/Booker/components/Header.tsx
+++ b/packages/features/bookings/Booker/components/Header.tsx
@@ -32,7 +32,7 @@ export function Header({
     [setLayout]
   );
 
-  if (isMobile || !enabledLayouts || enabledLayouts.length <= 1) return null;
+  if (isMobile || !enabledLayouts) return null;
 
   // Only reason we create this component, is because it is used 3 times in this component,
   // and this way we can't forget to update one of the props in all places :)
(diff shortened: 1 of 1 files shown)


packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -32,7 +32,7 @@ export function Header({
     [setLayout]
   );
 
-  if (isMobile || !enabledLayouts || enabledLayouts.length <= 1) return null;
```
JeroenReumkens, line 35: We can't return early since we need to show the header for the column and week view. That bar contains the title and prev/next week buttons. That's why the check is moved further down.

PeerRich (approved):

PR #9406, opened 2023-06-07
chore: moved viewport meta from document to PageWrapper

### LINKED ISSUE #9405: [CAL-1917] chore: meta viewport in _app not _document
```
import Head from "next/head";

import type { AppProps } from "@lib/app-providers";

import "../styles/globals.css";

function MyApp(props: AppProps) {
  const { Component, pageProps } = props;
  if (Component.PageWrapper !== undefined) return Component.PageWrapper(props);
  return (
    <>
      <Head>
        <meta
          name="viewport"
          content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"
        />
      </Head>
      <Component {...pageProps} />
    </>
  );
}

export default trpc.withTRPC(MyApp);

```

<sub>[CAL-1917](https://linear.app/calcom/issue/CAL-1917/chore-meta-viewport-in-app-not-document)</sub>

PeerRich:
fixes https://nextjs.org/docs/messages/no-document-viewport-meta


apps/web/pages/_app.tsx
```diff
@@ -7,7 +9,17 @@ import "../styles/globals.css";
 function MyApp(props: AppProps) {
   const { Component, pageProps } = props;
   if (Component.PageWrapper !== undefined) return Component.PageWrapper(props);
```
keithwillcode, line 11: Do we not need to also do this in the case this line? i.e. When a page uses PageWrapper?

PeerRich: not really sure to be honest

emrysal: I think we do 👍

apps/web/pages/_app.tsx
```diff
@@ -7,7 +9,17 @@ import "../styles/globals.css";
...
-  return <Component {...pageProps} />;
+  return (
+    <>
+      <Head>
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"
+        />
+      </Head>
```
roae, line 19: @PeerRich I think this should be in the PageWrapper Component in order to get the meta tag in all pages.

Only the calendar-cache page reach this code and this meta tag is useless there.

roae: I have already did this change, I think the PR title must change

imagen

keithwillcode (approved):
Nice. Looks good.

PR #9423, opened 2023-06-08
feat: WIP: tRPC/OpenAPI for v2

zomars:
## What does this PR do?

Enables openAPI endpoints for selected tRPC procedures. Right now we only added 2 testing endpoints. We need to discuss if it's worth the effort to adopt and maintain this instead of improving our current public API.

## Type of change

## How should this be tested?

- [ ] `yarn dev:api`
- [ ] Got to localhost:3002/v2/openapi.json

## Mandatory Tasks

## Checklist


packages/features/bookings/schemas/bookingReadPublic.schema.ts
```diff
@@ -0,0 +1,57 @@
+import { z } from "zod";
+
+import { _AttendeeModel, _BookingModel as Booking, _PaymentModel, _UserModel } from "@calcom/prisma/zod";
+import { EventTypeMetaDataSchema } from "@calcom/prisma/zod-utils";
+
+export const bookingReadPublicSchema = Booking.extend({
```
zomars, line 6: Taken from API libs.

The bad news is that we need to set and output schema for EVERY single tRPC procedure that we want to include.

The good news is that we already have most output schemas in the API libs.

packages/prisma/zod-utils.ts
```diff
@@ -110,7 +110,9 @@ export type BookingFieldType = typeof BookingFieldType extends z.Values<infer T>
 
 // Validation of user added bookingFields' responses happen using `getBookingResponsesSchema` which requires `eventType`.
 // So it is a dynamic validation and thus entire validation can't exist here
-export const bookingResponses = z
+// ⬇️ Prevents annoying Prisma TS error: "JsonValue isn't assignable to blah blah blah"
```
zomars, line 113: Hope this is self explanatory

packages/trpc/server/routers/viewer/bookings/_router.tsx
```diff
@@ -18,21 +20,26 @@ type BookingsRouterHandlerCache = {
 const UNSTABLE_HANDLER_CACHE: BookingsRouterHandlerCache = {};
 
 export const bookingsRouter = router({
-  get: authedProcedure.input(ZGetInputSchema).query(async ({ input, ctx }) => {
-    if (!UNSTABLE_HANDLER_CACHE.get) {
-      UNSTABLE_HANDLER_CACHE.get = await import("./get.handler").then((mod) => mod.getHandler);
-    }
+  get: authedProcedure
+    // Uncomment the below line to see a nasty error
+    // .meta({ openapi: { method: "GET", path: "/bookings" } })
```
zomars, line 25: This is where I got stuck. I still don't know if it's because a GET endpoint expects a input in the form of an object, but uncommenting this makes the openAPI to explode.

packages/trpc/server/routers/viewer/bookings/get.handler.ts
```diff
@@ -167,6 +167,7 @@ export const getHandler = async ({ ctx, input }: GetOptions) => {
         uid: true,
         recurringEventId: true,
         location: true,
+        responses: true,
```
zomars, line 170: These were missing to comply with the output schema

apps/api/pages/api/v2/openapi.json.ts
```diff
@@ -0,0 +1,16 @@
...
+import { appRouter } from "@calcom/trpc/server/routers/_app";
+
+const openApiDocument = generateOpenApiDocument(appRouter, {
+  title: "tRPC OpenAPI",
+  version: "1.0.0",
+  baseUrl: "http://localhost:3003",
+});
+
+const handler = (req: NextApiRequest, res: NextApiResponse) => {
+  res.status(200).send(openApiDocument);
```
zomars, line 13: You can go to /v2/openapi.json to see all available endpoints.

zomars (commented):
Here are my findings so far.

Cons:
- When it doesn't work it's pretty hard to make it work
- We need to set output schemas for every single procedure
- It will be a big effort to migrate


Pros:
- When it works it works pretty well
- We already have many output procedures in API code
- It will be way less effort to maintain separate API code/shared logic

packages/features/bookings/schemas/bookingReadPublic.schema.ts
alannnc: I'm up for this, it will make testing even easier I think.

PR #9428, opened 2023-06-09
fix: error when updating event type availability

CarinaWolli:
## What does this PR do?

Fixes error when updating the availability in the event type settings

Error message:
Screenshot 2023-06-09 at 09 03 36

Screenshot 2023-06-09 at 09 04 19

Fixes #9418

## Type of change

### DIFF
--- a/apps/web/pages/event-types/[type]/index.tsx
+++ b/apps/web/pages/event-types/[type]/index.tsx
@@ -372,8 +372,9 @@ const EventTypePage = (props: EventTypeSetupProps) => {
       throw new Error(t("seats_and_no_show_fee_error"));
     }
 
+    const { availability, ...rest } = input;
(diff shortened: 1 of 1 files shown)


hariombalhara (approved):
LGTM!

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -372,8 +372,9 @@ const EventTypePage = (props: EventTypeSetupProps) => {
       throw new Error(t("seats_and_no_show_fee_error"));
     }
 
+    const { availability, ...rest } = input;
     updateMutation.mutate({
```
Comment (hariombalhara) [inline apps/web/pages/event-types/[type]/index.tsx:376]: As a followup, we should either ignore these warnings explicitly, or better we should do something like {availability:_availability, ...rest}

PR #9441, opened 2023-06-09
fix: Prevent invalid host assignment

emrysal:
## What does this PR do?

Adds extra checks while adding hosts.


packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
```diff
@@ -149,7 +175,23 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
     };
   }
 
-  if (hosts) {
+  if (teamId && hosts) {
+    // check if all hosts can be assigned (memberships that have accepted invite)
+    const memberships =
+      (await prisma?.membership.findMany({
```
roae, line 181:
```suggestion
      (await ctx.prisma?.membership.findMany({
```

roae (dismissed):
this is failing because it is not getting prisma correctly, leave a comment in the line where the error is

roae (approved):
Nice job @emrysal 🙏

PR #9445, opened 2023-06-10
chore: Update reject message language on cancel event

### LINKED ISSUE #9421: [CAL-1923] Opt-in Reject Message
When an opt-in meeting is "rejected" the pop-up says that it's been "canceled." It can only be canceled if it was confirmed, then canceled. 

The language needs some updating:

"This event has been rejected" "This event has been declined" are a couple of options. 

I also did not have the opportunity to put a reason. 

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1923](https://linear.app/calcom/issue/CAL-1923/opt-in-reject-message)</sub>

wasimTQ:
## What does this PR do?

This PR changes the language of event rejected message.

Fixes #9421

image

## Type of change

## How should this be tested?

- [ ] Create a event
- [ ] Book the event using a different user
- [ ] Cancel the event

## Mandatory Tasks

### DIFF
--- a/apps/web/public/static/locales/en/common.json
+++ b/apps/web/public/static/locales/en/common.json
@@ -1039,7 +1039,7 @@
   "event_location": "Event's location",
   "reschedule_optional": "Reason for rescheduling (optional)",
   "reschedule_placeholder": "Let others know why you need to reschedule",
-  "event_cancelled": "This event is cancelled",
+  "event_cancelled": "This event has been cancelled",
   "emailed_information_about_cancelled_event": "We emailed you and the other attendees to let them know.",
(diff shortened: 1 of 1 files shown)


wasimTQ:
@roae Can you approve the vercel deployment?

deepak-likes-code (approved):
Seems good

emrysal:
Thanks for this contribution; but looking at the issue this is fixing; this fix is incorrect. The connected issue is about implementing a completely different message; when an booking is "cancelled" but not confirmed yet it's not actually cancelled but rather declined. The correct fix implements a different message for this scenario only. Will close as the correct fix requires fundamentally different code.

PR #9449, opened 2023-06-10
feat: Organization support for event types page

leog:
## What does this PR do?

Enables seeing the correct content when on an organization as well as seeing the correct options for creating a new event type.

<kbd>image</kbd>

<kbd>image</kbd>

## Type of change

## How should this be tested?

When inside an organization, go to the app and go to event-types page. Being the owner you can create event-types for any team as well as managed event-types.

## Mandatory Tasks


apps/web/pages/event-types/index.tsx
```diff
@@ -737,7 +739,9 @@ const EventTypeListHeading = ({
         )}
         {profile?.slug && (
           <Link href={`${CAL_URL}/${profile.slug}`} className="text-subtle block text-xs">
-            {`${CAL_URL?.replace("https://", "")}/${profile.slug}`}
+            {orgSlug
+              ? `${orgSlug}.${subdomainSuffix()}/${profile.slug}`
+              : `${CAL_URL?.replace("https://", "")}/${profile.slug}`}
```
leog, line 744: Making sure we show the correct Event Type Profile heading URL

packages/features/eventtypes/components/CreateEventTypeDialog.tsx
```diff
@@ -136,6 +139,9 @@ export default function CreateEventTypeDialog({
   });
 
   const flags = useFlagMap();
+  const urlPrefix = orgBranding
+    ? `${orgBranding.slug}.${subdomainSuffix()}`
+    : process.env.NEXT_PUBLIC_WEBSITE_URL;
```
leog, line 144: Abstracting the calculation of the URL to be used in the dialog

packages/features/eventtypes/components/OrganizationEventTypeFilter.tsx
```diff
@@ -23,7 +23,7 @@ export const OrganizationEventTypeFilter = () => {
   const isNotEmpty = !!teams?.length;
 
   return status === "success" ? (
-    <AnimatedPopover text={dropdownTitle} popoverTriggerClassNames="mb-0">
+    <AnimatedPopover text={dropdownTitle} popoverTriggerClassNames="!mb-0">
```
leog, line 26: This was not being applied, needed to be !important

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,50 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
     },
   });
 
+  const teamMemberships = user.teams.map((membership) => ({
+    teamId: membership.team.id,
+    membershipRole: membership.role,
+  }));
```
leog, line 199: Abstracting the membership of each team to then query

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,50 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
-        .map(mapEventType)
-        .filter((evType) => evType.userId === null || evType.userId === ctx.user.id)
-        .filter((evType) =>
-          membership.role === MembershipRole.MEMBER ? evType.schedulingType !== SchedulingType.MANAGED : true
-        ),
-    }))
+    user.teams
+      .filter((mmship) => {
+        const metadata = teamMetadataSchema.parse(mmship.team.metadata);
+        return !metadata?.isOrganization;
```
leog, line 206: We don't want to show organizations as possible teams to create event-types, at least for now, organizations don't have event-types, in their profile sub-teams are shown.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,50 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
+      })
+      .map((membership) => {
+        const orgMembership = teamMemberships.find(
+          (teamM) => teamM.teamId === membership.team.parentId
+        )?.membershipRole;
```
leog, line 211: In case we are in the presence of a sub-team belonging to an organization, we get the membership of the user in the organization as the user may be an owner and that has precedence over sub-team membership.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,50 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
+              ? !membership.team.parentId
+                ? `/team`
+                : "" + membership.team.slug
+              : null,
+          },
+          metadata: {
+            membershipCount: membership.team.members.length,
+            readOnly:
+              membership.role ===
+              (membership.team.parentId ? orgMembership ?? MembershipRole.MEMBER : MembershipRole.MEMBER),
```
leog, line 228: Make sure we evaluate membership in the org to be of precedence over sub-team membership.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,50 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
+        )?.membershipRole;
+        return {
+          teamId: membership.team.id,
+          membershipRole: orgMembership ?? membership.role,
+          profile: {
+            name: membership.team.name,
+            image: `${CAL_URL}/team/${membership.team.slug}/avatar.png`,
+            slug: membership.team.slug
+              ? !membership.team.parentId
+                ? `/team`
```
leog, line 220: We only want to show `/team` in the team URL on an non-org context.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,63 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
+  const teamMemberships = user.teams.map((membership) => ({
+    teamId: membership.team.id,
+    membershipRole: membership.role,
+  }));
+
+  const compareMembership = (mship1: MembershipRole, mship2: MembershipRole) => {
+    const mshipToNumber = (mship: MembershipRole) =>
+      Object.keys(MembershipRole).findIndex((mmship) => mmship === mship);
+    return mshipToNumber(mship1) > mshipToNumber(mship2);
+  };
```
leog, line 205: Having a way to compare two memberships relying on the order in which they are in the array representation of the object type definition, which enumerates them in order.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,63 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
+      .map((membership) => {
+        const orgMembership = teamMemberships.find(
+          (teamM) => teamM.teamId === membership.team.parentId
+        )?.membershipRole;
+        return {
+          teamId: membership.team.id,
+          membershipRole:
+            orgMembership && compareMembership(orgMembership, membership.role)
+              ? orgMembership
+              : membership.role,
```
leog, line 223: If there is an org membership and it is of higher order than the team membership, we choose the org membership, otherwise the team membership. This is because an Owner or Admin of the organization should be an Owner or Admin of each sub-team.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -190,27 +193,63 @@ export const getByViewerHandler = async ({ ctx }: GetByViewerOptions) => {
...
+          },
+          metadata: {
+            membershipCount: membership.team.members.length,
+            readOnly:
+              membership.role ===
+              (membership.team.parentId
+                ? orgMembership && compareMembership(orgMembership, membership.role)
+                  ? orgMembership
+                  : MembershipRole.MEMBER
+                : MembershipRole.MEMBER),
```
leog, line 241: In order to define if an event-type is read-only, we need to see the org membership for the user, because in case the user is an admin or owner of the org, same membership should apply to each sub-team

packages/features/eventtypes/components/CreateEventTypeDialog.tsx
sean-brydon: NIT/Future work: 
We do this in a few places across the whole of ORGs

Do you think we should extract this? 

Instead of a specific `useOrgBranding` context in the other PR maybe we can just introduce a `useOrg` context that has things like this without needing to do this/calculate username url etc every time

sean-brydon: Created CAL-1936

sean-brydon (approved):

PR #9465, opened 2023-06-12
fix: Multiple inPerson Location edit modals reading first location

### LINKED ISSUE #9454: In Person Organizer Address State Persistence and Display Issue
### Issue Summary

When adding a new In person organizer address multiple times, its state is saved in the component. Even read back after saving.

### Steps to Reproduce

https://www.loom.com/share/58e3a4517fcc43668066c1c7a07c55d3

JaideepGuntupalli:
## What does this PR do?

Fixes #9454

Screen Recording 2023-06-12 at 14 49 17

## Type of change

## How should this be tested?

- Try to make multiple in-person addresses
- Now every in-person edit modals shows the actual data instead of the first data as the default value

## Mandatory Tasks


Udit-takkar (approved):
@JaideepGuntupalli code works fine 

Approving now but there's too much hardcoding to check for inPerson location type and address property so I'll refactor the code later to simplify it.

JaideepGuntupalli:
> @JaideepGuntupalli code works fine
> 
> Approving now but there's too much hardcoding to check for inPerson location type and address property so I'll refactor the code later to simplify it.

So the reason for it is there was no way to check which location the user is editing other than just plainly comparing the Value, but the existing types didn't help move that value between those components. So I tried to satisfy the typescript and not break any existing code, hence the resultant code. I think some high-level overhaul may help, but not sure. Thanks for the feedback

PR #9466, opened 2023-06-12
fix: translations - email verify and conferencing app 

sean-brydon:
CleanShot 2023-06-12 at 11 16 08@2x
CleanShot 2023-06-12 at 11 15 51@2x


alishaz-polymath (approved):
🚀

packages/emails/src/templates/VerifyAccountEmail.tsx
```diff
@@ -44,13 +44,13 @@ export const VerifyAccountEmail = (
...
           <>
-            {props.language("happy_scheduling")} <br />
+            {props.language("happy_scheduling")}, <br />
             <a
               href={`mailto:${SUPPORT_MAIL_ADDRESS}`}
               style={{ color: "#3E3E3E" }}
               target="_blank"
               rel="noreferrer">
-              <>{props.language("the_calcom_team", { companyName: COMPANY_NAME })}</>
+              <>{props.language("the_calcom_team", { companyName: SENDER_NAME })}</>
```
hariombalhara, line 53: Can this be a breaking change?(for self hosters). I mean what if this env variable isn't set?

sean-brydon: Sure i will push these changes ... very good eyes @hariombalhara

pumfleet (approved):
Amazing, thanks!

hariombalhara (approved):

PR #9484, opened 2023-06-13
fix: Pass event duration to schedule api call

JeroenReumkens:
## What does this PR do?

Ensures we fetch new timeslots when changing duration in new booker.

Fixes CAL-1940 

## Type of change

## How should this be tested?

- [ ] Choose a multi duration event, change duration, and notice the slots update. 
- [ ] Events with only a single duration should also still show the correct timeslots.

### DIFF
--- a/packages/features/bookings/Booker/utils/event.ts
+++ b/packages/features/bookings/Booker/utils/event.ts
@@ -37,8 +37,8 @@ export const useEvent = () => {
 export const useScheduleForEvent = ({ prefetchNextMonth }: { prefetchNextMonth?: boolean } = {}) => {
   const { timezone } = useTimePreferences();
   const event = useEvent();
-  const [username, eventSlug, month] = useBookerStore(
-    (state) => [state.username, state.eventSlug, state.month],
+  const [username, eventSlug, month, duration] = useBookerStore(
+    (state) => [state.username, state.eventSlug, state.month, state.selectedDuration],
     shallow
   );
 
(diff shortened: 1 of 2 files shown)


packages/features/bookings/Booker/utils/event.ts
```diff
@@ -37,8 +37,8 @@ export const useEvent = () => {
 export const useScheduleForEvent = ({ prefetchNextMonth }: { prefetchNextMonth?: boolean } = {}) => {
   const { timezone } = useTimePreferences();
   const event = useEvent();
-  const [username, eventSlug, month] = useBookerStore(
-    (state) => [state.username, state.eventSlug, state.month],
+  const [username, eventSlug, month, duration] = useBookerStore(
+    (state) => [state.username, state.eventSlug, state.month, state.selectedDuration],
```
JeroenReumkens, line 41: Here we fetch the duration from the booker store, and pass it a few lines down to the useSchedule hook

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -35,6 +37,7 @@ export const useSchedule = ({
       endTime: (prefetchNextMonth ? nextMonthDayjs : monthDayjs).endOf("month").toISOString(),
       timeZone: timezone!,
       eventTypeId: eventId!,
+      duration: duration ? `${duration}` : undefined,
```
JeroenReumkens, line 40: `useSchedule` hook didn't event take duration into account yet. So this was definitely a bug!

sean-brydon (approved):
havent tested but code LGTM 

Does duration have to be a string here?

PR #9496, opened 2023-06-13
feat: add activepieces app

abuaboud:
## What does this PR do?

 Our users already using cal-com on activepieces to automate their tasks, I would love if it can be added there.

## Type of change

## How should this be tested?

- Make sure the page is working


PeerRich:
as far as i can see, there is no app-related code added yet?

abuaboud:
@PeerRich There is no in-app code required, It's self serve from inside activpeieces, I replaced the link with cal-com specific page https://www.activepieces.com/pieces/cal-com. and here is the experience from inside app:

image

Does this make sense? any suggestion?

PR #9455, opened 2023-06-11
test: Add E2E test to verify theme change

abhisheksunil2201:
## What does this PR do?

This PR adds end-to-end tests to verify if the theme has been applied successfully to public booking pages.
- change-theme.e2e.ts
  - Verify if the theme change has been applied to the respective public booking page successfully
- ThemeLabel.tsx
  - Added data-testid to the theme label

Fixes #5977
/claim #5977

## Type of change

- Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- [x] Run automated e2e tests

## Mandatory Tasks
- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

(all complete)

PR #9501, opened 2023-06-14
fix: add floating button tests

### LINKED ISSUE #7233: [CAL-1120] Add Floating Button popup tests.
```
          Also, e2e were passing. It means that we don't have floatingButton popup tests. Need to add those.
```

*Originally posted by @hariombalhara in* [*https://github.com/calcom/cal.com/pull/7212#discussion_r1112183659*](https://github.com/calcom/cal.com/pull/7212#discussion_r1112183659)

[There are ](https://github.com/calcom/cal.com/tree/4be2eee2da2f39395ba7b448af168606ee825c99/packages/embeds/embed-core/playwright/tests)`inline` embed  and 'click on an element ' embed tests already. These would be very similar to those with just one difference that the tests would execute on floating button popup

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1120](https://linear.app/calcom/issue/CAL-1120/add-floating-button-popup-tests)</sub>

mitjans:
## What does this PR do?

Add tests for floating button embeds. Closes #7233
/claim #7233

## Type of change

- Add new tests

## How should this be tested?

- [x] should open dark embed iframe on floating button click
- [x] should open light embed iframe on floating button click

## Mandatory Tasks


packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
```diff
@@ -72,7 +69,7 @@ test.describe("Popup Tests", () => {
     if (!embedIframe) {
       throw new Error("Embed iframe not found");
     }
-    const { uid: bookingId } = await bookFirstEvent("free", embedIframe, page);
+    const { uid: bookingId } = await bookNthEvent("free", embedIframe, page, 2);
```
hariombalhara, line 72: Why do you need to change the event number being used? relying on event index per test this way is bound to cause issues as someone can start using the same index in another test.
We should stick to booking a default event(the first one) unless the test explicitly needs to test any other event.

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
```diff
@@ -40,14 +40,11 @@ async function bookFirstFreeUserEventThroughEmbed({
...
   }
-  const booking = await bookFirstEvent("free", embedIframe, page);
+  const booking = await bookNthEvent("free", embedIframe, page, 1);
   return booking;
 }
 
 test.describe("Popup Tests", () => {
-  test.afterEach(async () => {
-    await deleteAllBookingsByEmail("embed-user@example.com");
-  });
```
hariombalhara, line 50: We should not stop doing this cleanup. This is important.

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
```diff
@@ -101,9 +98,63 @@ test.describe("Popup Tests", () => {
     });
   });
 
-  todo("Floating Button Test with Dark Theme");
+  test("should open dark embed iframe on floating button click", async ({
```
hariombalhara, line 101: We should do a single test that tests end to end booking.
For testing themes, we should use a different strategy that can be as simple as verifying that the embed has appropriate class on html tag. Refer to https://github.com/calcom/cal.com/pull/9455/files.
This would make the tests faster than the current approach of doing booking e2e.

hariombalhara (changes requested):
@mitjans Thanks for the contribution 🙏 Left some comments.

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
mitjans: The problem I faced is with race conditions when booking at the same time with different tests. I wasn't able to find any other case in your tests where this was the case, so I wasn't sure on how to handle it.

The solution I came up with is hardcoding the index of the day to be booked, so that each test can book a different day.

This is why I also deleted the cleanup `afterEach`.

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
mitjans: The 2 tests added test e2e booking using floating buttons. When starting the development, I noticed there were 2 TODO tests to test floating buttons with dark/light theme (I thought these were the once that had to be added).

As I understand it then, only 1 e2e test is needed for the floating button, and then 2 more tests that only test the theme of the popup, is that right? (In that case, if I remember correctly, the race condition wouldn't be a problem so the Nth code could be deleted)

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
mitjans: Right now there is a test that is failing because of this. Both `"should open embed iframe on click - Configured with light theme"` and `"should be able to reschedule"` use the same `"/free"` account and the same slot.

The second one (`"should be able to reschedule"`) always fails because it is not able to place booking for that slot.

> This is currently happing in the main branch (not in this PR).

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
hariombalhara: Yeah exactly.

hariombalhara (approved):

PR #9514, opened 2023-06-14
feat: Upstash implementation for rate limiting/redis

sean-brydon:
Loom: https://www.loom.com/share/6da8ea93bcb0438e8beea89a65c0d169


apps/web/pages/api/auth/forgot-password.ts
```diff
@@ -37,10 +34,12 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
 
   // 10 requests per minute
 
-  try {
-    limiter.check(10, ip);
-  } catch (e) {
-    return res.status(429).json({ message: "Too Many Requests." });
+  const limmiter = await rateLimit({
```
zomars, line 37:
```suggestion
  const limiter = await rateLimit({
```

packages/lib/rateLimit.ts
```diff
@@ -1,27 +1,42 @@
...
+    console.log("Rate limit skipped - UPSTASH not set");
+    return { success: true };
+  }
+
+  if (isIpInBanListString(identifier)) {
+    return await limitter.forcedSlowMode.limit(identifier);
+  }
+
+  return await limitter[rateLimitingType].limit(identifier);
+}
```
sean-brydon, line 40: Always default to core but can optionally pass in your own limiter function if required. 

if youre in ip ban list we force slow mode

packages/lib/rateLimit.ts
```diff
@@ -1,27 +1,49 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
```
zomars:
```suggestion
```

zomars (approved):

sean-brydon:
@keithwillcode @zomars had to move to a factory style approach due to this redis instance throwing errors when the env vars were not set. This way we can keep the same API and devex if you want to swap it out for an instance of your own too (self hosted)

PR #9541, opened 2023-06-15
feat: Enable new booker globally for 3.0

sean-brydon:
- Enables new-booker with the following env variables. It enables New Booker for non-embed only.
```
NEW_BOOKER_ENABLED_FOR_NON_EMBED=1 
NEW_BOOKER_ENABLED_FOR_EMBED=0
```
- Toggles ON booker-layout feature by default

## TODO
- Test New Booker on Prod with embeds(using the /api/newbooker/enable endpoint)
- Test some of the known embeds in productions with the new booker and then set `NEW_BOOKER_ENABLED_FOR_EMBED=1`
- After this has been in production for some time. Remove old booker code.

CleanShot 2023-06-15 at 14 27 46@2x
CleanShot 2023-06-15 at 14 27 30@2x


PeerRich (approved):

hariombalhara (approved):

packages/prisma/migrations/20230524105015_added_newbooker_feature_flag/migration.sql
```diff
@@ -3,7 +3,7 @@ INSERT INTO
 VALUES
   (
     'booker-layouts',
-    false,
+    true,
```
zomars, line 6: We cannot and should not modify already applied migrations.

```suggestion
    false,
```

zomars (approved):
Ship it

Issue #9373, opened 2023-06-06
[CAL-1887] Make E2E test: Can add Apple Calendar


roae:
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1887](https://linear.app/calcom/issue/CAL-1887/make-e2e-test-can-add-apple-calendar)</sub>

kanakpandit17:
Is there anyway I can get access to the threads?

ologbonowiwi:
Hey @PeerRich! Implemented testing of the Apple Calendar authentication. Can you address someone to review it?

PR #9545, opened 2023-06-15
test: Added e2e test for apple calendar

Ninad3721:
## What does this PR do?

This PR adds e2e test to see the apple calendar can be added.
Fixes # (issue)
https://github.com/calcom/cal.com/issues/9373

## Type of change

## How should this be tested?

npx playwright test apple-calendar


PeerRich:
this test does not use credentials thereby does not test if installing works

Ninad3721:
> this test does not use credentials thereby does not test if installing works

I inserted my apple account id and password but it gives this error 
image

PR #9557, opened 2023-06-16
fix: New Booker Rewrites

hariombalhara:
Solve the problem using just next.config.js. It solves two problems
- Vercel doesn't seem to be working with middleware in the same way it works local and thus Rewrites don't work there
- Avoids the use of middleware, so we don't need to add /:path* matcher saving function invocation costs.

### DIFF
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -70,11 +70,14 @@ const middleware: NextMiddleware = async (req) => {
 
   // Ensure that embed query param is there in when /embed is added.
   // query param is the way in which client side code knows that it is in embed mode.
-  if (url.pathname.endsWith("/embed") && typeof url.searchParams.get("embed") !== "string") {
-    console.log("url.pathname", url.pathname, url.searchParams.get("embed"));
-    url.searchParams.set("embed", "");
-    return NextResponse.redirect(url);
-  }
+
+  // NOTE: It is somehow causing infinite redirect for /free/30min/embed even though it set the embed param, it goes away
+  // The new logic doesn't really need query param to detect embed mode. It uses window.name as well. So, we should be fine disabling it for now.
+
+  // if (url.pathname.endsWith("/embed") && typeof url.searchParams.get("embed") !== "string") {
+  //   url.searchParams.set("embed", "");
+  //   return NextResponse.redirect(url);
+  // }
 
   // Don't 404 old routing_forms links
   if (url.pathname.startsWith("/apps/routing_forms")) {
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -68,9 +68,6 @@ const middleware: NextMiddleware = async (req) => {
     }
   }
 
-  // Ensure that embed query param is there in when /embed is added.
-  // query param is the way in which client side code knows that it is in embed mode.
-
   // Don't 404 old routing_forms links
   if (url.pathname.startsWith("/apps/routing_forms")) {
     url.pathname = url.pathname.replace("/apps/routing_forms", "/apps/routing-forms");
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -12,21 +12,10 @@ const middleware: NextMiddleware = async (req) => {
   const url = req.nextUrl;
   const requestHeaders = new Headers(req.headers);
   const { currentOrgDomain, isValidOrgDomain } = orgDomainConfig(req.headers.get("host") ?? "");
-  const isEmbedRequest = typeof url.searchParams.get("embed") === "string";
-
   /**
    * We are using env variable to toggle new-booker because using flags would be an unnecessary delay for booking pages
    * Also, we can't easily identify the booker page requests here(to just fetch the flags for those requests)
    */
-  // Enable New Booker for All but embed Requests
-  if (process.env.NEW_BOOKER_ENABLED_FOR_NON_EMBED === "1" && !isEmbedRequest) {
-    requestHeaders.set("new-booker-enabled", "1");
-  }
-
-  // Enable New Booker for Embed Requests
-  if (process.env.NEW_BOOKER_ENABLED_FOR_EMBED === "1" && isEmbedRequest) {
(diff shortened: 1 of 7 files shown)


apps/web/next.config.js
```diff
@@ -183,7 +183,12 @@ const nextConfig = {
     return config;
   },
   async rewrites() {
-    return [
+    // .* matches / as well(Note: *(i.e wildcard) doesn't match / but .*(i.e. RegExp) does)
+    // It would match /free/30min but not /bookings/upcoming because 'bookings' is an item in pages
+    // It would also not match /free/30min/embed because we are ensuring just two slashes
+    const userTypeRouteRegExp = `/:user((?!${pages.join("|")})[^/]*)/:type`;
```
hariombalhara, line 189: Using [^/]* ensures that we don't match something like /free/30min/embed

apps/web/next.config.js
```diff
@@ -219,68 +224,85 @@ const nextConfig = {
         source: "/cancel/:path*",
         destination: "/booking/:path*",
       },
+      // Keep cookie based booker enabled just in case we disable new-booker globally
+      ...[
```
hariombalhara, line 228: Nicely groups related rewrites.

apps/web/pages/[user]/[type].tsx
```diff
@@ -359,6 +359,7 @@ async function getDynamicGroupPageProps(context: GetServerSidePropsContext) {
 
 export async function getServerSideProps(context: GetServerSidePropsContext) {
   const { user: userParam } = paramsSchema.parse(context.query);
+  console.log("[user]/[type] - header", context.req.headers);
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:362]:
```suggestion
```

apps/web/pages/new-booker/[user]/[type].tsx
```diff
@@ -153,6 +153,8 @@ const paramsSchema = z.object({ type: z.string(), user: z.string() });
 // Booker page fetches a tiny bit of data server side, to determine early
 // whether the page should show an away state or dynamic booking not allowed.
 export const getServerSideProps = async (context: GetServerSidePropsContext) => {
+  console.log("[user]/[type] -new-booker - header", context.req.headers);
+
```
Comment (hariombalhara) [inline apps/web/pages/new-booker/[user]/[type].tsx:157]:
```suggestion
```

packages/embeds/embed-core/embed-iframe-init.ts
```diff
@@ -17,6 +20,8 @@ export default function EmbedInitIframe() {
       ? embedNameSpaceFromQueryParam
       : window.name.includes("cal-embed=")
       ? window.name.replace(/cal-embed=(.*)/, "$1").trim()
+      : hasEmbedPath
```
hariombalhara, line 23: Gives ability to derive default namespace, even if embed query param is not present

packages/embeds/embed-core/src/sdk-event.ts
```diff
@@ -2,9 +2,12 @@
...
  * This module is supposed to instantiate the SDK with appropriate namespace
  */
+import embedInit from "@calcom/embed-core/embed-iframe-init";
+
 import { SdkActionManager } from "./sdk-action-manager";
 
 export let sdkActionManager: SdkActionManager | null = null;
 if (typeof window !== "undefined") {
-  sdkActionManager = new SdkActionManager(new URL(document.URL).searchParams.get("embed"));
+  embedInit();
```
hariombalhara, line 11: Moved the init here so that window.getEmbedNamespace is defined for line 12.

packages/embeds/embed-core/src/sdk-event.ts
```diff
@@ -2,9 +2,12 @@
  * @file
  * This module is supposed to instantiate the SDK with appropriate namespace
  */
+import embedInit from "@calcom/embed-core/embed-iframe-init";
+
 import { SdkActionManager } from "./sdk-action-manager";
 
 export let sdkActionManager: SdkActionManager | null = null;
 if (typeof window !== "undefined") {
-  sdkActionManager = new SdkActionManager(new URL(document.URL).searchParams.get("embed"));
```
hariombalhara, line 9: Using getEmbedNamespace which has many ways to derive embed namespace reliably.

apps/web/middleware.ts
```diff
@@ -81,10 +70,14 @@ const middleware: NextMiddleware = async (req) => {
...
-    return NextResponse.redirect(url);
-  }
+
+  // NOTE: It is somehow causing infinite redirect for /free/30min/embed even though it set the embed param, it goes away
+  // The new logic doesn't really need query param to detect embed mode. It uses window.name as well. So, we should be fine disabling it for now.
+
+  // if (url.pathname.endsWith("/embed") && typeof url.searchParams.get("embed") !== "string") {
+  //   url.searchParams.set("embed", "");
+  //   return NextResponse.redirect(url);
+  // }
```
hariombalhara, line 80:
```suggestion
```

packages/embeds/embed-core/src/embed-iframe.ts
```diff
@@ -439,8 +438,7 @@ function keepParentInformedAboutDimensionChanges() {
 
 if (isBrowser) {
   log("Embed SDK loaded", { isEmbed: window?.isEmbed?.() || false });
-  // Exposes certain global variables/fns that are used by the app to get interface with the embed.
-  embedInit();
+
```
hariombalhara, line 441:
```suggestion
```

apps/web/next.config.js
emrysal: you could also do `/:user((?!${pages.join("|")})+)/:type` to loose the greediness

emrysal (approved):
Approved when test pass, seems to function

apps/web/next.config.js
hariombalhara: Did you mean something like .+**?** . I tried that as well even with lazy it would consume all the characters of /free/30min/embed. Because even if lazy it always tries to get a match in the end.

zomars:
Awesome job you two @hariombalhara @emrysal 🙏🏽

PR #9570, opened 2023-06-16
fix: New booker selected date issues

emrysal:
## What does this PR do?

* Fixes zustand deprecation warning
* Fixes setting selected date to current regardless of ?date param being passed.
* Fixes the selected bookingLayout not being applied from user profile settings.
* Disables revalidate call as the booker page is currently using getServerSideProps
* Fixes incorrect start day - which was based on the browsingDate rather than the start of the month. startOf("month") has some translation happening so I used date(1) to fix.

### DIFF
--- a/packages/features/calendars/DatePicker.tsx
+++ b/packages/features/calendars/DatePicker.tsx
@@ -109,7 +109,7 @@ const Days = ({
   nextMonthButton: () => void;
 }) => {
   // Create placeholder elements for empty days in first week
-  const weekdayOfFirst = browsingDate.day();
+  const weekdayOfFirst = browsingDate.date(1).day();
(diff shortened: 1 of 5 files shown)


packages/features/calendars/DatePicker.tsx
```diff
@@ -109,7 +109,7 @@ const Days = ({
   nextMonthButton: () => void;
 }) => {
   // Create placeholder elements for empty days in first week
-  const weekdayOfFirst = browsingDate.day();
+  const weekdayOfFirst = browsingDate.date(1).day();
```
alishaz-polymath, line 112: That's a good find. Super smart 🙌

alishaz-polymath (approved):
🚀

PR #9617, opened 2023-06-18
chore: Added unit tests for converting availability to string

keithwillcode:
## What does this PR do?

Adds unit tests for the `availabilityAsString` function.

## Type of change


alannnc (approved):
Looks good! As a part of testing a function I think we can also try to make it fail on purpose.

PR #9620, opened 2023-06-19
fix: Make layout of booker not overflow in safari

### LINKED ISSUE #9586: 🚨 Booking form buttons not visible in safari
### Issue Summary

When on the an event public booking form in Safari
The Back and Confirm buttons are not visible
They are rendered for a few milliseconds, but then are pushed way off screen
image
image

Works fine in Chromium browsers

### Steps to Reproduce

1. Use safari and navigate to an event page
2. Select a timeslot and wait for form to render
3. Check css for buttons

--
This only happens on the default booking view, not the new cool views

### Technical details
 Safari Version 16.5 (18615.2.9.11.4)

JeroenReumkens:
## What does this PR do?

* Tweaks to booker grid sizing to prevent buttons from overflow booker wrapper in Safari
* Also fixes another issue where in the weekview layout the overflow wasn't scrollable

Fixes #9586

CleanShot 2023-06-19 at 10 26 03@2x

## Type of change

## How should this be tested?

- [ ] Verify that the booker layout doesn't overflow without being scrollable in Safari. Test in other browsers as well.
- [ ] Verify mobile also still looks okay. 

## Mandatory Tasks

### DIFF
--- a/packages/features/bookings/Booker/Booker.tsx
+++ b/packages/features/bookings/Booker/Booker.tsx
@@ -139,12 +139,12 @@ const BookerComponent = ({
             <StickyOnDesktop
               key="meta"
               className={classNames(
-                "relative z-10 flex",
+                "relative z-10 flex [grid-area:meta]",
                 layout !== BookerLayouts.MONTH_VIEW && "sm:min-h-screen"
               )}>
               <BookerSection
                 area="meta"
-                className="max-w-screen flex w-full flex-col md:w-[var(--booker-meta-width)]">
+                className="max-w-screen flex h-full w-full flex-col md:w-[var(--booker-meta-width)]">
                 <EventMeta />
                 {layout !== BookerLayouts.MONTH_VIEW &&
                   !(layout === "mobile" && bookerState === "booking") && (
--- a/packages/features/bookings/Booker/components/AvailableTimeSlots.tsx
+++ b/packages/features/bookings/Booker/components/AvailableTimeSlots.tsx
@@ -73,7 +73,7 @@ export const AvailableTimeSlots = ({ extraDays, limitHeight, seatsPerTimeslot }:
     <div
       className={classNames(
         limitHeight && "flex-grow md:h-[400px]",
-        !limitHeight && "flex w-full flex-row gap-4"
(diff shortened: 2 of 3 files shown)


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -139,12 +139,12 @@ const BookerComponent = ({
             <StickyOnDesktop
               key="meta"
               className={classNames(
-                "relative z-10 flex",
+                "relative z-10 flex [grid-area:meta]",
```
JeroenReumkens, line 142: This wrapper div was the cause of the week view not being scrollable, because the grid area was actually defined in the child. Setting that on the parent makes the grid function like it should.

packages/features/bookings/Booker/components/AvailableTimeSlots.tsx
```diff
@@ -73,7 +73,7 @@ export const AvailableTimeSlots = ({ extraDays, limitHeight, seatsPerTimeslot }:
     <div
       className={classNames(
         limitHeight && "flex-grow md:h-[400px]",
-        !limitHeight && "flex w-full flex-row gap-4"
+        !limitHeight && "flex h-full w-full flex-row gap-4"
```
JeroenReumkens, line 76: Extend busy day blocks full height in column view.

packages/features/bookings/Booker/config.ts
```diff
@@ -70,7 +70,7 @@ export const resizeAnimationConfig: ResizeAnimationConfig = {
       "meta main main"
       `,
       gridTemplateColumns: "var(--booker-meta-width) var(--booker-main-width)",
-      gridTemplateRows: "auto",
+      gridTemplateRows: "1fr 0fr",
```
JeroenReumkens, line 73: This fixes the safari issue. 0fr for second row, because the 2 rows show the exact same template areas. So content will still be visible. Having 2 rows is needed (even though they are the same), to make the animations animate to the proper places. They would space out if suddenly a second row got added.

sean-brydon (approved):
Man css sucks at times! Nice fix

PR #9629, opened 2023-06-19
feat: Add divider and set spacing on modals with the type=creation prop (feat-add-divider)

gitstart-calcom:
## What does this PR do?

- Add the prop customDividerClassName to customize the CSS Divider if needed
- Add a divider on all modals with type="creation" prop, following the [figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/1.-Cal-DS?type=design&node-id=29898%3A100590&t=62LvZCZAEZm5zDdw-1) design
- Some type="creation" modals seem not to be this type, we pushed all changes with screenshots, if you want us to revert the changes on some modals we will revert

image

- Inline Embed: 
image

- Enable Two-Factor Authentication:
image

-Enable Two-Factor Authentication (step 2):
image

-Enable Two-Factor Authentication (step 3):
image

- Disable Two-Factor Authentication:
image

- Delete Account:
image

- Update Timezone:
image

- Duplicate Event Type:
image

- Add a New Form:
image

- Change Team Member Role:
image

- Impersonate:
image

- Bulk Update Event Types:
image

- Connecting with MS Teams:
image

- Set a Default App Link:
image

- OIDC configuration:
image

- SAML Configuration:
image

- How to use booking questions as variables?
image
image

- Invite Link Settings:
image

- Add Action:
image

- Edit Keys:
245192220-c669e6fe-01ba-4a66-9b03-96ab4e8f1809

- Create an Api Key:
image

- API key created successfully:
image

- Invite Team Member:
image

- Add a Question:
image

- Add a new event type:
image


apps/web/components/Embed.tsx
```diff
@@ -1144,8 +1145,12 @@ const EmbedTypeCodeAndPreviewDialogContent = ({
                     />
                   )}
                 </div>
-                <div className={router.query.embedTabName == "embed-preview" ? "block" : "hidden"} />
-                <div className="mt-8 flex flex-row-reverse gap-x-2">
+                <div className={router.query.embedTabName == "embed-preview" ? "mt-2 block" : "hidden"} />
+                <DialogFooter
+                  className="mt-8 flex flex-row-reverse gap-x-2"
+                  showDivider
+                  customDividerClassNames="w-2/3">
```
gitstart-calcom, line 1152: In this specific case we had to implement this to the divider fill only the desired component space:
Without this prop:
244967944-bda8f054-f3da-4c30-b075-7d279a477396

packages/features/ee/api-keys/components/ApiKeyDialogForm.tsx
```diff
@@ -62,7 +62,7 @@ export default function ApiKeyDialogForm({
           </div>
           <div>
             <div className="flex">
-              <code className="bg-subtle text-default mb-2 w-full truncate rounded-md rounded-r-none py-[6px] pl-2 pr-2 align-middle font-mono">
+              <code className="bg-subtle text-default w-full truncate rounded-md rounded-r-none py-[6px] pl-2 pr-2 align-middle font-mono">
```
gitstart-calcom, line 65: The "mb-2" className was removed to fix this wrong input size
245198211-6b727cf9-5bc0-410e-bcf7-865918194c89

PeerRich:
what is the original issue? can you link it by commenting "fixes #number"

apps/web/components/Embed.tsx
Jaibles: The divider should not cross into the gray part on the left ideally

Jaibles:
Overall this looks great. The only thing that looks off to me is the spacing above and below the body.. ie how the body content is spaced from the header and the footer.

CleanShot 2023-06-20 at 13 55 49@2x

Jaibles:
The following modals have Inter as the header instead of Cal Sans
- Change Team Member Role:
- Add a New Form

The following modals have the primary & secondary buttons order reversed (primary should be on the right):
- Enable Two-Factor Authentication (all steps)
- Disable Two-Factor Authentication
- Delete Account
- Change Team Member Role
- Add a New Form

In designs, the button that sits alongside the primary button always use the minimal style. It appears here we're using secondary for Cancel and minimal for Close. I think we can just make those the same but no strong feelings on that one.

apps/web/components/Embed.tsx
gitstart-calcom: Ok we did this, the screenshot above is just to show why we need this prop we added 😄

gitstart-calcom:
Hi @Jaibles we made the changes here, please take a look 😄

sean-brydon (approved):
I've double checked against @Jaibles comments and not noticed anything wrong. Thanks @gitstart-calcom will approve and if we notice anything else will create a issue to follow up. 

Great work

sean-brydon:
@sean-brydon merged this pull request with [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/9629).

PR #9551, opened 2023-06-16
feat: insights orgs [CAL-1888]

alannnc:
## What does this PR do?

- Handles insights if user logged in is owner/admin inside team org.

Fixes # (issue)
Fixes https://github.com/calcom/cal.com/issues/9360

https://www.loom.com/share/d149cc7b8dfd436487f404d472c1840c

## Type of change

- New feature (non-breaking change which adds functionality)

## How should this be tested?

- [x] Start with a new fresh db.
- [x] Setup an org from /settings/organizations/new.
- [x] In your terminal navigate to your packages/prisma folder and run command yarn seed-insights.
- [x] Modify newly created insights team to be child of team org.
- [x] Navigate to your insights page as org owner/admin and use all the features.

## Mandatory Tasks
- [X] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works

PR #9631, opened 2023-06-19
fix: Orgs/create child teams CAL-1986

### LINKED ISSUE #9979: [CAL-2118] Page Crashes after creating a team
https://www.loom.com/share/6749d3f14fd14182b0fcec05f448136d

<sub>[CAL-2118](https://linear.app/calcom/issue/CAL-2118/page-crashes-after-creating-a-team)</sub>

joeauyeung:
## What does this PR do?

https://www.loom.com/share/f09ccd5533ab488cb32621a1a4fe590d

Fixes #9979

## Type of change

## How should this be tested?

- Create a new org
- Go to `settings/teams/new` and create a new team
    - In the DB, the new team should have the `parentId` of the org id.
- Go to `/teams`
    - You should see the create team prompts
    - In the DB change your role to "member"
    - The create team prompts should disappear
- Go to the settings page
    - As the admin you should see the create team button on the side
    - As a member that button should disappear
- Go to `settings/teams/new` as a member and try creating a team
    - You should receive an error
- Create a team as a normal user without an org 

## Mandatory Tasks

## Checklist


apps/web/pages/teams/index.tsx
```diff
@@ -11,19 +12,28 @@ import PageWrapper from "@components/PageWrapper";
...
       subtitle={t("create_manage_teams_collaborative")}
       CTA={
-        <Button
-          variant="fab"
-          StartIcon={Plus}
-          type="button"
-          href={`${WEBAPP_URL}/settings/teams/new?returnTo=${WEBAPP_URL}/teams`}>
-          {t("new")}
-        </Button>
+        !session.data?.user.organizationId ? (
```
joeauyeung, line 32: If there is no `organizationId` then assume it is a normal user.

apps/web/public/static/locales/en/common.json
```diff
@@ -1870,5 +1870,6 @@
   "set_up_your_profile": "Set up your profile",
   "set_up_your_profile_description": "Let people know who you are within {{orgName}}, and when they engage with your public link.",
   "sender_id_info": "Name or number shown as the sender of an SMS (some countries do not allow alphanumeric sender IDs)",
+  "org_admins_can_create_new_teams": "Only the admin of your organization can create new teams",
```
joeauyeung, line 1873: image

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -373,10 +373,25 @@ export const AUTH_OPTIONS: AuthOptions = {
...
+          const orgAdminQuery = await prisma.membership.findFirst({
+            where: {
+              teamId: existingUser.organizationId,
+              userId: existingUser.id,
+              OR: [{ role: MembershipRole.ADMIN }, { role: MembershipRole.OWNER }],
+            },
+          });
+
+          isOrgAdmin = !!orgAdminQuery;
+        }
```
sean-brydon, line 388: CC: @zomars is another one of them cases where this should live on me query instead of session? Due to revalidation and caching? I swear we had this talk before with `belongsToActiveTeam`

packages/trpc/server/routers/viewer/teams/create.handler.ts
```diff
@@ -16,11 +16,41 @@ type CreateOptions = {
...
+    if (!org) throw new TRPCError({ code: "NOT_FOUND" });
+
+    // Check if the user has permission to create a team under the org
+    if (
+      !org?.members.some(
+        (member) =>
+          (member.userId === user.id && member.role === MembershipRole.OWNER) ||
+          member.role === MembershipRole.ADMIN
+      )
+    ) {
```
sean-brydon, line 44: NIT: I think we can move some of this into the query itself ? Might make it a tiny bit faster

apps/web/pages/teams/index.tsx
```diff
@@ -11,19 +12,28 @@ import PageWrapper from "@components/PageWrapper";
...
+          <NewTeamButton />
+        ) : session.data?.user.isOrgAdmin ? (
+          <NewTeamButton />
+        ) : undefined
```
alannnc, line 36:
```suggestion
    (!session.data?.user.organizationId || session.data?.user.isOrgAdmin) && <NewTeamButton />
```
Single line, let's try to avoid ternaries when more than 2 cases.

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -373,10 +373,25 @@ export const AUTH_OPTIONS: AuthOptions = {
         const belongsToActiveTeam = checkIfUserBelongsToActiveTeam(existingUser);
         const { teams, ...existingUserWithoutTeamsField } = existingUser;
 
+        let isOrgAdmin;
+        // Check if the existingUser is an org admin
```
alannnc, line 377:
```suggestion
        // Check if the existingUser is an org admin/owner
```

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -439,6 +455,7 @@ export const AUTH_OPTIONS: AuthOptions = {
           impersonatedByUID: token.impersonatedByUID as number,
           belongsToActiveTeam: token?.belongsToActiveTeam as boolean,
           organizationId: token?.organizationId,
+          token: token?.isOrgAdmin,
```
alannnc, line 458: Not sure if we should return a token or a boolean here

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -458,6 +475,7 @@ export const AUTH_OPTIONS: AuthOptions = {
           impersonatedByUID: token.impersonatedByUID as number,
           belongsToActiveTeam: token?.belongsToActiveTeam as boolean,
           organizationId: token?.organizationId,
+          isOrgAdmin: token?.isOrgAdmin as boolean,
```
alannnc, line 478: we can use !! instead of as boolean?

packages/features/ee/teams/components/TeamsListing.tsx
```diff
@@ -102,16 +117,13 @@ export function TeamsListing() {
...
-              </Button>
-            </ButtonGroup>
-          </div>
+          !session.data?.user.organizationId ? (
+            <CreateTeamButtonGroup />
+          ) : session.data?.user.isOrgAdmin ? (
+            <CreateTeamButtonGroup />
+          ) : (
+            <p>Only organization admins can create new teams</p>
+          )
```
alannnc, line 126:
```suggestion
          !session.data?.user.organizationId || session.data?.user.isOrgAdmin
             ? <CreateTeamButtonGroup />
             : <p>Only organization admins can create new teams</p>
```

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -333,13 +344,11 @@ const SettingsSidebarContainer = ({
...
-                  href={`${WEBAPP_URL}/settings/teams/new`}
-                  textClassNames="px-3 items-center mt-2 text-emphasis font-medium text-sm"
-                  icon={Plus}
-                  disableChevron
-                />
+                {!session.data?.user.organizationId ? (
+                  <CreateTeamButton />
+                ) : session.data?.user.isOrgAdmin ? (
+                  <CreateTeamButton />
+                ) : null}
```
alannnc, line 351:
```suggestion
                {
                    (!session.data?.user.organizationId || session.data?.user.isOrgAdmin) && <CreateTeamButton />
                 }
```

packages/trpc/server/routers/viewer/teams/create.handler.ts
alannnc:
```suggestion
    // Look if userId is owner/admin of team org
    const isOwnerAdminOfOrg = await ctx.prisma.membership.findFirst({
      where: {
        userId: ctx.user.id,
        teamId: childTeam.parentId,
        role: {
          in: ["OWNER", "ADMIN"],
        },
        accepted: true,
      },
    });
```
Something I used on in trpc router middleware
https://github.com/calcom/cal.com/pull/9551/files#diff-37aef050083ed9d17c03e39a0361b3d7782783311f576e46e1620e2cb2f5ee8a

alannnc (dismissed):
Left some comments

packages/trpc/server/routers/viewer/teams/create.handler.ts
joeauyeung: Thinking about this more, I can send the `isOrgAdmin` property from the frontend, whether that's from the session or the `me` query to the endpoint.

packages/features/auth/lib/next-auth-options.ts
joeauyeung: The property of the token is a boolean. Should I cast the type as boolean like what's done above?

packages/features/auth/lib/next-auth-options.ts
zomars: More specifically add it from here: 

https://github.com/calcom/cal.com/blob/main/packages/trpc/server/middlewares/sessionMiddleware.ts#L74-L79

zomars (changes requested):
Nice work so far @joeauyeung Imma have to agree with @sean-brydon on this one. Let's avoid using session for this data. Left more comments here: https://github.com/calcom/cal.com/pull/9631#discussion_r1237524530

apps/web/pages/teams/index.tsx
```diff
@@ -1,41 +1,47 @@
...
-  return {
-    props: {
-      ...(await serverSideTranslations("en", ["common"])),
-    },
-  };
+export const getServerSideProps = async (context: GetServerSidePropsContext) => {
+  const ssr = await ssrInit(context);
+  await ssr.viewer.me.prefetch();
+
+  return { props: { trpcState: ssr.dehydrate(), ...(await serverSideTranslations("en", ["common"])) } };
```
joeauyeung, line 45: Replaced `useSession` with the new data fetching method.

packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -75,6 +76,12 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
         select: {
           slug: true,
           metadata: true,
+          members: {
+            where: {
+              userId: session.user.id,
+              OR: [{ role: MembershipRole.ADMIN }, { role: MembershipRole.OWNER }],
+            },
+          },
```
joeauyeung, line 84: @sean-brydon filtered in the Prisma query

packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -96,10 +103,17 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
   // This helps to prevent reaching the 4MB payload limit by avoiding base64 and instead passing the avatar url
   user.avatar = rawAvatar ? `${WEBAPP_URL}/${user.username}/avatar.png` : defaultAvatarSrc({ email });
   const locale = user?.locale || ctx.locale;
+
+  const isOrgAdmin = !!user.organization?.members.length;
+  // Want to reduce the amount of data being sent
+  if (isOrgAdmin && user.organization?.members) {
+    user.organization.members = [];
+  }
```
joeauyeung, line 111: To reduce the payload being sent to the front we don't have to return the members array. It's just the user's information so it would have been duplicated.

packages/trpc/server/routers/viewer/teams/create.handler.ts
```diff
@@ -16,11 +16,20 @@ type CreateOptions = {
...
 
 export const createHandler = async ({ ctx, input }: CreateOptions) => {
+  const { user } = ctx;
   const { slug, name, logo } = input;
+  const isOrgChildTeam = !!user.organizationId;
+
+  // For orgs we want to create teams under the org
+  if (user.organizationId && !user.organization.isOrgAdmin) {
+    throw new TRPCError({ code: "FORBIDDEN", message: "org_admins_can_create_new_teams" });
+  }
```
joeauyeung, line 26: Rather than having a long query to check if the user can create a team, just use the new context added.

joeauyeung (commented):
@zomars addressed your feedback by replacing `useSession` with the `me` query.

packages/trpc/server/routers/viewer/teams/create.handler.ts
sean-brydon: @joeauyeung we never want to rely on frontend validation for this - a user could always modify the request to say they are an admin. We want to always check this permission on the backend

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -333,13 +334,15 @@ const SettingsSidebarContainer = ({
...
-                />
+                {!user?.organizationId || user?.organization.isOrgAdmin ? (
+                  <VerticalTabItem
+                    name={t("add_a_team")}
+                    href={`${WEBAPP_URL}/settings/teams/new`}
+                    textClassNames="px-3 items-center mt-2 text-emphasis font-medium text-sm"
+                    icon={Plus}
+                    disableChevron
+                  />
+                ) : null}
```
alannnc, line 345:
```suggestion
                {!user?.organizationId || user?.organization.isOrgAdmin && (
                  <VerticalTabItem
                    name={t("add_a_team")}
                    href={`${WEBAPP_URL}/settings/teams/new`}
                    textClassNames="px-3 items-center mt-2 text-emphasis font-medium text-sm"
                    icon={Plus}
                    disableChevron
                  />
                )}
```
Wouldn't this be better?

apps/web/pages/teams/index.tsx
```diff
@@ -1,41 +1,48 @@
...
-        </Button>
+        !user.organizationId || user.organization.isOrgAdmin ? (
+          <Button
+            variant="fab"
+            StartIcon={Plus}
+            type="button"
+            href={`${WEBAPP_URL}/settings/teams/new?returnTo=${WEBAPP_URL}/teams`}>
+            {t("new")}
+          </Button>
+        ) : undefined
```
alannnc, line 34:
```suggestion
        !user.organizationId || user.organization.isOrgAdmin && (
          <Button
            variant="fab"
            StartIcon={Plus}
            type="button"
            href={`${WEBAPP_URL}/settings/teams/new?returnTo=${WEBAPP_URL}/teams`}>
            {t("new")}
          </Button>
        )
```

alannnc: See if this works, I think CTA will expect something to be received.

packages/trpc/server/middlewares/sessionMiddleware.ts
zomars: Do we want full member data here?

packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -75,6 +76,12 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
         select: {
           slug: true,
           metadata: true,
+          members: {
+            where: {
```
zomars, line 80: To prevent querying all data. We only need to know if there are results or not.

```suggestion
						select: { userId: true },
            where: {
```

packages/trpc/server/middlewares/sessionMiddleware.ts
zomars: Can we delete instead of sending an empty array? This might cause unintended expectations

packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -96,10 +104,17 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
   // This helps to prevent reaching the 4MB payload limit by avoiding base64 and instead passing the avatar url
   user.avatar = rawAvatar ? `${WEBAPP_URL}/${user.username}/avatar.png` : defaultAvatarSrc({ email });
   const locale = user?.locale || ctx.locale;
+
+  const isOrgAdmin = !!user.organization?.members.length;
+  // Want to reduce the amount of data being sent
+  if (isOrgAdmin && user.organization?.members) {
+    user.organization.members = [];
```
zomars, line 111:
```suggestion
    delete user.organization.members;
```

zomars (changes requested):
Just noticing, why are we trying to merge this to the `feat/organizations-e2e` branch instead of `main`? 
image

EDIT: Tried to change the base branch but had to revert as too many unrelated changes. Looking into how to fix this.

zomars (dismissed):
Remade the PR with the correct base branch and re-applied only the changes from the original. Codewise LGTM. Made basic tests and seems to work fine. Also removed changes from SettignsLayout as @sean-brydon made changes in another PR that already hide the button.

leog (dismissed):
Works great, thanks for tackling this Joe. Would be awesome if you could incorporate what @sean-brydon worked on regarding being able to select pending members as part of the second step of the create team wizard, but can be worked on in a follow-up PR as Sean's PR is still open.

packages/trpc/server/middlewares/sessionMiddleware.ts
hariombalhara, line 1: I don't think we need to make any change in this file.

apps/web/pages/teams/index.tsx
hariombalhara: Screenshot 2023-07-07 at 11 52 51 AM
Don't think we need to call `serverSideTranslations` separately. ssrInit does that already.
```suggestion
export const getServerSideProps = async (context: GetServerSidePropsContext) => {
  const ssr = await ssrInit(context);
  await ssr.viewer.me.prefetch();

  return { props: { trpcState: ssr.dehydrate() } };
```

apps/web/pages/teams/index.tsx
```diff
@@ -1,41 +1,48 @@
...
-  return {
-    props: {
-      ...(await serverSideTranslations("en", ["common"])),
-    },
-  };
+export const getServerSideProps = async (context: GetServerSidePropsContext) => {
+  const ssr = await ssrInit(context);
+  await ssr.viewer.me.prefetch();
+
+  return { props: { trpcState: ssr.dehydrate(), ...(await serverSideTranslations("en", ["common"])) } };
```
hariombalhara, line 45:
```suggestion
  return { props: { trpcState: ssr.dehydrate() } };
```
Screenshot 2023-07-07 at 11 52 51 AM
Don't think we need to call `serverSideTranslations` separately. `ssrInit` does that already.

packages/trpc/server/middlewares/sessionMiddleware.ts
hariombalhara: It's needed as it's being used through ctx.

packages/trpc/server/routers/viewer/teams/create.handler.ts
```diff
@@ -17,11 +15,22 @@ type CreateOptions = {
...
+
   const currentOrgId = ctx.user.organization?.id;
+
   const slugCollisions = await prisma.team.findFirst({
     where: {
       slug: slug,
+      // If this is under an org, check that the team doesn't already exist
+      ...(isOrgChildTeam && { parentId: user.organizationId }),
       parentId: currentOrgId
```
hariombalhara, line 34: Isn't it doing the same thing that is at line 34. `currentOrgId` is equal to user.organization?.id

packages/trpc/server/routers/viewer/teams/create.handler.ts
```diff
@@ -51,25 +60,6 @@ export const createHandler = async ({ ctx, input }: CreateOptions) => {
...
-    }
-
-    const nameCollisions = await prisma.user.findFirst({
-      where: {
-        organizationId: ctx.user.organization.id,
-        username: slug,
-      },
-    });
-
-    if (nameCollisions) throw new TRPCError({ code: "BAD_REQUEST", message: "team_slug_exists_as_user" });
```
hariombalhara, line 68: Why is this code removed? Isn't it preventing the team slug to be same as user slug?

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -284,6 +284,9 @@ const SettingsSidebarContainer = ({
                 {teams &&
                   teamMenuState &&
                   teams.map((team, index: number) => {
+                    if (!teamMenuState[index]) {
```
hariombalhara, line 287: I made this change to fix a bug linked in PR description

hariombalhara (changes requested):
Have requested some changes @joeauyeung. Should be quick and we can merge after

packages/trpc/server/routers/viewer/teams/create.handler.ts
joeauyeung: There was an instance where @sean-brydon and I were working on the same thing, so had to clean up the file. I already talked to Sean about it.

hariombalhara (dismissed):

leog (approved):
LGTM

hariombalhara (approved):

PR #9632, opened 2023-06-19
fix: Creating teams for organizations in wizard

leog:
## What does this PR do?

When added the possibility to register an org checking if billing was enabled, we introduced a ramification where `requestedSlug` metadata property was not set but we relied on it to let the user create new teams. Expanding that check to cover the other scenario as well.

## Type of change

## How should this be tested?

Go to the org wizard and create teams, it should let you do it in a non-team-billing environment scenario.

## Mandatory Tasks


apps/web/public/static/locales/en/common.json
```diff
@@ -1889,5 +1889,6 @@
   "my_profile": "My Profile",
   "my_settings": "My Settings",
   "sender_id_info": "Name or number shown as the sender of an SMS (some countries do not allow alphanumeric sender IDs)",
+  "no_organization_slug": "There was an error creating teams for this organization. Missing URL slug.",
```
leog, line 1892: Adding a proper message for the error to not show coded to the end user

packages/features/ee/organizations/components/AddNewTeamsForm.tsx
```diff
@@ -50,7 +50,7 @@ export const AddNewTeamsForm = () => {
       }
     },
     onError: (error) => {
-      showToast(error.message, "error");
+      showToast(t(error.message), "error");
```
leog, line 53: Show the translated message to the end user, not the code

packages/trpc/server/routers/viewer/organizations/createTeams.handler.ts
```diff
@@ -18,10 +18,14 @@ type CreateTeamsOptions = {
 export const createTeamsHandler = async ({ ctx, input }: CreateTeamsOptions) => {
   const { teamNames, orgId } = input;
 
-  const organization = await prisma.team.findFirst({ where: { id: orgId }, select: { metadata: true } });
+  const organization = await prisma.team.findFirst({
+    where: { id: orgId },
+    select: { slug: true, metadata: true },
```
leog, line 23: Be sure to also get slug from the newly created org to check for it too

packages/trpc/server/routers/viewer/organizations/createTeams.handler.ts
```diff
@@ -18,10 +18,14 @@ type CreateTeamsOptions = {
...
+  });
   const metadata = teamMetadataSchema.parse(organization?.metadata);
 
-  if (!metadata?.requestedSlug) throw new TRPCError({ code: "BAD_REQUEST", message: "no_organization" });
+  if (!metadata?.requestedSlug || !organization?.slug)
```
leog, line 27: Not only rely on `requestedSlug` to show an error, when no team-billing is enabled, org `slug` is the one defined.

packages/trpc/server/routers/viewer/organizations/createTeams.handler.ts
```diff
@@ -18,10 +18,14 @@ type CreateTeamsOptions = {
...
+    throw new TRPCError({ code: "BAD_REQUEST", message: "no_organization_slug" });
```
leog, line 28: Making the error code more specific.

packages/trpc/server/routers/viewer/organizations/createTeams.handler.ts
```diff
@@ -18,10 +18,14 @@ type CreateTeamsOptions = {
...
 
   const userMembership = await prisma.membership.findFirst({
     where: {
```
emrysal, line 31: Not directly related to this PR, but this needs a check to confirm the membership isn't pending

emrysal (approved):
minor unrelated nit - looks good

packages/trpc/server/routers/viewer/organizations/createTeams.handler.ts
CarinaWolli: I am still getting the error, should that be && maybe?

emrysal: Sharp eyes :)

CarinaWolli (approved):
Works now :)

PR #9650, opened 2023-06-19
fix: old-behaviour-booking-cancel

alannnc:
## What does this PR do?

- Removes old behaviour for canceling booking that used to grab and try to cancel using only credentials
- adds try catch as wrapper for some async calls

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Book with a location type as Phone Number but also have any video credentials installed
- [ ] Try to cancel booking

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/features/bookings/lib/handleCancelBooking.ts
+++ b/packages/features/bookings/lib/handleCancelBooking.ts
@@ -278,18 +278,23 @@ async function handler(req: CustomRequest) {
       }
     }
 
-    await Promise.all(integrationsToDelete).then(async () => {
-      if (lastAttendee) {
-        await prisma.booking.update({
-          where: {
-            id: bookingToDelete.id,
-          },
-          data: {
-            status: BookingStatus.CANCELLED,
-          },
-        });
-      }
-    });
+    try {
+      await Promise.all(integrationsToDelete).then(async () => {
+        if (lastAttendee) {
+          await prisma.booking.update({
+            where: {
+              id: bookingToDelete.id,
+            },
+            data: {
+              status: BookingStatus.CANCELLED,
+            },
+          });
+        }
+      });
+    } catch (error) {
+      // Shouldn't stop code execution if integrations fail
+      // as integrations was already deleted
+    }
 
     const tAttendees = await getTranslation(attendee.locale ?? "en", "common");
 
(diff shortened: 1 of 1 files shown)


packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -278,18 +278,23 @@ async function handler(req: CustomRequest) {
...
+            data: {
+              status: BookingStatus.CANCELLED,
+            },
+          });
+        }
+      });
+    } catch (error) {
+      // Shouldn't stop code execution if integrations fail
+      // as integrations was already deleted
+    }
```
alannnc, line 297: If anything fails at least don't stop main execution

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -510,14 +515,6 @@ async function handler(req: CustomRequest) {
         apiDeletes.push(deleteMeeting(videoCredential, uid));
       }
     }
-    // For bookings made before this refactor we go through the old behaviour of running through each video credential
```
alannnc, line 513: This change has over 8 months so I say it's good to say removing it will be the right thing to do!

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -626,7 +623,11 @@ async function handler(req: CustomRequest) {
     });
 
     // We skip the deletion of the event, because that would also delete the payment reference, which we should keep
-    await apiDeletes;
+    try {
+      await apiDeletes;
+    } catch (error) {
+      console.error("Error deleting event", error);
+    }
```
alannnc, line 630: wrapper try catch

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -626,7 +623,11 @@ async function handler(req: CustomRequest) {
     });
 
     // We skip the deletion of the event, because that would also delete the payment reference, which we should keep
-    await apiDeletes;
+    try {
+      await apiDeletes;
+    } catch (error) {
+      console.error("Error deleting event", error);
```
sean-brydon, line 629: WOuld it be better to use logger in all these instances of console.error?

emrysal (approved):

emrysal:
Nicely spotted 👏

packages/features/bookings/lib/handleCancelBooking.ts
alannnc: It depends on what are we trying to achieve with logger. If we use a report tooling, it should take care wether we use console.error or logger IMO.

PR #9654, opened 2023-06-20
fix: Booker fix for team + user with same event + event slug

### LINKED ISSUE #8579: [CAL-1606] New Booker: Shows team event in place of user event if both have same slugs.
Screenshot 2023-04-30 at 6 35 12 PM

When there are two events with the same slugs(One a Team event and other an individual event), getPublicEvent just fetches the first one it can get.

Screenshot 2023-04-30 at 6 38 16 PM
Screenshot 2023-04-30 at 6 38 28 PM

<sub>[CAL-1606](https://linear.app/calcom/issue/CAL-1606/new-booker-shows-team-event-in-place-of-user-event-if-both-have-same)</sub>

JeroenReumkens:
## What does this PR do?

If there would a user that was part of a team that had the exact same slug as the user, AND also both the team and the user have an event with te same slug, it would show the user event on the team page. 

This PR changes that behaviour. Now we have the option in the booker the force it the only fetch a team event. That boolean is passed on the team pages, which results in the query only fetching team events, instead of running an `OR` query and fetching the first match. 

Fixes #8579

## Type of change

## How should this be tested?

- [ ] Login with team pro, change their profile slug to `seeded`. Go to the seeded team page and change the slug to `seeded` too. Then create an event in both the team and the teampro-profile with a slug `test`. With the changes in this PR, it should simply show both event pages correctly.


apps/web/pages/new-booker/team/[slug]/[type].tsx
```diff
@@ -30,6 +30,7 @@ export default function Type({ slug, user, booking, away, isBrandingHidden }: Pa
         rescheduleBooking={booking}
         isAway={away}
         hideBranding={isBrandingHidden}
+        isTeamEvent
```
Comment (JeroenReumkens) [inline apps/web/pages/new-booker/team/[slug]/[type].tsx:33]: New boolean in booker that can now force it to only fetch team events.

packages/features/bookings/Booker/store.ts
```diff
@@ -207,9 +217,29 @@ export const useInitializeBookerStore = ({
   rescheduleUid = null,
   rescheduleBooking = null,
   layout,
+  isTeamEvent,
 }: StoreInitializeType) => {
   const initializeStore = useBookerStore((state) => state.initialize);
   useEffect(() => {
-    initializeStore({ username, eventSlug, month, eventId, rescheduleUid, rescheduleBooking, layout });
-  }, [initializeStore, username, eventSlug, month, eventId, rescheduleUid, rescheduleBooking, layout]);
+    initializeStore({
```
JeroenReumkens, line 224: Prettier did it's thing. The `isTeamEvent` boolean is added to the effect here.

packages/features/bookings/Booker/utils/event.ts
```diff
@@ -16,9 +16,10 @@ import { useBookerStore } from "../store";
  */
 export const useEvent = () => {
   const [username, eventSlug] = useBookerStore((state) => [state.username, state.eventSlug], shallow);
+  const isTeamEvent = useBookerStore((state) => state.isTeamEvent);
```
JeroenReumkens, line 19: the `useEvent` hook takes the `isTeamEvent` boolean from the store and automatically passes that to the query so it fetches the right data.

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -141,24 +146,34 @@ export const getPublicEvent = async (username: string, eventSlug: string, prisma
     };
   }
 
+  const usersOrTeamQuery = isTeamEvent
```
JeroenReumkens, line 149: If it's a team event we now only query for the team, instead of running the OR query. Left the OR in place for all other cases, so team is even then still used as a fallback (for example in case a user in the future will use this as atom, and forget to pass in the `isTeamEvent` boolean.

apps/web/pages/new-booker/team/[slug]/[type].tsx
Comment (hariombalhara) [reply apps/web/pages/new-booker/team/[slug]/[type].tsx:33]: We would probably have to do it for private links as well. A private link can be a team event or user event.

packages/features/eventtypes/lib/getPublicEvent.ts
JeroenReumkens: Yes you are right, I've just now reverted this decision to NOT always fallback to the team event. The `isTeamEvent` boolean is not required though. So if you don't pass it in, it defaults to a user event and won't fetch the team event as fallback anymore.

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -141,24 +146,25 @@ export const getPublicEvent = async (username: string, eventSlug: string, prisma
...
+        team: {
+          slug: username,
+        },
+      }
+    : {
+        users: {
+          some: {
+            username,
+          },
+        },
```
hariombalhara, line 160:
```suggestion
        },
        team: null,
        
```

hariombalhara: This is needed, otherwise it would also return team events as well and then again we can serve team event instead of user only events.

hariombalhara (changes requested):
Requested one change. Otherwise looks good.
Renamed **seeded-team** slug to **teampro** and then did the following tests
- Visited /new-booker/teampro/30min - Serves user event.
- Visited /new-booker/team/teampro/30min(after creating 30min event in Seeded Team). Serves team event.

packages/features/eventtypes/lib/getPublicEvent.ts
JeroenReumkens: Wow, this makes a lot of sense yeah! Good spot!

JeroenReumkens:
> Requested one change. Otherwise looks good. Renamed **seeded-team** slug to **teampro** and then did the following tests
> 
> * Visited /new-booker/teampro/30min - Serves user event.
> * Visited /new-booker/team/teampro/30min(after creating 30min event in Seeded Team). Serves team event.

Thanks a lot for testing! Just fixed your suggested changing. Re-requesting review 😁

hariombalhara (approved):

PR #9657, opened 2023-06-20
test: e2e for workflows

CarinaWolli:
## What does this PR do?

Adds first e2e test for workflows + adds some comments for more possible e2e tests.

The test covers: 
- creating a default reminder workflow that is set active on an event type
- booking the active event type should create a workflow reminder with the correct scheduled date


apps/web/playwright/workflow.e2e.ts
```diff
@@ -0,0 +1,82 @@
...
+      expect(workflowReminders[0].scheduledDate.toISOString()).toBe(scheduledDate.toISOString());
+    });
+
+    // add all other actions to this workflow and test if they triggered
+    // cancel booking and test if workflow reminders are deleted
+    // test all other triggers
+  });
+
+  /*
+    test.describe("Team Workflows", () => {
```
hariombalhara, line 73: You can use todo fn instead of commenting these out.

packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -32,6 +33,8 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     return;
   }
 
+  const sandboxMode = IS_PRODUCTION ? false : true;
```
hariombalhara, line 36: Do we want to do it for non-e2e usage as well ?

hariombalhara (changes requested):
LGTM !! Just have one concern before the merge. 
https://github.com/calcom/cal.com/pull/9657/files#r1268966281

packages/features/ee/workflows/api/scheduleEmailReminders.ts
hariombalhara: Yeah that would be the best way then

CarinaWolli:
@hariombalhara can you re-review please 🙏

hariombalhara (approved):

PR #9661, opened 2023-06-20
feat: Handle deleting vercel domain when org is deleted

sean-brydon:
This PR deletes a domain in the vercel project when an organization is deleted. 

Please test :)


leog (approved):
LGTM, thanks for covering this.

PR #9668, opened 2023-06-20
fix: eventType-getByViewer subselect on children->users[]

emrysal:
## What does this PR do?

Pretty straightforward - exclude unneeded data in children select. Minor fixup that crashed some pages with many children users do to 4mb limit


zomars (dismissed):

Udit-takkar:
@emrysal type checks are failing

Udit-takkar (approved):

PR #9679, opened 2023-06-21
fix: Get brand colors from DB for owner, so it's reflected in booker

JeroenReumkens:
## What does this PR do?

Brand colors weren't pulled from DB for owner, so it wasn't reflected in booker. This PR fixes that. 

CleanShot 2023-06-21 at 10 02 39@2x

## Type of change

## How should this be tested?

- [ ] Change the brand color in the appearance settings and note it being reflected in the booker.


sean-brydon (approved):
Fixed for me :)

PR #9684, opened 2023-06-21
feat: Added back to today button for new booker week and column views

JeroenReumkens:
## What does this PR do?

Adds 'today' button in case user moves weeks ahead/back in week and column view. This way the user can quickly get back to today. 

**Draft because: Waiting for approving of my design by @Jaibles** 

Fixes CAL-1971

https://github.com/calcom/cal.com/assets/2969573/c901acc6-7911-44d9-a5fd-375b8b2954a6

## Type of change

## How should this be tested?

- [ ] Open column / week view in new booker, navigate back end forth between weeks and notice the 'today' button appear. Clicking it should bring you back to today.


packages/features/bookings/Booker/store.ts
```diff
@@ -112,8 +112,16 @@ export const useBookerStore = create<BookerStore>((set, get) => ({
   },
   selectedDate: getQueryParam("date") || null,
   setSelectedDate: (selectedDate: string | null) => {
+    const currentSelection = dayjs(get().selectedDate);
+    const newSelection = dayjs(selectedDate);
     set({ selectedDate });
     updateQueryParam("date", selectedDate ?? "");
+
+    // Setting month make sure small calendar in fullscreen layouts also updates.
+    if (newSelection.month() !== currentSelection.month()) {
```
JeroenReumkens, line 121: Updating the month when selecting a date makes sure that the mini calendar in week/column view also changes it's month if the month is different.

ciaranha (dismissed):
Thanks @JeroenReumkens! Really wanted this. What do you think about using the minimal button for the "Today" since the chevrons have it too. I've also communicated the spacing that I think looks most balance with all things added.

CleanShot 2023-06-21 at 16 24 33@2x

JeroenReumkens:
> Thanks @JeroenReumkens! Really wanted this. What do you think about using the minimal button for the "Today" since the chevrons have it too. I've also communicated the spacing that I think looks most balance with all things added.
> 
> CleanShot 2023-06-21 at 16 24 33@2x

@Jaibles  The minimal button felt a little bit lost to me. I think you don't straight away recognise that it's a button? But if you like it I'm more than happy to change it 😄 

CleanShot 2023-06-22 at 08 11 58@2x

As for the margins; I'm able to change them too. The only thing that's not easy to change is the margin on the right side of the months. In another issue we discussed that when changing months the arrows jumped around because the length of the month changes. So then we decided to give the months a fixed width to prevent that. 
I might be able to remove just a little bit more margin from the right side to make it a bit tighter, but replicating the design is impossible without removing the fixed width and introducing the jumping of prev/next arrow again because of longer/shorter month titles. 

Let me know what you want with this too.

Jaibles:
> We then we decided to give the months a fixed width to prevent that. I might be able to remove just a little bit more margin from the right side to make it a bit tighter

Yea that makes sense, I did have a prototype where I experienced that issue but it feels like we might have just gone a little too wide to compensate. If of course, you've found from testing that it needs to be that far wide, leave it, but if it can be tightened up, do tighten :)

> The minimal button felt a little bit lost to me. I think you don't straight away recognise that it's a button? But if you like it I'm more than happy to change it 😄

I can definitely see your point there. I think if we want to go outline on "Today" though, we gotta go outline on the left/right. What about this approach using our [combined button group component](https://share.cleanshot.com/5Q2h2sFF)?

CleanShot 2023-06-22 at 09 30 40@2x

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -73,6 +80,14 @@ export function Header({
             aria-label="Next Day"
             onClick={() => addToSelectedDate(extraDays + 1)}
           />
+          {selectedDateMin3DaysDifference && (
+            <Button
+              className="capitalize"
+              color="secondary"
```
PeerRich, line 86:
```suggestion
              color="minimal"
```

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -73,6 +80,14 @@ export function Header({
             aria-label="Next Day"
             onClick={() => addToSelectedDate(extraDays + 1)}
           />
+          {selectedDateMin3DaysDifference && (
+            <Button
+              className="capitalize"
+              color="minimal"
```
PeerRich, line 86:
```suggestion
              color="secondary"
```

PeerRich (approved):

PeerRich (approved):
very fancy

PR #9499, opened 2023-06-14
fix: redirect to 404 page for invalid email verification token

iamr-kumar:
## What does this PR do?

Fixes #9469 

## Type of change

- Bug fix (non-breaking change which fixes an issue)
Screenshot 2023-06-19 at 12 01 55 AM

## How should this be tested?

 
- [x] Send email verification link. Copy paste the link in browser and modify the token.
- [x] Go to `/api/auth/verify-email?token={some-invalid-token}`

## Mandatory Tasks
- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #9703, opened 2023-06-21
revert: "fix: redirect to 404 page for invalid email verification token"

zomars:
Reverts calcom/cal.com#9499 as it is causing other 404 pages to show as invalid token when it's not the case.


emrysal (approved):
👍

PR #9705, opened 2023-06-22
chore: both users who change either auth or email are shown same logout screen

JaideepGuntupalli:
## What does this PR do?

both users who change either auth or email are shown same logout screen


JaideepGuntupalli:
hey @emrysal, can you pls review and merge this so i could merge yours into the actual pull request?

apps/web/pages/auth/logout.tsx
```diff
@@ -29,11 +30,17 @@ export function Logout(props: Props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [props.query?.survey]);
   useEffect(() => {
-    if (props.query?.changeEmail === "true") {
-      setChangeEmail(true);
+    if (props.query?.emailChange === "true") {
```
emrysal, line 33: you can actually use the query instead of useState here

emrysal (approved):
Approved, but we should aim to keep state down.

apps/web/pages/auth/logout.tsx
JaideepGuntupalli: Thanks for the feedback! i thought of the same but i thought why change it and followed the existing pattern. Will keep that in mind.

PR #9727, opened 2023-06-22
fix: Make sure we get slug to decide things

leog:
## What does this PR do?

Organization update handler was removing slug from everywhere in case the input didn't include it. Wrapping the existing logic just to when slug is present.

## Type of change

## How should this be tested?

When creating an organization, going through the about step (3rd) was removing the slug as the requested one, resulting in not being able to create teams as it checks the existence of a valid slug.

## Mandatory Tasks


packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -66,26 +66,27 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
-      requestedSlug: input.slug,
-    };
-  } else {
-    data.slug = input.slug;
-
-    // If we save slug, we don't need the requestedSlug anymore
-    const metadataParse = teamMetadataSchema.safeParse(prevOrganisation.metadata);
-    if (metadataParse.success) {
-      const { requestedSlug: _, ...cleanMetadata } = metadataParse.data || {};
+  if (input.slug) {
```
emrysal, line 69:
```suggestion

  const cleanRequestedSlug = () => {
    const metadataParse = teamMetadataSchema.safeParse(prevOrganisation.metadata);
    if (metadataParse.success) {
      const { requestedSlug: _, ...cleanMetadata } = metadataParse.data || {};
      data.metadata = {
        ...cleanMetadata,
      };
    }
  }

  const requestOrCleanRequestedSlug = (slug: string) => {
    if (IS_TEAM_BILLING_ENABLED || prevOrganisation.slug) {
      data.slug = input.slug;
      cleanRequestedSlug(); // it's possible IS_TEAM_BILLING_ENABLED has not always been turned on.
      return;
    }
    data.metadata.requestedSlug = input.slug,
  };
  if (input.slug) {
    requestOrCleanRequestedSlug(input.slug);
  }
```

packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -66,26 +66,27 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
+    if (
+      IS_TEAM_BILLING_ENABLED &&
+      /** If the team doesn't have a slug we can assume that it hasn't been published yet. */
+      !prevOrganisation.slug
+    ) {
+      // Save it on the metadata so we can use it later
       data.metadata = {
-        ...cleanMetadata,
```
emrysal, line 87: This doesn't look right to me.

packages/trpc/server/routers/viewer/organizations/update.handler.ts
emrysal: Just an idea, NIT

emrysal (dismissed):
L87 needs some feedback, not sure about it; left some nits

packages/trpc/server/routers/viewer/organizations/update.handler.ts
leog: I don't understand, that line was never there to begin with 🤔

leog: Maybe it's a GitHub glitch, because that line actually is in line 87, not after line 76. https://github.com/calcom/cal.com/blob/093ea1f35d1e2b922d332bc93b47e3275d6bb276/packages/trpc/server/routers/viewer/organizations/update.handler.ts#L87

emrysal: oh yeah it's line 77 - is that OK? because it is gone and loses all metadata except requestedSlug

zomars (approved):

packages/trpc/server/routers/viewer/organizations/update.handler.ts
zomars: Let's do this in a follow up

PR #9728, opened 2023-06-22
chore: Subdomain blacklist when creating org

leog:
## What does this PR do?

When creating an organization, we shouldn't let users pick reserved subdomains such as docs, design, etc. This PR introduces a blacklist of subdomains to check before creating an org.

## Type of change

## How should this be tested?

Any user shouldn't be able to pick reserved subdomains as valid organization when creating one.

## Mandatory Tasks


packages/lib/constants.ts
```diff
@@ -70,3 +70,33 @@ export const IS_STRIPE_ENABLED = !!(
...
+  "go",
+  "status",
+  "api",
+  "saml",
+  "www",
+  "matrix",
+  "developer",
+  "cal",
+  "my",
+  "team",
```
baileypumfleet, line 101:
```suggestion
  "team",
  "support",
  "security",
  "blog",
  "learn",
  "admin",
```

pumfleet (commented):
Nice! Just added some more suggestions

packages/lib/constants.ts
leog: Added, feel free to approve.

zomars (commented):
Should selfhosters modify this file if they don't want to blacklist?

Should we put these in a comma separated list in an env variable?

packages/trpc/server/routers/viewer/organizations/update.handler.ts
zomars, line 1: Discarding changes to prevent conflicts with #9727

zomars (dismissed):

zomars:
- Moved constant to a env
- Added fallback when envs are undefined 
- Tested locally
- Fixed types

zomars:
Envs already set on Vercel as well

zomars:
Fixed unit tests. (my bad)

zomars (approved):

PR #9729, opened 2023-06-22
feat: Auto submit 2FA if all inputs have a value

### LINKED ISSUE #9726: [CAL-2031] auto submit 2FA if all inputs have a value
once these 6 have been filled, auto submit the form

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2031](https://linear.app/calcom/issue/CAL-2031/auto-submit-2fa-if-all-inputs-have-a-value)</sub>

afzalsayed96:
## What does this PR do?

Fixes #9726

## Type of change

## How should this be tested?

https://github.com/calcom/cal.com/assets/14029371/7a497f96-e966-4135-b169-dce83f43b318

## Mandatory Tasks


apps/web/components/security/EnableTwoFactorModal.tsx
```diff
@@ -127,6 +128,17 @@ const EnableTwoFactorModal = ({ onEnable, onCancel }: EnableTwoFactorModalProps)
...
+
+  const totpCode = form.watch("totpCode");
+
+  // auto submit 2FA if all inputs have a value
+  useEffect(() => {
+    if (totpCode?.trim().length === 6) {
+      form.handleSubmit(handleEnableRef.current)();
+    }
+  }, [form, handleEnableRef, totpCode]);
+
```
PeerRich, line 141: lets reuse this instead of duplicating

packages/lib/hooks/useCallbackRef.ts
```diff
@@ -0,0 +1,14 @@
...
+export const useCallbackRef = <C>(callback: C) => {
+  const callbackRef = useRef(callback);
+  useIsomorphicLayoutEffect(() => {
+    callbackRef.current = callback;
+  });
+
+  return callbackRef;
+};
+
+export default useCallbackRef;
```
PeerRich, line 14: this really needed? I wonder if we can even just check if the last input has a value and then submit

afzalsayed96: in order to submit we need a reference to `handleEnable`. Since, it's a function it either needs to be wrapped in a `useCallback` or a ref to get stable/most recent instance of the function.

Without the reference we could do something like finding the `form` instance in dom and calling `submit()` which isn't idiomatic in react but would simplify the code, yes.

something like:

```js
useEffect(() => {
   if (digits.at(-1)) {
      inputRef.current.form.submit()
   }
}, [digits])
```

sohamdata:
won't it give error/result in an **infinite loop of post requests** (returning 400) if we enter the **WRONG 2FA** code (`totpCode`), as `totpCode?.trim().length` **will remain 6**.

we would have to manually press backspace to stop that (`totpCode?.trim().length !== 6`) to prevent that.

apps/web/components/settings/DisableTwoFactorModal.tsx
```diff
@@ -82,6 +82,7 @@ const DisableTwoFactorAuthModal = ({
               {...form.register("password")}
               className="border-default mt-1 block w-full rounded-md border px-3 py-2 text-sm focus:border-black focus:outline-none focus:ring-black"
             />
+
```
emrysal, line 85:
```suggestion
```

emrysal (approved):
EnableTwoFactorModal seems duplicated, but that's not within the scope of this PR - approving - nice work @afzalsayed96

PR #9730, opened 2023-06-22
fix: Public Link of pending members is now disabled

### LINKED ISSUE #9701: Public pages button are enabled for pending users
### Issue Summary

After inviting users for a team, they are in pending status and their username is still `null`. The `Public page` buttons for those pending users are also enabled. After clicking on that button, it returns a 404 page as there are no users called `null`

### Steps to Reproduce

1. Create an organization and set up a team
2. Invite users for that team
3. Confirm that they are in pending status
4. Click on the `View public page` button for any pending user

### Actual Results

- `View public page` buttons are enabled for pending users
- A 404 page is shown after clicking on the buttons as there is no `null` user

### Expected Results

- `View public page` buttons should be disabled for pending users

Note: Do you think that we should have "reserved" usernames? Or kind of a list with not allowed usernames like that

### Technical details

- Tested on Chrome (114.0.5735.134) and Firefox (114.0.2)

### Evidence

image

image

AnitKrJha:
## What does this PR do?

Public pages' button for pending members of a team is now disabled.

Fixes #9701 

image

## Type of change


AnitKrJha:
@PeerRich @hariombalhara 

I would like to request your review of the recent changes I made. 

Additionally, while going through the codebase, 
I noticed an issue with the implementation of the `DropdownItem` component.

It seems that when the `href` prop is passed to the `DropdownItem` component, the `disabled` prop does not properly disable the button. Upon further examination, I discovered that the type of element is determined based on whether a link has been passed or not. When a link is provided, the underlying element becomes an `a` HTML element. Consequently, the `disabled` prop has no effect as the `a` element does not have a `disabled` property.

I wanted to bring this to your attention and kindly ask you to look into it.

PeerRich (approved):

PR #9731, opened 2023-06-22
fix: Various errors & 500 error on duplicate url when editing

emrysal:
## What does this PR do?

Fixes 500 error in logs, return human readable error; as well as various other translation related error messages.


keithwillcode (approved):
Looks good. Tested saving all sorts of event type changes and tested changing slugs to already-existing event type slug. No more 500s. Error shown to the user saying it would be a duplicate.

PR #9735, opened 2023-06-23
feat: auto submit 2fa values

### LINKED ISSUE #9726: [CAL-2031] auto submit 2FA if all inputs have a value

G3root:
## What does this PR do?

Fixes #9726. /claim #9726.

- [x] use [react pin input hook](https://www.npmjs.com/package/react-pin-input-hook)  for 2fa pin , which is customizable and actively maintained.
- [x] auto submit when 2fa pin are filled

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Mandatory Tasks


PeerRich (commented):
> use [react pin input hook](https://www.npmjs.com/package/react-pin-input-hook) for 2fa pin , which is customizable and actively maintained.

is it really worth adding another package just for this input? I mean... its working

preferably we fix this issue without adding a new package. in my eye it would just be a "check if notEmpty then submit form"

sean-brydon (changes requested):
Lets not add another hook-we can achieve this functionality with what we have already 

we can useDigit and just create a useEffect if length === desired length then call submit

PR #9755, opened 2023-06-23
fix: Added conditional error message for admin password banner

Ninad3721:
## What does this PR do?
The admin password banner was giving error for 2FA even if it is activated. So this PR puts a condition on the error message. It will show error message regarding the 2FA only if 2FA is disabled and about 2FA and password length both if both don't meet the requirement.

Fixes # (issue)

## Type of change

## How should this be tested?

1)Run a full production build. The error does not occur in a dev environment.
2)Logon as an admin user.
3)Disable two factor authentication.
4)Logoff then logon again.
5)Banner message appears.


a0m0rajab:
Hey @Ninad3721 just to pass the first test, you need to remove the space before the dots in the title so it should look like this: 
`fix: Added conditional error message for admin password banner`

Ninad3721:
> Hey @Ninad3721 just to pass the first test, you need to remove the space before the dots in the title so it should look like this: `fix: Added conditional error message for admin password banner`

Thank you for correcting 😊

a0m0rajab:
You are welcome, as for the linter you might fix it with: `npm run format` and `npm run lint:fix`, the last two tests require authorization which is not on you. 

which is visible in the package.json file: 

https://github.com/calcom/cal.com/blob/0aae07ae2f9561fe414cbd1060e9765168bfbcc7/package.json#L49-L51

Ninad3721:
> You are welcome, as for the linter you might fix it with: `npm run format` and `npm run lint:fix`, the last two tests require authorization which is not on you.
> 
> which is visible in the package.json file:
> 
> https://github.com/calcom/cal.com/blob/0aae07ae2f9561fe414cbd1060e9765168bfbcc7/package.json#L49-L51

I always use yarn lint:fix and then run the lint test and get all the tests passed but still don't know  why it fails test in github.

packages/features/users/components/AdminPasswordBanner.tsx
```diff
@@ -3,17 +3,19 @@ import Link from "next/link";
 
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { TopBanner } from "@calcom/ui";
+import { trpc } from "@calcom/trpc/react";
 
 function AdminPasswordBanner() {
   const { t } = useLocale();
   const { data } = useSession();
+  const { data: user, isLoading } = trpc.viewer.me.useQuery();
```
Udit-takkar, line 11: @sean-brydon should we add twoFactorEnabled in useSession? It would prevent this extra request with a lot of properties that are not required

packages/features/users/components/AdminPasswordBanner.tsx
```diff
@@ -3,17 +3,19 @@ import Link from "next/link";
...
 
   if (data?.user.role !== "INACTIVE_ADMIN") return null;
 
   return (
     <>
       <TopBanner
-        text={t("invalid_admin_password", { user: data.user.username })}
+       text={user?.twoFactorEnabled ? t("invalid_admin_password", { user: data.user.username }) : t("invalid_admin_password_and_no_2FA", { user: data.user.username }) }
```
Udit-takkar, line 18: @Ninad3721 we displayed the message "You are admin but you do not have a password length of at least 15 characters or no 2FA yet" before but after this change if a user updates a password to 15 characters and 2FA disabled he would see the message 

"You are admin but you do not have a password length of at least 15 characters and no 2FA yet" which is not accurate

PR #9756, opened 2023-06-23
feat: update insights filters UI

Udit-takkar:
## What does this PR do?

Fixes https://linear.app/calcom/issue/CAL-1874/insights-filters

Screenshot 2023-06-23 at 8 48 23 PM

Screenshot 2023-06-23 at 8 48 38 PM

Screenshot 2023-06-23 at 8 48 55 PM

Working:- 

https://github.com/calcom/cal.com/assets/53316345/a5a287d3-c696-4158-811b-00d071b33d8b

**NOTE:-**  _I will do a follow up PR in which i will change the implementation of insights filters so that we can reuse components. For eg TeamsFilter component is used in /event-types, /bookings and /workflows so i would try to directly use it insights also.  This PR just updates the UI according to the figma design._

## Type of change

## How should this be tested?

- Go to /insights 
- Try all filter combination


packages/features/insights/server/trpc-router.ts
```diff
@@ -90,6 +90,7 @@ const UserSelect = {
   name: true,
   email: true,
   avatar: true,
```
Udit-takkar, line 92: @alannnc is this correct usage of avatar. i think fetching avatar caused some issues in event-types.

alannnc (approved):
Made some fixes and it's ready to go to feat-insights-orgs branch!

PR #9775, opened 2023-06-24
feat: add route load progress indicator

### LINKED ISSUE #9742: [CAL-2080] UI inconsistency relating to high loading time

### Issue Summary

Some portion of the management page, when clicked, does not show any loading indicator. It stays on the same page and redirects to the intended page about 10-15 seconds later. I was consistently able to reproduce this issue on the App page. So, I'll be sharing it here.

### Steps to Reproduce

1. Land on the Cal.com management page
2. From the left pane, click Apps. Nothing will happen for the first 10 seconds
3. After approx. 10 seconds, you will be redirected to the app page

**Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?**

This hampers user experience. Users sometimes click 3-4 times and wonder why it's not loading (this happened in my case as well). See the technical details section for more context. If the load time is high, the user should see some kind of indicator that his request is being processed. Such indicators should be shown in all sections of the app, as required.

### Technical details

- Google Chrome - Version 114.0.5735.134 (Official Build) (64-bit)
- A quick peek at the Network tab reveals that a `GET` call on `https://app.cal.com/_next/data/KaRZJ2CMJ0Uyuf72oF6lh/en/apps.json` takes approx **10-13 seconds** to complete on initial load.
- The given load time might vary depending on network quality, but such load times are unusual. In any case, the user should have some kind of indicator feedback.

<sub>[CAL-2080](https://linear.app/calcom/issue/CAL-2080/ui-inconsistency-relating-to-high-loading-time)</sub>

G3root:
## What does this PR do?

Fixes #9742

[Screencast 2023-06-24 23:08:04.webm](https://github.com/calcom/cal.com/assets/84864519/fe28f72b-c9ab-4f70-9c3b-4297898ae086)

## Type of change

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -24,6 +24,7 @@
     "yarn": "3.4.1"
   },
   "dependencies": {
+    "@badrap/bar-of-progress": "^0.2.2",
     "@boxyhq/saml-jackson": "1.8.2",
     "@calcom/app-store": "*",
     "@calcom/app-store-cli": "*",
--- a/apps/web/pages/_app.tsx
+++ b/apps/web/pages/_app.tsx
@@ -9,7 +9,7 @@ import "../styles/globals.css";
 
 const progress = new ProgressBar({
   size: 2,
-  color: "#38bdf8",
+  color: "var(--cal-brand)",
   className: "bar-of-progress",
   delay: 100,
 });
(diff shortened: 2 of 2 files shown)


PeerRich:
ideally we fix the page load first instead of adding more loaders. there are ways to achieve a lot faster loading so lets put our energy into that

also, we dont need a package for this, its a simple tailwind component

PR #9808, opened 2023-06-26
chore: Removing the calendar-cache because of perf issues

keithwillcode:
## What does this PR do?

Our implementation of the calendar caching mechanism is too slow (600ms+ to fetch) and it's producing too many serverless function invocations. We need to rethink the caching strategy altogether.

## Type of change

## How should this be tested?

- [x] The booking page has no errors
- [x] The performance of the slots.getSchedule endpoint is not affected worse than 600ms on average

## Mandatory Tasks


packages/lib/server/revalidateCalendarCache.ts
```diff
@@ -1,18 +0,0 @@
-import type { NextApiResponse } from "next";
-
-import dayjs from "@calcom/dayjs";
-
-export const revalidateCalendarCache = (
```
keithwillcode, line 5: @roae @emrysal @zomars Let me know if you all would prefer to leave this as an empty implementation that we can come back to and replace with something like Redis cache. I don't think we should leave dead code around and we can always reference this commit when rethinking caching but I'm open.

roae: I prefer to remove it, we should avoid dead code.

roae (dismissed):
LGTM.

## Test performed
- Test the booking page.
- Add google calendar.
- Test the booking page, response time increased due to the connected calendar request.
- Disconnect google calendar successfully
- Test the booking page, the response time decreased

roae (approved):
The calendar cache pages were not deleted, I already deleted them

PeerRich:
merged @keithwillcode @roae 

wanna test this on staging?

PR #9811, opened 2023-06-26
fix: phone number disappearing issue

### LINKED ISSUE #9682: [CAL-2009] the phone input not prefilling (disappearing)
the phone input not prefilling (disappearing) is a regression of this PR: [https://github.com/calcom/cal.com/pull/8705](https://github.com/calcom/cal.com/pull/8705)
It changes the country on page load based on your IP address. But as a result of that country changing, it also clears the input field.

A note for the person picking this up; Locally the prefill is working because the API (packages/trpc/server/routers/publicViewer/countryCode.handler.ts) is always returning US. If you change that manually to a different country, then you will be able to replicate the behavior.

[https://www.loom.com/share/2ea2002b854648fa9b854a36bdf1c5a5?sid=552dda3f-6fb4-43cb-8ed8-a7124dd9337c](https://www.loom.com/share/2ea2002b854648fa9b854a36bdf1c5a5?sid=552dda3f-6fb4-43cb-8ed8-a7124dd9337c)

Created via Threads. See full discussion: [https://threads.com/34482909231](https://threads.com/34482909231)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2009](https://linear.app/calcom/issue/CAL-2009/the-phone-input-not-prefilling-disappearing)</sub>

anmol242:
Fixes the pre-fill issue for Phone number field while booking event or rescheduling event

Fixes https://github.com/calcom/cal.com/issues/9682

## Before Fix
https://www.loom.com/share/a6024364e1c744bf9db82e4b62504514?sid=8bb91e81-8055-474a-948e-ae7bc07520ff

## After Fix
https://www.loom.com/share/6ee121cf728d4769a915d7ce2172795e?sid=72a86bcb-84cf-41d3-9ef5-e7303a21358a

## Type of change

## How should this be tested?

-  Create an event type with options Phone number field (By going to advanced settings)
-  Choose `Identifier` as `phone` or whatever you want to fill
-  Open the URL for that event in another browser or in incognito mode by pre-filling the phone number identifier in the URL
-  You will see now phone number will be pre-filled for you and it will not disappear.
-  After you have created schedule, then try to reschedule the same event. It should pre-fill the phone number in that case too 

## Mandatory Tasks


packages/ui/form/PhoneInput.tsx
```diff
@@ -18,11 +18,19 @@ export type PhoneInputProps = {
 };
 
 function BasePhoneInput({ name, className = "", onChange, ...rest }: PhoneInputProps) {
+  useEffect(() => {
+    if (!rest.value) {
+      return
+    }
+    const formattedValue = rest.value.trim().replace(/^\+?/, "+") 
+    onChange(formattedValue);
+  }, []);
```
anmol242, line 27: Have done this because it was giving `Invalid phone number` whenever I try to submit it. Other wise it can be handled via input fields too

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,11 +18,19 @@ export type PhoneInputProps = {
...
+
   const defaultCountry = useDefaultCountry();
   return (
     <PhoneInput
       {...rest}
-      country={defaultCountry}
+      country={!rest.value ? defaultCountry : undefined}
```
anmol242, line 33: Have done this because if I haven't done this then it shows the default US country instead of selecting the country via itself.

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,21 +18,27 @@ export type PhoneInputProps = {
 };
 
 function BasePhoneInput({ name, className = "", onChange, ...rest }: PhoneInputProps) {
+  useEffect(() => {
+    if (!rest.value) {
+      return;
+    }
+    const formattedValue = rest.value.trim().replace(/^\+?/, "+");
```
anmol242, line 25: Have done this because it was giving an Invalid phone number whenever I try to submit it without changing the value so separated this logic to add `+` on useEffect instead of onChange. Otherwise, it can be handled via input fields too

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,21 +18,27 @@ export type PhoneInputProps = {
...
       disableSearchIcon
       inputProps={{
         name: name,
         required: rest.required,
         placeholder: rest.placeholder,
       }}
-      onChange={(value) => {
-        onChange("+" + value);
-      }}
+      onChange={(value) => onChange(value)}
```
anmol242, line 41: No need for this logic now as this will be handled via `useEffect`

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,17 +18,34 @@ export type PhoneInputProps = {
...
+      countryCode = parsePhoneNumber(rest.value)?.country || "";
+    } catch {}
+  }
+
   const defaultCountry = useDefaultCountry();
   return (
     <PhoneInput
       {...rest}
-      country={defaultCountry}
+      country={rest.value && countryCode ? undefined : defaultCountry}
```
anmol242, line 41: Have to done this change to see its an valid country code in phone number other wise it was changing to US

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,17 +18,34 @@ export type PhoneInputProps = {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  let countryCode;
+  if (rest.value) {
+    try {
+      countryCode = parsePhoneNumber(rest.value)?.country || "";
+    } catch {}
+  }
+
```
anmol242, line 36: This can be removed if we do not want to make US as default country for invalid phone inputs

packages/ui/form/PhoneInput.tsx
```diff
@@ -18,11 +18,27 @@ export type PhoneInputProps = {
...
```
Udit-takkar, line 36: I don't think we need this

packages/ui/form/PhoneInput.tsx
Udit-takkar: this hooks gets the country based on user ip on prod when deployed on vercel

anmol242: Understood 👍 , that makes sense

packages/ui/form/PhoneInput.tsx
anmol242: @Udit-takkar I will remove this if you can answer the below query of mine. Suppose I am in US country right now, But I pre-filled phone number that is outside US ( for example : `India`) In that case even if according to phone number it was supposed to show India country flag on `Dropdown` it will show `US` flag. So is that ok for you? If not then we don't need to remove this logic.

See the attached video what I am trying to say. See I filled the Indian number but it selected US. Then I tried to edit it, it changed to India Flag. I think this will be wrong behaviour

https://github.com/calcom/cal.com/assets/37592944/a49aee06-498d-49df-92f0-b44b1b2339ee

anmol242:
@PeerRich or anyone from the team can I get a re-review on this PR?

packages/ui/form/PhoneInput.tsx
Udit-takkar: No it should show the correct flag.

anmol242: So, @Udit-takkar we are okay to keep this logic or you want me to remove it?

Udit-takkar: you are not passing countryCode anywhere and i tried removing this logic and it still works fine with the correct flag.

Screenshot 2023-07-01 at 4 36 02 PM

Udit-takkar (approved):
LGTM.


https://github.com/calcom/cal.com/assets/53316345/cda152b9-5dc8-4013-a953-1fbd560cd1ea

packages/ui/form/PhoneInput.tsx
anmol242: No @Udit-takkar that logic was actually checking if my pre filled phone number is correct than only replace it with undefined on country input, other wise keep the default country flag in the input. See now we have removed that logic this will happen for wrong phone number, User will see no flag on the dropdown. Is that ok?

https://github.com/calcom/cal.com/assets/37592944/1ff5064e-93e4-44ff-8af1-9003a1b4b518

Udit-takkar: yeah that's fine for wrong number

anmol242: Video:  When we don't remove that logic. User will still see the default country flag even with wrong phone number.

https://github.com/calcom/cal.com/assets/37592944/172a59b2-6966-48af-be46-24388c59c477

anmol242: > yeah that's fine for wrong number

Aah ok @Udit-takkar, then that's fine and sorry if I sounded harsh I am not that good with english 😄

Udit-takkar: > Aah ok @Udit-takkar, then that's fine and sorry if I sounded harsh I am not that good with english 😄

Not at all. Thanks for contribution

anmol242:
@CarinaWolli or anyone from team can we merge this?

PR #9818, opened 2023-06-27
wip: Navigation history support in the new booker

roae:
## What does this PR do?

Add support to the navigation history in the new booker.

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Mandatory Tasks

### DIFF
--- a/packages/features/bookings/Booker/components/DatePicker.tsx
+++ b/packages/features/bookings/Booker/components/DatePicker.tsx
@@ -1,26 +1,25 @@
-import { useRouter } from "next/router";
-
 import type { Dayjs } from "@calcom/dayjs";
 import dayjs from "@calcom/dayjs";
 import { default as DatePickerComponent } from "@calcom/features/calendars/DatePicker";
 import { useNonEmptyScheduleDays } from "@calcom/features/schedules";
 import { weekdayToWeekIndex } from "@calcom/lib/date-fns";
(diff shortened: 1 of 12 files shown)


packages/features/bookings/Booker/components/DatePicker.tsx
```diff
@@ -25,16 +29,30 @@ export const DatePicker = () => {
...
+        if (syncChange) {
+          if (date.isSame(dayjs(), "month")) {
+            url.searchParams.set("date", dayjs().format("YYYY-MM-DD"));
+          } else {
+            url.searchParams.set("date", date.format("YYYY-MM-DD"));
+          }
+        } else {
+          url.searchParams.delete("date");
+        }
+        router.push(url, url, { shallow: true });
```
sean-brydon, line 50: Should we not just move all this logic into the zustand store? so we keep all the onChange logic in one place?  Same goes for other places it has been changed

roae: It could be, but also I think that keeping the navigation status in Zustand is not necessary, the query params should be the state source.

sean-brydon: Yeah i agree with it being the source! Just think we should keep it in the store incase we need it elsewhere also. Looks good man nice job!

PR #9827, opened 2023-06-27
fix: fixes event types toLowerCase issue on save

### LINKED ISSUE #9790: Saving Event Type automatically converts slug to lowercase

### Issue Summary

I use the API to create Event Types and often my generated slugs are a mixture of lower and uppercase letters and numbers. This works fine unless I use the UI to edit the event type. When I save the the Event Type, the URL (slug) is automatically converted to lowercase.

### Steps to Reproduce

1. Create an Event Type through the API and set the slug to have a mixture of uppercase and lowercase characters. E.g. FGcfd788guJB
2. Click into the newly created event type in the cal.com UI and the click 'Save'.
3. The slug is automatically changed to all lowercase. E.g. fgcfd788gujb

This makes it difficult to maintain the connection between my application and the cal.com data.

### Technical details

- Node.js version

Pradumn27:
## What does this PR do?

Fixes #9790

https://github.com/calcom/cal.com/assets/47187878/714b0f83-0164-4483-939d-05ce53c44699

- I have updated the slugify function to accommodate for the slugs where we don't need to convert the slug to lowercase.
- Have done these updates to all the places where event type slug is being created or updated.

## Type of change

## How should this be tested?

- Create an event type with a slug containing capital letters
- The event type slug should not turn the slug toLowerCase in both the cases while creating and editing

## Mandatory Tasks


packages/lib/slugify.ts
```diff
@@ -1,6 +1,9 @@
-export const slugify = (str: string) => {
+export const slugify = (str: string, toLowerCase?: boolean) => {
+  if (toLowerCase !== false) {
+    str = str.toLocaleLowerCase(); // Convert to lowercase
```
Udit-takkar, line 3: why are you using toLocaleLowerCase() here instead of toLoweCase() ?

Udit-takkar (commented):
LGTM. Left one comment

packages/lib/slugify.ts
Pradumn27: ahh! was fiddling with another project requiring localizations that got mixed up here, have rectified it please review.

CarinaWolli (changes requested):
Requesting changes to block merge as we probably don't want to allow upper-case slugs. Event-type links should also be case-insensitive like user profile links.

Pradumn27:
> Requesting changes to block merge as we probably don't want to allow upper-case slugs. Event-type links should also be case-insensitive like user profile links.

@CarinaWolli thanks for the review but as per issue #9790 faced by the user, if we are not going to allow uppercase slugs then shouldn't the event types created by the API also lowercase the provided slug?

emrysal:
@Pradumn27 Thanks for this; We've had quite a lengthy internal discussion and this has lead to the follow up strategy described in https://github.com/calcom/cal.com/issues/9934 - so yes; entirely right; we'll be updating the API to reflect this strategy.

PR #9848, opened 2023-06-29
chore: Adds auto increment ID to members table

sean-brydon:
This PR adds auto increment Id to membership table.

This will allow cursor based pagination allowing us to make use of infiniteQuery inside the new users table. 

Just about every table we have uses ID so its a bit weird that this one didn't have it.

**Has migrations that need to be tested to ensure they do not drop any data. It did not locally just want to make sure**

**Check prod data to ensure no duplicates exist ** 

All memberships locally are still intact
CleanShot 2023-06-29 at 13 53 19@2x


packages/prisma/migrations/20230629124638_membership_autoinc_id/migration.sql
```diff
@@ -0,0 +1,14 @@
+/*
+  Warnings:
+
+  - The primary key for the `Membership` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - A unique constraint covering the columns `[userId,teamId]` on the table `Membership` will be added. If there are existing duplicate values, this will fail.
+
```
sean-brydon, line 6: There will not be any duplicates due to the fact this was used as the PK before hand.

keithwillcode (approved):
Looks good.

emrysal:
lgtm

PR #9849, opened 2023-06-29
chore: Adds createdAt and updatedAt to all missing models

sean-brydon:
Adds createdAt and updatedAt to all models it was missing in schema.

I dont see any reason why this wasnt added for all models as it is common best practice especially in SQL.

Note: I updatedAt has been added to models that may not get updated by the user they have all been set to optional.


sean-brydon:
Closing this as it probably requires more thought than adding it to all models @emrysal

PR #9873, opened 2023-07-01
fix: [CAL-2040] add spinner when bookerState is loading

### LINKED ISSUE #9758: [CAL-2040] Booker shows loading of monthly, if only view is weekly or column
We shouldn't see a flash of the monthly view is the only calendar available is monthly or column

[https://github.com/calcom/cal.com/assets/4536123/9726b5e4-dddd-4079-8af6-aa09e8d9ba07](https://github.com/calcom/cal.com/assets/4536123/9726b5e4-dddd-4079-8af6-aa09e8d9ba07)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2040](https://linear.app/calcom/issue/CAL-2040/booker-shows-loading-of-monthly-if-only-view-is-weekly-or-column)</sub>

afrieirham:
## What does this PR do?

Add loading spinner when bookerState is loading to avoid showing monthly view if the default is set to weekly or column.

Fixes #9758 

https://github.com/calcom/cal.com/assets/32460534/0a865096-5585-4d5f-b092-b506f8e44709

## Type of change

## How should this be tested?

- [x] Use user free@example.com and choose weekly or column view in appearance. Then visit booker's page.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/features/bookings/Booker/Booker.tsx
+++ b/packages/features/bookings/Booker/Booker.tsx
@@ -10,7 +10,6 @@ import classNames from "@calcom/lib/classNames";
 import useMediaQuery from "@calcom/lib/hooks/useMediaQuery";
 import { BookerLayouts, defaultBookerLayoutSettings } from "@calcom/prisma/zod-utils";
 
-import { Spinner } from "../../calendars/weeklyview/components/spinner/Spinner";
 import { AvailableTimeSlots } from "./components/AvailableTimeSlots";
 import { BookEventForm } from "./components/BookEventForm";
 import { BookFormAsModal } from "./components/BookEventForm/BookFormAsModal";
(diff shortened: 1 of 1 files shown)


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -132,6 +133,11 @@ const BookerComponent = ({
   };
 
   const shouldShowFormInDialog = shouldShowFormInDialogMap[layout];
+
+  if (bookerState === "loading") {
+    return <Spinner />;
```
CarinaWolli, line 138: I don't think we want to show a spinner. I would prefer to return `null` if `bookerState` is still loading. 

Let me know in case I am missing something and it was mentioned somewhere to use a spinner here (cc @Jaibles )

CarinaWolli (changes requested):

packages/features/bookings/Booker/Booker.tsx
afrieirham: updated


https://github.com/calcom/cal.com/assets/32460534/9931f492-bedb-4df6-8ff3-6cfdaf927d26

CarinaWolli (approved):
Looks good to me now. Thank you 🙏

PR #9875, opened 2023-07-01
fix: incomplete vercel api error message on org creation

### LINKED ISSUE #9874: Incomplete error message for domain clash while creating a new organization.

### Issue Summary
Incomplete error message for domain clash while creating a new organization.

### Steps to Reproduce

1. Go to `settings/new/organization`
2. Fill the first form with an organization email that is already registered
3. Fill name and URL
4. Submit and then enter the verification code.
5. `domain_taken_team` message is displayed instead of proper error message.

Screenshot 2023-07-02 at 2 55 08 AM

### Technical details

- Tested on Brave 1.52.129 on production

iamr-kumar:
## What does this PR do?

Fixes #9874 

## Type of change

## How should this be tested?

- [x] Create an organization with an org email that is already taken.

## Mandatory Tasks


packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -21,7 +21,7 @@ type CreateOptions = {
   input: TCreateInputSchema;
 };
 
-const vercelCreateDomain = async (domain: string) => {
+const vercelCreateDomain = async (domain: string): Promise<boolean> => {
```
sean-brydon, line 24:
```suggestion
const vercelCreateDomain = async (domain: string) => {
```
Lets just use the inferred type

sean-brydon (dismissed):
LGTM - lets just use inferred type in the function above

packages/trpc/server/routers/viewer/organizations/create.handler.ts
iamr-kumar: I made this change because I was getting warnings from the linter. But yes, as I have looked at many other places in the code, inferred type has been used. So I will make this change.

apps/web/public/static/locales/en/common.json
```diff
@@ -1479,6 +1479,8 @@
   "must_enter_team_name": "Must enter a team name",
   "team_url_required": "Must enter a team URL",
   "url_taken": "This URL is already taken",
+  "domain_taken_team": "Domain is already owned by another team but you can request delegation to access it.",
+  "domain_taken_project": "Domain is already being used by a different project.",
```
leog, line 1483: As these messages refer to Vercel projects and teams, I think it would be best to just have one single message here like `"problem_registering_domain": "There was a problem with registering the subdomain, please try again or contact an administrator"` and use it in both scenarios 🙏

leog (changes requested):
Thanks for your contribution @iamr-kumar. Left a comment, once you can tackle it, please resolve it and ask a re-review so it comes back to me.

leog (approved):
Great job, thanks for your contribution @iamr-kumar

PR #9880, opened 2023-07-02
fix: Routing form accepts blank name value

ParbhatKataria1:
This PR makes the "name" value mandatory in the Routing Form.

Fixes #9871 

/claim #9871

Screenshot 2023-07-02 161448

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

1. Install the routing form app
2. Click to add a new routing form
3. Click to add a new routing form
4. Type a single space
5. Click on the Continue button
6. Observe

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #9882, opened 2023-07-02
fix: made the name field required in ROUTING FORMS

### LINKED ISSUE #9871: Routing form accepts blank as its name
### Issue Summary

To have a name is something mandatory for the routing form, however, it accepts blank as a name.
 
### Steps to Reproduce

1. Install the routing form app
2. Click to add a new routing form
3. Type a single space
4. Click on the Continue button
5. Observe

### Actual Results

- It's accepting blank as a name

### Expected Results

- It should validate that there is no valid value

### Technical details

- Tested on Chrome (114.0.5735.134)

### Evidence

image

image

image

AnitKrJha:
## What does this PR do?

Makes the name field in the routing form compulsory

image

/claim #9871 
Fixes #9871 

## Type of change

`there was no need to add additional logic , as the forms name are private to the logged in user and has no role in creating the URLs for the same`


PeerRich (approved):
simplest solution so far. works for me!

PeerRich:
i think this is required: https://github.com/calcom/cal.com/pull/9880

PeerRich:
are you sure "required" is enough? someone pointed out the field already had requried

AnitKrJha:
> are you sure "required" is enough? someone pointed out the field already had requried

It is enough ! The form name provides no purpose in link creation ! It is there just for the creator of the form to identify them. 
I don't see a reason to pack in additional JavaScript! 

Also if we want to have atleast one character in the form field ! Let's just add pattern(regex) to the input field ! 

No custom logic!

PeerRich:
awesome this worked.

PeerRich (approved):

PeerRich:
@hariombalhara the routing form check is failing, i think because the check doesnt need a title

PeerRich:
@AnitKrJha would you be down to fix the test?

PeerRich:
https://github.com/calcom/cal.com/actions/runs/5445419924/jobs/9904917679

AnitKrJha:
@PeerRich you can just close this issue , When going through the tests , I realized i added the required attr to the form desc .
Title field was already required ! 


My bad !
That required was a blind spot for both of us 😂

AnitKrJha:
I don't think this issue needs to be further addressed!

PR #9888, opened 2023-07-02
chore: Button Component Tidy up

sean-brydon:
For some reason we were using this function to apply conditional styles. I think we can do without it. 

Also added the correct disabled state for destructive buttons. 

Note: Focus state isnt correctly emulated in storybook in some browsers. Please tab over them :D 
CleanShot 2023-07-02 at 20 22 24@2x

Design Spec 📐: https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/Cal-DS---Components?type=design&node-id=309-41600&mode=dev


PeerRich (dismissed):
good refactor. tested and looks fine!

PeerRich:
some types failing interestingly
image

sean-brydon:
> some types failing interestingly image

Investigating

PeerRich (approved):

pumfleet (approved):
Code looks good to me

pumfleet (commented):
Should we be worried about the failing UI check?

PR #9889, opened 2023-07-02
feat: [RFC] Draft - Audit Log Exploration

sean-brydon:
See Threads RFC: https://threads.com/34485285457?s=qi1iudhtXBfgVmFsCogk2E


sean-brydon:
Closing as we need to think about this - no point leaving it open

PR #9897, opened 2023-07-03
test: Create unit tests for react components in packages/ui/components/alert (tests-alert-component)

gitstart-calcom:
Here is the GitStart Ticket for this pull request: [feat:tests-alert-component](https://developers.gitstart.com/client/calcom/ticket/feat:tests-alert-component)

## What does this PR do?
This PR is part of a larger task, with the objective of creating unit tests for components present in the `packages/ui/components/` folder. In addition, in this PR we are setting up the `vitest` for this folder, to be used in the next PRs

## Demo

250400577-92218568-6b9b-404f-849b-45d492e3bf81


package.json
```diff
@@ -77,6 +77,7 @@
     "@deploysentinel/playwright": "^0.3.3",
     "@playwright/test": "^1.31.2",
     "@snaplet/copycat": "^0.3.0",
+    "@testing-library/jest-dom": "^5.16.5",
```
PeerRich, line 80: why are we adding a new testing library? We already have playwright

gitstart-calcom: Hi @PeerRich! We are using vitest to make the tests (a lib already installed on the project), we are just using the RTL to render the components. We can try to use only playwright for it but this lib is used to test pages instead only specifc components, but this can be more complicated

packages/ui/components/alert/alert.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    expect(screen.getByTestId("alert").classList.toString()).toContain("component-class");
+  });
+
+  test("Should render custom icon with class names", () => {
+    render(
+      <Alert severity="info" CustomIcon={UserIcon} customIconColor="icon-color" iconClassName="icon-class" />
+    );
+
+    expect(screen.getByTestId("custom-icon")?.classList.toString()).toContain("icon-color");
+    expect(screen.getByTestId("custom-icon")?.classList.toString()).toContain("icon-class");
```
keithwillcode, line 25: Asserting classes set feels like it will lead to brittle tests and doesn't give a ton of value. I would suggest we don't add tests like these.

keithwillcode (changes requested):
Overall, I like these since they test the inner logic around severity controlling what is rendered. 

In regards to asserting specific classes though, I'm in favor of not adding tests like those due to the brittle nature it brings to the test.

gitstart-calcom:
Hi @keithwillcode we made the requested changes now, thanks for the feedback! 😄

keithwillcode (approved):

PR #9899, opened 2023-07-03
fix: app crash issue on setting a greater recurring count than the ma…

### LINKED ISSUE #9606: App is crashing if user type a huge number of events

### Issue Summary

When we create a recurring event, there are 20 events as a limitation, however, users can set the number of events to have recurred. As this is not limited to 20, users can add huge numbers here and it can cause a crash.

### Steps to Reproduce

1. Create a recurring event with any number of events, limited to 20
2. In another tab, try to book for this event type
3. Set a day
4. Check that it's possible to have more than the limitation from the event type (20 in that case)
5. Type 999999999 as the number of events
6. Set a time
7. Check that the app will crash 
8. If refresh the page, it cost a lot of time to recover 
9. After recovery, it shows that there are 20805 more events, but if you try to mouse over that, it will crash again

### Technical details

- Google Chrome v114.0.5735.134 (Official Build) (64-bit)
- Tested on Version preview for that PR (https://cal-5luk9i0hk-cal.vercel.app/event-types)

image

image

Pradumn27:
## What does this PR do?

Fixes the app crash issue that happens when a user enters a recurring count greater than the one specified by the author. Have changed the behaviour to default to the author's max count on submit if the provided count is greater than that.

Fixes #9606

https://github.com/calcom/cal.com/assets/47187878/e52aabe3-4c42-4b39-835c-1eb622907cd7

## Type of change

## How should this be tested?

- Create a recurring event and specify a recurring count
- On another go to the link of the event
- In the recurring count manually enter a number greater than the author-specified count
- On submitting the count defaults to the author's specified maximum count

## Mandatory Tasks

### DIFF
--- a/apps/web/components/booking/pages/AvailabilityPage.tsx
+++ b/apps/web/components/booking/pages/AvailabilityPage.tsx
@@ -204,7 +204,13 @@ const AvailabilityPage = ({ profile, eventType, ...restProps }: Props) => {
                               isFullWidth={false}
                               className="me-2 inline w-16"
                               onChange={(event) => {
-                                setRecurringEventCount(parseInt(event?.target.value));
+                                const count =
+                                  eventType?.recurringEvent?.count &&
+                                  parseInt(event?.target.value) > eventType?.recurringEvent?.count
+                                    ? eventType.recurringEvent.count
+                                    : parseInt(event?.target.value);
+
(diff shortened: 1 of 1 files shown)


emrysal (approved):
Tested, works 🙏

PR #9903, opened 2023-07-04
fix: toLowerCase/slugify of booking page params

emrysal:
## What does this PR do?

Use slugify as a standarized way of handling event slugs/ usernames - this logic is used similarly whilst writing into the frontend and should be consistent.

We should future proof this with function parameters to prevent these from being incompatible on write/read. Ideally we'd use the slugify library to achieve this - we shouldn't have rolled our own here.

We need test coverage for this too; so we can move to a more robust system in the future or just decide never to change our slugify methods.

### DIFF
--- a/packages/lib/defaultEvents.ts
+++ b/packages/lib/defaultEvents.ts
@@ -2,6 +2,7 @@ import type { Prisma, Credential } from "@prisma/client";
 
 import { DailyLocationType } from "@calcom/app-store/locations";
 import { getBookingFieldsWithSystemFields } from "@calcom/features/bookings/lib/getBookingFields";
+import slugify from "@calcom/lib/slugify";
 import { PeriodType, SchedulingType } from "@calcom/prisma/enums";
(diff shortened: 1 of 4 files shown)


packages/lib/defaultEvents.ts
```diff
@@ -177,14 +178,8 @@ export const getUsernameList = (users: string | string[] | undefined): string[]
   // So, even though this code handles even if individual user is dynamic link, that isn't a possibility right now.
   users = arrayCast(users);
 
-  const allUsers = users.map((user) =>
-    user
```
emrysal, line 181: this looks like it was wrong, if you lowercase and replace " " with + before splitting you end up splitting too late. Split should happen first before spaces are changed into their replacement character;

zomars (commented):
Codewise seems like a nice improvement. Will test before approval. Nice job @emrysal 🙏🏽

zomars (approved):
Tested and it seems good to go 🙏🏽

PR #9919, opened 2023-07-04
fix: Few minor UI issues in Orgs

leog:
## What does this PR do?

Fixing a few minor UI issues in Orgs

Team avatars were not showing up in new event-type dropdown, now they do.
<kbd>image</kbd>

Insights section had an empty div with margins pushing the entire content down, now fixed.
<kbd>image</kbd>

New org wizard showed "undefined" in the slug when no slug from query parameter was present (to support new org 404 CTA)
<kbd>image</kbd>

## Type of change

## How should this be tested?

Look at the screenshots and see the described difference marked with green


sean-brydon (approved):
Tested - all the above issues appear to be fixed!

PR #9938, opened 2023-07-05
fix: premium-username-callback

### LINKED ISSUE #9932: bug: 500 when redirected back from checkout

### Issue Summary

When upgrading to a premium username, I got a 500 after the checkout was finished:

CleanShot 2023-07-04 at 22 54 40@2x

I still got the username though:

CleanShot 2023-07-04 at 23 02 37@2x

The upgrade was a sponsored one though, so not sure if that has something to do with it still going through?

### Steps to Reproduce

1. Go to account page
2. Enter a premium username you want to upgrade to
3. Go to the billing portal
4. Enter the sponsored discount code, and your card details
5. See the screenshot posted above when redirected back to Cal.

alannnc:
## What does this PR do?

- Adds full url to stripe callbackUrl

Fixes #9932 

## Type of change

## How should this be tested?

- [ ] Setup your project constants so your cal.com instance run as a production and not a self host so premium usernames are available
- [ ] Setup website api to be listening
- [ ] Try to update your username to something premium.
- [ ] Do the payment checkout flow
- [ ] Callback url should be working now

## Mandatory Tasks

## Checklist


apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
```diff
@@ -116,7 +117,7 @@ const PremiumTextfield = (props: ICustomUsernameProps) => {
 
   const paymentLink = `/api/integrations/stripepayment/subscription?intentUsername=${
     inputUsernameValue || usernameFromStripe
-  }&action=${usernameChangeCondition}&callbackUrl=${router.asPath}`;
+  }&action=${usernameChangeCondition}&callbackUrl=${WEBAPP_URL}${router.asPath}`;
```
alannnc, line 120: Recent changes on next broke callbackUrl to be only path adding full url now should solve the issue.

packages/trpc/server/routers/loggedInViewer/stripeCustomer.handler.ts
```diff
@@ -30,12 +30,29 @@ export const stripeCustomerHandler = async ({ ctx }: StripeCustomerOptions) => {
...
+        id: userId,
+      },
+      data: {
+        metadata: {
+          ...metadata,
+          stripeCustomerId: customer.id,
+        },
+      },
+    });
+    stripeCustomerId = customer.id;
```
alannnc, line 52: Little improvement, in the case user doesn't have stripeCustomerId in user metadata it should create one for him.

apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
```diff
@@ -135,7 +135,7 @@ const UsernameTextfield = (props: ICustomUsernameProps & Partial<React.Component
           />
           {currentUsername !== inputUsernameValue && (
             <div className="absolute right-[2px] top-6 flex flex-row">
-              <span className={classNames("mx-2 py-2.5")}>
+              <span className={classNames("mx-2 py-3.5")}>
```
alannnc, line 138: Little visual fix

apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
alannnc: https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object

packages/trpc/server/routers/loggedInViewer/stripeCustomer.handler.ts
emrysal: In what situation would this happen?

apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
emrysal: How did you decide to implement within the frontend instead of backend?

alannnc: To be fair I don't remember but having url on server will mean that it can only handle one url, instead of having the UI component in different places and they should be on charge of their url to return, this means if component gets moved around places it will be itself in charge of having to provide return url.

packages/trpc/server/routers/loggedInViewer/stripeCustomer.handler.ts
alannnc: - An error during signup
- Not sure if getting invited would create you stripe customer id

emrysal (approved):

PR #9948, opened 2023-07-05
chore: enable adding mailhog and postgres port from env in docker compose

dammy001:
## What does this PR do?

This PR enables adding mailhog and postgres ports from `env` to prevent the "port in use" error in docker-compose

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Mandatory Tasks


keithwillcode (changes requested):
@dammy001 Can you revert the yarn.lock change please?

dammy001:
Okay @dammy001

nicktrn:
I'm not sure what issue this fixes, unless you already have other services running on those ports. But as they are non-standard that is not very likely.

Is the idea to set those vars and edit related ones in the root `.env` like this?

```sh
# .env (excerpt)

FORWARD_POSTGRES_PORT=1234
FORWARD_MAILHOG_PORT=2345
FORWARD_MAILHOG_DASHBOARD_PORT=3456

DATABASE_URL="postgresql://postgres:@localhost:${FORWARD_POSTGRES_PORT}/calendso"
EMAIL_SERVER_PORT=$FORWARD_MAILHOG_PORT
```

`docker compose` won't even know about these vars, so you'll have to either pass them directly to `yarn dx` or create `.env` files in both directories where the `docker-compose.yml` files are located.

Then again, maybe I'm missing something.

PR #9950, opened 2023-07-05
refactor: Eventtype app context

joeauyeung:
## What does this PR do?

Stacked off of #9337 

This PR refactors the `eventtypeAppContext` from an array to an object, in order to grab only what is needed.

Fixes # (issue)

## Type of change

## How should this be tested?

- Install event type apps
- Enable & edit the data to ensure they are being saved to the event type metadata
- Create a managed event type and enable some apps
    - In a child event type go to the apps page. Make sure they are enabled and not editable

## Mandatory Tasks

## Checklist


packages/app-store/EventTypeAppContext.tsx
```diff
@@ -5,13 +5,20 @@ export type GetAppData = (key: string) => unknown;
...
-  () => ({}),
-  undefined,
-  undefined,
-]);
+const EventTypeAppContext = React.createContext<AppContext>({
+  getAppData: () => ({}),
+  setAppData: () => ({}),
+  LockedIcon: undefined,
+  disabled: undefined,
+});
```
joeauyeung, line 21: I moved the context from an array into an object. Not every item was needed from the context, so with an object, we're able to call on only the things we needed rather than everything.

packages/app-store/_components/AppCard.tsx
```diff
@@ -16,24 +16,22 @@ export default function AppCard({
   switchOnClick,
   switchChecked,
   children,
-  setAppData,
```
joeauyeung, line 19: We can call the `setAppData` from the context, so we don't need to pass this down as a prop.

packages/app-store/_components/AppCard.tsx
```diff
@@ -16,24 +16,22 @@ export default function AppCard({
...
   returnTo,
   teamId,
-  disableSwitch,
-  LockedIcon,
+  appDataSchema,
```
joeauyeung, line 21: We do need to pass the `appDataSchema` though so the context gets the right types

packages/app-store/fathom/components/EventTypeAppCardInterface.tsx
```diff
@@ -4,19 +4,17 @@ import useIsAppEnabled from "@calcom/app-store/_utils/useIsAppEnabled";
 import type { EventTypeAppCardComponent } from "@calcom/app-store/types";
 import { TextField } from "@calcom/ui";
 
-import type { appDataSchema } from "../zod";
+import { appDataSchema } from "../zod";
 
 const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({ app, eventType }) {
-  const [getAppData, setAppData, LockedIcon, disabled] = useAppContextWithSchema<typeof appDataSchema>();
+  const { getAppData, setAppData, disabled } = useAppContextWithSchema<typeof appDataSchema>();
```
joeauyeung, line 10: This is an example where we don't need the `LockedIcon` so with an object we can call on what we do need. Also makes it easier if we have to add more data to the context in the future.

packages/app-store/qr_code/components/EventTypeAppCardInterface.tsx
```diff
@@ -7,11 +7,11 @@ import type { EventTypeAppCardComponent } from "@calcom/app-store/types";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { Tooltip, TextField } from "@calcom/ui";
 
-import type { appDataSchema } from "../zod";
+import { appDataSchema } from "../zod";
 
 const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({ eventType, app }) {
   const { t } = useLocale();
-  const [getAppData, setAppData, LockedIcon, disabled] = useAppContextWithSchema<typeof appDataSchema>();
+  const { disabled } = useAppContextWithSchema<typeof appDataSchema>();
```
joeauyeung, line 14: Another good example is where we only need the `disabled` prop from the context. Better than grabbing everything.

hariombalhara:
I will focus on this PR after v3.1 release.

hariombalhara (changes requested):
There are merge conflicts

PR #9967, opened 2023-07-06
fix: better slot starting times

CarinaWolli:
## What does this PR do?

Currently, we start the first slot always at the nearest 15 minutes. This is not ideal as for some duration other slot starting time make more sense. So with this PR the starting times are defined as follow: 

- Frequency is exact hours (60, 120, 180, ...), slot start time is a full hour
- Frequency is half hours (30, 90, ...), slot start time is half or full hours (8:00, 8:30, ...)
- Same with 20-minute events (20, 40, ...) and 10-minute events 
- Everything else will start at the nearest 15 min slot

It also fixes that slot times are shifted when there is a busy slot with a different duration. Here is a before and after of a 30-min event with a 5-minute busy slot at 1:00 pm

Before: 
Screenshot 2023-07-07 at 13 31 45

After
Screenshot 2023-07-07 at 13 34 01

#### 30 Minute events, availability starting at 7:15

Before: 
Screenshot 2023-07-06 at 12 40 00

After:
Screenshot 2023-07-06 at 12 40 42

## Type of change

## How should this be tested?

- Check if slot times are shown as described
- Test with different intervals/durations
- Test with busy times
- Test with different availabilities 

## Mandatory Tasks


packages/lib/slots.ts
```diff
@@ -160,19 +160,52 @@ function buildSlotsWithDateRanges({
...
-            .add(slotStartTime.hour() * 60 + Math.ceil(slotStartTime.minute() / 15) * 15, "minute")
-        : slotStartTime;
+    let previousStartTime;
+    // not sure here (we want to check if it is the same day in the organizer's timezone)
+    if (
+      slots.length &&
+      dayjs
+        .utc(range.start)
+        .add(range.start.utcOffset())
+        .isSame(dayjs.utc(slots[slots.length - 1].time).add(slots[slots.length - 1].time.utcOffset()), "day")
```
CarinaWolli, line 176:
```suggestion
 range.start.isSame(slots[slots.length - 1].time, "day")
```
@emrysal Just to make sure, this wouldn't work, right? I think `isSame' would compare in the system timezone but we want to check if it is the same day in the timezone of the organizer

packages/lib/slots.ts
```diff
@@ -160,19 +160,47 @@ function buildSlotsWithDateRanges({
...
+    ) {
+      previousStartTime = slots[slots.length - 1].time;
+    }
+
+    if (!previousStartTime) {
```
CarinaWolli, line 181: If we don't have a perviousStartTime, meaning there aren't any existing slots already on that day, then we need to find the start time of the first slot

packages/lib/slots.ts
```diff
@@ -160,19 +160,47 @@ function buildSlotsWithDateRanges({
...
+      let interval = 15;
+
+      const intervalsWithDefinedStartTimes = [60, 30, 20, 10];
+
+      for (let i = 0; i < intervalsWithDefinedStartTimes.length; i++) {
+        if (frequency % intervalsWithDefinedStartTimes[i] === 0) {
+          interval = intervalsWithDefinedStartTimes[i];
+          break;
```
CarinaWolli, line 189: If we find an interval that matches then we jump out of the loop as we always want the longest matching interval here. 
e.g. Otherwise, 60-minute frequencies would end up in an interval of 10

packages/lib/slots.ts
```diff
@@ -160,19 +160,47 @@ function buildSlotsWithDateRanges({
...
+        }
+      }
+
+      slotStartTime =
+        slotStartTime.utc().minute() % interval !== 0
+          ? slotStartTime
+              .startOf("hour")
+              .add(Math.ceil(slotStartTime.minute() / interval) * interval, "minute")
+          : slotStartTime;
+    } else {
```
CarinaWolli, line 199: In the else part, we make sure that the slot time matches the slots times that already exist on that day

packages/lib/slots.ts
emrysal: range start is various timezones, so wouldn't no

zomars (approved):
Tested and code LGTM. Nice work @CarinaWolli 🙏

zomars:
@zomars merged this pull request with [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/9967).

PR #9973, opened 2023-07-06
fix: Enable Salesforce & CRMs for teams & orgs [CAL-2116]

joeauyeung:
## What does this PR do?

This PR adds state that contains a teamId in order to install CRM apps to teams.

Fixes # (issue)

## Type of change

## How should this be tested?

- Create a team
- Install Salesforce
- Create a booking
- The new attendee should be created in Salesforce

## Mandatory Tasks

## Checklist


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -653,6 +653,16 @@ async function handler(
     eventType.schedulingType === SchedulingType.COLLECTIVE ||
     eventType.schedulingType === SchedulingType.ROUND_ROBIN;
 
+  const teamCredentials: Credential[] = [];
```
joeauyeung, line 656: We need to pass `teamCredentials` to the event manager in order to trigger any CRM apps

packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -79,7 +79,7 @@ export default class SalesforceCalendarService implements Calendar {
           .sobject("Contact")
           .create({
             FirstName,
-            LastName,
+            LastName: LastName || "",
```
joeauyeung, line 82: Requires a last name but need to fallback in case the attendee only supplies a first name

leog (dismissed):
Tested locally and it works, great job

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -837,6 +838,53 @@ async function handler(
...
+        credentials: true,
+      },
+    });
+    if (teamCredentialsQuery?.credentials) {
+      teamCredentials.push(...teamCredentialsQuery?.credentials);
+    }
+  }
+
+  // If the user is a part of an organization, query for the organization's credentials
+  if (organizerUser?.organizationId) {
```
hariombalhara, line 873: Maybe move it to a separate function and call it here. The handler function is too big already

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1115,7 +1163,7 @@ async function handler(
         },
       });
 
-      const credentials = await refreshCredentials(organizerUser.credentials);
+      const credentials = await refreshCredentials([...organizerUser.credentials, ...teamCredentials]);
```
hariombalhara, line 1166: Better to create a variable(maybe possibleCredentials) and use that everywhere.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1839,7 +1887,7 @@ async function handler(
   }
 
   // After polling videoBusyTimes, credentials might have been changed due to refreshment, so query them again.
-  const credentials = await refreshCredentials(organizerUser.credentials);
+  const credentials = await refreshCredentials([...organizerUser.credentials, ...teamCredentials]);
```
hariombalhara, line 1890: The logic downstream is going to pick the latest added credential first and thus there is no visibility and direct control on which credential would be picked. 

So, from a user perspective if there is Zoom app installed on Team, Org, and organizer level and the eventType for the organizer is booked, the meeting can go to any of the Team, Org, organizer connected accounts and it would keep on varying depending on which was installed most recently.


I think we should provide a way to be very clearly select which credential to use for an App for an EventType

hariombalhara (changes requested):
Have added some comments. Waiting on @Jaibles's input here and can be merged after.

hariombalhara (approved):
Merging this PR as it's mostly working as expected. I have some confusions that I have pinged @joeauyeung about but I don't want to block the merge as we have a huge timezone gap and that would slow down things.

PR #9987, opened 2023-07-07
feat: temp admin user management 

sean-brydon:
## What does this PR do?
Loom: https://www.loom.com/share/361ca6ca9b5748079cba59ed2f348d90

Not all items on edit from update the user. This is very temp page to help Milos

How to test:
Login as admin user
settings > admin > users
search - see filtered users
scroll to bottom of list - notice we refetch before you hit the bottom so you dont notice items loading
delete user -> user is deleted and removed from list (radix bug in main causes dialog to freeze refresh to make this go away)
edit user -> Edit users email/username/identityprovider. Not all items on this update work just yet. These will be picked up when we come to implement the proper user table.

### DIFF
--- a/packages/features/ee/users/components/UserForm.tsx
+++ b/packages/features/ee/users/components/UserForm.tsx
@@ -103,7 +103,7 @@ export const UserForm = ({
       identityProvider: {
         value: defaultValues?.identityProvider || identityProviderOptions[0].value,
         label:
-          identityProviderOptions.find((option) => option.value === defaultValues?.role)?.label ||
+          identityProviderOptions.find((option) => option.value === defaultValues?.identityProvider)?.label ||
           identityProviderOptions[0].label,
       },
     },
--- a/packages/features/ee/users/components/UsersTable.tsx
+++ b/packages/features/ee/users/components/UsersTable.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useRef, useMemo, useCallback, useEffect } from "react";
 
 import { trpc } from "@calcom/trpc/react";
 import { Badge, ConfirmationDialogContent, Dialog, DropdownActions, showToast, Table } from "@calcom/ui";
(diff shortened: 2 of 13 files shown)


packages/features/ee/users/components/UserForm.tsx
```diff
@@ -103,7 +103,7 @@ export const UserForm = ({
       identityProvider: {
         value: defaultValues?.identityProvider || identityProviderOptions[0].value,
         label:
-          identityProviderOptions.find((option) => option.value === defaultValues?.role)?.label ||
+          identityProviderOptions.find((option) => option.value === defaultValues?.identityProvider)?.label ||
```
sean-brydon, line 106: never pulled back the correct label

packages/features/ee/users/components/UsersTable.tsx
```diff
@@ -1,20 +1,52 @@
...
+  Avatar,
+} from "@calcom/ui";
+import { Edit, Trash, Lock } from "@calcom/ui/components/icon";
 
 import { withLicenseRequired } from "../../common/components/LicenseRequired";
 
-const { Body, Cell, ColumnTitle, Header, Row } = Table;
+const { Cell, ColumnTitle, Header, Row } = Table;
+
+const FETCH_LIMIT = 25;
```
sean-brydon, line 22: Fetch 25 as a base line - we can increase this if needed

packages/features/ee/users/components/UsersTable.tsx
```diff
@@ -1,20 +1,52 @@
...
+            pageParams: [],
+          };
+        }
+        return {
+          ...data,
+          pages: data.pages.map((page) => ({
+            ...page,
+            rows: page.rows.filter((row) => row.id !== userToDelete),
+          })),
+        };
```
sean-brydon, line 48: We manipulate the cache here to remove it instead of potentially fetching up to the current cursor. This could be like 50k users down the list. Not ideal.

packages/lib/hooks/useDebounce.tsx
```diff
@@ -0,0 +1,17 @@
...
+      setDebouncedValue(value);
+    }, delay);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [value, delay]);
+
+  return debouncedValue;
+}
```
sean-brydon, line 17: Lets try and stop using lodash to debounce values in react? Not sure how this is only being added now.

packages/trpc/server/routers/viewer/admin/listPaginated.handler.ts
```diff
@@ -0,0 +1,68 @@
...
+      timeZone: true,
+      role: true,
+    },
+  });
+
+  let nextCursor: typeof cursor | undefined = undefined;
+  if (users && users.length > limit) {
+    const nextItem = users.pop();
+    nextCursor = nextItem!.id;
+  }
```
sean-brydon, line 59: Pop so we can use it as the next cursor but not add it to the users array

hariombalhara (approved):
LGTM !! Tested changing the identity provider and password reset email only.

hariombalhara:
@hariombalhara merged this pull request with [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/9987).

zomars (commented):
LGTM @sean-brydon Nice work! 🙏🏽

PR #9989, opened 2023-07-07
chore: Cache i18n lookups in localStorage

### LINKED ISSUE #9840: [CAL-2070] reduce i18n skeletons
we have a lot of skeletons in place where we are loading i18n strings (translated words)

if we dont have these skeletons, it would show something like "event_types_title" for a brief second

if we do SSR for all pages, we'd wait for the string to load and then load the entire page, causing no skeletons, no string flickering but slower pages (we are bascially blocking the UI till all strings are loaded)

\_\_

is there no way to load the entire en/common.json once, and keep it somehow warm/around without constant skeletons?

ideally we get away without SSR

EDIT by @zomars:

Made a more detailed explanation of the desired solution here: [https://github.com/calcom/cal.com/issues/10198](https://github.com/calcom/cal.com/issues/10198)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2070](https://linear.app/calcom/issue/CAL-2070/reduce-i18n-skeletons)</sub>

rjackson:
## What does this PR do?

With this change we're add localStorage persistance of locale choice & translation strings, which means translated text will appear much sooner for users who have visited Cal.com before.

The original i18n request is still present, and the frontend & locale data is refreshed as soon as it finishes – so the appropriate language & latest translations will always come through to the frontend soon after – likely within 1 second. The localStorage persistence is more of a micro-optimisation to make the website _feel_ more responsive & smooth.

Because we're tracking the trpc endpoint response, invalidating it via `utils.viewer.public.i18n.invalidate()` still kicks off a new request and will update the locale info. This ensures that the latest translations come through when a user changes language, for example.

Fixes #9840
/claim #9840

## Type of change

## How should this be tested?

To test this locally, I introduced a false delay to `i18n.handler.ts`:

```
  await new Promise((resolve) => setTimeout(resolve, 5000));
```

- [x] Load the page from scratch with empty caches: No translated strings visible until i18n network response
- [x] Refresh the page: See translated text before i18n has finished loading
- In settings, change language
  - [x] See translated text change once i18n request has processed
  - [x] Refresh the page: See translated text, in correct language, before i18n has finished loading

## Mandatory Tasks

## Checklist

- I haven't ran e2e tests locally. Hopefully we'll see those run via Github Actions.


wesleymatosdev (commented):
It is not necessary early return when we already have the data? @rjackson 

Also, should we add any kind of  clean on the localStorage? When the language changes or after x time (to have updated translations)

rjackson:
@ologbonowiwi Nah, I'd recommend we leave the original trpc call even if we have the data in local storage. Main reasoning is as you say – handling language changes or updated translations.

For non-server-rended pages we'll always have to have _some_ trpc call to fetch the user's chosen locale, in case they've happened to change it on another browser or device. The existing call does that, but with the bonus of bringing across any up-to-date translations.

That does mean the resulting change is more of a micro optimisation, but it makes the app _feel_ much quicker & responsive. Or rather, it felt really slow & chunky without.

zomars:
@rjackson Thank you for your contribution. Just some questions regarding on the pursued approach.

Right now the `trpc.viewer.public.i18n` is an "uncache-able" state since it doesn't receive a desired locale to load as an input. Leaving us into the need of recurring of workaround as using local storage to cache.

Wouldn't it make more sense to refactor the actual query so it receives an actual input for desired locale to load. That way we can actually enable the server cache as we have here:

https://github.com/calcom/cal.com/blob/64147d3618812a3db7e7f9d80c4cb68cc0f62b16/packages/trpc/server/createNextApiHandler.ts#L61-L67

This would allow us to just take advantage of browsers caching as well without the need to implement our own invalidations/refreshing logic.

rjackson:
@zomars Yeah that's an approach I considered & prototyped, but ultimately threw away. Let me walk through what my thoughts were when I was exploring that approach:

On entirely static pages the frontend doesn't know what the user's locale is, so we're always going to need an API call to find that out.

Once we have that locale, we can then look up the locale files. In this case we don't even need a trpc endpoint, because they're already publicly accessible & should be appropriately cached by Vercel: https://cal.com/static/locales/en/common.json

```txt
$ curl -Ik https://cal.com/static/locales/en/common.json
HTTP/2 200
date: Mon, 17 Jul 2023 17:22:04 GMT
content-type: application/json; charset=utf-8
access-control-allow-origin: *
age: 7
cache-control: public, max-age=0, must-revalidate
cf-cache-status: DYNAMIC
content-disposition: inline; filename="common.json"
etag: W/"059face8fbc4c1f03c05979c4127e8c3"
nel: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
referrer-policy: strict-origin-when-cross-origin
report-to: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=6797o4AqERUxf4lvyjREjl%2Beg1TxBNf%2FOYNYaEnuwBBKfEA%2FkYx4UV8Zjo8tZqYc4rARRBZ2osCAwyhzX23ylUVAb37RXvmm1DFAZW9FQL2GHzVUNBA%2BDCCeUKaB"}],"group":"cf-nel","max_age":604800}
strict-transport-security: max-age=63072000
x-content-type-options: nosniff
x-matched-path: /static/locales/en/common.json
x-vercel-cache: HIT
x-vercel-id: lhr1:lhr1:lhr1::z56j8-1689614524612-1149d1329040
server: cloudflare
cf-ray: 7e842a3a9f95222a-MAN
```

We can configure next-i18next to load these files directly, but it introduces the flash of untranslated tokens and "Text content does not match server-rendered HTML" errors.

And this approach didn't actually help with the goal of this issue: It ended up being a whole lot of code changes & new problems to overcome, to still end up being blocked by an API call – and therefore customers still waiting for translations to load.

So instead of spending more time trying to make that work I thought if we're going to need an API call anyway, then what the current i18n endpoint does of sending by sending the translations alongside is actually perfectly fine. I just need a quicker way of loading that content if a customer has seen it before – which is what the Local Storage wrapper does.

That does introduce the issues of the Local Storage wrapper holding outdated content, especially if the user changes language – hence the decision to leave the i18n lookup in. In my mind the primary goal of that endpoint is to find out the user's current language, and secondarily send across the translation strings.

zomars:
@rjackson Fair points indeed. Although IMO it is preferable to default to browser locale first (we don't need an API call for this) Also for SSR pages we already know the user configured locale and can just serve the correct prefetched data. For SSG it is not a big of a problem. The fact that we need to refactor the endpoint in order for it to be cache-able still stands.

Will have to discuss with the team if we can accept this approach or try to solve this from the root instead of keep adding workarounds.

zomars:
Thanks again for the attempt @rjackson. Talked to the team and although this technically works we're opting to do this in a way that allows to scale and benefit all users instead of on a per client basis. I'm sharing the team feedback in order to have more context on the decision:

> Personally don’t know the intricacies of the library that he’s referring to but my take is that it feels like a non-ideal solution to have to use localStorage versus very standard response headers that the browser can use to cache files. And then when content changes, it gets invalidated (using correct mechanisms for this of course)

> Also, localStorage benefits one person. While response cache would benefit everyone as that can be cached on edge. So, even a new user visiting [cal.com](http://cal.com/) would have benefit of that.
> 
> In addition to that, localStorage isn't available in third party context. So, embeds which will always be in third party context, can't access that and thus they won't benefit from that.

> Nothing more to add to what has already been said - in full agreement, with the perf improvements on the cards I'd like to investigate improving our i18n setup "the right way".

Will close this PR for now, feel free to continue the discussion in the original issue. ​May you be well 🙏

rjackson:
@zomars Gotcha.

I have some experimental work-in-progress from trying to configure next-i18next the proper way, but it introduced issues with:

- Flashes of untranslated tokens
- Server hydration mismatch errors
- Issues invalidating & reloading translations when a user changes language

https://github.com/calcom/cal.com/compare/main...rjackson:cal.com:i18n-pain?expand=1

The commit there was in-progress and incomplete, so I'm not 100% sure exactly what functioning state it was left it. Nonetheless, it could be useful for taking this forward and getting i18next integrated the right way.

As I said on #9840, as we're not really using any Next.js' localisation features directly we might have a cleaner integration by writing our custom integration layer against react-i18next, rather than trying to use `next-i18next` and hacking around to try and undo or skip some of its automatic behaviour.

For statically generated pages, I still don't think this will make a meaningful difference in terms of performance as we'll still need an API look up to determine the user's language before we can load the appropriate localisation files.

As you've suggested we can flash english content whilst that request is pending, but in that case I'd still question whether or not that's a pleasant experience for users. I'd still recommend a local storage adapter as a micro optimisation, although with a proper integration we should be able to leverage `i18next-localstorage-backend` directly for that rather than managing it in Cal.com code.

PR #9992, opened 2023-07-07
feat: legal notice component for imprint and privacy

### LINKED ISSUE #8427: add legal notice for imprint and privacy

We need legal notices for the public pages. There should be a possibility to link an imprint and a privacy policy. The URLs are configurable in .env file

grafik
grafik
grafik

servusrene:
## What does this PR do?

fixes #8427 

grafik
grafik

## Testing

To test the customization, `yarn db-studio` must be run. The next step is to open the `http://localhost:5555` website. There, under the `Deployment` Table, the values `imprintLink` and `privacyLink` must be set.


rene-mueller:
@PeerRich  i'm done, with the privacy and imprint feature.

PeerRich:
@rene-mueller this is dope. do you have a screenshot where an admin enters this during setup?

PeerRich:
> @rene-mueller this is dope. do you have a screenshot where an admin enters this during setup?

and where to edit after, if the URL changes

rene-mueller:
@PeerRich currently the changes are only possible via the database. The options should be changeable in the admin area (next to the branding). I currently have no time to change this.

PeerRich:
> @PeerRich currently the changes are only possible via the database. The options should be changeable in the admin area next to the branding and the other deployment settings. I currently have no time to change this.

ok i understand. how would I best test this?

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -173,13 +174,11 @@ const BookerComponent = ({
...
               key="meta"
               className={classNames(
-                "relative z-10 flex [grid-area:meta]",
+                "relative z-10 flex flex-col [grid-area:meta]",
                 // Important: In Embed if we make min-height:100vh, it will cause the height to continuously keep on increasing
                 layout !== BookerLayouts.MONTH_VIEW && !isEmbed && "sm:min-h-screen"
               )}>
-              <BookerSection
-                area="meta"
-                className="max-w-screen flex w-full flex-col md:w-[var(--booker-meta-width)]">
```
PeerRich, line 182: this looks dangerous. is this needed?

rene-mueller: @PeerRich I have adjusted it because area has no effect on the display. From my point of view this is a remnant and its not needed.

PeerRich:
i wonder how this will work with embeds tho

rene-mueller:
> i wonder how this will work with embeds tho

i have a check in the legalnotice component it may be better to move each to the outside.

leog (commented):
Thanks for the contribution @rene-mueller. Tested locally and works, also wondering about embed.

Let us know how can we help, we can also summon @Jaibles in case design is needed, his blessing on how it looks is always a good thing to have 😁

Also @PeerRich, should we add these new links in the Admin onboarding wizard? Sounds like a good thing to do to be sure nobody misses setting those up avoiding manual setup through `Deployment` database table

rene-mueller:
@leog @PeerRich why shouldn't it work for embed?

I got the layout from @PeerRich so it should be fine.

It would be great when you can add the setup wizard for that feature.

Jaibles:
In terms of design.. I just have one question... 
CleanShot 2023-08-09 at 16 13 54@2x
If so.. can we make it the same?

And can we have the casing be:
Imprint  |  Privacy

Instead of 
imprint  |  privacy

PR #9997, opened 2023-07-07
chore: Teams empty screen

leog:
## What does this PR do?

When an org admin creates an org and goes to teams page, a blank page was shown. This introduces an empty state for that scenario.

<kbd>image</kbd>

## Type of change

## How should this be tested?

Create an org and go to teams page, you should see an empty screen now.


CarinaWolli (approved):
Looks good! 👍

Jaibles:
That empty state looks perfect to me. The empty state for teams in Orgs does not need to be as promo-y as the regular ones as we're not trying to get them to upgrade here. They're already on the highest plan so all they need to see here is a clear action to create a team. I'm sure we can do something better in the future but this is perfect for now

PR #10001, opened 2023-07-07
fix: typo string to strong

sydwardrae:
fixed a quick little spelling error


leog (approved):
Thanks for taking care of this

PR #10012, opened 2023-07-08
fix: Cleanup after removing connected calendar 

### LINKED ISSUE #9988: Cleanup after removing connected calendar is incomplete

### Issue Summary

When we remove a connected calendar, it doesn't get removed entirely. It doesn't show up in the list of calendars anymore, but it persists somewhere within the DB (potentially in the Selected Calendars table) and it becomes a problem as they are used during the freebusy query in the getAvailability.

Ideally, we would want it to be removed entirely instead of persisting.

### Video showing example

[Loom Video](https://www.loom.com/share/be444227a73b43758268a12aa661bc49?sid=354eea45-48d1-4d3a-b3c6-4c78d525abef)

### Steps to Reproduce

1. Add a calendar
2. Notice the new calendar connection entry in selected calendar table
3. Remove this calendar
4. The calendar doesn't show up in the calendar list in UI, but persists in the selected calendar table and is processed in the get availability call.

### Technical details

- Tested on version [v.3.0.14-sh](https://go.cal.com/releases) locally

Pradumn27:
## What does this PR do?

Fixes the database inconsistency caused when a calendar is being deleted, as even after being deleted it was still present in the SelectedCalendars table.

Fixes #9988 

https://github.com/calcom/cal.com/assets/47187878/61f9f865-41ef-4753-82de-7a89b94efde1

## Type of change

## How should this be tested?

- Connect a calendar
- Check that it is present in the database's SelectedCalendar table
- delete it
- See that it is also removed from the SelectedCalendar table

## Mandatory Tasks


joeauyeung (dismissed):
@Pradumn27 thank you for you contribution.

Testing this, only the primary calendar is deleted from the `SelectedCalendars` table. The expected behaviour is all calendars associated with that calendar integration should be deleted.

Pradumn27:
Thanks for the review @joeauyeung , have done the changes accordingly and now all the selected calendars associated with that account are getting deleted.

keithwillcode:
Curious if this fixes #10066

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -313,6 +302,27 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
     }
   }
 
+  // If it's a calendar remove it from the SelectedCalendars
+  if (credential.app?.categories.includes(AppCategories.calendar)) {
+    const calendar = await getCalendar(credential);
+
+    const calendars = await calendar?.listCalendars();
```
joeauyeung, line 309: I like this approach for our current selected calendars, but I think we can improve this. I'm going to approve for now as this is a great solution.

joeauyeung (approved):
LGTM. Thanks for your contribution @Pradumn27

PR #10019, opened 2023-07-09
fix: Refresh session

### LINKED ISSUE #9965: Session not getting refreshed
### Issue Summary

Changing the username will not alter the data stored in session, which means that the data retrieved from session and utilized will remain unaffected. An example of this is the public URL page in the sidebar, which will not be updated.

### Steps to Reproduce

1. Change the username
2. Check Whether the public url in sidebar is changed 

When you login with google provider an username will be autogenerated, when you change this username that will not be reflected in ui immediately due to the session problem.

Suggestion: Next-auth v4.22.1 has a function to update the sesison

The-Robin-Hood:
## What does this PR do?
Fixes #9965

## Type of change

- Chore (workflow improvements)

## How should this be tested?

- [ ] After changing username check whether it gets updated in next session retrieval or not.
- [ ] Check public page url gets updated properly after the username change
- [ ] Check premium username change works properly
- [ ] Signin with google and test whether the session works properly (change username when entered and it should be reflect in dashboard)
- [ ] Ensure the update doesnt break the session datas


zomars (dismissed):
Thank you for your contribution 🙏 @The-Robin-Hood

Upon further inspection it seems like this will help for sure when updating the username. But this PR made me realize that it will remain inconsistent with every single instance where we use the updateProfile mutation.

image


How about this?

We should extract the useMutation hook into a custom hook that already invalidates/refresh mostly affected endpoint AND session data. And use that hook everywhere the mutation is used. I'm thinking something like this:

image

apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
```diff
@@ -69,6 +71,13 @@ const UsernameTextfield = (props: ICustomUsernameProps & Partial<React.Component
       onSuccessMutation && (await onSuccessMutation());
       setOpenDialogSaveUsername(false);
       setCurrentUsername(inputUsernameValue);
+      await update({
+        ...session,
+        user: {
+          ...session?.user,
+          username: inputUsernameValue,
+        },
+      });
```
zomars, line 80: We can also just use this and let the server revalidate, otherwise we would need to validate the input again.

```suggestion
      await update();
```

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -349,7 +349,10 @@ export const AUTH_OPTIONS: AuthOptions = {
   },
   providers,
   callbacks: {
-    async jwt({ token, user, account }) {
+    async jwt({ token, user, account, trigger, session }) {
+      if (trigger === "update") {
+        return { ...token, ...session.user };
+      }
```
zomars, line 355: I know this is faster, but let the server do it's thing.

```suggestion
```

apps/web/components/ui/UsernameAvailability/PremiumTextfield.tsx
```diff
@@ -94,6 +96,13 @@ const PremiumTextfield = (props: ICustomUsernameProps) => {
   const updateUsername = trpc.viewer.updateProfile.useMutation({
     onSuccess: async () => {
       onSuccessMutation && (await onSuccessMutation());
+      await update({
+        ...session,
+        user: {
+          ...session?.user,
+          username: inputUsernameValue,
+        },
+      });
```
zomars, line 105: Same as my previous comment. Extra points for encapsulating this behavior in a custom `trpc.viewer.updateProfile.useMutation` hook and just use that.

```suggestion
      await update();
```

The-Robin-Hood: LGTM

emrysal (approved):
Just gave it a test and it works, @zomars made some good points we should tackle in a follow up

PR #10025, opened 2023-07-09
fix: fixes issue with webhook not being triggered for managed events

### LINKED ISSUE #9962: [CAL-2111] Managed event does not fire Webhooks if it has multiple team members

### Issue Summary

A managed event doesn't fire webhook if the it contains multiple team members. If it has a single team member, the webhooks fire.

### Steps to Reproduce

1. Create a managed event
2. Add a webhook to this event with BOOKING.CREATED trigger enabled
3. Make booking and confirm that the webhook works
4. Now add more team members to the event
5. Make another booking, and how it shouldn't trigger the webhook

### Technical details

- Production

<sub>[CAL-2111](https://linear.app/calcom/issue/CAL-2111/managed-event-does-not-fire-webhooks-if-it-has-multiple-team-members)</sub>

Pradumn27:
## What does this PR do?

This PR fixes the issue that the Booking Created webhook was not being triggered for managed events. The reason was that the eventTypes created for every assigned team member did not contain the teamId which was needed to find the right webhook while triggering.

We can't add teamId to this eventType either because [slug,teamId] is a unique identifier for the event types and it's important that it stays that way so that we can't have a team having multiple events with the same slug. So what I've done is while booking I'll check whether it's a managed event and then will extract the eventType from the eventType of this slug which contains the teamId this way we will be able to access the teamId.

Also, the event type that was being created when a team member was assigned a managed event did not contain the schedulingType so have added that too.

Fixes #9962 

https://github.com/calcom/cal.com/assets/47187878/19ffda33-0362-4fbf-82a4-7e188495c94e

## Type of change

## How should this be tested?

- Create a managed event 
- Configure a webhook for this event with BOOKING.CREATED trigger enabled
- Add members to this event
- Make a booking
- Check that the webhook was triggered

## Mandatory Tasks


CarinaWolli:
@Jaibles (or @leog, @alishaz-polymath) should managed event types trigger the user's webhooks, or should they trigger the team webhooks? 

@Pradumn27 I see in your screen share you show that it triggers for a team webhooks, however, that's not working for me 🤔

Pradumn27:
> @Pradumn27 I see in your screen share you show that it triggers for a team webhooks, however, that's not working for me 🤔

@CarinaWolli are you using any already created user's event? Because then it will not work as I've mentioned in the pr description "The event type that was being created when a team member was assigned a managed event did not contain the schedulingType so have added that too." that might be the possible case, do let me know will debug this further then.

CarinaWolli:
Ah I see, sorry I missed that. That was the case for me! 

We also recognize a managed event type by it having a `parent`. So we could just simply check if it has a parent and then check the parent's teamId

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2164,11 +2164,33 @@ async function handler(
...
+      },
+      select: {
+        teamId: true,
+      },
+    });
+
+    if (managedEvent?.teamId) {
+      teamId = managedEvent.teamId;
+    }
+  }
```
hariombalhara, line 2187: Let's move this code out of `handler` function into it's own function(say getTeamId())

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2164,11 +2164,33 @@ async function handler(
...
     smsReminderNumber: booking?.smsReminderNumber || undefined,
   };
+
+  let teamId = eventType.team?.id;
+
+  // If it's a managed event we need to find the teamId for it from the originally created event with this slug
+  if (!teamId && eventType.schedulingType === SchedulingType.MANAGED) {
+    const managedEvent = await prisma.eventType.findFirst({
+      where: {
+        slug: req.body.eventTypeSlug,
```
hariombalhara, line 2174: This is wrong way to get the parent managed event-type. Slugs can be duplicate.
We should get by parentId of the event.

hariombalhara (changes requested):
Thanks for the PR 🙏 But it needs some fixing

Pradumn27:
@CarinaWolli @hariombalhara thanks for the review, wasn't aware of the parentId field, have made the requested adjustments please review again.

packages/lib/defaultEvents.ts
```diff
@@ -92,6 +92,7 @@ const commons = {
   userId: 0,
   owner: null,
   workflows: [],
+  parentId: null,
```
Pradumn27, line 95: Added to remove the type warnings for getEventType.

packages/features/ee/managed-event-types/lib/handleChildrenEventTypes.ts
```diff
@@ -205,6 +205,7 @@ export default async function handleChildrenEventTypes({
               },
             },*/
             hashedLink: hashedLinkQuery(userId),
+            schedulingType: updatedEventType.schedulingType,
```
hariombalhara, line 208: How is this newly added schedulingType property being used ?

hariombalhara:
> > @Pradumn27 I see in your screen share you show that it triggers for a team webhooks, however, that's not working for me 🤔
> 
> @CarinaWolli are you using any already created user's event? Because then it will not work as I've mentioned in the pr description "The event type that was being created when a team member was assigned a managed event did not contain the schedulingType so have added that too." that might be the possible case, do let me know will debug this further then.

Do we need schedulingType? The requirement is to simply get the parent team ID which is already available in parentId. My concern is to to get it working for existing event-types as well

packages/features/ee/managed-event-types/lib/handleChildrenEventTypes.ts
Pradumn27: It isn't used currently, but for database consistency shouldn't this event still have this schedulingType?

Pradumn27:
> Do we need schedulingType? The requirement is to simply get the parent team ID which is already available in parentId. My concern is to to get it working for existing event-types as well

@hariombalhara As per the current changes it will work for the existing event-types as well, in the case of schedulingType I figured if we want to have database consistency shouldn't these managed events also have this schedulingType?

hariombalhara (approved):
Awesome work @Pradumn27 🙏 Thank you !!

PR #10037, opened 2023-07-10
feat: Auto accept on org verify

### LINKED ISSUE #10005: Members in pending status should be accepted automatically after an Organization is verified

Currently, if an unverified Organization has members not confirmed yet (Pending status), after the acceptance, its members from the same Organization are still in Pending status.

All members from the same Organization (slug) should be accepted automatically.

It could be configurable with a toggle in the admin settings.

sean-brydon:
## What does this PR do?
Fixes #10005

Loom: https://www.loom.com/share/e271813e06af43e69cf36e99e54173d0

How to test: 
Create a org - add a bunch of members from your domain + other domains
Add a child team - add some members to that also

In a new window login as an account with ADMIN (root level perms not team)
Goto - settings>admin>organization
Verify a domain

All users matching the domain should auto be accepted
Login as one of the test users -> Test they are apart of the organization and that they get presented organization in settings


packages/trpc/server/routers/viewer/organizations/adminVerify.handler.ts
```diff
@@ -57,5 +57,69 @@ export const adminVerifyHandler = async ({ input }: AdminVerifyOptions) => {
...
+        some: {
+          teamId: input.orgId,
+        },
+      },
+    },
+    select: {
+      id: true,
+      email: true,
+    },
+  });
```
sean-brydon, line 75: Find only users with this email domain that have a membership with this email domain

sean-brydon: We could make this Distinct on userId not sure if its needed tho

packages/trpc/server/routers/viewer/organizations/adminVerify.handler.ts
```diff
@@ -57,5 +57,69 @@ export const adminVerifyHandler = async ({ input }: AdminVerifyOptions) => {
...
+
+  const userIds = foundUsersWithMatchingEmailDomain.map((user) => user.id);
+  const userEmails = foundUsersWithMatchingEmailDomain.map((user) => user.email);
+
+  await prisma.$transaction([
```
sean-brydon, line 80: Transaction cause none of these depend on eachother

leog (approved):
LGTM, thanks for covering that scenario

packages/trpc/server/routers/viewer/organizations/adminVerify.handler.ts
zomars: What happens if there's no found users?

sean-brydon: All the above below should just not have any effect - we can early return tho i guess

ericrommel:
Is it need to send an invitation email if the users will be auto-accepted? Okay, there is a checkbox where you can decide if an invitation should be sent... IMO, if it will auto-accept then either the checkbox should be disabled/not shown or it even marked should be ignored and the email won't be sent. What do you think?

image

PR #10042, opened 2023-07-10
feat: Add password checks for password-reset

sean-brydon:
## What does this PR do?

for some reason forgot password-reset page lost its validation to check password validity 
CleanShot 2023-07-10 at 17 41 33@2x
CleanShot 2023-07-10 at 17 41 37@2x
CleanShot 2023-07-10 at 17 41 41@2x
CleanShot 2023-07-10 at 17 42 04@2x

How to test: 
* Invalid password -> Error
* Valid password -> success

### DIFF
--- a/apps/web/pages/api/auth/reset-password.ts
+++ b/apps/web/pages/api/auth/reset-password.ts
@@ -52,9 +52,23 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       },
     });
 
+    await expireResetPasswordRequest(rawRequestId);
+
     return res.status(201).json({ message: "Password reset." });
   } catch (reason) {
     console.error(reason);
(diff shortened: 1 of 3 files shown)


CarinaWolli:
@sean-brydon could you take a look at the failed e2e tests?

zomars (changes requested):
Blocking to made some adjustments. We can drop useState in the form. Also we don't mark a password request as already used in the API which allows to reuse the same URL to reset the password multiple times.

EDIT:

We need to updates e2e tests as well.

zomars (dismissed):
- Got rid of extra state in favor of form state
- Added request expiration after being used
- Updated test to match new password policy

zomars (approved):

PR #10046, opened 2023-07-10
test: Create unit tests for react components in packages/ui/components/avatar (CALCOM-612)

gitstart-calcom:
Here is the GitStart Ticket for this pull request: [CALCOM-612](https://developers.gitstart.com/client/calcom/ticket/CALCOM-612)

## DEMO

Run 
```
yarn test avatar
```

image


packages/ui/components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const ImageMock = (props: { [key: string]: string; alt: string }) => {
+    return <img {...props} alt={props.alt} />;
+  };
+  const actual = (await vi.importActual("@radix-ui/react-avatar")) as { default: object };
+  return {
+    ...actual.default,
+    Image: ImageMock,
+  };
+});
+
```
gitstart-calcom, line 23: The Radix Image component needs to be mocked because it only renders after the image loads,
https://www.radix-ui.com/docs/primitives/components/avatar#image
and the image doesn't load because js-dom doesn't have some browser APIs (See below about ResizerObserve).

packages/ui/components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+          setOpen(isOpen);
+        }}
+      />
+    );
+  };
+  return {
+    Tooltip: TooltipMock,
+  };
+});
+
```
gitstart-calcom, line 51: All these mocks are for Tooltip testing, radix uses ResizeObserver which js-dom doesn't have,
so I'm simulating its behavior, as it's just a mock, radix doesn't actually render the tooltip content in the resulting HTML,
but in this way, I managed to prove that the event is detected in the component and that once the state is updated, 
the `data-state` attribute, which comes from the radix, is modified internally (`data-state` starts as `closed `and ends as `instant-open`, see test block)

packages/ui/components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+const forcedWaitFor = async () => {
+  await waitFor(
+    async () => {
+      await new Promise((r) => setTimeout(r, 1000));
+    },
+    { timeout: 2000 }
+  );
+};
+
```
gitstart-calcom, line 66: These two serve to test the fallback, it is necessary to wait for the image to load to fail, for the fallback to appear and that is why the mock of the `Image`, needs to be undone since the default behavior of the radix needs to happen

gitstart-calcom:
As said on our slack group, we will remove the radix components tests and update the PR 😄

ericrommel (changes requested):
Hey @gitstart-calcom, sorry, but unit testing for ui components from a third-party library doesn't make any sense. Can you elaborate on how valuable for us to test them? unit tests are linked to actions on our behaviours, not from external components. We can achieve UI behavioural expectations from the high-level testing part, like UI Testing, Acceptance testing and others that can be even automated. What do you think to discuss about it (if needed)?

gitstart-calcom:
Hi @ericrommel, how are you? 😄 
It was discussed in our Slack group about creating these tests and we were allowed to create these PRs, unit tests for UI components offer valuable benefits, even if they come from a third-party library. here are some points that you should consider:

- They allow isolated testing of internal logic and behavior specific to the components.
- Unit tests provide a faster feedback loop, helping identify and address component-specific issues promptly.
- They contribute to overall test coverage and enhance confidence in the codebase's stability and reliability.
- Unit tests for UI components complement high-level testing approaches, focusing on internal behaviors.

In summary, incorporating unit tests for UI components in the calcom brings advantages in terms of isolated testing, faster feedback, improved test coverage, and enhanced confidence. What do you think?

gitstart-calcom:
The Avatar component doesn't use any special logic, it just renders the Radix component and some children. We will close this PR

PR #10048, opened 2023-07-10
fix: fixes issue of being able to reinvite an already invited member

### LINKED ISSUE #9998: It's possible to invite a member that was already invited in the Organization tab
### Issue Summary
It's possible to invite a member that was already invited in the Organization tab. If you try it from the `Invite individual` tab, it's not.

**Note: It happens to the `Bulk import` tab as well**

 
### Steps to Reproduce

1. Create an organization (/settings/organizations/new) and add a team
2. Add a few members to the Organization (/settings/organizations/members)
4. Go to add members to the team created (/settings/teams/[id]/members)
5. From the Invite individual tab, add a specific member
6. Try to add this same member again.
7. Note that a message `Member has already been invited` will be shown
8. From the Organization tab, select a member from the list to add to the team
9. Try it again selecting the same member
10. Note that a notification that the member was added will be shown and also another invitation email will be sent

### Actual Results

- Members that were already invited can be invited again

### Expected Results

- If a member was already invited, the system should be able to avoid inviting it again.

### Technical details

- Tested on Chrome (114.0.5735.134)

### Evidence

https://www.loom.com/share/0c9376ad8b2d4ec1ba01774ba3d33890

Pradumn27:
## What does this PR do?

Fixes the issue that it's possible to invite a member that was already invited in the Organization tab and the Bulk Invite tab

Fixes #9998

https://github.com/calcom/cal.com/assets/47187878/31bf78ea-5131-43d1-9029-0e96b9388b65

## Type of change

## How should this be tested?

- Create an organization (/settings/organizations/new) and add a team
- Add a few members to the Organization (/settings/organizations/members)
- Go to add members to the team created (/settings/teams/[id]/members)
- From the Invite individual tab, add a specific member
- Try to add this same member again.
- Note that a message Member has already been invited will be shown
- From the Organization tab, select a member from the list to add to the team
- Try it again selecting the same member
- Note that a message Member has already been invited will be shown
- Similar process with the Bulk Invite tab

## Mandatory Tasks


sean-brydon:
@ericrommel this looks good on my end - would you be able to look at the preview URL / this locally to confirm you no longer experience the issue?

emrysal (changes requested):
It would make sense to tackle this only in the backend; instead of only in the frontend. - this is an ideal case of a good error to bubble up from the inviteMember backend call. @Pradumn27 would you be up for attempting that approach instead?

Pradumn27:
> It would make sense to tackle this only in the backend; instead of only in the frontend. - this is an ideal case of a good error to bubble up from the inviteMember backend call. @Pradumn27 would you be up for attempting that approach instead?

Thanks for the review @emrysal , This event is actually already handled on the backend I checked the code and instead of throwing an error it was being logged, to clarify what you're suggesting is that this case gets handled only on the backend and when there is an error with existing invite we show it in a toast or show it as I've shown in the recording?

sean-brydon:
> It would make sense to tackle this only in the backend; instead of only in the frontend. - this is an ideal case of a good error to bubble up from the inviteMember backend call. @Pradumn27 would you be up for attempting that approach instead?

Just tested this - we seem to already do this on backend - were you able to get around that in anyway Alex?

PR #10051, opened 2023-07-11
fix: prevents warnings on branding call when unauthed

zomars:
## What does this PR do?

IDK if this endpoint should be public or not so I've disable the called until we're logged in.

## Type of change

## How should this be tested?

- See console when unearthed. The warning should be gone
- When authed and if on an org, data should return normally.

## Mandatory Tasks


emrysal (approved):

hariombalhara (approved):
LGTM!!

PR #10054, opened 2023-07-11
fix: Gets Google Calendar working locally even when not using localhost:3000 domain to access the app

hariombalhara:
## What does this PR do?

The problem is as follows:
- Google Calendar(and most others integrations) don't allow adding a non `localhost` domain which is not secure and doesn't have a valid tLD. e.g app.cal.local has `.local` tLD which isn't a valid registered TLD. Also, even if we use a valid TLD, it would require you to use an HTTPS url which most people don't setup locally.
- But we can't use http://localhost:3000 for testing organizations.
- So, we use http://localhost:3000 only in Google Calendar and when setting up OAuth, we do that with the same domain but when adding credentials to DB, we redirect to http://app.cal.local so that the authentication is still possible.

Note that you an still work it around by running an ngrok tunnel but that changes it's URL on every restart and is not a good experience at all.

## Requirement/Documentation
We need to be able to test Google Calendar integration locally

## Type of change
  - [x] Bug fix (non-breaking change which fixes an issue) existing functionality to not work as expected)

## How should this be tested?
- Follow the steps "Obtaining the Google API Credentials" from README.md to be able to test Google Calendar locally.
	- 	Make sure to use http://localhost:3000 for redirect URLs whitelisting in Google Cloud.
- Access the app at app.cal.local:3000(which is a requirement when testing org stuff). Now try installing Google Calendar. You should be able to install it now. It isn't possible to do so in main,

## Mandatory Tasks

## Checklist


apps/web/next.config.js
```diff
@@ -524,6 +524,25 @@ const nextConfig = {
         destination: "/apps/installed/conferencing",
         permanent: true,
       },
+      // OAuth callbacks when sent tolocalhost:3000(that would be expected) should be redirected to corresponding to WEBAPP_URL
```
hariombalhara, line 527:
```suggestion
      // OAuth callbacks when sent to localhost:3000(which would be expected) should be redirected to corresponding to WEBAPP_URL
```

packages/app-store/googlecalendar/api/add.ts
```diff
@@ -1,7 +1,7 @@
 import { google } from "googleapis";
 import type { NextApiRequest, NextApiResponse } from "next";
 
-import { WEBAPP_URL } from "@calcom/lib/constants";
+import { WEBAPP_URL_FOR_OAUTH } from "@calcom/lib/constants";
```
hariombalhara, line 4: We can do the same change in other integrations as well but we would need to test them first before making the change. I have tested Google Calendar, so adding it here.

keithwillcode (approved):

PR #10062, opened 2023-07-11
fix: avatars not loading for org events

hariombalhara:
For org users and teams, correct context of org should be used which is available only when using org domain and thus

Fixes CAL-2140 

Also, it fixes a crash for the team page due to inviteToken being sent from gSSP
Screenshot 2023-07-11 at 5 33 01 PM

Also, disables Org disbanding. It would be enabled later https://linear.app/calcom/issue/CAL-2138/ability-to-disband-organizations
Owner view for General now doesn't have Disband CTA
image

## Avatars in different scenarios

Screenshot 2023-07-11 at 6 21 58 PM

Screenshot 2023-07-11 at 6 22 22 PM

Screenshot 2023-07-11 at 6 22 35 PM

Screenshot 2023-07-11 at 6 24 03 PM

Screenshot 2023-07-11 at 6 30 03 PM

Screenshot 2023-07-11 at 6 35 12 PM

Screenshot 2023-07-11 at 6 35 36 PM


packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -137,7 +136,7 @@ export const getPublicEvent = async (
         username: users[0].username,
         name: users[0].name,
         weekStart: users[0].weekStart,
-        image: `${WEBAPP_URL}/${users[0].username}/avatar.png`,
+        image: `/${users[0].username}/avatar.png`,
```
hariombalhara, line 139: For public booking pages, I am using relative URL. There seems to be no need of using full URL and there isn't a quick way to get full URL without using orgBranding which requires the user to be logged in.

Using relative URL seems to be the ideal way to me.

leog (approved):
LGTM, great catch

PeerRich:
image
checks failing

PR #10063, opened 2023-07-11
fix: TS error in main

hariombalhara:
There is parentId already there. It happened due to two different PRs adding the same property and the second one before merge wasn't upto date with the main


emrysal (approved):

PeerRich (approved):

PeerRich:
force merging because the test would fail

PR #10086, opened 2023-07-12
Add huntr.dev to SECURITY.md

huntr-helper:
As requested through the platform by @calcom, this will point your security policy to [huntr.dev](https://huntr.dev/repos/calcom/cal.com})


ConnorGargano:
Not applicable, we are not looking into using Huntr.dev

PR #10090, opened 2023-07-12
fix: ternary operator with lenght causing 0

### LINKED ISSUE #10082: Dangling zero in the dashboard
### Issue Summary

There's a dangling zero at the bottom of the dashboard

image

### Steps to Reproduce

1. Click edit on any of your existing events - for example: https://app.cal.com/event-types/123456?tabName=advanced
2. Scroll to the bottom
3. Find a dangling zero down there

### Actual Results

  - There's a random zero at the bottom of the dashboard

### Expected Results

  - There should not be a random zero at the bottom of the dashboard

PeerRich:
PSA: dont use `length` with `&&` to prevent trailing 0's

### DIFF
--- a/apps/web/pages/event-types/[type]/index.tsx
+++ b/apps/web/pages/event-types/[type]/index.tsx
@@ -503,7 +503,7 @@ const EventTypePage = (props: EventTypeSetupProps) => {
         </Form>
       </EventTypeSingleLayout>
 
-      {slugExistsChildrenDialogOpen.length && (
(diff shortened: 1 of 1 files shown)


alishaz-polymath (approved):

emrysal:
!! works too and is cleaner 👍

odicho:
I think a cleaner approach here is to actually do a strict comparison to 0. That's the only time the ternary operator will evaluate to false anyways. I try to avoid the `? <> : null` syntax whenever possible.

PR #10096, opened 2023-07-12
chore: Update GitHub env caching to use current PR details

keithwillcode:
## What does this PR do?

Our caching for environment variables was causing old variables to get injected into GitHub actions whenever new variables were added. This is directly affecting us from adding e2e tests for workflows.

## Type of change

## Mandatory Tasks


keithwillcode:
@zomars Note that GitHub isn't picking up changes properly in these workflow files and so it's failing builds. Here's a workflow that is run directly against this branch which allows GitHub to use the proper workflow files. https://github.com/calcom/cal.com/actions/runs/5545197033

emrysal (approved):
looks good :)

zomars:
@keithwillcode @emrysal seems like old cache from main is no longer working: https://github.com/calcom/cal.com/actions/runs/5546144414/jobs/10126045140

PR #10098, opened 2023-07-12
chore: Relying on CALCOM_ENV

leog:
## What does this PR do?

Making sure we rely on our own Environment variable defined from NODE_ENV so we can better manage environments in Vercel as talked with @keithwillcode 

## Type of change


packages/lib/constants.ts
```diff
@@ -39,7 +39,8 @@ export const IS_SELF_HOSTED = !(
   new URL(WEBAPP_URL).hostname.endsWith(".cal.dev") || new URL(WEBAPP_URL).hostname.endsWith(".cal.com")
 );
 export const EMBED_LIB_URL = process.env.NEXT_PUBLIC_EMBED_LIB_URL || `${WEBAPP_URL}/embed/embed.js`;
-export const IS_PRODUCTION = process.env.NODE_ENV === "production";
+export const CALCOM_ENV = process.env.CALCOM_ENV || process.env.NODE_ENV;
```
keithwillcode, line 42: I tend to like having apps fail fast, especially when it comes to configuration. This code allows for potential configuration to get mixed up in the event we don't set the CALCOM_ENV.

Ideally we would make sure CALCOM_ENV is set, otherwise an error is thrown immediately.

Is this to preserve backwards compatibility?

keithwillcode: @leog Bump^^

leog: Yes it is, so you were implying we should make it like the following?
```js
 export const CALCOM_ENV = process.env.CALCOM_ENV || "development";
```

There are places we need to rely on `process.env.CALCOM_ENV` such as `next.config.js`, which wouldn't have a default value if we forgot to set it.

keithwillcode: No, just wanted to confirm it was for backwards compatibility. If we change it to this and someone upgrades without seeing they need this new env var, their system will be running in "development" mode. I think what you have originally make sense.

keithwillcode: This does make me wonder if we need to update all of the places where we are using process.env.CALCOM_ENV to use this constant or IS_PRODUCTION instead. Reason, someone upgrades without setting this value, all of the lines of code that reference it will not work correctly.

leog:
@keithwillcode is there anything you want me to change/update? pinging @calcom/core to review this too

keithwillcode:
> @keithwillcode is there anything you want me to change/update? pinging @calcom/core to review this too

https://github.com/calcom/cal.com/pull/10098#discussion_r1263984167
Yeah I feel like we should implement this^^

apps/web/next.config.js
```diff
@@ -25,8 +25,10 @@ if (!process.env.NEXTAUTH_URL && process.env.NEXT_PUBLIC_WEBAPP_URL) {
 if (!process.env.NEXT_PUBLIC_WEBSITE_URL) {
   process.env.NEXT_PUBLIC_WEBSITE_URL = process.env.NEXT_PUBLIC_WEBAPP_URL;
 }
-
-if (process.env.CSP_POLICY === "strict" && process.env.NODE_ENV === "production") {
+if (
+  process.env.CSP_POLICY === "strict" &&
+  (process.env.CALCOM_ENV === "production" || process.env.NODE_ENV === "production")
```
leog, line 30: This is to be able to have backward compatibility in case `CALCOM_ENV` is not set

apps/web/pages/_document.tsx
```diff
@@ -49,7 +51,7 @@ class MyDocument extends Document<Props> {
           <meta name="msapplication-TileColor" content="#ff0000" />
           <meta name="theme-color" media="(prefers-color-scheme: light)" content="#f9fafb" />
           <meta name="theme-color" media="(prefers-color-scheme: dark)" content="#1C1C1C" />
-          {(process.env.NODE_ENV === "development" || process.env.VERCEL_ENV === "preview") && (
+          {(!IS_PRODUCTION || process.env.VERCEL_ENV === "preview") && (
```
leog, line 54: Relying more on `IS_PRODUCTION` to be better prepared for backward compatibility.

leog:
> > @keithwillcode is there anything you want me to change/update? pinging @calcom/core to review this too
> 
> [#10098 (comment)](https://github.com/calcom/cal.com/pull/10098#discussion_r1263984167) Yeah I feel like we should implement this^^

@keithwillcode thanks for the pointer, totally missed it. Done!

keithwillcode (approved):

keithwillcode:
@leog Looks good but would like @zomars to take a look as well before we merge.

leog:
> I think I'm failing to understand. What is this PR solving? What is the issue?

@zomars I probably failed to explain it in the description. The thing is that Vercel takes control over the NODE_ENV variable without the possibility to alter us to our needs. In order to separate ourselves from what Vercel forces us to use in that env variable, we thought we could create a variable that could depend on NODE_ENV but also take our own established value depending on what we intend to get form the set up environment. @keithwillcode feel free to add anything I might have missed.

zomars:
> The thing is that Vercel takes control over the NODE_ENV variable without the possibility to alter us to our needs.

How so? Everything we deploy is a "production build" so even previews run production builds. A production build is what you get by running `yarn build`.

I get the confusion between a "production build" and a "production deployment" in Vercel. If we want to conditionally run code in the production deployment then we should rely on the [VERCEL_ENV](https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables) variable. 

IMO this PR is adding more confusion into these concepts.

keithwillcode:
> > The thing is that Vercel takes control over the NODE_ENV variable without the possibility to alter us to our needs.
> 
> How so? Everything we deploy is a "production build" so even previews run production builds. A production build is what you get by running `yarn build`.
> 
> I get the confusion between a "production build" and a "production deployment" in Vercel. If we want to conditionally run code in the production deployment then we should rely on the [VERCEL_ENV](https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables) variable.
> 
> IMO this PR is adding more confusion into these concepts.

@zomars Using only NODE_ENV or VERCEL_ENV, there's no way to distinguish between a production environment and test environment. Both run production builds from Vercel's perspective and so NODE_ENV and VERCEL_ENV would both be `production`. Also, relying on VERCEL_ENV makes it harder when not deploying on Vercel.

There are many places in our code where we only want to run the code if it's a production environment, but there's no way to distinguish that properly at the moment. Introducing this CALCOM_ENV variables allows us to tell the system that it's a "staging" environment for instance.

zomars:
What about

```
const IS_CAL_PROD = process.env.WEBAPP_URL === "app.cal.com"
```

zomars:
I'm not against the PR but just trying to understand the actual use case

zomars:
> > What about
> > ```
> > const IS_CAL_PROD = process.env.WEBAPP_URL === "app.cal.com"
> > ```
> 
> What about self-hosters?
> 
> the idea is giving the installer the control to set the running environment of the app since there are places in the code base that only should be executed when in prod and vice versa. Example: handling stack traces.

Ok, It seems like we're talking about different things then. That's what NODE_ENV is for.

IMO we should differentiate what a "build" and a "deployment" is.

- On localhost we run a development build, we don't deploy
- On cal.vercel.app we run a production build, but it is not a production environment
- On cal.com, we run both a production build and it is a production environment

Are there use cases when we need to run code on a "production build" and "non production deployment"?

keithwillcode:
> > > What about
> > > ```
> > > const IS_CAL_PROD = process.env.WEBAPP_URL === "app.cal.com"
> > > ```
> > 
> > 
> > What about self-hosters?
> > the idea is giving the installer the control to set the running environment of the app since there are places in the code base that only should be executed when in prod and vice versa. Example: handling stack traces.
> 
> Ok, It seems like we're talking about different things then. That's what NODE_ENV is for.
> 
> IMO we should differentiate what a "build" and a "deployment" is.
> 
> * On localhost we run a development build, we don't deploy
> * On cal.vercel.app we run a production build, but it is not a production deployment
> * On cal.com, we run both a production build and it is a production deployment
> 
> Are there use cases when we need to run code on a "production build" and "non production deployment"?

Yeah that’s exactly what QA and Preview environments are environments

Issue #10078, opened 2023-07-12
[CAL-2150] + shouldn't be allowed in username, conflicts with dynamic bookings


hariombalhara:
* "+ " shouldn't be allowed in username. It is reserved for dynamic bookings
  We should slugify + and replace it with '-'

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2150](https://linear.app/calcom/issue/CAL-2150/shouldnt-be-allowed-in-username-conflicts-with-dynamic-bookings)</sub>

AyushGovil:
Hey @hariombalhara , can you elaborate the issue with a small example or share me the steps to reproduce?

PeerRich:
we have dynamic. booking links such as cal.com/peer+bailey, thereby we need to ban "+" in usernames. adding a + should convert it to - in the username input

PeerRich:
![CleanShot 2023-07-12 at 13 54 24@2x](https://github.com/calcom/cal.com/assets/8019099/f42e664f-fce2-4dac-9d9c-4c58b0f7343a)
luckily there are no users yet with "+" in their username

AyushGovil:
I can work on this one if no one minds.

AyushGovil:
Can someone check this => https://github.com/calcom/cal.com/pull/10142. Need review.

valdemartti:
Although the related PRs have been closed, I believe this has been already implemented with `slugify` in the `updateProfileHandler` function and in the registration flow ([1eeb91a](https://github.com/calcom/cal.com/commit/1eeb91a793e577b10c3037a2158d7ad052659b20) and [f6005b8](https://github.com/calcom/cal.com/commit/f6005b8c70838421bf975aa876cbfde79b628bed)). I did notice that the username with `+` still persists in the user's session after changing a username, but believe that should be a different non-critical issue. Suggesting to close this issue as solved.

PR #10101, opened 2023-07-12
fix: username + replaced with -

AyushGovil:
## What does this PR do?

replaces all '+' with '-' in the username string.

Fixes # [(issue)](https://github.com/calcom/cal.com/issues/10078)

## Type of change

## How should this be tested?

image

Did console.log to check the new string in the browser console. Got the desired log in the browser. Could not test the functionality e2e in local because sign up functionality doesn't work in local therefore used console.log to check the final object.

image

Inputs used as username to test => ayush+gov, ayush-gov+,  +ayush-gov+, ++a 

## Mandatory Tasks


emrysal:
Hi @AyushGovil Have a look at our slugify (packages/lib/slugify.ts) - this is not the only place the username is slugified. Needs a global solution.

emrysal (changes requested):
Hi @AyushGovil Have a look at our slugify (packages/lib/slugify.ts) - this is not the only place the username is slugified. Needs a global solution.

AyushGovil:
Hey @emrysal, I see that cal has a sign up page in this format but in local project setup, the sign up page is different.

image

image

I asked this because I have to handle all possible ways in which the username can be created.

hariombalhara:
> Hey @emrysal, I see that cal has a sign up page in this format but in local project setup, the sign up page is different.
> 
> image
> 
> image
> 
> What is the reason for this? I asked this because I have to handle all possible ways in which the username can be created.

cal.com/signup is coming from the website repo which is private. If you would go to app.cal.com/signup, you would see the same.

hariombalhara (changes requested):
As @emrysal mentioned there would be more occurences.
Also let's make the change in `slugify` as well so if any existing + in username, that's also taken care of.

Other places we might need to handle are
- Username change from within the app(/settings/my-account)
- username change during onboarding

Also, I think we should call `slugify` function itself before saving the username

AyushGovil:
Hey @hariombalhara , @emrysal  I have made the changes => https://github.com/calcom/cal.com/pull/10142. Also regarding the slugify function there was no modification required because it was already converting the + to -. I have created a separate branch with new pr because my branch for this pr got deleted because of some issue in my local.

PR #10102, opened 2023-07-12
fix: Password reset tests & better expiry checking

emrysal:
## What does this PR do?

* By using Prisma instead of using NEXT_PUBLIC_IS_E2E we both improve the test and reduce differences between test & production.
* The API endpoint had no check on whether the password reset was expired, which is a low risk vulnerability.
* Added an extra check in the tests to confirm a used password reset cannot be re-used.

### DIFF
--- a/apps/web/pages/api/auth/forgot-password.ts
+++ b/apps/web/pages/api/auth/forgot-password.ts
@@ -88,17 +88,9 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
       resetLink,
     });
 
-    /** So we can test the password reset flow on CI */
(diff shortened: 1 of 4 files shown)


apps/web/pages/api/auth/forgot-password.ts
```diff
@@ -88,17 +88,9 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
       resetLink,
     });
 
-    /** So we can test the password reset flow on CI */
-    if (process.env.NEXT_PUBLIC_IS_E2E) {
```
zomars, line 92: Nice workaround 🙏🏽

zomars (approved):

PR #10103, opened 2023-07-12
fix: slot reservation uses wrong event length

CarinaWolli:
## What does this PR do?

Fixes that for events with multiple duration, the slot reservation system uses the wrong event length to reserve slots.

## Type of change

## How should this be tested?

- create event type with multiple durations choose 15 min and 120 min
- make 120 min the default duration 
- go to the event type link
- pick any 15min slot (for example 11:00-11:15) => this slot is reserved now
- Open a new browser tab with the event-type link
- check available slots for 15 min duration at the day you booked the event: 
   - expected slots: all slots except 11:00 should be here
   - before: all slots between 9:00 and 11:00 were missing


zomars (approved):
LGTM

PR #10104, opened 2023-07-12
fix: Use user locale for sending password resets

emrysal:
## What does this PR do?

Instead of passing through the browser locale it makes more sense of sending the email according to the locale configured by the user.


zomars (approved):
LGTM

PR #10108, opened 2023-07-12
fix: Creating a default availability for org owner [CAL-2142]

leog:
## What does this PR do?

When a user creates an organization, they don't go through the profile onboarding, preventing them to have a default availability created for them.

This PR introduces the creation of a default availability when the owner account is created. Empty state was already there, just not accessible through UI.

## Type of change

## How should this be tested?

1. Create an organization
2. Go to the Availability page
3. See a default availability be created

> **Note**
> As you will not be able to delete the only availability to see the empty state, check it deleting availability from DB

## Mandatory Tasks


packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -77,13 +78,31 @@ export const createHandler = async ({ input }: CreateOptions) => {
 
   const emailDomain = adminEmail.split("@")[1];
 
+  const t = await getTranslation(ctx.user.locale ?? "en", "common");
+  const availability = getAvailabilityFromSchedule(DEFAULT_SCHEDULE);
```
leog, line 82: Luckily it is this easy to create a default availability 🌟

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -77,13 +78,31 @@ export const createHandler = async ({ input }: CreateOptions) => {
...
+              createMany: {
+                data: availability.map((schedule) => ({
+                  days: schedule.days,
+                  startTime: schedule.startTime,
+                  endTime: schedule.endTime,
+                })),
+              },
+            },
+          },
+        },
```
leog, line 105: And then assigning it to the newly owner created user

apps/web/pages/availability/index.tsx
```diff
@@ -95,6 +95,7 @@ export function AvailabilityList({ schedules }: RouterOutputs["viewer"]["availab
             Icon={Clock}
             headline={t("new_schedule_heading")}
             description={t("new_schedule_description")}
+            className="w-full"
```
leog, line 98: Empty screen was not filling the whole space

sean-brydon (approved):
Havent tested but code looks good

emrysal (approved):

PR #10109, opened 2023-07-12
fix: hasTeamPlan with accepted prop

### LINKED ISSUE #7974: [CAL-1353] Show empty view in Insights when user has no team but team invite
### Issue Summary

If a user is not part of a team but has an invite open it currently looks like that:
Screenshot 2023-03-28 at 11 46 06

<sub>[CAL-1353](https://linear.app/calcom/issue/CAL-1353/show-empty-view-in-insights-when-user-has-no-team-but-team-invite)</sub>

alannnc:
## What does this PR do?

- Reported bug was there was no info to show when an user invited but not accepted membership can view insights page but without any data.
- Empty state on the page shouldn't be anymore since a user with any team membership will always have their default insights.
- Add accepted = true when querying if user hasTeamPlan enabled.

Fixes #7974

## Type of change

## How should this be tested?

  - What are the minimal test data to have? Running yarn seed-insights can help
  - What is expected (happy path) to have (input and output)? User invited to team but not accepted should not be able to see insights page.
  - Any other important info that could help to test that PR. None

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/trpc/server/routers/viewer/teams/hasTeamPlan.handler.ts
+++ b/packages/trpc/server/routers/viewer/teams/hasTeamPlan.handler.ts
@@ -12,6 +12,7 @@ export const hasTeamPlanHandler = async ({ ctx }: HasTeamPlanOptions) => {
 
   const hasTeamPlan = await prisma.membership.findFirst({
     where: {
+      accepted: true,
       userId,
       team: {
         slug: {


CarinaWolli (approved):
Nice fix 👏🏻

PR #10116, opened 2023-07-13
fix: Remove/Update behaviour of date-overrides

### LINKED ISSUE #8043: [CAL-1382] Issue when deleting date overrides
[https://www.loom.com/share/cd0410734b294d3596e600c00fadcb71](https://www.loom.com/share/cd0410734b294d3596e600c00fadcb71)
[https://www.loom.com/share/cd0410734b294d3596e600c00fadcb71](https://www.loom.com/share/cd0410734b294d3596e600c00fadcb71)

Created via Threads. See full discussion: [https://threads.com/34460761924](https://threads.com/34460761924)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1382](https://linear.app/calcom/issue/CAL-1382/issue-when-deleting-date-overrides)</sub>

emrysal:
## What does this PR do?

This is not the cleanest approach, but the existing mechanism didn't work in a few situations because of the date sort. The update caused a re-render that caused an invalid unsortedFieldArrayMap.

Reproduction/how to test: 

**Updating bug was hit when you changed the date and it went before another date (re-ordering)**; which then modified the wrong date override. This PR replaces the entire array which retains the correct sort order by brute force.

**Delete bug was encountered if the sorted order was not the same as the filtered order, which depends on the order you created the date overrides in**. This PR implements an original order before the sort causing the right item to be removed.

Fixes #8043

### DIFF
--- a/apps/web/pages/availability/[schedule].tsx
+++ b/apps/web/pages/availability/[schedule].tsx
@@ -53,7 +53,7 @@ type AvailabilityFormValues = {
 };
 
 const DateOverride = ({ workingHours }: { workingHours: WorkingHours[] }) => {
-  const { remove, append, update, fields } = useFieldArray<AvailabilityFormValues, "dateOverrides">({
+  const { remove, append, replace, fields } = useFieldArray<AvailabilityFormValues, "dateOverrides">({
     name: "dateOverrides",
   });
   const { t } = useLocale();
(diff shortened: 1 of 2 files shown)


zomars (approved):
LGTM

zomars:
@zomars merged this pull request with [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/10116).

PR #10130, opened 2023-07-13
fix: Important Booking Page fixes

hariombalhara:
## What does this PR do?
Fixes following issues:

- Wrong org's event served when having same usename and event slug in another org
`vwo` org serving `404bug` org's bookme event.
Screenshot 2023-07-13 at 2 47 55 PM

- [Crash of user page when going from Team Page to User Page](https://www.loom.com/share/6e6c2d3e39534f8fb394a5f81ff91fa2)

- Also fixes route params being forwarded as query params. Not a bug exactly but adds additional query params in URL that are not needed.

- Fixes dynamic bookings allowing users outside of org.
  free user doesn't exist for `wingify` org
image

## Type of change

## How should this be tested?
- Test booking pages - Team and User events for both Org and Non Org members.

## Mandatory Tasks

## Checklist


apps/web/pages/[user]/[type].tsx
```diff
@@ -26,13 +26,15 @@ export default function Type({ slug, user, booking, away, isBrandingHidden, resc
...
         hideBranding={isBrandingHidden}
+        org={org}
       />
       <Booker
         username={user}
         eventSlug={slug}
         bookingData={booking}
         isAway={away}
         hideBranding={isBrandingHidden}
+        org={org}
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:37]: Makes booker aware of org so that it can send the org back to fetch the event through public.event tRPC call.

apps/web/pages/api/user/avatar.ts
```diff
@@ -62,6 +64,9 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   const img = identity?.avatar;
   // If image isn't set or links to this route itself, use default avatar
   if (!img) {
+    if (identity?.org) {
+      res.setHeader("x-cal-org", identity.org);
```
hariombalhara, line 68: It's helpful for debugging. One can just send the request and know if the domain has been labelled as an org subdomain or not.

apps/web/pages/[user]/[type].tsx
```diff
@@ -26,13 +26,15 @@ export default function Type({ slug, user, booking, away, isBrandingHidden, resc
         eventSlug={slug}
         rescheduleUid={rescheduleUid ?? undefined}
         hideBranding={isBrandingHidden}
+        org={org}
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:29]: Informs SEO component as well about the org

apps/web/pages/org/[orgSlug]/[user]/[type].tsx
```diff
@@ -55,8 +55,8 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
 type Props = UserTypePageProps | TeamTypePageProps;
 
 export default function Page(props: Props) {
-  if ((props as TeamTypePageProps)?.teamId) return TeamTypePage(props as TeamTypePageProps);
```
Comment (hariombalhara) [inline apps/web/pages/org/[orgSlug]/[user]/[type].tsx:58]: Converting these function calls to JSX fixes the issue because with function calls we end up with wrong hooks order.

apps/web/pages/org/[orgSlug]/[user]/index.tsx
```diff
@@ -33,8 +33,8 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
 type Props = UserPageProps | TeamPageProps;
 
 export default function Page(props: Props) {
-  if ((props as TeamPageProps)?.team) return TeamPage(props as TeamPageProps);
```
Comment (hariombalhara) [inline apps/web/pages/org/[orgSlug]/[user]/index.tsx:36]: Same here, converting function calls to JSX

apps/web/pages/team/[slug].tsx
```diff
@@ -64,7 +64,7 @@ function TeamPage({ team, isUnpublished, markdownStrippedBio, isValidOrgDomain }
   }
 
   // slug is a route parameter, we don't want to forward it to the next route
```
Comment (hariombalhara) [inline apps/web/pages/team/[slug].tsx:66]:
```suggestion
  // We don't want to forward route params to the next route. Intention is to just pass any additional query param eg. for prefilling fields.
```

apps/web/pages/team/[slug]/[type].tsx
```diff
@@ -67,30 +76,33 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
   if (!team) {
     return {
       notFound: true,
-    };
+    } as const;
```
Comment (hariombalhara) [inline apps/web/pages/team/[slug]/[type].tsx:79]: It allows reusing team/[slug]/[type] in team/[slug]/[type]/embed without TS errors.

apps/web/pages/team/[slug]/[type]/embed.tsx
```diff
@@ -1,104 +1,7 @@
-import type { GetServerSidePropsContext } from "next";
-import { z } from "zod";
+import withEmbedSsr from "@lib/withEmbedSsr";
```
Comment (hariombalhara) [inline apps/web/pages/team/[slug]/[type]/embed.tsx:1]: It was a copy paste of team/[slug]/[type], so importing that only here like it's done for other embed routes.

leog (approved):
Thanks for taking care of this @hariombalhara. Changes look good, but I was wondering if there is an alternative way to get the org context without having to pass it down everywhere like `org={org}`. The orgId is present in the user session, although that is for a logged in user. We could revisit later to improve in future releases, not a showstopper at least for me, approving.

hariombalhara:
@leog I thought about it. There are two reasons for that:
- <Booker> is supposed to be used as a react component by our customers. So this way they can explicitly set which org they want to work with when embedding the react component on their page
- Secondly, as you mentioned, session isn't available on booker page.

PR #10137, opened 2023-07-13
test: add apple calendar testing

### LINKED ISSUE #9373: [CAL-1887] Make E2E test: Can add Apple Calendar

wesleymatosdev:
- chore: include data-testids
- test: create apple calendar basic test case
- refactor: create fixture to install apple calendar

## What does this PR do?

Fixes #9373
/claim #9373

## Type of change

## How should this be tested?

      - `E2E_TEST_APPLE_CALENDAR_EMAIL`: Your iCloud e-mail
      - `E2E_TEST_APPLE_CALENDAR_PASSWORD`: The app password that you generated on iCloud
        - To generate an app password, go for https://appleid.apple.com/account/manage
        - Open the "App-Specific Password"
        - Click on "Generate an app-specific password" (or in the "+" button if you already have other app passwords)
        - Set a name to this app password
        - Copy the password and paste on the end of your `.env` file setting the `E2E_TEST_APPLE_CALENDAR_PASSWORD` variable
    - Paste `yarn e2e apple-calendar.e2e.ts` on your terminal and wait
    - It should successfully login into the Apple Calendar and the test should pass


apps/web/playwright/apple-calendar.e2e.ts
```diff
@@ -0,0 +1,34 @@
...
+test.afterEach(({ users }) => users.deleteAll());
+
+// eslint-disable-next-line turbo/no-undeclared-env-vars
+const APPLE_CALENDAR_EMAIL = process.env.E2E_TEST_APPLE_CALENDAR_EMAIL;
+// eslint-disable-next-line turbo/no-undeclared-env-vars
+const APPLE_CALENDAR_PASSWORD = process.env.E2E_TEST_APPLE_CALENDAR_PASSWORD;
+
+test.describe("Apple Calendar", () => {
+  if (!APPLE_CALENDAR_EMAIL || !APPLE_CALENDAR_PASSWORD)
+    throw new Error("It should run only it has the testing credentials");
```
ologbonowiwi, line 17: I'm aware that we usually do `test.skip` for checking variable values (as in [integrations-stripe.e2e.ts](https://github.com/calcom/cal.com/blob/6c7d7be9cc96a9d3359564e350117ca25e0f82e2/apps/web/playwright/integrations-stripe.e2e.ts#L20)), but in this case, I throw an error to have `APPLE_CALENDAR_EMAIL` and `APPLE_CALENDAR_PASSWORD` as strings, instead of manually do an `as string` when filling the inputs (don't handle types at all would cause a transpilation error on typescript because the `fill` method expects a string and can't be an undefined value)

ologbonowiwi: I noticed that this is NOT what should happen at all, I'll change it to `skip` to avoid breaking tests on pipelines that are not set up yet.

But not sure about how I should handle the typescript warning about these variables being undefined; can I just `const APPLE_CALENDAR_EMAIL = value as string`?

.env.example
```diff
@@ -215,3 +215,8 @@ PROJECT_ID_VERCEL=
 TEAM_ID_VERCEL=
 # Get it from: https://vercel.com/account/tokens
 AUTH_BEARER_TOKEN_VERCEL=
+
+#   - APPLE CALENDAR
+# Used for E2E tests on Apple Calendar
+E2E_TEST_APPLE_CALENDAR_EMAIL=""
+E2E_TEST_APPLE_CALENDAR_PASSWORD=""
```
PeerRich, line 222: can we not hardcode these instead? since its just for testing? have a throway account. or is it possible to change the password?

ologbonowiwi: It is possible to change the password. Apple doesn't allow you to login into external apps using the main iCloud pass, so you must create an app-specific password.

Honestly, since the code base is public, I guess we could create a throwaway account but add this on the test pipeline rather than have it hard coded since people can try to access the account with bad intentions, and IDK how Apple protects stuff like IP of who created the account, the e-mail attached, etc.

Even though, I'm with you if you say go for the hard-coded. If so, please provide me an e-mail + app password to put in there. I tried to create one myself but had no success without attaching it to a valid e-mail/phone number.

shivamklr (commented):
The code works on local. Feel free to review the code.

apps/web/playwright/apple-calendar.e2e.ts
zomars: You can null assert them like this: APPLE_CALENDAR_EMAIL!

ologbonowiwi: This `APPLE_CALENDAR_EMAIL!` also checks for undefined @zomars? Asking because the value it'll come as `undefined` probably

.env.example
shivamklr: @keithwillcode Can we add the above to GH .env before merging this PR?

apps/web/playwright/apple-calendar.e2e.ts
zomars: We don't need to check for undefined here. That's what the skip is for. So if they're set we can assume they're not null.

zomars (dismissed):
Made some fixes and as far as I'm concerned this is ready to merge. From our side we need to create a throwaway account and set it up in GH. But we can safely merge now.

Thank you for your contribution 🙏 @ologbonowiwi

zomars:
> need you to re-review when possible @zomars. I addressed the changes regarding throwing error/using skip.

I've already addressed the changes myself. Can we revert those? Of not I'll handle it from here.

zomars (approved):

PR #10142, opened 2023-07-13
fix: slugify username at sign up and username change within app

AyushGovil:
## What does this PR do?

replaces all '+' with '-' in the username string.

Fixes # https://github.com/calcom/cal.com/issues/10078

## Type of change

## How should this be tested?

Did console.log to check the new string in the browser console. Got the desired log in the browser. Could not test the functionality e2e in local for sign up because sign up functionality doesn't work in local.

image

Testing for username change within app defined with images below.

image

image

image

some test cases I used and result
++x  => x
ayush+go => ayush-go
v+1 => v-1
v+++ => v
++ => empty string

Reviewer can also try once using my code on local and test further. 

## Mandatory Tasks

### DIFF
--- a/apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
+++ b/apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
@@ -2,7 +2,7 @@ import classNames from "classnames";
 import { debounce, noop } from "lodash";
 import type { RefCallback } from "react";
 import { useEffect, useMemo, useState } from "react";
-
+import slugify from "@calcom/lib/slugify";
 import { fetchUsername } from "@calcom/lib/fetchUsername";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import type { TRPCClientErrorLike } from "@calcom/trpc/client";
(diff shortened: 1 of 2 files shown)


AyushGovil:
Need review @emrysal , @hariombalhara .

PeerRich:
i am not sure if using useEffects with states is the right approach here. cc @sean-brydon and @hario

apps/web/components/ui/UsernameAvailability/UsernameTextfield.tsx
```diff
@@ -36,19 +36,21 @@ const UsernameTextfield = (props: ICustomUsernameProps & Partial<React.Component
...
         setMarkAsError(!data.available);
         setUsernameIsAvailable(data.available);
+        setFinalUserNameValue(username);
       }, 150),
     []
   );
 
   useEffect(() => {
-    if (!inputUsernameValue) {
+    if (!inputUsernameValue || !slugify(inputUsernameValue)) {
```
AyushGovil, line 53: Adding !slugify(inputUsernameValue) will handle cases like these:

User Input => '+++'
value of !inputUsernameValue = false
value of !slugify(inputUsernameValue) = true because this function would return '' for this user input.
It will enter the condition and cancel the debounce preventing unneccesary api call.

PR #10157, opened 2023-07-14
fix: replace all async foreach callbacks

### LINKED ISSUE #10156: Async forEach callbacks cause unpredictable results and prevent error handling
### Issue Summary

Found many instances of `forEach(async ..)` pattern.

JS engine will just fire off the statements, go to the next line, and essentially run those async callbacks in the background.

It's impractical to handle or even catch errors with these constructs. It also creates race conditions that result in unexpected and unpredictable results.

This _could_ be intentional, i.e. hackish way to implement "backround tasks" to speed up parent function returns. But I would *strongly* advise against this for the above reasons.

### Steps to Reproduce

Example from wipemycal:
```js
try {
  bookingRefsFiltered.forEach(async (bookingRef) => {
    // some code
  });
} catch (error) {
  // THIS WILL NEVER FIRE
  if (error instanceof Error) {
    logger.error(error.message);
  }
}
// THIS WILL RUN BEFORE THE ASYNC CODE
await sendRequestRescheduleEmail(..)
```

### Actual Results

Parallel task execution, with:
- uncaught errors + no error handling
- unpredictable results

### Expected Results

Parallel task execution, with:
- error handling
- predictable results

### Evidence

`forEach()` [expects a synchronous function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#:~:text=expects%20a%20synchronous%20function)

nicktrn:
## What does this PR do?

- `forEach(asyncFn)` => `await Promise.all(map(asyncFn))`
- some error logging and refactoring
- add eslint rule to prevent reoccurence

Fixes #10156
/claim #10157

Also fixes:
- flaky tests (possibly)
- longterm outstanding issues (likely)
- code quality (definitely)
- tears, sweat, and banging your head against the wall (definitely)

## Type of change

## Caveat

Since we wait for promises to resolve, this will add parent function execution time - hopefully not much. In any case, this is likely preferable to *unpredictable* results.

## Todo

- [x] add eslint rule to prevent reoccurence
- [ ] fix similar issues
  * async function calls without await - there are many
  * implicit async .forEach with promise return
- [ ] rfc: consider additional eslint rules
  * no-async-promise-executor
  * no-await-in-loop
  * no-return-await
  * require-await
- [ ] rfc: task queues with error handling, concurrency limits, timeout, retry, rollback, etc
- [ ] rfc: use queues to add real background task runners
- [ ] rfc: standardise error logging / compliance

## Mandatory Tasks

## Checklist


packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -449,13 +450,16 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
 
       //cancel all workflow reminders from steps that were edited
-      remindersToUpdate.forEach(async (reminder) => {
+      const pDeleteReminders = remindersToUpdate.map((reminder) => {
         if (reminder.method === WorkflowMethods.EMAIL) {
           deleteScheduledEmailReminder(reminder.id, reminder.referenceId);
         } else if (reminder.method === WorkflowMethods.SMS) {
           deleteScheduledSMSReminder(reminder.id, reminder.referenceId);
+        } else if (reminder.method === WorkflowMethods.WHATSAPP) {
+          deleteScheduledWhatsappReminder(reminder.id, reminder.referenceId);
```
nicktrn, line 459: seems this was forgotten here, but similar calls found elsewhere in the file

nicktrn:
Came across even more instances of this in different forms. It's impossible to tell in many places if this is desired behaviour or not.

Might make sense to be explicit by requiring the use of a wrapper function like this:

```js
type NoArgFunction = () => unknown;

const asyncWrap = async (fn: NoArgFunction) => fn();

export const sendToBackgroundAndNeverThrow = (fn: NoArgFunction | NoArgFunction[]): void => {
  if (typeof fn === "function") {
    Promise.allSettled([asyncWrap(fn)]);
  } else {
    Promise.allSettled(fn.map(asyncWrap));
  }
};
```

Having to call `sendToBackgroundAndNeverThrow(() => someTask(args))` would make the intentions very clear. Those calls can then also be more easily found and migrated whenever background task queues are implemented.

nicktrn:
Closing as this was rather premature, and now outdated.

keithwillcode:
> Closing as this was rather premature, and now outdated.

This is a great initiative @nicktrn. Want to pick it back up? 

/bounty 50

packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -354,29 +357,27 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
-              eventTypeId: chEventType.id,
-            },
-          });
-        });
-      }
+    await ctx.prisma.workflowsOnEventTypes.createMany({
+      data: activeOnEventTypes.map((eventType) => ({
+        workflowId: id,
+        eventTypeId: eventType.id,
+      })),
```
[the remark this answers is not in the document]
hariombalhara: Let's do it as a separate performance optimization in another PR.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -257,13 +257,13 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
         },
       });
 
-      steps.forEach(async (step) => {
+      const pSteps = steps.map(async (step) => {
```
hariombalhara, line 260:
```suggestion
      const promiseSteps = steps.map(async (step) => {
```
Let's be clear here.

nicktrn:
@keithwillcode Sure thing, will try to resolve those merge conflicts first of all. It's been a while.

There are some decisions to be made on how to proceed though:

1. Should this PR only deal with async forEach or un-awaited promises in general?
2. I can replace all occurences with `await Promise.all()` but this could mean adding execution time to very commonly accessed routes. Is this desired?

An alternative could be to add a wrapper as suggested in https://github.com/calcom/cal.com/pull/10157#issuecomment-1638997077 for now, and include error logging.

nicktrn:
@hariombalhara thanks for the review pointers so far, will help to keep this focused.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
hariombalhara: That's interesting. @CarinaWolli can you confirm that the flow was missing whatsapp here.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -455,13 +453,16 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
+      const pDeleteReminders = remindersToUpdate.map((reminder) => {
         if (reminder.method === WorkflowMethods.EMAIL) {
           deleteScheduledEmailReminder(reminder.id, reminder.referenceId);
         } else if (reminder.method === WorkflowMethods.SMS) {
           deleteScheduledSMSReminder(reminder.id, reminder.referenceId);
+        } else if (reminder.method === WorkflowMethods.WHATSAPP) {
+          deleteScheduledWhatsappReminder(reminder.id, reminder.referenceId);
         }
       });
+      await Promise.all(pDeleteReminders);
```
hariombalhara, line 465: How do you think about wrapping the code itself in Promise.all vs storing the result in `pDeleteReminders` and then doing await.

I find the earlier approach easier to read.

hariombalhara: So this becomes, 
```
await Promise.all(remindersToUpdate.map((reminder) => {
        if (reminder.method === WorkflowMethods.EMAIL) {
          deleteScheduledEmailReminder(reminder.id, reminder.referenceId);
        } else if (reminder.method === WorkflowMethods.SMS) {
          deleteScheduledSMSReminder(reminder.id, reminder.referenceId);
        } else if (reminder.method === WorkflowMethods.WHATSAPP) {
          deleteScheduledWhatsappReminder(reminder.id, reminder.referenceId))
```

packages/app-store/zapier/lib/nodeScheduler.ts
```diff
@@ -57,38 +58,39 @@ export async function cancelScheduledJobs(
   appId?: string | null,
   isReschedule?: boolean
 ) {
-  try {
-    let scheduledJobs = booking.scheduledJobs || [];
+  let scheduledJobs = booking.scheduledJobs || [];
+  if (!booking.scheduledJobs) return;
```
hariombalhara, line 62: What's the point of fallback initialization with empty array if we are still early returning in that case.

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -647,11 +657,13 @@ async function handler(req: CustomRequest) {
   });
 
   // delete scheduled jobs of cancelled bookings
+  // FIXME: async calls into ether
```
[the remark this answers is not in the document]
hariombalhara: Let's tackle these unexplicit async `forEach` in a separate PR. I guess we should use `allSettled` in here to identify which booking's scheduledJob/reminders can't be cancelled and let other cancellations go on.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
nicktrn: I think there have been quite a few changes here since the last merge, probably fixed now!

packages/trpc/server/routers/viewer/workflows/update.handler.ts
nicktrn: Sure that works. For longer mapping functions I tend to prefer prior assignment, but this is rather short.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
nicktrn: Not fixed on main, will revert it for now.

hariombalhara:
> @keithwillcode Sure thing, will try to resolve those merge conflicts first of all. It's been a while.
> 
> There are some decisions to be made on how to proceed though:
> 
> 1. Should this PR only deal with async forEach or un-awaited promises in general?
> 2. I can replace all occurences with `await Promise.all()` but this could mean adding execution time to very commonly accessed routes. Is this desired?
> 
> An alternative could be to add a wrapper as suggested in [#10157 (comment)](https://github.com/calcom/cal.com/pull/10157#issuecomment-1638997077) for now, and include error logging.

1. I think we can do it in two parts because unawaited promises in general is difficult to find out without eslint plugin setup(I guess?)
2. That should be fine IMHO. The alternate approach isn't easy to read and difficult to reason a bit. We should go for that if that really makes sense.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
nicktrn: Will actually revert the whole thing. One of those implicit async calls that could probably do better with `Promise.allSettled()` in another PR. Error handling should really be discussed.

packages/trpc/server/routers/viewer/workflows/update.handler.ts
hariombalhara: I was also thinking that maybe classes would make sense here. We can have classes for all these channels(Whatsapp, SMS, Email) and the code here would simply do `channel.deleteReminder`. This would avoid having these if elses everywhere. This code is just too much involved with what's going on.

But classes might increase complexity in some other way.

packages/trpc/server/routers/viewer/bookings/requestReschedule.handler.ts
```diff
@@ -198,17 +198,20 @@ export const requestRescheduleHandler = async ({ ctx, input }: RequestReschedule
...
-    if (bookingRef.type.endsWith("_calendar")) {
-      const calendar = await getCalendar(credentialsMap.get(bookingRef.type));
-      return calendar?.deleteEvent(bookingRef.uid, builder.calendarEvent, bookingRef.externalCalendarId);
-    } else if (bookingRef.type.endsWith("_video")) {
-      return deleteMeeting(credentialsMap.get(bookingRef.type), bookingRef.uid);
-    }
-  });
-  await Promise.all(pCancellations);
+  // FIXME: error-handling
+  await Promise.allSettled(
```
nicktrn, line 203: Think this one is better suited for `allSettled`. Failure to delete a meeting should probably not prevent reschedule.

nicktrn:
> 1. I think we can do it in two parts because unawaited promises in general is difficult to find out without eslint plugin setup(I guess?)
> 2. That should be fine IMHO. The alternate approach isn't easy to read and difficult to reason a bit. We should go for that if that really makes sense.

1. ESLint will definitely make this easier!
2. Have gone with that for now.

Have to say I'm quite worried by this PR. Will definitely need to be followed up with proper error handling / retries, at the very least additional logging.

A big issue is that those async calls are now back in function scope, i.e. any rejected promise will prevent normal return. The good thing is this should highlight previously hidden errors.

Are there any plans for background tasks? Think I saw some mention of redis-backed solutions a while back. Could also consider [Trigger.dev](https://trigger.dev/).

packages/trpc/server/routers/viewer/workflows/update.handler.ts
CarinaWolli: > That's interesting. @CarinaWolli can you confirm that the flow was missing whatsapp here.

Yes, definitely looks like this was missing here. Thanks for finding that 🙏

nicktrn: Back in it goes :smile:

keithwillcode:
> > @keithwillcode Sure thing, will try to resolve those merge conflicts first of all. It's been a while.
> > There are some decisions to be made on how to proceed though:
> > 
> > 1. Should this PR only deal with async forEach or un-awaited promises in general?
> > 2. I can replace all occurences with `await Promise.all()` but this could mean adding execution time to very commonly accessed routes. Is this desired?
> > 
> > An alternative could be to add a wrapper as suggested in [#10157 (comment)](https://github.com/calcom/cal.com/pull/10157#issuecomment-1638997077) for now, and include error logging.
> 
> 1. I think we can do it in two parts because unawaited promises in general is difficult to find out without eslint plugin setup(I guess?)
> 2. That should be fine IMHO. The alternate approach isn't easy to read and difficult to reason a bit. We should go for that if that really makes sense.

Agree with this approach 🙏🏼. @nicktrn make sure you claim this PR so you get the bounty pay out when it merges

nicktrn:
Ah yes, thanks @keithwillcode. Appreciate the gesture but really wasn't necessary!

zomars (dismissed):
LGTM. Thank you for your contribution 🙏 

Ship it

zomars (approved):

PR #10163, opened 2023-07-14
fix: Embed  `theme` not working using Embed API

### LINKED ISSUE #10187: [CAL-2177] Theme Issues with New Booker
* Theme specified using Embed API stopped working
* Booker Pages' themes are affected by App Theme. When App is opened in the same browser instance as booker page. One affects the other. It is a regression.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2177](https://linear.app/calcom/issue/CAL-2177/theme-issues-with-new-booker)</sub>

hariombalhara:
## What does this PR do?

Fixes #10187
See [Tests Done](https://www.loom.com/share/f03e0191b60143d8b45a505042dbfa11)

## Type of change

## How should this be tested?
- [x] Configure embed to use `dark` theme and verify that dark theme is shown on event booking page(when user has light theme set). This is failing in main
- Additional Tests for embed to avoid any new regression
	- [x] - Configure "auto" theme using embed API and see it reacts to system theme
	- [x] - Don't configure any theme and see that "light" theme is shown even when we switch system theme(Because User has configured light theme in App)
-  Tests outside embed to avoid any new regression
	- [x] - See that light theme is shown even after switching system theme
	- [x] - Now, switch the user theme to dark and see that it reflects the change. 

## Mandatory Tasks

 [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.


packages/features/bookings/Booker/utils/use-brand-colors.tsx
```diff
@@ -17,10 +15,7 @@ export const useBrandColors = ({
     lightVal: brandColor,
     darkVal: darkBrandColor,
   });
-  useCalcomTheme(brandTheme);
-  const { setTheme } = useTheme(theme);
 
-  useEffect(() => {
-    if (theme) setTheme(theme);
```
hariombalhara, line 24: We shouldn't set theme here. useTheme takes care of setting the theme automatically considering embed preferences.

apps/web/lib/app-providers.tsx
```diff
@@ -48,6 +48,12 @@ type AppPropsWithChildren = AppProps & {
   children: ReactNode;
 };
 
+const getEmbedNamespace = (query: Record<string, string>) => {
+  // Mostly embed query param should be available on server. Use that there.
+  // Use the most reliable detection on client
+  return typeof window !== "undefined" ? window.getEmbedNamespace() : query.embed;
```
hariombalhara, line 54: It avoids possible flash of different theme by getting the correct storage key from backend.
It wasn't possible to do reliably earlier because embed query param was getting lost. After recent fixes, embed query param is reliably forwarded throughh booking page navigations

apps/web/components/ui/form/Select.tsx
```diff
@@ -30,7 +30,7 @@ function Select<
   Group extends GroupBase<Option> = GroupBase<Option>
 >({ className, ...props }: SelectProps<Option, IsMulti, Group>) {
   const [mounted, setMounted] = useState<boolean>(false);
-  const { resolvedTheme, forcedTheme } = useTheme();
+  const { resolvedTheme, forcedTheme } = useGetTheme();
```
hariombalhara, line 33: `useTheme` by default won't return these values. This is to avoid accidentally modifying theme of a page where the purpose is just to get the values.
So, we have a separate hook to just get the values.

apps/web/lib/app-providers.tsx
```diff
@@ -158,10 +164,10 @@ function getThemeProviderProps({
     ? ThemeSupport.None
     : ThemeSupport.App;
 
-  const isBookingPageThemSupportRequired = themeSupport === ThemeSupport.Booking;
+  const isBookingPageThemeSupportRequired = themeSupport === ThemeSupport.Booking;
```
hariombalhara, line 167: Fixes spelling typo

apps/web/pages/[user]/[type].tsx
```diff
@@ -40,6 +40,7 @@ export default function Type({ slug, user, booking, away, isBrandingHidden, resc
   );
 }
 
+Type.isBookingPage = true;
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:43]: New Booker pages were missing this which was causing them to use the same storage key as app and thus app could override booking pages theme if opened side by side.

apps/web/pages/booking/[uid].tsx
```diff
@@ -278,7 +278,7 @@ export default function Success(props: SuccessProps) {
 
   // This is a weird case where the same route can be opened in booking flow as a success page or as a booking detail page from the app
   // As Booking Page it has to support configured theme, but as booking detail page it should not do any change. Let Shell.tsx handle it.
-  useTheme(isSuccessBookingPage ? props.profile.theme : undefined);
+  useTheme(isSuccessBookingPage ? props.profile.theme : "system");
```
Comment (hariombalhara) [inline apps/web/pages/booking/[uid].tsx:281]: useTheme now explicilty needs "system" to be passed to support system theme instead of applying system theme when undefined or no argument is passed. This is more intuitive and less error prone.

packages/embeds/embed-core/playground.ts
```diff
@@ -331,7 +332,7 @@ Cal("init", "routingFormDark", {
 
 if (only === "all" || only == "ns:floatingButton") {
   Cal.ns.floatingButton("floatingButton", {
-    calLink: "pro",
+    calLink: calLink || "pro",
```
hariombalhara, line 335: Allow opening any link in floating popup playground. Used by e2e.

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
```diff
@@ -102,56 +102,6 @@ test.describe("Popup Tests", () => {
...
-    });
-
-    if (!embedIframe) {
-      throw new Error("Embed iframe not found");
-    }
-
-    const html = embedIframe.locator("html");
-    await expect(html).toHaveAttribute("class", "dark");
-  });
-
```
hariombalhara, line 154: Moved these tests in separate describe block

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
```diff
@@ -186,4 +137,112 @@ test.describe("Popup Tests", () => {
...
+        const html = embedIframe.locator("html");
+        // Expect "dark" theme as configured in App for pro user.
+        await expect(html).toHaveAttribute("class", "dark");
+        const { uid: bookingId } = await bookFirstEvent("pro", embedIframe, page);
+        const booking = await getBooking(bookingId);
+
+        expect(booking.attendees.length).toBe(3);
+      });
+
+      test("should open embed iframe according to system theme when configured with 'auto' theme using Embed API", async ({
```
hariombalhara, line 171: Added new 'auto' theme embed API test

packages/embeds/embed-core/playwright/tests/action-based.e2e.ts
```diff
@@ -186,4 +137,112 @@ test.describe("Popup Tests", () => {
...
+
+        if (!embedIframe) {
+          throw new Error("Embed iframe not found");
+        }
+
+        const html = embedIframe.locator("html");
+        await expect(html).toHaveAttribute("class", "dark");
+      });
+
+      test("should open embed iframe(Event Booking Page) with dark theme when configured with dark theme using Embed API", async ({
```
hariombalhara, line 223: This would have been a failing test in main.

packages/lib/hooks/useTheme.tsx
```diff
@@ -2,29 +2,28 @@ import { useTheme as useNextTheme } from "next-themes";
...
   const { resolvedTheme, setTheme, forcedTheme, theme: activeTheme } = useNextTheme();
   const embedTheme = useEmbedTheme();
 
   useEffect(() => {
-    // If themeToSet is not provided the app intends to not apply a specific theme
-    if (!themeToSet) {
-      // But if embedTheme is set then the Embed API intends to apply that theme or it wants "system" theme which is the default
-      setTheme(embedTheme || "system");
+    // Undefined themeToSet allow the hook to be used where the theme is fetched after calling useTheme hook
+    if (getOnly || themeToSet === undefined) {
```
hariombalhara, line 19: This is important to handle undefined so that we can ignore the hook call with usages like
profile?.theme where it would be undefined till profile is fetched.

packages/lib/hooks/useTheme.tsx
```diff
@@ -33,10 +32,25 @@ export default function useTheme(themeToSet?: Maybe<string>) {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [themeToSet, setTheme, embedTheme]);
-  return {
-    resolvedTheme,
-    setTheme,
-    forcedTheme,
-    activeTheme,
-  };
+
+  if (getOnly) {
```
hariombalhara, line 36: As mentioned earlier. We no more return these values unless getOnly is passed as true, so that dev doesn't accidentally use useTheme for getting the values. There is useGetTheme for that.

Also, removed setTheme method because we never want theme to be set from outside as that won't handle embed theme.

hariombalhara (commented):
Ready for Review

zomars (approved):
[cool](https://app.graphite.dev/macros/calcom)

PR #10184, opened 2023-07-16
fix: iw to he (hebrew)

PeerRich:
seems like we use iw right now which is modern Hebrew but in crowdin we are using traditional hebrew which is "he"


PeerRich:
tested and works. pls approve a

sean-brydon (approved):

PR #10185, opened 2023-07-16
chore: fixing some initial re-renders 

sean-brydon:
Fixes inital re-renders in event-types pages. 

Will carry on adding to this PR/a new PR with additional fixes changes


apps/web/components/eventtype/EventAvailabilityTab.tsx
```diff
@@ -71,86 +71,91 @@ const format = (date: Date, hour12: boolean) =>
...
-  selectedScheduleValue,
-}: {
-  isManagedEventType: boolean;
-  selectedScheduleValue: AvailabilityOption | undefined;
-}) => {
-  const { data: loggedInUser } = useMeQuery();
-  const timeFormat = loggedInUser?.timeFormat;
-  const { t, i18n } = useLocale();
-  const { watch } = useFormContext<FormValues>();
+const EventTypeScheduleDetails = memo(
```
sean-brydon, line 74: Just wrapped this in a memo - was only being re-rendered by the parent component and not any props that would cause a re-render

packages/features/shell/Shell.tsx
```diff
@@ -791,15 +791,18 @@ const getOrganizationUrl = (slug: string) =>
...
-  const router = useRouter();
   const orgBranding = useOrgBrandingValues();
-  const publicPageUrl = orgBranding?.slug ? getOrganizationUrl(orgBranding?.slug) : "";
+
+  const publicPageUrl = useMemo(() => {
+    if (!user?.organizationId) return `${process.env.NEXT_PUBLIC_WEBSITE_URL}/${user?.username}`;
+    const publicPageUrl = orgBranding?.slug ? getOrganizationUrl(orgBranding?.slug) : "";
+    return publicPageUrl;
+  }, [orgBranding?.slug, user?.organizationId, user?.username]);
+
```
sean-brydon, line 801: Refactored to use memo for this - we didnt need to calculate each time + re-render if there wasnt a org. 

(default should be no org imho so i made these changes to default to no org value)

packages/features/shell/Shell.tsx
```diff
@@ -898,11 +900,6 @@ function SideBar({ bannersHeight, user }: SideBarProps) {
                   isLocaleReady ? "hover:bg-emphasis hover:text-emphasis" : "",
                   index === 0 && "mt-3"
                 )}
-                aria-current={
-                  defaultIsCurrent && defaultIsCurrent({ item: { href: item.href }, router })
-                    ? "page"
-                    : undefined
```
sean-brydon, line 904: Can't be current with the options 

view_public_page -> opens in a new tab
copy_public_page_link -> copies and isnt a page
settings -> layout changes and so no point doing this calc

packages/features/tips/Tips.tsx
```diff
@@ -175,3 +172,5 @@ export default function Tips() {
     </div>
   );
 }
+
+export default memo(Tips);
```
sean-brydon, line 176: This just renders from internal data - no point re-rendering on every parent change (since this was in shell it was re-rendering a lot)

keithwillcode (changes requested):
Looks good. 

There’s some type fixes failing though.

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -20,13 +19,13 @@ export function useViewerI18n() {
...
  */
-const I18nLanguageHandler = (): null => {
+const useI18nLanguageHandler = () => {
   const { i18n } = useTranslation("common");
   const locale = useViewerI18n().data?.locale || i18n.language;
 
   useEffect(() => {
     // bail early when i18n = {}
-    if (isEmpty(i18n)) return;
+    if (Object.keys(i18n).length === 0) return;
```
sean-brydon, line 28: No need for lodash here

apps/web/components/PageWrapper.tsx
```diff
@@ -29,6 +29,7 @@ const calFont = localFont({
 });
 
 function PageWrapper(props: AppProps) {
+  useI18nLanguageHandler();
```
sean-brydon, line 32: Converted to a hook - it was a component always returning null

packages/features/insights/filters/EventTypeList.tsx
```diff
@@ -14,7 +16,7 @@ const mapEventTypeToOption = (eventType: EventType): Option => ({
   label: eventType.teamId ? `${eventType.title} (${eventType.team?.name})` : eventType.title,
 });
 
-export const EventTypeList = () => {
+export const EventTypeList = memo(() => {
```
sean-brydon, line 19: Parents were causing this to re-render on prop change - we didnt need this as all state is internal here

packages/lib/hooks/useTraceUpdate.tsx
```diff
@@ -0,0 +1,21 @@
+import { useRef, useEffect } from "react";
+
+export function useTraceUpdate(props: { [s: string]: unknown } | ArrayLike<unknown>) {
```
sean-brydon, line 3: Helper to be used while debugging - call it in any component and see why it it updated/ how many times it re-rendered

keithwillcode (dismissed):

zomars (dismissed):
Definitively a step in the right direction. Awesome work @sean-brydon 

Ship it

PeerRich (approved):

PR #10190, opened 2023-07-17
fix: Move to minmax instead of auto grid

sean-brydon:
Fixes the issue on safari where grid template rows - auto breaks as they dont support it
Before:
CleanShot 2023-07-17 at 13 02 54

After:
CleanShot 2023-07-17 at 13 01 49


PeerRich (approved):
works. good stuff

hariombalhara (approved):

hariombalhara (commented):
@sean-brydon be like
[cathack](https://app.graphite.dev/macros/calcom)

PR #10199, opened 2023-07-17
fix: issues on embed popup

gitstart-calcom:
The empty screen problem is solved just by running the embed-dev command in the packages/embeds/embed-core folder(`yarn workspace @calcom/embed-core dev`). The problem only happens when the files haven't been generated yet, so the API doesn't work, the files are generated in `apps/web/public/embed/`, once generated, there is no need to run the command again (unless you are editing the corresponding .ts files).

## DEMO

https://github.com/calcom/cal.com/assets/121884634/d6af59fa-4e43-4cb8-a56d-62bb03b9cd32


apps/web/pages/event-types/index.tsx
```diff
@@ -925,12 +925,12 @@ const EventTypesPage = () => {
 
                 {data.eventTypeGroups.length === 0 && <CreateFirstEventTypeView />}
 
-                <EmbedDialog />
                 {router.query.dialog === "duplicate" && <DuplicateDialog />}
               </>
             );
           }}
         />
+        <EmbedDialog />
```
gitstart-calcom, line 933: This solves the problem of the popup appearing and disappearing.

In `apps/web/components/eventtype/EventTypeSingleLayout.tsx`

The component is a direct child of `Shell`, and the problem does not happen on the single event page.

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -42,7 +43,12 @@ export function Header({
     [setLayout, layout]
   );
 
-  if (isMobile || !enabledLayouts) return null;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => {
+    setIsEmbed(window?.isEmbed?.() ?? false);
+  });
+
+  if (isMobile || isEmbed || !enabledLayouts) return null;
```
gitstart-calcom, line 51: Once the files are generated, it will be possible to see the calendar, but `Next.js` complains about a hydration error, and this happens because this component used the window before being defined. Was inserted this `useEffect` because in the previous code the function was called on every render

hariombalhara, line 51: Why have we moved this check in here? We don't want to hide the entire header. Just the layout

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -42,7 +43,12 @@ export function Header({
     [setLayout, layout]
   );
 
-  if (isMobile || !enabledLayouts) return null;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => {
```
hariombalhara, line 47: I would prefer to not use useEffect. Is there a way to fix the hydration error without useEffect ?
https://react.dev/learn/you-might-not-need-an-effect

hariombalhara (dismissed):
Thanks @gitstart. Requesting 2 changes wrt hydration error fix.

The hydration thing isn't a requirement for this fix. If you want you can open a separate PR as well.

packages/features/bookings/Booker/components/Header.tsx
gitstart-calcom: We will make this changes

packages/features/bookings/Booker/components/Header.tsx
gitstart-calcom: We will try to fix it without useEffect

PeerRich:
@gitstart-calcom pls address @hariombalhara's feedback and look into merge conflict

PeerRich (commented):
@Jaibles will provide an svg for the email embed dialog

gitstart-calcom:
Hi @hariombalhara, we solved the conflicts and made the requested changes here. Now we are fixing the problem without using useEffect 😄

hariombalhara:
@gitstart I am not sure what this PR is fixing now. 
Can you add a before and after loom? Also, update the PR description accordingly.


I don't know how the following description relates to what is being fixed.
"The empty screen problem is solved just by running the embed-dev command in the packages/embeds/embed-core folder(yarn workspace @calcom/embed-core dev). The problem only happens when the files haven't been generated yet, so the API doesn't work, the files are generated in apps/web/public/embed/, once generated, there is no need to run the command again (unless you are editing the corresponding .ts files)."

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -146,5 +147,13 @@ const LayoutToggle = ({
...
-  return <ToggleGroup onValueChange={onLayoutToggle} defaultValue={layout} options={layoutOptions} />;
+  const ToggleGroupLoader = (): JSX.Element => (
+    <ToggleGroup onValueChange={onLayoutToggle} defaultValue={layout} options={layoutOptions} />
+  );
+
+  const ToggleGroupNoSSR = dynamic(() => Promise.resolve(ToggleGroupLoader), {
+    ssr: false,
+  });
+
+  return <ToggleGroupNoSSR />;
```
hariombalhara, line 158: Why are we dynamically loading ToggleGroup ?

hariombalhara (changes requested):
Left some comments.

PR #10201, opened 2023-07-17
fix: customTemplate() function fails when no locale is given

CarinaWolli:
## What does this PR do?

Fixes error `Something went wrong RangeError: Incorrect locale information provided at new DateTimeFormat` that sometimes was thrown when `customTemplate()` was called. We now fall back to "en" if no locale is given. 

## Type of change


packages/trpc/server/routers/viewer/workflows/activateEventType.handler.ts
```diff
@@ -162,21 +162,21 @@ export const activateEventTypeHandler = async ({ ctx, input }: ActivateEventType
...
         }),
         organizer: booking.user
           ? {
               name: booking.user.name || "",
               email: booking.user.email,
               timeZone: booking.user.timeZone,
-              language: { locale: booking.user.locale || "" },
+              language: { locale: booking.user.locale || "en" },
             }
           : { name: "", email: "", timeZone: "", language: { locale: "" } },
```
alannnc, line 175: @CarinaWolli Should this too be locale: ""?
```suggestion
          : { name: "", email: "", timeZone: "", language: { locale: "en" } },
```

alannnc (approved):

shivamklr (commented):
lgtm.

PR #10203, opened 2023-07-17
fix: Fix bookingPage UI issues in dark mode 

gitstart-calcom:
Here is the GitStart Ticket for this pull request: [fix-bookingPage](https://developers.gitstart.com/client/calcom/ticket/fix-bookingPage)

[Figma](https://www.figma.com/file/9MOufQNLtdkpnDucmNX10R/Cal-DS---Components?type=design&node-id=784-39271&mode=design)

## DEMO

https://github.com/calcom/cal.com/assets/121884634/38396671-da6f-499d-84fa-21f6660c0144


packages/ui/form/radio-area/Radio.tsx
```diff
@@ -20,7 +20,7 @@ export const Radio = (props: RadioGroupPrimitive.RadioGroupItemProps & { childre
 export const Indicator = ({ disabled }: { disabled?: boolean }) => (
   <RadioGroupPrimitive.Indicator
     className={classNames(
-      "after:bg-default dark:after:bg-inverted relative flex h-full w-full items-center justify-center rounded-full bg-black after:h-[6px] after:w-[6px] after:rounded-full after:content-['']",
+      "after:bg-default relative flex h-full w-full items-center justify-center rounded-full bg-black after:h-[6px] after:w-[6px] after:rounded-full after:content-['']",
```
sean-brydon, line 23: I think we may need to take into consideration the active state for this. 

A user (even tho warned not to) can set a really bad accessable colour that doesnt pass the contrast checks. 

we could use something like dark:after:bg-brand-accent (not sure if its in our tailwind config but its at --cal-brand-accent) 

Sine we use brand colour for the selected BG it would make sense to have the right contrast colour there? Does that make sense

gitstart-calcom: Yes it does. Thanks @sean-brydon

sean-brydon (approved):
LGTM - thanks for quick action on the changes :) Nice job!

gitstart-calcom:
Hi @sean-brydon thanks for the accurate feedback! 👍 😄

PR #10209, opened 2023-07-18
fix: booking error on invalid timezone

### LINKED ISSUE #10189: [CAL-2180] Booking with invalid/unsupported timezones goes through
### Issue Summary

Currently, it is possible for a booking to go through with an unsupported Timezone as we can define unsupported timezone in the Booking API POST request in Timezone, and it goes through. Later, when one attempts to open the [cal.com/bookings](http://cal.com/bookings) page, it would hard crash.

### Steps to Reproduce

1. Create a booking using the API with an unsupported Timezone `eg: America/Pacific`. Note that this isn't listed in the UI, but one can create a booking with this timezone via API
2. Once the booking goes through, open `cal.com/bookings` and that should return in hard crash

### Actual Results

* The booking goes through even with unsupported Timezone and later crashes in UI

### Expected Results

Ideally,

* we want to be able to parse and confirm that the timezone passed is supported. If not, we should return error and the booking shouldn't go through.
* We should also handle such situations in the UI a bit better, instead of hard crashing. Also, ideally, we should provide a nice mapping somewhere which allows the user to find out what their timezone is in our DB, like the closest supported alternative

### Technical details

* In production

### Evidence

Reported by a couple of users, confirmed that we **do not** currently have ways to handle it in place.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2180](https://linear.app/calcom/issue/CAL-2180/booking-with-invalidunsupported-timezones-goes-through)</sub>

jaypitroda12:
## What does this PR do?

- Fixes #10189 

## Type of change

## How should this be tested?

  - What are the minimal test data to have? Invalid timezone
  - What is expected (happy path) to have (input and output)? Booking API throws invalid timezone message 
  - Any other important info that could help to test that PR - None

## Mandatory Tasks


jaypitroda12:
@zomars Initially I used `dayjs().tz(timezone)` to validate timezone, but I found an existing zod validation function and used that instead.

packages/prisma/zod-utils.ts
```diff
@@ -13,6 +13,7 @@ import type {
   ZodTypeAny,
 } from "zod";
 
+import { timeZone } from "@calcom/api/lib/validations/shared/timeZone";
```
Udit-takkar, line 16: I don't think we use tzdata package which is inside /shared/timeZone file. we use react-timezone-select to get the lost of available timezones

jaypitroda12: I also had second thoughts about using that package. Here's an alternate version without using list of all timezones.

```typescript
  timeZone: z.string().refine(
    (value: string) => {
      try {
        dayjs().tz(value);
        return true;
      } catch (error) {
        return false;
      }
    },
    { message: "Invalid timezone" }
  )
```

image

Udit-takkar: @jaypitroda12 we use react-timezone-select and city-timezones in TimezoneSelect.tsx. I am not sure if  dayjs().tz(value) would work exactly same as them

jaypitroda12: @Udit-takkar day.js uses has IANA database for timezones [ref](https://day.js.org/docs/en/timezone/timezone)

both `react-timezone-select` and `city-timezones` has different set of timezones from what I've checked, so maybe we can use `city-timezones` in this case for validation?

alishaz-polymath: @Udit-takkar I think it is best if we take consider the set we use in parsing the timezone when creating a booking in the backend, or in in the `cal.com/bookings` page instead, no? The issue was highlighted by API users alone so far. Plus, we can potentially adjust react-timezone-select to only include the supported timezones which should arguably be the easier thing to do 🤔

jaypitroda12: @alishaz-polymath thank you for taking over the critical part. 🙌🏼 

I've updated the PR, is there something more to be covered for the issue?

packages/prisma/zod-utils.ts
```diff
@@ -202,7 +203,7 @@ export const bookingCreateBodySchema = z.object({
   rescheduleUid: z.string().optional(),
   recurringEventId: z.string().optional(),
   start: z.string(),
-  timeZone: z.string(),
+  timeZone: z.string().refine((value: string) => isSupportedTimeZone(value), { message: "Invalid timezone" }),
```
alishaz-polymath, line 206: I wish we could use a translation string here instead, but it's a minor NIT 🙏

alishaz-polymath (approved):
Left a minor NIT, but can be revisited in a follow up.

Thanks for this fix @jaypitroda12 , amazing work 🙌 
🚀 🚀

emrysal (approved):

PR #10210, opened 2023-07-18
fix: Avoid resetting state one every render.

### LINKED ISSUE #10087: Preview State in Embed Dialog is getting revert to default state after changing the tabs
### Issue Summary

The preview state is reverted to the default state after changing the tab. All the selected colors and themes are being lost to the default state. The main issue is due to the state being managed by the query params, and because of this, the EmbedDialog Component is being rendered repeatedly after we change the tab.

### Steps to Reproduce

1. Go to event types and click on more options and select the embed option
2. The dialog box will open, change the color and theme and select the new tab  (preview or react).
3. the preview state gets reverted to the default one.

### Actual Results

  - The state is being lost when we change the tab.

### Expected Results

  - The state should be preserved even when changing the tab.

### Evidence

  - Below is the issue
cal-com-embed-dialog-state-error

hariombalhara:
## What does this PR do?
Fixes #10087
Removes usage of withQuery(which is deprecated). withQuery returns new function instance on each call which causes the component to lose it's state on every render.

Before: https://user-images.githubusercontent.com/72548403/252946101-bc1b985a-eced-4d3f-b354-e3ac96ac7bb8.gif

After fix [demo](https://www.loom.com/share/05989ca73e7e4fe6945f4f8d79c138b0) 

## Type of change

## How should this be tested?
- See the loom

## Mandatory Tasks

## Checklist


sean-brydon (approved):
Nice! This was on my todos actually cause i hate withquery :P 

Tested and it seems to work fine for me! I think in the future we should spit up loading eventTypes for users and times so we can do them in separate queries. (Not related to this PR at all :P )

shivamklr (commented):
Tested locally, works great.

PR #10225, opened 2023-07-18
fix: Hotfix-bookings list breaks on unsupported timezone

### LINKED ISSUE #10189: [CAL-2180] Booking with invalid/unsupported timezones goes through

alishaz-polymath:
## What does this PR do?

This PR adds a fallback to unsupported attendee timezone which can be fed via the API, resulting in hard crash of the `/bookings` page in the cal.com UI. Currently we can create a booking with certain unsupported timezones via the API, however that is being sorted as well. The code changes in this PR would still be important for backwards compatibility, otherwise `/bookings/past` would remain broken due to old bookings of this nature.

Partly Fixes #10189 

[Loom Video](https://www.loom.com/share/41d770d44fb147978412398627af6a39?sid=8a742664-8f44-4df5-a26b-41c699dbf7ba)
[Loom Video #2](https://www.loom.com/share/ee9210314ef44887ae65daf0f017212d?sid=f3d854c8-5306-478a-9560-a40b613e8090)

## Type of change

## How should this be tested?

    Create a new booking (either via API or via UI) and make sure that it has a unsupported timezone such as "America/Pacific". You can change the timezone of the attendee by going into the attendee table in prisma studio, if you create the booking via UI. 

     The /bookings/ page containing this booking should crash. Once my changes are applied, it shouldn't crash anymore

## Mandatory Tasks

### DIFF
--- a/packages/lib/date-fns/index.ts
+++ b/packages/lib/date-fns/index.ts
@@ -38,8 +38,12 @@ export const formatTime = (
  * 
  */
 export const isSupportedTimeZone = (timeZone: string) => {
-  const validTimezones = Intl.supportedValuesOf("timeZone");
-  return validTimezones.includes(timeZone);
+  try {
+    dayjs().tz(timeZone);
+    return true;
+  } catch (error) {
+    return false;
+  }
 };
 
 
--- a/packages/lib/date-fns/index.ts
+++ b/packages/lib/date-fns/index.ts
@@ -31,6 +31,19 @@ export const formatTime = (
 };
 
 /**
+ * Checks if a provided timeZone string is recognized as a valid timezone by the Intl API.
+ * 
+ * @param {string} timeZone - The timezone string to be verified.
+ * @returns {boolean} - Returns 'true' if the provided timezone string is recognized as a valid timezone by the Intl API. Otherwise, returns 'false'.
+ * 
+ */
+export const isSupportedTimeZone = (timeZone: string) => {
+  const validTimezones = Intl.supportedValuesOf("timeZone");
+  return validTimezones.includes(timeZone);
+};
+
+
+/**
  * Returns a localized and translated date or time, based on the native
  * Intl.DateTimeFormat available to JS. Undefined values mean the browser's
  * locale will be used.
--- a/packages/lib/date-fns/index.ts
+++ b/packages/lib/date-fns/index.ts
@@ -31,10 +31,10 @@ export const formatTime = (
 };
 
 /**
- * Checks if a provided timeZone string is recognized as a valid timezone by the Intl API.
+ * Checks if a provided timeZone string is recognized as a valid timezone by dayjs.
  * 
  * @param {string} timeZone - The timezone string to be verified.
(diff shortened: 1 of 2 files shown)


packages/lib/date-fns/index.ts
```diff
@@ -30,6 +30,19 @@ export const formatTime = (
...
+ * @param {string} timeZone - The timezone string to be verified.
+ * @returns {boolean} - Returns 'true' if the provided timezone string is recognized as a valid timezone by the Intl API. Otherwise, returns 'false'.
+ * 
+ */
+export const isSupportedTimeZone = (timeZone: string) => {
+  const validTimezones = Intl.supportedValuesOf("timeZone");
+  return validTimezones.includes(timeZone);
+};
+
+
```
alishaz-polymath, line 45: Self explanatory

packages/ui/components/popover/MeetingTimeInTimezones.tsx
```diff
@@ -34,8 +35,12 @@ const MeetingTimeInTimezones = ({
...
 }: MeetingTimeInTimezonesProps) => {
   if (!userTimezone || !attendees.length) return null;
-
-  const attendeeTimezones = attendees.map((attendee) => attendee.timeZone);
+  
+  // If attendeeTimezone is unsupported, we fallback to host timezone. Unsupported Attendee timezone can be used due to bad API booking request in the past | backward-compatibility
+  
+  const attendeeTimezones = attendees.map((attendee) => {
+    return isSupportedTimeZone(attendee.timeZone) ? attendee.timeZone : userTimezone;
+  });
```
alishaz-polymath, line 43: When an unsupported timezone is picked for attendee, we use userTimezone (which is host's timezone) instead.

alishaz-polymath (commented):
Self Review added

emrysal (dismissed):

PeerRich (dismissed):

hariombalhara (approved):
LGTM!! 
There might be other occurences where we might need to fix this but if it solves the immediate problem, it's good to go

PR #10232, opened 2023-07-18
fix: update list for countries that don't support alphanumeric sender ids

CarinaWolli:
## What does this PR do?

This PR adds three missing country codes to the list of countries that don't support alphanumeric sender ids. I compared the whole list from [Twilio](https://support.twilio.com/hc/en-us/articles/223133767-International-support-for-Alphanumeric-Sender-ID) with our existing noAphanumericSenderIdSupport list. 

## Type of change


packages/features/ee/workflows/lib/alphanumericSenderIdSupport.ts
```diff
@@ -79,11 +79,13 @@ const noAlphanumericSenderIdSupport = [
   "+216",
   "+90",
   "+256",
-  "+971",
```
CarinaWolli, line 82: duplicate

PeerRich (approved):

PR #10238, opened 2023-07-18
perf: Event-type config page

### LINKED ISSUE #9944: Loading time after clicking edit
### Issue Summary

Short lag happens when user clicks on the edit button for any event in the event types page. Possible improvement could be to add a loader to improve the user experience.

### Steps to Reproduce

* Go to the event types page.

https://github.com/calcom/cal.com/assets/33975149/80e212d5-021f-4f79-873d-58e9a5b79155

We can add a loader same as given below. It will enhance the user's experience.

image

leog:
## What does this PR do?

Follows some suggestion to reduce re-renders on a couple of high profile components needed for event-type config page to improve performance. One is related to avoid changing children to re-render by componentizing to behave as an atomic block and the other is to avoid relying on objects when passing down properties.

Fixes #9944 

## Type of change

- [X] Performance fix (non-breaking change which intends to improve performance)

## How should this be tested?

Open an Event Type config page, it should have to process less re-renders, reducing the TTI (time-to-interactive).

## Mandatory Tasks

### DIFF
--- a/apps/web/components/Embed.tsx
+++ b/apps/web/components/Embed.tsx
@@ -191,8 +191,8 @@ export default function MyApp() {
       ${uiInstructionCode}
     })();
   }, [])
-  return <Cal 
-    calLink="${calLink}" 
+  return <Cal
+    calLink="${calLink}"
     style={{width:"${width}",height:"${height}",overflow:"scroll"}}
     ${previewState.layout ? "config={{layout: '" + previewState.layout + "'}}" : ""}${
         IS_SELF_HOSTED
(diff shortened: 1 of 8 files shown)


apps/web/components/Embed.tsx
sean-brydon, line 1: These changes needed? or do we have a prettier mis match

sean-brydon (approved):
Changes look good - Thanks

apps/web/components/Embed.tsx
leog: Prettier yeah

PR #10248, opened 2023-07-19
fix: Fixes error for unknown JSON when creating an account 

sean-brydon:
This happened on signup when a user was using a + in their email.

Since we stringify the email in signup api it was being removed/replace with a charater that wasnt valid. The user was being created with a empty username and breaking.

Hopefully preventing a + symbol in here will fix this issue.


keithwillcode (commented):
Feels like we should be able to support these characters instead of just filtering it out or is it too much complication for the value?

sean-brydon:
> Feels like we should be able to support these characters instead of just filtering it out or is it too much complication for the value?

I agree - i know we mentioned in threads supporting "." etc soon also. Maybe that is something we can do instead of this

keithwillcode:
> > Feels like we should be able to support these characters instead of just filtering it out or is it too much complication for the value?
> 
> I agree - i know we mentioned in threads supporting "." etc soon also. Maybe that is something we can do instead of this

Yeah, seems like we should either support symbols or we don't. A more generic solution versus just checking for + would be nice.

shivamklr (approved):
Tested, works. Feel free to review code.

PeerRich:
> Yeah, seems like we should either support symbols or we don't. A more generic solution versus just checking for + would be nice.

@keithwillcode we do allow "-" for example

PR #10271, opened 2023-07-20
chore: Make embed and app-store tests required

hariombalhara:
Make Embed and AppStore tests required.
These tests are skipped automatically if there isn't a related change but if they ran, we should not let a PR merge without them being passing.


keithwillcode (approved):

zomars:
This is breaking checks on main. Will revert.

PR #10279, opened 2023-07-20
fix: Default to Daily if Viable

joeauyeung:
## What does this PR do?

This PR fixes a bug where an event type would default to Daily even if it would throw an error.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a new event type, do not set the location
- Create a booking under that new event type
    - The `EventManager` should not throw an error

## Mandatory Tasks

## Checklist


emrysal (changes requested):
image

packages/core/EventManager.ts
```diff
@@ -94,7 +95,22 @@ export default class EventManager {
...
+        },
+        select: {
+          keys: true,
+          enabled: true,
+        },
+      });
+
+      const calVideoKeys = calVideoKeysSchema.safeParse(calVideo?.keys);
+
+      if (calVideo?.enabled && calVideoKeys.success) evt["location"] = "integrations:daily";
```
joeauyeung, line 112: Add Zod safeParse to ensure that the keys are correct, and that means we don't have to do weird type hacks for `Prisma.JsonValue`

@emrysal

emrysal (dismissed):
Works!

joeauyeung:
> @joeauyeung e2e tests are failing and it is related to your changes

Tests are failing because the result is expecting Daily as the video location. Should I change the tests to not expect it?

CarinaWolli:
> Tests are failing because the result is expecting Daily as the video location. Should I change the tests to not expect it?

Yes, let's change that

CarinaWolli (approved):

PR #10283, opened 2023-07-20
fix: Flaky 2FA tests

joeauyeung:
## What does this PR do?

This PR fixes the "enable 2FA test" by waiting for the enable 2FA dialog box to disappear and wait to see if the enable 2FA switch is set to `true`

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Run "Enable 2FA" test

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/playwright/login.2fa.e2e.ts
+++ b/apps/web/playwright/login.2fa.e2e.ts
@@ -45,9 +45,7 @@ test.describe("2FA Tests", async () => {
       });
 
       await page.waitForSelector(`[data-testid=two-factor-switch]`);
-      await expect(page.locator(`[data-testid=two-factor-switch]`)).toBeTruthy();
-
-      // await expect(page.locator(`[data-testid=two-factor-switch][data-state="checked"]`)).toBeVisible();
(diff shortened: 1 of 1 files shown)


apps/web/playwright/login.2fa.e2e.ts
```diff
@@ -44,7 +44,10 @@ test.describe("2FA Tests", async () => {
         secret: secret!,
       });
 
-      await expect(page.locator(`[data-testid=two-factor-switch][data-state="checked"]`)).toBeVisible();
+      await page.waitForSelector(`[data-testid=two-factor-switch]`);
+      await expect(page.locator(`[data-testid=two-factor-switch]`)).toBeTruthy();
```
joeauyeung, line 48: Rather than just waiting for the switch to appear. Also expect that it is set to `true`

emrysal: I don't think this checks if it is checked- rather whether the locator for the switch is truthy, so whenever it's present. There's an `isChecked` param in page. though.

zomars (approved):
Noice

PR #10285, opened 2023-07-20
fix: Add attendees to getLuckyUser algorithm

emrysal:
## What does this PR do?

Implementing fixed hosts brought in a breaking change to the leastRecently algorithm, excluding non-organizer RR attendees. This fixes that, fixed hosts are not part of the pool so are not added to the availableUsers array.

This is urgent, but it needs tests; so that's a follow up as high priority.


zomars (approved):
LGTM

PR #10288, opened 2023-07-20
fix: 2FA from CAL to GOOGLE

### LINKED ISSUE #7891: [CAL-1318] Disabling 2FA is broken when authenticating with Google
As I am authenticating with Google, I do not have a password.

* Also note the double label.

image

<sub>[CAL-1318](https://linear.app/calcom/issue/CAL-1318/disabling-2fa-is-broken-when-authenticating-with-google)</sub>

leog:
## What does this PR do?

Switching login from CAL credentials to Google having 2FA enabled was failing due to:
1. Password couldn't be empty which is what it was when entering TOTP code in the 2FA form above the user/password form
2. Condition within authorize logic that missed a check if user was indeed using CAL as provider before asking for TOTP code before setting new provider

Also fixes #7891 regarding being able to disable 2FA for converted CAL to Google providers and UI issues.

## Type of change

## How should this be tested?

1. Sign up using a Google email without using "Sign up with Google"
2. Turn on 2FA
3. Logout
4. Sign in using "Sign in with Google"
5. Provide a TOTP code and should work
6. Try to disable 2FA and you will be able to do that

## Mandatory Tasks


apps/web/components/settings/DisableTwoFactorModal.tsx
```diff
@@ -12,7 +12,7 @@ import TwoFactorAuthAPI from "./TwoFactorAuthAPI";
 interface DisableTwoFactorAuthModalProps {
   open: boolean;
   onOpenChange: () => void;
-
+  disablePassword?: boolean;
```
leog, line 15: Needed a way to tell disable 2FA that the password should be asked, as when provider is not CAL, password is now correctly set to null

leog, line 15: Needed a way to tell disable 2FA that the password should be asked, as when provider is not CAL, password is now correctly set to null

apps/web/components/settings/DisableTwoFactorModal.tsx
```diff
@@ -75,19 +76,21 @@ const DisableTwoFactorAuthModal = ({
...
-            />
+            {!disablePassword && (
+              <PasswordField
+                labelProps={{
+                  className: "block text-sm font-medium text-default",
+                }}
+                {...form.register("password")}
+                className="border-default mt-1 block w-full rounded-md border px-3 py-2 text-sm focus:border-black focus:outline-none focus:ring-black"
+              />
+            )}
```
leog, line 87: Removing the password when flag comes in as true

apps/web/components/settings/DisableTwoFactorModal.tsx
```diff
@@ -75,19 +76,21 @@ const DisableTwoFactorAuthModal = ({
...
             <TwoFactor center={false} />
 
             {errorMessage && <p className="mt-1 text-sm text-red-700">{errorMessage}</p>}
           </div>
 
-          <DialogFooter showDivider className="mt-5 sm:mt-4 sm:flex sm:flex-row-reverse">
+          <DialogFooter showDivider className="relative mt-5">
```
leog, line 93: Fixing UI issue

apps/web/components/settings/EnableTwoFactorModal.tsx
```diff
@@ -185,7 +185,7 @@ const EnableTwoFactorModal = ({ onEnable, onCancel, open, onOpenChange }: Enable
               )}
             </div>
           </WithStep>
-          <DialogFooter className="mt-8 sm:flex sm:flex-row-reverse" showDivider>
+          <DialogFooter className="mt-8" showDivider>
```
leog, line 188: Fixing UI issue

apps/web/pages/api/auth/two-factor/totp/disable.ts
```diff
@@ -28,17 +29,19 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     return res.status(401).json({ message: "Not authenticated" });
   }
 
-  if (!user.password) {
+  if (!user.password && user.identityProvider === IdentityProvider.CAL) {
```
leog, line 32: When disabling an account with 2FA with a provider not CAL, no password will come so no need to fail

apps/web/pages/api/auth/two-factor/totp/disable.ts
```diff
@@ -28,17 +29,19 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
   }
 
   if (!user.twoFactorEnabled) {
     return res.json({ message: "Two factor disabled" });
   }
 
-  const isCorrectPassword = await verifyPassword(req.body.password, user.password);
-  if (!isCorrectPassword) {
-    return res.status(400).json({ error: ErrorCode.IncorrectPassword });
+  if (user.password && user.identityProvider === IdentityProvider.CAL) {
```
leog, line 40: When disabling an account with 2FA with a provider not CAL, no password will come so no need to fail when is not correct

apps/web/pages/auth/login.tsx
```diff
@@ -57,7 +57,10 @@ export default function Login({
         .string()
         .min(1, `${t("error_required_field")}`)
         .email(`${t("enter_valid_email")}`),
-      password: z.string().min(1, `${t("error_required_field")}`),
+      password: z
+        .string()
+        .min(1, `${t("error_required_field")}`)
+        .or(z.literal("")),
```
leog, line 63: Password can be empty when entering a TOTP code in the UI, which has the user/password form behind it getting the email prefilled from JWT token

apps/web/pages/settings/security/two-factor-auth.tsx
```diff
@@ -34,14 +34,15 @@ const TwoFactorAuthView = () => {
   if (isLoading) return <SkeletonLoader />;
 
   const isCalProvider = user?.identityProvider === "CAL";
+  const canSetupTwoFactor = !isCalProvider && !user?.twoFactorEnabled;
```
leog, line 37: We need to check if 2FA is already enabled before disabling interacting with the configuration, otherwise converted CAL to GOOGLE users will not be able to disable 2FA

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -115,7 +115,7 @@ const providers: Provider[] = [
         throw new Error(ErrorCode.IncorrectEmailPassword);
       }
 
-      if (user.password || !credentials.totpCode) {
+      if (user.password && !credentials.totpCode) {
```
leog, line 118: We should check if password is correct when signing in when user has provided user and no TOTP code, otherwise treat this as a 2FA login

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -560,7 +560,7 @@ export const AUTH_OPTIONS: AuthOptions = {
                 console.error("Error while linking account of already existing user");
               }
             }
-            if (existingUser.twoFactorEnabled) {
+            if (existingUser.twoFactorEnabled && existingUser.identityProvider === IdentityProvider.CAL) {
```
leog, line 563: This missing check was not letting a user with a different provider than CAL to set the password as null and changing the provider with the new one which is later in the logic

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -600,7 +600,11 @@ export const AUTH_OPTIONS: AuthOptions = {
 
         if (existingUserWithEmail) {
           // if self-hosted then we can allow auto-merge of identity providers if email is verified
-          if (!hostedCal && existingUserWithEmail.emailVerified) {
+          if (
+            !hostedCal &&
+            existingUserWithEmail.emailVerified &&
+            existingUserWithEmail.identityProvider !== IdentityProvider.CAL
```
leog, line 606: This missing check was also not identifying the user a a converted provider later on

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -147,7 +147,11 @@ const useTabs = () => {
       tab.name = user?.name || "my_account";
       tab.icon = undefined;
       tab.avatar = avatar?.avatar || WEBAPP_URL + "/" + session?.data?.user?.username + "/avatar.png";
-    } else if (tab.href === "/settings/security" && user?.identityProvider === IdentityProvider.GOOGLE) {
+    } else if (
+      tab.href === "/settings/security" &&
+      user?.identityProvider === IdentityProvider.GOOGLE &&
+      !user?.twoFactorEnabled
```
leog, line 153: We need to show 2FA option in settings for people with 2FA enabled even if they have a different provider than CAL, otherwise converted provider users couldn't disable 2FA

leog (commented):
Self-review

apps/web/pages/auth/login.tsx
alannnc: This would allow receiving password: ""? Is there a way to use if when case here?
// Pseudo
If (prop.disabledPassword or 2fa flag) {
password = z.string().min(1);
} else {
password = z.literal("");
}

alannnc (dismissed):
Code looks good! Just left a small question.

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -560,7 +560,7 @@ export const AUTH_OPTIONS: AuthOptions = {
                 console.error("Error while linking account of already existing user");
               }
             }
-            if (existingUser.twoFactorEnabled) {
+            if (existingUser.twoFactorEnabled && existingUser.identityProvider === idP) {
```
leog, line 563: This missing check was not letting a user with a different provider than CAL to set the password as null and changing the provider with the new one which is later in the logic

apps/web/pages/auth/login.tsx
```diff
@@ -57,7 +57,7 @@ export default function Login({
         .string()
         .min(1, `${t("error_required_field")}`)
         .email(`${t("enter_valid_email")}`),
-      password: z.string().min(1, `${t("error_required_field")}`),
+      password: !!totpEmail ? z.literal("") : z.string().min(1, `${t("error_required_field")}`),
```
leog, line 60: @alannnc thanks for the suggestion, here it is, if we are in a TOTP code context, we allow empty password, otherwise apply the minimum limit

emrysal (approved):
Let's go!

PR #10289, opened 2023-07-20
fix: team event emails; sending & email subject

emrysal:
## What does this PR do?

* Ensures email is sent to organizer & team members
* Fixes title for > 1 person events; preferring the team name (defaulting to host if not available)
* Removes another case of `avatar`


CarinaWolli (approved):
tested round-robin, collective and dynamic events. Looks all good to me 👍

PR #10296, opened 2023-07-21
chore: enable i18n-unused GH action in `website`

p6l-richard:
## What does this PR do?

- adds `i18n-unused` as a devDep to yarn workspace's root
- adds a config file for `i18n-unused` to the root to make it run in `website`
- `website` is now able to run `yarn i18n-unused display-unused` from the workspace root in a GH Action

Fixes:
Currently, there isn't a way to run a github action with a package dependency in `website`. 
The reason:
- `cal.com`'s yarn removes `website`'s dependencies for commits where `website` isn't checked out locally. 

An example is @sean-brydon's commit, which [removed the i18n-unused](https://github.com/calcom/cal.com/commit/610c86837a5592e69576cfc2525dac70f105912e#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL4797) dependency in `website`: 
image

When Sean checked out `website` locally, yarn added the dependency to the lock file again. However, not every dev in `cal.com` has the latest `website` checked out locally (and probably won't need it), so that I don't see an ergonomic way to ensure a two-way sync for the lock file other than including this in the workspace root as a devDependency (ensuring it won't get removed).

This enables to run: `yarn i18n-unused display-unused` from the workspace root with `website` as the target:
CleanShot 2023-07-21 at 13 33 09

## Type of change

## How should this be tested?

- After the merge, follow up with a PR in `website` to run a GitHub action with `i18n-unused` and have the github action run through

## Mandatory Tasks

### DIFF
--- a/i18n-unused.config.js
+++ b/i18n-unused.config.js
@@ -1,0 +1,35 @@
+const path = require("path");
+
+/**
+ * # Regex for translation keys
+ * This regex matches the following function calls:
+ * 1. `t("<some_key>")`,
+ * 2. `t("<some_key>", {key: "<some_string>"})`, or
+ * 3. `<Trans i18nKey="home_hero_subtitle" t={t}>`
+ *
+ * It also ensure that we don't match any other similar function calls (e.g. `format("dddd")`).
+ *
+ * ## Explanation of the regex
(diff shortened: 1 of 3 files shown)


emrysal (approved):
Regenerated yarn.lock, should be good now

PR #10305, opened 2023-07-22
fix: changes for showing error message in case of INTERNAL_SERVER_ERROR in event types

### LINKED ISSUE #10205: No length limit for duration causes error 500
### Issue Summary

When setting a duration, if the number is higher than 899999999 (INT4 - 32-bit signed integer), it causes error 500 from the server. From the duplicate event type option, it shows a nice notification but from the event type configuration, it shows the DB error happened from the server.

### Steps to Reproduce

1. In an Event type, click on the `Duplicate` option
2. Set `9000000000` as the length value
3. Click on the `Continue` button
4. See the notification error
5. Save it with a valid number, like 15
6. From the Event Type configuration, set the duration to `9000000000`
7. Click on the `Save` button
8. See the notification

### Actual Results

  - An error 500 is thrown on both pages but in the Event type configuration page, the notification is the error from the server

### Expected Results

  - A nice notification should be shown. Ideally, both fields should have a value limitation that can be typed.

### Technical details

  - Tested on `app.cal.dev`
  - Tested on: Firefox (v115.0.2) and Chrome (v114.0.5735.199)

### Evidence

image

image

N-NeelPatel:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/10205

## Type of change

## How should this be tested?

(These steps are from the issue mentioned above)
- In an Event type, click on the Duplicate option
- Set 9000000000 as the length value
- Click on the Continue button
- See the notification error
- Save it with a valid number, like 15
- From the Event Type configuration, set the duration to 9000000000
- Click on the Save button
- See the notification

## Mandatory Tasks

### Before Change
image

### After Change
image


apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -205,6 +205,12 @@ const EventTypePage = (props: EventTypeSetupProps) => {
         message = `${err.data.code}: ${t(err.message)}`;
       }
 
+      if (err.data?.code === "INTERNAL_SERVER_ERROR") {
+        const message = t("unexpected_error_try_again");
+        showToast(message, "error");
+        return;
```
Comment (leog) [inline apps/web/pages/event-types/[type]/index.tsx:211]:
```suggestion
        message = t("unexpected_error_try_again");
```

leog (changes requested):
Thanks for your contribution @N-NeelPatel and for taking care of this edge case to make it nicer to the end user. I suggested a change, with that it should be good to go 👍

apps/web/pages/event-types/[type]/index.tsx
Comment (N-NeelPatel) [reply apps/web/pages/event-types/[type]/index.tsx]: @leog - thanks for reviewing the code, I have made the above suggested change

leog (approved):
Looks good now! Thanks again

PR #10318, opened 2023-07-23
fix: `nameless` in sidebar for orgs

### LINKED ISSUE #10310: [CAL-2218] Organizations: Shows as nameless in top left profile
I just created an org and the org shows as Nameless. This should have the org name

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2218](https://linear.app/calcom/issue/CAL-2218/organizations-shows-as-nameless-in-top-left-profile)</sub>

hariombalhara:
## What does this PR do?

- Revert stale Infinite from getBranding query so that it is updated when org updates. It is easier then manually retching when some data changes. This actually fixes the bug where nameless is shown when a new org is created.
- Replace `useOrgBrandingValues` with `useOrgBranding` which gets the value from Provider. This would avoid unnecessary refetch of getBranding data which doesn't change frequently.

Fixes #10310 

## Type of change

## Mandatory Tasks

## Checklist


packages/features/ee/organizations/context/provider.ts
```diff
@@ -22,19 +22,19 @@ export type OrganizationBranding =
...
  * Accesses the branding for an organization from context.
- *
  * You need to render a <OrgBrandingProvider /> further up to be able to use
  * this component.
+ *
+ * @returns `undefined` when data isn't available yet, `null` when there's an error, and the data(which can be `null`) when it's available
  */
 export function useOrgBranding() {
   const orgBrandingContext = useContext(OrganizationBrandingContext);
-  if (orgBrandingContext === null)
```
hariombalhara, line 35: The value can itself be null if the user isn't an org member. So, make a wrapper object that has orgBrand and that wrapper objet can be checked if it's nullish or not.

sean-brydon (approved):
Tested -> Works great!

Nice work on fetching from provider instread of trpc each time we need it. We could likely do this across the app for a few things!

PR #10346, opened 2023-07-24
fix: Onboarding wizard improvements [CAL-2229]

leog:
## What does this PR do?

When a user created their user through an org invitation, they picked up a username that will then be presented again to be picked in the onboarding wizard. That part was hidden when coming from signup.

<kbd></kbd>

Also, a verification email was being sent to the invited user even after creating an account accepting the invitation clicking on the button of the invitation email, which was redundant, so the verification email was also refactored to no be sent for org signup.

## Type of change

## How should this be tested?

Invite a user to an organization and click the button of the email to create an account. Username in onboarding wizard should not be shown, and verification email should not be sent.

## Mandatory Tasks


zomars (approved):
LGTM

PR #10347, opened 2023-07-24
fix: ratelimit - updates

sean-brydon:
Adds a few changes for rate limiting


packages/lib/rateLimit.ts
```diff
@@ -57,6 +57,12 @@ export function rateLimiter() {
       prefix: "ratelimit",
       limiter: Ratelimit.fixedWindow(10, "60s"),
     }),
+    common: new Ratelimit({
+      redis,
+      analytics: true,
+      prefix: "ratelimit",
+      limiter: Ratelimit.fixedWindow(200, "60s"),
+    }),
```
sean-brydon, line 65: New limiting type that will be used across higher called apis

keithwillcode (approved):

packages/lib/rateLimit.ts
```diff
@@ -57,6 +57,12 @@ export function rateLimiter() {
       prefix: "ratelimit",
       limiter: Ratelimit.fixedWindow(10, "60s"),
     }),
+    common: new Ratelimit({
```
zomars, line 60:
```suggestion
    common: () => new Ratelimit({
```
 We should invoque the limiters otherwise we're loading all of them instead of only the ones needed. Not a big deal since there's only 3 but a good practice for scaling

sean-brydon: Getting some errors doing this - will create issue n follow up!

PR #10357, opened 2023-07-25
fix: Embed booker top margin embed

hariombalhara:
## What does this PR do?
- Removes top margin again from embed which got added after the merge of #9873. It's a weird side effect of that change.
- Now, top margin is removed in a better way, mt-0 should haven't worked already(but it was working somehow) because the margin was coming due to flex align-items center and not through some explicitly added margin.
- Also, fixes the failing embed test in main(which was in that PR but the PR got merged because the embed tests doesn't block the merge). PR would follow soon to make embed tests required.

## Type of change

## How should this be tested?
- Simply go to https://localhost:3000/free/30min/embed and notice that booker is centered vertically. With this fix, it would be back to top.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/embeds/embed-core/playwright/tests/embed-pages.e2e.ts
+++ b/packages/embeds/embed-core/playwright/tests/embed-pages.e2e.ts
@@ -6,17 +6,20 @@ test.describe("Embed Pages", () => {
   test("Event Type Page: should not have margin top on embed page", async ({ page }) => {
     await page.goto("http://localhost:3000/free/30min/embed");
     // Checks the margin from top by checking the distance between the div inside main from the viewport
-    const marginFromTop = await page.evaluate(() => {
-      const mainElement = document.querySelector("main");
-      const divElement = mainElement?.querySelector("div");
+    const marginFromTop = await page.evaluate(async () => {
+      return await new Promise((resolve) => {
+        (function tryGettingBoundingRect() {
+          const mainElement = document.querySelector(".main");
 
-      if (mainElement && divElement) {
-        // This returns the distance of the div element from the viewport
-        const divRect = divElement.getBoundingClientRect();
-        return divRect.top;
-      }
-
-      return null;
+          if (mainElement) {
+            // This returns the distance of the div element from the viewport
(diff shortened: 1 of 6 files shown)


sean-brydon (dismissed):

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -183,9 +183,7 @@ const BookerComponent = ({
             (layout === BookerLayouts.MONTH_VIEW || isEmbed) && "border-subtle rounded-md border",
             !isEmbed && "sm:transition-[width] sm:duration-300",
             isEmbed && layout === BookerLayouts.MONTH_VIEW && "border-booker sm:border-booker-width",
-            !isEmbed && layout === BookerLayouts.MONTH_VIEW && "border-subtle",
-            // We don't want any margins for Embed. Any margin needed should be added by Embed user.
-            layout === BookerLayouts.MONTH_VIEW && isEmbed && "mt-0"
```
hariombalhara, line 188: Removing the margin in the correct way now. It was working accidentally somehow but after an unrelated fix of returning null from the component in a case, it stopped working.

packages/embeds/embed-core/playwright/tests/embed-pages.e2e.ts
```diff
@@ -6,17 +6,20 @@ test.describe("Embed Pages", () => {
...
+      return await new Promise((resolve) => {
+        (function tryGettingBoundingRect() {
+          const mainElement = document.querySelector(".main");
+
+          if (mainElement) {
+            // This returns the distance of the div element from the viewport
+            const mainElBoundingRect = mainElement.getBoundingClientRect();
+            resolve(mainElBoundingRect.top);
+          } else {
+            setTimeout(tryGettingBoundingRect, 500);
```
hariombalhara, line 19: With the recent change in Booker, main element isn't SSRd. SSR has null element rendered. It shows on client and thus we need to wait for the element to be available on client now.

sean-brydon (approved):

PR #10358, opened 2023-07-25
chore: make Embed and AppStore Tests required

hariombalhara:
## What does this PR do?

Make all tests required now. 
A recent regression that could have been avoided if embed tests were required #10357

Side Note: Marked it High Priority as it's quick to merge and super important to have these tests block PR merge considering recent regression as well.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -84,21 +84,18 @@ jobs:
 
   e2e-app-store:
     name: E2E App Store tests
-    if: ${{ needs.changes.outputs.app-store == 'true' }}
     needs: [changes, lint, build]
     uses: ./.github/workflows/e2e-app-store.yml
     secrets: inherit
 
   e2e-embed:
     name: E2E embeds tests
-    if: ${{ needs.changes.outputs.embed == 'true' }}
     needs: [changes, lint, build]
     uses: ./.github/workflows/e2e-embed.yml
     secrets: inherit
 
   e2e-embed-react:
     name: E2E React embeds tests
-    if: ${{ needs.changes.outputs.embed-react == 'true' }}
     needs: [changes, lint, build]
     uses: ./.github/workflows/e2e-embed-react.yml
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yml
```diff
@@ -84,21 +84,18 @@ jobs:
...
   e2e-embed:
     name: E2E embeds tests
-    if: ${{ needs.changes.outputs.embed == 'true' }}
     needs: [changes, lint, build]
     uses: ./.github/workflows/e2e-embed.yml
     secrets: inherit
 
   e2e-embed-react:
     name: E2E React embeds tests
-    if: ${{ needs.changes.outputs.embed-react == 'true' }}
```
hariombalhara, line 101: Now there is no condition to run these checks.

.github/workflows/pr.yml
```diff
@@ -109,7 +106,7 @@ jobs:
     secrets: inherit
 
   required:
-    needs: [lint, type-check, test, build, e2e]
+    needs: [lint, type-check, test, build, e2e, e2e-embed, e2e-embed-react, e2e-app-store]
```
hariombalhara, line 109: As there is no condition on these checks, we can require them here.

keithwillcode (changes requested):
@hariombalhara to avoid adding more minutes of wait time back to builds, let’s implement the matrix strategy like we have on the main e2e test suite.

hariombalhara:
@keithwillcode I have pushed the changes for matrix support. Yet to test them. But I noticed that the embed and AppStore tests don't take more time than core e2e tests.
AFAIK, All these tests run in parallel, so even without sharding they would compete much earlier than core tests.

If I am right in that analysis, we don't need sharding.

image

keithwillcode:
IMG_7832

@hariombalhara 13 minutes is what they used to take before I put the matrix in place. Now they only take 2-3 minutes.

keithwillcode (approved):

PR #10359, opened 2023-07-25
fix: Feature flag user cache

sean-brydon:
Adds feature flags to cache after its ran once. We very rarley change this and it isnt needed to be fetched each time 

CleanShot 2023-07-25 at 09 52 17


packages/features/flags/components/FlagAdminList.tsx
```diff
@@ -32,7 +32,7 @@ const FlagToggle = (props: { flag: Flag }) => {
     flag: { slug, enabled },
   } = props;
   const utils = trpc.useContext();
-  const mutation = trpc.viewer.features.toggle.useMutation({
+  const mutation = trpc.viewer.admin.toggleFeatureFlag.useMutation({
```
sean-brydon, line 35: Moved to admin so we can set the cache control on the createNextAPIHandler to true as a result of isPublic=true

packages/features/flags/server/router.ts
```diff
@@ -17,14 +14,4 @@ export const featureFlagRouter = router({
...
-  toggle: authedAdminProcedure
-    .input(z.object({ slug: z.string(), enabled: z.boolean() }))
-    .mutation(({ ctx, input }) => {
-      const { prisma, user } = ctx;
-      const { slug, enabled } = input;
-      return prisma.feature.update({
-        where: { slug },
-        data: { enabled, updatedBy: user.id },
-      });
-    }),
```
sean-brydon, line 29: Moved to admin as its a better fit there IMHO

packages/trpc/server/createNextApiHandler.ts
```diff
@@ -61,9 +61,10 @@ export function createNextApiHandler(router: AnyRouter, isPublic = false) {
         const cacheRules = {
           session: `no-cache`,
           i18n: `no-cache`,
-          // Revalidation time here should be 1 second, per https://github.com/calcom/cal.com/pull/6823#issuecomment-1423215321
+          // Revalidation time here should be 1 second, per https://github.com/calcoSm/cal.com/pull/6823#issuecomment-1423215321
```
sean-brydon, line 64:
```suggestion
          // Revalidation time here should be 1 second, per https://github.com/calcom/cal.com/pull/6823#issuecomment-1423215321
```

keithwillcode (dismissed):

packages/trpc/server/createNextApiHandler.ts
```diff
@@ -64,6 +64,7 @@ export function createNextApiHandler(router: AnyRouter, isPublic = false) {
           // Revalidation time here should be 1 second, per https://github.com/calcom/cal.com/pull/6823#issuecomment-1423215321
           "slots.getSchedule": `no-cache`, // FIXME
           cityTimezones: `max-age=${ONE_DAY_IN_SECONDS}, stale-while-revalidate`,
+          map: `max-age=${ONE_DAY_IN_SECONDS}, stale-while-revalidate`, // "map" - Feature Flag Map
```
hariombalhara, line 67: This approach of mapping cacheRules can give unexpected results as it isn't checking for full path. So, there can be features.map and example.map. This cache header would be applied to both I guess?

This isn't related to the PR, so it's non blocking but should be fixed I guess.

sean-brydon: Oh i guess youre right -> We could rename this route to something more specific so this doesnt happen?

hariombalhara: Also, this is going to consider the response fresh for a day. We might not be able to disable/enable a feature for our users within a good amount of time which is an important thing IMHO.

Also, I suspect a weird issue with this
We might end up in a scenario like this
- Say, a feature F1 was disabled and thus it's cached across all our users.
- We disable Feature F1, now, backend code knows immediately that F1 is disabled but frontend won't realize it and it might keep showing F1 in UI and taking action on that feature would be rejected by server(as it knows feature is disabled).

hariombalhara: I think the approach to solve the multiple requests for flag might be to configure react-query in such a way that the request goes only once from one pageview. Only, on hard refresh we send a new request.

Above scenario can still happen in here, but user can always refresh the page

hariombalhara (dismissed):
Left some comments

packages/trpc/server/createNextApiHandler.ts
zomars: Using stale while revalidate will prevent stale data for more than a few seconds

zomars: Pushing a fix for the namespace issue

image

apps/web/pages/api/trpc/features/[trpc].ts
```diff
@@ -1,4 +1,4 @@
 import { featureFlagRouter } from "@calcom/features/flags/server/router";
 import { createNextApiHandler } from "@calcom/trpc/server/createNextApiHandler";
 
-export default createNextApiHandler(featureFlagRouter);
+export default createNextApiHandler(featureFlagRouter, true, "features");
```
Comment (zomars) [inline apps/web/pages/api/trpc/features/[trpc].ts:4]: We can now optionally namespace api handlers to distinguish them @sean-brydon @hariombalhara

zomars (approved):
Tested and addressed @hariombalhara concerns 🙏🏽

keithwillcode (approved):

PR #10366, opened 2023-07-25
fix: switching between months leads to incorrect autoselect first available date

### LINKED ISSUE #10295: When switching between months put me on 1st of month - desired?
### Issue Summary

Hey, I'm testing this from the perspective of a user, and I'm a bit confused about what's happening when I want to book and looking the right time to suite me going to next month and back. 

### Steps to Reproduce

1. Go to https://cal.com/yangwao/25min
2. Click on next month 
3. Click on previous months

When I'm looking for dates I see available slots for next month, but when I click back, shows me all booked at first sight, however, I need to click on the first available date to see slots

### Actual Results

- Shows "All booked", even there are available days in that month

### Expected Results

- I would like to see choose the first available slot from previous months

### Evidence

https://github.com/calcom/cal.com/assets/5887929/aa3fe7c3-70cb-45c0-b86b-102ffc10cf46

MehulZR:
## What does this PR do?

Fixes #10295  (issue)

https://github.com/calcom/cal.com/assets/86384652/3cb6403b-b7b2-48f4-a0c7-b3be5ba45b19

## Type of change

## Mandatory Tasks


packages/features/calendars/DatePicker.tsx
```diff
@@ -138,6 +139,21 @@ const Days = ({
     ? availableDates(props.includedDates)
     : props.includedDates;
 
+  useEffect(() => {
+    if (selected && selected.date()) {
```
hariombalhara, line 143: We prefer early return https://handbook.cal.com/engineering/best-practices/prefer-early-returns

packages/features/calendars/DatePicker.tsx
```diff
@@ -138,6 +139,21 @@ const Days = ({
...
+      if (selected.format("YYYY-MM") === browsingDate.format("YYYY-MM")) return;
```
hariombalhara, line 144: This is really difficult for someone new to understand what's going on.
Would it make sense to abstract it out in `areWeOnTheSameMonth` function ?

packages/features/calendars/DatePicker.tsx
```diff
@@ -138,6 +139,21 @@ const Days = ({
...
+      if (!includedDates) props.onChange(browsingDate.set("date", 1));
```
hariombalhara, line 145: This code repeats the logic that decides which date would be enabled. We should avoid that as it can cause an inconsistency in future.
As the requirement is to select the first enabled date, we can refactor a bit and do exactly that


image

hariombalhara: Also, while you are fixing this bug, please see if you can also fix the scenario when date query param that is present in the URL is no longer an enabled date e.g. http://org2.cal.local:3000/test/d?date=2023-08-05&month=2023-08
It shows up like this. In this case also, we can select the first enabled date.
image

hariombalhara (dismissed):
Thanks @MehulZR. Great work!! I have tested and it works correctly but needs some code cleanup.

MehulZR:
In order to solve both the main issue and the query param issue, I think we should allow the the datePicker component to select a *null* date. The store permits it, just the datePicker component needed to be adapted for it.

With that I was able to the issues in a concise and elegant manner which didn't involve any janky solutions.

It this behaviour desirable?

https://github.com/calcom/cal.com/assets/86384652/4044ef29-1e6f-4db5-b1a6-ff098adf933a

packages/features/calendars/DatePicker.tsx
hariombalhara: We need to have firstEnabledDate of current month. I am not sure why that is happening as in the code I shared above I am finding the firstEnabledDate from days that are to be rendered, so it shouldn't have previous month day.

hariombalhara:
I am not sure how allowing `null` day to be selected would solve that. In general, allowing selection of null day doesn't make sense. But could you please push the new code in a commit, I would be able to review better then.

packages/features/calendars/DatePicker.tsx
MehulZR: firstEnabledDate looks into the daysToRender. But what if that particular month does not have available dates?. It is completely booked. Then there will be no firstEnabledDate which leads to this misbehaviour by avaliableTimeSlots.

packages/features/calendars/DatePicker.tsx
```diff
@@ -1,5 +1,8 @@
+import { useEffect } from "react";
 import { shallow } from "zustand/shallow";
 
+
+
```
hariombalhara, line 5:
```suggestion
```

packages/features/calendars/DatePicker.tsx
```diff
@@ -144,6 +146,32 @@ const Days = ({
...
+  const doesMonthContainAvailableDates = firstAvailableDate ? true : false;
+
+  const isSelectedDateAvailable =
+    selected && doesMonthContainAvailableDates
+      ? daysToRender.some(({ day, disabled }) => {
+          if (day && yyyymmdd(day) === yyyymmdd(selected) && !disabled) return true;
+        })
+      : false;
+
+  useEffect(() => {
```
hariombalhara, line 169: We don't need to use effect here
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
Made some changes that avoid it. When I avoided that, a problem that I saw was that includedDates was initially empty array and later on it get's populated which was causing disabled logic to initially show all dates as enabled and then later some of the dates got disabled.
I fixed that by making sure `includedDates` is null when slots are loading and after that it becomes an array. This was anyway misleading as empty array(earlier) could mean slots loading or no slots.

packages/features/schedules/lib/use-schedule/useNonEmptyScheduleDays.ts
```diff
@@ -3,7 +3,7 @@ import { useMemo } from "react";
 import type { Slots } from "../use-schedule";
 
 export const getNonEmptyScheduleDays = (slots?: Slots) => {
-  if (typeof slots === "undefined") return [];
+  if (typeof slots === "undefined") return null;
```
hariombalhara, line 6: If slots aren't given yet we can't compute slots, so return null to differentiate b/w empty array when no slots are actually available in a month.

This handles the case of the past month where we would get 0 slots in the month and thus an empty array.

packages/features/calendars/DatePicker.tsx
```diff
@@ -190,10 +205,7 @@ const Days = ({
               onClick={() => {
                 props.onChange(day);
               }}
-              disabled={
-                (includedDates && !includedDates.includes(yyyymmdd(day))) ||
```
hariombalhara, line 194: includedDates can actually be null now(while slots request is loading). Earlier, it was an empty array always.

hariombalhara:
@MehulZR I fixed the conflicts, handled some changes for `selected` being a possible array now and also avoided `useEffect` usage. Could you please review and test from your end as well.

It's good from my side now. Thinking about adding tests for the new behaviour handling
- Selecting first available date when navigating to next/prev month
- When directly opening through query params, selecting the first available date of the month. If that's not available it should show the banner that No dates are available and all dates should be disabled there. Plus, timeslots section should be hidden.

packages/features/calendars/DatePicker.tsx
```diff
@@ -289,4 +301,41 @@ const DatePicker = ({
...
+  selected,
+  onChange,
+}: {
+  daysToRenderForTheMonth: { day: Dayjs | null; disabled: boolean }[];
+  selected: Dayjs | Dayjs[] | null | undefined;
+  onChange: (date: Dayjs | null) => void;
+}) => {
+  // Let's not do something for now in case of multiple selected dates as behaviour is unclear and it's not needed at the moment
+  if (selected instanceof Array) {
+    return;
```
Udit-takkar, line 318: instanceof doesn't work properly in iframes. can we use Array.isArray() here?

Udit-takkar (commented):
Works great. Left one comment

packages/features/calendars/DatePicker.tsx
hariombalhara: I don't think there should be an issue wrt iframes. Such an issue might happen if we do some cross iframe data passing, I guess. But here selected and Array are always of same window.

Udit-takkar (approved):

packages/features/schedules/lib/use-schedule/useNonEmptyScheduleDays.ts
dmonn: @hariombalhara Could this lead to some issues if all slots for the current month are filled? Using the state after this merge triggered this error on our self-hosted instance.

```
TypeError: Cannot read properties of null (reading 'filter')
    at BookerComponent (/var/task/apps/web/.next/server/chunks/35157.js:313:150)
```

When navigating to a booking page that doesn't have any spots left for the current month (i.e. next two days).

Changing this back fixed the issue, but I'm not sure if it leads to different issues elsewhere.

Udit-takkar: @dmonn yeah this is causing type error

PR #10369, opened 2023-07-25
fix: app.cal.com signup changes

sean-brydon:
INTERNAL: https://threads.com/34513158436?s=9ddGBqVUA4Dwmwfmixk8qG

### DIFF
--- a/apps/web/pages/api/auth/signup.ts
+++ b/apps/web/pages/api/auth/signup.ts
@@ -2,9 +2,9 @@ import type { NextApiRequest, NextApiResponse } from "next";
 import { z } from "zod";
 
 import dayjs from "@calcom/dayjs";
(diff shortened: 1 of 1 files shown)


apps/web/pages/api/auth/signup.ts
```diff
@@ -128,6 +114,37 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
 
     if (team) {
       const teamMetadata = teamMetadataSchema.parse(team?.metadata);
+
+      if (
+        IS_CALCOM &&
+        (!teamMetadata?.isOrganization || (!!team.parentId && !teamMetadata?.isOrganization))
```
emrysal, line 120:
```suggestion
        (!teamMetadata?.isOrganization)
```

Reason: The second or will never be true 'more' than the first.

emrysal (approved):
Tested, works

PR #10375, opened 2023-07-25
fix: Username check for sign-up with invitation in org context

leog:
## What does this PR do?

Usernames were not being checked correctly when creating an account using sign-up flow with an invitation token in an organization context.

## Type of change

## How should this be tested?

Invite someone to an org and have them follow the create account link, selecting a knowingly unused username to see go through sign-up without any error.


keithwillcode (approved):
Looks good.

PR #10353, opened 2023-07-25
perf: Refactor `getEnabledApps` to filter on credentials passed

joeauyeung:
## What does this PR do?

Stacked on #10351 

`getEnabledApps` used to return the metadata for every app. This response would sometimes be sent to the frontend. We can improve performance by only returning the metadata of credentials being passed into the function.

Fixes # (issue)

## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is ab UI/UX design document, please, share it here.

## Type of change

- [x] Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- Install a conferencing app
- Go to the pages where `getEnabledApps` is used
    - Installed app page
    - Event type page
    - On the event type setup page, open the edit location modal

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #10383, opened 2023-07-25
feat: Enable Conferencing Apps for Team Events [CAL-1925]

### LINKED ISSUE #9430: [CAL-1925] Better handling of Conferencing apps for team event-types
### **Current problem**

* When selecting a conferencing app as a location for a team event-type, it's very unclear whose link the conferencing app selected will use.
* For example, if we have a Round Robin event and select **Zoom**, it will use the **Zoom** link of the person who created the event type - not the person who is actually selected as the host for the call.

\----

### **The current work-around**

* A stop-gap solution we built added an option to "Use organizers default app". This goes some way to solving the problem, but it provides little or no control to admins.

\----

### **Goal / ideal outcome**

Enable an admin to be able to:

* select Zoom/Teams etc. for a team event type
* the event will use the Zoom account of the host
* In the future, they will also be able to see if there is anyone in the event type who does not have the relevant account installed/connected (and ping them to connect.)

The two apps that are most popular in these use cases are Zoom and Teams. We will focus our initial investigations around those.

We also need to factor in apps that use links like Whereby/Around/etc. One option for these, is to let individuals install the app, and in the installation process paste in their link. The other is to have the option to paste in links for each person in the event-type in the event-type settings.

\-------

### **Alex's thoughts:**

> Eventually we were looking into basically having a location dropdown per round robin attendee, so  this would be configurable to anything - you could then go to the event type as a round robin member and adjust your settings from there, default app would still be supported but only as default.

I like this idea, but also think it's worth doing a phase of design to see if there are any other possible solutions before jumping into just one.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1925](https://linear.app/calcom/issue/CAL-1925/better-handling-of-conferencing-apps-for-team-event-types)</sub>

joeauyeung:
## What does this PR do?

Please review and merge #10353 first.

This PR enables conferencing apps to be installed to a team/org and used for team event types. 

The current conferencing apps that will be enabled in this PR are:
- MS Teams
- Jitsi
- Webex
- Huddle01

Fixes #9430 

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create an org and a team under the org
    - The org should have two members
- Create a collective, round-robin, and a managed event type
    - The collective event should only have one member assigned
- For each app listed above book the following scenarios and ensure that the meeting URLs are unique for each
    - Book a collective event
    - Book two round robin events so both members are assigned
    - Book a managed event type under both members	

## Mandatory Tasks

## Checklist


apps/web/components/AppListCard.tsx
```diff
@@ -96,7 +97,7 @@ export default function AppListCard(props: AppListCardProps) {
                   className="mr-2"
                   alt={credentialOwner.name || "Nameless"}
                   size="xs"
-                  imageSrc={credentialOwner.avatar}
+                  imageSrc={getPlaceholderAvatar(credentialOwner.avatar, credentialOwner?.name as string)}
```
joeauyeung, line 100: Fix placeholder avatar for teams

apps/web/components/apps/App.tsx
```diff
@@ -56,6 +56,7 @@ const Component = ({
   descriptionItems,
   isTemplate,
   dependencies,
+  concurrentMeetings,
```
joeauyeung, line 59: New property in app metadata to enable concurrent meetings for conferencing apps

apps/web/components/apps/App.tsx
```diff
@@ -416,7 +420,7 @@ const InstallAppButtonChild = ({
     },
   });
 
-  if (!userAdminTeams?.length || appCategories.some((category) => ["calendar", "conferencing"].includes(category))) {
+  if (!userAdminTeams?.length || appCategories.some((category) => category === "calendar" || category === "conferencing" && !concurrentMeetings)) {
```
joeauyeung, line 423: Don't show the dropdown to install a conferencing app for a team/org if the `concurrentMeetings` prop isn't true.

apps/web/components/dialog/EditLocationDialog.tsx
```diff
@@ -298,8 +299,21 @@ export const EditLocationDialog = (props: ISetLocationDialog) => {
...
+                    credentialId: values.credentialId,
+                  };
                 }
+
+                if (values.teamName) {
+                  details = {
+                    ...details,
+                    teamName: values.teamName,
+                  };
+                }
```
joeauyeung, line 313: Save the credentialId and the team name to the event type location. That way users know who owns the conferencing app that is selected. 

Since team names rarely change, we don't have to query the DB for a team name every time.

apps/web/pages/apps/installed/[category].tsx
```diff
@@ -169,7 +169,7 @@ const IntegrationsList = ({ data, handleDisconnect, variant }: IntegrationsListP
                   <Button StartIcon={MoreHorizontal} variant="icon" color="secondary" />
                 </DropdownMenuTrigger>
                 <DropdownMenuContent>
-                  {!appIsDefault && variant === "conferencing" && (
+                  {!appIsDefault && variant === "conferencing" && !item.credentialOwner?.teamId && (
```
Comment (joeauyeung) [inline apps/web/pages/apps/installed/[category].tsx:172]: On a user level, should disable setting team/org scoped conferencing apps as their default conferencing app

packages/app-store/huddle01video/api/add.ts
joeauyeung, line 1: Even though this is depreciated, still adding functionality

packages/app-store/locations.ts
```diff
@@ -359,7 +361,7 @@ export const getLocationValueForDB = (
       bookingLocation = location[eventLocationType.defaultValueVariable] || bookingLocation;
     }
   });
-  return bookingLocation;
+  return { bookingLocation, conferenceCredentialId };
```
joeauyeung, line 364: We need to query the event type for the location information and if a credentialId was saved

packages/app-store/utils.ts
joeauyeung, line 1: The `getLocationGroupedOptions` was refactored in this PR https://github.com/calcom/cal.com/pull/10384

I mixed up stacking PRs so this got merged early

packages/trpc/server/routers/loggedInViewer/locationOptions.handler.ts
```diff
@@ -15,31 +12,11 @@ type LocationOptionsOptions = {
...
-      type: true,
-      key: true,
-      userId: true,
-      teamId: true,
-      appId: true,
-      invalid: true,
-    },
-  });
-
-  const integrations = await getEnabledApps(credentials, true);
```
joeauyeung, line 38: This logic is moved to the `getLocationGroupedOptions` refactor

apps/web/components/AppListCard.tsx
hariombalhara: Screenshot 2023-07-27 at 2 45 52 PM
Possible Followup: Avatar component has fallback image support but it uses gravatar.com maybe we can just switch it with ui-avatars.com and automatically enable fallback.
Screenshot 2023-07-27 at 2 45 16 PM

packages/types/App.d.ts
```diff
@@ -144,6 +144,8 @@ export interface App {
   __template?: string;
   /** Slug of an app needed to be installed before the current app can be added */
   dependencies?: string[];
+  /** Enables video apps to be used for team events */
+  concurrentMeetings?: boolean;
```
hariombalhara, line 148: nit: I am not sure why we are calling it `concurrentMeetings`, wouldn't `teamInstallEnabled` make more sense?

packages/types/App.d.ts
```diff
@@ -144,6 +144,8 @@ export interface App {
   __template?: string;
   /** Slug of an app needed to be installed before the current app can be added */
   dependencies?: string[];
+  /** Enables video apps to be used for team events */
```
hariombalhara, line 147:
```suggestion
  /** Enables video apps to be used for team events. Non Video/Conferencing apps don't honor this as they support team installation always. */
```

packages/ui/components/apps/AppCard.tsx
```diff
@@ -200,7 +205,9 @@ const InstallAppButtonChild = ({
 
   if (
     !userAdminTeams?.length ||
-    appCategories.some((category) => category === "calendar" || category === "conferencing")
```
hariombalhara, line 203: Can we extract it out as `doesAppSupportTeamInstall()` to be used across components.

apps/web/components/dialog/EditLocationDialog.tsx
```diff
@@ -298,8 +299,21 @@ export const EditLocationDialog = (props: ISetLocationDialog) => {
...
+
+                console.log("🚀 ~ file: EditLocationDialog.tsx:310 ~ handleSubmit={ ~ details:", details);
```
hariombalhara, line 315:
```suggestion
```

packages/app-store/utils.ts
```diff
@@ -6,6 +6,8 @@ import type { TFunction } from "next-i18next";
 import { appStoreMetadata } from "@calcom/app-store/appStoreMetaData";
 import type { EventLocationType } from "@calcom/app-store/locations";
 import { defaultLocations } from "@calcom/app-store/locations";
+import getEnabledApps from "@calcom/lib/apps/getEnabledApps";
+import { prisma } from "@calcom/prisma";
```
hariombalhara, line 10: This module is imported client side(in `booking/[uid].tsx`) and thus it can't have prisma as a dependency in it.
image

We should move `getLocationGroupedOption` into it's own module and the prisma dependency would move there along with it.

packages/app-store/utils.ts
```diff
@@ -38,13 +40,85 @@ const credentialData = Prisma.validator<Prisma.CredentialArgs>()({
...
+  team?: {
+    name: string;
+  } | null;
+};
+
 export const ALL_APPS = Object.values(ALL_APPS_MAP);
 
-export function getLocationGroupedOptions(integrations: ReturnType<typeof getApps>, t: TFunction) {
+export async function getLocationGroupedOptions(
+  idToSearchOn: { userId: number } | { teamId: number },
```
hariombalhara, line 52:
```suggestion
  useOrTeamIds: { userId: number } | { teamId: number },
```

packages/app-store/utils.ts
```diff
@@ -38,13 +40,85 @@ const credentialData = Prisma.validator<Prisma.CredentialArgs>()({
...
+    } else {
+      idToSearchObject = { teamId };
+    }
+  } else {
+    idToSearchObject = { userId: idToSearchOn.userId };
+  }
+
+  const credentials = await prisma.credential.findMany({
+    where: {
+      ...idToSearchObject,
```
hariombalhara, line 97: Screenshot 2023-07-27 at 3 46 10 PM
I think it is solvable using one prisma API call.

hariombalhara: Also, we are not allowing team installed apps to be available for Team member Events. Is this long term or we are going to support that?

packages/app-store/utils.ts
hariombalhara: Nevermind: I think your approach would be much faster because you directly found out the parent teamId and then using just an **in** query.

packages/app-store/utils.ts
hariombalhara: Nevermind: I think your approach would be much faster because you directly found out the parent teamId and then using just an **in** query.

hariombalhara (changes requested):
Great work @joeauyeung !! Left some comments to resolve before merging can be done

packages/app-store/utils.ts
```diff
@@ -184,4 +123,19 @@ export function getAppFromLocationValue(type: string): AppMeta | undefined {
   return ALL_APPS.find((app) => app?.appData?.location?.type === type);
 }
 
+/**
+ *
+ * @param appCategories - from app metadata
+ * @param concurrentMeetings - from app metadata
+ * @returns - true if app supports team install
+ */
+export function doesAppSupportTeamInstall(
```
joeauyeung, line 132: @hariombalhara abstracted this logic

packages/app-store/server.ts
```diff
@@ -0,0 +1,144 @@
+import type { TFunction } from "next-i18next";
+
+import type { CredentialDataWithTeamName } from "@calcom/app-store/utils";
+import getEnabledApps from "@calcom/lib/apps/getEnabledApps";
+import { prisma } from "@calcom/prisma";
+import { AppCategories } from "@calcom/prisma/enums";
+
+import { defaultLocations } from "./locations";
+
+export async function getLocationGroupedOptions(
```
joeauyeung, line 10: @hariombalhara took your suggestion and moved it under `app-store/server`

packages/types/App.d.ts
hariombalhara: Didn't know this. But I still feel that the flag should be named something like `teamInstallEnabled`. You can set teamInstallEnabled based on this conccurentMeetings criteria or any other critera. Hypothetical other criteria might be self hosters wanting to disable certain apps for teams.

It's a nit only.

apps/web/components/AppListCard.tsx
hariombalhara: Created a ticket to track this.https://github.com/calcom/cal.com/issues/10475

hariombalhara:
In 'Edit Location' from booking flow, we don't add the context of the team/org the Location App belongs to
image

Should have been 'Jitsi Video(org2)' like at other places.

hariombalhara (approved):
LGTM!! Thanks @joeauyeung 🙏

hariombalhara (changes requested):
Not merging it yet as the target branch https://github.com/calcom/cal.com/pull/10383 have some changes requested and that would make it difficult to review that.

hariombalhara:
@joeauyeung Some merge conflicts in here probably from the merge of the target branch. Let's resolve them and then it's good to merge.

hariombalhara (approved):

Issue #10334, opened 2023-07-24
[CAL-2232] Add more instructions for contributors


Udit-takkar:
We need to add more instructions for open source contributors in docs or [CONTRIBUTING.md](http://CONTRIBUTING.md)

some common mistakes in the PR

1. Not using prettier formatting

<img width="684" alt="Screenshot 2023-07-24 at 3 13 23 PM" src="https://github.com/calcom/cal.com/assets/53316345/ff7396a4-0d0a-495d-8f6e-b80b586b5239">

2. Commiting yarn.lock file even when there is no change in the package.json file and how to uncommit it if pushed to the branch
3. Not linking the PR to issue.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2232](https://linear.app/calcom/issue/CAL-2232/add-more-instructions-for-contributors)</sub>

nicktrn:
> 1. Not using prettier formatting

There is an issue with husky hooks not being set up as yarn 2+ ignores prepare scripts. Will submit a very brief PR to address this. My guess is that this hasn't been working since the upgrade from v1 in #7738.

PR #10393, opened 2023-07-26
fix: make husky install compatible with yarn v3

nicktrn:
## What does this PR do?

Yarn v2+ ignores prepare scripts, so `husky install` hasn't been working since the upgrade in March.

I followed the [official advice](https://typicode.github.io/husky/getting-started.html#yarn-2) and moved it to the postinstall script.

Related:
- #10334 

## Type of change

## How should this be tested?

```sh
git config --unset-all core.hookspath .husky && yarn clean && yarn && git config --get core.hookspath
```
Expected output: `.husky`

## Mandatory Tasks


emrysal (approved):
Thanks for this; you're entirely right of course & it should be in the postinstall.

PR #10409, opened 2023-07-26
test: Create unit tests for react components in packages/ui/components/form/checkbox

gitstart-calcom:
## DEMO

Run 
```
yarn test checkbox
```
image


sean-brydon (approved):
Amazing thank you. I wonder if in the future we should swap this out for the Radix Checkbox defined in the same file.

PR #10411, opened 2023-07-26
chore: Remove extra params from the users prop.

emrysal:
* Removes redundant data from the user props, which also keeps some data out of the frontend.

### DIFF
--- a/apps/web/pages/[user].tsx
+++ b/apps/web/pages/[user].tsx
@@ -1,8 +1,10 @@
+import type { DehydratedState } from "@tanstack/react-query";
 import classNames from "classnames";
-import type { GetServerSidePropsContext } from "next";
+import type { GetServerSideProps, GetServerSidePropsContext, InferGetServerSidePropsType } from "next";
 import Link from "next/link";
 import { useRouter } from "next/router";
 import { Toaster } from "react-hot-toast";
+import type { z } from "zod";
 
 import {
   sdkActionManager,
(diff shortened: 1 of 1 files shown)


apps/web/pages/[user].tsx
```diff
@@ -78,25 +78,23 @@ export function UserPage(props: UserPageProps) {
             isEmbed ? "border-booker border-booker-width  bg-default rounded-md border" : "",
             "max-w-3xl px-4 py-24"
           )}>
-          {isSingleUser && ( // When we deal with a single user, not dynamic group
```
Comment (emrysal) [inline apps/web/pages/[user].tsx:81]: isSingleUser is no longer possible because dynamic events are now redirected to [type]

apps/web/pages/[user].tsx
```diff
@@ -195,9 +193,38 @@ const getEventTypesWithHiddenFromDB = async (userId: number) => {
...
+    | "length"
+    | "hidden"
+    | "requiresConfirmation"
+    | "price"
+    | "currency"
+    | "recurringEvent"
+  >)[];
+} & EmbedProps;
+
+export const getServerSideProps: GetServerSideProps<UserPageProps> = async (context) => {
```
Comment (emrysal) [inline apps/web/pages/[user].tsx:226]: Explicitly declaring the return props gives both a degree of safety and also speeds up typescript

Udit-takkar (approved):
LGTM

sean-brydon (approved):
Tested - works as expected.

PR #10416, opened 2023-07-27
fix: changed Japanese translations on booking page to better match th…

sajanlamsal:
## What does this PR do?

From a Japanese perspective, after booking an event the booking successful page displayed has a Japanese language that needs to be changed. Currently, it seems to be translated without context. I have fixed a few areas where the language is better suited. 

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

et your browser language to Japanese, and book for an event. 

- no
- set your browser language to Japanese, and book for an event

## Mandatory Tasks

- [ X] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

スクリーンショット 2023-07-27 19 37 55


keithwillcode (approved):
Thanks!

PR #10417, opened 2023-07-27
fix: Embed Code Generator - Fixes 404 in preview of embed for orgs

hariombalhara:
## What does this PR do?
Demo https://www.loom.com/share/6513fbd2cda442a589306c46e676b525

## Type of change

## Tests
-  See the loom

## Mandatory Tasks

## Checklist


apps/web/components/Embed.tsx
```diff
@@ -323,18 +344,22 @@ ${uiInstructionCode}`;
 ${uiInstructionCode}`;
     },
   },
-};
+} satisfies Record<string, Record<string, (...args: any[]) => string>>;
```
hariombalhara, line 347: satisfies does what I really needed. It works now with Next.js

apps/web/components/Embed.tsx
```diff
@@ -687,7 +730,14 @@ ${getEmbedTypeSpecificString({ embedFramework: "HTML", embedType, calLink, previ
 
 /* If you are using npm */
 // npm install @calcom/embed-react
-${getEmbedTypeSpecificString({ embedFramework: "react", embedType, calLink, previewState })}
+${getEmbedTypeSpecificString({
```
hariombalhara, line 733: Thought of making it a hook. But embedFramework is conditional and using a hook with conditional would require me to make a state which is unnecessary. So, keeping it as a regular function only

packages/embeds/embed-core/src/preview.ts
```diff
@@ -1,7 +1,11 @@
-const WEBAPP_URL =
-  import.meta.env.EMBED_PUBLIC_WEBAPP_URL || `https://${import.meta.env.EMBED_PUBLIC_VERCEL_URL}`;
-const EMBED_LIB_URL = import.meta.env.EMBED_PUBLIC_EMBED_LIB_URL || `${WEBAPP_URL}/embed/embed.js`;
-
```
hariombalhara, line 4: Orgs change forced this logic to improve. No need to read from env variables here.

zomars (dismissed):
[cool](https://app.graphite.dev/macros/calcom)

zomars (approved):
Fixed conflicts

PR #10418, opened 2023-07-27
perf: Remove IP ban and tweak matcher accordingly

emrysal:
## What does this PR do?

* The IP ban feature in CloudFlare (for cal.com) actually works, we can use that instead of middleware and reduce our middleware executions.
* Also tweaks the matchers


hariombalhara (approved):
Awesome !!

PR #9105, opened 2023-05-24
refactor: next/router hooks with next/navigation hooks

zomars:
Retaking efforts made here by @grzpab  #8843

## What does this PR do?

fixes #9929

Replaces all usages of hooks from `next/router` to `next/navigation`

## Type of change
- Chore (refactoring code, technical debt, workflow improvements)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How should this be tested?
Full application check, including visual regression checks.

## Checklist
- I haven't checked if my PR needs changes to the documentation
- I haven't added tests that prove my fix is effective or that my feature works

PR #10434, opened 2023-07-28
fix: Broken Embed Dialog in Routing Forms due to recent rewrite

hariombalhara:
## What does this PR do?

Fix broken Embed Code Generator Dialog due to rewrite on /routing-forms/forms.

## Root Cause 
[Next.js wrongly interprets Embed Dialog route in Routing Forms as matching [user]/[type]](https://www.loom.com/share/a088fb9e828a485a848cca320ad78886). The reason for that is that the route file /routing-forms/forms doesn't exist and it seems to be an issue with Rewrites.

## Approach to fix it
Remove the rewrite for now. It isn't necessary to have /routing-forms working, we can just redirect it to /apps/routing-forms

Also, I have made the code more clear by creating files for routes corresponding to /routing-forms**. This is a better approach because we determine if a page is a booking page by a RegExp which is supposed to exclude these pages. If the file exists, exclusion happens automatically.

After:
https://www.loom.com/share/d71372c2cee64b62923cbb213ef2d76e

## Type of change

## How should this be tested?
- Go through the various Dialogs in app

## Mandatory Tasks

## Checklist


apps/web/components/Embed.tsx
hariombalhara, line 1: All code from this file taken from https://github.com/calcom/cal.com/pull/9105

packages/ui/components/dialog/Dialog.tsx
```diff
@@ -14,44 +14,56 @@ export type DialogProps = React.ComponentProps<(typeof DialogPrimitive)["Root"]>
...
-          undefined,
-          { shallow: true }
-        );
+        router.push(`${pathname}?${newSearchParams.toString()}`);
       }
-      setOpen(open);
+      setDialogState(open ? DIALOG_STATE.OPEN : DIALOG_STATE.CLOSING);
     };
-    // handles initial state
-    if (!open && router.query["dialog"] === name) {
```
hariombalhara, line 49: This was quite a tricky handling here. On line 35 we modify query which is a reference to router.query and thus router.query would have query param `dialog` removed instantly. In reality, URL would still have that query param. With next/router, it stopped working because searchParams is readOnly and creating a new copy of it would loose the reference.

So, I have introduced a life cycle state CLOSING for this case. As soon as Dialog is dismissed, it doesn't go to CLOSED state. It goes to CLOSING state instead and then navigates to CLOSED state when the query param is removed.

packages/features/shell/Shell.tsx
```diff
@@ -539,7 +539,7 @@ const navigation: NavigationItemType[] = [
     isCurrent: ({ router, item }) => {
       const path = router.asPath.split("?")[0];
       // During Server rendering path is /v2/apps but on client it becomes /apps(weird..)
-      return path.startsWith(item.href) || path.startsWith("/v2" + item.href);
```
hariombalhara, line 542: There is no v2 now

zomars (dismissed):
LGTM

zomars (approved):
Solved conflicts and build fixes

PR #10404, opened 2023-07-26
fix: weekly limits

nicktrn:
## What does this PR do?

This addresses two bugs related to weekly events, relating to:

1. All non-variable length event types
2. Weeks spanning two months

Fixes #10361
Fixes #10402

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

1. **Fixed length events**
- Limit a 30min event to 30min per week
- Book once
- See that the entire week is blocked

2. **Week in two months**
- Limit any event to 1 booking (or 1 length of event) per week
- Book in week that ends in next month, e.g. Mon 31 Jul
- See that you are unable to book 1-5 Aug

Availability and booking page tests are passing, but might make sense to add a new test to prevent regressions. Bug (2) was probably always there, but not (1).

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works

PR #10435, opened 2023-07-28
fix: Limit Booking Frequency

### LINKED ISSUE #10361: [CAL-2237] "Limit booking frequency" setting doesn't work
### Issue Summary

I have a [Cal.com](http://Cal.com) event at [this link](https://cal.com/quanx/boba). In this event's settings > "Limits" > "Limit booking frequency", I added a rule of 1 per week. After I've set this rule, people could still book > 1 per week: evidently, I already have 1 booking this week and another next week, yet my public event page still allows people to book during these weeks.

### Steps to Reproduce

1. In an event's settings > "Limits" > "Limit booking frequency", add a rule of 1 per week
2. Book a meeting this week for that event
3. Return to the booking page to see that this week is available for more bookings

### Actual Results

* People can still book after I've set a rule of only one event booking per week

### Expected Results

* If this week already has a booking, people should see that this week no longer has any open slots for more booking

### Technical details

* I'm using Chrome / Arc browsers
* Confirmed that this setting has been saved and was set for several days now

### Evidence

CleanShot 070349@2x

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2237](https://linear.app/calcom/issue/CAL-2237/limit-booking-frequency-setting-doesnt-work)</sub>

rkreddy99:
## What does this PR do?

fixes a booking issue which allowed for event bookings even though the limits were placed.

Fixes #10361 

## Type of change

## How should this be tested?

-Put a limit on weekly bookings to 1 and create an event towards the end of the month (say 31st Jul) and the entire week (1st Aug) should be blocked. 
- Put a limit on monthly bookings to 1 and make sure that there is a booking in the same month but before today. There should not be any slots available in the month.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/core/getBusyTimes.ts
+++ b/packages/core/getBusyTimes.ts
@@ -211,3 +211,120 @@ export async function getBusyTimes(params: {
 }
 
 export default getBusyTimes;
+
+// this funcationality cannot be imolpemented in getBusyTime because it has optional and eventTypeId and while searching for bookings it doesnt use eventTypeId filter which can fetch unnecessary bookings
+export async function getEventBookingsForPeriod(params: {
+  userId: number;
+  eventTypeId: number;
+  startTime: string;
+  endTime: string;
+  seatedEvent?: boolean;
+}) {
+  const { userId, eventTypeId, startTime, endTime, seatedEvent } = params;
+  logger.silly(
+    `Fetch bookings from Cal Bookings in range ${startTime} to ${endTime} for input ${JSON.stringify({
+      userId,
+      eventTypeId,
+      status: BookingStatus.ACCEPTED,
+    })}`
+  );
+  // get user email for attendee checking.
+  const user = await prisma.user.findUniqueOrThrow({
+    where: {
+      id: userId,
+    },
+    select: {
+      email: true,
+    },
+  });
+
+  /**
+   * A user is considered busy within a given time period if there
+   * is a booking they own OR attend.
+   *
+   * Performs a query for all bookings where:
+   *   - The given booking is owned by this user, or..
+   *   - The current user has a different booking at this time he/she attends
+   *
+   * See further discussion within this GH issue:
+   * https://github.com/calcom/cal.com/issues/6374
+   *
+   * NOTE: Changes here will likely require changes to some mocking
+   *  logic within getSchedule.test.ts:addBookings
+   */
+  performance.mark("prismaBookingGetStart");
+
+  const sharedQuery = {
+    startTime: { gte: new Date(startTime) },
+    endTime: { lte: new Date(endTime) },
+    status: {
+      in: [BookingStatus.ACCEPTED],
+    },
+  };
+  // Find bookings that block this user from hosting further bookings.
+  const bookings = await prisma.booking.findMany({
+    where:
+      // User is primary host (individual events, or primary organizer)
+      {
+        ...sharedQuery,
+        eventTypeId,
+        userId,
+      },
+    select: {
+      id: true,
+      startTime: true,
+      endTime: true,
+      title: true,
+      eventType: {
+        select: {
+          id: true,
+          seatsPerTimeSlot: true,
+        },
+      },
+      ...(seatedEvent && {
+        _count: {
+          select: {
+            seatsReferences: true,
+          },
+        },
+      }),
+    },
+  });
+
+  const bookingSeatCountMap: { [x: string]: number } = {};
+  const busyTimes = bookings.reduce(
+    (aggregate: EventBusyDetails[], { id, startTime, endTime, eventType, title, ...rest }) => {
+      if (rest._count?.seatsReferences) {
+        const bookedAt = dayjs(startTime).utc().format() + "<>" + dayjs(endTime).utc().format();
+        bookingSeatCountMap[bookedAt] = bookingSeatCountMap[bookedAt] || 0;
+        bookingSeatCountMap[bookedAt]++;
+        // Seat references on the current event are non-blocking until the event is fully booked.
+        if (
+          // there are still seats available.
+          bookingSeatCountMap[bookedAt] < (eventType?.seatsPerTimeSlot || 1) &&
+          // and this is the seated event, other event types should be blocked.
+          eventTypeId === eventType?.id
+        ) {
+          // then we do not add the booking to the busyTimes.
+          return aggregate;
+        }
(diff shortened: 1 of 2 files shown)


packages/core/getUserAvailability.ts
```diff
@@ -187,27 +187,67 @@ export async function getUserAvailability(
...
+    for (const key of intervalLimitKeys) {
+      if ((bookingLimits && key in bookingLimits) || (durationLimits && key in durationLimits)) {
+        const filter = key.split("_")[1].toLowerCase() as "week" | "month" | "day";
+        startDate = dayjs.min(startDate, startDate.startOf(filter));
+        endDate = dayjs.max(endDate, endDate.endOf(filter));
+      }
+    }
+
+    const currentEventBookingsBeforeDateFromPromise = getEventBookingsForPeriod({
+      startTime: startDate.toISOString(),
```
rkreddy99, line 211: getBusy function fetches the data for `dateFrom - dateTo`

the following code fetches for `startDate - dateFrom` and `dateTo - endDate`

nicktrn:
> I believe the merged fix wont work in the case of `PER_MONTH` limit.

Indeed, we need to check bookings for the whole month, not just what's left of it.

> we can just change start and end dates to dayjs.min(startDate.startOf(month), startDate.startOf(week)) and dayjs.max(endDate.endOf(month), endDate.endOf(week))

That should be enough to fix this, and we already do that when the current date is in the first week of the month.

Re performance, I think a conditional approach like you suggest is best - especially now that we're talking about possibly extending by 1-2 months, querying the db + checking linked calendars.

There are plenty of things that aren't great about the current getUserAvailability function. That's why I don't like to mess around with too much in core functions. Fix one thing, break another. In this case, performance - which already isn't great.

Maybe just "bandaging" this for now and conditionally fetching additional bookings _only for monthly limits_ would be enough? Maybe we could even use the existing functions that are already used for yearly limits?

rkreddy99:
>  Maybe we could even use the existing functions that are already used for yearly limits?
Yeah this was my initial thought as well but for this path we have to optimise the queries otherwise the performance will take a hit. I am still working on that.

So I tried to use the existing functionality as much as possible without introduced new functions.

The newly created function to fetch event bookings is also almost a replica of getbusytimes but specifically for this use case.

This is definitely a temporary solution using the existing flow.

packages/core/getBusyTimes.ts
```diff
@@ -211,3 +211,120 @@ export async function getBusyTimes(params: {
 }
 
 export default getBusyTimes;
+
+// this funcationality cannot be imolpemented in getBusyTime because it has optional and eventTypeId and while searching for bookings it doesnt use eventTypeId filter which can fetch unnecessary bookings
+export async function getEventBookingsForPeriod(params: {
```
rkreddy99, line 216: We can implement this functionality using `getBusyTimes` function by
- adding `checkGivenEventOnly` - a `boolean` param, which if enabled will only get bookings where `eventId === eventTypeId`
- provide `credentials = []` and `selectedCalendars = []`

CarinaWolli:
I can't reproduce the issue anymore on latest main (I tried to do what you described in the 'How should this be tested?' part)

> Put a limit on weekly bookings to 1 and create an event towards the end of the month (say 31st Jul) and the entire week (1st Aug) should be blocked.

That got solved here already: https://github.com/calcom/cal.com/pull/10404

> Put a limit on monthly bookings to 1 and make sure that there is a booking in the same month but before today. There should not be any slots available in the month.

I don't know where this one got fixed, but it is working for me. @rkreddy99 can you confirm that?

rkreddy99:
> I don't know where this one got fixed, but it is working for me. @rkreddy99 can you confirm that?

@CarinaWolli  Does the booked date and current date fall in the same week? If yes then you wont be able to reproduce the error, sorry for the miscommunication. The merged fix already checks for dates between beginning of dateFrom and end of week of dateTo.

So to reproduce the error book an event at 1st and get the user availability from 21st to 30th. 

Here's the error in the logic (before and after merged fix), 

1. The code (prev to merged fix) fetches user bookings and meetings from the current date to the end of the month (for the calender view in bookings page), by calling `getBusyTimes.` 
2. So this does not include all the prev meetings in that month.
3. After that bookingLimits and durationLimits are fetched.
4. Now to calculate the busy times for month, week or day limits, bookings fetched in the step 1 are used, which dont have any bookings from the prev time range.
5. Yearly limits are check separately using db calls in `getBusyTimesFromBookingLimits` and `getBusyTimesFromDurationLimits` functions, but not for PER_MONTH, PER_WEEK and PER_DAY.

The merged fix, changes the range for step1 from `dateFrom - dateTo` to `beginningOfWeekOf(dateFrom) - endOfWeekOf(dateTo)`, which does not solve the case for monthly limits, but has introduced another bug -> `getBusyTimes` function adds `afterEventBuffer` and `beforeEventBuffer` to the booked times to get user availability. Now when we just extend the range of dates for step1, we add that duration as booked time, which will result in wrong duration limits calculation (1hr limit, 15min event duration + 40 min buffer time, durationLimit exceeded for other bookings). Also we dont really require all the things like calender events, events attended by user (as attendee) etc for the `1st - currentDate`.

packages/core/getUserAvailability.ts
```diff
@@ -187,27 +187,67 @@ export async function getUserAvailability(
     source: query.withSource ? a.source : undefined,
   }));
 
-  const bookings = busyTimes.filter((busyTime) => busyTime.source?.startsWith(`eventType-${eventType?.id}`));
```
rkreddy99, line 190: @CarinaWolli bookings are derived from busyTimes, which does not include the whole month's booking and also use afterEventBuffer and beforeEventBuffer.

nicktrn:
The issue with monthly limits is easier to visualise if we pretend today is the middle of a month:

missing-limit-data

What PR #10404 did is extend the bookings we fetch by the blue dates.

rkreddy99:
"A picture is worth a thousand words" 😅, @CarinaWolli checkout the above image by @nicktrn and then read my 1000 words😂

zomars:
I'm assuming since https://github.com/calcom/cal.com/pull/10480#pullrequestreview-1570736438 got merged we can close this one? @nicktrn

nicktrn:
Yes, I picked those commits with correct attribution. @zomars

alannnc:
Closing as #10480 should consider this changes too.

PR #10442, opened 2023-07-28
test: Create unit tests for react components in packages/ui/components/form/step

gitstart-calcom:
## Demo

Run 
```
yarn test step
```
image


packages/ui/components/form/step/steps.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  navigateToStep: mockNavigateToStep,
+  stepLabel: (currentStep: number, totalSteps: number) => `Test Step ${currentStep} of ${totalSteps}`,
+};
+
+describe("Tests for Steps Component", () => {
+  test("Should render the correct number of steps", () => {
+    const { getByTestId, getByText } = render(<Steps {...Props} />);
+
+    for (let i = 0; i < Props.maxSteps; i++) {
+      const step = getByTestId(`step-indicator-${i}`);
```
hariombalhara, line 23: `getByTestId` gets by data-testid attribute only. So, it would always have` toHaveAttribute("data-testid")` assertion passing.
I don't think we need `toHaveAttribute` assertion, instead we should check that the steps are clickable. There are 2 things in there
- Presence of` cursor-pointer` class. I know that we shouldn't test implementation details but this class is something that must always be present as long as we are using tailwind
- Clicking every step till current step and checking that navigateToStep is being called.

packages/ui/components/form/step/steps.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+    expect(getByText(`Test Step ${CURRENT_STEP} of ${MAX_STEPS}`)).toBeInTheDocument();
+  });
+  test("Should navigate to the correct step when clicked", () => {
+    const { getByTestId, getByText, rerender } = render(<Steps {...Props} />);
+
+    const stepIndicators = getByTestId("step-indicator-1");
+    fireEvent.click(stepIndicators);
+    expect(mockNavigateToStep).toHaveBeenCalledTimes(1);
+
+    CURRENT_STEP = CURRENT_STEP + 1;
```
hariombalhara, line 37: A test shouldn't modify external variable.

hariombalhara (changes requested):
Thanks @gitstart. Left some comments

gitstart-calcom:
Hello @hariombalhara 
The changes have been made, can you review them, please?

hariombalhara (approved):
Thanks @gitstart-calcom 🙏 Looks good now!!

PR #10448, opened 2023-07-28
feat: Modify sort with if/elseif/else randomize when the dates are equal

### LINKED ISSUE #10339: [CAL-2235] Least recently booked should assign to the next booker at random if all else is equal.
When a new round robin is created or a few users are created at the time they have been least recently booked, these can be identical for multiple users when they have just been added. When this is the case we should randomise the next assignee instead of going through the users in the order in which they got added.

<sub>[CAL-2235](https://linear.app/calcom/issue/CAL-2235/least-recently-booked-should-assign-to-the-next-booker-at-random-if)</sub>

emrysal:
## What does this PR do?

Fixes #10339


Udit-takkar (approved):
LGTM. works fine

Udit-takkar (dismissed):
So i tested with 4 members with one of the member have availablility mon - wed while rest of them had normal working days( mon -fri) but when i tried booking on friday multiple times i would get the same user as host even though the user was not the least recently booked.

the key value pair is userIdAndAtCreatedPair and user with id 11 was most recently booked and not least recently but it was picked by the leastRecentlyBookedUser logic.

Screenshot 2023-08-01 at 5 10 59 PM

packages/lib/server/getLuckyUser.ts
```diff
@@ -92,7 +92,10 @@ async function leastRecentlyBookedUser<T extends Pick<User, "id" | "email">>({
   }
 
   const leastRecentlyBookedUser = availableUsers.sort((a, b) => {
-    return userIdAndAtCreatedPair[a.id] > userIdAndAtCreatedPair[b.id] ? 1 : -1;
+    if (userIdAndAtCreatedPair[a.id] > userIdAndAtCreatedPair[b.id]) return 1;
+    else if (userIdAndAtCreatedPair[a.id] < userIdAndAtCreatedPair[b.id]) return 0;
```
hariombalhara, line 96: It doesn't sort 3,2,1 to 1,2,3. That seems like an unintended behaviour change. Can you confirm @emrysal

Udit-takkar: yeah it should be 

```suggestion
    else if (userIdAndAtCreatedPair[a.id] < userIdAndAtCreatedPair[b.id]) return -1;
```

Udit-takkar: or we should probably use days() to compare dates

emrysal: comparing dates works fine as they are numeric in this comparison, good spot on the 0! It should indeed be -1

hariombalhara (approved):

CarinaWolli:
@emrysal can you take a look at why the unit test 'can find lucky user with maximize availability' fails?

zomars:
Heads up @emrysal this is causing `yarn test` to randomly fail.

zomars:
Made a fix https://github.com/calcom/cal.com/pull/10581

PR #10449, opened 2023-07-28
fix: use cloudflare real IP

sean-brydon:
use cloudflare real ip instead of forwarded for if exists

### DIFF
--- a/packages/lib/getIP.ts
+++ b/packages/lib/getIP.ts
@@ -10,8 +10,14 @@ export function parseIpFromHeaders(value: string | string[]) {
  * @see https://github.com/vercel/examples/blob/main/edge-functions/ip-blocking/lib/get-ip.ts
  **/
 export default function getIP(request: Request | NextApiRequest) {
-  const xff =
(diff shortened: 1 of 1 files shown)


packages/lib/getIP.ts
```diff
@@ -10,8 +10,14 @@ export function parseIpFromHeaders(value: string | string[]) {
  * @see https://github.com/vercel/examples/blob/main/edge-functions/ip-blocking/lib/get-ip.ts
  **/
 export default function getIP(request: Request | NextApiRequest) {
-  const xff =
-    request instanceof Request ? request.headers.get("x-forwarded-for") : request.headers["x-forwarded-for"];
+  let xff =
+    request instanceof Request
+      ? request.headers.get("cf-connecting-ip")
+      : request.headers["cf-connecting-ip"];
```
sean-brydon, line 16: https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#cf-connecting-ip

emrysal (approved):
lgtm (let's see if it works)

PR #10450, opened 2023-07-28
feat: Auto join team if already in org [CAL-2248]

### LINKED ISSUE #10401: [CAL-2248] Orgs: Invites to teams when you're already in the org should be auto-accepted
Currently invites are showing up here, even though I'm already in the org. If an admin adds you to a team, you should see it already.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2248](https://linear.app/calcom/issue/CAL-2248/orgs-invites-to-teams-when-youre-already-in-the-org-should-be-auto)</sub>

sean-brydon:
Auto joins a team if already in the organization that the team is apart of. 

Fixes: #10401 
Fixes: CAL-2248

Loom: https://www.loom.com/share/a4993ec59f62494ab7397b05402b4135

how to test: 
invite user from organization to a team - auto join should happen. 
Invite a user not from organization - auto join should be false and the invite/onboarding flow happens.

### DIFF
--- a/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
+++ b/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
@@ -70,13 +70,12 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
     } else {
       throwIfInviteIsToOrgAndUserExists(invitee, team, input.isOrg);
 
-      if (
-        await createAndAutoJoinIfInOrg({
-          invitee,
-          role: input.role,
-          team,
-        })
-      ) {
+      const shouldAutoJoinOrgTeam = await createAndAutoJoinIfInOrg({
+        invitee,
+        role: input.role,
+        team,
+      });
+      if (shouldAutoJoinOrgTeam.autoJoined) {
         // Continue here because if this is true we dont need to send an email to the user
         // we also dont need to update stripe as thats handled on an ORG level and not a team level.
         continue;
--- a/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
+++ b/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
@@ -21,6 +21,7 @@ import {
   createProvisionalMembership,
   getIsOrgVerified,
   sendVerificationEmail,
+  createAndAutoJoinIfInOrg,
 } from "./utils";
 
 type InviteMemberOptions = {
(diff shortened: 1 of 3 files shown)


packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
```diff
@@ -325,3 +325,39 @@ export function getIsOrgVerified(
...
+    return {
+      autoJoined: false,
+    };
+  }
+
+  // If the user is invited to a child team, they are automatically accepted into the parent org
+  await prisma.membership.create({
+    data: {
+      userId: invitee.id,
+      teamId: team.id,
```
joeauyeung, line 354: If the user is being added to the org shouldn't the `teamId` be the team `parentId`?

packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
```diff
@@ -325,3 +325,39 @@ export function getIsOrgVerified(
...
+  invitee: User;
+  role: MembershipRole;
+}) {
+  if (invitee.organizationId && invitee.organizationId !== team.parentId) {
+    return {
+      autoJoined: false,
+    };
+  }
+
+  if (!team.parentId) {
```
emrysal, line 344: What if he joins the main organisation team?

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
```diff
@@ -69,6 +70,18 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
     } else {
       throwIfInviteIsToOrgAndUserExists(invitee, team, input.isOrg);
 
+      if (
+        await createAndAutoJoinIfInOrg({
+          invitee,
+          role: input.role,
+          team,
+        })
```
emrysal, line 78:
```suggestion
        }).autoJoined
```

packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
sean-brydon: This doesnt handle to org. This is only invites to teams apart of the users org. 

Let me now if i need to comment to explain that

packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
sean-brydon: then parentId is null and this is skipped - we want them to have the full onboarding flow. 

if autojoined here is true we skip the invite etc

emrysal (dismissed):
Looks good to me

emrysal (approved):

PR #10454, opened 2023-07-28
fix: signup password hint validation

### LINKED ISSUE #10379: [CAL-2258] Password hints don't match
Here we can see that we don't have any lowercase nor uppercase letter, but we still show it green.

Here we see that we have at least a number but we display the error anyways

Here we see bot upper and lowecase letter but the error remains

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2258](https://linear.app/calcom/issue/CAL-2258/password-hints-dont-match)</sub>

MehulZR:
## What does this PR do?

Fixes #10379 (issue)

https://github.com/calcom/cal.com/assets/86384652/34994639-ba82-4ebf-9772-79c5b3ed4ebb

## Type of change

## Mandatory Tasks


packages/ui/components/form/inputs/HintOrErrors.tsx
```diff
@@ -3,27 +3,38 @@ import { useFormContext } from "react-hook-form";
...
-  const fieldErrors: FieldErrors<T> | undefined = formState.errors[fieldName];
+  const formFieldErrors: FieldErrors<T> | undefined = formState.errors[fieldName];
+
+  let Errors = validationErrors ? validationErrors : [];
+  if (formFieldErrors?.message.length && !Errors.includes(formFieldErrors.message))
+    Errors.push(formFieldErrors.message);
 
-  if (!hintErrors && fieldErrors && !fieldErrors.message) {
-    // no hints passed, field errors exist and they are custom ones
+  // no hints passed, field errors exist and they are custom ones
```
MehulZR, line 32: Was not able to figure out how this scenario happens.
Some information regarding this will be helpful in inspecting it further.

Udit-takkar (changes requested):
@MehulZR Awesome work but we already have common function that validate password that we use in other parts of the codebase.

you should do something like

```
const passwordSchema = z.string().superRefine((data, ctx) => {
  const isStrict = false;
  const result = isPasswordValid(data, true, isStrict);

  Object.keys(result).map((key: string) => {
    if (!result[key as keyof typeof result]) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        path: [key],
        message: key,
      });
    }
  });
});

```

check out apps/web/components/setup/AdminUser.tsx file. I don't think you would require to make changes in HintOrErrors or Input file. the only difference would be that AdminUser.tsx uses isStrict = true because it requires password of length of 15.

you can have a look at isPasswordValid function you might have to change the condition to >=7 from > 7 to allow 7 characters length.

Also this signup page make a request to /api/auth/singup which uses z.string().min(7) so you should create a common schema and use it on both pages/signup and api/auth/signup

packages/features/auth/lib/isPasswordValid.ts
```diff
@@ -10,15 +10,11 @@ export function isPasswordValid(password: string, breakdown?: boolean, strict?:
...
     min = false, // Eight characters, or fifteen in strict mode.
     admin_min = false;
-  if (password.length > 7 && (!strict || password.length > 14)) min = true;
+  if (password.length >= 7 && (!strict || password.length > 14)) min = true;
   if (strict && password.length > 14) admin_min = true;
-  for (let i = 0; i < password.length; i++) {
-    if (!isNaN(parseInt(password[i]))) num = true;
-    else {
-      if (password[i] === password[i].toUpperCase()) cap = true;
-      if (password[i] === password[i].toLowerCase()) low = true;
```
MehulZR, line 19: These conditions passes if the given input is a symbol (< > , ? /).

apps/web/pages/api/auth/signup.ts
```diff
@@ -1,5 +1,5 @@
 import type { NextApiRequest, NextApiResponse } from "next";
-import { z } from "zod";
+import { apiSignupSchema as signupSchema } from "pages/signup";
```
Udit-takkar, line 2: NIT: We can move apiSignupSchema to zod-utils

Udit-takkar (dismissed):
LGTM. Great work @MehulZR

Udit-takkar:
TODO -

- [ ]  fix conflicts

Udit-takkar (approved):
LGTM. Fixed conflicts

PR #10456, opened 2023-07-28
fix: #10423 private link glitch

### LINKED ISSUE #10423: "Generate private link" event setting resets after save
### Issue Summary

`Generate private link` advanced event setting resets after repeated save

### Steps to Reproduce

1. Go to event Settings -> Advanced panel
2. Enable `Generate private link` toggle and Save
3. Reload the page
4. Change something else, e.g. event title and Save
5. Reload the page
6. `Generate private link` is not toggled off

### Actual Results

I did not expect it to be toggled off

### Expected Results

`Generate private link` should still be on?

### Technical details

Probably not important, reproduced both in production and in dev main branch.

### Evidence

https://github.com/calcom/cal.com/assets/12836237/2704cbe9-e0b9-4c9e-8f33-e4bb7ebb0050

A1zenSousuke:
## What does this PR do?
Fixes: #10423 

The problem in code was that if value of private link is changed then the link will still be visible but if after the reload we change the event name to something else and save it then the hashed link value is not added so when refreshed it again it wasnt able to recieve the previous hashedLink value. so I have just added a conditional value if value is already there then it will use previous value or it will be undefined. I hope this solution is correct.

@PeerRich can you check this PR

https://github.com/calcom/cal.com/assets/30045569/b4fff678-450f-4eff-9029-d51338747528

### DIFF
--- a/apps/web/pages/event-types/[type]/index.tsx
+++ b/apps/web/pages/event-types/[type]/index.tsx
@@ -247,6 +247,7 @@ const EventTypePage = (props: EventTypeSetupProps) => {
       durationLimits: eventType.durationLimits || undefined,
       length: eventType.length,
       hidden: eventType.hidden,
+      hashedLink: eventType.hashedLink?.link || undefined,
       periodDates: {
         startDate: periodDates.startDate,
         endDate: periodDates.endDate,


Udit-takkar (approved):
LGTM. Tested works as expected

PR #10459, opened 2023-07-28
test: Create unit tests for react components in packages/ui/components/form/select

gitstart-calcom:
Run 
```
yarn test select
```
image


packages/ui/components/form/select/select.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+      expect(screen.getByText("Option 3")).toBeInTheDocument();
+      expect(screen.getByRole("form")).toHaveFormValues({ test: "option3" });
+    });
+
+    test("Should select and have the correct value after selection of an option", async () => {
+      await waitFor(async () => {
+        await selectOption("Option 2");
+      });
+
+      expect(onChangeMock).toHaveBeenCalled();
```
hariombalhara, line 61: Would be best to test the callback args as well.

packages/ui/components/form/select/select.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+    test("Should keep the selected value after the user has selected an option and clicked out of the selection element", async () => {
+      await waitFor(async () => {
+        await selectOption("Option 2");
+        const outsideButton = screen.getByText("Click Outside");
+        fireEvent.click(outsideButton);
+      });
+
+      expect(screen.getByRole("form")).toHaveFormValues({ test: "option2" });
+    });
+  });
```
hariombalhara, line 88: We don't seem to be testing the code that we have written in `Select` component. 
In the component, we are just configuring the props for react-select.
Screenshot 2023-07-31 at 12 45 44 PM

- We need a test for verifying that any classes passed to Select are passed to ReactSelect
- We might need to test certain classes as well. It isn't recommended to test implementation details so may be we keep a color map for tests and verify the background color of the DOM element is CAL_BG_SUBTLE when isDisabled prop is given.
- Similarly we should probably test the result of passing all these props - isFocussed, isDisabled, isSelected. It might be overkill as well

hariombalhara (dismissed):
Thanks @gitstart. Left some comments.

gitstart-calcom:
Hello @hariombalhara
The changes have been made, and they are ready for review

keithwillcode (dismissed):

keithwillcode (approved):

Issue #10292, opened 2023-07-21
Erroe ESlint / Rule prettier/prettier


MyPsyOnline:
Hello,

i have an error when building fresh install :

### Issue Summary
Fresh install from github version 3.1.4
I'm able to access at webinterface when i'm on the stage "yarn dev"

But when i do : yarn build, i have this error : 

`@calcom/web:build: - error ESLint: Cannot read properties of undefined (reading 'buildError') Occurred while linting /home/debian/cal.com/apps/web/components/apps/App.tsx:1 Rule: "prettier/prettier"`

### Actual Results

I'm stucked as yarn build doesn't success

### Technical details

- nodejs version 18.16.1

PeerRich:
cant reproduce. 

i can run yarn build and yarn dev just fine with 3.1.4

MyPsyOnline:
Thanks for this answer.
I reproduce the same error when i try to update an actual running version of cal.com.

is it possible That this issue is specific to my configuration ?
In this case, any idea to guide me to a solution ?

nicktrn:
I've had the same error before, but could still proceed with the build. Maybe try to run with `CI=1 yarn build`

MyPsyOnline:
> I've had the same error before, but could still proceed with the build. Maybe try to run with `CI=1 yarn build`

Yes, Build has ending with success  !
Thank you so much !!

PR #10491, opened 2023-07-31
fix: eslint crashing

O4epegb:
## What does this PR do?

`yarn lint` crashes for me locally (and on CI too apparently, e.g. other random PR https://github.com/calcom/cal.com/actions/runs/5718689486/job/15495018755?pr=10488#step:5:18) because in App.tsx one of the type has the same name as App component, both `App`, basically shadowing
It's actually valid TS code, but it seems it requires some package updates to parse without errors, and since right now it's hard to update package versions and yarn.lock correctly, I decided to just fix this one line.

## Type of change

## How should this be tested?

Run `yarn lint`, it should not crash

## Mandatory Tasks

### DIFF
--- a/apps/web/components/apps/App.tsx
+++ b/apps/web/components/apps/App.tsx
@@ -1,23 +1,22 @@
 import Link from "next/link";
 import type { IframeHTMLAttributes } from "react";
-import React, { useState, useEffect } from "react";
-import { useRouter } from "next/router";
-import { CAL_URL } from "@calcom/lib/constants";
-import type { RouterOutputs } from "@calcom/trpc/react";
+import React, { useState } from "react";
 
 import useAddAppMutation from "@calcom/app-store/_utils/useAddAppMutation";
 import { InstallAppButton, AppDependencyComponent } from "@calcom/app-store/components";
 import DisconnectIntegration from "@calcom/features/apps/components/DisconnectIntegration";
+import { Spinner } from "@calcom/features/calendars/weeklyview/components/spinner/Spinner";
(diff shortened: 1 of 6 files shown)


apps/web/components/apps/App.tsx
```diff
@@ -1,23 +1,22 @@
...
 import type { UserAdminTeams } from "@calcom/features/ee/teams/lib/getUserAdminTeams";
 import Shell from "@calcom/features/shell/Shell";
 import classNames from "@calcom/lib/classNames";
+import { CAL_URL } from "@calcom/lib/constants";
 import { APP_NAME, COMPANY_NAME, SUPPORT_MAIL_ADDRESS } from "@calcom/lib/constants";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
+import type { RouterOutputs } from "@calcom/trpc/react";
 import { trpc } from "@calcom/trpc/react";
-import type { App as AppType } from "@calcom/types/App";
+import type { App as AppType, AppFrontendPayload } from "@calcom/types/App";
```
O4epegb, line 18: This is the only change basically, everything else is autoformat etc

O4epegb:
And now after the fix there are 55 eslint-prettier errors, should I fix them here too? Or make them warnings?

nicktrn:
Can't help here but came to say well done! Saw that error before myself but never got the full call stack, so decided to ignore it in the end. Others have had the problem too, see #10292.

emrysal (dismissed):
Amazing fix! Turns out our lint:report command has been failing silently and our lint analyse actually succeeding, so the errors that you found made it in over the last week.

lgtm

emrysal (approved):

PR #10492, opened 2023-07-31
fix: Address translation problem

### LINKED ISSUE #10476: [CAL-2274] In Person Attendee Address translation key is shown instead of value.
If In Person(Attendee Address) is selected, the Current Location during 'Edit Location' flow of the booking shows the key instead of the label.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2274](https://linear.app/calcom/issue/CAL-2274/in-person-attendee-address-translation-key-is-shown-instead-of-value)</sub>

anikdhabal:
## What does this PR do?

Before:-
Screenshot 2023-08-01 045220
Screenshot 2023-08-01 045237

Now:-
Screenshot 2023-08-01 045154
Screenshot 2023-08-01 045118

Fixes #10476 

## Requirement/Documentation

## Type of change

## Mandatory Tasks

### DIFF
--- a/packages/app-store/locations.ts
+++ b/packages/app-store/locations.ts
@@ -303,13 +303,13 @@ export const getHumanReadableLocationValue = (
 
   // Just in case linkValue is a `locationType.type`(for old bookings)
   const eventLocationType = getEventLocationType(linkValue);
-
+  const isDefault = eventLocationType?.default;
   if (eventLocationType) {
     // If we can find a video location based on linkValue then it means that the linkValue is something like integrations:google-meet and in that case we don't have the meeting URL to show.
     // Show a generic message in that case.
-    return `${eventLocationType.label}`;
+    return isDefault ? translationFunction(eventLocationType.label) : `${eventLocationType.label}`;
   }
-  // Otherwise just show the available link value which can be a Phone number, a URL or a physical address of a place.
+  // Otherwise just show the available link value.
   return linkValue || "";
 };
 


Udit-takkar (commented):
@anikdhabal works fine but i am not sure this is what we want. we should display the actual value of In person address or phone number that the attendee or organiser entered like what what we see on the booking page 

cc: @hariombalhara

hariombalhara (approved):
This code is seriously messy:( But you fixed it in the best possible way I can think of. Thanks @anikdhabal 🙏

hariombalhara:
@Udit-takkar, @anikdhabal's explanation is correct here.

packages/app-store/locations.ts
```diff
@@ -303,13 +303,13 @@ export const getHumanReadableLocationValue = (
 
   // Just in case linkValue is a `locationType.type`(for old bookings)
   const eventLocationType = getEventLocationType(linkValue);
-
+  const isDefault = eventLocationType?.default;
   if (eventLocationType) {
     // If we can find a video location based on linkValue then it means that the linkValue is something like integrations:google-meet and in that case we don't have the meeting URL to show.
     // Show a generic message in that case.
-    return `${eventLocationType.label}`;
+    return isDefault ? translationFunction(eventLocationType.label) : `${eventLocationType.label}`;
```
hariombalhara, line 310: @anikdhabal I just thought after the merge. We should translate both the labels, default and non-default. Right now  App location labels aren't using translation keys but when they start to do it, they will face the same issue.
Screenshot 2023-08-01 at 5 00 49 PM

anikdhabal: > @anikdhabal I just thought after the merge. We should translate both the labels, default and non-default. Right now App location labels aren't using translation keys but when they start to do it, they will face the same issue. Screenshot 2023-08-01 at 5 00 49 PM

Yes, it's a valid point. @hariombalhara okay will make a PR then.

PR #10513, opened 2023-08-01
fix: random multiple duration value input [CAL-887]

### LINKED ISSUE #6433: Custom duration isn't loaded from the URL

### Issue Summary

My link is:
https://cal.com/abc/my-meeting?duration=60
Ideally when the page loads it would select 60 minutes but it still selects my default of 30 minutes.

### Steps to Reproduce

1. Create an event with "Allow booker to select duration". Add two durations
2. Go to event link and select the non default link. The url will now update
3. Refresh the page - the selected slot is still the default and not the one from the URL.

leog:
## What does this PR do?

New booker allowed any duration to be passed by query parameter to override event type duration, even a non-multiple-duration event-type.

Fixes #6433 

## Type of change

## How should this be tested?

- Go to `/pro/30min` event-type and add `?duration=45` to the URL, which will be deleted and ignored now. 
- Go to `/pro/multiple-duration` event-type and modify the 75 default value for the `duration` query parameter to an allowed value and see it go through, and modify it to a non-allowed value and see it change to the default value 75

## Mandatory Tasks


apps/web/components/Embed.tsx
```diff
@@ -1001,6 +1001,7 @@ const EmailEmbed = ({ eventType, username }: { eventType?: EventType; username:
     eventSlug: eventType?.slug ?? "",
     eventId: eventType?.id,
     layout: BookerLayouts.MONTH_VIEW,
+    durationConfig: eventType?.metadata?.multipleDuration,
```
leog, line 1004: Passing down the multiple duration config to delegate duration calculation

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -100,6 +100,7 @@ const BookerComponent = ({
     layout: defaultLayout,
     isTeamEvent,
     org: entity.orgSlug,
+    durationConfig: event?.data?.metadata.multipleDuration,
```
leog, line 103: Passing down the multiple duration config to delegate duration calculation

packages/features/bookings/Booker/store.ts
```diff
@@ -25,6 +25,7 @@ type StoreInitializeType = {
   verifiedEmail?: string | null;
   rescheduleUid?: string | null;
   seatReferenceUid?: string;
+  durationConfig?: number[] | null;
```
leog, line 28: We now accept a duration configuration to enable calculation of allowed durations

CarinaWolli (approved):
LGTM!

PR #10521, opened 2023-08-02
chore: mailhog

shivamklr:
## What does this PR do?

Enable Mailhog for e2e tests. These changes should be non-breaking and should only start the mailhog container for future commit usage. 
This PR should be merged prior to the PR to test PoC for mailhog with the Github workflows.

## Requirement/Documentation

Setup the following github secrets
- EMAIL_SERVER_HOST
- EMAIL_SERVER_PORT
- EMAIL_SERVER_USER
- EMAIL_SERVER_PASSWORD

## Type of change


shivamklr:
@keithwillcode For the reference I am using these exact .env variables locally. And the mailhog service is able to capture the emails.
```
EMAIL_SERVER_HOST='localhost'
EMAIL_SERVER_PORT=1025
EMAIL_SERVER_USER='<office365_emailAddress>'
EMAIL_SERVER_PASSWORD='<office365_password>'
```

keithwillcode (commented):
We should probably use these same settings for the other e2e jobs as well, right?

nicktrn:
Just a quick heads up that I plan to open an RFC today re fully parallel e2e email testing. Does not require messing around with mailhog, auth, new env vars, etc. Did some prototyping and think this could work well.

zomars (commented):
Do we need to do this in e2e-appstore and embed as well?

shivamklr:
@zomars If they use the same endpoints/components then no.

shivamklr:
@zomars Just to clear things up. I will be doing this for the core e2e first. Once we cover that, we can move the changes so it covers all the e2e.

keithwillcode (approved):

PR #10536, opened 2023-08-02
fix: Redact prisma errors

hariombalhara:
## What does this PR do?

Redacts error coming from Prisma through 
- tRPC
- public API endpoints. 
- Page routes gSSP and gSP function call errors
While doing this it allows easily redacting any other error

## Error from API
**Before**
image
**After**
image

## tRPC Error 
**Before**
image
**After**
image

## Page Route Error redaction
image

## Follow up
- [ ] Handle Next.js api endpoints - There is no middleware for them right now. So, it would be done as a followup. Using next-auth-middleware might be a good option. We use the same in public API

## Type of change

## How should this be tested?
Change username in DB connection string to get a connection error and then
1. Test any tRPC endpoint that connects to DB
2. Test any public API endpoint that connects to DB

## Mandatory Tasks

## Checklist


apps/api/lib/helpers/captureErrors.ts
```diff
@@ -1,14 +1,20 @@
...
+import { redactError } from "@calcom/lib/redactError";
+
 export const captureErrors: NextMiddleware = async (_req, res, next) => {
   try {
     // Catch any errors that are thrown in remaining
     // middleware and the API route handler
     await next();
   } catch (error) {
     Sentry.captureException(error);
-    console.log(error);
```
hariombalhara, line 11: redactError takes care of logging the error.

apps/api/lib/helpers/withMiddleware.ts
```diff
@@ -28,10 +28,17 @@ const withMiddleware = label(
     customPrismaClient,
     extendRequest,
     pagination: withPagination,
-    sentry: captureErrors,
+    captureErrors,
```
hariombalhara, line 31: captureErrors is not sentry specific.

apps/api/lib/helpers/captureErrors.ts
```diff
@@ -1,14 +1,20 @@
...
+    const redactedError = redactError(error);
+    if (redactedError instanceof Error) {
```
keithwillcode, line 14: Is returning 400 (Bad Request) the correct error here? Are we able to properly return a 500 when it's an unexpected error or is it not possible at the moment based on how we are processing errors?

keithwillcode (approved):
Nice work. Just had a comment on the 400 vs 500 status code

PR #10546, opened 2023-08-02
fix: profile banner delay for orgs + small DX chore

leog:
## What does this PR do?

When creating an org, the owner should see a profile banner to set up their profile once they land on the app. That was delayed since we didn't trigger rendering the corresponding component once things change.

image

Also, when I was creating an org to replicate the issue and see the solution, noticed we sent the email with the code for lower envs either way and it was not needed since we introduced random totp codes for lower envs to stop relying on unstable dev mail services.

## Type of change

## How should this be tested?

Have orgs enabled locally, create one going to '/settings/organizations/new` and after org wizard, see the banner appear when you land in the app.

## Mandatory Tasks


apps/web/pages/event-types/index.tsx
```diff
@@ -929,7 +929,7 @@ const EventTypesPage = () => {
       !!orgBranding && !document.cookie.includes("calcom-profile-banner=1") && !user?.completedOnboarding
     );
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [orgBranding, user]);
```
leog, line 932: This is the missing ingredient to fix this

packages/features/shell/Shell.tsx
```diff
@@ -132,7 +132,6 @@ function useRedirectToLoginIfUnauthenticated(isPublic = false) {
 }
 
 function AppTop({ setBannersHeight }: { setBannersHeight: Dispatch<SetStateAction<number>> }) {
-  const router = useRouter();
```
leog, line 135: Unneeded code and also deprecated now

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -131,6 +136,7 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
 
     return { user: { ...createOwnerOrg, password } };
   } else {
+    if (!IS_PRODUCTION) return { checked: true };
```
leog, line 139: DX improvement to not send totp code email when creating an org on lower envs

apps/web/pages/event-types/index.tsx
```diff
@@ -929,7 +929,7 @@
       !!orgBranding && !document.cookie.includes("calcom-profile-banner=1") && !user?.completedOnboarding
     );
```
sean-brydon, line 930: <i class='graphite__hidden'>[Re: lines 928 to 930]<br/></i>

NIT - this entire set should be in a separate useEffect as it has different dependancies  compared to the show intercom bubble.

<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/calcom/cal.com/10546?utm_source=unchanged-line-comment">Graphite</a>.</span>

leog: Good catch, done

sean-brydon (approved):
Thanks for quick fix <3

PR #10556, opened 2023-08-03
fix: auth redirect

### LINKED ISSUE #10495: 'auth' Can be Registered as User
### Issue Summary
image

### Steps to Reproduce

1. try going to cal.com/auth

This is a bug and mostly a ambiguity bug.
### Actual Results

- Can create a user 'auth' into 
### Expected Results
User should not be able to create users with api keywords like auth.
Else:
path should be changed to somthing like localhost/user/{username}/{...}

### Evidence
image
image

I was able to create event with login leading to path as localhost/auth/login-> Now technically points to 2 locations
When I tried : localhost/auth-> 
image

When I clicked Login Page it led me Login in page rather than to calendar for scheduling

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/10495

https://github.com/calcom/cal.com/assets/53316345/b47f9965-316b-4397-9f5f-42fdd497dc61

## Type of change

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -379,7 +379,7 @@ const nextConfig = {
       {
         source: "/auth",
         destination: "/auth/login",
-        permanent: true,
+        permanent: false,
       },
       {
         source: "/settings",
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -377,6 +377,11 @@ const nextConfig = {
         permanent: true,
       },
       {
+        source: "/auth",
+        destination: "/auth/login",
+        permanent: true,
+      },
+      {
         source: "/settings",
         destination: "/settings/my-account/profile",
         permanent: true,


apps/web/next.config.js
```diff
@@ -376,6 +376,11 @@ const nextConfig = {
         destination: "/signup",
         permanent: true,
       },
+      {
+        source: "/auth",
+        destination: "/auth/login",
+        permanent: true,
+      },
```
Udit-takkar, line 383: https://github.com/calcom/cal.com/issues/10495#issuecomment-1663574950

apps/web/next.config.js
```diff
@@ -376,6 +376,11 @@ const nextConfig = {
         destination: "/signup",
         permanent: true,
       },
+      {
+        source: "/auth",
+        destination: "/auth/login",
+        permanent: true,
```
hariombalhara, line 382:
```suggestion
        permanent: false,
```

This is a temporary thing so we shouldn't make it permanent. Once a permanent redirect is done on browser, it is cached forever and can't be changed

hariombalhara (changes requested):
Left a blocking comment, otherwise it's good

hariombalhara (approved):

PR #10557, opened 2023-08-03
perf: When a Google access token is expired and can't be refreshed, w…

emrysal:
…rite invalid state.

This PR is the start of doing this; but only enables it on Google Calendar.


keithwillcode (approved):
Looks good. Agree with handling the invalid_grant situation better moving forward.

PR #10569, opened 2023-08-03
feat: secret booking questions

### LINKED ISSUE #10365: Secret booking questions
Can you please add the function that you can set booking questions as "secret"? So that you or the owner of the appointment can see these questions or answers, but the guest does not get to see them?

Example: Sales teams often exchange their internal notes on appointments with other sales people. For example, a qualification meeting takes place with a prospect, and the salesperson takes notes on the meeting. Afterwards, this salesperson then schedules a sales call with the prospect, but another salesperson conducts the call. Therefore, the 1st sales person now makes the appointment booking via Cal.com for the sales call, using the appointment link of the 2nd sales person (or via Round Robin). In order for salesperson 2 to receive the notes from the qualification meeting, a booking question is stored in his calendar for this purpose, where salesperson 1 stores all the notes.

The problem with this is that the prospect also receives all the notes from the call, as these are displayed on the confirmation page (when rescheduling and canceling the appointment), as well as in their appointment confirmation email and in their calendar. However, this should not be the case, as these notes are only for internal sharing.

rkreddy99:
## What does this PR do?

Creates a new feature for adding secret questions in the booking flow.

Fixes #10365 
 Loom Video: https://www.loom.com/share/0fc24982eaa0488a94a0a79d3fd22259?sid=9f95e41b-cbef-4ff4-b31a-61fd89d63b62

## Type of change

## How should this be tested?

- Enable secret question by adding field in advanced section of event
- Book the event, the secret question should only be shared with the owner and not the guests.

## Mandatory Tasks

## Checklist

/claim #10365

EDIT: 
Updated video:

https://www.loom.com/share/d50bf1ac34aa42f5b952a15cb1eec949?sid=3df385f3-c71a-41f4-917d-36e9200e7ba8


rkreddy99:
@PeerRich @baileypumfleet (i dont know if there is any specific team member to tag to discuss this)

I need inputs to solve this issue

- For emails and calender events, i have taken care of sending the secret answer only to the organizer,
- not showing it on reschedule page can also be done easily
- I am stuck on how to hide this in the booking success page, on the ui i can handle it with if the logged in user is organizer then show o/w not. But this will still be present in the network calls section right. 

- Am I overthinking this issue, and just use loggedin user cond and show accordingly? 
- Is this case too specific and should not solve it? 
- Or should  i add another column in db for secret responses?

rkreddy99:
**INCOGNITO TAB IS BOOKERS PERSPECTIVE**

https://www.loom.com/share/d50bf1ac34aa42f5b952a15cb1eec949?sid=3df385f3-c71a-41f4-917d-36e9200e7ba8

rkreddy99:
Hi @PeerRich @baileypumfleet 

Is this feature still required? If yes, could you please review the PR? Thanks.

tagging @CarinaWolli since she has worked on booking flow. please ignore if you are not the correct person.

packages/features/bookings/Booker/components/BookEventForm/BookingFields.tsx
```diff
@@ -29,6 +29,7 @@ export const BookingFields = ({
     // The logic here intends to make modifications to booking fields based on the way we want to specifically show Booking Form
     <div>
       {fields.map((field, index) => {
+        if (field.editable === "user-owner-secret" && currentView === "reschedule") return null;
```
zomars, line 32: Can we add a comment here explaining why this is needed?

zomars (commented):
The feature works as intended. My main concern is on the code to achieve this. It feels quite intrusive for a low priority feature. I think we can achieve it in a non intrusive way. Tagging @hariombalhara to get his opinion on this.

packages/features/bookings/Booker/components/BookEventForm/BookingFields.tsx
rkreddy99: @zomars Idea was that a secret question and response should not be visible to a user once booking is completed. Now since if we click on reschedule link and can see all the responses, I thought of adding this condition to not show that.

packages/types/Calendar.d.ts
```diff
@@ -188,6 +188,8 @@ export interface CalendarEvent {
 
   // It just has responses to only the user fields. It allows to easily iterate over to show only user fields
   userFieldsResponses?: CalEventResponses | null;
+  // It just has responses to only the user-owner-secret fields. It allows to easily iterate over to show only user fields
+  userOwnerSecretFieldsResponses?: CalEventResponses | null;
```
hariombalhara, line 192: We shouldn't introduce a new property here that could just be an attribute in userFieldsResponse. I think we should have a prop `isSecretQuestion` per field response in userFieldsResponses.

packages/features/form-builder/schema.ts
```diff
@@ -28,6 +28,7 @@ export const EditableSchema = z.enum([
   "system-but-hidden", // Can't be deleted, name can't be edited, will be shown
   "user", // Fully editable
   "user-readonly", // All fields are readOnly.
+  "user-owner-secret", // Only shared between owner and attendee, no one else can see this response
```
hariombalhara, line 31: The idea of EditableSchema was that it talks about different editability scenarios of a field. It doesn't make sense to have a `user-owner-secret` here.

apps/web/pages/booking/[uid].tsx
```diff
@@ -1068,7 +1073,23 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
...
-  const bookingInfo = getBookingWithResponses(bookingInfoRaw);
+  const bookingInfoWithSecretResponses = getBookingWithResponses(bookingInfoRaw);
+
+  const filteredResponses = bookingInfoWithSecretResponses.responses;
+
+  if (!session || eventTypeRaw.owner?.id !== session.user.id) {
+    const bookingFields = eventTypeRaw.bookingFields;
+    const parsedBookingFields = bookingFields ? eventTypeBookingFields.parse(bookingFields) : null;
+    parsedBookingFields?.forEach(
+      (field) => field.editable === "user-owner-secret" && delete filteredResponses[field.name]
```
Comment (hariombalhara) [inline apps/web/pages/booking/[uid].tsx:1084]: Couldn't we check for `field.secretQuestion` here?

Comment (hariombalhara) [reply apps/web/pages/booking/[uid].tsx:1084]: Instead of doing a delete on `filteredResponses` which is actually a reference of `bookingInfoWithSecretResponses.responses`, we should use array `filter` here.

packages/emails/src/components/UserOwnerSecretFieldsResponses.tsx
hariombalhara, line 1: We should be able to avoid this new Component and module by having `isSecret` prop on userFields https://github.com/calcom/cal.com/pull/10569/files#r1302863711

hariombalhara:
@rkreddy99 Could you define a secret question in the PR description itself clearly.

apps/web/pages/booking/[uid].tsx
```diff
@@ -1068,7 +1073,23 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
...
+    );
+  }
```
Comment (hariombalhara) [inline apps/web/pages/booking/[uid].tsx:1086]: We should refactor it to move out of the component in it's own function

rkreddy99:
> @rkreddy99 Could you define a secret question in the PR description itself clearly.

Check this comment from the issue https://github.com/calcom/cal.com/issues/10365#issuecomment-1664300884

hariombalhara:
I am also thinking that should we call it a **secret question** or an **internal note**? Making it an internal note would mean that it should be accessible to all the team members and the person who posted that(In the demo as the booker is providing that, he can access that note)

Right now, it is not accessible to other team members which I think it should be.

Also, by making it an internal note we can provide other ways to set that note e.g. through API or say during/after the meeting with some UI. 

Overall, I think we should create specs for it before we implement and ship it.

cc @ciaranha

ciaranha:
> I am also thinking that should we call it a **secret question** or an **internal note**? Making it an internal note would mean that it should be accessible to all the team members and the person who posted that(In the demo as the booker is providing that, he can access that note)
> 
> Right now, it is not accessible to other team members which I think it should be.
> 
> Also, by making it an internal note we can provide other ways to set that note e.g. through API or say during/after the meeting with some UI.
> 
> Overall, I think we should create specs for it before we implement and ship it.


"Internal note" is a much better term for this than a "secret question". 

Also, thoughts on this being a completely separate question option @hariombalhara? I don't like that we are going to present this very niche feature request to every single additional question that every user adds. It's pretty unique to want to show a booking question on the booking page, but not after IMO.

CleanShot 2023-10-11 at 06 01 40@2x
CleanShot 2023-10-11 at 06 04 33@2x
CleanShot 2023-10-11 at 06 04 39@2x

I do think it would be great to have this last "info" bit also shown in the dropdown when selecting items but we could look at that later.

PeerRich:
going with an "internal note" approach instead https://github.com/calcom/cal.com/issues/12222

PR #10572, opened 2023-08-03
fix: Only enforce min seats for orgs

joeauyeung:
## What does this PR do?

This PR fixes the logic to only change team subscription amounts for orgs and not normal teams

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Enable team billing
- Create a team and invite members
- Delete some members
- On Stripe the subscription count should be updated

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/lib/constants.ts
+++ b/packages/lib/constants.ts
@@ -83,8 +83,7 @@ export const IS_STRIPE_ENABLED = !!(
   process.env.STRIPE_PRIVATE_KEY
 );
 /** Self hosted shouldn't checkout when creating teams unless required */
-// export const IS_TEAM_BILLING_ENABLED = IS_STRIPE_ENABLED && (!IS_SELF_HOSTED || HOSTED_CAL_FEATURES);
-export const IS_TEAM_BILLING_ENABLED = true;
+export const IS_TEAM_BILLING_ENABLED = IS_STRIPE_ENABLED && (!IS_SELF_HOSTED || HOSTED_CAL_FEATURES);
 export const FULL_NAME_LENGTH_MAX_LIMIT = 50;
 export const MINUTES_TO_BOOK = process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "5";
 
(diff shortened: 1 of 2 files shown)


zomars (changes requested):
We have done types failing

zomars (changes requested):
Some Qs, I'm trying to understand the why of this.

- Where does the ORGANIZATION_MIN_SEATS number coming from?
- Will this affect non orgs teams sub update logic? How can we verify this?
- Is this related to the previous discussion where repeated members of different teams of the same org should not be charged?

packages/lib/constants.ts
```diff
@@ -90,3 +90,5 @@ export const MINUTES_TO_BOOK = process.env.NEXT_PUBLIC_MINUTES_TO_BOOK || "5";
 // Needed for orgs
 export const ALLOWED_HOSTNAMES = JSON.parse(`[${process.env.ALLOWED_HOSTNAMES || ""}]`) as string[];
 export const RESERVED_SUBDOMAINS = JSON.parse(`[${process.env.RESERVED_SUBDOMAINS || ""}]`) as string[];
+
+export const ORGANIZATION_MIN_SEATS = 30;
```
alannnc, line 94: I think we can move this to it's own file? Unless it's being used multiple times on several files.

alannnc (approved):
Looks good, left a NIT comment.

packages/lib/constants.ts
joeauyeung: Do you have a suggestion on where to move it? I figured the constants file would be an easy place to find this variable.

joeauyeung:
> Some Qs, I'm trying to understand the why of this.
> 
>     * Where does the ORGANIZATION_MIN_SEATS number coming from?
> 
>     * Will this affect non orgs teams sub update logic? How can we verify this?
> 
>     * Is this related to the previous discussion where repeated members of different teams of the same org should not be charged?

- The minimum number is coming from @shirazdole
- The problem was this logic before was affecting non-org teams. I tested this locally and checked the test Stripe dashboard, and the number of seats the customer was being charged is being updated with this fix. Regular teams will have normal members and not orgMembers
- That logic is handled somewhere else. This is only for removing team members

zomars (approved):
Thanks fro clarifying. LGTM

LGTM

PR #10576, opened 2023-08-03
perf: call getSchedule at the start 

### LINKED ISSUE #10564: [CAL-2307] Unneeded wait time with getSchedule after session
> Fix issue where we wait to load getSchedule after the session call has completed. It’s a public endpoint and doesn’t need a session, correct?

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2307](https://linear.app/calcom/issue/CAL-2307/unneeded-wait-time-with-getschedule-after-session)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/10564

slots.getSchedule api is now called as the same time as session and public.event instead of waiting for public.event for 0.5s - 3s+ earlier. 

Screenshot 2023-08-03 at 11 20 39 PM

## Type of change

- [x] Performance improvement


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -165,10 +165,6 @@ const BookerComponent = ({
 
   const shouldShowFormInDialog = shouldShowFormInDialogMap[layout];
 
-  if (bookerState === "loading") {
-    return null;
-  }
-
```
Udit-takkar, line 171: this was blocking slots.getSchedule( inside AvailableTimeSlots) till event loading finished.

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -36,17 +41,14 @@ export const useSchedule = ({
       // if `prefetchNextMonth` is true, two months are fetched at once.
       endTime: (prefetchNextMonth ? nextMonthDayjs : monthDayjs).endOf("month").toISOString(),
       timeZone: timezone!,
-      eventTypeId: eventId!,
       duration: duration ? `${duration}` : undefined,
+      eventTypeId: eventId,
+      org,
+      isTeamEvent,
```
Udit-takkar, line 47: we need either eventTypeId or org, isTeamEvent, username and eventTypeSlug to get the eventTypeId on the server side

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -36,17 +41,14 @@ export const useSchedule = ({
...
     },
     {
       refetchOnWindowFocus: false,
-      enabled:
-        Boolean(username) &&
-        Boolean(eventSlug) &&
-        (Boolean(eventId) || eventId === 0) &&
-        Boolean(month) &&
-        Boolean(timezone),
+      enabled: Boolean(username) && Boolean(eventSlug) && Boolean(month) && Boolean(timezone),
```
Udit-takkar, line 51: we don't depend on eventTypeId now so we don't wait for useEvent request to return response and then start with this query (slots.getSchedule) and get eventTypeId on slots.getSchedule

keithwillcode:
@Udit-takkar In the screenshot you posted in the description, it shows that `slots.getSchedule` is called alone but I'm seeing it get called with `session` and `event`. Is that expected?
Screen Shot 2023-08-04 at 4 32 23 PM

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -249,6 +249,71 @@ export const getPublicEvent = async (
   };
 };
 
+export const getPublicEventId = async (
```
Udit-takkar, line 252: It returns eventId and doesn't fetch all the data like getPublicEvent()

packages/features/bookings/Booker/Booker.tsx
hariombalhara: This would cause again the flicker that was fixed [here](https://github.com/calcom/cal.com/pull/9873) 
Maybe we can keep the booker children invisible instead till the state is loading. 
https://github.com/calcom/cal.com/pull/10588/files#r1285364468

apps/web/components/Embed.tsx
```diff
@@ -1058,7 +1058,7 @@ const EmailEmbed = ({ eventType, username }: { eventType?: EventType; username:
     shallow
   );
   const event = useEvent();
-  const schedule = useScheduleForEvent();
+  const schedule = useScheduleForEvent({ eventId: eventType?.id });
```
hariombalhara, line 1061: Not passing the org here?

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -200,6 +201,19 @@ export async function getSchedule(input: TGetScheduleInputSchema) {
     logger.setSettings({ minLevel: "silly" });
   }
   const startPrismaEventTypeGet = performance.now();
+
+  if (!input.eventTypeId && input.eventTypeSlug && !!input.usernameList) {
```
hariombalhara, line 205: I think we shouldn't get eventTypeId here. The Booker already had the eventTypeId and it should pass the info to getSchedule.

Even if we need to do this, let's abstract it out in a function like getEventTypeId

Udit-takkar: actually the booker get the eventTypeId from trpc.event.public so slots.getSchedule would only run after trpc.event.public ( in useEvent) return us result but here i created a function getPublicEventId which does what trpc.public.event ( getPublicEvent) does but fetches nothing extra only id so it should be a bit faster and we don't wait for trpc.public.event api to finish

packages/features/bookings/Booker/Booker.tsx
Udit-takkar: okay thanks

apps/web/components/Embed.tsx
Udit-takkar: yeah because we are passing eventId directly here

Screenshot 2023-08-08 at 6 19 54 PM

packages/trpc/server/routers/viewer/slots/util.ts
hariombalhara: public.event is prefetched on server, so on client it should be available on first render. There was a bug in that, which I fixed in my PR

Udit-takkar:
@hariombalhara now i am getting the eventTypeId from prefetched response. removed the server side logic

Udit-takkar:
https://github.com/calcom/cal.com/pull/10607 this has been merged

PR #10580, opened 2023-08-03
chore: optimize `ssrInit` method with `Promise.all`

O4epegb:
## What does this PR do?

Tiny optimization for `ssrInit` method. Right now it has 3 sequential async calls, but it would be better to use `Promise.all` instead. On my local machine it makes this call almost 2x faster (from 30ms to 15ms on average), it's really tiny gain, but it's basically free, so why not. Not sure how big the increase will be on production, but anyway, seems a good practice to not have waterfall calls like that. There are some other places like that, in the end it could be nice improvement piled up.

## Type of change

## Mandatory Tasks

### DIFF
--- a/apps/web/server/lib/ssr.ts
+++ b/apps/web/server/lib/ssr.ts
@@ -16,7 +16,7 @@ import { appRouter } from "@calcom/trpc/server/routers/_app";
 export async function ssrInit(context: GetServerSidePropsContext) {
   const ctx = await createContext(context);
   const locale = getLocaleFromHeaders(context.req);
-  const i18n = await serverSideTranslations(getLocaleFromHeaders(context.req), ["common", "vital"]);
+  const i18n = await serverSideTranslations(locale, ["common", "vital"]);
 
   const ssr = createProxySSGHelpers({
     router: appRouter,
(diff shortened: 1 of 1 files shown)


apps/web/server/lib/ssr.ts
```diff
@@ -16,20 +16,22 @@ import { appRouter } from "@calcom/trpc/server/routers/_app";
...
     ctx: { ...ctx, locale, i18n },
   });
 
-  // always preload "viewer.public.i18n"
-  await ssr.viewer.public.i18n.fetch();
-  // So feature flags are available on first render
-  await ssr.viewer.features.map.prefetch();
-  // Provides a better UX to the users who have already upgraded.
-  await ssr.viewer.teams.hasTeamPlan.prefetch();
+  await Promise.all([
```
Udit-takkar, line 27: we are trying to reduce the usage of Promise.all to prevent reaching the limit of database connections. not sure if we should do this. what do you think @emrysal  ?

O4epegb: Oh, didn't know that! Well, if so, then yes, it will make 1 additional connection here.

PeerRich: https://www.youtube.com/watch?v=f2Z1v3cqgDI

O4epegb: > https://www.youtube.com/watch?v=f2Z1v3cqgDI

Agreed that it can be overused, but in this case we just need to run 3 queries with no side effects and no error handling, and since this method runs basically on every page, it makes sense to speed it up a little bit.

If the goal is to limit db connections, then I agree that it's not a good idea to parallelize things, but at the same time I think it would be better to solve this problem with a db bouncer instead of relying on waterfall code?

PR #10581, opened 2023-08-03
fix: random test failure when booking are on same date

zomars:
## What does this PR do?

Fixes random tests failures introduced in #10448

## Requirement/Documentation

## Type of change

## How should this be tested?

- Run `yarn test` multiple times. It should always succeed

## Mandatory Tasks

### DIFF
--- a/apps/web/test/lib/team-event-types.test.ts
+++ b/apps/web/test/lib/team-event-types.test.ts
@@ -13,18 +13,21 @@ it("can find lucky user with maximize availability", async () => {
     email: "test@example.com",
     bookings: [
       {
-        createdAt: new Date("2022-01-25"),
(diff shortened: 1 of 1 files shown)


apps/web/test/lib/team-event-types.test.ts
```diff
@@ -13,18 +13,21 @@ it("can find lucky user with maximize availability", async () => {
...
   });
   const user2 = buildUser({
-    id: 1,
-    username: "test",
-    name: "Test User",
-    email: "test@example.com",
+    id: 2,
+    username: "test2",
+    name: "Test User 2",
+    email: "tes2t@example.com",
```
zomars, line 27: Both users had same id, email, etc. Making this test more confusing than needed.

alannnc (approved):
Nice catch

PR #10582, opened 2023-08-03
fix: don't show team unaccepted members

leog:
## What does this PR do?

When adding team members, unaccepted members will now not show up as options. This is just a feature for organizations. Also fixes the broken page when adding members in a subteam from the team creation wizard, which had an unstyled border.

## Type of change

## How should this be tested?

Try to create a regular team, add members and create a Collective/Round robin event type, see that pending members will not be an option. On the other hand, for org subteams, pending members will be an option when creating a subteam event-type.

## Mandatory Tasks


packages/features/ee/teams/components/AddNewTeamMembers.tsx
```diff
@@ -213,7 +213,10 @@ const PendingMemberItem = (props: { member: TeamMember; index: number; teamId: n
   return (
     <li
       key={member.email}
-      className={classNames("flex items-center justify-between p-6 text-sm", index !== 0 && "border-t")}
+      className={classNames(
+        "flex items-center justify-between p-6 text-sm",
+        index !== 0 && "border-subtle border-t"
```
leog, line 218: Border color missing, on dark mode it was being shown as really white

packages/lib/getEventTypeById.ts
```diff
@@ -120,6 +120,7 @@ export default async function getEventTypeById({
           members: {
             select: {
               role: true,
+              accepted: true,
```
leog, line 123: Need to know acceptance status for members

packages/lib/getEventTypeById.ts
```diff
@@ -264,8 +265,6 @@ export default async function getEventTypeById({
     giphy: getEventTypeAppData(eventTypeWithParsedMetadata, "giphy", true),
   };
 
-  // TODO: How to extract metadata schema from _EventTypeModel to be able to parse it?
-  // const parsedMetaData = _EventTypeModel.parse(newMetadata);
```
leog, line 268: Unneeded comment, metadata is already there

packages/lib/getEventTypeById.ts
```diff
@@ -363,18 +362,21 @@ export default async function getEventTypeById({
     bookingFields: getBookingFieldsWithSystemFields(eventType),
   });
 
+  const isOrgEventType = !!eventTypeObject.team?.parentId;
```
leog, line 365: Establishing org context for the event-type

packages/lib/getEventTypeById.ts
```diff
@@ -363,18 +362,21 @@ export default async function getEventTypeById({
...
-          avatar: `${CAL_URL}/${member.user.username}/avatar.png`,
-        };
-        return {
-          ...user,
-          eventTypes: user.eventTypes.map((evTy) => evTy.slug),
-          membership: member.role,
-        };
-      })
+    ? eventTypeObject.team.members
+        .filter((member) => member.accepted || isOrgEventType)
```
leog, line 368: Filtering member if is not accepted and there is no org context

packages/lib/server/queries/teams/index.ts
```diff
@@ -101,8 +101,6 @@ export async function getTeamWithMembers(args: {
   if (userId) where.members = { some: { userId } };
   if (orgSlug) {
     where.parent = getSlugOrRequestedSlug(orgSlug);
-  } else {
-    where.parentId = null;
```
leog, line 105: This was preventing the onboarding members step in team creation wizard to load, since was setting `parentId` to null for a subteam, which has a `parentId`

packages/lib/getEventTypeById.ts
zomars, line 368: Shouldn't it be?
```suggestion
        .filter((member) => member.accepted || !isOrgEventType)
```

leog: Not really, we need to filter out (return false) only when `Accepted = F` and `isOrgEventType = F`. With your alternative we would be filtering out members when `Accepted = F` and `!isOrgEventType = F => isOrgEventType = T` and we want that case to return the member.

| P = Accepted    | Q = isOrgEventType | P ∨ Q |  !Q | P ∨ !Q |
| -------- | ------- | ------- | ------- | ------- |
| T | T | T | F | T |
| T | F | T | T | T |
| F | T | T | F | F |
| F | F | F | T | T |

Udit-takkar (approved):
LGTM

sean-brydon (approved):
Fixes the issue :) Nice!

PR #10588, opened 2023-08-04
Booker Perf improvements

hariombalhara:
## What does this PR do?

- Earlier BookerComponent wasn't rendering child components when bookerState was **loading** and thus delaying `getSchedule` call. Now, it's rendered in a hidden state and thus the call isn't delayed and sent along with session.
- `username`, `eventSlug`, `isTeamEvent`, `org` values are available to Booker component and for them we don't need to wait for store to initialize. Let `useEvent` handle these initial values as well
- Prefetched event query on server is correctly reused now. Earlier it wasn't reused because a new property `isTeamEvent` was missing from prefetch query at server. So, even though **event** call is sent, that's just background prefetch of data.
	- [Refresh experience also improved due to it](https://www.loom.com/share/4d3596a607a749b19019f33e272872de)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/pages/[user]/[type].tsx
+++ b/apps/web/pages/[user]/[type].tsx
@@ -165,6 +165,7 @@ async function getUserPageProps(context: GetServerSidePropsContext) {
   const eventData = await ssr.viewer.public.event.fetch({
     username,
     eventSlug: slug,
+    isTeamEvent: false,
     org,
   });
 
--- a/apps/web/pages/_document.tsx
+++ b/apps/web/pages/_document.tsx
@@ -49,7 +49,7 @@ class MyDocument extends Document<Props> {
           <meta name="msapplication-TileColor" content="#ff0000" />
           <meta name="theme-color" media="(prefers-color-scheme: light)" content="#f9fafb" />
           <meta name="theme-color" media="(prefers-color-scheme: dark)" content="#1C1C1C" />
-          {(process.env.NODE_ENV === "development" || process.env.VERCEL_ENV === "preview") && (
+          {process.env.VERCEL_ENV === "preview" && (
             // eslint-disable-next-line @next/next/no-sync-scripts
             <script
               data-project-id="KjpMrKTnXquJVKfeqmjdTffVPf1a6Unw2LZ58iE4"
--- a/packages/features/bookings/Booker/Booker.tsx
+++ b/packages/features/bookings/Booker/Booker.tsx
@@ -13,6 +13,7 @@ import { BookerLayouts, defaultBookerLayoutSettings } from "@calcom/prisma/zod-u
 import { AvailableTimeSlots } from "./components/AvailableTimeSlots";
 import { BookEventForm } from "./components/BookEventForm";
 import { BookFormAsModal } from "./components/BookEventForm/BookFormAsModal";
+import { DatePicker } from "./components/DatePicker";
 import { EventMeta } from "./components/EventMeta";
 import { Header } from "./components/Header";
 import { LargeCalendar } from "./components/LargeCalendar";
@@ -28,9 +29,6 @@ import { useBrandColors } from "./utils/use-brand-colors";
 
 const PoweredBy = dynamic(() => import("@calcom/ee/components/PoweredBy"));
 const UnpublishedEntity = dynamic(() => import("@calcom/ui").then((mod) => mod.UnpublishedEntity));
-const DatePicker = dynamic(() => import("./components/DatePicker").then((mod) => mod.DatePicker), {
-  ssr: false,
-});
 
 const BookerComponent = ({
   username,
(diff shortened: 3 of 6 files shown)


packages/features/bookings/Booker/store.ts
```diff
@@ -304,6 +304,7 @@ export const useInitializeBookerStore = ({
   org,
 }: StoreInitializeType) => {
   const initializeStore = useBookerStore((state) => state.initialize);
+
```
hariombalhara, line 307:
```suggestion
```

packages/features/bookings/components/BookerSeo.tsx
```diff
@@ -16,12 +17,14 @@ interface BookerSeoProps {
...
 
 export const BookerSeo = (props: BookerSeoProps) => {
-  const { eventSlug, username, rescheduleUid, hideBranding, isTeamEvent, entity } = props;
+  const { eventSlug, username, rescheduleUid, hideBranding, isTeamEvent = false, entity } = props;
   const { t } = useLocale();
-  const { data: event } = trpc.viewer.public.event.useQuery(
-    { username, eventSlug, isTeamEvent, org: entity.orgSlug ?? null },
-    { refetchOnWindowFocus: false }
-  );
+  const { data: event } = useEvent({
```
hariombalhara, line 22: Reuse useEvent instead of using the query directly.

packages/features/bookings/Booker/utils/event.ts
```diff
@@ -14,15 +14,35 @@ import { useBookerStore } from "../store";
...
+    refetchOnWindowFocus: false,
+    enabled: isQueryEnabled,
+  });
 
-  return trpc.viewer.public.event.useQuery(
-    { username: username ?? "", eventSlug: eventSlug ?? "", isTeamEvent, org: org ?? null },
-    { refetchOnWindowFocus: false, enabled: Boolean(username) && Boolean(eventSlug) }
-  );
+  if (res.isLoading && isQueryEnabled) {
+    console.warn(
```
hariombalhara, line 39: This would help us in catching the case when event query stops being prefetched in future.

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -165,12 +170,8 @@ const BookerComponent = ({
 
   const shouldShowFormInDialog = shouldShowFormInDialogMap[layout];
 
-  if (bookerState === "loading") {
-    return null;
-  }
-
   return (
-    <>
+    <div className={bookerState === "loading" ? "invisible" : "visible"}>
```
hariombalhara, line 174: Instead of not rendering the child components, let them render invisible so that they can fetch their own data already.

apps/web/pages/_document.tsx
```diff
@@ -49,7 +49,7 @@ class MyDocument extends Document<Props> {
           <meta name="msapplication-TileColor" content="#ff0000" />
           <meta name="theme-color" media="(prefers-color-scheme: light)" content="#f9fafb" />
           <meta name="theme-color" media="(prefers-color-scheme: dark)" content="#1C1C1C" />
-          {(process.env.NODE_ENV === "development" || process.env.VERCEL_ENV === "preview") && (
+          {process.env.VERCEL_ENV === "preview" && (
```
hariombalhara, line 52: It was discussed earlier in threads that we don't need Meticulous on local.

packages/features/bookings/Booker/utils/event.ts
```diff
@@ -14,15 +14,35 @@ import { useBookerStore } from "../store";
...
+}) => {
   const [username, eventSlug] = useBookerStore((state) => [state.username, state.eventSlug], shallow);
   const isTeamEvent = useBookerStore((state) => state.isTeamEvent);
   const org = useBookerStore((state) => state.org);
+  const params = {
+    username: username ?? initialValues?.username ?? "",
+    eventSlug: eventSlug ?? initialValues?.eventSlug ?? "",
+    isTeamEvent: isTeamEvent ?? initialValues?.isTeamEvent ?? false,
+    org: org ?? initialValues?.org ?? null,
+  };
```
hariombalhara, line 31: Store doesn't provide these values on first render, Booker has them so, use the provided values as the initial values and later let the store handle any changes in them.

apps/web/pages/[user]/[type].tsx
```diff
@@ -165,6 +165,7 @@ async function getUserPageProps(context: GetServerSidePropsContext) {
   const eventData = await ssr.viewer.public.event.fetch({
     username,
     eventSlug: slug,
+    isTeamEvent: false,
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:168]: `isTeamEvent` was missing and this was the reason this prefetched query response was unused during hydration

PR #10604, opened 2023-08-04
perf: Use Intl.DateTimeFormat instead of Dayjs.tz().format()

emrysal:
## What does this PR do?

Comments added in code.

### DIFF
--- a/packages/trpc/server/routers/viewer/slots/util.ts
+++ b/packages/trpc/server/routers/viewer/slots/util.ts
@@ -408,17 +408,27 @@ export async function getSchedule(input: TGetScheduleInputSchema) {
   }
 
   availableTimeSlots = availableTimeSlots.filter((slot) => isTimeWithinBounds(slot.time));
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -408,17 +408,27 @@ export async function getSchedule(input: TGetScheduleInputSchema) {
   }
 
   availableTimeSlots = availableTimeSlots.filter((slot) => isTimeWithinBounds(slot.time));
+  // fr-CA uses YYYY-MM-DD
+  const formatter = new Intl.DateTimeFormat("fr-CA", {
+    year: "numeric",
+    month: "2-digit",
+    day: "2-digit",
+    timeZone: input.timeZone,
+  });
```
PeerRich, line 417: really we have to write these kind of exceptions 😆

keithwillcode (approved):
Confirmed this resulted in a huge perf improvement. Down to 6ms from 600ms.

PR #10560, opened 2023-08-03
test: add emails fixture and apply to existing tests

nicktrn:
## What does this PR do?

Adds fixture for email testing. Updates some tests for demo purposes.

```js
// this:
const { id } = await prisma.resetPasswordRequest.findFirstOrThrow({
  where: {
    email,
  },
  select: {
    id: true,
  },
  orderBy: {
    createdAt: "desc",
  },
})
await page.goto(`/auth/forgot-password/${id}`)

// becomes that:
const email = await emails.waitForOne(email)
await email.clickCta()

// while ensuring we receive the appropriate email, and that the cta works
```

Brief API overview:

```js
/* Fixture */

// string: just look for recipient
// object: filter by all email props - to, from, subject, (custom) headers, content, ..
type Filter = string | Record<string, string>

// wait until we receive an email matching a single filter
emails.waitForOne(filter: Filter): Email

// wait until we receive an email for each supplied filter
emails.waitForMany(filters: Filter[]): Email[]

// returns namespaced address that will be delivered to current test only
emails.generateAddress()


/* Email */

// renders email
email.open()

// renders email and clicks cta link
// shortcut for: await email.open() && page.getByTestId("cta-link").click()
email.clickCta()

// returns href via partial string match, e.g. "forgot" => "..example.com/forgot-password/"
email.findHref(partial: string)
```

Fixes #10535 

## Architectural overview

```mermaid
sequenceDiagram
    participant Client
    box Server
    participant WebSocket
    participant SMTP
    end
    participant Mailer

    Client-)WebSocket: subscribe
    Note right of Client: namespaced
    Mailer-)SMTP: send mail
    SMTP->>+WebSocket: get subscribers
    WebSocket-->>-SMTP: subscribers[ ]
    loop every subscriber
    SMTP-)WebSocket: forward mail
    end
    WebSocket-)Client: forward mail
    Note right of Client: namespaced
```

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How should this be tested?

```sh
yarn e2e playwright/{booking-pages,auth/forgot}
```
Runs fine locally, ~~but still some things to sort out for ci.~~ and now also on ci

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

Issue #10535, opened 2023-08-02
RFC: e2e email testing


nicktrn:
## Is your proposal related to a problem?
![emailissues](https://github.com/calcom/cal.com/assets/55853254/52104308-369a-48f4-94e5-599e77a74a3a)

I believe a lot of this could be prevented by integrating emails into current e2e test flows.

There are many existing features that would greatly benefit from this, e.g. password reset, email verification, opt-in bookings, reminders, confirmations, etc. But also proposed features such as #8007.

Many different attempts to use emails in tests have already been made, this is merely the proposal of yet another way.

### Current ideas
- querying the db https://github.com/calcom/cal.com/blob/65384d7b1fd57ed6f8000e3af751da969d4b8f55/apps/web/playwright/auth/forgot-password.e2e.ts#L21-L25
- using global vars https://github.com/calcom/cal.com/blob/65384d7b1fd57ed6f8000e3af751da969d4b8f55/apps/web/playwright/integrations.e2e.ts#L138-L144
- using mailhog + api #9437 #10521
![mailhog](https://github.com/calcom/cal.com/assets/55853254/cea5a768-0e3c-4f5f-b812-aa7e4d5ce083)

### Problems
- flaky tests
- non-e2e
- bypasses test isolation
- untested key features
- unmaintained software written in Go

## Describe the solution you'd like

```js
await bookFirstAvailable(page)
const email = await emails.waitForOne(hostEmailAddress)
await email.clickCta()
await expect(bookingConfirmedLocator).toBeVisible()
```

Key features:
- test isolation -> fully parallel support
- simple api that "just works"
- minimal or NO changes to existing code

## Describe alternatives you've considered

- external APIs / cloud offerings
- mailhog & co
- existing packages

I believe all of these are incomplete, cumbersome so extend, or have other issues that make them unworkable.

## Proposal

PR with working examples at #10560

nicktrn:
Tagging some people who have previously worked on this

@hariombalhara @keithwillcode @shivamklr

nicktrn:
Closing as per current email testing plans https://github.com/calcom/cal.com/pull/10881#issuecomment-1688185403

PR #10606, opened 2023-08-04
feat: mailhog fixture

shivamklr:
## What does this PR do?
This PR is a followup on these PRs, #10521 and #10637.
This PR is a PoC for mailhog that can be used to assert emails.

## Type of change

- [x] Chore/Test (refactoring code, technical debt, workflow improvements)

## How should this be tested?
- Make sure the `mailhog` container is running. Ideally, it should run with`yarn dx`. 
- The default env variables should work `EMAIL_SERVER_HOST='localhost'` and `EMAIL_SERVER_PORT=1025`.
- Test `booking-pages.e2e.ts`


nicktrn:
- #10535 
- #10560

leog (changes requested):
Code looks good, very good addition to our test suite.
I just couldn't run it locally to verify it works as expected, getting a `Error: connect ECONNREFUSED ::1:8025` error. Is there any extra step needed to make it work?

nicktrn:
@leog Mailhog needs to be running for this to work. I've taken a different approach in #10560 where playwright starts a mail server instance directly. Would be worth discussing pros and cons.

leog (approved):
Got it to work using `yarn dx` and since the changes to make it work in GH actions were taken in a different PR already merged, this looks good to me now. Thanks!

apps/web/playwright/fixtures/users.ts
```diff
@@ -330,7 +331,19 @@ export const createUsersFixture = (page: Page, workerInfo: WorkerInfo) => {
...
+        const emailMessages = await emails.search(user.email);
+        if (emailMessages && emailMessages.count > 0) {
+          emailMessages.items.forEach((item) => {
+            emailMessageIds.push(item.ID);
+          });
+        }
+      }
+      for (const id of emailMessageIds) {
+        await emails.deleteMessage(id);
+      }
```
zomars, line 346: @leog @shivamklr I think this is why I get so many errors running e2e locally on production builds.

image

Can we make this optional? Only check for mailhog if it's enabled. I spent quite a while figuring out why all my tests were failing locally. At least we should document and update the README with instructions.

shivamklr: Sure. I will take a look and update the users fixture.

PR #10607, opened 2023-08-04
perf: no wait for session when calling getschedule 10552 cal 2311

### LINKED ISSUE #10552: [CAL-2311] Perf: No wait for session when calling getSchedule
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2311](https://linear.app/calcom/issue/CAL-2311/perf-no-wait-for-session-when-calling-getschedule)</sub>

alannnc:
## What does this PR do?

- Fixes usePublicPage hook.
- Moves data fetching from getSchedule to top of render
- prefetches session
- getSchedule can handle eventSlug only and fetch everything needed there.

Fixes #10552 

## Requirement/Documentation
Before:
image
image

After:
image
image

This is just an example, to be certain we need to measure it many times with many different cpu configurations. Also we can start looking for best Started time for both scenarios.
But from the waterfall graphic we can see that it's now being called in parallel with other networks calls as before was waiting for a lot of things to load before calling trpc.

## Type of change

## Mandatory Tasks

## Checklist


apps/web/lib/app-providers.tsx
```diff
@@ -243,7 +243,11 @@ function OrgBrandProvider({ children }: { children: React.ReactNode }) {
 }
 
 const AppProviders = (props: AppPropsWithChildren) => {
-  const session = trpc.viewer.public.session.useQuery().data;
+  const session = trpc.viewer.public.session.useQuery(undefined, {
+    trpc: {
+      context: { skipBatch: true },
+    },
+  }).data;
```
alannnc, line 250: Since some components waits for session, we shouldn't be batching this one.

apps/web/lib/hooks/usePublicPage.ts
```diff
@@ -1,9 +1,12 @@
-import { usePathname } from "next/navigation";
+import { usePathname, useSearchParams } from "next/navigation";
 
 export default function usePublicPage() {
   const pathname = usePathname();
-  const isPublicPage = ["/[user]", "/booking", "/cancel", "/reschedule"].find((route) =>
-    pathname?.startsWith(route)
-  );
-  return isPublicPage;
+  const isPublicPage = ["/booking", "/cancel", "/reschedule"].some((route) => pathname?.startsWith(route));
```
alannnc, line 5: /[user] url wasn't executing properly, so we now look for searchParams with the new next/navigation

apps/web/server/lib/ssr.ts
```diff
@@ -31,5 +31,7 @@ export async function ssrInit(context: GetServerSidePropsContext) {
   // Provides a better UX to the users who have already upgraded.
   await ssr.viewer.teams.hasTeamPlan.prefetch();
 
+  await ssr.viewer.public.session.prefetch();
```
alannnc, line 34: prefetching session on SSR as suggested by @zomars

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -41,6 +41,17 @@ const BookerComponent = ({
...
+   * Prioritize dateSchedule load
+   * Component will render but use data already fetched from here, and no duplicate requests will be made
+   * */
+  useScheduleForEvent({
+    prefetchNextMonth: false,
+    username,
+    eventSlug,
+    month,
+    duration: undefined,
+  });
```
alannnc, line 54: Calling hook right away in the render.

packages/features/bookings/Booker/utils/event.ts
```diff
@@ -27,7 +29,8 @@ export const useEvent = () => {
 
 /**
  * Gets schedule for the current event and current month.
- * Gets all values from the booker store.
+ * Gets all values right away and not the store because it increases network timing, only for the first render.
+ * We can read from the store if we want to get the latest values.
```
alannnc, line 33: Self explanatory

apps/web/lib/hooks/usePublicPage.ts
hariombalhara: I think now we can move to using `isBookingPage` Component prop(which wasn't there earlier) as that is more reliable 
image

apps/web/lib/hooks/usePublicPage.ts
```diff
@@ -1,9 +1,12 @@
...
+
+  const searchParams = useSearchParams();
+  const userParam = searchParams.get("user");
+  const teamParam = searchParams.get("team");
+
+  return !isPublicPage && (userParam || teamParam);
```
hariombalhara, line 11:
```suggestion
  return isPublicPage || (userParam || teamParam);
```

packages/features/bookings/Booker/Booker.tsx
hariombalhara: It isn't intuitive as we are not using the returned value in this component. Instead, I like this approach of letting the Component render it's child and let the child component fetch `useSchedule` as it's the one who needs it
https://github.com/calcom/cal.com/pull/10588/files#diff-d8cdd443610a4483db45dcc4da6a02bcff302ad6c591aea6fa1ee8646a52aedbL168-L173

packages/features/bookings/Booker/utils/event.ts
```diff
@@ -36,21 +39,37 @@ export const useEvent = () => {
...
+  prefetchNextMonth?: boolean;
+  username?: string | null;
+  eventSlug?: string | null;
+  eventId?: number | null;
+  month?: string | null;
+  duration?: number | null;
+} = {}) => {
+  useTimePreferences();
+
+  const timezone = localStorage.getItem(timezoneLocalStorageKey) || dayjs.tz.guess();
```
hariombalhara, line 59: Why are we not using `useTimePreferences`? It seems to be doing the same thing.

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -29,24 +29,31 @@ export const useSchedule = ({
...
+      },
       refetchOnWindowFocus: false,
       enabled:
         Boolean(username) &&
-        Boolean(eventSlug) &&
-        (Boolean(eventId) || eventId === 0) &&
         Boolean(month) &&
-        Boolean(timezone),
+        Boolean(timezone) &&
+        // Should only wait for one or the other, not both.
```
hariombalhara, line 55: Screenshot 2023-08-07 at 10 01 14 AM
It might be breaking dynamic booking.

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -29,24 +29,31 @@ export const useSchedule = ({
   return trpc.viewer.public.slots.getSchedule.useQuery(
     {
       usernameList: getUsernameList(username ?? ""),
-      eventTypeSlug: eventSlug!,
+      // Prioritize slug over id, since slug is the first value we get available.
```
hariombalhara, line 32: Non dynamic events need eventTypeId only
Screenshot 2023-08-07 at 10 03 48 AM

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -29,24 +29,31 @@ export const useSchedule = ({
...
+      trpc: {
+        context: {
+          skipBatch: true,
+        },
+      },
       refetchOnWindowFocus: false,
       enabled:
         Boolean(username) &&
-        Boolean(eventSlug) &&
-        (Boolean(eventId) || eventId === 0) &&
```
Udit-takkar, line 47: i don't think you should remove eventId from here because it is required on the backend https://github.com/calcom/cal.com/blob/main/packages/trpc/server/routers/viewer/slots/util.ts#L189. only dynamic event types have eventId.

Udit-takkar (commented):
@alannnc it is not session api call but public.event api and https://github.com/calcom/cal.com/pull/10576/files#r1284253215 that was blocking the slots.getSchedule api call.

In this PR the useScheduleForEvent gets the eventTypeId from useEvent() hook so public.event API response would be required to get the eventId before executing slots.getSchedule. I tackled this problem by fetching the eventId in the slots.getSchedule  https://github.com/calcom/cal.com/pull/10576/files#r1285625321

packages/features/bookings/Booker/Booker.tsx
emrysal: Neither way is particularly great tbh

hariombalhara: Yeah I agree.

The problem actually is that `useScheduleForEvent` needs to have `eventTypeId`(and other props) set in zustand booker store. So, we are solving the problem by calling `useScheduleForEvent` here itself and providing it the values directly. If we can solve it in zustand store(bookerStore) directly that would be the best IMHO(but that wasn't possible I tried), but I feel next best thing is that the custom hook `useEvent` gives you the correct data wherever you use it. So, I fixed it in `useEvent` itself. So, once it's fixed, every component and hook would get the correct data.


Another approach might be to pass these props directly to the components needing them instead of them using `useEvent` but that might be a lot of prop drilling which feels unnecessary and counter intuitive to me.

emrysal (dismissed):
Looks good to me - edit: @alannnc tests are failing

hariombalhara (dismissed):
Blocking merge as there are possible functional issues with the code right now in terms of eventTypeId being considered optional

apps/web/lib/hooks/usePublicPage.ts
alannnc: Should we rename? isBookingPage just tells that its a booking page but it will be better if we use isPublicPage? or auth = true?
https://next-auth.js.org/getting-started/client#custom-client-session-handling

packages/features/bookings/Booker/Booker.tsx
alannnc: My guess is this will be solved once we reach react server components. But for now since this is a core component we can use this for now and let the rest keep working as usual.

packages/features/schedules/lib/use-schedule/useSchedule.ts
alannnc: Didn't removed just moved to last conditional

apps/web/lib/hooks/usePublicPage.ts
```diff
@@ -1,9 +1,12 @@
...
-    pathname?.startsWith(route)
-  );
-  return isPublicPage;
+  const isPublicPage = ["/booking", "/cancel", "/reschedule"].some((route) => pathname?.startsWith(route));
+
+  const searchParams = useSearchParams();
+  const userParam = searchParams.get("user");
+  const teamParam = searchParams.get("team");
+
+  return isPublicPage || (userParam || teamParam);
```
alannnc, line 11:
```suggestion
  return isPublicPage || userParam || teamParam;
```

packages/features/schedules/lib/use-schedule/useSchedule.ts
alannnc: Shouldn't a failed test for dynamic pop up here? I think we have that.

alannnc:
> @alannnc it is not session api call but public.event api and https://github.com/calcom/cal.com/pull/10576/files#r1284253215 that was blocking the slots.getSchedule api call.
> 
> In this PR the useScheduleForEvent gets the eventTypeId from useEvent() hook so public.event API response would be required to get the eventId before executing slots.getSchedule. I tackled this problem by fetching the eventId in the slots.getSchedule https://github.com/calcom/cal.com/pull/10576/files#r1285625321

Yeah the thing was not session related but just keep name on the PR to be the same as the issue. Also while digging found some things we can improve related to public pages.

alannnc:
> Blocking merge as there are possible functional issues with the code right now in terms of eventTypeId being considered optional

Is there a dynamic test that takes care of this? IMO i didn't turned to optional if there isn't a slug we wait for eventTypeId, if there is a slug we can fetch omitting eventTypeId since our trpc requires slug or username[] and eventTypeId.

apps/web/lib/hooks/usePublicPage.ts
hariombalhara: The purpose of isPublicPage was to disable intercom on booking pages that are public. So, I guess we would be fine renaming it to useIsBookingPage

alannnc: Agree

packages/features/schedules/lib/use-schedule/useSchedule.ts
alannnc: This still pending as I think test related to dynamic should fail here.

hariombalhara: image

Regular events(Non Dynamic) aren't showing availability if they don't have slug 15min,60min,30min. 
This is due to `getSchedule` not receiving eventTypeId and the above logic considering the request as dynamic booking availability request.

packages/features/schedules/lib/use-schedule/useSchedule.ts
hariombalhara: Dynamic is working fine because it needs `eventSlug` and due to whatever reasons `eventSlug` is there always, but regular bookings availability are broken.
See https://github.com/calcom/cal.com/pull/10607/files#r1286568174

hariombalhara:
Verified that it's breaking regular events booking availability https://github.com/calcom/cal.com/pull/10607/files#r1286568174

packages/features/schedules/lib/use-schedule/useSchedule.ts
alannnc: I couldn't reproduce it. Mine availability is working fine here. Can you give me exact steps to do so?
image

packages/features/schedules/lib/use-schedule/useSchedule.ts
alannnc: Regular event types working fine for me. Also dynamics.

packages/features/schedules/lib/use-schedule/useSchedule.ts
hariombalhara: That is super weird. Giving you a loom

hariombalhara: Here it is https://www.loom.com/share/058304b8913d4a38be70b73c6af88acc

PeerRich:
@alannnc any movement here?

alannnc:
> @alannnc any movement here?

Working on something to solve slug and eventTypeId issue.

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -70,9 +70,31 @@ export const checkIfIsAvailable = ({
...
+      },
+      select: {
+        id: true,
+      },
+    });
+    if (!eventType) {
+      throw new TRPCError({ code: "NOT_FOUND" });
+    }
+    eventTypeId = eventType.id;
+  }
```
alannnc, line 93: We should be better be fetching eventTypeId on the server than waiting in the frontend.

alannnc: Also getSchedules allows two combinations of inputs, slug and usernameList or eventTypeId. But then wasn't using  slug and usernameList combination to fetch the event.

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -175,7 +197,7 @@ export async function getDynamicEventType(input: TGetScheduleInputSchema) {
 }
 
 export function getRegularOrDynamicEventType(input: TGetScheduleInputSchema) {
-  const isDynamicBooking = !input.eventTypeId;
+  const isDynamicBooking = input.usernameList && input.usernameList.length > 1;
```
alannnc, line 200: In the frontend (but SSR) we decide if it's dynamic using different sentence. I think we should use the same.
image

packages/trpc/server/routers/viewer/slots/util.ts
emrysal: > We should be better be fetching eventTypeId on the server than waiting in the frontend.

Working on that 🙏

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -70,9 +70,46 @@ export const checkIfIsAvailable = ({
...
+        id: true,
+      },
+    });
+
+    if (!eventType) {
+      throw new TRPCError({ code: "NOT_FOUND" });
+    }
+
+    eventTypeId = eventType.id;
+  }
```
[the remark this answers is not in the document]
emrysal: @hariombalhara I'm working on deprecating getServerSideProps entirely & also get the eventTypeId serverside.

emrysal (approved):
🙌

packages/trpc/server/routers/viewer/slots/util.ts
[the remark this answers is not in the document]
Udit-takkar: @alannnc can this query handle orgs event types? (we also have some extra logic in getPublicEvent function to handle this)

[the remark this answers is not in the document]
alannnc: @Udit-takkar I fixed so it can handle team/[slug]/[type].tsx and I see [orgSlug]/team/[slug]/[type].tsx reuses that logic. So it should work too and we have some test for that too.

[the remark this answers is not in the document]
hariombalhara: @alannnc It's not handling org event-types. I had two users in my DB with `pro` username(one in an Org so duplicate username is possible) and `getUserIdFromUsername` got me the wrong user and thus even though the booking page loaded, the schedule call gave me 404.

image

image

PR #10610, opened 2023-08-05
perf: Improved buildDateRanges performance + cleanup

emrysal:
## What does this PR do?

* Cleans up workingHours/dateOverrides
* Dayjs isSame is very slow - using valueOf (to get the numeric value) and checking for sameness is much faster.
* Cleans up the getAggregatedAvailability typing.


zomars (approved):
LGTM

PR #10615, opened 2023-08-06
fix: #10614 - made nodemailer compatible with mailhog smtp server.

### LINKED ISSUE #10614: [BUG]: Mailhog server auth issue with nodemailer

### Issue Summary

When you send mail using nodemailer with smtp client mailhog it throws an ERROR: **Missing Credentials for "PLAIN"** 

### Steps to Reproduce

1. Set up project locally as given in Readme section of Repository.
2. Go to localhost:3000.
3. Login with the unverified account.
4. At the top bar Click on verify Email.

### Actual Results

mailhog error

### Expected Results

It should not throw any error.

RohitSaini15:
## What does this PR do?

Allow nodemailer to send message locally to mailhog smtp server.

Fixes #10614

https://github.com/calcom/cal.com/assets/71682713/ea66294f-e19a-4e39-90d9-0b0381071238

### DIFF
--- a/packages/lib/serverConfig.ts
+++ b/packages/lib/serverConfig.ts
@@ -13,10 +13,12 @@ function detectTransport(): SendmailTransport.Options | SMTPConnection.Options |
     const transport = {
       host: process.env.EMAIL_SERVER_HOST,
       port,
-      auth: {
-        user: process.env.EMAIL_SERVER_USER,
-        pass: process.env.EMAIL_SERVER_PASSWORD,
-      },
+      auth: process.env.EMAIL_SERVER_PASSWORD
+        ? {
+            user: process.env.EMAIL_SERVER_USER,
+            pass: process.env.EMAIL_SERVER_PASSWORD,
+          }
+        : undefined,
       secure: port === 465,
       tls: {
         rejectUnauthorized: !isENVDev,
--- a/packages/lib/serverConfig.ts
+++ b/packages/lib/serverConfig.ts
@@ -13,17 +13,18 @@ function detectTransport(): SendmailTransport.Options | SMTPConnection.Options |
     const transport = {
       host: process.env.EMAIL_SERVER_HOST,
       port,
-      auth: process.env.EMAIL_SERVER_PASSWORD
-        ? {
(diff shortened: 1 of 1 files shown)


packages/lib/serverConfig.ts
```diff
@@ -13,10 +13,12 @@ function detectTransport(): SendmailTransport.Options | SMTPConnection.Options |
     const transport = {
       host: process.env.EMAIL_SERVER_HOST,
       port,
-      auth: {
-        user: process.env.EMAIL_SERVER_USER,
-        pass: process.env.EMAIL_SERVER_PASSWORD,
-      },
+      auth: process.env.EMAIL_SERVER_PASSWORD
```
iamr-kumar, line 16: I think a better approach here would be to use `isENVProd` check instead of the server password. Would make the use case more clearer

iamr-kumar, line 16: I think a better approach here would be to use `isENVProd` check instead of the server password. Would make the use case more clearer

RohitSaini15: what if the user tries to use google smtp server on the local machine then this won't work with isENVProd right?

emrysal: I think that is happening here is that mailhog requires no authentication, and setting auth: to `process.env.EMAIL_SERVER_PASSWORD` disables auth implicitly (as this is probably an empty string).

instead of this, I'd suggest removing the "auth" key entirely as per https://stackoverflow.com/a/73117417 when the pass/user are empty

emrysal (changes requested):

shivamklr (changes requested):
@RohitSaini15 The type-check is failing. Kindly fix the error.

shivamklr (commented):
Tested lgtm. Feel free to review the code team.

shivamklr (approved):

emrysal (approved):
Looks good now, thought of an edge case where the user is set but the email server password is an empty string; but that's unlikely to be the case anywhere; we'll cross that bridge when it comes up.

PR #10619, opened 2023-08-06
perf: Remove organisation as it's unused in getBusyTimes (?)

emrysal:
## What does this PR do?

* Removes credentials from the getEventType call (part of other performance refactors)
* Remove organization slug as it was unused in getBusyTimes - but it was prop drilled all the way from getEventType, so I believe it can be removed.


leog (approved):
Good that we are removing deprecated code if not needed by new ways of caching calendar. LGTM

PR #10638, opened 2023-08-07
perf: Correct use of next-auth session

sean-brydon:
Internal discussion about this: https://threads.com/34523046158?s=oQMEHAJUVdjCkudo4EhB8h


zomars (approved):
Ship it

shivamklr (commented):
lgtm

PR #10646, opened 2023-08-08
perf: Avoid unmounting of Shell on navigation and thus reduce number of paints 

hariombalhara:
## What does this PR do?

- Avoid state reset for Shell Sidebar on navigation. See [comparison b/w production and this PR
](https://www.loom.com/share/6b6b58ac11d24942b9a725345e8c4fd5)
- It automatically improves App experience as well as Avatar and Tips aren't recreated which user could see clearly earlier.
- Also adds @calcom/debugging package that can be used for various debugging utilities. Right now it has `RenderCounter` component. `useTraceUpdate` should also be moved here.

Followup:
- Handle Routing Forms Navigation as well without state reset
- Handle sub navigations like /event-types/[id]

## Type of change

## How should this be tested?
- Browse through all the main navigations and see that they are working

## Mandatory Tasks

## Checklist


apps/web/components/apps/layouts/AppsLayout.tsx
```diff
@@ -22,7 +22,7 @@ export default function AppsLayout({ children, actions, emptyStore, ...rest }: A
 
   if (session.status === "loading") return <></>;
   return (
-    <Shell {...rest} actions={actions?.("block")} hideHeadingOnMobile>
+    <ShellMain {...rest} actions={actions?.("block")} hideHeadingOnMobile>
```
hariombalhara, line 25: Have been using this strategy for Routing Forms navigations. Using `ShellMain` here allows configuring shell on per route basis. `MainLayout` renders `Shell` without main content

packages/features/bookings/layout/BookingLayout.tsx
hariombalhara, line 1: It was imported in just one route. So, I moved it there itself.

apps/web/pages/event-types/index.tsx
```diff
@@ -932,25 +933,25 @@ const EventTypesPage = () => {
...
         description="Create events to share for people to book on your calendar."
       />
-      <Shell
-        withoutSeo
-        heading={t("event_types_page_title")}
-        hideHeadingOnMobile
-        subtitle={t("event_types_page_subtitle")}
-        afterHeading={showProfileBanner && <SetupProfileBanner closeAction={closeBanner} />}
-        beforeCTAactions={<Actions />}
-        CTA={<CTA data={data} />}>
```
hariombalhara, line 947: These properties being specific to this route are now moved to `ShellMain`

apps/web/lib/hooks/usePublicPage.ts
```diff
@@ -2,7 +2,7 @@ import { usePathname } from "next/navigation";
 
 export default function usePublicPage() {
   const pathname = usePathname();
-  const isPublicPage = ["/[user]", "/booking", "/cancel", "/reschedule"].find((route) =>
+  const isPublicPage = ["/[user]", "/booking/", "/cancel", "/reschedule"].find((route) =>
```
hariombalhara, line 5: It was causing bookings/upcoming page to be considered public page and thus was removing Intercom Provider from there which was causing the DOM to be reconstructed for entire page.

zomars (changes requested):
Locally works as expected. nice job @hariombalhara Can we check why e2e is failing?

zomars (approved):
Tests are passing now!

PR #10651, opened 2023-08-08
fix: UI mismatch with form state

### LINKED ISSUE #10390: [CAL-2261] If the input type "Name" is selected, the label can't be changed from our default label "Your Name"
[https://app.birdeatsbug.com/sessions/-ZDXRm7sd1JSyC-NEoXm8ZHSyTbWMWezvkAllOzlsN2Z](https://app.birdeatsbug.com/sessions/-ZDXRm7sd1JSyC-NEoXm8ZHSyTbWMWezvkAllOzlsN2Z)
Created via Threads. See full discussion: [https://threads.com/34513965502](https://threads.com/34513965502)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2261](https://linear.app/calcom/issue/CAL-2261/if-the-input-type-name-is-selected-the-label-cant-be-changed-from-our)</sub>

### LINKED ISSUE #10647: [CAL-2516] Form state changes doesn't lead to change in UI for some fields.
### Issue Summary

When making some change in the eventType form. The form changes are not reflected in the UI for some fields. They are registerd in the form state but for the UI rendering logic they are dependant on the data fetched from backend. 

This can lead to bad UX. The customer can not be sure what is the current form state due to UI mismatch.
The customer may change a field (eg. enabling recurring event), switches the tabs and decides to disables the recurring event. When he switches back to the recurring event tab. He might assume that his changes did not persisted and the recurring event is disabled ( which infact is enabled in the form state ) and leads to unexpected behaivour for the customer.

### Steps to Reproduce

1. Edit a event type(meeting).
2. Make some changes in a tab and switch between tabs back and forth. ( for eg. lets change the recurring event setting)
3. The change will be registed in the form state but will not be reflected in the UI.

### Actual Results

https://github.com/calcom/cal.com/assets/86384652/6d2591bc-b2e0-4790-916a-512c63dd7431

### Expected Results

https://github.com/calcom/cal.com/assets/86384652/f0629779-1ae5-4b69-9b5c-879dfae832cb

### Technical details

- Browser version: Chrome 115
- Node.js version: 18.16.1

<sub>[CAL-2516](https://linear.app/calcom/issue/CAL-2516/form-state-changes-doesnt-lead-to-change-in-ui-for-some-fields)</sub>

MehulZR:
## What does this PR do?

Fixes #10647  (issue)
~~Fixes #10390  (issue)~~

https://github.com/calcom/cal.com/assets/86384652/677e684f-b73b-4da9-a6dd-750ed8d85577

## Type of change

## Mandatory Tasks


apps/web/public/static/locales/en/common.json
```diff
@@ -33,7 +33,6 @@
   "event_request_cancelled": "Your scheduled event was canceled",
   "organizer": "Organizer",
   "need_to_reschedule_or_cancel": "Need to reschedule or cancel?",
-  "no_options_available": "No options available",
```
MehulZR, line 36: duplicate keys
- no_options_available
- people

apps/web/public/static/locales/en/common.json
```diff
@@ -1985,5 +1983,7 @@
   "attendee_last_name_variable": "Attendee last name",
   "attendee_first_name_info": "The person booking's first name",
   "attendee_last_name_info": "The person booking's last name",
-  "ADD_NEW_STRINGS_ABOVE_THIS_LINE_TO_PREVENT_MERGE_CONFLICTS": "↑↑↑↑↑↑↑↑↑↑↑↑↑ Add your new strings above here ↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑"
+  "ADD_NEW_STRINGS_ABOVE_THIS_LINE_TO_PREVENT_MERGE_CONFLICTS": "↑↑↑↑↑↑↑↑↑↑↑↑↑ Add your new strings above here ↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑",
+  "example_first_name":"John",
```
MehulZR, line 1987: Will need translation for other languages.

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -551,7 +551,19 @@ function FieldLabel({ field }: { field: RhfFormField }) {
...
+  let label = "";
+  if (variant === "firstAndLastName") {
+    label =
+      variantsConfig.variants[variant].fields[0]?.label?.length &&
+      variantsConfig.variants[variant].fields[1]?.label?.length
+        ? `${variantsConfig.variants[variant].fields[0].label}, ${variantsConfig.variants[variant].fields[1].label}`
+        : t(fieldTypeConfigVariants[variant as keyof typeof fieldTypeConfigVariants].label);
+  } else {
+    label = variantsConfig.variants[variant].fields[0]?.label?.length
+      ? variantsConfig.variants[variant].fields[0].label
```
MehulZR, line 563: Leads to type issue.

hariombalhara:
@MehulZR Could you please remove the changes related to #10390 from this PR as that's already solved(and being solved) elsewhere already.

Also, you should open separate PRs for separate fixes unless they are really really related. It makes it easy to review and merge the PR

Udit-takkar (commented):
@MehulZR please remove conflicts and changes related to https://github.com/calcom/cal.com/issues/10390

MehulZR:
While updating the PR, I noticed that it is possible to book a event with *length* of 0. This is possible when saving the form from a tab other than the *Event Setup*.

image


We can solve it in two ways.
1. Either not letting the form submit via RHF.
2. On Submit, checking the value and throwing an Error like we do when there are no multiple durations selected.

> I prefer the 2nd approach because If we use the 1st approach The user will not get any indication of an error on tabs other than *Event Setup*. And it also looks more consistent.

If you would like the 2nd approach too. What Error message do you want to be shown?

image

packages/features/form-builder/FormBuilder.tsx
```diff
@@ -551,7 +551,19 @@ function FieldLabel({ field }: { field: RhfFormField }) {
       "Field has `variantsConfig` but no `defaultVariant`" + JSON.stringify(fieldTypeConfigVariantsConfig)
```
hariombalhara, line 551: @MehulZR I still see some unrelated changes here in this file related to #10390. Let's remove those first as I mentioned earlier.

hariombalhara (changes requested):
Let's address merge conflicts and unrelated changes first.

Udit-takkar (commented):
@MehulZR do not push yarn.lock unless you change package.json.

Use this command to undo the changes in yarn.lock

Screenshot 2023-09-04 at 5 13 45 PM

Udit-takkar:
@MehulZR can you fix the conflicts in this PR?

Udit-takkar:
@MehulZR there are still some conflicts

apps/web/components/eventtype/EventTypeSingleLayout.tsx
```diff
@@ -171,11 +181,17 @@ function EventTypeSingleLayout({
     t("locked_fields_member_description")
   );
 
+  const enabledWebhooks = formMethods.watch("enabledWebhooks");
+  const length = formMethods.watch("length");
+  const multipleDuration = formMethods.watch("metadata")?.multipleDuration;
+
```
hariombalhara, line 187: Instead of maintaining `enabledWebhooks` separately, we can simply invalidate eventType tRPC query like this [here](https://github.com/calcom/cal.com/blob/f87eac193fb0a6b244489a61b97e5f65b73e64b9/packages/features/webhooks/components/WebhookListItem.tsx#L53) 

```javascript
if (webhook.eventTypeId) {
    await utils.viewer.eventTypes.get.invalidate({ id: webhook.eventTypeId });
  }
```

apps/web/components/eventtype/EventWebhooksTab.tsx
```diff
@@ -23,6 +24,11 @@ export const EventWebhooksTab = ({ eventType }: Pick<EventTypeSetupProps, "event
 
   const { data: webhooks } = trpc.viewer.webhook.list.useQuery({ eventTypeId: eventType.id });
 
+  const enabledWebhooks = webhooks?.filter((webhook) => webhook.active).length;
+  const formMethods = useFormContext<FormValues>();
+  useEffect(() => {
+    if (enabledWebhooks !== undefined) formMethods.setValue("enabledWebhooks", enabledWebhooks);
+  }, [enabledWebhooks]);
```
hariombalhara, line 31: As mentioned earlier, it won't be needed if we just do the invalidation

hariombalhara (dismissed):
@MehulZR I resolved the conflicts and left some more comments.

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -203,7 +205,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps,
       <RequiresConfirmationController
         eventType={eventType}
         seatsEnabled={seatsEnabled}
-        metadata={eventType.metadata}
+        metadata={formMethods.getValues("metadata")}
```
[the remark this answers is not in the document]
MehulZR: There are some cases when we need the original `eventType` prop alongside `formMethods` values.
For ex. having both the count of webhooks and other form values in the same component.

what we can do is instead of calling `formMethods.getValues("some_property_name")` everywhere, we just do `formValues = formMethods.getValues()` saving the execution calls and cleaning up things. 
but formMethods.getValues isn't reactive (value change doesn't cause re-render), so for some cases we will still be using formMethods.watch().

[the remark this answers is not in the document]
MehulZR: I have tested a bit with using the `formValues` strategy, as expected it is not reliable. We should use `formMethods.getValues("some_prop")`

joeauyeung:
@MehulZR just a heads up there's more conflicts

joeauyeung (dismissed):
Double checked everything and tested some fields. I fixed the `noShowFeeEnabled` since it was returning true even if Stripe was disabled. 

I did notice one thing that's consistent in `main` as well but for managed event types when changing from a set availability and back to "member's default availability" the option doesn't change. 
CleanShot 2024-01-23 at 12 22 09


I'm ok approving this PR since that bug wasn't caused by the changes made and it's such a huge QOL improvement.

emrysal (approved):
Fixed merge conflicts, approved

PR #10657, opened 2023-08-09
perf: Replace un-needed context fetching 

sean-brydon:
This PR replaces some fields that were always fetched in trpc session contect to only why they are needed. 

These often made this query exponentially slower whenever there was values in 
* credentials 
* avatar


packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -122,7 +108,6 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
       metadata: orgMetadata,
     },
     id,
-    rawAvatar,
```
emrysal, line 125: rawAvatar has a bit more dependencies in the code - see L40 😱

emrysal (changes requested):

apps/web/pages/event-types/index.tsx
```diff
@@ -715,10 +715,7 @@ const EventTypeListHeading = ({
       <Avatar
         alt={profile?.name || ""}
         href={teamId ? `/settings/teams/${teamId}/profile` : "/settings/my-account/profile"}
-        imageSrc={
-          `${orgBranding?.fullDomain ?? WEBAPP_URL}/${teamId ? "team/" : ""}${profile.slug}/avatar.png` ||
-          undefined
-        }
+        imageSrc={`${orgBranding?.fullDomain ?? WEBAPP_URL}/${profile.slug}/avatar.png` || undefined}
```
emrysal, line 718: Fixed a 404, team avatar here was generated as /team/team/seeded-team/avatar.png

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
```diff
@@ -41,7 +41,23 @@ export const createHandler = async ({ ctx, input }: CreateOptions) => {
...
+        key: true,
+        userId: true,
+        appId: true,
+        invalid: true,
+        teamId: true,
+      },
+      orderBy: {
+        id: "asc",
+      },
+    });
```
zomars, line 60: Would love to abstract these repeated calls. Not a blocker tho.

zomars (approved):
Nice improvements @sean-brydon :pray:

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
sean-brydon: I can do this :)

zomars (dismissed):
Checking locally and it seems like we're missing credentials in quite a lot of places. Will block to check. I'll try to see if we can create a "authedProcedureWithCredentials" to use it where needed.

emrysal (approved):
Should be good to go if tests pass 🙏

PR #10660, opened 2023-08-09
feat: add support for apple travel time

### LINKED ISSUE #10474: Support Apple's Travel Time

Apple calendar has a feature called "Travel Time" that is supposed to block out travel time needed to get to an event location. Cal.com doesn't seem to support this, as it still lets people book on top of my "travel time".

Cal.com should handle the "travel time" and consider it to be blocked off.

It's possible to just manually schedule a block with an event for travel time, but that's clutters the calendar with events that aren't really events - and is why Apple created the Travel Time feature.

### Additional context

The data seems to be already available from CalDAV as someone posted here:
https://community.home-assistant.io/t/caldav-add-support-for-travel-time/525502

### Requirement/Document

N/A

nicktrn:
## What does this PR do?

Adds support for Apple's [Travel Time](https://support.apple.com/en-gb/guide/calendar/icl43600/mac) feature

See [JSFiddle](https://jsfiddle.net/kv1czg05/) for example payload and parsing.

Fixes #10474
/claim #10474 

## Requirement/Documentation

- https://community.home-assistant.io/t/caldav-add-support-for-travel-time/525502
- https://github.com/nextcloud/calendar/issues/581

## Type of change

## How should this be tested?

1. Add travel time via Apple Calendar
2. Check if it's now unavailable to book via Cal

## Mandatory Tasks

## Checklist


alishaz-polymath:
Hey @nicktrn Thank you for the PR. 
Please make sure to ready it for review once you're happy with the changes. Please ensure you've gone through [Contributor's Guide to PR](https://cal.com/docs/knowledgebase/contributors-guide/pull-requests) and ensure self-review is done when you push it to be reviewed 🙏

nicktrn:
Hi @alishaz-polymath Thanks for having a look this quickly, but this is only a draft for now. As you know there are still things to discuss. By the way, ical.js is a pain to work with.

alishaz-polymath:
> Hi @alishaz-polymath Thanks for having a look this quickly, but this is only a draft for now. As you know there are still things to discuss. By the way, ical.js is a pain to work with.

Oh for sure. And yes, iCal.js is painful 😅

packages/lib/CalendarService.ts
```diff
@@ -60,6 +60,30 @@ function getFileExtension(url: string): string {
...
+    if (!Number.isInteger(travelSeconds)) return 0;
+    return travelSeconds;
+  } catch (e) {
+    log.error("[ERROR] invalid travelDuration?", e);
+    return 0;
+  }
+};
+
+const applyTravelDuration = (event: ICAL.Event, seconds: number) => {
+  if (seconds <= 0) return event;
```
nicktrn, line 81: Negative travel times should be impossible, but better to be safe than sorry

packages/lib/CalendarService.ts
```diff
@@ -60,6 +60,30 @@ function getFileExtension(url: string): string {
   return fileName.substring(fileName.lastIndexOf(".") + 1);
 }
 
+// for Apple's Travel Time feature only (for now)
+const getTravelDurationInSeconds = (vevent: ICAL.Component, log: typeof logger) => {
+  const travelDuration: ICAL.Duration = vevent.getFirstPropertyValue("x-apple-travel-duration");
```
nicktrn, line 65: I did consider checking if this is indeed an ICAL.Duration, but their helpers are incredibly useless here. Will let any string with `P` as first or second character pass. See:

https://github.com/kewisch/ical.js/blob/35533497954b2c6e20a902789fe11e95740c3cf6/lib/ical/duration.js#L36-L38

packages/lib/CalendarService.ts
```diff
@@ -60,6 +60,30 @@ function getFileExtension(url: string): string {
...
+  if (!travelDuration) return 0;
+
+  // we can't rely on this being a valid duration and it's painful to check, so just try and catch if anything throws
+  try {
+    const travelSeconds = travelDuration.toSeconds();
+    // integer validation as we can never be sure with ical.js
+    if (!Number.isInteger(travelSeconds)) return 0;
```
nicktrn, line 72: This is also rather incredible. There are no type checks in the function and we could be getting a very long random string back instead. I think `isInteger()` is the best bet here.

packages/lib/CalendarService.ts
```diff
@@ -60,6 +60,30 @@ function getFileExtension(url: string): string {
...
+    return travelSeconds;
+  } catch (e) {
+    log.error("[ERROR] invalid travelDuration?", e);
+    return 0;
```
nicktrn, line 76: Just in case anything fails. Will actually change the error message quickly, only just realised I replaced it with the proper logger in the end.

packages/types/ical.d.ts
```diff
@@ -168,6 +168,7 @@ declare module "ical.js" {
     public isNegative: boolean;
     public icalclass: string;
     public icaltype: string;
+    public toSeconds(): number;
```
nicktrn, line 171: I did consider using `number | string` or even `any` type here to be accurate, but the rest of the ambient def doesn't do this so I chose not to.

nicktrn (commented):
Ready for review by a Certified iCal.js Expert :tm:

alishaz-polymath:
I'm wondering if it would be better to have travel time as separate events attached to the primary event, or extend the duration of the event instead (as done in this case) 🤔 
My thoughts:

1. If we extend the duration of the same event, it obviously looks cleaner on the calendar, however, we lose vital information about when an event is supposed to start versus when it shows as blocked. Is there a way we can emphasise/identify right off the first look what the actual event start time is (excluding the adjusted time, that is)?
2. We could think about linking pre and post meeting travel time as separate events, keeping them linked to the main event, such that if and when the primary event is removed/rescheduled, the necessary changes take place in the travel time events as well. Would this be a better approach? Are there any pitfalls here?
cc: @joeauyeung @PeerRich @nicktrn

nicktrn:
@alishaz-polymath Thanks for reviewing!

I think mutating the ICAL.Event startDate directly is safe and the simplest solution in this case, as we are just checking availability and returning `EventBusyDate[]`, i.e. an array of start/end dates without additional info.

The data is only used for those availability checks and discarded when the function returns, so this should not have any unintended side-effects.

In summary, this will **only** affect availability of users with linked Apple Calendars who set travel times for specific events.

Apologies if my comments in the related issue confused matters.

This is what it looks like in Calendar on Mac, and this is exactly what we will be blocking:

787430E6-59F1-4263-B4D9-3F95B7ABDD90
_(borrowed from the linked Nextcloud Calendar issue)_

zomars (approved):
Simple. Non intrusive. And it seems to work. Will test before merging. Thank you for your contribution 🙏 @nicktrn

PR #10661, opened 2023-08-09
fix: Timing  format related bugs fixed and booking page dates made consistent for recurring events

Chiranjeev-droid:
## What does this PR do?
When we are taking availability as input from the user , we are taking as 00:00 format but while showing/ rendering at some places, we are showing as 24 which looks too odd. In some places, it is showing 00 while at some places it is showing time as 24 .This pr resolves the issue and now user will be seeing only 00:min and not 24:min . 

Also there is a issue on booking page which we get after booking is confirmed/rescheduled or cancelled. When we go for recurring events more than 4 , lets we have chosen for 12 recuring, then it will show the rest of 8 as last recurring day which is wrong.

Fixes / Changes Made:-
1. Changed the time format from 24 to 00 and also the recurring dates inconsistency removed.
Untitled (3)
After making changes:- 
Untitled (5)
2. Availability Page before making changes
Untitled (6)
After making changes
Untitled (7)

3. go to event types ⇒ select an event ⇒ click on edit ⇒  click on availability →
 now one will see timing like this 24 hr while we are setting availability we are setting as 00:00
Untitled (8)

after making changes or after editing EventAvailabilityTab which is present in eventtype, we get
Untitled (9)

Before making changes in [schedule.tsx] inside web pages availability,  we see

Untitled (10)

after making changes
Untitled (11)

before making changes in  packages/features/schedules/components/ScheduleListItem.tsx we see 
Untitled (12)

after making changes
Untitled (13)

before editing  /packages/features/schedules/components/DateOverrideList.tsx.   (date overrides ) 
Untitled (14)

after making changes
Untitled (15)
 
before applying changes , if meeting is scheduled/ rescheduled/ cancelled
Untitled (16)

after applying changes -
Screenshot 2023-08-09 at 1 25 21 PM

## Type of change

## How should this be tested?
- create a recurring event  with more than 4 occurences , after successfully registering , expand the event day details and one will see the inconsistency. Also if we have taken time as 00:followed_by_mins then we will be seeing as 24:followed_by_mins.
- similarly while booking event , when we select slot we will select slot like 00 in case of 24 hr format but after selcting slot and before confirming ,we will see the slot at 24 which is not expected.This happens only in case of non recurring events.
- also while creating a schedule for a event type or adding overrides for a particular day, we will be entering as 00 but after adding,it will be shown to user as 24 instead of 00.


Chiranjeev-droid:
@emrysal can you please review the changes🥹

apps/web/components/eventtype/EventAvailabilityTab.tsx
```diff
@@ -118,10 +118,12 @@ const EventTypeScheduleDetails = memo(
                     {filterDays(index).map((dayRange, i) => (
                       <div key={i} className="text-default flex items-center leading-4">
                         <span className="w-16 sm:w-28 sm:text-left">
-                          {format(dayRange.startTime, timeFormat === 12)}
+                          {format(dayRange.startTime, timeFormat === 12).replace(/24(?=:)/g, "00")}
```
CarinaWolli, line 121: let's make this directly in the format function (https://github.com/calcom/cal.com/pull/10661/files#diff-99c36cb0362623749ba84d29249cbb05823ac6fb0d8dad23efa322e063121accR67)

apps/web/pages/availability/[schedule].tsx
```diff
@@ -171,7 +171,10 @@ export default function Availability() {
             .filter((availability) => !!availability.days.length)
             .map((availability) => (
               <span key={availability.id}>
-                {availabilityAsString(availability, { locale: i18n.language, hour12: timeFormat === 12 })}
+                {availabilityAsString(availability, {
```
Comment (CarinaWolli) [inline apps/web/pages/availability/[schedule].tsx:174]: here too, let's make the changes in `availabilityAsString` directly

apps/web/pages/booking/[uid].tsx
```diff
@@ -793,8 +793,18 @@ export function RecurringBookings({
             <div key={idx} className={classNames("mb-2", isCancelled ? "line-through" : "")}>
               {formatToLocalizedDate(dayjs.tz(dateStr, tz), language, "full", tz)}
               <br />
-              {formatToLocalizedTime(dayjs(dateStr), language, undefined, !is24h, tz)} -{" "}
-              {formatToLocalizedTime(dayjs(dateStr).add(duration, "m"), language, undefined, !is24h, tz)}{" "}
+              {formatToLocalizedTime(dayjs(dateStr), language, undefined, !is24h, tz).replace(
```
Comment (CarinaWolli) [inline apps/web/pages/booking/[uid].tsx:796]: let's make changes in `formatToLocalizedTime` directly

CarinaWolli (changes requested):
Looks good overall, but I would suggest making the changes directly in the functions itself

PR #10681, opened 2023-08-09
fix: Allows SAML audience to be configured

deepakprabhakara:
## What does this PR do?

Added env var SAML_AUDIENCE. This is needed to allow Cal SSO to operate in different environments.

## Type of change

## How should this be tested?

This should be tested like other SSO features. First set the env var SAML_AUDIENCE to a value other than the default `https://saml.cal.com` and then configure a SAML app with the new SAML audience value. Navigate to the SSO settings and configure the SAML metadata and test the SSO login flow. Login should succeed and the new audience should be verified.

## Mandatory Tasks


PeerRich:
> @zomars Who can help review this small PR to make the SAML audience configurable?

@deepakprabhakara what is needed here? UI? if yes, where?

pmibal-hf:
@PeerRich @deepakprabhakara Any chance this will make it in a release?

PR #10696, opened 2023-08-10
perf: getSchedule - Running queries in parallel

keithwillcode:
## What does this PR do?

This breaks the queries needed to fulfill `getEventType` into separate async calls not relying on Prisma when calling `getSchedule`. On the dev DB, this is proving to be a consistent improvement of around 300ms.

## Type of change

## How should this be tested?

- What are the minimal test data to have? **basic event types, dynamic event types, team event types**
- What is expected (happy path) to have (input and output)? **The getSchedule call functions exactly the same as before**

## Mandatory Tasks


packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -89,63 +93,130 @@ export async function getEventType(input: TGetScheduleInputSchema) {
       }
     }
 
-    const eventType = await prisma.eventType.findFirst({
+    const startFirstPrismaEventTypeGet = performance.now();
+    eventType = await prisma.eventType.findFirst({
```
emrysal, line 97: I'd start off by preferring the eventTypeId - also if the eventTypeId exists you can probably safely to an findUniqueOrThrow to bail early, then only if there is no event type yet in the follow up try by slug+userId/teamId.

keithwillcode: So the problem is this code gets reached first because the eventTypeId doesn't exist so you have to find it searching by slug, teamId and userId. Are you thinking of a different workaround?

keithwillcode:
Closing this as we've determined with the prisma client running more queries like this in actually slower (verified using tracing)

PR #10697, opened 2023-08-10
fix: reschedule timezone

Udit-takkar:
https://threads.com/34524526047?s=iBF5JA9HyxtgCPWMv4fuRW

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -643,7 +643,6 @@ async function handler(
     notes: additionalNotes,
     smsReminderNumber,
     rescheduleReason,
-    bookingAttendees,
     ...reqBody
   } = getBookingData({
     req,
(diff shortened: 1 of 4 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -643,6 +643,7 @@ async function handler(
     notes: additionalNotes,
     smsReminderNumber,
     rescheduleReason,
+    bookingAttendees,
```
CarinaWolli, line 646: Can't we just get the attendees from the `orginalRescheduleBooking`?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -848,6 +847,46 @@ async function handler(
...
+      !!eventType.seatsPerTimeSlot
+    );
+    if (!originalRescheduledBooking) {
+      throw new HttpError({ statusCode: 404, message: "Could not find original booking" });
+    }
+  }
+
+  const isOrganizerRescheduling = organizerUser.id === userId;
+
+  const attendeeInfoOnReschedule =
```
CarinaWolli, line 880: only if the organizer reschedules we want the orignal tz and language. Attendee might be in different tz now and that's why he rescheduled with different tz

CarinaWolli: @emrysal what do you think about that? 

@Udit-takkar's thought: i thought of this edge but if organiser clicks on reschedule from email or just open the reschedule on another device where he is not logged in then we won't be able to know if it is organiser or attendee who is rescheduling

emrysal (approved):
lgtm 🚀

shivamklr (commented):
Found a minor [issue](https://github.com/calcom/cal.com/issues/10735).

PR #10698, opened 2023-08-10
test: Create unit tests for react components is packages/ui/components/navigation

gitstart-calcom:
## Testing
```jsx
yarn test navigation
```

image


packages/ui/components/navigation/tabs/navigation.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+      render(<HorizontalTabs tabs={mockTabs} />);
+      mockTabs.forEach((tab) => {
+        const tabLabelElement = screen.getByTestId(`horizontal-tab-${tab.name}`);
+        expect(tabLabelElement).toBeInTheDocument();
+
+        const name = screen.getByText(tab.name);
+        expect(name).toBeInTheDocument();
+        expect(tabLabelElement).toHaveAttribute("href", tab.href);
+
+        const spanSibling = name.querySelector("span");
```
hariombalhara, line 53: `span` seems to be implementation detail. User doesn't see a span, he would see the content in it.
We should check what that span shows to user.

packages/ui/components/navigation/tabs/navigation.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+        const tabName = screen.getByText(tab.name);
+        expect(tabName).toBeInTheDocument();
+
+        const aTag = screen.getByTestId(`vertical-tab-${tab.name}`);
+        const tabContainer = tabName.closest("a");
+        const infoElement = tabContainer?.querySelector("p[title='info']");
+
+        expect(chevronRight.length).toEqual(mockTabs.length - 1);
+        if (tab.disabled) {
+          expect(aTag).toHaveClass("pointer-events-none !opacity-30");
```
hariombalhara, line 121: nit: I would prefer these to be abstracted. Like assertTabIsDisabled()

packages/ui/components/navigation/tabs/navigation.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+          expect(aTag).toHaveAttribute("target", "_self");
+        }
+      });
+
+      expect(externalLink.length).toEqual(1);
+      expect(iconElement.length).toEqual(1);
+
+      const svg = document.querySelectorAll("svg");
+      const svgLength = svg?.length;
+      expect(svgLength).toEqual(mockTabs.length + 1);
```
hariombalhara, line 144: Why would there be 1 svg more than tabs ? This isn't clear here.
From a user perspective I would see an icon along with every item. So, we should check the presence per item and not in total.

Also one additional SVG is probably for chevron which should be checked for it's presence in appropriate place.

packages/ui/components/navigation/tabs/navigation.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+          expect(aTag).toHaveAttribute("target", "_blank");
+        } else {
+          expect(aTag).toHaveAttribute("target", "_self");
+        }
+      });
+
+      expect(externalLink.length).toEqual(1);
+      expect(iconElement.length).toEqual(1);
+
+      const svg = document.querySelectorAll("svg");
```
hariombalhara, line 142: Is it right to use document here? Shouldn't we use screen?

hariombalhara (changes requested):
Thanks @gitstart-calcom. Left some comments.

packages/ui/components/navigation/tabs/navigation.test.tsx
gitstart-calcom: Hi @hariombalhara That span tag doesn't have any content, please take a look here:
image

gitstart-calcom:
Hi @hariombalhara we made the requested changes, please take a look 😄

packages/ui/components/navigation/tabs/navigation.test.tsx
hariombalhara: @gitstart-calcom Then why are we checking for `span` tag. What I meant was that we should test for what a user sees. If he can't see the span(it has no content), then why are we testing its presence

hariombalhara (changes requested):
Left a reply.

packages/ui/components/navigation/tabs/navigation.test.tsx
gitstart-calcom: Makes sense, we have updated the test now

hariombalhara (approved):

Issue #9669, opened 2023-06-20
[CAL-2007] chore: Use computed fields to display avatars/logos


zomars:
### Is your proposal related to a problem?

There have been several occasions where endpoints have timed-out due to downloading large payloads for user avatars and team logos. 

Right now it is too easy to include a `include: { user: true }` which reintroduces the problem all over again.

### Describe the solution you'd like

Instead of manually patching every single query that comes up. We can instead replace the `avatar/logo` fields with a virtual one that just returns the formed URL to the avatar endpoint.

- Rename `avatar` column to `avatarRaw`
- Add a computed field called `avatar` which returns a formed URL: `${CAL_URL}/${user.username}/avatar.png`

<sub>[CAL-2007](https://linear.app/calcom/issue/CAL-2007/chore-use-computed-fields-to-display-avatarslogos)</sub>

zomars:
Mostly fixed with the new @emrysal avatar implementation.

PR #10700, opened 2023-08-10
feat: User Avatar and Org Logo

joeauyeung:
## What does this PR do?

This PR adds the organization logo to the user user avatars where the "lg" size was used.

-User onboarding step
CleanShot 2023-08-10 at 20 55 43@2x

- User public page
CleanShot 2023-08-10 at 20 44 40@2x

- User profile settings
CleanShot 2023-08-10 at 20 44 01@2x

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create an org and add a member
- Go to that memeber's public page. The organization logo should be displayed with the avatar
- Go to the member's profile setting's page. The organization logo should be displayed with the avatar
- Go to those two pages with a normal user. The avatar should render itself

## Mandatory Tasks

## Checklist


packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -62,6 +62,8 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
           id: true,
           slug: true,
           metadata: true,
+          name: true,
+          logo: true,
```
Udit-takkar, line 66: is it fine to fetch logo here? https://github.com/calcom/cal.com/issues/9669

hariombalhara: Nice catch !! @Udit-takkar 
We should support an org avatar endpoint and then we don't need to return the logo from here,

hariombalhara: https://app.gitbook.com/o/6snd8PyPYMhg0wUw6CeQ/s/VXRprBTuMlihk37NQgUU/~/changes/101/engineering/performance-best-practices#say-no-to-big-payloads-in-endpoints
Added in the Performance best practices Change Request on gitbook

We should probably have a way to detect and prevent that. Maybe we can have a middleware that runs on dev and that can detect data URI in a property and throw error or warning. 
Making it dev only would ensure no performance impact in production due to middleware.

apps/web/pages/[user].tsx
```diff
@@ -92,7 +92,13 @@ export function UserPage(props: InferGetServerSidePropsType<typeof getServerSide
...
           )}>
           <div className="mb-8 text-center">
-            <Avatar imageSrc={profile.image} size="xl" alt={profile.name} />
+            <Avatar
+              imageSrc={profile.image}
+              size="xl"
+              alt={profile.name}
+              organizationLogo={profile.organizationLogo}
+              organizationName={profile.organizationName}
+            />
```
Comment (hariombalhara) [inline apps/web/pages/[user].tsx:101]: How about creating an HOC <BigAvatar> that can make `organizationLogo` and `organizationName` required props. Right now these are optional props because other sizes avatars don't need org logo.

We can pass explicit null when user isn't org member. This would help in ensuring that we never miss org logo with big avatars.

hariombalhara (dismissed):
Left some comments. Majorly blocking is @Udit-takkar's comment.

apps/web/pages/[user].tsx
Comment (sean-brydon) [reply apps/web/pages/[user].tsx:101]: I envision this as 
image
Very common for avatars to have a so called indicator e.g we use it for online/away status in user dropdown.

Or something like how chakra-ui does this
CleanShot 2023-08-11 at 15 35 56

Comment (leog) [reply apps/web/pages/[user].tsx:101]: I was going to suggest the same thing as Sean, we should probably let Avatar component put any kind of bottom badge so we can cover online status and org logo. I think making it configurable from the outside should be enough, so you can pass the avatar either the org logo or the online status as an additional property and if that's present, put it as is in the badge area.

Comment (joeauyeung) [reply apps/web/pages/[user].tsx:106]: Stacked PR addressing this feedback
https://github.com/calcom/cal.com/pull/10788

apps/web/pages/[user].tsx
```diff
@@ -322,9 +331,15 @@ export const getServerSideProps: GetServerSideProps<UserPageProps> = async (cont
...
     brandColor: user.brandColor,
     darkBrandColor: user.darkBrandColor,
+    organizationName: user.organization?.name,
+    organizationLogo: user.organization?.logo,
     allowSEOIndexing: user.allowSEOIndexing ?? true,
   };
 
+  if (user.organization) {
+    profile.organizationLogo = `/org/${user.organization.slug}/avatar.png`;
+  }
```
Comment (joeauyeung) [inline apps/web/pages/[user].tsx:339]: @hariombalhara @Udit-takkar using this new API endpoint now for the logo

apps/web/pages/api/user/avatar.ts
joeauyeung, line 1: @hariombalhara @leog I noticed there's this endpoint and another endpoint `api/logo`, that also retrieves team logos. What is the difference between the two?

sean-brydon:
Looks good - some type errors failing for the gravitar fallback tho 🤔

sean-brydon:
LGTM from a UI side - not tested any of the other changes however

packages/features/ee/organizations/components/OrganizationAvatar.tsx
```diff
@@ -0,0 +1,41 @@
+import classNames from "@calcom/lib/classNames";
+import { Avatar } from "@calcom/ui";
+import type { AvatarProps } from "@calcom/ui";
+
+type OrganizationAvatarProps = Pick<
+  AvatarProps,
+  "size" | "imageSrc" | "alt" | "indicator" | "gravatarFallbackMd5"
+> & {
+  organizationSlug?: string | null;
```
hariombalhara, line 9: I would prefer it to be required, so that we don't miss passing the `organizationSlug` property.
```suggestion
  organizationSlug: string | null | undefined;
```

packages/features/ee/organizations/components/OrganizationAvatar.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const OrganizationAvatar = ({
+  size,
+  imageSrc,
+  alt,
+  gravatarFallbackMd5,
+  organizationSlug,
```
hariombalhara, line 17: It's not supporting `...rest` props. `OrganizationAvatar` being a subtype of `Avatar` should accept all props of `Avatar`

apps/web/pages/api/user/avatar.ts
```diff
@@ -64,6 +65,28 @@ async function getIdentityData(req: NextApiRequest) {
       avatar: team?.logo || getPlaceholderAvatar(null, teamname),
     };
   }
+  if (orgSlug) {
+    const org = await prisma.team.findFirst({
+      where: {
+        slug: orgSlug,
```
hariombalhara, line 71: You would need to check `requestedSlug` as well. See, the usage of `getSlugOrRequestedSlug` above

apps/web/pages/api/user/avatar.ts
```diff
@@ -64,6 +65,28 @@ async function getIdentityData(req: NextApiRequest) {
...
+        slug: true,
+        logo: true,
+        name: true,
+      },
+    });
+    return {
+      org: org?.slug,
+      name: org?.name,
+      email: null,
+      avatar: org?.logo || getPlaceholderAvatar(null, org?.name),
```
hariombalhara, line 87:
```suggestion
      avatar: getPlaceholderAvatar(org?.logo, org?.name),
```

`getPlaceholderAvatar` already takes care of handling nullish org.logo

apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -404,7 +408,13 @@ const ProfileForm = ({
           name="avatar"
           render={({ field: { value } }) => (
             <>
-              <Avatar alt="" imageSrc={value} gravatarFallbackMd5="fallback" size="lg" />
+              <OrganizationAvatar
+                alt=""
```
hariombalhara, line 412: We should use username here as it is done at other places.

packages/trpc/server/routers/loggedInViewer/me.handler.ts
```diff
@@ -24,7 +25,7 @@ export const meHandler = async ({ ctx }: MeOptions) => {
     locale: user.locale,
     timeFormat: user.timeFormat,
     timeZone: user.timeZone,
-    avatar: `${WEBAPP_URL}/${user.username}/avatar.png`,
+    avatar: `${userOrgDomain}/${user.username}/avatar.png`,
```
hariombalhara, line 28: There is `getBookerUrl` already that does this job. But that is async and does a query to fetch the org from orgId.

Maybe we can create a synchronous `getBookerUrl` which can accept the org directly and return the bookerUrl.

This is important because there is more logic in it than `getUserOrgDomain` has currently. It needs to check for `requestedSlug` as well.

packages/ui/components/avatar/AvatarGroup.tsx
```diff
@@ -36,7 +36,6 @@ export const AvatarGroup = function AvatarGroup(props: AvatarGroupProps) {
             imageSrc={item.image}
             title={item.title}
             alt={item.alt || ""}
-            accepted={props.accepted}
```
hariombalhara, line 39: What's the reason to remove it?

Also, if we are removing it we should cleanup `AvatarGroupProps` type as well.

packages/ui/components/avatar/Avatar.tsx
```diff
@@ -63,17 +64,7 @@ export function Avatar(props: AvatarProps) {
...
           </>
         </AvatarPrimitive.Fallback>
-        {props.accepted && (
-          <div
-            className={classNames(
-              "text-inverted absolute bottom-0 right-0 block rounded-full bg-green-400 ring-2 ring-white",
-              size === "lg" ? "h-5 w-5" : "h-2 w-2"
-            )}>
-            <div className="flex h-full items-center justify-center p-[2px]">
-              {size === "lg" && <Check />}
```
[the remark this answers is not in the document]
joeauyeung: This wasn't being used anywhere in the app so removed for now. #10788

packages/features/ee/organizations/lib/orgDomains.ts
```diff
@@ -64,3 +65,9 @@ export function getSlugOrRequestedSlug(slug: string) {
     ],
   } satisfies Prisma.TeamWhereInput;
 }
+
+export function getUserOrgDomain(organization: RouterOutputs["viewer"]["me"]["organization"]): string {
+  const orgSlug = organization.slug ?? organization.metadata?.requestedSlug;
```
joeauyeung, line 70: @hariombalhara would this satisfy this comment?
https://github.com/calcom/cal.com/pull/10700#discussion_r1299798360

sean-brydon:
CleanShot 2023-09-07 at 09 08 52
Overall LGTM - Noticed one weird case in /profile where the team logo is tiny. This can be handled in a follow up tho. 

Nice job

sean-brydon (approved):

PR #10703, opened 2023-08-10
fix: count without where clause is total users regardless of selection

emrysal:
## What does this PR do?

Straightforward fix to keep the count function from reporting the total user count but rather the paginated result.


alishaz-polymath (approved):
Nice! 🚀

PR #10709, opened 2023-08-10
fix: multiple duration getSchedule calls [CAL-2336]

### LINKED ISSUE #10707: [CAL-2336] getSchedule call fires twice

As getSchedule is our most resource intensive call we need to fix this urgently.

Additional context: seems to only happen on events with multiple durations

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2336](https://linear.app/calcom/issue/CAL-2336/getschedule-call-fires-twice)</sub>

leog:
## What does this PR do?

Due to the nature of multiple duration and recent changes that aimed to call as fast as possible the `getSchedule` result to populate the available slots, that call was made multiple times, first with an undefined duration and then the selected one.

Now we preemptively get the duration from query param if present, check if it's a possibility within the given event type options and return a valid duration if the event type has multiple duration configuration, either the query parameter value or the default duration if not valid. That duration is then passed down to the `getSchedule` call to effectively gain the performance optimization that was expected.

Fixes #10707 

## Type of change

## How should this be tested?

Go to `/pro/multiple-duration` event type and check in Dev Tools Network tab that `getSchedule` is called once when loading the page. Before this change it was called twice.

## Mandatory Tasks


packages/features/bookings/lib/get-booking.ts
```diff
@@ -224,3 +224,15 @@ export const getBookingForSeatedEvent = async (uid: string) => {
...
+  multipleDurationConfig: number[] | undefined,
+  queryDuration: string | string[] | undefined,
+  defaultValue: number
+) => {
+  return multipleDurationConfig
+    ? multipleDurationConfig.includes(Number(queryDuration))
+      ? Number(queryDuration)
+      : defaultValue
+    : null;
+};
```
zomars, line 238: Legibility nitpick

[Prefer Early Returns](https://handbook.cal.com/engineering/best-practices/prefer-early-returns) 

```suggestion
export const getMultipleDurationValue = (
  multipleDurationConfig: number[] | undefined,
  queryDuration: string | string[] | undefined,
  defaultValue: number
) => {
  if (!multipleDurationConfig) return null;
  if (multipleDurationConfig.includes(Number(queryDuration))) return Number(queryDuration);
  return defaultValue;
};
```

zomars (dismissed):

zomars (approved):

PR #10723, opened 2023-08-11
fix: include orgs slug in getUserByUsername just like [user]/[type] page

alannnc:
## What does this PR do?

- Handles ORG when looking up userId by username. This is the same way we do it on the SSR of booking pages.

Fixes # (issue)

## Type of change

## How should this be tested?

- What are the minimal test data to have? An Org with a username that it's the same as other without organization.
- What is expected (happy path) to have (input and output)? They should have their unique availability on their page if they both have the same eventSlug.

## Mandatory Tasks

## Checklist


packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -211,20 +215,28 @@ export async function getDynamicEventType(input: TGetScheduleInputSchema) {
...
-  return isDynamicBooking ? getDynamicEventType(input) : getEventType(input);
+  return isDynamicBooking ? getDynamicEventType(input) : getEventType(input, organizationDetails);
 }
 
-export async function getAvailableSlots(input: TGetScheduleInputSchema) {
+export async function getAvailableSlots(
+  ctx: { req: { headers: { host?: string | null } } },
+  input: TGetScheduleInputSchema
+) {
+  const orgDetails = orgDomainConfig(ctx.req.headers.host ?? "");
```
alannnc, line 230: Getting org slug from contex just as we do on booking pages.

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -480,10 +492,15 @@ export async function getAvailableSlots(input: TGetScheduleInputSchema) {
...
-async function getUserIdFromUsername(username: string) {
+async function getUserIdFromUsername(
+  username: string,
+  organizationDetails: { currentOrgDomain: string | null; isValidOrgDomain: boolean }
+) {
+  const { currentOrgDomain, isValidOrgDomain } = organizationDetails;
   const user = await prisma.user.findFirst({
     where: {
       username,
+      organization: isValidOrgDomain && currentOrgDomain ? getSlugOrRequestedSlug(currentOrgDomain) : null,
```
alannnc, line 503: This was required if an org user have the same username with another one without org.

zomars:
We got a type error 
image

zomars (changes requested):
Let's fix the types first 🙏🏽

alannnc:
@zomars No more type erros on this :)

Udit-takkar (dismissed):
Getting 404 on getSchedule

If a user creates a team "engineering" and another user has a team name "engineering" but in his organisation then i get 404 on getSchedule when i go the booking page.

Screenshot 2023-08-23 at 12 27 22 PM

alannnc:
> Getting 404 on getSchedule
> 
> If a user creates a team "engineering" and another user has a team name "engineering" but in his organisation then i get 404 on getSchedule when i go the booking page.
> 
> Screenshot 2023-08-23 at 12 27 22 PM

Nice find @Udit-takkar, it was a different bug from what I was solving but related as I applied the same logic to solve team names collition.

zomars (approved):
LGTM

PR #10727, opened 2023-08-12
fix: If the labels are empty for inputs then show default values and also show the last label if the inputs are in split mode in event types.

ZainGulbaz:
## What does this PR do?

This PR rectifies the label issue when  the label for field types of config variants in field form of event's booking question is split and shows default label if the label fields are empty

Fixes # (10390)

Before:
https://www.loom.com/share/63704afcc62b4253b26477f448de94ac
After:
https://www.loom.com/share/99a142ebd593499aa741251aac07226b

## Type of change

## How should this be tested?
The user can edit the event and from advanced setting he can edit the "name" field and change the label in split mode.
Add empty label in all fields or any one field.

## Checklist

### DIFF
--- a/packages/features/form-builder/FormBuilder.tsx
+++ b/packages/features/form-builder/FormBuilder.tsx
@@ -29,7 +29,7 @@ import { ArrowDown, ArrowUp, X, Plus, Trash2 } from "@calcom/ui/components/icon"
 import { fieldTypesConfigMap } from "./fieldTypes";
 import { fieldsThatSupportLabelAsSafeHtml } from "./fieldsThatSupportLabelAsSafeHtml";
 import type { fieldsSchema } from "./schema";
-import { getVariantsConfig } from "./utils";
+import { getVariantsConfig, getVariantFieldLabel } from "./utils";
 
 type RhfForm = {
   fields: z.infer<typeof fieldsSchema>;
@@ -532,6 +532,7 @@ function FieldLabel({ field }: { field: RhfFormField }) {
   const variantsConfig = field.variantsConfig;
   const variantsConfigVariants = variantsConfig?.variants;
   const defaultVariant = fieldTypeConfigVariantsConfig?.defaultVariant;
+
   if (!fieldTypeConfigVariants || !variantsConfig) {
     if (fieldsThatSupportLabelAsSafeHtml.includes(field.type)) {
       return (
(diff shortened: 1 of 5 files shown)


packages/features/form-builder/FormBuilder.tsx
```diff
@@ -553,8 +553,20 @@ function FieldLabel({ field }: { field: RhfFormField }) {
     );
   }
   const label =
-    variantsConfigVariants?.[variant as keyof typeof fieldTypeConfigVariants]?.fields?.[0]?.label || "";
-  return <span>{t(label)}</span>;
+    variantsConfigVariants?.[variant as keyof typeof fieldTypeConfigVariants]?.fields
```
hariombalhara, line 556: This is difficult to read, better would be to store it in a variable.

PeerRich:
why should it prefill with the name of the person logged in? that makes no sense

packages/features/form-builder/FormBuilderField.tsx
```diff
@@ -184,10 +184,12 @@ function getAndUpdateNormalizedValues(field: RhfFormFields[number], t: TFunction
     Object.entries(field.variantsConfig.variants).forEach(([variantName, variant]) => {
       variant.fields.forEach((variantField) => {
         const fieldTypeVariantsConfig = fieldTypesConfigMap[field.type]?.variantsConfig;
-        const defaultVariantFieldLabel =
-          fieldTypeVariantsConfig?.variants?.[variantName]?.fieldsMap[variantField.name]?.defaultLabel;
-
-        variantField.label = variantField.label || t(defaultVariantFieldLabel || "");
+        const fieldTypeVariantConfig = fieldTypeVariantsConfig?.variants[variantName];
+        variantField.label = getVariantFieldLabel({
```
hariombalhara, line 188: @ZainGulbaz Reused `getVariantFieldLabel` here to make the implementation consistent with the app that you implemented. Earlier, if just spaces are entered, then booker was showing just the spaces and not the fallback, now with your fix it would fallback to default.

packages/features/form-builder/utils.ts
```diff
@@ -67,3 +70,18 @@ export const getVariantsConfig = (field: Pick<z.infer<typeof fieldSchema>, "vari
...
+export function getVariantFieldLabel({
+  variantField,
+  fieldTypeVariantConfig,
+  t,
+}: {
+  variantField: VariantConfig["fields"][number];
+  fieldTypeVariantConfig: FieldTypeVariantConfig | undefined;
+  t: TFunction;
+}) {
+  if (!variantField.label?.trim()) {
```
hariombalhara, line 83: Your logic is here now. Also removed length check, it's a string and we can simply check for truthy.

hariombalhara:
Thank you for the PR @ZainGulbaz and double thanks @MehulZR for pointing out the issue. I have made a small refactor and it looks good to me.
@ZainGulbaz can you test from your end as well once.

PR #10728, opened 2023-08-12
chore: Simplified date overrides

emrysal:
## What does this PR do?

Saw something weird happening on Sean's side, so rewrote this logic in a way that should be DST compatible as well, and simplifying.


sean-brydon (approved):
Great! What a weird bug we found <3

PR #10730, opened 2023-08-13
fix: invitation to a team using username

### LINKED ISSUE #10726: [CAL-2341] Can't invite team members with username
### Issue Summary

Can't invite team members with username even though the input label is 'Email or Username'. On submitting the form with username the error is 'Invite failed because "username" is not a valid email address'.

### Steps to Reproduce

1. Try to add a member in your team with their username

### Screenshot

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2341](https://linear.app/calcom/issue/CAL-2341/cant-invite-team-members-with-username)</sub>

prane-eth:
## What does this PR do?

Fixes: #10726 
The issue mentions that we are unable to invite a user to a team using a username and have to enter an email address.

Before fix: (An error gets displayed on the bottom-right corner)
before-fix
After fix: (We can invite by username, and an invitation will show as pending without any errors)
after-fix

## Type of change

## Mandatory Tasks

## Note

- I haven't updated tests, as it was added by someone else under the same issue.

### DIFF
--- a/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
+++ b/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
@@ -22,6 +22,7 @@ import {
   getIsOrgVerified,
   sendVerificationEmail,
   createAndAutoJoinIfInOrg,
+  throwUsernameDoesNotExistError,
 } from "./utils";
 
 type InviteMemberOptions = {
@@ -51,11 +52,14 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
     });
     const invitee = await getUserToInviteOrThrowIfExists({
       usernameOrEmail,
-      orgId: input.teamId,
       isOrg: input.isOrg,
     });
 
     if (!invitee) {
+      const isInputEmail = isEmail(usernameOrEmail);
+      if (!isInputEmail) {
+        throwUsernameDoesNotExistError(usernameOrEmail);
+      }
       checkInputEmailIsValid(usernameOrEmail);
 
       // valid email given, create User and add to team
--- a/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
+++ b/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
@@ -14,7 +14,6 @@ import {
   getTeamOrThrow,
   getEmailsToInvite,
   getUserToInviteOrThrowIfExists,
-  checkInputEmailIsValid,
   getOrgConnectionInfo,
   createNewUserConnectToOrgIfExists,
   throwIfInviteIsToOrgAndUserExists,
@@ -61,8 +60,6 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
       if (!isEmail(usernameOrEmail)) {
         throwUsernameDoesNotExistInOrgError(usernameOrEmail);
       }
-
-      checkInputEmailIsValid(usernameOrEmail);
 
       // valid email given, create User and add to team
       await createNewUserConnectToOrgIfExists({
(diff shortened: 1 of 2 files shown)


packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
```diff
@@ -51,11 +52,14 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
...
       usernameOrEmail,
-      orgId: input.teamId,
       isOrg: input.isOrg,
     });
 
     if (!invitee) {
+      const isInputEmail = isEmail(usernameOrEmail);
+      if (!isInputEmail) {
+        throwUsernameDoesNotExistError(usernameOrEmail);
+      }
```
ologbonowiwi, line 62: This doesn't make too much sense. Throw that user doesn't exist without even search for it doesn't make too much sense for me 🤔

packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
```diff
@@ -66,17 +66,15 @@ export async function getEmailsToInvite(usernameOrEmail: string | string[]) {
...
   isOrg,
 }: {
   usernameOrEmail: string;
-  orgId: number;
   isOrg?: boolean;
 }) {
   // Check if user exists in ORG or exists all together
   const invitee = await prisma.user.findFirst({
     where: {
-      OR: [{ username: usernameOrEmail, organizationId: orgId }, { email: usernameOrEmail }],
```
InderdeepBajwa, line 79: Removing `organizationId: orgId` may not retain the right logic as [inviteMember.handler.ts#70](https://github.com/calcom/cal.com/blob/main/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts#L70) relies on the check that the provided username is already a part of the organization.

prane-eth: @InderdeepBajwa, Thank you so much for the feedback.
I updated the code. Can you please check?

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
```diff
@@ -56,6 +57,11 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
     });
 
     if (!invitee) {
+      // if input is a username, throw error
+      if (!isEmail(usernameOrEmail)) {
+        throwUsernameDoesNotExistInOrgError(usernameOrEmail);
+      }
+
```
Udit-takkar, line 64: checkInputEmailIsValid on L65 does this same thing

Udit-takkar (commented):
@sean-brydon i think there is some issue with getUserToInviteOrThrowIfExists function. 

```
  const invitee = await prisma.user.findFirst({
    where: {
      OR: [{ username: usernameOrEmail, organizationId: orgId }, { email: usernameOrEmail }],
    },
  });
```
This only checks username with organisationId and we are passing input.teamId to organisationId so it might not fetch the user if we pass username in input and teamId is normal team and not organisation

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
prane-eth: @Udit-takkar, Thank you for finding out.
I just removed checkInputEmailIsValid as it is not required.
If the input is not a valid email, the code above will assume it as a username.

Udit-takkar (changes requested):
Screenshot 2023-08-29 at 12 37 05 PM

This doesn't work and previous comment was not addressed by you.

prane-eth:
@Udit-takkar, it works if the username exists within the organization. According to the current functionality, we need to send an invitation if the username exists within the organization.

https://github.com/calcom/cal.com/assets/48318416/e9d9c999-b3d6-4969-9151-5db9624beb0a

PR #10740, opened 2023-08-14
feat: Availability slider for orgs

sean-brydon:
Implements a WTB type of display for users availability and timezones for use in ORGS only

How to test: 
Create a ORG
Invite a few users
go through onboarding with them and change the timezone of each user.
Visit /availability you should see a tab to switch to your team or visit /availability?type=team directly.
Compare generated hours with the worldtimebuddy equivalent. 
Ensure availability matches up with the users default schedule 

https://www.loom.com/share/e0508a3b048e4f4fb4a67682f4373ec1
CleanShot 2023-08-14 at 14 25 52@2x
CleanShot 2023-08-14 at 14 25 49@2x


apps/web/pages/availability/index.tsx
```diff
@@ -129,14 +132,54 @@ const WithQuery = withQuery(trpc.viewer.availability.list as any);
...
+                defaultValue={searchParams?.get("type") ?? "mine"}
+                onValueChange={(value) => {
+                  router.push(`${pathname}?${createQueryString("type", value)}`);
+                }}
+                options={[
+                  { value: "mine", label: t("my_availability") },
+                  { value: "team", label: t("team_availability") },
+                ]}
+              />
+            )}
```
sean-brydon, line 171: If we are in an org -> Display toggle to switch between your availability and your teams

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -0,0 +1,165 @@
...
+        accessorFn: (data) => data.timeZone,
+        header: "Timezone",
+        cell: ({ row }) => {
+          const { timeZone } = row.original;
+          const timeRaw = dayjs().tz(timeZone);
+          const time = timeRaw.format("HH:mm");
+          const utcOffsetInMinutes = timeRaw.utcOffset();
+          const hours = Math.abs(Math.floor(utcOffsetInMinutes / 60));
+          const minutes = Math.abs(utcOffsetInMinutes % 60);
+          const offsetFormatted = `${utcOffsetInMinutes < 0 ? "-" : "+"}${hours
```
sean-brydon, line 79: Math to calculate the offset from UTC

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -0,0 +1,165 @@
...
+      value={createTimezoneBuddyStore({
+        browsingDate: browsingDate.toDate(),
+      })}>
+      <div className="relative">
+        <DataTable
+          tableContainerRef={tableContainerRef}
+          columns={memorisedColumns}
+          data={flatData}
+          isLoading={isLoading}
+          // tableOverlay={<HoverOverview />}
```
sean-brydon, line 159: Will be added in a future PR

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  useEffect(() => {
+    fetchMoreOnBottomReached(tableContainerRef.current);
+  }, [fetchMoreOnBottomReached]);
+
+  return (
+    <TBContext.Provider
+      value={createTimezoneBuddyStore({
+        browsingDate: browsingDate.toDate(),
+      })}>
```
sean-brydon, line 152: As per zustand docs for creating stores where you pass in the value via props. 

We are doing this wrong in other places

packages/features/timezone-buddy/components/HoverOverview.tsx
sean-brydon, line 1: This file is not currently used and will be fully implemented in a follow up PR to emulate the selection style that WTB/the designs in the figma

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -0,0 +1,126 @@
...
+  offset: number;
+}) {
+  if (!dateRanges) return false;
+  const currentHour = cellDate.hour();
+
+  return dateRanges.some((time) => {
+    if (!time) null;
+
+    const startHour = dayjs(time.start).subtract(offset, "hour");
+    const endHour = dayjs(time.end).subtract(offset, "hour");
```
sean-brydon, line 37: Apply the offset from the users browsing timezone

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -0,0 +1,126 @@
...
+  if (!store) throw new Error("Missing TBContext.Provider in the tree");
+  const browsingDate = useStore(store, (s) => s.browsingDate);
+
+  const usersTimezoneDate = dayjs(browsingDate).tz(timezone);
+
+  const nowDate = dayjs(browsingDate);
+
+  const offset = (nowDate.utcOffset() - usersTimezoneDate.utcOffset()) / 60;
+
+  const hours = Array.from({ length: 24 }, (_, i) => i - offset + 1);
```
sean-brydon, line 63: Ensure the hours are correct basedo n the UTC offset diff from the nowDate of the provider and the timezone of the current row users tz

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const days = [
+    hours.filter((i) => i < 0).map((i) => (i + 24) % 24),
+    hours.filter((i) => i >= 0 && i < 24),
+    hours.filter((i) => i >= 24).map((i) => i % 24),
+  ];
```
sean-brydon, line 69: Separate the hours into days.

If I is < 0 we add 24 and get the remainder.
if its in the range of 0-24 its fine
Above we do the reverse

Allows us to split into 3 sections. Yesterday - today - tomorrow. 

All can be present in each timezone depending on scale

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -0,0 +1,126 @@
...
+          if (!day.length) return null;
+          const dateWithDaySet = usersTimezoneDate.add(i - 1, "day");
+          return (
+            <div key={i} className={classNames("flex flex-none overflow-hidden rounded-lg border border-2")}>
+              {day.map((h) => {
+                const hours = Math.floor(h); // Whole number part
+                const fractionalHours = h - hours; // Decimal part
+
+                // Convert the fractional hours to minutes
+                const minutes = fractionalHours * 60;
```
sean-brydon, line 84: Calculate the offset into minutes -> 9.5 -> 9h. 30min

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  <div
+                    key={h}
+                    className={classNames(
+                      "flex h-8 flex-col items-center justify-center",
+                      isInRange ? "text-emphasis bg-success" : "",
+                      hours ? "" : "bg-subtle font-medium"
+                    )}
+                    style={{
+                      width: `${DAY_CELL_WIDTH}px`,
+                    }}>
```
sean-brydon, line 104: Extracted width to constant in constants.ts will be used when dealing with the hover state

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        <div className="flex flex-col text-center text-xs leading-3">
+                          <span>{hourSet.format("H")}</span>
+                        </div>
+                      </div>
+                    ) : (
+                      <div className="flex flex-col text-center text-xs leading-3">
+                        <span>{hourSet.format("MMM")}</span>
+                        <span>{hourSet.format("DD")}</span>
+                      </div>
+                    )}
```
sean-brydon, line 116: If hours is 0 (midnight) we display new day with date instead of the default hour (0)

packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  let nextCursor: typeof cursor | undefined = undefined;
+  if (teamMembers && teamMembers.length > limit) {
+    const nextItem = teamMembers.pop();
+    nextCursor = nextItem!.id;
+  }
+
+  const dateFrom = dayjs(input.startDate).tz(input.loggedInUsersTz).subtract(1, "day");
+  const dateTo = dayjs(input.endDate).tz(input.loggedInUsersTz).add(1, "day");
```
sean-brydon, line 66: We have to -1 and +1 to get the range. 

As this fetches on YOUR browser timezone it would not include anything before that point. In cases of other timezones they may have 5-6 hours left of the day

packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  const buildMembers = teamMembers?.map(async (member) => {
+    if (!member.user.defaultScheduleId) {
+      return {
+        id: member.user.id,
+        username: member.user.username,
+        email: member.user.email,
+        timeZone: member.user.timeZone,
+        role: member.role,
+        dateRanges: [] as DateRange[],
```
sean-brydon, line 76: If there is no defaultScheduleId we just return no slots

packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -0,0 +1,112 @@
...
+      };
+    }
+
+    const schedule = await prisma.schedule.findUnique({
+      where: { id: member.user.defaultScheduleId },
+      select: { availability: true, timeZone: true },
+    });
+    const timeZone = schedule?.timeZone || member.user.timeZone;
```
sean-brydon, line 84: We ideally want to use the schedule timezone but if one isnt found we fallback to the users timezone defined in profile

packages/ui/components/data-table/index.tsx
```diff
@@ -164,6 +166,7 @@ export function DataTable<TData, TValue>({
               </tr>
             )}
           </TableBody>
+          {tableOverlay && tableOverlay}
```
sean-brydon, line 169: Option to have a tableoverlay to implement custom hovers / anything that needs to be relative to the table.

Udit-takkar (commented):
Screenshot 2023-08-16 at 4 14 22 PM

No team availability and heading switch in mobile

Edit: This is intended behaviour

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -0,0 +1,165 @@
...
+          const utcOffsetInMinutes = timeRaw.utcOffset();
+          const hours = Math.abs(Math.floor(utcOffsetInMinutes / 60));
+          const minutes = Math.abs(utcOffsetInMinutes % 60);
+          const offsetFormatted = `${utcOffsetInMinutes < 0 ? "-" : "+"}${hours
+            .toString()
+            .padStart(2, "0")}:${minutes.toString().padStart(2, "0")}`;
+
+          return (
+            <div className="flex flex-col gap-2 leading-none">
+              <span className="text-default text-sm font-medium">{time}</span>
```
Udit-takkar, line 85: [Not a big issue] I have 12 hour time format selected but i still see everything 24 hour format.

WTB supports 12 hour format 

Screenshot 2023-08-16 at 4 20 17 PM

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -0,0 +1,165 @@
...
+export interface SliderUser {
+  id: number;
+  username: string | null;
+  email: string;
+  timeZone: string;
+  role: MembershipRole;
+  dateRanges: DateRange[];
+}
+
+export function AvailabilitySliderTable() {
```
Udit-takkar, line 23: Screenshot 2023-08-16 at 4 36 27 PM

I had my availability set up between 9 - 4:30PM but i see 9 - 5PM

packages/features/timezone-buddy/components/HoverOverview.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    function handleMouseUp() {
+      setIsPressed(false);
+    }
+
+    document.addEventListener("mousemove", handleMouseMove);
+    document.addEventListener("mousedown", handleMouseDown);
+    document.addEventListener("mouseup", handleMouseUp);
+
```
Udit-takkar, line 63: should we use throttling for these events ?

Udit-takkar (commented):
@sean-brydon Great PR 🚀

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -0,0 +1,165 @@
...
+      },
+    ];
+
+    return cols;
+  }, [browsingDate]);
+
+  //we must flatten the array of arrays from the useInfiniteQuery hook
+  const flatData = useMemo(() => data?.pages?.flatMap((page) => page.rows) ?? [], [data]) as SliderUser[];
+  const totalDBRowCount = data?.pages?.[0]?.meta?.totalRowCount ?? 0;
+  const totalFetched = flatData.length;
```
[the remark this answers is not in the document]
sean-brydon: We can but it actually leads to more code -> Since we can't use memo or have an easy way to define variables inside there. You have to useState/let. This is the cleanest way i've found to do this rn. Will likely try to transform this whole pagination thing into a hook at some point in the future

packages/features/timezone-buddy/components/HoverOverview.tsx
sean-brydon: Great suggestion! I will deffo consider this when i come to implementing the hover view. Not 100% sure if its needed as the only thing this will be rendering is the x position of a box. So it won't be causing re-renders down the tree

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
sean-brydon: Will cover this in a follow up i think. We can do with 24h support for now. It requires some weirdness with dates and displaying that we need to figure out. @Jaibles we should probably figure out how we're going to display this. Likely the same approach as WTB also

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
sean-brydon: Wait this is right isnt it? Yes youre only available for 1/2 of 4:30 but your still available in the 4 slot? I wonder if we should display an amber cell here let me try

sean-brydon: Screenshot above appears to be right for 9-4:30? 

Granted 4 shell should probably be amber. Will play around in figma to see if we can get something to show this.

Will tackle in a follow up if you believe this isnt a blocker @Udit-takkar

Udit-takkar: @sean-brydon this is a bit confusing because 9 - 5 also shows this same UI. the screenshot shows slots 9 - 10, 10- 11, 11-12, 12 - 13, 13- 14, 14-15, 15- 16, 16-17.  the last slots should be 16:00 -  16:30. maybe because there isn't any clear boundary that help differentiate where does 4pm starts and ends. I expected 16:30 to be where "16" is displayed in the screenshot because 9:00 AM in the middle of 8 and 9 so middle of 16 and 17 would be 17:00 and not 16:30

sean-brydon: I understand - we need to make some compromises here. As we can't seperate into 30 min cells as we'd just get the same issue with 45 min offset timezones. @Jaibles has cleared this up in the designs for the selection hover
CleanShot 2023-08-17 at 09 19 10@2x

packages/features/timezone-buddy/components/TimeDial.tsx
```diff
@@ -68,57 +102,114 @@ export function TimeDial({ timezone, dateRanges }: TimeDialProps) {
...
+                      rangeGradients.darkTextGradient = `linear-gradient(90deg, var(--cal-text-inverted, white) ${
+                        rangeOverlap === 50 ? "50%" : Math.round(rangeOverlap / 100) * 100 + "%"
+                      }, var(--cal-text-emphasis, #111827) 0%, var(--cal-text-emphasis, #111827) 0%)`;
+                    }
+                  }
+
+                  const minuteOffsetStyles: { marginLeft?: string } = {};
+                  if (hours !== 0 && !minuteOffsetApplied) {
+                    minuteOffsetApplied = true;
+                    minuteOffsetStyles.marginLeft = `${DAY_CELL_WIDTH * (offset % 1)}px`;
```
sean-brydon, line 160: @emrysal X%1 is always 0 no?

emrysal (approved):
Tested, works :)

packages/features/timezone-buddy/components/TimeDial.tsx
emrysal: yes - that's intentional it should only shift for decimal places like 3.5 and 4.75

PR #10762, opened 2023-08-15
chore: add constant for production builds

zomars:
## What does this PR do?

Adds specific constant for knowing if it's a production build.

## Type of change


keithwillcode (dismissed):

packages/lib/constants.ts
```diff
@@ -4,6 +4,7 @@ const HEROKU_URL = process.env.HEROKU_APP_NAME ? `https://${process.env.HEROKU_A
 const RENDER_URL = process.env.RENDER_EXTERNAL_URL ? `https://${process.env.RENDER_EXTERNAL_URL}` : "";
 export const CALCOM_ENV = process.env.CALCOM_ENV || process.env.NODE_ENV;
 export const IS_PRODUCTION = CALCOM_ENV === "production";
+export const IS_PRODUCTION_BUILD = process.env.NODE_ENV === "production";
```
leog, line 7: Wouldn’t `IS_PRODUCTION` achieve the same result when no `process.env.CALCOM_ENV` is defined?

zomars: I still think it's important to differentiate deployments and builds

leog (approved):
Looks good, although couldn't find a place where it's being used.

PR #10765, opened 2023-08-15
test: Create unit tests for react components in packages/ui/components/form/TimezoneSelect

gitstart-calcom:
command to run these tests: `yarn test timezone`

image

obs: the props that are passed to the component, have a specific and unique effect in the CSS, so the assertions are made with the hasClass


ThyMinimalDev:
hey there, not sure to understand what is the goal with these tests, we seem to mostly assess if the right css class is assigned but we are not really testing the behaviour of the component here, is it wanted and if yes what is the added benefit of these tests ?

gitstart-calcom:
HI @ThyMinimalDev in this specific case, since we are using a dropdown from the react-select library, we can customize any specific parts of this component using these props (valueContainer, input, control, etc). So we are testing if this props is working properly in this case. We are also testing if the component have the correct behavior when is loading.

keithwillcode (dismissed):

keithwillcode:
@zomars we are getting heap issues here in the unit tests. Any recs based on recent changes you’ve made around this?

keithwillcode (approved):

PR #10770, opened 2023-08-15
perf: Remove duplicate call for user event types

### LINKED ISSUE #10505: perf: Improve performance of /event-types
### Issue Summary

This page is performing far too many database queries. We need to do a deep dive into what's truly needed for this page to operate and limit the queries it's running.

### Steps to Reproduce

1. Log in and go to /event-types

### Actual Results

- Takes multiple seconds to load this page, specifically the `getByViewer` tRPC endpoint

### Expected Results

- This should be extremely fast to load

### Evidence

- As any user, log into the app and go to /event-types. It will be slow.

keithwillcode:
## What does this PR do?

Removes a duplicate call to get all event types linked directly to a user.

## Type of change

## How should this be tested?

- Ensure that all event types created by the user load
- Ensure managed event types load for users

## Mandatory Tasks

### DIFF
--- a/packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
+++ b/packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
@@ -160,24 +160,6 @@ export const getByViewerHandler = async ({ ctx, input }: GetByViewerOptions) => 
   });
 
   const userEventTypes = user.eventTypes.map(mapEventType);
-  // backwards compatibility, TMP:
-  const typesRaw = (
-    await prisma.eventType.findMany({
-      where: {
-        userId: getPrismaWhereUserIdFromFilter(ctx.user.id, input?.filters),
-        team: null,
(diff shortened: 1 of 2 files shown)


emrysal (approved):
Don't see a problem with this; lgtm 👍

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -160,24 +160,6 @@ export const getByViewerHandler = async ({ ctx, input }: GetByViewerOptions) =>
   });
 
   const userEventTypes = user.eventTypes.map(mapEventType);
-  // backwards compatibility, TMP:
-  const typesRaw = (
-    await prisma.eventType.findMany({
-      where: {
```
keithwillcode, line 166: This query is exactly the same as what's above this in the main query to get the user.

PR #10771, opened 2023-08-15
fix: add 2fa and email verification grace period

nicktrn:
## What does this PR do?

Accept the previous TOTP code in addition to the current one.

This is in line with the recommendations in [RFC 6238](https://datatracker.ietf.org/doc/html/rfc6238#section-5.2:~:text=We%20RECOMMEND%20that%20at%20most%20one%0A%20%20%20time%20step%20is%20allowed%20as%20the%20network%20delay).

If you submit a TOTP token towards the end of the period, it will be invalid by the time you submit it. Came across this while working on #10600 and it's rather annoying.

This also fixes an issue where email verification codes could be invalid by the time they are received. With these changes, the codes are now valid for a duration of 15-30 minutes, instead of the previous 0-15 minutes.

## Type of change

## How should this be tested?

1. Ensure 2FA is enabled.
2. Enter the current TOTP code during login
3. Only submit it in the next window, i.e. when a new code is showing in your authenticator
4. The previous code should be accepted with this fix.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/lib/totp.ts
+++ b/packages/lib/totp.ts
@@ -12,7 +12,11 @@ import { keyDecoder, keyEncoder } from "@otplib/plugin-thirty-two";
  * @param opts - The AuthenticatorOptions object.
  * @param opts.window - The amount of past and future tokens considered valid. Either a single value or array of `[past, future]`. Default: `[1, 0]`
  */
-export const totpAuthenticatorCheck = (token: string, secret: string, opts: AuthenticatorOptions) => {
+export const totpAuthenticatorCheck = (
+  token: string,
+  secret: string,
+  opts: Partial<AuthenticatorOptions> = {}
+) => {
   const { window = [1, 0], ...rest } = opts;
   const authenticator = new Authenticator({
     createDigest,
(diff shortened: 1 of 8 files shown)


packages/lib/totp.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Checks the validity of a TOTP token using a raw secret.
+ *
+ * @param token - The token.
+ * @param secret - The raw hex-encoded shared secret.
+ * @param opts - The TOTPOptions object.
+ * @param opts.window - The amount of past and future tokens considered valid. Either a single value or array of `[past, future]`. Default: `[1, 0]`
+ */
+export const totpRawCheck = (token: string, secret: string, opts: Partial<TOTPOptions> = {}) => {
```
nicktrn, line 40: The only difference here is the encoding. Authenticator secrets are base32-encoded, these are just hex-encoded. Necessary for use with the current email verification implementation.

Udit-takkar (approved):
@nicktrn LGTM 🚀🚀

nicktrn:
:pray:

PR #10772, opened 2023-08-15
chore: Don't render certain components when running meticulous tests

keithwillcode:
## What does this PR do?

Fixes the red X's we see when Meticulous runs and detects visual changes that are triggered by URL or version number changes, which are not real changes.

## Type of change


keithwillcode:
@shivamklr Want to take this over and fix the places where we show a URL, which per deployment is different and causing meticulous to think there are differences?

shivamklr:
@keithwillcode Taking a look.

keithwillcode (commented):
@shivamklr Nice work. @emrysal and I chatted about this and think we should put something like `const IS_VISUAL_REGRESSION_TESTING = Boolean(globalThis.window?.Meticulous?.isRunningAsTest);` in `constants.ts` for the following reasons:
- It's more obvious as to what the code is doing
- It shields the dependency of using Meticulous behind a "wall" instead of being repeated multiple times throughout the code.

keithwillcode (commented):
@shivam I see sessions running on localhost. Can we update the code such that Meticulous isn't recording those.

shivamklr:
This PR is not yet ready to merge. I am trying to figure out what would meticulous show in the dashboard after the following changes.

sean-brydon (commented):
I'd honestly rather remove meticulous than have to manage this across the app. Let me know what you think.

keithwillcode:
> I'd honestly rather remove meticulous than have to manage this across the app. Let me know what you think.

Yeah I see what you mean. Fortunately it seems to just be a few places we need this. If it were more, I’d definitely say it would be a harder sell. 

I think a good plan is to get this running and passing, see the value we get and then decide from there.

hariombalhara (approved):
Agree with both @sean-brydon and @keithwillcode

Let's see if benefits outweigh the drawbacks. The biggest benefit that I see is it being low cost in maintenance.

~~Also, which components in general we should not render?~~
Edit: I saw that in PR description

PR #10779, opened 2023-08-15
feat: collapse event type details

### LINKED ISSUE #10768: Individual user profile page - less event info shown in overview
Is it possible to only display a short summary of the events in the individual user profil? 

I'd like to only see the first couple of lines as I see them in my event type page and not the whole text - making it kind of untidy. 

For example, check my current profil: www.cal.com/buchen

Another useful workaround / feature would be to categorize event types i. e. individual / group / online, so that visitors of the profile page would have to choose from the category before they see the single events with description.

Thanks for your response!

ghost:
## What does this PR do?

Fixes #10768

Adds a feature in which user can expand or collapse huge event type's description

https://github.com/calcom/cal.com/assets/138296762/12991452-44ad-4034-8744-0f416fbbbceb

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How should this be tested?

- Create an event type with huge description.
- Go to your public page.
- Event type with huge description is clamped and can be expanded and collapsed

## Mandatory Tasks

### DIFF
--- a/apps/web/pages/[user].tsx
+++ b/apps/web/pages/[user].tsx
@@ -149,7 +149,7 @@ export function UserPage(props: InferGetServerSidePropsType<typeof getServerSide
                       <div className="flex flex-wrap items-center">
                         <h2 className=" text-default pr-2 text-sm font-semibold">{type.title}</h2>
                       </div>
-                      <EventTypeDescription eventType={type} isPublic={true} />
+                      <EventTypeDescription eventType={type} isPublic={true} shortenDescription />
                     </Link>
                   </div>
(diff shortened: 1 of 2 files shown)


Udit-takkar (commented):
@sean-brydon you added line-clamp-3 on the event type description https://github.com/calcom/cal.com/pull/10795. should we also add expland/collapse dropdown ?

vineetpradhan-dev:
@Udit-takkar Can I put the same logic with line-clamp-3 instead of line-clamp-4

PR #10781, opened 2023-08-15
refactor: Eventtype app context after orgs

joeauyeung:
## What does this PR do?

This is based on #9950.

This PR refactors the eventtypeAppContext from an array to an object to grab only what is needed.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Enable event type apps
- Go to the event type app page
- Install apps
    - Ensure that changes can be made and saved
- Create a managed event type
    - Ensure that the owner can make changes to the event type apps
    - Go to the child event type and ensure that changes can not be made

## Mandatory Tasks

## Checklist


packages/app-store/EventTypeAppContext.tsx
```diff
@@ -5,13 +5,20 @@ export type GetAppData = (key: string) => unknown;
...
-  () => ({}),
-  undefined,
-  undefined,
-]);
+const EventTypeAppContext = React.createContext<AppContext>({
+  getAppData: () => ({}),
+  setAppData: () => ({}),
+  LockedIcon: undefined,
+  disabled: undefined,
+});
```
joeauyeung, line 21: I moved the context from an array into an object. Not every item was needed from the context, so with an object, we're able to call on only the things we needed rather than everything.

https://github.com/calcom/cal.com/pull/9950#discussion_r1253489346

packages/app-store/_components/AppCard.tsx
```diff
@@ -16,24 +16,22 @@ export default function AppCard({
   switchOnClick,
   switchChecked,
   children,
-  setAppData,
```
joeauyeung, line 19: We can call the setAppData from the context, so we don't need to pass this down as a prop.

https://github.com/calcom/cal.com/pull/9950#discussion_r1253491752

packages/app-store/_components/AppCard.tsx
```diff
@@ -16,24 +16,22 @@ export default function AppCard({
...
   returnTo,
   teamId,
-  disableSwitch,
-  LockedIcon,
```
joeauyeung, line 23: These managed event type fields can also be called from the context now.

packages/app-store/_components/AppCard.tsx
```diff
@@ -16,24 +16,22 @@ export default function AppCard({
...
+  appDataSchema,
```
joeauyeung, line 21: We do need to pass the appDataSchema though so the context gets the right types.

https://github.com/calcom/cal.com/pull/9950#discussion_r1253492034

packages/app-store/fathom/components/EventTypeAppCardInterface.tsx
```diff
@@ -4,19 +4,17 @@ import useIsAppEnabled from "@calcom/app-store/_utils/useIsAppEnabled";
 import type { EventTypeAppCardComponent } from "@calcom/app-store/types";
 import { TextField } from "@calcom/ui";
 
-import type { appDataSchema } from "../zod";
+import { appDataSchema } from "../zod";
 
 const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({ app, eventType }) {
-  const [getAppData, setAppData, LockedIcon, disabled] = useAppContextWithSchema<typeof appDataSchema>();
+  const { getAppData, setAppData, disabled } = useAppContextWithSchema<typeof appDataSchema>();
```
joeauyeung, line 10: This is an example where we don't need the LockedIcon so with an object we can call on what we do need. Also makes it easier if we have to add more data to the context in the future.

https://github.com/calcom/cal.com/pull/9950#discussion_r1253492838

packages/app-store/qr_code/components/EventTypeAppCardInterface.tsx
```diff
@@ -7,11 +7,11 @@ import type { EventTypeAppCardComponent } from "@calcom/app-store/types";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { Tooltip, TextField } from "@calcom/ui";
 
-import type { appDataSchema } from "../zod";
+import { appDataSchema } from "../zod";
 
 const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({ eventType, app }) {
   const { t } = useLocale();
-  const [getAppData, setAppData, LockedIcon, disabled] = useAppContextWithSchema<typeof appDataSchema>();
+  const { disabled } = useAppContextWithSchema<typeof appDataSchema>();
```
joeauyeung, line 14: Another good example is where we only need the disabled prop from the context. Better than grabbing everything.

https://github.com/calcom/cal.com/pull/9950#discussion_r1253493983

packages/app-store/EventTypeAppContext.tsx
```diff
@@ -5,13 +5,20 @@ export type GetAppData = (key: string) => unknown;
...
-  () => ({}),
-  () => ({}),
-  undefined,
-  undefined,
-]);
+const EventTypeAppContext = React.createContext<AppContext>({
+  getAppData: () => ({}),
+  setAppData: () => ({}),
+  LockedIcon: undefined,
+  disabled: undefined,
```
hariombalhara, line 20: I think we can remove this undefined assignment now. Earlier it was needed because it was an array and it has to have all the items for clarity.

packages/app-store/_components/AppCard.tsx
hariombalhara, line 21: We are using `appDataSchema` for it's type only. So, I don't think we should pass it as prop.
Also, I don't think we need to pass in the schema type to `useAppContextWithSchema`, we can work without that.

packages/app-store/EventTypeAppContext.tsx
hariombalhara: Fixed it.

packages/app-store/_components/AppCard.tsx
hariombalhara: Fixed it.

hariombalhara (approved):

PR #10782, opened 2023-08-15
chore: add DatePicker in storybook

gitstart-calcom:
## Loom

https://www.loom.com/share/38d89f3aa5a140e2a3e526dc99b3d3c5

The DatePicker currently has no support for dark mode, so it doesn't look very good in the dark background


gitstart-calcom:
We will push another PR with the same changes 😄

PR #10788, opened 2023-08-15
feat: organization avatar component

joeauyeung:
## What does this PR do?

	
This PR adds the `indicator` prop to the `Avatar` component and adds an `OrganizationAvatar` component.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/ui/components/avatar/Avatar.tsx
```diff
@@ -66,27 +64,7 @@ export function Avatar(props: AvatarProps) {
...
-          <div
-            className={classNames(
-              "text-inverted absolute bottom-0 right-0 block rounded-full bg-green-400 ring-2 ring-white",
-              size === "lg" ? "h-5 w-5" : "h-2 w-2"
-            )}>
-            <div className="flex h-full items-center justify-center p-[2px]">
-              {size === "lg" && <Check />}
-            </div>
-          </div>
-        )}
```
joeauyeung, line 79: I looked through the code base for `accepted={` and only found it being used by `AvatarGroup`. That was the only result so I'm assuming that this wasn't being used.

packages/features/ee/organizations/components/OrganizationAvatar.tsx
```diff
@@ -0,0 +1,39 @@
...
+}: OrganizationAvatarProps) => {
+  return (
+    <Avatar
+      size={size}
+      imageSrc={imageSrc}
+      alt={alt}
+      indicator={
+        (organizationLogo || organizationName) && (
+          <div
+            className={classNames("absolute bottom-0 right-0 z-10", size === "lg" ? "h-5 w-5" : "h-10 w-10")}>
```
sean-brydon, line 26: I think the sizes are wrong here  - Pushing a storybook file change so we can easily test this also 
CleanShot 2023-08-16 at 09 48 09@2x

sean-brydon, line 26:
```suggestion
            className={classNames("absolute bottom-0 right-0 z-10", size === "lg" ? "h-3 w-3" : "h-10 w-10")}>
```

h-5 w-5 doesnt exist in tailwind

sean-brydon (approved):
Added a storybook component and change the sizes as w-5 and h-5 doesnt exist. 

We have some merge conflicts tho :/

PR #10792, opened 2023-08-16
fix: currency for event types

### LINKED ISSUE #10682: [CAL-2329] default currency not showing on new-booker or payment page
Add default currency form payment.currency table.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2329](https://linear.app/calcom/issue/CAL-2329/default-currency-not-showing-on-new-booker-or-payment-page)</sub>

alannnc:
## What does this PR do?

- Fixes event type metadata that wasn't getting payment app so it would save wrong data.
- fixes getEventTypeAppData to favour eventType price and currency over wrong app metadata.

Fixes #10682 
## Requirement/Documentation

## Type of change

## How should this be tested?

### DIFF
--- a/apps/web/pages/booking/[uid].tsx
+++ b/apps/web/pages/booking/[uid].tsx
@@ -96,7 +96,7 @@ const querySchema = z.object({
 });
 
 export default function Success(props: SuccessProps) {
-  const { t } = useLocale();
+  const { t, i18n } = useLocale();
   const router = useRouter();
   const routerQuery = useRouterQuery();
   const pathname = usePathname();
@@ -477,6 +477,21 @@ export default function Success(props: SuccessProps) {
                           ) : (
                             locationToDisplay
                           )}
+                        </div>
+                      </>
+                    )}
+                    {props.paymentStatus && (
+                      <>
+                        <div className="mt-3 font-medium">
+                          {props.paymentStatus.paymentOption === "HOLD"
+                            ? t("complete_your_booking")
+                            : t("payment")}
+                        </div>
+                        <div className="col-span-2 mb-2 mt-3">
+                          {new Intl.NumberFormat(i18n.language, {
+                            style: "currency",
+                            currency: props.paymentStatus.currency,
+                          }).format(props.paymentStatus.amount / 100.0)}
                         </div>
                       </>
                     )}
@@ -1117,6 +1132,9 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
     select: {
       success: true,
       refunded: true,
+      currency: true,
+      amount: true,
+      paymentOption: true,
     },
   });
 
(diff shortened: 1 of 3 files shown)


packages/lib/getEventTypeById.ts
```diff
@@ -227,7 +227,7 @@ export default async function getEventTypeById({
       app: {
         enabled: true,
         categories: {
-          hasSome: [AppCategories.conferencing, AppCategories.video],
+          hasSome: [AppCategories.conferencing, AppCategories.video, AppCategories.payment],
```
alannnc, line 230: While saving any payment app since payment category wasn't included here it will save default stripe app metadata.

alannnc: It can be confusing but I think eventType.price and currency should only live there and not on metadata. I can confirm that both value should have the correct one.

packages/app-store/_utils/getEventTypeAppData.ts
```diff
@@ -18,12 +18,14 @@ export const getEventTypeAppData = <T extends EventTypeAppsList>(
     return allowDataGet
       ? {
           ...appMetadata,
+          // We should favor eventType's price and currency over appMetadata's price and currency
+          price: eventType.price || appMetadata.price,
+          currency: eventType.currency || appMetadata.currency,
```
alannnc, line 23: Since adding an app to an event type was saving default wrong data we should favour eventType price and currency and not from metadata.

apps/web/pages/booking/[uid].tsx
```diff
@@ -480,6 +480,21 @@ export default function Success(props: SuccessProps) {
...
+                            : t("payment")}
+                        </div>
+                        <div className="col-span-2 mb-2 mt-3">
+                          {new Intl.NumberFormat(i18n.language, {
+                            style: "currency",
+                            currency: props.paymentStatus.currency,
+                          }).format(props.paymentStatus.amount / 100.0)}
+                        </div>
+                      </>
+                    )}
```
Comment (joeauyeung) [inline apps/web/pages/booking/[uid].tsx:497]: Adds the payment information on the booking page
CleanShot 2023-08-15 at 23 28 55

joeauyeung (approved):
This LGTM but two points

- @alannnc I added the pricing to the booking page. Let me know what you think in this commit https://github.com/calcom/cal.com/pull/10792/commits/070893b6ab1222be6ea5f479ad7e4fd72651deea
- I thought moving forward we wanted to get all event type payment information in the metadata because of `getPaymentAppData`. This PR fixes the problem but we should consider this.

Udit-takkar (approved):
LGTM

apps/web/pages/booking/[uid].tsx
```diff
@@ -480,6 +480,21 @@ export default function Success(props: SuccessProps) {
...
+                        <div className="mt-3 font-medium">
+                          {props.paymentStatus.paymentOption === "HOLD"
+                            ? t("complete_your_booking")
+                            : t("payment")}
+                        </div>
+                        <div className="col-span-2 mb-2 mt-3">
+                          {new Intl.NumberFormat(i18n.language, {
+                            style: "currency",
+                            currency: props.paymentStatus.currency,
+                          }).format(props.paymentStatus.amount / 100.0)}
```
Comment (Abdel-Monaam-Aouini) [inline apps/web/pages/booking/[uid].tsx:494]: can you make it a function with amount props :)

PR #10808, opened 2023-08-16
perf: Create a seperate component for dependencies of BookEventForm

emrysal:
## What does this PR do?

Currently when loading `BookFormAsModal` we add the following deps already, which are unneeded for rendering until actually displaying the BookForm.

If we extract this out as follows that cuts down on rendering.

```typescript
  const { t } = useLocale();
  const selectedTimeslot = useBookerStore((state) => state.selectedTimeslot);
  const selectedDuration = useBookerStore((state) => state.selectedDuration);
  const { data } = useEvent();
  const parsedSelectedTimeslot = dayjs(selectedTimeslot);
  const { timeFormat, timezone } = useTimePreferences();
```


sean-brydon (approved):
LGTM nice change

PR #10830, opened 2023-08-18
feat: add dependabot.yml to fetch new versiosn of dependencies

Abdel-Monaam-Aouini:


joeauyeung:
@keithwillcode @zomars @emrysal Is this something we want to introduce into our pipeline?

PR #10836, opened 2023-08-18
chore: add dateRangePicker in storybook (CALCOM-10760)

gitstart-calcom:
## DEMO
https://www.loom.com/share/4ac79c06acae41c59c53ec7fae73f56e

We don't use className because this property doesn't exist in dateRangePickerComponet, when this component is executed in different files they pass classNames related to the context with tailwind.

props:

image


PeerRich:
this doesnt look like the right date picker

gitstart-calcom:
Hi @PeerRich, should we look for another dateRangePicker component or update the classNames to this current tested component? We looked for another dateRangePicker in Calcom and didn't find it
image

PeerRich:
incorrect DatePicker

PR #10840, opened 2023-08-18
feat: existing typos, add `cspell` in pre-commit to avoid future typos

xnivaxhzne:
## What does this PR do?

Current version of the code have typos which counts to around 120 unique word. I feel a well maintained project should not have typos. So, I have fixed them. 

Also to void future typos, I have added a config setup in pre-commit hook using the [cspell](https://www.npmjs.com/package/cspell) package. This will prevent users from committing code with unintentional mis spellings. 

Created a [file](https://github.com/kavinkuma6/cal.com/blob/add-cspell-pre-commit-to-fix-typos/.cspell/projectWords.txt), which contains the project specific words or words not found dictionary. This can be used by the package. In future, incase if the user wants to add a non dictionary word to the code,  the same should be added here. I have added a terminal instruction for the same [here](https://github.com/kavinkuma6/cal.com/blob/add-cspell-pre-commit-to-fix-typos/.cspell/cspell.sh#L7) which would be shown in pre commit hook.

Incase, if we want to check the spelling for the whole project, unlike only the staged files as mentioned above, we can use the newly added script `npm run spell-check`

## Requirement/Documentation

## Type of change

## How should this be tested?

- Are there environment variables that should be set? - No
- What are the minimal test data to have? - NA
- What is expected (happy path) to have (input and output)? - NA
- Any other important info that could help to test that PR: I agree we have lot of variable name changes. But we can do most of the testing in the PR review process itself. Then we can use the other test setups to verify the overall workings. 

## Mandatory Tasks


apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -347,6 +347,7 @@ export const EventSetupTab = (
             ) && (
               <div className="text-default flex text-sm">
                 <Check className="mr-1.5 mt-0.5 h-2 w-2.5" />
+                {/* cspell:ignore requres */}
                 <Trans i18nKey="event_type_requres_google_cal">
```
kavinkuma6, line 351: I dont find `event_type_requres_google_cal` defined anywhere. So ignored cspell temporarily

apps/web/public/static/locales/es/common.json
```diff
@@ -1623,7 +1623,7 @@
   "teams_plan_required": "Se requiere un plan de Equipos",
   "routing_forms_are_a_great_way": "Los formularios de enrutamiento son una excelente manera de dirigir sus clientes potenciales entrantes a la persona adecuada. Actualice a un plan de Equipos para acceder a esta función.",
   "choose_a_license": "Elija una licencia",
-  "choose_license_description": "Cal.com viene con una licencia AGPLv3 accesible y gratuita, la cual tiene algunas limitaciones. Esta licencia se puede actualizar a una licencia Empresarial en cualquier momento. Puede actualizarla más tarde.",
```
kavinkuma6, line 1626: Hope this a typo.

packages/features/ee/organizations/pages/settings/profile.tsx
```diff
@@ -57,7 +57,7 @@ const OrgProfileView = () => {
     },
     async onSuccess() {
       await utils.viewer.teams.get.invalidate();
-      showToast(t("your_organisation_updated_sucessfully"), "success");
+      showToast(t("your_organisation_updated_sucessfully"), "success"); // cspell:ignore sucessfully
```
kavinkuma6, line 60: I dont find the declaration anywhere. So, for safety ignored it.

packages/features/ee/support/components/HelpMenuItem.tsx
```diff
@@ -186,6 +186,7 @@ export default function HelpMenuItem({ onHelpItemSelect }: HelpMenuItemProps) {
       <div className="text-subtle bg-muted w-full p-5">
         <p className="">{t("specific_issue")}</p>
         <button
+          // cspell:ignore defualt
```
kavinkuma6, line 189: In some places, this `text-defualt` class is used. Can any one confirm whether it is intentional or bug?

For now, added `ignore` for safety.

packages/types/ical.d.ts
```diff
@@ -75,7 +75,7 @@ declare module "ical.js" {
 
     public constructor(
       component?: Component | null,
-      options?: { strictExceptions: boolean; exepctions: Array<Component | Event> }
+      options?: { strictExceptions: boolean; exceptions: Array<Component | Event> }
```
kavinkuma6, line 78: I dont see this prop used. So corrected it. Pls confirm.

zomars:
Hi @kavinkuma6 Thank you for your contribution 🙏 I love the fact the we're actually fixing typos. I got some questions about the proposed workflow tho. Does this means that a spell check can prevent a commit being made? Doesn't it feels too obtrusive to a regular dev workflow?

kavinkuma6:
> Hi @kavinkuma6 Thank you for your contribution 🙏 I love the fact the we're actually fixing typos. I got some questions about the proposed workflow tho. Does this means that a spell check can prevent a commit being made? Doesn't it feels too obtrusive to a regular dev workflow?

@zomars 
 when a commit is being made, it will run the spell check only on the staged files. This will get executed after the current lint check implementation on the staged files. Yes, it will prevent the commit similar to how lint check prevents a commit.

Since it runs only on the staged files, it won't be too obtrusive. Also, it won't take much time than the time taken by the lint check on the staged files. 

This is my input. But you can also run it once and share your feedback.

PeerRich (changes requested):
imo this PR is quite intruisive, touches a ton of files and I am not sure if we want an automated spell checker at this point

PR #10843, opened 2023-08-18
perf: Cache translations per language/per release

### LINKED ISSUE #10198: [CAL-2192] RFC: Make the i18n serving endpoint "cache-able"

Relevant to #9840

Right now is impossible to cache this endpoint since every single user query the same endpoint with the same input:

```
GET /api/trpc/viewer.public.i18n?input={"json":null} 
```

And this same endpoint MIGHT return: english, spanish, etc...

This impedes the use of caching, if we try to cache this endpoint we will see the mixed-languages bug that we saw a while ago since from the server point of view we're all querying for the same data.

I think the first quick win we can do here is to make the client request the actual locale from the query itself.

If we update the endpoint so we explicilty ask for the locale then we can start caching again since all these example will always return the same language:

```
GET /api/trpc/viewer.public.i18n?input={"json":{"locale":"es"}} 
GET /api/trpc/viewer.public.i18n?input={"json":{"locale":"en"}}
GET /api/trpc/viewer.public.i18n?input={"json":{"locale":"de"}} 
```

This will allow us to cache more aggressively.

There have been other suggestions including caching the returned locales using local storage and preventing a call as done here #9989. But IMO this is introducing other blind spots where we will need to introduce revalidating logic (which browsers already handle)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2192](https://linear.app/calcom/issue/CAL-2192/rfc-make-the-i18n-serving-endpoint-cache-able)</sub>

keithwillcode:
## What does this PR do?

Fixes #10198

## Type of change

## How should this be tested?

- Changing language saves the setting and updates the interface to that new language.
- Coming to the site as a logged out user uses the language preference from your machine/browser
- The requested URL for i18n always includes the locale and the version number (if it doesn't include the locale and the version number, we risk caching data for too long and missing new translations upon upgrade)

## Mandatory Tasks


apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -2,26 +2,28 @@ import { useTranslation } from "next-i18next";
 import { useEffect } from "react";
 
 import { trpc } from "@calcom/trpc/react";
+import { CalComVersion } from "@calcom/ui/components/credits/Credits";
```
keithwillcode, line 5: Ideally this goes into a different file than its current component. Then, it can be reused across both components that need it.

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -58,30 +52,26 @@ const GeneralQueryView = () => {
...
 
   const mutation = trpc.viewer.updateProfile.useMutation({
-    onSuccess: async () => {
-      // Invalidate our previous i18n cache
-      await utils.viewer.public.i18n.invalidate();
+    onSuccess: async (res) => {
+      await utils.viewer.me.invalidate();
       reset(getValues());
       showToast(t("settings_updated_successfully"), "success");
+      update(res);
```
zomars, line 68: This was causing the locale to not change when updating your profile

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -58,30 +52,26 @@ const GeneralQueryView = () => {
...
     },
     onError: () => {
       showToast(t("error_updating_settings"), "error");
     },
     onSettled: async () => {
-      await utils.viewer.public.i18n.invalidate();
+      await utils.viewer.me.invalidate();
```
zomars, line 74: No need to invalidate the i18n endpoint anymore. We just refetch the updated user

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -1,27 +1,33 @@
...
-    /**
-     * i18n should never be clubbed with other queries, so that it's caching can be managed independently.
-     * We intend to not cache i18n query
-     **/
-    trpc: {
-      context: { skipBatch: true },
-    },
-  });
+// eslint-disable-next-line turbo/no-undeclared-env-vars
+const vercelCommitHash = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA || "NA";
```
zomars, line 8: The commit hash should be enough for now. WDYT @keithwillcode

zomars (approved):
I think this is in a merge-able state. Nice work @keithwillcode 🙏🏽

zomars (changes requested):
Spoke too soon. Blocking to push some fixes.

apps/web/components/security/DisableUserImpersonation.tsx
```diff
@@ -12,9 +12,6 @@ const DisableUserImpersonation = ({ disableImpersonation }: { disableImpersonati
       showToast(t("your_user_profile_updated_successfully"), "success");
       await utils.viewer.me.invalidate();
     },
-    async onSettled() {
-      await utils.viewer.public.i18n.invalidate();
```
zomars, line 16: No need to invalidate now

packages/trpc/server/middlewares/localeMiddleware.ts
zomars, line 1: Unused now

zomars (approved):
All green now

Ship it

apps/web/components/I18nLanguageHandler.tsx
keithwillcode: Yeah that works too!

keithwillcode: @zomars thinking about this more; if someone is self-hosting and not on Vercel, this would request the i18n endpoint with NA, which means the next time they upgrade, the new translations wouldn’t be pulled by the browser because the request URL would be the same.

shivamklr (commented):
Tested, lgtm

apps/web/components/I18nLanguageHandler.tsx
leonardocustodio: I was having this exact problem, when running locally when I changed the translation files everything would work fine. But when deployed to remote the translation would always be the old ones. Using an env var of commit hash is far from being a good approach. Especially if it is not written anywhere about it.

leonardocustodio: I believe I'm having an issue related with this, when I deploy the changes to production the translations are never updated. Could it be the reason?

PR #10850, opened 2023-08-19
fix: client locale inference

zomars:
## What does this PR do?

Follow up to #10843 

- Had to add 1KB to the bundle to parse the locale correctly from the client 

image

## Requirement/Documentation

## Type of change

## How should this be tested?

- On incognito windows, public pages should match the browser locale

## Mandatory Tasks


apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -11,6 +11,7 @@ export function useViewerI18n(locale: string) {
   return trpc.viewer.public.i18n.useQuery(
     { locale, CalComVersion: vercelCommitHash },
     {
+      initialData: { locale: "en", i18n: {} },
```
zomars, line 14: Instead of doing

```
useViewerI18n().data ?? {
    locale: "en",
  };
```

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -21,13 +22,27 @@ export function useViewerI18n(locale: string) {
...
 
+function useClientLocale() {
+  const session = useSession();
+  // If the user is logged in, use their locale
+  if (session.data?.user.locale) return session.data.user.locale;
+  // If the user is not logged in, use the browser locale
+  if (typeof window !== "undefined") return window.navigator.language;
+  // If the browser is not available, use English
+  return "en";
+}
```
zomars, line 33: Hope is clear enough

apps/web/lib/app-providers.tsx
```diff
@@ -69,11 +69,7 @@ const CustomI18nextProvider = (props: AppPropsWithoutNonce) => {
   /**
    * i18n should never be clubbed with other queries, so that it's caching can be managed independently.
    **/
-  const session = useSession();
```
zomars, line 72: useClientViewerI18n already checks for session

apps/web/lib/core/i18n/i18n.utils.ts
zomars, line 1: Not used anymore

apps/web/server/lib/ssr.ts
```diff
@@ -2,7 +2,7 @@ import type { GetServerSidePropsContext } from "next";
 import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import superjson from "superjson";
 
-import { getLocaleFromHeaders } from "@calcom/lib/i18n";
+import { getLocaleFromRequest } from "@calcom/lib/i18n";
```
zomars, line 5: Renamed to better reflect new behavior

packages/lib/i18n.ts
```diff
@@ -1,15 +1,22 @@
...
 import type { Maybe } from "@calcom/trpc/server";
 
 const { i18n } = require("@calcom/config/next-i18next.config");
 
-export function getLocaleFromHeaders(req: IncomingMessage): string {
+export async function getLocaleFromRequest(
+  req: NextApiRequest | GetServerSidePropsContext["req"]
+): Promise<string> {
+  const session = await getServerSession({ req });
+  if (session?.user?.locale) return session.user.locale;
```
zomars, line 14: Session takes preference

packages/lib/i18n.ts
```diff
@@ -1,15 +1,22 @@
...
   let preferredLocale: string | null | undefined;
   if (req.headers["accept-language"]) {
-    preferredLocale = parser.pick(i18n.locales, req.headers["accept-language"]) as Maybe<string>;
+    preferredLocale = parser.pick(i18n.locales, req.headers["accept-language"], {
+      loose: true,
```
zomars, line 18: This transforms `es-MX` to `es`. It had issues in my Safari browser.

packages/trpc/server/createContext.ts
```diff
@@ -62,7 +62,7 @@ export async function createContextInner(opts: CreateInnerContextOptions) {
  * @link https://trpc.io/docs/context
  */
 export const createContext = async ({ req, res }: CreateContextOptions, sessionGetter?: GetSessionFn) => {
-  const locale = getLocaleFromHeaders(req);
+  const locale = await getLocaleFromRequest(req);
```
zomars, line 65: Is now async due to session checks. It should not add too much overhead since sessions are cached.

packages/trpc/server/routers/publicViewer/i18n.schema.ts
```diff
@@ -1,7 +1,11 @@
+import parser from "accept-language-parser";
 import { z } from "zod";
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { i18n } = require("@calcom/config/next-i18next.config");
+
 export const i18nInputSchema = z.object({
-  locale: z.string(),
+  locale: z.string().transform((locale) => parser.pick(i18n.locales, locale, { loose: true })),
```
zomars, line 8: This allows to send `es-MX`, `es-419`, etc and transform it to `es`, etc.

playwright.config.ts
```diff
@@ -101,6 +102,7 @@ const config: PlaywrightTestConfig = {
       },
       use: {
         ...devices["Desktop Chrome"],
+        locale: "en-US",
```
zomars, line 105: My system is in Spanish, added this so tests are consistent regardless system locale.

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -1,16 +1,20 @@
...
 import { useSession } from "next-auth/react";
 import { useTranslation } from "next-i18next";
 import { useEffect } from "react";
 
+import { locales } from "@calcom/lib/i18n";
 import { trpc } from "@calcom/trpc/react";
 
 // eslint-disable-next-line turbo/no-undeclared-env-vars
-const vercelCommitHash = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA || "NA";
+const CalComVersion = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA || "NA";
```
keithwillcode, line 10: @zomars i think we need to use the version from package.json here actually. 

If a self-hoster not running on Vercel runs this, it will only be able to call the i18n endpoint with the NA value, which means it will indefinitely cache what it gets back. Next time they upgrade, no new translations would be shown because they browser would still have the file cached

keithwillcode (changes requested):
Nice changes. Just think we need to fix the version part before merging.

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -1,16 +1,20 @@
...
 
-export function useViewerI18n(locale: string) {
+function useViewerI18n(locale: string) {
   return trpc.viewer.public.i18n.useQuery(
-    { locale, CalComVersion: vercelCommitHash },
+    { locale, CalComVersion },
     {
+      initialData: { locale: null, i18n: {} },
+      refetchOnMount: "always",
```
hariombalhara, line 17: I am not sure why we need this. As it's tied to commit hash, shouldn't we just fetch it once?

apps/web/components/I18nLanguageHandler.tsx
```diff
@@ -21,13 +25,30 @@ export function useViewerI18n(locale: string) {
   );
 }
 
+function useClientLocale() {
+  const session = useSession();
+  // If the user is logged in, use their locale
+  if (session.data?.user.locale) return session.data.user.locale;
+  // If the user is not logged in, use the browser locale
+  if (typeof window !== "undefined") {
+    const browserLocale = parser.pick(locales, window.navigator.language, { loose: true });
```
hariombalhara, line 34: Why do we need to parse it client side? We are already handling parser pick on server side. Maybe I am missing something?

apps/web/server/lib/ssr.ts
```diff
@@ -19,23 +19,24 @@ const CalComVersion = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA || "NA";
...
 
-  // always preload "viewer.public.i18n"
-  await ssr.viewer.public.i18n.fetch({ locale, CalComVersion });
-  // So feature flags are available on first render
-  await ssr.viewer.features.map.prefetch();
-  // Provides a better UX to the users who have already upgraded.
-  await ssr.viewer.teams.hasTeamPlan.prefetch();
-
-  await ssr.viewer.public.session.prefetch();
+  await Promise.allSettled([
```
hariombalhara, line 31: TIL about allSettled

hariombalhara (changes requested):
I have an important concern about the usage of parser client side. Maybe we can avoid that as I commented.

apps/web/components/I18nLanguageHandler.tsx
zomars: This is due using initialData so data is never undefined. But using it will prevent the initial fetch being made. Hence the referchObMount

apps/web/components/I18nLanguageHandler.tsx
zomars: To prevent duplicate caches. Without this we were generating caches for es, es-ES, es-MX, etc. when they all respond with the same data. Third ensures the initial request is always sent as 'es'. Also without this locales didn't load in time causing tests to fail due to revalidation.

apps/web/components/I18nLanguageHandler.tsx
zomars: Yeah.  I was trying to come up with a way to use the version without being too intrusive

hariombalhara (dismissed):
LGTM

keithwillcode (approved):
Amazing. Thanks so much for pushing this through.

shivamklr (commented):
Tested, lgtm

PR #10853, opened 2023-08-19
perf: GCal reduce calendar payload

joeauyeung:
## What does this PR do?

Explored reducing the payload size from GCal queries.

Reducing the freebusy query had no difference in payload size.

Reducing the calendar query payload of 4 calendars, the size went from 1895 to 434 bytes.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Connect a GCal account
- The list of calendars should be listed on the calendar page
- On the booking page, any events on the GCal should be blocked

## Mandatory Tasks

## Checklist


hariombalhara (approved):
Nice optimization @joeauyeung !! 

As a followup, I was thinking how we can prevent having a scenario where we are accessing a field that we didn't fetch. I don't think google calendar types prevent that from happening.

Maybe adding a zod schema would help . It doesn't seem necessary right now as the fields are accessed almost below the place where they are fetched. But it can be problematic if they are accessed somewhere else e.g. in a separate function.

Udit-takkar (approved):

PR #10860, opened 2023-08-21
fix: Make identifier conform to RHF field requirement

### LINKED ISSUE #9540: [CAL-1956] Identifier while creating a Booking Field has to be slugified
See what happens if it has certain special characters - [https://www.loom.com/share/cb23fc26985a48c9bb903736c054631f?sid=af9206c8-6e86-4d0e-a267-afd0c9f176d5](https://www.loom.com/share/cb23fc26985a48c9bb903736c054631f?sid=af9206c8-6e86-4d0e-a267-afd0c9f176d5)
The problem is solved if we simply slugify what is entered by the user in `Identifier` field.

So if user enters 'Please enter your name' we will automatically convert it to 'please-enter-your-name'

We need to do it at 2 places

* When user enters some text in it and focusses away(by either clicking on Submit button or focussing on another field), we can slugify the text
* We should also add a zod transform on identifier(if zod transformer not possible, we can manually call slugify(from packages/lib/slugify) on it before saving.
* We have to cater POST, PATCH API endpoint as well for event-type

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1956](https://linear.app/calcom/issue/CAL-1956/identifier-while-creating-a-booking-field-has-to-be-slugified)</sub>

hariombalhara:
## What does this PR do?

Form Builder identifiers would be automatically transformed to RHF compatible field names as you type. [Demo](https://www.loom.com/share/972e350955cb4085a641246fa2005327)
	- Any existing field identifiers would be transformed when that EventType is re-saved from Advanced Tab. [Demo](https://www.loom.com/share/b577582779fe458cbd68b892921df2f1)

Added unit tests as well. Added them for `slugify` too as the change was related to that and I possibly found a bug there

Fixes #9540

## Type of change

## How should this be tested?

## Mandatory Tasks

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,6 +7,7 @@ public
 
 *.lock
 *.log
+*.test.ts
 
 .gitignore
 .npmignore
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,7 +7,6 @@ public
 
 *.lock
 *.log
-*.test.ts
 
 .gitignore
 .npmignore
--- a/packages/features/bookings/lib/getBookingFields.ts
+++ b/packages/features/bookings/lib/getBookingFields.ts
@@ -3,6 +3,7 @@ import type { z } from "zod";
 
 import { SMS_REMINDER_NUMBER_FIELD } from "@calcom/features/bookings/lib/SystemField";
 import { fieldsThatSupportLabelAsSafeHtml } from "@calcom/features/form-builder/fieldsThatSupportLabelAsSafeHtml";
+import { getFieldIdentifier } from "@calcom/features/form-builder/utils/getFieldIdentifier";
 import { markdownToSafeHTML } from "@calcom/lib/markdownToSafeHTML";
 import slugify from "@calcom/lib/slugify";
 import { EventTypeCustomInputType } from "@calcom/prisma/enums";
@@ -285,7 +286,9 @@ export const ensureBookingInputsHaveSystemFields = ({
 
   const missingSystemBeforeFields = [];
   for (const field of systemBeforeFields) {
-    const existingBookingFieldIndex = bookingFields.findIndex((f) => f.name === field.name);
+    const existingBookingFieldIndex = bookingFields.findIndex(
+      (f) => getFieldIdentifier(f.name) === getFieldIdentifier(field.name)
+    );
     // Only do a push, we must not update existing system fields as user could have modified any property in it,
     if (existingBookingFieldIndex === -1) {
       missingSystemBeforeFields.push(field);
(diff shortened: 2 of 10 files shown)


.prettierignore
```diff
@@ -7,7 +7,6 @@ public
 
 *.lock
 *.log
-*.test.ts
```
hariombalhara, line 10: Screenshot 2023-08-21 at 6 38 42 PM
Was setting these lint errors in test files. So, I am removing test files from `.prettierignore`. If we discover any issues due to that, we should fix the root cause. cc @leog

packages/lib/slugify.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  // So that user can freely add spaces and any other character iteratively and it get's converted to - later on.
+  it("should remove dashes from start and end.", () => {
+    expect(slugify("hello-there-")).toEqual("hello-there");
+    expect(slugify("hello-there_")).toEqual("hello-there");
+    expect(slugify("_hello-there_")).toEqual("hello-there");
+    expect(slugify("$hello-there_")).toEqual("hello-there");
+  });
+
+  // This is failing, if we want to fix it, one approach is as used in getValidRhfFieldName
+  it.skip("should remove unicode and emoji characters", () => {
```
hariombalhara, line 32: I think this should work with `slugify`. So added it for now as skipped.

packages/features/form-builder/schema.ts
```diff
@@ -31,7 +33,7 @@ export const EditableSchema = z.enum([
 ]);
 
 const baseFieldSchema = z.object({
-  name: z.string(),
+  name: z.string().transform(getValidRhfFieldName),
```
hariombalhara, line 36: It would allow the transformation to happen in both API and if someone just intercepts the request at client and modifies the identifier.

packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -285,7 +286,9 @@ export const ensureBookingInputsHaveSystemFields = ({
 
   const missingSystemBeforeFields = [];
   for (const field of systemBeforeFields) {
-    const existingBookingFieldIndex = bookingFields.findIndex((f) => f.name === field.name);
+    const existingBookingFieldIndex = bookingFields.findIndex(
+      (f) => getFieldIdentifier(f.name) === getFieldIdentifier(field.name)
```
hariombalhara, line 290: Do an apples to apples comparison.

.prettierignore
hariombalhara: Had to revert the .prettierignore change because it would require fixing prettier errors in many test files. Job for a different PR.

zomars (commented):
Code LGTM let's wait for the tests

Udit-takkar (dismissed):
LGTM. Works fine

Udit-takkar (approved):
Fixed tests. approving again

CarinaWolli (approved):

hariombalhara:
Thanks a lot @Udit-takkar 🙏🙏

PR #10875, opened 2023-08-22
feat: admin org list now shows all, dns check added

leog:
## What does this PR do?

Administrator Organization list now shows all organizations, verified or not, and a DNS check was added to flag which organizations successfully registered a subdomain in a Vercel hosting environment or which were not configured at all, sending an email notifications to the Cal.com instance admins.

Here is the list of orgs for admins:
<kbd></kbd>

Verification actions:
<kbd></kbd>

Verification actions plus mark DNS as configured:
<kbd></kbd>

Email sent to instance admins:
<kbd></kbd>

## Type of change

## How should this be tested?

Create a bunch of orgs in a non-Vercel environment (such as localhost) and see in the Organizations Admin Settings all the created orgs. Verify some of them and you will still see them in the list without the badge Unverified. Mark some of them as DNS configured and see them in the list without the DNS missing badge.


apps/web/pages/settings/admin/organizations/index.tsx
```diff
@@ -1,9 +1,9 @@
...
+import AdminOrgsPage from "@calcom/features/ee/organizations/pages/settings/admin/AdminOrgPage";
 
 import type { CalPageWrapper } from "@components/PageWrapper";
 import PageWrapper from "@components/PageWrapper";
 
-const Page = UnverifiedOrgsPage as CalPageWrapper;
+const Page = AdminOrgsPage as CalPageWrapper;
 Page.PageWrapper = PageWrapper;
 
 export default Page;
```
leog, line 9: Changing name of component as it lists all orgs now

packages/emails/email-manager.ts
```diff
@@ -376,3 +378,7 @@ export const sendDailyVideoRecordingEmails = async (calEvent: CalendarEvent, dow
 export const sendOrganizationEmailVerification = async (sendOrgInput: OrganizationEmailVerify) => {
   await sendEmail(() => new OrganizationEmailVerification(sendOrgInput));
 };
+
+export const sendAdminOrganizationNotification = async (input: OrganizationNotification) => {
+  await sendEmail(() => new AdminOrganizationNotification(input));
+};
```
leog, line 384: This is the email to send the admins about an unconfigured organization in terms of subdomain and DNS

packages/prisma/zod-utils.ts
```diff
@@ -328,6 +328,7 @@ export const teamMetadataSchema = z
     subscriptionItemId: z.string().nullable(),
     isOrganization: z.boolean().nullable(),
     isOrganizationVerified: z.boolean().nullable(),
+    isOrganizationConfigured: z.boolean().nullable(),
```
leog, line 331: Whoops, added a new property here @keithwillcode. I will surely revise it after today's Engineering Round Table

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -43,12 +39,10 @@ const vercelCreateDomain = async (domain: string) => {
...
 
   // Domain is already owned by another team but you can request delegation to access it
-  if (data.error?.code === "forbidden")
-    throw new TRPCError({ code: "CONFLICT", message: "domain_taken_team" });
+  if (data.error?.code === "forbidden") return false;
 
   // Domain is already being used by a different project
-  if (data.error?.code === "domain_taken")
-    throw new TRPCError({ code: "CONFLICT", message: "domain_taken_project" });
+  if (data.error?.code === "domain_taken") return false;
```
leog, line 45: We don't want to fail creating the org because there was an error with Vercel the same way we want to support other providers which is why a custom DNS missing check was added, in order to let the admins keep track which orgs are effectively well configured.

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -85,9 +79,32 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
 
   const t = await getTranslation(ctx.user.locale ?? "en", "common");
   const availability = getAvailabilityFromSchedule(DEFAULT_SCHEDULE);
+  let isOrganizationConfigured = false;
 
   if (check === false) {
-    if (IS_CALCOM) await vercelCreateDomain(slug);
+    // eslint-disable-next-line turbo/no-undeclared-env-vars
+    if (process.env.VERCEL) {
```
leog, line 86: This env var exists in Vercel platform to detect it's a Vercel environment

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -85,9 +79,32 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
...
+      });
+      if (instanceAdmins.length) {
+        await sendAdminOrganizationNotification({
+          instanceAdmins,
+          orgSlug: slug,
+          ownerEmail: adminEmail,
+          t,
+        });
+      } else {
+        console.warn("Organization created: subdomain not configured and couldn't notify adminnistrators");
```
leog, line 105: No Vercel, no email sent to admins, so we warn in console and we mark the org as DNS missing until admins mark it as configured from settings

packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -20,11 +21,13 @@ type UpdateOptions = {
 
 export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
   // A user can only have one org so we pass in their currentOrgId here
-  const currentOrgId = ctx.user?.organization?.id;
+  const currentOrgId = ctx.user?.organization?.id || input.id;
```
leog, line 24: Needed to pass down an org Id instead of inferring from user orgId as an admin doesn't have an org assigned.

packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -20,11 +21,13 @@ type UpdateOptions = {
...
 
-  if (!currentOrgId) throw new TRPCError({ code: "UNAUTHORIZED" });
+  if (!currentOrgId || ctx.user.role !== UserPermissionRole.ADMIN)
```
leog, line 26: Checking current user role to include admin updates to org

packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -20,11 +21,13 @@ type UpdateOptions = {
...
+    throw new TRPCError({ code: "UNAUTHORIZED" });
 
-  if (!(await isOrganisationAdmin(ctx.user?.id, currentOrgId))) throw new TRPCError({ code: "UNAUTHORIZED" });
+  if (!(await isOrganisationAdmin(ctx.user?.id, currentOrgId)) || ctx.user.role !== UserPermissionRole.ADMIN)
```
leog, line 29: Checking current user role to include admin updates to org

packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -64,6 +68,10 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
     timeZone: input.timeZone,
     weekStart: input.weekStart,
     timeFormat: input.timeFormat,
+    metadata: {
+      ...prevMetadata,
+      ...input.metadata,
+    },
```
leog, line 74: Taking into account passed down metadata letting it overwrite existing values

packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -80,13 +88,11 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
-      if (metadataParse.success) {
-        const { requestedSlug: _, ...cleanMetadata } = metadataParse.data || {};
-        data.metadata = {
-          ...cleanMetadata,
-        };
-      }
+      const { requestedSlug: _, ...cleanMetadata } = prevMetadata || {};
+      data.metadata = {
+        ...cleanMetadata,
+        ...input.metadata,
```
leog, line 94: Taking into account passed down metadata letting it overwrite existing values

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -85,9 +79,32 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
...
+      // within a Vercel context
+      isOrganizationConfigured = await vercelCreateDomain(slug);
+    } else {
+      // Otherwise, we proceed to send an administrative email to admins regarding
+      // the need to configure DNS registry to support the newly created org
+      const instanceAdmins = await prisma.user.findMany({
+        where: { role: UserPermissionRole.ADMIN },
+        select: { email: true },
+      });
+      if (instanceAdmins.length) {
```
sean-brydon, line 97: NIT: Couldn't this techincally be 0 admins

leog: Yes it could, there's an else for that case, and there's nothing we can do other than mark the org as DNS Missing so admins can see something pending on the admin org list

packages/features/ee/organizations/pages/settings/admin/AdminOrgPage.tsx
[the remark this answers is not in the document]
leog: I think we will end up working with the latest available components for v2, this is a follow up on v1

packages/trpc/server/routers/viewer/organizations/update.schema.ts
```diff
@@ -22,6 +25,7 @@ export const ZUpdateInputSchema = z.object({
   timeZone: z.string().optional(),
   weekStart: z.string().optional(),
   timeFormat: z.number().optional(),
+  metadata: teamMetadataSchema.optional(),
```
zomars, line 28: Should we allow every single key to be update-able from the update mutation? One thing is to have a schema for reading the metadata. But IMO we should have another schema so not all keys are editable from here.

packages/trpc/server/routers/viewer/organizations/update.handler.ts
zomars: We have the same issue here: It will overwrite the metadata when setting requestedSlug:

https://github.com/calcom/cal.com/blob/376f023e0fd69c0f30569152785b56c277a32bc8/packages/trpc/server/routers/viewer/organizations/update.handler.ts#L76-L78

packages/lib/getMetadataHelpers.ts
```diff
@@ -0,0 +1,32 @@
...
+ * The intention behind these helpers is to make it easier to work with metadata.
+ * @param schema This is the zod schema that you want to use to parse the metadata
+ * @param rawMetadata This is the metadata that you want to parse
+ * @returns An object with the parsed metadata, a get function to get a specific key of the metadata, and a mergeMetadata function to merge new metadata with the old one
+ * @example
+ * const { mergeMetadata } = getMetadataHelpers(teamMetadataSchema, team.metadata);
+ * const newMetadata = mergeMetadata({ someKey: "someValue" });
+ * prisma.team.update({ ..., data: { metadata: newMetadata } });
+ */
+export function getMetadataHelpers<T extends SomeZodObject>(schema: T, rawMetadata: unknown) {
```
zomars, line 15: Would like to get you input on this pattern @leog 🙏🏽

zomars (changes requested):
Left some comments to address before approving 🙏🏽

packages/trpc/server/routers/viewer/organizations/update.schema.ts
```diff
@@ -22,6 +25,7 @@ export const ZUpdateInputSchema = z.object({
   timeZone: z.string().optional(),
   weekStart: z.string().optional(),
   timeFormat: z.number().optional(),
+  metadata: teamMetadataSchema.unwrap().pick({ isOrganizationConfigured: true }),
```
leog, line 28: @zomars here is the restriction to only have `isOrganizationConfigured` be passed down to be updated

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -1,19 +1,21 @@
 import { createHash } from "crypto";
+import { lookup } from "dns";
```
leog, line 2: Added to detect the IP of Cal.com app so it can be passed down to the email

zomars (approved):

PR #10876, opened 2023-08-22
perf: Lazy load framer motion stuff as much as we can

hariombalhara:
## What does this PR do?

It doesn't reduce bundle size significantly but lazy-loads whatever we can in from framer and uses[ strict mode ](https://www.framer.com/motion/lazy-motion/###strict) to ensure that we can't do non-lazy framer motion stuff in there.

Significant improvement will come from https://github.com/calcom/cal.com/issues/10880 but it's not trivial 

## Type of change
- [x] Performance Improvement

## How should this be tested?
- Browse to the booking form and observe the animation

## Mandatory Tasks

## Checklist


zomars (approved):
Everything helps! 🙏🏽

PR #10881, opened 2023-08-22
chore: expect emails e2e

### LINKED ISSUE #10855: [CAL-2366] Running E2E locally yields ECONNREFUSED
```
          @leog @shivamklr I think this is why I get so many errors running e2e locally on production builds.
```

Can we make this optional? Only check for mailhog if it's enabled. I spent quite a while figuring out why all my tests were failing locally. At least we should document and update the README with instructions.

\*Originally posted by @zomars in \*[*https://github.com/calcom/cal.com/pull/10606#discussion_r1299447778*](https://github.com/calcom/cal.com/pull/10606#discussion_r1299447778)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2366](https://linear.app/calcom/issue/CAL-2366/running-e2e-locally-yields-econnrefused)</sub>

shivamklr:
## What does this PR do?

Fixes #10855 
- Made emails fixture optional
- Add env variable to enable mailhog checks
- Refactor code

## Type of change

## How should this be tested?

- Case 1:
    - Set env `E2E_TEST_MAILHOG_ENABLED="1"`
    - Run mailhog container
    - Run e2e tests
- Case 2:
    -  Set env `E2E_TEST_MAILHOG_ENABLED=`
    - Run e2e tests

## Mandatory Tasks


.env.example
```diff
@@ -126,8 +126,9 @@ TWILIO_WHATSAPP_PHONE_NUMBER=
 NEXT_PUBLIC_SENDER_ID=
 TWILIO_VERIFY_SID=
 
-# This is used so we can bypass emails in auth flows for E2E testing
```
shivamklr, line 129: No longer bypasses emails

apps/web/playwright/lib/fixtures.ts
```diff
@@ -71,7 +72,11 @@ export const test = base.extend<Fixtures>({
...
   },
   emails: async ({}, use) => {
-    const mailhogAPI = mailhog();
-    await use(mailhogAPI);
+    if (IS_MAILHOG_ENABLED) {
+      const mailhogAPI = mailhog();
+      await use(mailhogAPI);
+    } else {
+      await use(undefined);
+    }
```
shivamklr, line 80: Optional fixture

shivamklr (commented):
Self Review

.env.example
```diff
@@ -126,8 +126,9 @@ TWILIO_WHATSAPP_PHONE_NUMBER=
...
-# Set it to "1" if you need to run E2E tests locally
+# Set it to "1" if you need to run E2E tests locally. 
+# If you need to run email tests with mailhog locally,
```
keithwillcode, line 130: This comment is out of place now that the value is on line 180

keithwillcode (changes requested):
Nice work! Just one little change I saw to make

nicktrn:
@shivamklr @keithwillcode @zomars 

Not not downplay the effort that has been put into the current email testing framework, but I think the real issue is that an external (unmaintained) third party app is required for successful test runs.

I don't think adding more complexity / env vars, and conditionally disabling email tests is a workable solution.

What happens if a test relies on information from emails to continue, e.g. clicking the CTA link?

With mailhog, the workarounds already start when trying to disable authentication for local testing. Its REST API is rather limited - you will run into more issues requiring unsightly workarounds here too. Extending the API requires patches in Go for the server itself, in addition to patching the JS package used here.

You could have a look at #10560 for an alternative in pure JS. The [SMTP server](https://nodemailer.com/extras/smtp-server/) is automatically started by Playwright, following the same pattern used for Next.

I may be completely off-track but I think it's much more maintainable and extensible. It should also solve other problems that will show up on the current path.

(Haven't rebased recently as I would have to rollback most of the mailhog-related changes.)

keithwillcode:
> @shivamklr @keithwillcode @zomars
> 
> Not not downplay the effort that has been put into the current email testing framework, but I think the real issue is that an external (unmaintained) third party app is required for successful test runs.
> 
> I don't think adding more complexity / env vars, and conditionally disabling email tests is a workable solution.
> 
> What happens if a test relies on information from emails to continue, e.g. clicking the CTA link?
> 
> With mailhog, the workarounds already start when trying to disable authentication for local testing. Its REST API is rather limited - you will run into more issues requiring unsightly workarounds here too. Extending the API requires patches in Go for the server itself, in addition to patching the JS package used here.
> 
> You could have a look at #10560 for an alternative in pure JS. The [SMTP server](https://nodemailer.com/extras/smtp-server/) is automatically started by Playwright, following the same pattern used for Next.
> 
> I may be completely off-track but I think it's much more maintainable and extensible. It should also solve other problems that will show up on the current path.
> 
> (Haven't rebased recently as I would have to rollback most of the mailhog-related changes.)

Hey @nicktrn, appreciate your comment here and the work you put into #10560. At the moment, we aren't going to take on maintaining something built from scratch with the complexity as in #10560. It's a very cool and clever solution, but will be more work to maintain moving forward.

keithwillcode (dismissed):

zomars (dismissed):
Lovely!

nicktrn:
@keithwillcode thanks for having a look and the positive feedback. I certainly understand your point re complexity. It's sad there are no well-established solutions out there yet, bar some online services that come with their own issues. Will let you all get on with the current email strategy. Any major roadblocks, just give me a ping.

keithwillcode (approved):

PR #10882, opened 2023-08-22

### LINKED ISSUE #10789: Double Booking (for requests that require confirmation)

### Issue Summary

If two different users try to book the same slot that requires Confirmation. The admin is able to Confirm both, and this creates a double booking.

### Steps to Reproduce

1. 1st user books a slot (requires confirmation)
2. 2nd user books the same slot (requires confirmation)
3. Admin Confirms first slot booking
4. Admin Confirms 2nd booking for the same slot

### Actual Results

- It is allowing Admin to Confirm the same slot from two different requests

### Expected Results

- It should not allow Admin to confirm a slot, once it has been confirmed 

### Technical details

using hosted app.cal.com

### Evidence
Screneshot
https://www.awesomescreenshot.com/image/42238430?key=bccaaa338af85f1393a0a9bdd56189e6

vijayraghav-io:
## What does this PR do?

Consider the scenario - User/Admin has more than one booking confirmation requests for same time slot. 
Fixes #10789 

API - Check is implemented in 'bookings.confirm' end point. Throws trpc error if slot is not available for the user.

## Requirement/Documentation

PFB snapshots and videos explaining the scenario and the fix implemented.
Scenario - User has two booking requests that require confirmation.
CallCom_1
After user confirms one of the booking, hitting the "Confirm" button on second booking request returns error.

If the user cancels the first confirmed booking, then the second booking can be confirmed successfully. This is also tested.

Video - Before changes:
https://www.loom.com/share/f4c6747ac0144187bf6a9780c9562c88?sid=38741956-2ce1-430e-b7db-fd18a798b2d0

Video - After implementation:
https://www.loom.com/share/4733a402c8ce4711a6fa3518252c26be?sid=4d514a07-6a88-4c9e-a4b2-b2ee1f33b788

## Type of change

## How should this be tested?

Refer the videos above.

## Mandatory Tasks

## Checklist

- Added test case - "Cannot confirm booking for a slot, if another confirmed booking already exists for same slot." in booking-pages.e2e.ts

FYI @mywebmasteruk  @keithwillcode


vijayraghav-io:
Found the reason - mailhog was not running, hence most of the tests were failing.
Now after running with mailhog, still 7 tests are failing, may be due to some integrations like apple-calendar.

Will create new testcases for double booking and update.

vijayraghav-io:
Added test case - "double booking cannot be done for same slot" in booking-pages.e2e.ts
and verified

vijayraghav-io:
@Udit-takkar , required tests added , PR can be reviewed.

vijayraghav-io:
@Udit-takkar , updated the existing testcase "can book an event that requires confirmation and then that booking can be accepted by organizer" as it was failing. Also verified, it is passing now.

apps/web/components/booking/BookingListItem.tsx
```diff
@@ -115,6 +116,38 @@ function BookingListItem(booking: BookingItemProps) {
     return booking.seatsReferences[0].referenceUid;
   };
 
+  // Check user's availability/slots for the current booking request,
+  const isSlotAvailable = () => {
+    const resp = trpc.viewer.public.slots.getSchedule.useQuery(
+      {
```
Udit-takkar, line 122: is there any other way to check this. this API is sometimes very slow on production and can slow this page also if user has many unconfirmed bookings

packages/trpc/server/routers/viewer/bookings/confirm.handler.ts
```diff
@@ -113,6 +116,25 @@ export const confirmHandler = async ({ ctx, input }: ConfirmOptions) => {
...
+  // Check for user's slot availability for current booking start and end time
+  if (confirmed) {
+    const slotsAvailable = await getAvailableSlots({
+      startTime: booking?.startTime.toISOString(),
+      endTime: booking?.endTime.toISOString(),
+      eventTypeId: booking?.eventType?.id,
+      eventTypeSlug: booking?.eventType?.slug,
+      timeZone: user?.timeZone,
+      usernameList: getUsernameList(user?.username ?? ""),
+      isTeamEvent: false,
```
Udit-takkar, line 128: why did you pass false here ?

Udit-takkar (commented):
left some comments above

packages/trpc/server/routers/viewer/bookings/confirm.handler.ts
vijayraghav-io: Yes, updated.
The result would be same for this scenario, even if we default to false, as here we are concerned about single user's or admin's slots only. But anyway updated to be in consistent.

Udit-takkar (dismissed):
@vijayraghav-io ok it's great now that you have removed the query from frontend but it feels very slow while confirming a meeting.
Screenshot 2023-09-09 at 12 43 29 AM

vijayraghav-io:
@Udit-takkar , I checked different scenarios and confirm endpoint response time is < 600ms only.

Screenshot 2023-09-09 at 10 47 46 PM


Screenshot 2023-09-09 at 11 07 56 PM


Screenshot 2023-09-09 at 10 41 01 PM


also compared with main branch not much significant difference is there

zomars:
Thank you for your contribution 🙏 I'm adding a more specific error toast before merging:

image

zomars (approved):
Thanks again! 🙏🏽

zomars:
Hi @vijayraghav-io we had to revert as we found some undesired behavior that needs to be discussed further. Feel free to follow the discussion in the new PR #11595

PR #10885, opened 2023-08-22
fix: Improves the Login button for SAML/OIDC

alishaz-polymath:
## What does this PR do?

The login button for both SAML and OIDC says `Login with SAML`, when it should state Login with SAML/OIDC to include them both.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## Mandatory Tasks


keithwillcode (commented):
Looks good. Should we delete "signin_with_saml"?

alishaz-polymath:
> Looks good. Should we delete "signin_with_saml"?

I left it there just in case we decide to conditionally show `Sign in with SAML` or `Sign in with OIDC` later on, in which case we can then delete the Sign in with OIDC. But we can delete `signin_with_saml` if that's not a route we intend to take down the line.

keithwillcode:
> > Looks good. Should we delete "signin_with_saml"?
> 
> I left it there just in case we decide to conditionally show `Sign in with SAML` or `Sign in with OIDC` later on, in which case we can then delete the Sign in with OIDC. But we can delete `signin_with_saml` if that's not a route we intend to take down the line.

Ah ok cool. Yeah, that could be possible actually. I could see us possibly refactoring to have separate settings for SAML vs OIDC. Right now they are put into the same env vars.

keithwillcode (approved):

PR #10898, opened 2023-08-23
refactor: improvements on german translation

kamildemirci:
## What does this PR do?

Improve the German translation for in person meetings. Currently it's translated as "Persönlich" and "In Person" which is not so common. In German, the booking type is called "Vor-Ort-Termin" and the location description is "vor Ort".

## Requirement/Documentation

none

## Type of change

## How should this be tested?

- Change the language to German
- Set up an in person meeting for a booking type
- Schedule a meeting

## Mandatory Tasks

## Checklist


kamilberke:
An alternative would be to call **both** "Persönlich" or "In Person".

PeerRich (approved):

PR #10928, opened 2023-08-24
feat: Adds filter-by-email on /users/_get endpoint for admin calls

alishaz-polymath:
## What does this PR do?

Adds the ability to filter by user emails in the `/users/_get` endpoint for Admin Calls.

From now on, an Admin can query single or multiple users by email when making a `GET` request for `/users` on the instance:

1. For a single user: Simply attach &email=<user1@email.com>
    Eg. `{URL}/users?apiKey=<api_key>&email=teampro@example.com`
2. For multiple users: Simply keep appending &email=<user-n@email.com>
    Eg. `{URL}/users?apiKey=<api_key>&email=teampro@example.com&email=pro@example.com`

## Type of change


apps/api/pages/api/users/_get.ts
```diff
@@ -39,6 +51,30 @@ export async function getHandler(req: NextApiRequest) {
   const where: Prisma.UserWhereInput = {};
   // If user is not ADMIN, return only his data.
   if (!isAdmin) where.id = userId;
+
+  if (req.query.email) {
+    const validationResult = schemaQuerySingleOrMultipleUserEmails.safeParse(req.query);
```
zomars, line 56: We can simplify by using parse. `defaultResponder` already handles validations errors.
```suggestion
    const validationResult = schemaQuerySingleOrMultipleUserEmails.parse(req.query);
```

alishaz-polymath: Oh Cool! 👍

apps/api/pages/api/users/_get.ts
```diff
@@ -39,6 +51,30 @@ export async function getHandler(req: NextApiRequest) {
   const where: Prisma.UserWhereInput = {};
   // If user is not ADMIN, return only his data.
   if (!isAdmin) where.id = userId;
+
+  if (req.query.email) {
+    const validationResult = schemaQuerySingleOrMultipleUserEmails.parse(req.query);
```
emrysal, line 56: This probably should be `safeParse`

emrysal (changes requested):

emrysal (approved):

apps/api/lib/validations/shared/queryUserEmail.ts
```diff
@@ -10,7 +10,7 @@ export const schemaQueryUserEmail = baseApiParams.extend({
 });
 
 export const schemaQuerySingleOrMultipleUserEmails = z.object({
-  email: z.union([z.string().email(), z.array(z.string().email())]).optional(),
+  email: z.union([z.string().email(), z.array(z.string().email())]),
```
alishaz-polymath, line 13: If we remove optional here, would the endpoint still work if no email is provided?  🤔

emrysal: Then req.email is falsy:
image

alishaz-polymath: Oh yes, good call!

PR #10946, opened 2023-08-25
fix: meeting ended trigger for webhooks and zapier sometimes not working

### LINKED ISSUE #9347: [CAL-1868] Meeting ended trigger for webhooks and zapier not always working
### Issue Summary

Didn't have enough time to investigate this, but I think the trigger only works if the time of when the meeting ends is not far away from when the time the meeting was initially booked. I think it is because of `node-schedule` which is used to schedule a job to trigger the webhooks.

<sub>[CAL-1868](https://linear.app/calcom/issue/CAL-1868/meeting-ended-trigger-for-webhooks-and-zapier-not-always-working)</sub>

montocoder:
## What does this PR do?

This PR addresses the issue of the Zapier and workflow meeting ended trigger not working properly. To fix this issue, we replaced the scheduling approach from using `node-schedule` with a cron instead.

The reason we cannot use `node-schedule` for this use case is that this library is only intended for in-process scheduling. Here is an excerpt from their [documentation](https://www.npmjs.com/package/node-schedule):

> Note that Node Schedule is designed for in-process scheduling, i.e. scheduled jobs will only fire as long as your script is running, and the schedule will disappear when execution completes. If you need to schedule jobs that will persist even when your script isn't running, consider using actual [cron](http://unixhelp.ed.ac.uk/CGI/man-cgi?crontab+5).

Fixes: https://github.com/calcom/cal.com/issues/9347

## Type of change

## How should this be tested?

1. Set up Zapier and a webhook with a meeting ended trigger.
2. Schedule multiple meetings at different times (recommended: add a description with "should be triggered at xx:xx").
3. Wait and see if the triggers are executed at the correct time.
4. Check the database to ensure that finished jobs are cleaned correctly.

## Mandatory Tasks

## Checklist

- I have read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)
- My code does follow the style guidelines of this project
- I commented my code, particularly in hard-to-understand areas
- I have checked if my PR needs changes to the documentation
- I have checked if my changes generate no new warnings
- I have checked if new and existing unit tests pass locally with my changes


packages/features/webhooks/lib/scheduleTrigger.ts
```diff
@@ -9,44 +7,32 @@ export async function scheduleTrigger(
...
       },
       data: {
         scheduledJobs: {
-          push: job.name,
+          push: jobName,
         },
       },
     });
+
+    await prisma.$transaction([createTrigger, updateBooking]);
```
Udit-takkar, line 35: nice. i was about to add transaction here

CarinaWolli (approved):
Nice, code looks good and it works! Thank you for this important fix 🙏

PR #10950, opened 2023-08-25
test: Upcoming bookings sorting order

### LINKED ISSUE #10945: E2E Test: tRPC Route bookings/get
We should do an e2e test here with a scenario of many different type of bookings(Teams, where the user is an attendee, where the user is an organizer, seated bookings)

hariombalhara:
## What does this PR do?
Adds upcoming bookings test that verifies that attendee bookings are also shown and in correct sorted order

Fixes #10945

## Type of change

- [x] Test


keithwillcode (approved):
Awesome work @hariombalhara. Love it

PR #10957, opened 2023-08-25
fix: use team id for availbility slider

sean-brydon:
Uses teamId...

### DIFF
--- a/packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
+++ b/packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
@@ -1,5 +1,3 @@
-import { Prisma } from "@prisma/client";
-
 import type { Dayjs } from "@calcom/dayjs";
 import dayjs from "@calcom/dayjs";
 import type { DateRange } from "@calcom/lib/date-ranges";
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -116,23 +103,15 @@ async function getInfoForAllTeams({ ctx, input }: GetOptions) {
...
   }
 
-  const getTotalMembers = await prisma.$queryRaw<{
-    count: number;
-  }>(Prisma.sql`
-      SELECT
-        COUNT(DISTINCT "userId") as "count"
-      FROM "Membership"
-      WHERE "teamId" IN (${Prisma.join(teamIds)})
-`);
```
sean-brydon, line 134: We actually do the distinct in the membership query so we can rely on that for the correct count value

packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -116,23 +103,15 @@ async function getInfoForAllTeams({ ctx, input }: GetOptions) {
       },
       select: {
         id: true,
+        teamId: true,
       },
     })
-    .then((memberships) => memberships.map((membership) => membership.id));
+    .then((memberships) => memberships.map((membership) => membership.teamId));
```
sean-brydon, line 109: Use the teamId...

emrysal (approved):

PR #10958, opened 2023-08-25
fix: webhooks for managed event types

### LINKED ISSUE #10940: trigger user and team webhooks for managed events
### Issue Summary

Currently, for managed event types only the team webhook triggeres. We also want the user webhooks to trigger. Same with Zapier, a managed event type should trigger for the team and the user.

CarinaWolli:
## What does this PR do?

Webhooks for managed event types should trigger for team webhooks and user webhooks. Before they only triggerd for team webhooks. Also fixes that listBookings (for Zapier) now returns bookings from managed events when using user or team API key. 

Fixes #10940

## Type of change

## How should this be tested?

- Create a managed event type
- Create a team webhook and a user webhook that triggers when booking is created (use webhook.site)
- Book managed event type, see that webhooks triggered twice (once for users, once for team)
- Test other webhook triggers + test booking normal user event type and team event type (round robin, collective)

### DIFF
--- a/packages/app-store/zapier/api/subscriptions/listBookings.ts
+++ b/packages/app-store/zapier/api/subscriptions/listBookings.ts
@@ -28,7 +28,6 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
       };
     } else {
       where.userId = validKey.userId;
-      s;
     }
 
     const bookings = await prisma.booking.findMany({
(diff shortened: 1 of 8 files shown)


packages/app-store/zapier/api/subscriptions/listBookings.ts
```diff
@@ -22,8 +22,15 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
     const where: Prisma.BookingWhereInput = {};
-    if (validKey.teamId) where.eventType = { teamId: validKey.teamId };
-    else where.userId = validKey.userId;
+    if (validKey.teamId) {
+      where.eventType = {
+        OR: [{ teamId: validKey.teamId }, { parent: { teamId: validKey.teamId } }],
+      };
+    } else {
+      where.userId = validKey.userId;
+      s;
```
zomars, line 31:
```suggestion
```

zomars (approved):
LGTM

Issue #11007, opened 2023-08-29
Module not Found


PeterKwesiAnsah:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

My local setup is failing to compile because of a missing module 

### Steps to Reproduce

1. Follow the developer setup guide

### Actual Results

- Failing Compilation

### Expected Results

- The app should be able to run successfully

### Technical details

- Node.js version-v18.17.0

### Evidence
![Screenshot (47)](https://github.com/calcom/cal.com/assets/31078314/c8b70d2c-37ad-4397-949c-1445f5330bdc)

suyash5053:
@PeterKwesiAnsah Try running yarn again, the dependecy might got skipped.

Taha-1005:
Hi @PeterKwesiAnsah , I also faced the same issue. I solved it by deleting the repo, cloning it again and then running yarn again. Hope this helps.

PeterKwesiAnsah:
@Taha-1005 I tried what you said and it's working now.

PR #11010, opened 2023-08-29
fix: Module Not Found

Suvankar621:
Fix Bug: #11007

- **Fix Module not Found :** 

In the code base the file name extension must be **ssg.js** instead of **ssg.ts**

It will fix the issue


Udit-takkar (commented):
@Suvankar621 can you elaborate more about this bug?

I never got module not found

Suvankar621:
> @Suvankar621 can you elaborate more about this bug?
> 
> I never got module not found



> In the code base the file name extension must be **ssg.js** instead of **ssg.ts**

@Udit-takkar
 Check Module Installation: First and foremost, ensure that you have the @trpc/react-query package installed in your project. Run the following command to install it:

**Copy code
npm install @trpc/react-query**

**Verify Module Path:** **Verify that the path @trpc/react-query/ssg** is correct. 
The module path should match the actual folder structure and module hierarchy within the package. Double-check the package's documentation or source code to confirm the correct path.

**Case Sensitivity:** Module paths in JavaScript are case-sensitive. Ensure that the casing of the module path matches the actual file structure. For instance, if the file is named **ssg.js** or **ssg.ts**, ensure that your module path uses lowercase **(@trpc/react-query/ssg)** and not a different casing.

**Check Package Structure:** Make sure that the **@trpc/react-query** package actually contains an **ssg sub-module**. This might involve looking through the package's source code or documentation.

**File Extension:** Depending on the environment and tools you're using, you might need to specify the file extension of the module you're importing. If the module is a TypeScript module, ensure you're using **.ts** or **.tsx**. If it's a JavaScript module, use **.js**.

**Typescript Configuration:** If you're using TypeScript, check your **tsconfig.json file**.
 Ensure that the module resolution settings are correctly configured to match your project structure and module paths.

After doing all these things i changed file name  **ssg.ts** instead of **ssg.js**

**../../packages/trpc/react/ssg.ts to ../../packages/trpc/react/ssg.js** and Module Not Found issue is solved

Udit-takkar:
@Suvankar621 it's throwing a type error

PR #10998, opened 2023-08-28
fix: Fixes username invite issue

sean-brydon:
Fixes: CAL-2393 #10993 

Fixes the ability for users to invite members to team via usernames. 

This works in and out of org context.

How to test: 
Create a team
invite an existing user via email
invite an existing user via username

PR #11012, opened 2023-08-29
fix: fix-team-invite-username

### LINKED ISSUE #10993: [CAL-2393] Team Invite Broken: Cannot Invite Userrnames only Emails
When you try to invite a user to a team using "username," it says this user cannot be invited because  email cannot be found. We should be able to invite via username AND email 

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2393](https://linear.app/calcom/issue/CAL-2393/team-invite-broken-cannot-invite-userrnames-only-emails)</sub>

samueldenzil:
## What does this PR do?

Fixes #10993 
Fixes the issue where users can invite members using username and email

I modified the query to include a conditional check `isOrg` to determine whether organizationId should be considered in the filter.

When isOrg is true, the query filters for the combination of username and organizationId and when isOrg is false, the query only filters based on username.

## Type of change

### DIFF
--- a/packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
+++ b/packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
@@ -76,7 +76,12 @@ export async function getUserToInviteOrThrowIfExists({
   // Check if user exists in ORG or exists all together
   const invitee = await prisma.user.findFirst({
     where: {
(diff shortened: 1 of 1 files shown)


Udit-takkar (commented):
already fixed by https://github.com/calcom/cal.com/pull/10998

PR #11032, opened 2023-08-30
fix: multiple duration when booking

leog:
## What does this PR do?

When on a multiple duration event type of any sort, you could choose a different duration in the booking form which could be outside availability. This restricts choosing a different duration from that booker state.

https://github.com/calcom/cal.com/assets/467258/5bcd1023-9bd1-46ad-8034-f9d1a333645f

## Type of change

## How should this be tested?

Create a multiple duration event type and when booking see that you can select a duration but can't choose a different one in the booking form state.


alannnc (approved):
It works!

shivamklr (commented):
Tested, lgtm.

PR #11041, opened 2023-08-31
fix: Handle disabling of ORGANIZATIONS_ENABLED correctly.

hariombalhara:
## What does this PR do?

- Improve `ORGANIZATIONS_ENABLED` check so that it is only enabled when it is set to "1" or "true" as per the [comment](https://github.com/calcom/cal.com/blob/4e5efd6ad6510d8d6a6b1ad2b743669de2ac3fba/.env.example#L212). It allows disabling it by setting it to "0" or "false"
- Adding header logging for redirects only when the env variable is enabled to avoid false logs.

## Type of change

## How should this be tested?
- Simply go to app.cal.local:3000 with the env variable disabled and verify that the headers starting with `X-Cal-Org` aren't showing up.
- Enable the env variable and repeat the test to see that now they are visible.

## Mandatory Tasks

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -14,7 +14,8 @@ const {
 
 if (!process.env.NEXTAUTH_SECRET) throw new Error("Please set NEXTAUTH_SECRET");
 if (!process.env.CALENDSO_ENCRYPTION_KEY) throw new Error("Please set CALENDSO_ENCRYPTION_KEY");
-
+const isOrganizationsEnabled =
+  process.env.ORGANIZATIONS_ENABLED === "1" || process.env.ORGANIZATIONS_ENABLED === "true";
 // To be able to use the version in the app without having to import package.json
 process.env.NEXT_PUBLIC_CALCOM_VERSION = version;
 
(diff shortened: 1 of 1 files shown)


apps/web/next.config.js
```diff
@@ -226,7 +227,7 @@ const nextConfig = {
   async rewrites() {
     const beforeFiles = [
       // These rewrites are other than booking pages rewrites and so that they aren't redirected to org pages ensure that they happen in beforeFiles
-      ...(process.env.ORGANIZATIONS_ENABLED
```
hariombalhara, line 229: Earlier we were checking the emptiness of the env variable and thus there was no way to disable it by setting it to 0 or false

apps/web/next.config.js
```diff
@@ -333,44 +334,46 @@ const nextConfig = {
           },
         ],
       },
-      ...[
```
hariombalhara, line 336: Add the logging headers only when the redirect is actually happening. [Same check is already there for redirects](https://github.com/calcom/cal.com/blob/81116abd48072996d8f162e73267617f7adb15ae/apps/web/next.config.js#L230).

leog (approved):
Great catch, thanks for taking care of this.

PR #11048, opened 2023-08-31
fix: include app data and credentials from DB

### LINKED ISSUE #11040: refactor: show apps of user query
the current implementation of https://github.com/calcom/cal.com/pull/11036 wont scale and will be slow for showing 

we likely need to add the app data to the user query to not slow down

alannnc:
## What does this PR do?

- Includes app data in main team user select query to increase performace.
- Adds connectedApps to member obj and uses map to read app data

Fixes #11040 

## Requirement/Documentation

## Type of change

## How should this be tested?

- What are the minimal test data to have? You can use teampro | teamfree@example accounts with some installed apps.
- What is expected (happy path) to have (input and output)? Visit Seeded team members page and see installed app in each user.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/lib/server/queries/teams/index.ts
+++ b/packages/lib/server/queries/teams/index.ts
@@ -31,7 +31,12 @@ export async function getTeamWithMembers(args: {
     selectedCalendars: true,
     credentials: {
       include: {
-        app: true,
+        app: {
+          select: {
+            slug: true,
+            categories: true,
+          },
+        },
         destinationCalendars: {
           select: {
             externalId: true,
@@ -129,6 +134,9 @@ export async function getTeamWithMembers(args: {
 
   if (!team) return null;
 
+  // This should improve performance saving already app data found.
+  const appDataMap = new Map();
+
   const members = team.members.map((obj) => {
     return {
       ...obj.user,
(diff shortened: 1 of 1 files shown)


packages/lib/server/queries/teams/index.ts
```diff
@@ -124,43 +133,39 @@ export async function getTeamWithMembers(args: {
...
-        disableImpersonation: obj.disableImpersonation,
-        avatar: `${WEBAPP_URL}/${obj.user.username}/avatar.png`,
-        connectedApps,
-      };
-      return rest;
-    })
-  );
+
+  // This should improve performance saving already app data found.
+  const appDataMap = new Map();
```
alannnc, line 138: Using Map here to avoid array look up on already fetched appData

packages/lib/server/queries/teams/index.ts
```diff
@@ -32,7 +31,17 @@ export async function getTeamWithMembers(args: {
     selectedCalendars: true,
     credentials: {
       include: {
-        app: true,
+        app: {
+          select: {
+            slug: true,
+            categories: true,
```
alannnc, line 37: Only fetch required data

packages/lib/server/queries/teams/index.ts
```diff
@@ -124,43 +133,39 @@ export async function getTeamWithMembers(args: {
...
+        const appSlug = cred.app?.slug;
+        let appData = appDataMap.get(appSlug);
+
+        if (!appData) {
+          appData = getAppFromSlug(appSlug);
+          appDataMap.set(appSlug, appData);
+        }
+
+        const isCalendar = cred?.app?.categories.includes("calendar");
+        const externalId = isCalendar ? cred.destinationCalendars[0]?.externalId : undefined;
```
alannnc, line 157: Getting [0] index should give us the main destinationCalendar, but I don't think this would be more than 1 item here since it's 1 calendar per credential.

zomars (approved):
Nice improvements. Thank you @alannnc 🙏🏽

PR #11052, opened 2023-08-31
fix: resolve zod versions to allow AI to pass build

DexterStorey:
## What does this PR do?

The AI app requires a newer version of zod (3.22.2) than the rest of the mono repo (3.20.2)

If the newer version is included in the ai package.json, the other apps will use that package and there will be errors, so we must resolve the older version in the root package and specify the newer version in the ai folder.

## Type of change

## Mandatory Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,8 @@
     "@apidevtools/json-schema-ref-parser": "9.0.9",
     "@types/node": "16.9.1",
     "@types/react": "18.0.26",
-    "@types/react-dom": "^18.0.9"
+    "@types/react-dom": "^18.0.9",
+    "zod": "^3.20.2"
   },
   "lint-staged": {
     "(apps|packages)/**/*.{js,ts,jsx,tsx}": [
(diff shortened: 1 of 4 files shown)


DexterStorey:
@emrysal I didn't commit the new yarn.lock but I think it might need to be updated to support the new resolutions.

If there is any easy way for me build the lock file locally without the massive diff, please let me know.

DexterStorey:
@baileypumfleet This PR should allow the ai build to pass on vercel

PeerRich (approved):

package.json
```diff
@@ -102,7 +102,8 @@
     "@apidevtools/json-schema-ref-parser": "9.0.9",
     "@types/node": "16.9.1",
     "@types/react": "18.0.26",
-    "@types/react-dom": "^18.0.9"
+    "@types/react-dom": "^18.0.9",
+    "zod": "^3.20.2"
```
PeerRich, line 106: @zomars was this a mistake by me merging?

zomars: Zod isn't needed on root AFAIK.

PR #11057, opened 2023-08-31
fix: team sub quantity update stripe

alannnc:
## What does this PR do?

- Fixes sub team quantity assuming the new member it's already listed when fetching team.members.

## Type of change

## Mandatory Tasks

## Checklist


packages/features/ee/teams/lib/payments.ts
```diff
@@ -117,10 +117,8 @@ export const updateQuantitySubscriptionFromStripe = async (teamId: number) => {
       return;
     }
 
-    const newQuantity = membershipCount - subscriptionQuantity;
-
```
alannnc, line 121: We fetch team.members earlier here, my guess this function was being called before adding team member. But now as we fetch team.member and we already count the latest added we can update quantity to membershipCount.

packages/features/ee/teams/lib/payments.ts
```diff
@@ -117,10 +117,8 @@ export const updateQuantitySubscriptionFromStripe = async (teamId: number) => {
...
     await stripe.subscriptions.update(subscriptionId, {
-      items: [{ quantity: membershipCount + newQuantity, id: subscriptionItemId }],
+      items: [{ quantity: membershipCount, id: subscriptionItemId }],
```
alannnc, line 121: membershipCount was 3 but subscriptionQty was 2 so this will result in 4 instead of 3.
```suggestion
     const newQuantity = membershipCount - subscriptionQuantity; // 3 - 2 = 1
     ....
     items: [{ quantity: membershipCount + newQuantity, id: subscriptionItemId }], // 3 + 1 = 4 Which is wrong
```

zomars (approved):
LFG!

Ship it

PR #11059, opened 2023-08-31
feat: Sync app credentials between Cal.com & self-hosted platforms

joeauyeung:
## What does this PR do?

This PR allows self-hosters that have their own app store, to sync their user's 3rd party credentials with their instance of Cal.com. This allows their app store to be the source of truth for apps.

Fixes # (issue)

## Requirement/Documentation

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How should this be tested?

- Set up the required .env variables that were added in the PR
- Hit the `api/webhook/app-credential` endpoint with the correct webhook secret and encrypted payload
- The new credentials should be created in the DB
- For the affected apps, force a refresh of the tokens. The endpoint they are hitting should simulate what a self-hoster would have made
- Force a refresh of the tokens but this time do not enable app credential syncing

## Mandatory Tasks

## Checklist


apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,76 @@
...
+  // Check that credential sharing is enabled
+  if (!APP_CREDENTIAL_SHARING_ENABLED) {
+    return res.status(403).json({ message: "Credential sharing is not enabled" });
+  }
+
+  // Check that the webhook secret matches
+  if (req.headers["calcom-webhook-secret"] !== process.env.CALCOM_WEBHOOK_SECRET) {
+    return res.status(403).json({ message: "Invalid webhook secret" });
+  }
+
```
joeauyeung, line 26: TODO: Add decoding step here

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,76 @@
...
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  // Check that credential sharing is enabled
+  if (!APP_CREDENTIAL_SHARING_ENABLED) {
+    return res.status(403).json({ message: "Credential sharing is not enabled" });
+  }
+
+  // Check that the webhook secret matches
+  if (req.headers["calcom-webhook-secret"] !== process.env.CALCOM_WEBHOOK_SECRET) {
+    return res.status(403).json({ message: "Invalid webhook secret" });
+  }
```
joeauyeung, line 25: Would this secret also be used in future webhooks as well? If we add this to the SaSS product then would need UI to generate a secret based on userId.

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,76 @@
...
+import { appStoreMetadata } from "@calcom/app-store/appStoreMetaData";
+import { APP_CREDENTIAL_SHARING_ENABLED } from "@calcom/lib/constants";
+import prisma from "@calcom/prisma";
+
+const appCredentialWebhookRequestBodySchema = z.object({
+  // UserId of the cal.com user
+  userId: z.number().int(),
+  //   The dirname of the app under packages/app-store
+  appDirName: z.string(),
+  keys: z.record(z.any()),
```
joeauyeung, line 13: Thinking each app should have their own keys schema

joeauyeung (commented):
Initial review

apps/web/pages/api/webhook/app-credential.ts
joeauyeung, line 1: The webhook endpoint used to send app credentials to Cal

packages/app-store/_utils/oauth/createOAuthAppCredential.ts
```diff
@@ -3,8 +3,8 @@ import type { NextApiRequest } from "next";
 import { HttpError } from "@calcom/lib/http-error";
 import prisma from "@calcom/prisma";
 
-import { decodeOAuthState } from "./decodeOAuthState";
-import { throwIfNotHaveAdminAccessToTeam } from "./throwIfNotHaveAdminAccessToTeam";
+import { decodeOAuthState } from "../oauth/decodeOAuthState";
```
joeauyeung, line 6: Moved these OAuth helpers into their own folder in `app-store/_utils`

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { APP_CREDENTIAL_SHARING_ENABLED } from "@calcom/lib/constants";
+
+const minimumTokenReponseSchema = z.object({
+  access_token: z.string(),
+  //   Assume that any property with a number is the expiry
+  [z.string().toString()]: z.number(),
+  //   Allow other properties in the token response
+  [z.string().optional().toString()]: z.unknown().optional(),
+});
```
joeauyeung, line 11: If we are being sent tokens then they should at least contain the access token and the expiry date. The expiry date could be named different things such as expiry_date, expires_in, etc

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
```diff
@@ -0,0 +1,32 @@
...
+    refreshTokenResponse = schema.parse(response);
+  }
+
+  if (!refreshTokenResponse.success) {
+    throw new Error("Invalid refreshed tokens were returned");
+  }
+
+  if (!refreshTokenResponse.data.refresh_token) {
+    refreshTokenResponse.data.refresh_token = "refresh_token";
+  }
```
joeauyeung, line 27: Some app schemas expect this property still.

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
```diff
@@ -0,0 +1,22 @@
...
+    // Customize the payload based on what your endpoint requires
+    // The response should only contain the access token and expiry date
+    const response = await fetch(process.env.CALCOM_CREDENTIAL_SYNC_ENDPOINT, {
+      method: "POST",
+      body: new URLSearchParams({
+        calcomUserId: userId.toString(),
+        appSlug,
+      }),
+    });
+    return response;
```
joeauyeung, line 15: If app credential sharing is enabled then make a call to the self-hosted platform for new tokens.

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
joeauyeung, line 1: This function is used to parse the response from the self-hosted platform or the app provider when requesting tokens.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -47,11 +49,18 @@ export default class GoogleCalendarService implements Calendar {
...
       try {
-        const { res } = await myGoogleAuth.refreshToken(googleCredentials.refresh_token);
+        const res = await refreshOAuthTokens(
+          async () => {
+            const fetchTokens = await myGoogleAuth.refreshToken(googleCredentials.refresh_token);
+            return fetchTokens.res;
+          },
+          "google-calendar",
+          credential.userId
+        );
```
joeauyeung, line 59: If credential sharing is not enabled then request tokens as normal.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -47,11 +49,18 @@ export default class GoogleCalendarService implements Calendar {
...
         const token = res?.data;
         googleCredentials.access_token = token.access_token;
         googleCredentials.expiry_date = token.expiry_date;
-        const key = googleCredentialSchema.parse(googleCredentials);
+        const key = parseRefreshTokenResponse(googleCredentials, googleCredentialSchema);
```
joeauyeung, line 63: If credential sharing is enabled then parse the response from the self-hosted platform or from the 3rd party app.

packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -60,6 +72,32 @@ export default class SalesforceCalendarService implements Calendar {
...
+    const accessTokenParsed = parseRefreshTokenResponse(accessTokenJson, salesforceTokenSchema);
+
+    if (!accessTokenParsed.success) {
+      return Promise.reject(new Error("Invalid refreshed tokens were returned"));
+    }
+
+    await prisma.credential.update({
+      where: { id: credential.id },
+      data: { key: { ...accessTokenParsed.data, refresh_token: credentialKey.refresh_token } },
+    });
```
joeauyeung, line 99: Using JSforce it abstracted refreshing the access token. It also wasn't being saved to the credential record so we were always refreshing the access tokens. If we want to sync credentials then we need to hit the endpoint manually if credential sharing isn't enabled.

packages/app-store/zohocrm/api/_getAdd.ts
```diff
@@ -13,7 +14,20 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+    access_type: "offline",
+    state,
+    prompt: "consent",
+  };
+
+  // const redirectUri = WEBAPP_URL + "/api/integrations/zohocrm/callback";
+  // const url = `https://accounts.zoho.com/oauth/v2/auth?scope=ZohoCRM.modules.ALL,ZohoCRM.users.READ,AaaServer.profile.READ&client_id=${client_id}&response_type=code&access_type=offline&redirect_uri=${redirectUri}&state=${state}&prompt=consent`;
+  const query = stringify(params);
+  const url = `https://accounts.zoho.com/oauth/v2/auth?${query}`;
+
```
joeauyeung, line 31: This change is also made in https://github.com/calcom/cal.com/pull/11161

joeauyeung (commented):
Didn't test all apps because some mechanics that the apps use to refresh tokens were repeated. I did have to make changes to Salesforce as the method of getting new tokens was abstracted by JSforce.

alishaz-polymath:
Do you think it would make sense to place loggers to confirm exactly where the performance hit is? Is it the fetching, parsing or something else?

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,90 @@
...
+// https://github.com/colinhacks/zod/discussions/839#discussioncomment-6488540
+const [firstAppKey, ...otherAppKeys] = Object.keys(appStoreMetadata) as (keyof typeof appStoreMetadata)[];
+
+const appMetadataEnum = z.enum([firstAppKey, ...otherAppKeys]);
+
+const appCredentialWebhookRequestBodySchema = z.object({
+  // UserId of the cal.com user
+  userId: z.number().int(),
+  // The dirname of the app under packages/app-store
+  appDirName: appMetadataEnum,
```
keithwillcode, line 18: If this is tied to the directory name, does that mean we will break functionality if we rename/move it?

Even if not (which would be ideal), having this value tied what is a directory name feels wrong. Can we rename to something like `appId` or `appSlug`?

Reason: An external webhook should see our system as a block box. It shouldn't need to send a body param named `appDirName`.

joeauyeung: On line 47 we use the `appDirName` to find the app metadata in `appStoreMetadata`. I do agree that app slug does feel better but that means we would have to make a Prisma query to find the dirName from the slug or another identifier. I was also thinking becuase this endpoint would mostly be used for people self-hosting I think they would be a more comfortable digging into our folder structure. I think if we have a doc around this it should clear up any confusion.

keithwillcode: I think doing a query to find it is fine. Honestly think passing a directory name to an endpoint is way too tightly coupled to ship like that, even if for self-hosters.

.env.example
```diff
@@ -229,3 +229,16 @@ AUTH_BEARER_TOKEN_VERCEL=
...
+
+# - APP CREDENTIAL SYNC ***********************************************************************************
+# Used for self-hosters that are implementing Cal.com into their applications that already have certain integrations
+# Under settings/admin/apps ensure that all app secrets are set the same as the parent application
+# You can use: `openssl rand -base64 32` to generate one
+CALCOM_WEBHOOK_SECRET=""
+# This is the header name that will be used to verify the webhook secret. Should be in lowercase
+CALCOM_WEBHOOK_HEADER_NAME="calcom-webhook-secret"
+CALCOM_CREDENTIAL_SYNC_ENDPOINT=""
+# Should key should match on Cal.com and your application
```
zomars, line 241: typo?

```suggestion
# Key should match on Cal.com and your application
```

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
zomars: Can we make this part of the Base CalendarService Class instead? That way we can invoque this like `this.refreshOAuthTokens` and share/override functionality if needed.

packages/app-store/office365video/lib/VideoApiAdapter.ts
```diff
@@ -57,16 +58,21 @@ const o365Auth = async (credential: CredentialPayload) => {
...
-      method: "POST",
-      headers: { "Content-Type": "application/x-www-form-urlencoded" },
-      body: new URLSearchParams({
-        client_id,
-        refresh_token: refreshToken,
-        grant_type: "refresh_token",
-        client_secret,
-      }),
-    });
+    const response = await refreshOAuthTokens(
```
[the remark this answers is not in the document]
zomars: Basic class inheritance should also work in this case.

packages/app-store/zoho-bigin/api/add.ts
```diff
@@ -14,7 +14,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     const clientId = typeof appKeys.client_id === "string" ? appKeys.client_id : "";
     if (!clientId) return res.status(400).json({ message: "Zoho Bigin client_id missing." });
 
-    const redirectUri = WEBAPP_URL + `/api/integrations/${appConfig.slug}/callback`;
+    const redirectUri = WEBAPP_URL + `/api/integrations/zoho-bigin/callback`;
```
zomars, line 17: Is this change intended?

zomars (dismissed):
No major code objections. LGTM. Although we're going to need some help testing all the apps since this touch every calendar app.

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  if (!appDirName) {
+    return res.status(404).json({ message: "App not found" });
+  }
+
+  //   Search for the app's slug and type
+  const appMetadata = appStoreMetadata[appDirName.slug as keyof typeof appStoreMetadata];
+
+  if (!appMetadata) {
+    return res.status(404).json({ message: "App not found. Ensure that you have the correct appDir" });
```
keithwillcode, line 53: This message is now wrong. Should mention the slug

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+  const reqBody = appCredentialWebhookRequestBodySchema.parse(req.body);
+
+  // Check that the user exists
+  const user = await prisma.user.findUnique({ where: { id: reqBody.userId } });
+
+  if (!user) {
+    return res.status(404).json({ message: "User not found" });
+  }
+
+  const appDirName = await prisma.app.findUnique({
```
keithwillcode, line 40: This should now be “app”, yeah?

packages/app-store/office365video/lib/VideoApiAdapter.ts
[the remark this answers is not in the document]
emrysal: Don't think so, this is more a decorator thing; don't think class inheritance would work at all

joeauyeung:
> No major code objections. LGTM. Although we're going to need some help testing all the apps since this touch every calendar app.

I tested every app this PR touches but I think it's good to get a second pair of eyes on it.

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
joeauyeung: What if we have an app that users OAuth tokens but isn't a calendar app? Should we create an App class that has this method and CalendarService can inherit that App class?

joeauyeung: Talked with @keithwillcode about this. This is a nice to have but non-blocking.

zomars (dismissed):
Nice work @joeauyeung 

Ship it

packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -262,14 +262,6 @@ export default class Office365CalendarService implements Calendar {
...
       const tokenResponse = parseRefreshTokenResponse(responseJson, refreshTokenResponseSchema);
       o365AuthCredentials = { ...o365AuthCredentials, ...(tokenResponse.success && tokenResponse.data) };
-      if (!tokenResponse.success) {
-        console.error(
-          "Outlook error grabbing new tokens ~ zodError:",
-          tokenResponse.error,
-          "MS response:",
-          responseJson
-        );
-      }
```
joeauyeung, line 265: Error handled in `parseRefreshTokenResponse`

emrysal (approved):

zomars:
This had to be reverted. It was preventing bookings with gCal (other cals could be affected) https://github.com/calcom/cal.com/pull/11449#pullrequestreview-1634246333

mfeuerstein (commented):
## PR Review — approved

Reviewed 30 files. 0 high-severity issues found. Verdict: approved.

### `packages/app-store/_utils/oauth/encodeOAuthState.ts` (low)

- Reviewed packages/app-store/_utils/oauth/encodeOAuthState.ts — looks good

### `packages/app-store/_utils/oauth/createOAuthAppCredential.ts` (low)

- Reviewed packages/app-store/_utils/oauth/createOAuthAppCredential.ts — looks good

### `packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts` (low)

- Reviewed packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts — looks good

### `packages/app-store/_utils/oauth/refreshOAuthTokens.ts` (low)

- Reviewed packages/app-store/_utils/oauth/refreshOAuthTokens.ts — looks good

### `packages/app-store/googlecalendar/api/add.ts` (low)

- Reviewed packages/app-store/googlecalendar/api/add.ts — looks good

### `packages/app-store/googlecalendar/lib/CalendarService.ts` (low)

- Reviewed packages/app-store/googlecalendar/lib/CalendarService.ts — looks good

### `packages/app-store/_utils/oauth/decodeOAuthState.ts` (low)

- Reviewed packages/app-store/_utils/oauth/decodeOAuthState.ts — looks good

### `packages/app-store/googlecalendar/api/callback.ts` (low)

- Reviewed packages/app-store/googlecalendar/api/callback.ts — looks good

### `packages/app-store/hubspot/api/add.ts` (low)

- Reviewed packages/app-store/hubspot/api/add.ts — looks good

### `packages/app-store/hubspot/api/callback.ts` (low)

- Reviewed packages/app-store/hubspot/api/callback.ts — looks good

### `packages/app-store/hubspot/lib/CalendarService.ts` (low)

- Reviewed packages/app-store/hubspot/lib/CalendarService.ts — looks good

### `packages/app-store/larkcalendar/api/callback.ts` (low)

- Reviewed packages/app-store/larkcalendar/api/callback.ts — looks good

### `apps/web/pages/api/webhook/app-credential.ts` (low)

- Reviewed apps/web/pages/api/webhook/app-credential.ts — looks good

### `.env.example` (low)

- Reviewed .env.example — looks good

### `packages/app-store/office365calendar/api/add.ts` (low)

- Reviewed packages/app-store/office365calendar/api/add.ts — looks good

### `packages/app-store/office365calendar/api/callback.ts` (low)

- Reviewed packages/app-store/office365calendar/api/callback.ts — looks good

### `packages/app-store/larkcalendar/lib/CalendarService.ts` (low)

- Reviewed packages/app-store/larkcalendar/lib/CalendarService.ts — looks good

### `packages/app-store/office365video/api/callback.ts` (low)

- Reviewed packages/app-store/office365video/api/callback.ts — looks good

### `packages/app-store/salesforce/api/add.ts` (low)

- Reviewed packages/app-store/salesforce/api/add.ts — looks good

### `packages/app-store/office365video/api/add.ts` (low)

- Reviewed packages/app-store/office365video/api/add.ts — looks good

### `packages/app-store/office365video/lib/VideoApiAdapter.ts` (low)

- Reviewed packages/app-store/office365video/lib/VideoApiAdapter.ts — looks good

### `packages/app-store/office365calendar/lib/CalendarService.ts` (low)

- Reviewed packages/app-store/office365calendar/lib/CalendarService.ts — looks good

### `packages/app-store/salesforce/api/callback.ts` (low)

- Reviewed packages/app-store/salesforce/api/callback.ts — looks good

### `packages/app-store/salesforce/lib/CalendarService.ts` (low)

- Reviewed packages/app-store/salesforce/lib/CalendarService.ts — looks good

### `packages/app-store/larkcalendar/api/add.ts` (low)

- Reviewed packages/app-store/larkcalendar/api/add.ts — looks good

### `packages/app-store/webex/lib/VideoApiAdapter.ts` (low)

- Reviewed packages/app-store/webex/lib/VideoApiAdapter.ts — looks good

### `packages/app-store/tandemvideo/api/callback.ts` (low)

- Reviewed packages/app-store/tandemvideo/api/callback.ts — looks good

### `packages/app-store/stripepayment/api/callback.ts` (low)

- Reviewed packages/app-store/stripepayment/api/callback.ts — looks good

### `packages/app-store/zoho-bigin/api/add.ts` (low)

- Reviewed packages/app-store/zoho-bigin/api/add.ts — looks good

### `packages/app-store/webex/api/callback.ts` (low)

- Reviewed packages/app-store/webex/api/callback.ts — looks good

ron-x5labs (commented):
## Code Review: feat: Sync app credentials between Cal.com & self-hosted platforms

### Problem
This PR enables self-hosters with their own app store to sync users' 3rd-party OAuth credentials with their Cal.com instance, making the self-hosted app store the source of truth. It adds an inbound webhook (`/api/webhook/app-credential`) that decrypts and upserts credentials, and redirects each app's OAuth token-refresh flow to a self-hosted endpoint (`CALCOM_CREDENTIAL_SYNC_ENDPOINT`) when `APP_CREDENTIAL_SHARING_ENABLED`.

### Solution Reviewed
Two new utilities (`refreshOAuthTokens`, `parseRefreshTokenResponse`) centralize the refresh/parse logic across ~10 calendar and video apps. OAuth helpers were moved into `_utils/oauth/`. Four new env vars are passed through `turbo.json`. The webhook authenticates via a shared header secret, decrypts keys with AES256, and upserts into the `Credential` table.

### Summary
This PR has multiple blocking defects that caused it to be reverted post-merge ("preventing bookings with gCal"). The root cause is a return-type contract regression: `parseRefreshTokenResponse` returns the full `safeParse` result `{success, data}` instead of `.data`, and the Google Calendar caller stores that wrapper directly as `credential.key`, corrupting the stored credential so every subsequent gCal auth parse throws. Beyond the revert cause, there are contract mismatches (slug vs dirName, credentialId vs userId), a broken zod schema, missing error handling on the webhook, and a tenant-isolation gap where anyone holding the shared webhook secret can overwrite any user's credentials.

### Verification
- Reviewed all 40 changed files against the PR commit (9fde0e90) in a local worktree.
- Verified `appStoreMetadata` is keyed by dirName (`googlecalendar`, `zoomvideo`) not slug (`google-calendar`, `zoom`) — confirmed in `apps.metadata.generated.ts`.
- Verified `parseRefreshTokenResponse` returns `refreshTokenResponse` (the full safeParse result) at line 29, and `googlecalendar/lib/CalendarService.ts:97` stores it verbatim as `key` at line 100.
- Verified zoho-bigin passes `credentialId` (credential PK) not `credential.userId` at line 93.
- Verified Salesforce builds `jsforce.Connection` with stale `credentialKey.access_token` at line 106 after refreshing.
- Type-check/build not run: the PR is from 2023 against a much older main; running against current main is not meaningful.

### Issues Found

#### Blocking

1. **googlecalendar/lib/CalendarService.ts:97** — `parseRefreshTokenResponse` returns `{success, data}` (line 29 of the util), not `.data`. Line 97 stores this wrapper directly as `credential.key` (line 100: `data: { key }`). Pre-PR used `googleCredentialSchema.parse()` (unwrapped). The stored key becomes `{success:true, data:{...}}`; the next `googleCredentialSchema.parse(credential.key)` fails → every gCal booking throws. **This is the regression that forced the revert.** Other callers (office365:264, zoom:111, salesforce:98) correctly use `.data`.

2. **_utils/oauth/refreshOAuthTokens.ts:15** — The sharing-enabled branch returns a raw `fetch Response`, but the non-sharing branch returns the app-native shape (GaxiosResponse `.res.data`, axios `.data`, Hubspot token object). Callers that read `.data` (google:94, zoho-bigin:96, zohocrm:217) get `undefined` → `TypeError` in sharing mode. The sync branch must normalize to a shape each caller expects.

3. **app-credential.ts:50** — `appStoreMetadata[app.slug]` indexes the metadata map by `App.slug` (e.g. `google-calendar`, `zoom`), but the map is keyed by dirName (`googlecalendar`, `zoomvideo`). The lookup returns `undefined` → 404 for every app where slug ≠ dirName (Google, Lark, Outlook Calendar, MS Teams, Zoom). Should use `app.dirName`, and the Prisma `select` at line 42 must include `dirName`.

4. **app-credential.ts:31** — `appCredentialWebhookRequestBodySchema.parse(req.body)` uses `.parse` (not `safeParse`). A malformed body throws an uncaught `ZodError` → 500 with stack trace and request-body shape echoed to the caller (information leak). Use `safeParse` and return a structured 400.

5. **app-credential.ts:57** — `JSON.parse(symmetricDecrypt(...))` is unguarded. `symmetricDecrypt` throws on wrong/missing key, bad padding, or corrupt ciphertext; `JSON.parse` throws on non-JSON plaintext. No try/catch → 500 with stack trace. Additionally, `symmetricDecrypt` uses AES-256-CBC without a MAC, so tampered ciphertext produces distinct error shapes (padding vs JSON) — a padding-oracle distinguisher on an internet-facing endpoint. Wrap in try/catch returning 400; consider AES-GCM or an HMAC.

6. **app-credential.ts:31 (userId)** — Tenant isolation failure: `userId` is taken directly from the request body with no session/org binding. The only authentication is a single global shared webhook secret. Anyone holding `CALCOM_WEBHOOK_SECRET` can upsert OAuth credentials for **any** Cal.com user ID — a cross-tenant credential-overwrite primitive. Bind the caller to a permitted userId scope or require per-user signing.

7. **salesforce/lib/CalendarService.ts:86** — `if (response.statusText !== "OK")` is unreliable: HTTP/2 has no `statusText` (empty string on 200), so legitimate responses are wrongly rejected as 400. Use `response.ok` / `response.status`. Additionally, line 106 builds the `jsforce.Connection` with `credentialKey.access_token` (the **stale** token) instead of the freshly-refreshed `accessTokenParsed.data.access_token` — the persisted new token is never used by the connection.

8. **zoho-bigin/lib/CalendarService.ts:93** — Passes `credentialId` (the Credential row PK, set at line 9) instead of `credential.userId` to `refreshOAuthTokens`. The sync endpoint receives the credential PK as `calcomUserId`; when it calls the webhook back, `prisma.user.findUnique({ where: { id: reqBody.userId } })` runs against the credential PK → 404. All other apps correctly pass `credential.userId`.

9. **zoomvideo/lib/VideoApiAdapter.ts:92** — Passes appSlug `"zoomvideo"` (the dirName), but the webhook resolves the app via `prisma.app.findUnique({ where: { slug } })` expecting `"zoom"` (the App.slug). A sync callback carrying `zoomvideo` finds no App row → 404.

10. **_utils/oauth/parseRefreshTokenResponse.ts:15** — Gating mismatch with `refreshOAuthTokens`: the router checks `APP_CREDENTIAL_SHARING_ENABLED && CALCOM_CREDENTIAL_SYNC_ENDPOINT && userId`, but the parser checks only the first two (no `userId`). Consequence: (a) team credentials (`userId` null) run the native refresh but are parsed with the stripping minimum schema; (b) Salesforce uses `parseRefreshTokenResponse` but never calls `refreshOAuthTokens`, so when sharing is enabled its direct `login.salesforce.com` response is parsed with the minimum schema, dropping `instance_url`/`signature`/`scope`/`token_type` and breaking the jsforce connection.

#### Non-blocking

- **app-credential.ts:25** — Webhook secret compared with `!==` (not constant-time). Use `crypto.timingSafeEqual` on equal-length Buffers.
- **app-credential.ts:17** — No HTTP method guard. A GET with no body proceeds to `.parse(undefined)` → 500. Add `if (req.method !== "POST") return res.status(405).end();`.
- **app-credential.ts:62** — TOCTOU race: `findFirst` → `update`/`create` is not atomic. The `Credential` model has no `@@unique([userId, appId])` (only indexes), so concurrent/replayed webhooks can both create duplicate rows. Use `upsert` with a unique constraint or wrap in a transaction.
- **refreshOAuthTokens.ts:8** — `fetch(CALCOM_CREDENTIAL_SYNC_ENDPOINT, ...)` has no https enforcement, no timeout/AbortSignal, and no `response.ok` check. An `http://` endpoint leaks `calcomUserId` + `appSlug` in plaintext.
- **parseRefreshTokenResponse.ts:8** — `minimumTokenResponseSchema` computed keys `[z.string().toString()]` resolve to the literal `"[object Object]"`, not wildcards. The schema only enforces `access_token: string`; the "expiry must be a number" rule is dead and extras aren't actually allowed as documented. Use `z.object({ access_token: z.string() }).passthrough()` with an explicit numeric expiry check.
- **parseRefreshTokenResponse.ts:26** — When `refresh_token` is missing, the code injects the literal string `"refresh_token"` as a placeholder. This bogus sentinel gets persisted into the credential key and later sent to refresh endpoints. Either fail or omit the field.
- **parseRefreshTokenResponse.ts:22** — `throw new Error("Invalid refreshed tokens were returned")` swallows the zod error and raw response. The PR also removed the O365 `console.error` logging, making token-refresh failures undiagnosable across all apps. Include/log the parse error.
- **refreshOAuthTokens.ts:11** — Contract asymmetry: sends `calcomUserId` but the webhook expects `userId`. The self-hoster must know to remap — undocumented and error-prone.
- **salesforce/lib/CalendarService.ts:75** — The token refresh now runs eagerly in `getClient` (called from the constructor) on every instantiation — a network call + DB write per instance, with no caching.

#### Suggestions

- Dead code: `if (!accessTokenParsed.success)` (salesforce:92), `if (!parsedToken.success)` (zoom:108), and `...(tokenResponse.success && tokenResponse.data)` (office365:264) are unreachable — `parseRefreshTokenResponse` already throws on `!success`. Remove or restructure so the failure path is actually handled.
- No tests were added for the webhook endpoint or the two new utilities (the PR description confirms this). The handler has 6+ distinct failure modes that are regression-prone.
- No changelog or documentation for the new feature, endpoint contract, or env-var setup. Self-hosters have no reference for the AES256 encryption requirement or the `calcomUserId` → `userId` remapping.

### Verdict
Recommend changes before merge (retrospective — PR was merged then reverted). The revert was caused by the `parseRefreshTokenResponse` return-type regression corrupting gCal credential keys. The security posture of the webhook (global shared secret + attacker-controlled userId + uncaught decryption) also needs hardening before re-landing.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -81,11 +83,18 @@ export default class GoogleCalendarService implements Calendar {
...
```
ron-x5labs, line 97: 🔴 **Blocking — revert root cause.** `parseRefreshTokenResponse` returns the full `safeParse` result `{success, data}` (util line 29), not `.data`. Line 97 stores this wrapper directly as `credential.key` (line 100: `data: { key }`). Pre-PR used `googleCredentialSchema.parse()` (unwrapped). The stored key becomes `{success:true, data:{...}}`; the next `googleCredentialSchema.parse(credential.key)` fails → every gCal booking throws. This is the regression that forced the revert. Use `parseRefreshTokenResponse(...).data` here, as office365:264, zoom:111, and salesforce:98 correctly do.

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
ron-x5labs, line 15: 🔴 **Blocking.** The sharing-enabled branch returns a raw `fetch Response` (line 15), but the non-sharing branch returns the app-native shape (GaxiosResponse `.res.data`, axios `.data`, Hubspot token object). Callers reading `.data` (google:94, zoho-bigin:96, zohocrm:217) get `undefined` → `TypeError` in sharing mode. The sync branch must normalize to a shape each caller expects, or callers need to branch on the response type.

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+    where: { slug: reqBody.appSlug },
+    select: { slug: true },
+  });
+
+  if (!app) {
+    return res.status(404).json({ message: "App not found" });
+  }
+
+  //   Search for the app's slug and type
+  const appMetadata = appStoreMetadata[app.slug as keyof typeof appStoreMetadata];
```
ron-x5labs, line 50: 🔴 **Blocking.** `appStoreMetadata[app.slug]` indexes the metadata map by `App.slug` (e.g. `google-calendar`, `zoom`), but the map is keyed by **dirName** (`googlecalendar`, `zoomvideo`). The lookup returns `undefined` → 404 for every app where slug ≠ dirName (Google, Lark, Outlook Calendar, MS Teams, Zoom). Use `app.dirName`, and add `dirName: true` to the Prisma `select` at line 42.

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  // Check that the webhook secret matches
+  if (
+    req.headers[process.env.CALCOM_WEBHOOK_HEADER_NAME || "calcom-webhook-secret"] !==
+    process.env.CALCOM_WEBHOOK_SECRET
+  ) {
+    return res.status(403).json({ message: "Invalid webhook secret" });
+  }
+
+  const reqBody = appCredentialWebhookRequestBodySchema.parse(req.body);
```
ron-x5labs, line 31: 🔴 **Blocking (two issues).** (1) `.parse()` not `safeParse()` — a malformed body throws an uncaught `ZodError` → 500 with stack trace and request-body shape echoed to the caller (information leak). Use `safeParse` and return a structured 400. (2) **Tenant isolation:** `userId` is taken directly from the request body with no session/org binding. The only auth is a single global shared webhook secret, so anyone holding it can upsert OAuth credentials for **any** Cal.com user ID — a cross-tenant credential-overwrite primitive.

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  //   Search for the app's slug and type
+  const appMetadata = appStoreMetadata[app.slug as keyof typeof appStoreMetadata];
+
+  if (!appMetadata) {
+    return res.status(404).json({ message: "App not found. Ensure that you have the correct app slug" });
+  }
+
+  // Decrypt the keys
+  const keys = JSON.parse(
```
ron-x5labs, line 57: 🔴 **Blocking.** `JSON.parse(symmetricDecrypt(...))` is unguarded. `symmetricDecrypt` throws on wrong/missing key, bad padding, or corrupt ciphertext; `JSON.parse` throws on non-JSON. No try/catch → 500 with stack trace. Also, `symmetricDecrypt` uses AES-256-CBC without a MAC (`packages/lib/crypto.ts`), so tampered ciphertext produces distinct error shapes (padding vs JSON) — a padding-oracle distinguisher on an internet-facing endpoint. Wrap in try/catch returning 400; consider AES-GCM or an HMAC over the ciphertext.

packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -60,6 +72,32 @@ export default class SalesforceCalendarService implements Calendar {
...
+      body: new URLSearchParams({
+        grant_type: "refresh_token",
+        client_id: consumer_key,
+        client_secret: consumer_secret,
+        refresh_token: credentialKey.refresh_token,
+        format: "json",
+      }),
+    });
+
+    if (response.statusText !== "OK") throw new HttpError({ statusCode: 400, message: response.statusText });
```
ron-x5labs, line 86: 🔴 **Blocking (two issues).** (1) `response.statusText !== "OK"` is unreliable — HTTP/2 has no `statusText` (empty string on 200), so legitimate responses are wrongly rejected. Use `response.ok` / `response.status`. (2) Line 106 builds `jsforce.Connection` with `credentialKey.access_token` (the **stale** token) instead of the freshly-refreshed `accessTokenParsed.data.access_token` — the persisted new token (line 98) is never used by the connection.

packages/app-store/zoho-bigin/lib/CalendarService.ts
```diff
@@ -81,11 +82,16 @@ export default class BiginCalendarService implements Calendar {
...
-    });
+    const tokenInfo = await refreshOAuthTokens(
+      async () =>
+        await axios.post(accountsUrl, qs.stringify(formData), {
+          headers: {
+            "Content-Type": "application/x-www-form-urlencoded;charset=utf-8",
+          },
+        }),
+      "zoho-bigin",
+      credentialId
```
ron-x5labs, line 93: 🔴 **Blocking.** Passes `credentialId` (the Credential row PK, set at line 9) instead of `credential.userId` to `refreshOAuthTokens`. The sync endpoint receives the credential PK as `calcomUserId`; when it calls the webhook back, `prisma.user.findUnique({ where: { id: reqBody.userId } })` runs against the credential PK → 404. All other apps correctly pass `credential.userId`.

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -74,17 +76,22 @@ const zoomAuth = (credential: CredentialPayload) => {
...
+          headers: {
+            Authorization: authHeader,
+            "Content-Type": "application/x-www-form-urlencoded",
+          },
+          body: new URLSearchParams({
+            refresh_token: refreshToken,
+            grant_type: "refresh_token",
+          }),
+        }),
+      "zoomvideo",
```
ron-x5labs, line 92: 🔴 **Blocking.** Passes appSlug `"zoomvideo"` (the dirName), but the webhook resolves the app via `prisma.app.findUnique({ where: { slug } })` expecting `"zoom"` (the `App.slug`). A sync callback carrying `zoomvideo` finds no App row → 404. The appSlug passed here must match the `App.slug` the webhook queries.

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
```diff
@@ -0,0 +1,32 @@
+import { z } from "zod";
+
+import { APP_CREDENTIAL_SHARING_ENABLED } from "@calcom/lib/constants";
+
+const minimumTokenResponseSchema = z.object({
+  access_token: z.string(),
+  //   Assume that any property with a number is the expiry
+  [z.string().toString()]: z.number(),
```
ron-x5labs, line 8: 🔴 **Blocking — two issues.** (1) `minimumTokenResponseSchema` computed keys `[z.string().toString()]` (line 8) and `[z.string().optional().toString()]` (line 10) resolve to the literal `"[object Object]"`, not wildcards. The schema only enforces `access_token: string`; the "expiry must be a number" rule is dead. Use `z.object({ access_token: z.string() }).passthrough()` with an explicit numeric expiry check. (2) **Gating mismatch** (line 15): `refreshOAuthTokens` checks `&& userId` but this parser doesn't — team credentials and Salesforce (no `refreshOAuthTokens`) get the stripping minimum schema, dropping critical fields like `instance_url`.

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
```diff
@@ -0,0 +1,32 @@
...
+  } else {
+    refreshTokenResponse = schema.safeParse(response);
+  }
+
+  if (!refreshTokenResponse.success) {
+    throw new Error("Invalid refreshed tokens were returned");
+  }
+
+  if (!refreshTokenResponse.data.refresh_token) {
+    refreshTokenResponse.data.refresh_token = "refresh_token";
```
ron-x5labs, line 26: 🟡 **Non-blocking.** When `refresh_token` is missing, the code injects the literal string `"refresh_token"` as a placeholder. This bogus sentinel gets persisted into the credential key and later sent to refresh endpoints, producing confusing downstream failures. Either fail when `refresh_token` is absent or omit the field — do not inject a placeholder.

apps/web/pages/api/webhook/app-credential.ts
ron-x5labs, line 50: 🔴 **Blocking — wrong lookup key.** `appStoreMetadata` is keyed by `dirName` (`googlecalendar`, `zoomvideo`, `office365calendar`, …), but `app.slug` is the public page slug (`google-calendar`, `zoom`, `office365-calendar`). For every app where slug ≠ dirName this returns `undefined` → 404. The canonical lookup (`_appRegistry.ts:17`) uses `app.dirName`.

```ts
const app = await prisma.app.findUnique({
  where: { slug: reqBody.appSlug },
  select: { slug: true, dirName: true },
});
// ...
const appMetadata = appStoreMetadata[app.dirName as keyof typeof appStoreMetadata];
```

packages/app-store/googlecalendar/lib/CalendarService.ts
ron-x5labs, line 97: 🔴 **Blocking — stores the wrapper, not the data.** `parseRefreshTokenResponse` returns the full `safeParse` result `{success, data}` (util line 29), but here it is stored verbatim as `credential.key`. Pre-PR used `googleCredentialSchema.parse(googleCredentials)` (unwrapped). The persisted key becomes `{success:true,data:{...}}`, so the next `googleCredentialSchema.parse(credential.key)` throws and every Google booking fails — for ALL users, not just sharing mode.

```ts
const parsed = parseRefreshTokenResponse(googleCredentials, googleCredentialSchema);
await prisma.credential.update({
  where: { id: credential.id },
  data: { key: parsed.data },
});
```

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
ron-x5labs, line 15: 🔴 **Blocking — response-shape mismatch in sharing mode.** The sharing branch returns a raw `fetch` Response, but the non-sharing branch returns each caller's native shape (gaxios `res` for Google, axios response for Zoho, Hubspot token object). Callers that read `.data` (google: `res?.data`, zoho-bigin: `tokenInfo.data`, zohocrm) or `.access_token` (hubspot) get `undefined` → `TypeError` when sharing is enabled. The sharing branch must normalize to a shape each caller expects (or callers must branch on the response type).

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
```diff
@@ -0,0 +1,32 @@
+import { z } from "zod";
+
+import { APP_CREDENTIAL_SHARING_ENABLED } from "@calcom/lib/constants";
+
+const minimumTokenResponseSchema = z.object({
```
ron-x5labs, line 5: 🔴 **Blocking — the lenient schema strips the expiry.** `[z.string().toString()]` is a single literal computed key: `z.string().toString()` evaluates to `"[object Object]"`, so both lines collapse to the same key and zod's default strip mode drops every property except `access_token`. Verified: `safeParse({access_token:'t',expiry_date:123,refresh_token:'r'}).data` → `{access_token:'t'}`. The expiry the comment claims to capture is lost, and `refresh_token` is then overwritten with the literal `"refresh_token"`. Sharing mode persists credentials with no expiry and a bogus refresh token.

To accept arbitrary keys, use `.passthrough()` or `.catchall()`:
```ts
const minimumTokenResponseSchema = z.object({ access_token: z.string() }).passthrough();
```

packages/app-store/zoho-bigin/lib/CalendarService.ts
ron-x5labs, line 93: 🔴 **Blocking — wrong id passed.** `credentialId` is the credential record id (`credential.id`), but `refreshOAuthTokens`'s third parameter is `userId` and is sent to the sync endpoint as `calcomUserId`. In sharing mode the endpoint looks up the wrong user (or none) and returns wrong/empty tokens. Pass the user id:
```ts
"zoho-bigin",
credential.userId
```

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const appCredentialWebhookRequestBodySchema = z.object({
+  // UserId of the cal.com user
+  userId: z.number().int(),
+  appSlug: z.string(),
+  // Keys should be AES256 encrypted with the CALCOM_APP_CREDENTIAL_ENCRYPTION_KEY
+  keys: z.string(),
+});
+/** */
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
```
ron-x5labs, line 17: 🟡 No HTTP method guard — any method (GET, PUT, …) that presents the secret upserts credentials. Reject non-POST before doing any work:
```ts
if (req.method !== "POST") return res.status(405).json({ message: "Method not allowed" });
```

apps/web/pages/api/webhook/app-credential.ts
```diff
@@ -0,0 +1,93 @@
...
+  // Check that credential sharing is enabled
+  if (!APP_CREDENTIAL_SHARING_ENABLED) {
+    return res.status(403).json({ message: "Credential sharing is not enabled" });
+  }
+
+  // Check that the webhook secret matches
+  if (
+    req.headers[process.env.CALCOM_WEBHOOK_HEADER_NAME || "calcom-webhook-secret"] !==
```
ron-x5labs, line 25: 🟡 The secret is compared with `!==` (timing-unsafe), and `appCredentialWebhookRequestBodySchema.parse`, `symmetricDecrypt`, and `JSON.parse` all throw on bad input and surface as unhandled 500s instead of 400s. Use a constant-time compare for the secret and wrap parsing/decryption in try/catch returning a 400.

packages/app-store/salesforce/lib/CalendarService.ts
ron-x5labs, line 86: 🟡 This manual refresh runs unconditionally for ALL Salesforce users (no `APP_CREDENTIAL_SHARING_ENABLED` gate), changing the previously jsforce-abstracted refresh+persist behavior for everyone. Also `response.statusText !== "OK"` is fragile — HTTP/2 responses carry no statusText; use `!response.ok`. Note Salesforce bypasses `refreshOAuthTokens`, so self-hosters can't actually sync Salesforce credentials — inconsistent with the PR's goal.

ron-x5labs (changes requested):
## Code Review: feat: Sync app credentials between Cal.com & self-hosted platforms

### Problem
Enable self-hosters with their own app store to sync users' 3rd-party OAuth credentials into Cal.com, making the self-hosted app store the source of truth. Adds an inbound webhook (`/api/webhook/app-credential`) that decrypts and upserts credentials, and redirects each app's OAuth token-refresh path to a self-hosted endpoint when `APP_CREDENTIAL_SHARING_ENABLED`.

### Solution Reviewed
Two new utilities (`refreshOAuthTokens`, `parseRefreshTokenResponse`) centralize refresh/parse logic across ~10 calendar and video apps; OAuth helpers were moved into `_utils/oauth/`; four env vars added. The approach is reasonable, but the cross-app abstraction has several contract mismatches that break both the new sharing mode *and* (for Google) normal SaaS mode.

### Summary
Blocking. The Google Calendar change stores the full zod `safeParse` wrapper as `credential.key` instead of `.data`, which breaks every Google booking regardless of sharing mode — a regression that should not merge. The sharing path itself is broken in multiple ways: the sync endpoint returns a raw `fetch` Response while callers expect app-native shapes, the metadata lookup in the webhook is keyed by the wrong field, the lenient token schema silently strips expiry/refresh_token, and Zoho Bigin passes the credential id as the user id.

### Files Reviewed
- `apps/web/pages/api/webhook/app-credential.ts` — deeply reviewed (new inbound credential-sync endpoint; auth + data path)
- `packages/app-store/_utils/oauth/refreshOAuthTokens.ts` — deeply reviewed (new shared refresh router)
- `packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts` — deeply reviewed (new shared token parser; schema is broken)
- `packages/app-store/googlecalendar/lib/CalendarService.ts` — deeply reviewed (regression in non-sharing mode)
- `packages/app-store/zoho-bigin/lib/CalendarService.ts` — deeply reviewed (wrong id passed)
- `packages/app-store/salesforce/lib/CalendarService.ts` — deeply reviewed (unconditional behavior change, fragile error check)
- `packages/app-store/{hubspot,larkcalendar,office365calendar,office365video,webex,zohocrm,zoomvideo}/...` — lightly reviewed (mechanical refreshOAuthTokens wiring; response-shape contract varies per caller)
- OAuth helper renames + import-path updates across `add.ts`/`callback.ts` — lightly reviewed (mechanical, correct)
- `.env.example`, `packages/lib/constants.ts`, `turbo.json` — lightly reviewed (env passthrough consistent)

### Verification
- `node` zod repro of `minimumTokenResponseSchema` — confirmed both computed keys collapse to `"[object Object]"` and zod strips every key except `access_token` (expiry + refresh_token lost).
- Inspected `apps.metadata.generated.ts` + `_appRegistry.ts` — confirmed `appStoreMetadata` is keyed by `dirName`, and the canonical lookup is `appStoreMetadata[app.dirName]`, not `app.slug`.
- Inspected `schema.prisma` — confirmed `App.dirName` exists and is `@unique`.
- Type-check / build — skipped: worktree has no `node_modules` installed.

### Issues Found

#### 🔴 Blocking
1. **app-credential.ts:50** — `appStoreMetadata[app.slug]` indexes the metadata map by `App.slug` (e.g. `google-calendar`), but the map is keyed by `dirName` (`googlecalendar`). Lookup returns `undefined` → 404 for every app where slug ≠ dirName (Google, Zoom, Lark, Outlook Calendar, MS Teams).
2. **googlecalendar/CalendarService.ts:97** — `parseRefreshTokenResponse` returns the full `safeParse` result (`{success,data}`), but this stores it directly as `credential.key`. Pre-PR used `googleCredentialSchema.parse()` (unwrapped). The stored key becomes `{success:true,data:{...}}`; the next `googleCredentialSchema.parse(credential.key)` fails → every gCal booking throws. This affects ALL users, not just sharing mode.
3. **refreshOAuthTokens.ts:15** — the sharing branch returns a raw `fetch` Response, but the non-sharing branch returns each app's native shape (gaxios `res`, axios response, Hubspot token object). Callers reading `.data` (google, zoho-bigin, zohocrm) or `.access_token` (hubspot) get `undefined` → `TypeError` in sharing mode.
4. **parseRefreshTokenResponse.ts:5** — `minimumTokenResponseSchema` uses `[z.string().toString()]` as computed keys; both evaluate to `"[object Object]"` (duplicate key) and zod's default strip mode drops every property except `access_token`. The expiry field the comment claims to capture is stripped, and `refresh_token` is overwritten with the literal string `"refresh_token"`. Sharing mode therefore persists credentials with no expiry and a bogus refresh token.
5. **zoho-bigin/CalendarService.ts:93** — `refreshOAuthTokens(..., "zoho-bigin", credentialId)` passes the credential record id as the `userId` argument. In sharing mode the sync endpoint receives `calcomUserId=<credentialId>` and looks up the wrong user.

#### 🟡 Non-blocking
6. **app-credential.ts:17** — No HTTP method guard. Any method (GET, PUT, …) that presents the secret upserts credentials. Reject non-POST early.
7. **app-credential.ts:25** — Webhook secret compared with `!==` (timing-unsafe), and `zod.parse` / `symmetricDecrypt` / `JSON.parse` failures propagate as unhandled 500s rather than 400s. Wrap parsing in try/catch and use a constant-time compare for the secret.
8. **salesforce/CalendarService.ts:86** — The manual `fetch` to `login.salesforce.com` runs unconditionally for ALL Salesforce users (not gated by `APP_CREDENTIAL_SHARING_ENABLED`), changing refresh+persist behavior for everyone; `response.statusText !== "OK"` is also fragile (HTTP/2 has no statusText — use `response.ok`). Salesforce also bypasses `refreshOAuthTokens`, so self-hosters cannot sync Salesforce credentials.

### Verdict
Recommend changes before merge. The Google Calendar regression (#2) alone is a ship-stopper; the remaining sharing-path bugs (#1, #3, #4, #5) mean the feature does not work end-to-end for the apps it targets.

apps/web/pages/api/webhook/app-credential.ts
ron-x5labs, line 17: 🟡 **Non-blocking — no HTTP method guard.** Any method (GET, PUT, DELETE, etc.) that presents the secret reaches body parsing and credential upsert. Reject non-POST before doing any work:
```ts
if (req.method !== "POST") return res.status(405).json({ message: "Method not allowed" });
```
_Prior review status: persistent (code unchanged)._

apps/web/pages/api/webhook/app-credential.ts
ron-x5labs, line 25: 🟡 **Non-blocking — timing-unsafe secret comparison.** The webhook secret is compared with `!==`, which is vulnerable to timing side-channels. Use constant-time comparison:
```ts
import crypto from "crypto";
const secret = process.env.CALCOM_WEBHOOK_SECRET || "";
const header = (req.headers[process.env.CALCOM_WEBHOOK_HEADER_NAME || "calcom-webhook-secret"] as string) || "";
if (secret.length !== header.length || !crypto.timingSafeEqual(Buffer.from(header), Buffer.from(secret))) {
  return res.status(403).json({ message: "Invalid webhook secret" });
}
```
_Prior review status: persistent._

apps/web/pages/api/webhook/app-credential.ts
ron-x5labs, line 31: 🔴 **Blocking — `.parse()` throws on malformed input.** `appCredentialWebhookRequestBodySchema.parse(req.body)` throws an uncaught `ZodError` → 500 with stack trace. Use `safeParse()` and return 400:
```ts
const parsed = appCredentialWebhookRequestBodySchema.safeParse(req.body);
if (!parsed.success) return res.status(400).json({ message: "Invalid request body" });
const reqBody = parsed.data;
```
_Prior review status: persistent._

apps/web/pages/api/webhook/app-credential.ts
ron-x5labs, line 50: 🔴 **Blocking — wrong lookup key.** `appStoreMetadata` is keyed by **dirName** (`googlecalendar`, `zoomvideo`, `office365video`), but `app.slug` is the public slug (`google-calendar`, `zoom`, `msteams`). For 5 of 10 apps slug≠dirName, this returns `undefined` → 404 for all sync requests targeting those apps. Select `dirName` and use it for the lookup:
```ts
const app = await prisma.app.findUnique({
  where: { slug: reqBody.appSlug },
  select: { slug: true, dirName: true },
});
// ...
const appMetadata = appStoreMetadata[app.dirName as keyof typeof appStoreMetadata];
```
_Prior review status: persistent._

apps/web/pages/api/webhook/app-credential.ts
ron-x5labs, line 57: 🔴 **Blocking — unguarded `JSON.parse(symmetricDecrypt(...))`.** `symmetricDecrypt` throws on wrong/missing key, bad padding, or corrupt ciphertext; `JSON.parse` throws on non-JSON. No try/catch → uncaught 500. Additionally, `symmetricDecrypt` uses AES-256-CBC without a MAC, exposing a padding oracle via timing differences. Wrap in try/catch:
```ts
let keys;
try {
  keys = JSON.parse(symmetricDecrypt(reqBody.keys, process.env.CALCOM_APP_CREDENTIAL_ENCRYPTION_KEY || ""));
} catch {
  return res.status(400).json({ message: "Invalid encrypted keys" });
}
```
_Prior review status: persistent._

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
ron-x5labs, line 8: 🔴 **Blocking — broken computed property keys.** `[z.string().toString()]` and `[z.string().optional().toString()]` (line 10) both evaluate to the literal `"[object Object]"` — the second overwrites the first. The resulting schema validates only `access_token` as a string. All other fields (expiry, scope, token_type, etc.) are silently stripped by zod's default strip behavior. Use `.passthrough()` to accept additional fields:
```ts
const minimumTokenResponseSchema = z.object({
  access_token: z.string(),
}).passthrough();
```
_Prior review status: persistent._

packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts
ron-x5labs, line 26: 🟡 **Non-blocking — bogus `refresh_token` placeholder.** When `refresh_token` is missing from the response, it's set to the literal string `"refresh_token"`. This bogus sentinel gets persisted into the credential key and used in the next refresh attempt, which will fail. Consider omitting the field or handling the missing-token case explicitly.
_Prior review status: persistent._

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
```diff
@@ -0,0 +1,22 @@
+import { APP_CREDENTIAL_SHARING_ENABLED } from "@calcom/lib/constants";
+
+const refreshOAuthTokens = async (refreshFunction: () => any, appSlug: string, userId: number | null) => {
+  // Check that app syncing is enabled and that the credential belongs to a user
+  if (APP_CREDENTIAL_SHARING_ENABLED && process.env.CALCOM_CREDENTIAL_SYNC_ENDPOINT && userId) {
+    // Customize the payload based on what your endpoint requires
+    // The response should only contain the access token and expiry date
+    const response = await fetch(process.env.CALCOM_CREDENTIAL_SYNC_ENDPOINT, {
```
ron-x5labs, line 8: 🟡 **Non-blocking — no fetch error handling.** The `fetch()` call has no try/catch and no `response.ok` check. Network errors throw uncaught → 500; non-2xx responses are returned as-is and will crash callers expecting parsed JSON. Add error handling:
```ts
let response: Response;
try {
  response = await fetch(...);
} catch (err) {
  throw new Error(`Credential sync fetch failed: ${err instanceof Error ? err.message : String(err)}`);
}
if (!response.ok) throw new Error(`Sync endpoint returned ${response.status}`);
```
_Prior review status: persistent._

packages/app-store/_utils/oauth/refreshOAuthTokens.ts
ron-x5labs, line 15: 🔴 **Blocking — incompatible return shapes.** The sharing branch returns a raw `fetch` Response, but the non-sharing branch returns each caller's native shape (gaxios `res` for Google, axios response for Zoho, token object for Hubspot). In sharing mode, callers like `googlecalendar` (line 94: `res?.data`) and `zoho-bigin` (line 96: `tokenInfo.data`) crash with TypeError. Normalize the response shape — e.g. `return response.json()`.
_Prior review status: persistent._

packages/app-store/googlecalendar/lib/CalendarService.ts
ron-x5labs, line 97: 🔴 **Blocking — stores wrapper, not `.data`.** `parseRefreshTokenResponse` returns the full `safeParse` result `{success, data}`, but line 97 stores the entire wrapper as `credential.key`. On the next load, `googleCredentialSchema.parse(credential.key)` fails because expected fields are nested under `.data`. This corrupts every refreshed Google Calendar credential on second use. Store `.data`:
```ts
const parsed = parseRefreshTokenResponse(googleCredentials, googleCredentialSchema);
const key = parsed.data;
```
_Prior review status: persistent._

packages/app-store/salesforce/lib/CalendarService.ts
ron-x5labs, line 86: 🔴 **Blocking — `statusText` unreliable in HTTP/2.** `response.statusText !== "OK"` rejects legitimate responses because HTTP/2 responses carry no `statusText` (empty string on 200). Use `response.ok` or `response.status`:
```ts
if (!response.ok) throw new HttpError({ statusCode: response.status, message: response.statusText });
```
_Prior review status: persistent._

packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -60,6 +72,32 @@ export default class SalesforceCalendarService implements Calendar {
...
+
+    const accessTokenJson = await response.json();
+
+    const accessTokenParsed = parseRefreshTokenResponse(accessTokenJson, salesforceTokenSchema);
+
+    if (!accessTokenParsed.success) {
```
ron-x5labs, line 92: 💡 **Suggestion — unreachable dead code.** `parseRefreshTokenResponse` has `if (!refreshTokenResponse.success) throw new Error(...)` at lines 21-23 of the util, covering both the sharing and non-sharing paths. The function never returns `{success: false}` — it always throws. So this `if (!accessTokenParsed.success)` guard is unreachable dead code. Same dead pattern at `zoomvideo/VideoApiAdapter.ts:104`. Remove the guard.
_New finding._

packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -60,6 +72,32 @@ export default class SalesforceCalendarService implements Calendar {
...
+      return Promise.reject(new Error("Invalid refreshed tokens were returned"));
+    }
+
+    await prisma.credential.update({
```
ron-x5labs, line 96: 🔴 **Blocking — missing `prisma` import.** `prisma.credential.update()` is called here but `prisma` is never imported in this file. Every other CalendarService/VideoApiAdapter that uses prisma imports it (`import prisma from "@calcom/prisma"` — confirmed in googlecalendar:11, zoho-bigin:6, zoomvideo:4, office365calendar:9, office365video:5). This fails type-checking and would throw `ReferenceError` at runtime. Add:
```ts
```
_New finding (not in prior reviews)._

packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -60,6 +72,32 @@ export default class SalesforceCalendarService implements Calendar {
...
+      where: { id: credential.id },
+      data: { key: { ...accessTokenParsed.data, refresh_token: credentialKey.refresh_token } },
+    });
+
     return new jsforce.Connection({
```
ron-x5labs, line 101: 🔴 **Blocking — uses stale access token.** The `jsforce.Connection` (line 106) is created with `credentialKey.access_token` (the original, potentially expired token) instead of the freshly-refreshed `accessTokenParsed.data.access_token`. The new token is persisted to DB (line 96) but never used by the current connection — it only takes effect on the next request. Use the fresh token:
```ts
accessToken: accessTokenParsed.data.access_token,
```
_Prior review status: persistent._

packages/app-store/zoho-bigin/lib/CalendarService.ts
ron-x5labs, line 93: 🔴 **Blocking — passes credential ID instead of user ID.** `credentialId` (set from `credential.id` at line 53) is the credential record PK, but `refreshOAuthTokens`'s third parameter is `userId` and is sent to the sync endpoint as `calcomUserId`. The sync endpoint would receive the wrong ID and fail the user lookup (`prisma.user.findUnique({ where: { id: <credentialId> } })`). Use `credential.userId`:
```ts
credential.userId
```
_Prior review status: persistent._

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
ron-x5labs, line 92: 🔴 **Blocking — wrong app slug.** Passes `"zoomvideo"` (the dirName) as `appSlug`, but the webhook resolves apps via `prisma.app.findUnique({ where: { slug } })` expecting `"zoom"` (the `App.slug`). A sync callback carrying `"zoomvideo"` fails the DB lookup → 404. Use `"zoom"`.
_Prior review status: persistent._

packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -94,7 +101,7 @@ const zoomAuth = (credential: CredentialPayload) => {
       }
     }
     // We check the if the new credentials matches the expected response structure
-    const parsedToken = zoomRefreshedTokenSchema.safeParse(responseBody);
+    const parsedToken = parseRefreshTokenResponse(responseBody, zoomRefreshedTokenSchema);
```
ron-x5labs, line 104: 💡 **Suggestion — unreachable dead code.** `parseRefreshTokenResponse` has `if (!refreshTokenResponse.success) throw new Error(...)` at lines 21-23 of the util, covering both sharing and non-sharing paths. The function never returns `{success: false}`. So the `if (!parsedToken.success)` guard at line 108 is unreachable dead code. Remove the guard.
_New finding._

ron-x5labs (changes requested):
## Code Review: App credential syncing for self-hosted deployments

### Problem
This PR adds an app-credential-syncing mechanism for self-hosted Cal.diy instances. A webhook endpoint (`app-credential.ts`) receives encrypted OAuth tokens from a sync source and upserts them into the credential table. A `refreshOAuthTokens` utility and `parseRefreshTokenResponse` helper gate OAuth refresh calls through a central sync endpoint when `APP_CREDENTIAL_SHARING_ENABLED` is set.

### Solution Reviewed
New files: `app-credential.ts` (webhook handler), `refreshOAuthTokens.ts` (sync gate), `parseRefreshTokenResponse.ts` (token validator). Modified refresh paths in 7 app integrations: googlecalendar, salesforce, zoho-bigin, zoomvideo, office365video, office365calendar, plus hubspot/lark/webex via shared utils.

### Summary
Multiple blocking issues across all new files and several modified callers. The core syncing mechanism is broken: the webhook can't resolve app metadata for 5/10 apps (slug≠dirName), the zod schema collapses to validating only `access_token`, return shapes are incompatible between sharing and non-sharing modes, and Google Calendar stores a wrapper object instead of parsed data. Salesforce additionally has a missing `prisma` import that would fail compilation. Cannot merge in current state.

### Prior Review Status
Two prior reviews from this skill exist (IDs 4920350113, 4943367646). All previously reported findings remain **persistent** (code unchanged since last review). New findings in this review: missing `prisma` import in salesforce (line 96), and unreachable dead-code guards (salesforce:92, zoomvideo:104).

### Files Reviewed
- `apps/web/pages/api/webhook/app-credential.ts` — deeply reviewed (new, high-risk: auth/crypto)
- `packages/app-store/_utils/oauth/parseRefreshTokenResponse.ts` — deeply reviewed (new, high-risk: data validation)
- `packages/app-store/_utils/oauth/refreshOAuthTokens.ts` — deeply reviewed (new, high-risk: network/security)
- `packages/app-store/googlecalendar/lib/CalendarService.ts` — deeply reviewed (modified, high-risk: OAuth)
- `packages/app-store/salesforce/lib/CalendarService.ts` — deeply reviewed (modified, high-risk: OAuth)
- `packages/app-store/zoho-bigin/lib/CalendarService.ts` — deeply reviewed (modified, medium-risk)
- `packages/app-store/zoomvideo/lib/VideoApiAdapter.ts` — deeply reviewed (modified, medium-risk)
- `packages/app-store/office365video/lib/VideoApiAdapter.ts` — lightly reviewed (same pattern)
- `packages/app-store/office365calendar/lib/CalendarService.ts` — lightly reviewed (same pattern)
- `packages/lib/constants.ts` — lightly reviewed (env var config)
- Other moved/renamed files — lightly reviewed (import path changes only)

### Verification
- `grep` for `import prisma` across affected files — confirmed salesforce CalendarService.ts is the only file missing the import that all 5 others have
- `grep` for `slug`/`dirName` in `_metadata.ts`/`config.json` — confirmed 5 apps have slug≠dirName (google-calendar, lark-calendar, office365-calendar, msteams, zoom)
- Type check / build — skipped (requires full monorepo setup; but missing `prisma` import is a definite compile error)

### Verdict
Recommend changes before merge — the syncing mechanism is broken in multiple independent ways: metadata lookup fails for half the apps, the zod schema doesn't capture expiry, return shapes crash non-fetch callers, Google Calendar credential corruption on second use, and Salesforce won't compile due to a missing `prisma` import.

PR #11073, opened 2023-09-01
fix: #10973 back button not working as expected in embed

### LINKED ISSUE #10973: Booking Timeout Broken
### Issue Summary

There is a period, apparently 5 minutes in length, where after selecting a day and time for a booking, that time is  locked out or "reserved", and not available to be chosen by other users.  Unfortunately, it is also not available to be chosen again by the same user who "reserved" it.   Consequently, having arrived at the "Confirm" page, a user who backs up to reconsider alternative event-types or timeslots will find that their initial timeslot choice is  is now unavailable.   If they try  the same day  with the very same, or another, event-type, the time that they just chose will not be available.   The timeslot doesn't become available again for 5 minutes, for anybody.

This is very confusing.   What happened to that time that was available a few seconds ago?  There is no indication that the availability is only temporarily blocked, and the timeslot will become available again in 5 minutes, perhaps less.   Users will think the booking system is broken.

### Steps to Reproduce

1. Select  an event type.

2. Click a date, then click a time.

3. On the Confirm screen, click "Back".  This will take you back to the screen where you
just picked a time.  (You can also get back to this screen other ways.)  The time you just selected 
will no longer be shown as available, even though you just selected it.   The only way it will be available is if you come back after 5 minutes.

By doing these three steps repeatedly,  picking different events/timeslots each time, it seems to be possible for one user to lock out a lot of timeslots, at least for the timeout period, without actually "Confirming" anything.

### Expected Results

Possibilities:
(1) No reservation/lockout period.  Everybody is in a race to click Confirm first.  If the user clicks "Confirm", there is the possibility that the chosen timeslot, or part of it, has been grabbed by somebody else, especially if the user has been too slow.  Note that the system has to deal with this possibility anyway because the reservation is some fixed time, and after that time, clicking Confirm can fail because someone else grabbed the timeslot, or part of it.

(2) Reservation period, but it doesn't apply against the person who made the reservation.  If someone backs up  from the Confirm screen to reconsider times, days, or other event-types, that user can still get "their" orignally-selected time slot, within the "reservation" limit. 

(3) Reservation period, but the reservation expires if the user who made the reservation navigates
away from the page with the "Confirm" button.   Things are as they were, and the user (or anybody else) can now book that time-slot.

### Actual Results

The reservation expires after 5 minutes, and a user who navigates away from the "Confirm" button
must wait the 5 minutes to be able to get that time again.  During that 5 minutes, the timeslot is not
available again for the event type selected (or any other event-type) to anybody, including the user who initiated the reservation.

vijayraghav-io:
## What does this PR do?

Fixes #10973 

Below is the brief description or reproduction steps :
In **embed** scenario, 
1. Select an event type.
2. Select a date and then click on a time slot.
3. On the Confirm screen, click "Back". This will take you back to the screen where you just picked a time. The time you just selected will no longer be shown as available, even though you just selected it. The only way it will be available is if you come back after 5 minutes.

Video : Before fix : (Issue)
Slot is selected for Sep6th 10AM, Click on back button, same slot is **Not** selectable
https://www.loom.com/share/de09f1f4d43640c8989b929becf544d9?sid=5558af13-6384-4ef5-8929-bdb340307009

Video : After fix :
Slot is selected for Sep6th 10AM, Click on back button, same slot is selectable
https://www.loom.com/share/b4c8110bed594c31aaa188bc2f37eccc?sid=cc9ae96c-471a-4fba-9133-c9eb60c59a00
 

Root cause of issue:
In case of embed, the cookies are not set in requests to trpc.
The trpc endpoint "slots.removeSelectedSlotMark" relies on cookies in request to get the uid. And uses this uid to delete the selection mark in db.
Hence for embed case, it's (endpoint) not able to get uid from cookie, and the selection mark is not removed. So the slot is still reserved (even after user clicks on back button)

Implementaion / Fix :
uid is passed as an optional argument to the endpoint "slots.removeSelectedSlotMark"

## Type of change
- [v] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?
Refer above for steps to reproduce the issue and verify its working fine.

## Mandatory Tasks
- [v] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

### DIFF
--- a/packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
+++ b/packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
@@ -45,8 +45,18 @@ type BookEventFormProps = {
 type DefaultValues = Record<string, unknown>;
 
 export const BookEventForm = ({ onCancel }: BookEventFormProps) => {
+  let uid = "";
   const reserveSlotMutation = trpc.viewer.public.slots.reserveSlot.useMutation({
-    trpc: { context: { skipBatch: true } },
+    trpc: {
+      context: {
+        skipBatch: true,
+      },
+    },
+    onSuccess: (data) => {
+      // uid is required to pass as input to removeSelectedSlotMark endpoint
+      // as in case of embed, uid is not passed in cookie
+      uid = data?.uid;
+    },
   });
   const removeSelectedSlot = trpc.viewer.public.slots.removeSelectedSlotMark.useMutation({
     trpc: { context: { skipBatch: true } },
@@ -82,7 +92,7 @@ export const BookEventForm = ({ onCancel }: BookEventFormProps) => {
 
     return () => {
       if (eventType) {
-        removeSelectedSlot.mutate();
+        removeSelectedSlot.mutate({ uid: uid });
       }
       clearInterval(interval);
     };
--- a/packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
+++ b/packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
@@ -45,7 +45,7 @@ type BookEventFormProps = {
 type DefaultValues = Record<string, unknown>;
 
 export const BookEventForm = ({ onCancel }: BookEventFormProps) => {
-  let uid = "";
+  const uid = useRef("");
   const reserveSlotMutation = trpc.viewer.public.slots.reserveSlot.useMutation({
     trpc: {
       context: {
@@ -55,7 +55,7 @@ export const BookEventForm = ({ onCancel }: BookEventFormProps) => {
     onSuccess: (data) => {
       // uid is required to pass as input to removeSelectedSlotMark endpoint
       // as in case of embed, uid is not passed in cookie
-      uid = data?.uid;
+      uid.current = data?.uid;
     },
   });
   const removeSelectedSlot = trpc.viewer.public.slots.removeSelectedSlotMark.useMutation({
@@ -92,7 +92,7 @@ export const BookEventForm = ({ onCancel }: BookEventFormProps) => {
 
     return () => {
       if (eventType) {
-        removeSelectedSlot.mutate({ uid: uid });
+        removeSelectedSlot.mutate({ uid: uid.current });
       }
       clearInterval(interval);
     };
--- a/packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
+++ b/packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
@@ -34,6 +34,7 @@ import { Alert, Button, EmptyScreen, Form, showToast } from "@calcom/ui";
 import { Calendar } from "@calcom/ui/components/icon";
 
 import { useBookerStore } from "../../store";
+import { useSlotReservationId } from "../../useSlotReservationId";
 import { useEvent } from "../../utils/event";
 import { BookingFields } from "./BookingFields";
 import { FormSkeleton } from "./Skeleton";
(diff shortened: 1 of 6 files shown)


packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
```diff
@@ -45,8 +45,18 @@ type BookEventFormProps = {
 type DefaultValues = Record<string, unknown>;
 
 export const BookEventForm = ({ onCancel }: BookEventFormProps) => {
+  let uid = "";
```
Udit-takkar, line 48: why not a state variable here ?

vijayraghav-io: This 'uid' info is very much local to BookEventForm (that too for embed scenario only) and not used in any other Component (also dint forsee future use), that's why did not make it as State variable.

Udit-takkar: you could also use useRef

Udit-takkar (commented):
@vijayraghav-io @hariombalhara 

I tried selecting a slot in the booking page and then opening a new browser and i still see my selected slot as available. does reserve slots work in case of embeds?  

i used one of the embed from http://127.0.0.1:3100/embed/

cd packages/embeds/embed-core
yarn dev

packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
vijayraghav-io: yes useRef can be used, updated the same

vijayraghav-io:
@Udit-takkar @hariombalhara , yes reserve slot is working in embed scenario (given that second attempt to book same slot is within NEXT_PUBLIC_MINUTES_TO_BOOK or 5mins)

Udit-takkar (dismissed):
LGTM.

Will test https://github.com/calcom/cal.com/pull/11073#pullrequestreview-1610491951 on prod once this is live

packages/features/bookings/Booker/useSlotReservationId.ts
```diff
@@ -0,0 +1,15 @@
+// TODO: It would be lost on refresh, so we need to persist it.
+// Though, we are persisting it in a cookie(`uid` cookie is set through reserveSlot call)
+// but that becomes a third party cookie in context of embed and thus isn't accessible inside embed
+// So, we need to persist it in top window as first party cookie in that case.
+let slotReservationId: null | string = null;
+
+export const useSlotReservationId = () => {
```
hariombalhara, line 7: I moved the slot reservation Id logic here @vijayraghav-io @Udit-takkar. This is so that we can improve upon this so that it's not lost on refresh.

hariombalhara (approved):
Tested: https://www.loom.com/share/b0184fd963284ae5a835c5a7724d4226
LGTM!!

packages/features/bookings/Booker/useSlotReservationId.ts
hariombalhara: I didn't use useRef because the goal is to keep it persisted as much as we can in memory. Using useRef would cause it to be lost when BookEventForm is unmounted/remounted.

packages/features/bookings/Booker/useSlotReservationId.ts
```diff
@@ -0,0 +1,15 @@
+// TODO: It would be lost on refresh, so we need to persist it.
+// Though, we are persisting it in a cookie(`uid` cookie is set through reserveSlot call)
+// but that becomes a third party cookie in context of embed and thus isn't accessible inside embed
```
hariombalhara, line 3: I plan to implement a storage layer for embed that would persist data in top window(embedder page) on behalf of the iframe so that things stay consistent b/w embed and non-embed.

hariombalhara:
Thankyou @vijayraghav-io 🙏 Great work !!

vijayraghav-io:
Thankyou! @hariombalhara , Great Improvement too! 🙏

PR #11078, opened 2023-09-01
fix: Revert #10366: Undoing then redo fix differently

emrysal:
## What does this PR do?

Reverts setting the selectedDate to null which causes the month to be broken downstream.


sean-brydon (approved):
Works - Looks like you accidentally fixed a bug in the process :D

PR #11080, opened 2023-09-01
fix: support subteam in org redirection

leog:
## What does this PR do?

Using the slug "support" for a subteam within an org was clashing with our "Support" redirection which opened up Intercom in the event types page. This PR fixes that.

## Type of change

## How should this be tested?

1. Create an org
2. Create a subteam with slug "support"
3. Visit public profile of the created subteam
4. You should be able to see the profile page and not be redirected to event types page with intercom opened

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -450,6 +450,13 @@ const nextConfig = {
       },
       {
         source: "/support",
+        missing: [
+          {
+            type: "header",
+            key: "host",
+            value: orgHostPath,
+          },
+        ],
         destination: "/event-types?openIntercom=true",
         permanent: true,
       },


apps/web/next.config.js
```diff
@@ -450,6 +450,13 @@ const nextConfig = {
       },
       {
         source: "/support",
+        missing: [
+          {
+            type: "header",
+            key: "host",
+            value: orgHostPath,
+          },
+        ],
```
leog, line 459: Making sure there is no org context to apply the redirection

joeauyeung (approved):
If my assumption is right, if the header `host` is not null then it'll continue to the proper org URL else it'll continue as normal when going to `/support.

PeerRich (approved):

PeerRich:
merged. we can try on app.cal.dev

PR #11082, opened 2023-09-01
fix: Remove background colour from avatar 

sean-brydon:
Hard to see but the avatar had a tiny gap of margin 

We dont use the background for this anymore so its not needed.

Before: 
CleanShot 2023-09-01 at 16 42 38

After: 
CleanShot 2023-09-01 at 16 41 26


leog (approved):
Great, thanks for taking care of this.

PR #11111, opened 2023-09-03

### LINKED ISSUE #11106: Add placeholder design to the dates to override panel.

image

image

razmishra:
https://github.com/calcom/cal.com/assets/127204999/082f4bce-1611-4534-8e6b-04b43944c61a

fixes #11106 

What I did-
I have added an else condition in DateOverrideInputDialog.tsx file such that, if date is not selected then this dialog box will be visible otherwise it will show the selected date.


shammirbaig:
I guess border line in between those two main divs is missing the video

razmishra:
> I guess border line in between those two main divs is missing the video

Added the separator line between both divs.

packages/features/schedules/components/DateOverrideInputDialog.tsx
```diff
@@ -192,6 +188,30 @@ const DateOverrideForm = ({
...
+            width: "35%",
+            margin: "50px 50px",
+          }}>
+          <div
+            style={{
+              padding: "10px",
+              color: "#999",
+              textAlign: "center",
+              fontWeight: "bold",
+            }}>
```
sean-brydon, line 211: Please use tailwind

sean-brydon (dismissed):

packages/features/schedules/components/DateOverrideInputDialog.tsx
```diff
@@ -192,6 +188,12 @@ const DateOverrideForm = ({
             <DialogClose onClick={onClose} />
           </div>
         </div>
+      ) : (
+        <div className="flex items-center justify-center">
+          <div className="rounded-8 h-50 ml-8 w-80 border-2 border-dashed border-gray-500 p-6 brightness-50 filter">
```
PeerRich, line 193: have you tested this with light mode too?

PeerRich:
i dont think we need an empty screen here. why dont we make the dialog smaller and animate it when it gets larger after the selection?

PeerRich (dismissed):
i dont think we need an empty screen here. why dont we just make the initial dialog smaller and then expand it after the selection?

razmishra:
> i dont think we need an empty screen here. why dont we just make the initial dialog smaller and then expand it after the selection?

That's also a good idea to go with. Let me make the changes accordingly. Thank you for the suggestion.

packages/ui/components/dialog/Dialog.tsx
```diff
@@ -89,7 +89,7 @@ export const DialogContent = React.forwardRef<HTMLDivElement, DialogContentProps
         <DialogPrimitive.Content
           {...props}
           className={classNames(
-            "fadeIn bg-default scroll-bar fixed left-1/2 top-1/2 z-50 w-full max-w-[22rem] -translate-x-1/2 -translate-y-1/2 rounded-md text-left shadow-xl focus-visible:outline-none sm:align-middle",
+            "fadeIn bg-default scroll-bar fixed left-1/2 top-1/2 z-50 max-w-[22rem] -translate-x-1/2 -translate-y-1/2 rounded-md text-left shadow-xl focus-visible:outline-none sm:align-middle",
```
PeerRich, line 92: this could have side effects and need to be careful with

razmishra: > this could have side effects and need to be careful with

I was into the same problem while making changes over here.
The issue was that the dialog-creation class was being created dynamically and the <form> was taking the width of the dialog-creation.
<form> had datepPicker and an area where selected dates will be shown. That's why I made a change that dialog-creation should not take full width always. It should take full width if there is content to display.
And this solved our issue. 
Initially, we had nothing in the right side of datePicker that's why the size of the dialog-creation gets reduced to only the width of datePicker.

Most importantly, it is not creating size issues over other places because it will take the full size if there is content to display. I hope it makes sense.
By the way I did checked with another popups available in the app, all were working fine as expected.

shivamklr (changes requested):
This has broken the functionality on mobile screens. The modal is not usable when we click on a date. Kindly revisit the changes for mobile screens.
image

PR #11123, opened 2023-09-04
fix: Require an 'Email' placeholder in the 'Add guest' field.

### LINKED ISSUE #11094: Require an 'Email' placeholder in the 'Add guest' field."
Require an 'Email' placeholder in the 'Add guest' field."

When I click on "Add guest," it's not clear what information to input, whether it's a name or an email. However, when I hover over the cross mark, it displays "Remove email," indicating that an email should be added. The placeholder in the "Add guest" text field should include the label "Email."

image

Answer:
Simply add the placeholder "Email" to the "Add Guest" field input.

Madhan13K:
## What does this PR do?

This add the Email placeholder to Add Guests

Fixes #11094

## Type of change

## Mandatory Tasks


packages/features/bookings/lib/getBookingFields.ts
```diff
@@ -251,6 +251,7 @@ export const ensureBookingInputsHaveSystemFields = ({
       type: "multiemail",
       editable: "system-but-optional",
       name: "guests",
+      defaultPlaceholder: "Enter Email...",
```
hariombalhara, line 254: We use translation keys here. It has to go first in en/common.json and then key would be used here. Also, **Enter Email...** doesn't look good. How about using "Email Address" which has a translation _email_address_

cc @PeerRich

hariombalhara (dismissed):
Blocking on a review from @PeerRich

packages/features/bookings/lib/getBookingFields.ts
Madhan13K: Hi @hariombalhara, I will change the text and send it. Thanks for the update.

Madhan13K: I have fixed the issue...

alannnc (approved):
Thank you for your contribution!

PR #11140, opened 2023-09-04
fix: getSchedule appearing twice error corrected #11138

### LINKED ISSUE #11138: [CAL-2404] getSchedule called twice on booking page
Due to recent changes getSchedule get called twice on booking page, we need to investigate further down.

related changes to have in mind: [https://github.com/calcom/cal.com/pull/9815](https://github.com/calcom/cal.com/pull/9815)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2404](https://linear.app/calcom/issue/CAL-2404/getschedule-called-twice-on-booking-page)</sub>

FaizAdhikari:
## What does this PR do?

getSchedule was calling twice on booking page and it's corrected in this PR

Fixes #11138 

## Type of change

The function buildWhereClause is calling twice in the code . This is because the if statement and the else statement both call the function.

To correct this, I have moved the call to the buildWhereClause function outside of the if statement. This will ensure that the function is only called once, regardless of whether the if statement is true or false.


alannnc:
I've publish my solution to this issue on https://github.com/calcom/cal.com/pull/11141 if you want to take a look of that.

Before closing this PR, your fix attacks something else you found? So we can adjust accordingly.

alannnc:
Gonna close this PR but if you think what you found it's a bug please file an issue with full report details.

PR #11153, opened 2023-09-05
chore: removed gravatar

### LINKED ISSUE #11151: remove gravatar alltogether and replace default image with static one
we are making unnecessary calls to gravatar and in general, we should reduce dependencies. 

instead we should be showing a static default avatar photo and urge people to upload their own image

PeerRich:
CleanShot 2023-09-05 at 18 11 09@2x

CleanShot 2023-09-05 at 18 11 21@2x

CleanShot 2023-09-05 at 18 11 31@2x

CleanShot 2023-09-05 at 18 11 40@2x


sean-brydon (approved):
Tested and works fine. Nicer fallback than what we had can't lie. 

We should follow up with a console PR to remove the one of the fallback.

PR #11158, opened 2023-09-05
chore: unify org data access

leog:
## What does this PR do?

**TL;DR**: Rely on session org data for client through existing hook

Before this refactor, org data was being accessed through a couple of ways in the client side, using different asynchronous calls to get the same basic org info. This refactor unifies org data access to rely on user session. This tackles in-app data, not public pages which rely heavily on browser URL through router host.

## Type of change

## How should this be tested?

App should work as always, maybe a bit snappier as we rely on session data which is available more promptly than using `getBrand` on client as well we had before.


joeauyeung (dismissed):
What a great idea and LGTM

PeerRich:
waiting for @zomars to approve adding more to session

zomars (dismissed):
Way more clean! Nice work @leog 

LGTM

PeerRich:
E2E failing @leog

keithwillcode (approved):

PR #11181, opened 2023-09-06
fix: paypal save event type and payment

alannnc:
## What does this PR do?

- Fixes capture return waiting for body when it's not there anymore.
- Fixes save on event type price | currency on event type for stripe and paypal.

## How should this be tested?

- What are the minimal test data to have? Paypal app in an event type.

## Mandatory Tasks

## Checklist


packages/app-store/paypal/lib/Paypal.ts
```diff
@@ -128,6 +128,7 @@ class Paypal {
       if (captureResult.ok) {
         const result = await captureResult.json();
         if (result.body.status === "COMPLETED") {
+        if (result?.status === "COMPLETED") {
```
zomars, line 131: Is this expected? @alannnc

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
```diff
@@ -241,13 +241,13 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
   }
 
-  if (input?.price || input.metadata?.apps?.stripe?.price) {
-    data.price = input.price || input.metadata?.apps?.stripe?.price;
+  if (input.metadata?.apps?.stripe?.price) {
+    data.price = input.metadata?.apps?.stripe?.price;
     const paymentCredential = await ctx.prisma.credential.findFirst({
       where: {
         userId: ctx.user.id,
         type: {
```
zomars, line 249: Should we used appId instead? since type is deprecated.

https://github.com/calcom/cal.com/blob/main/packages/prisma/schema.prisma#L128-L129

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
```diff
@@ -241,13 +241,13 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
     }
   }
 
-  if (input?.price || input.metadata?.apps?.stripe?.price) {
-    data.price = input.price || input.metadata?.apps?.stripe?.price;
+  if (input.metadata?.apps?.stripe?.price) {
```
zomars, line 244: Also instead of multiple conditionals, can we abstract this into a function? getPaymentCredential from metadata or something like that.

zomars (changes requested):
Have some code quality comments before merging 🙏🏽

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
ubinatus: Hi! I'm currently working in adding `MercadoPago` payment integration to Cal.com (work in progress [here](https://github.com/ubinatus/cal.com/tree/feature/mercadopago). Should create a PR later today). 

Regarding @zomars  comment, I've thought about the following:

```ts
  // TODO: This could be a function to abstract the payment setters in the eventType as only one is allowed.
  const paymentAppIds = ["stripe", "paypal", "mercadopago"];
  if (input?.price) {
    const paymentCredentials = await ctx.prisma.credential.findFirst({
      where: {
        userId: ctx.user.id,
        appId: {
          in: paymentAppIds,
        },
      },
      select: {
        type: true,
        key: true,
      },
    });

    paymentAppIds.forEach((paymentAppId) => {
      if (
        paymentCredentials.appId === paymentAppId &&
        input.metadata?.apps &&
        input.metadata?.apps[paymentAppId] &&
        input.metadata?.apps[paymentAppId].enabled
      ) {
        data.price = input.price || input.metadata?.apps[paymentAppId].price;
        if (paymentAppId === "stripe") {
          const { default_currency } = stripeDataSchema.parse(paymentCredential.key);
          data.currency = default_currency;
        } else {
          data.currency = input.metadata?.apps[paymentAppId].currency.toLowerCase();
        }
      }
    });
  }
```

That should only set the price/currency from the enabled payment app, since the user could now have more than 1 payment `Credential` but not necessarily the first found is the one that belongs to the `EventType`.

Best,
JA

ubinatus: Made some cleaning, here's the updated code:

```ts
  const paymentAppIds = ["stripe", "paypal", "mercadopago"];
  if (input.metadata?.apps) {
    const paymentCredentials = await ctx.prisma.credential.findMany({
      where: {
        userId: ctx.user.id,
        appId: {
          in: paymentAppIds,
        },
      },
      select: {
        type: true,
        appId: true,
        key: true,
      },
    });

    paymentCredentials.forEach((paymentCredential) => {
      const appId = paymentCredential.appId! as keyof typeof input.metadata.apps;
      const appMetadata = input.metadata?.apps ? input.metadata?.apps[appId] : null;
      if (appMetadata?.enabled) {
        data.price = input.price || appMetadata.price;
        if (appId === "stripe") {
          const { default_currency } = stripeDataSchema.parse(paymentCredential.key);
          data.currency = default_currency;
        } else {
          data.currency = appMetadata.currency.toLowerCase();
        }
      }
    });
  }
  ```

packages/app-store/paypal/lib/Paypal.ts
alannnc: Yes, somehow new fetcher doesn't retrieve body this time.

alannnc:
@zomars Changes were addressed and the rest will be handled on MercadoPago PR, can we get an approve and merge?

zomars (approved):

PR #11182, opened 2023-09-06

### LINKED ISSUE #11106: Add placeholder design to the dates to override panel.

razmishra:
## What does this PR do?

Fixes #11106 

what I did-
I removed w-full class from dialog.tsx file and added condition based w-full class in dateOverrideInputDialog.tsx file.
This fixes the issue and modal is responsive on smaller screen also.

## Mandatory Tasks


packages/ui/components/dialog/Dialog.tsx
```diff
@@ -89,7 +89,7 @@ export const DialogContent = React.forwardRef<HTMLDivElement, DialogContentProps
         <DialogPrimitive.Content
           {...props}
           className={classNames(
-            "fadeIn bg-default scroll-bar fixed left-1/2 top-1/2 z-50 w-full max-w-[22rem] -translate-x-1/2 -translate-y-1/2 rounded-md text-left shadow-xl focus-visible:outline-none sm:align-middle",
```
PeerRich, line 92: will removing w-full have side-effects @sean-brydon or @Udit-takkar ?

razmishra: Untill anyone of them reply, allow me to explain what w-full was doing generally..
w-full was allowing to take full width over every screen whether its small medium or large device.
Since dialog box is one if the child div that's why dislog was also taking full width irrespective of screen size.
In the similar manner, form component(which allow to overwrite date and has two divs, one to show date picker and one to show the selected dates) is child div of dialog and is taking full width, due to this right div (div to show dates selected) place is always visible even if dates are not selected.

Therefore to reduce size of form div, we have to fix the default size of the dialog and to fix the dialog we should fix its parent class.
Therefore what I did, I removed the default w-full class (which was by default for every device), and said if the screen is mobile then only take full width otherwise don't.

In this manner, the dialog will take the width if there is content to display, otherwise, it won't take any extra empty space.
This is how I came to fix it.

I hope it makes sense, by the way, I'm open to feedback and will fix this if needed again.

PeerRich: lemme test this and merge asap! 🙏 thank you

PeerRich (changes requested):
some dialogs are smaller now than before


before:
CleanShot 2023-09-07 at 17 31 17@2x


after:
CleanShot 2023-09-07 at 17 30 54@2x


or

before:
CleanShot 2023-09-07 at 17 32 08@2x


after:
CleanShot 2023-09-07 at 17 31 53@2x

emrysal:
Agreed, removing w-full from all dialogs is not the right fix, we need to overwrite just for date overrides

razmishra:
> some dialogs are smaller now than before
> 
> before: CleanShot 2023-09-07 at 17 31 17@2x
> 
> after: CleanShot 2023-09-07 at 17 30 54@2x
> 
> or
> 
> before: CleanShot 2023-09-07 at 17 32 08@2x
> 
> after: CleanShot 2023-09-07 at 17 31 53@2x

Oops... its actually weird, let me see what else is approach to fix this without changing w-full class.

Udit-takkar (commented):
This PR https://github.com/calcom/cal.com/pull/11250 has been merged

PR #11185, opened 2023-09-06
feat: Calendar Cache

zomars:
## What does this PR do?

- Implements a basic third-party calendar cache that lasts for a minute (for v1)
- First call will be as slow as now, but subsequent calls should be nearly instant.
- Migrates gCal CalendarService from callbacks to async/awaits

## Possible downsides

- If a profile is a high traffic one, we might run into a race condition where the cache is de-synced with the actual availability and get a "Could not book the meeting." error.

## TODO

- [x] Implement a cron for cleanup expired caches
- [x] Implement an opt-in feature mechanism 
- [ ] Implement a way to bypass cache altogether (either via request header or search parameter)
- [x] Make the cache more aggressive (save full months, even if only a day was requested)

## Type of change

## How should this be tested?

- Run migrations
- Install Google Calendar
- Enable some calendars
- As an ADMIN, enable the calendar-cache feature flag
- Go to a booking page
- Check if CalendarCache table is being populated correctly
- Check availability works as normal

## Mandatory Tasks


PeerRich:
ouhh exciting

emrysal:
The problem is that the current bottleneck is not the calendars - the bottleneck is the database right now 😓

zomars:
> The problem is that the current bottleneck is not the calendars - the bottleneck is the database right now 😓

This is not an attempt to fix that bottleneck. This is another mechanism altogether. But agreed.

zomars (commented):
Ready for review and manual testing

PeerRich:
> If a profile is a high traffic one, we might run into a race condition where the cache is de-synced with the actual availability and get a "Could not book the meeting." error.


for this, cant we use our reservation queue system?

we need to prevent "Could not book the meeting." errors at all cost

zomars:
>  > If a profile is a high traffic one, we might run into a race condition where the cache is de-synced with the actual availability and get a "Could not book the meeting." error.
> 
> 
> for this, cant we use our reservation queue system?
> 
> we need to prevent "Could not book the meeting." errors at all cost

Could alleviate but we can't really guarantee

zomars (commented):
I do think this is ready to merge. Even if we deploy to prod it won't be active until we activate it via feature flag.

apps/web/vercel.json
```diff
@@ -0,0 +1,8 @@
+{
+  "crons": [
+    {
+      "path": "/api/cron/calendar-cache-cleanup",
+      "schedule": "0 5 * * *"
```
alannnc, line 5:
```suggestion
      // Every day At 05:00.” 
      "schedule": "0 5 * * *"
```

alannnc: A little hint to the eye

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -24,13 +25,45 @@ interface GoogleCalError extends Error {
   code?: number;
 }
 
+const ONE_MINUTE_MS = 60 * 1000;
+const CACHING_TIME = ONE_MINUTE_MS;
+
+/** Expand the start date to the start of the month */
+function getTimeMin(timeMin: string) {
```
alannnc, line 32:
```suggestion
function getStartDateOfTheMonth(timeMin: string) {
```

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -24,13 +25,45 @@ interface GoogleCalError extends Error {
...
+  const dateMin = new Date(timeMin);
+  return new Date(dateMin.getFullYear(), dateMin.getMonth(), 1, 0, 0, 0, 0).toISOString();
+}
```
alannnc, line 35:
```suggestion
function getStartDateOfTheMonth(dateStart: string) {
  const dateMin = new Date(dateStart);
  return new Date(dateMin.getFullYear(), dateMin.getMonth(), 1, 0, 0, 0, 0).toISOString();
}
```

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -24,13 +25,45 @@ interface GoogleCalError extends Error {
...
+
+/** Expand the end date to the end of the month */
+function getTimeMax(timeMax: string) {
+  const dateMax = new Date(timeMax);
+  return new Date(dateMax.getFullYear(), dateMax.getMonth() + 1, 0, 0, 0, 0, 0).toISOString();
+}
```
alannnc, line 41:
```suggestion
function getEndTimeOfTheMonth(endTime: string) {
  const dateMax = new Date(endTime);
  return new Date(dateMax.getFullYear(), dateMax.getMonth() + 1, 0, 0, 0, 0, 0).toISOString();
}
```

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -24,13 +25,45 @@ interface GoogleCalError extends Error {
...
+ * Enable or disable the expanded cache
+ * TODO: Make this configurable
+ * */
+const ENABLE_EXPANDED_CACHE = true;
+
+/**
+ * By expanding the cache to whole months, we can save round trips to the third party APIs.
+ * In this case we already have the data in the database, so we can just return it.
+ */
+function handleMinMax(min: string, max: string) {
```
alannnc, line 53:
```suggestion
function handleStartAndEndOfMonth(min: string, max: string) {
```

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -96,345 +138,340 @@ export default class GoogleCalendarService implements Calendar {
...
-        payload["conferenceData"] = calEventRaw.conferenceData;
-      }
-      const calendar = google.calendar({
-        version: "v3",
+    if (calEventRaw.conferenceData && calEventRaw.location === MeetLocationType) {
+      payload["conferenceData"] = calEventRaw.conferenceData;
+    }
+    const calendar = await this.authedCalendar();
+    // Find in calEventRaw.destinationCalendar the one with the same credentialId
+
```
alannnc, line 183:
```suggestion
    
    // Find in calEventRaw.destinationCalendar the one with the same credentialId
```

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -96,345 +138,340 @@ export default class GoogleCalendarService implements Calendar {
...
-          eventId: uid,
-          sendNotifications: false,
-          sendUpdates: "none",
+    if (cached) return cached.value as unknown as calendar_v3.Schema$FreeBusyResponse;
+
+    const apires = await calendar.freebusy.query({
+      requestBody: { timeMin, timeMax, items },
+    });
+
+    await prisma.calendarCache.upsert({
```
alannnc, line 389: Can we omit this await? To make it respond faster.

alannnc (dismissed):
Code looks good! Left some NIT changes. Still since it's moving important calendar service methods we still need to test it.

apps/web/vercel.json
zomars: I don't think we can put comments on naked json files

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -129,306 +170,302 @@ export default class GoogleCalendarService implements Calendar {
...
+    try {
+      const event = await calendar.events.insert({
+        calendarId: selectedCalendar,
+        requestBody: payload,
+        conferenceDataVersion: 1,
+        sendUpdates: "none",
+      });
+
+      if (event && event.data.id && event.data.hangoutLink) {
+        await calendar.events.patch({
```
zomars, line 214: There was a missing await here.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -129,306 +170,302 @@ export default class GoogleCalendarService implements Calendar {
...
-          if (err) {
-            console.error("There was an error contacting google calendar service: ", err);
-            return reject(err);
-          }
-
-          if (evt && evt.data.id && evt.data.hangoutLink && event.location === MeetLocationType) {
-            calendar.events.patch({
-              // Update the same event but this time we know the hangout link
-              calendarId: selectedCalendar,
-              auth: myGoogleAuth,
```
zomars, line 271: calendar is already authed

apps/web/vercel.json
zomars: image

packages/app-store/googlecalendar/lib/CalendarService.ts
zomars: Will keep timeMin/Max to match what gCal is expecting.

zomars (commented):
Self review done

packages/app-store/googlecalendar/lib/CalendarService.test.ts
```diff
@@ -0,0 +1,99 @@
+import { afterEach, expect, test, vi } from "vitest";
+
+import prismaMock from "../../../../tests/libs/__mocks__/prisma";
```
emrysal, line 3: lil lint fix needed

emrysal (approved):
It looks good to me :)

PR #11187, opened 2023-09-06
feat: Add user's avatar to POST and PATCH requests in users endpoints

HamzaFouad:
## What does this PR do?

- Adds avatar to `POST` and `PATCH` requests in users endpoint
- Update users docs

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


apps/api/lib/validations/user.ts
```diff
@@ -104,6 +104,7 @@ const schemaUserEditParams = z.object({
     .optional()
     .nullable(),
   locale: z.nativeEnum(locales).optional().nullable(),
+  avatar: z.string().optional(),
```
keithwillcode, line 107: We should add validation that this string is actually a base64 string.

keithwillcode (dismissed):

apps/api/lib/validations/user.ts
HamzaFouad: @keithwillcode, Could you please check the new changes?

I separated the `isValidBase64` method in utils files, so it could be used for both images and texts, I don't know if I am missing something. hopefully not :)

apps/api/pages/api/users/[userId]/_patch.ts
```diff
@@ -62,6 +62,8 @@ import { schemaUserEditBodyParams, schemaUserReadPublic } from "~/lib/validation
  *               locale:
  *                 description: The user's locale. Acceptable values are one of [EN, FR, IT, RU, ES, DE, PT, RO, NL, PT_BR, ES_419, KO, JA, PL, AR, IW, ZH_CH, ZH_TW, CS, SR, SV, VI]
  *                 type: string
+ *                 description: The user's avatar, in base64 format
+ *                 type: string
```
Comment (alishaz-polymath) [inline apps/api/pages/api/users/[userId]/_patch.ts:66]:
```suggestion
 *               avatar:
 *                 description: The user's avatar, in base64 format
 *                 type: string
```

apps/api/lib/utils/isValidBase64.ts
```diff
@@ -0,0 +1,17 @@
...
+      if (parts.length === 2) {
+        // metadata and base64Data
+        base64Data = parts[1]; // Extract the base64 data from the second part
+      }
+    }
+    const buffer = Buffer.from(base64Data, "base64").toString("base64");
+    return buffer === base64Data;
+  } catch (error) {
+    return false;
+  }
```
ThyMinimalDev, line 16: couldn't we replace this whole thing with a regexp ? 

Also I think the naming of the function is unclear, we are here checking if the data inside of the data-uri string is encoded in base64.

HamzaFouad: Sure we can. but why do yo prefer using regex over a native lib?
--
I tried to name the extension method to be as generic as possible to accommodate all base64-encoding-string checks, not just for images.

HamzaFouad:
@alishaz-polymath  Could you please take the initiative and help this PR to move forward? It has been stale for very long time with no progress in reviews.

alishaz-polymath:
> @alishaz-polymath Could you please take the initiative and help this PR to move forward? It has been stale for very long time with no progress in reviews.

I'll review it this week. Thank you @HamzaFouad 🙏

apps/api/lib/utils/isValidBase64.ts
```diff
@@ -0,0 +1,17 @@
+export function isValidBase64(input: string): boolean {
+  try {
```
alishaz-polymath, line 2:
```suggestion
  const base64Regex = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/;

    try {
```

Can potentially use a regex here for strict checking for base64
This pattern assumes a character set of A-Z, a-z, 0-9, +, / and possibly ending with = or ==.

apps/api/lib/utils/isValidBase64.ts
```diff
@@ -0,0 +1,17 @@
...
+    let base64Data = input;
+    // Check if the input starts with a data URI scheme (data:content/type;base64,...)
+    if (input.startsWith("data:")) {
+      const parts = input.split(",");
+      if (parts.length === 2) {
+        // metadata and base64Data
+        base64Data = parts[1]; // Extract the base64 data from the second part
+      }
```
alishaz-polymath, line 10:
```suggestion
      if (parts.length !== 2) return false;
      
      base64Data = parts[1]; // Extract the base64 data from the second part
```

We can return false when parts length !== 2 to not go about further computations as it would be false already.

apps/api/lib/utils/isValidBase64.ts
```diff
@@ -0,0 +1,17 @@
...
+    }
+    const buffer = Buffer.from(base64Data, "base64").toString("base64");
```
alishaz-polymath, line 12:
```suggestion

    if (!base64Regex.test(base64Data)) return false;
    
    Buffer.from(base64Data, "base64");
    return true;
```

We can use regex to test and return false if it doesn't adhere to the regex pattern. The next line would fail if not base64, at which point it should return false thanks to catch.

apps/api/lib/utils/isValidBase64.ts
```diff
@@ -0,0 +1,17 @@
...
+    return buffer === base64Data;
```
alishaz-polymath, line 13:
```suggestion
    
```
Don't need to recheck it now that all checks are already done, this is redundant.

alishaz-polymath (changes requested):
Requested a few changes to improve strictness and efficiency.

Please **do test locally** to confirm they work and adjust accordingly. The rest looks good to me 🎉 , but I'll wait on the requested changes. 🙏

apps/api/lib/utils/isValidBase64.ts
HamzaFouad: I think we no longer need to use this line too anymore, right?
``` Buffer.from(base64Data, "base64"); ```

as the condition before it should be enough to cover base64 validity

File: apps/api/lib/utils/isValidBase64.ts (hunk 5 above)
Comment (HamzaFouad) [reply apps/api/lib/utils/isValidBase64.ts:10]: I made this condition to not just cover base64-images, but also to cover any base64 string checks you might have in the app, so the string may be a single part or multiple part based on the usecase.

if we are only going to make this method exclusive to base64-images then there is no need for the above check
and we can tweak the regex you implemented and implement the method as follows:

```export function isValidBase64Image(input: string): boolean {
  const dataURISchemeRegex = /^data:[A-Za-z0-9;:,.=()+\/-]+;base64,/;
  const base64Regex = new RegExp(
    `^(?:${dataURISchemeRegex.source})?(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$`
  );

  return base64Regex.test(input);
}
```
we will also rename the method to be image exclusive `isValidBase64Image`

apps/api/lib/utils/isValidBase64.ts
alishaz-polymath: Actually we still do.

- The regex test checks if the string looks like valid base64
- The buffer conversion checks if the string can be successfully decoded as base64

alishaz-polymath: @ThyMinimalDev 
What do you think?
The regex **is** fairly strict so there shouldn't be many strings that pass the regex and fail Base64, but what is the safer approach here?

ThyMinimalDev: I Think using a  regexp is the best approach here, I don't think there is a bulletproof way to make sure that the base64 string is actually a valid image, but as long as it's a valid base64 it is fine.

alishaz-polymath (approved):
LGTM 🚀 🚀

ThyMinimalDev (approved):

apps/api/lib/validations/user.ts
```diff
@@ -4,6 +4,7 @@ import { checkUsername } from "@calcom/lib/server/checkUsername";
 import { _UserModel as User } from "@calcom/prisma/zod";
 import { iso8601 } from "@calcom/prisma/zod-utils";
 
+import { isValidBase64Image } from "~/lib/utils/isValidBase64";
```
ThyMinimalDev, line 7: missed this during review, this file is now named `sValidBase64Image.ts` not `isValidBase64.ts`

PR #10791, opened 2023-08-16
fix: update profile input metadata

alannnc:
## What does this PR do?

- Better parsing when handling input.metadata for updateProfile trpc.

- [X] Bug fix (non-breaking change which fixes an issue)

- I haven't added tests that prove my fix is effective or that my feature works

PR #11188, opened 2023-09-06
fix: updateProfile metadata overwrite

zomars:
## What does this PR do?

followup to #10791

Fixes a condition where user metadata might be replaced when updating booking layout or session timeout 

## Requirement/Documentation

## Type of change

## How should this be tested?

On user settings, update booking layout or session timeout. The rest of the metadata should remain untouched.

## Mandatory Tasks


alannnc (approved):
Looks good!

PR #11190, opened 2023-09-07
chore: Enable prettier on vitest files without causing test failures

hariombalhara:
## What does this PR do?
Earlier prettier was disabled for test files as it was causing test failures because of mock imports being reordered. It fixes that.

- Ensures that prettier formatting doesn't fail tests because of `__mock__` imports reordering to somewhere at the bottom of imports.
- Formats all .test.ts files as earlier they were ignored by .prettierignore

Note: Except `.prettierignore` and `prettier-preset` all changes are automatically done by prettier.


PeerRich (approved):

zomars (commented):
I was wondering why we were ignoring test formatting . Nice  work @hariombalhara 🙏

PR #11224, opened 2023-09-07
perf: Improve getAggregatedAvailability by reducing reliance on Day.js

emrysal:
## What does this PR do?

On large event types with a large pool of users and availabilities, the performance of day.js `diff` and `isAfter` add up. This change saves ~200ms on a 6 person event type and probably ends up shaving seconds on big event types.

### DIFF
--- a/packages/core/getAggregatedAvailability.ts
+++ b/packages/core/getAggregatedAvailability.ts
@@ -22,25 +22,33 @@ export const getAggregatedAvailability = (
   return mergeOverlappingDateRanges(availability);
 };
 
+function isSameDay(date1: Date, date2: Date) {
+  return (
+    date1.getUTCFullYear() === date2.getUTCFullYear() &&
+    date1.getUTCMonth() === date2.getUTCMonth() &&
+    date1.getUTCDate() === date2.getUTCDate()
+  );
+}
+
 function mergeOverlappingDateRanges(dateRanges: DateRange[]) {
-  const sortedDateRanges = dateRanges.sort((a, b) => a.start.diff(b.start)); //is it already sorted before?
+  dateRanges.sort((a, b) => a.start.valueOf() - b.start.valueOf());
 
   const mergedDateRanges: DateRange[] = [];
 
-  let currentRange = sortedDateRanges[0];
+  let currentRange = dateRanges[0];
   if (!currentRange) {
     return [];
   }
 
-  for (let i = 1; i < sortedDateRanges.length; i++) {
-    const nextRange = sortedDateRanges[i];
+  for (let i = 1; i < dateRanges.length; i++) {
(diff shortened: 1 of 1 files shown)


packages/core/getAggregatedAvailability.ts
```diff
@@ -22,25 +22,33 @@ export const getAggregatedAvailability = (
...
   }
 
-  for (let i = 1; i < sortedDateRanges.length; i++) {
-    const nextRange = sortedDateRanges[i];
+  for (let i = 1; i < dateRanges.length; i++) {
+    const nextRange = dateRanges[i];
     if (
-      currentRange.start.utc().format("DD MM YY") === nextRange.start.utc().format("DD MM YY") &&
-      currentRange.end.isAfter(nextRange.start)
+      isSameDay(currentRange.start.toDate(), nextRange.start.toDate()) &&
```
emrysal, line 46: Native compare is faster than string formatting. Not a huge diff but it is faster so worth doing.

packages/core/getAggregatedAvailability.ts
```diff
@@ -22,25 +22,33 @@ export const getAggregatedAvailability = (
...
+      currentRange.end.valueOf() > nextRange.start.valueOf()
```
emrysal, line 47: Hugely faster, valueOf is very fast (effectively `+currentRange.end`)

packages/core/getAggregatedAvailability.ts
```diff
@@ -22,25 +22,33 @@ export const getAggregatedAvailability = (
...
+  return (
+    date1.getUTCFullYear() === date2.getUTCFullYear() &&
+    date1.getUTCMonth() === date2.getUTCMonth() &&
+    date1.getUTCDate() === date2.getUTCDate()
+  );
+}
+
 function mergeOverlappingDateRanges(dateRanges: DateRange[]) {
-  const sortedDateRanges = dateRanges.sort((a, b) => a.start.diff(b.start)); //is it already sorted before?
+  dateRanges.sort((a, b) => a.start.valueOf() - b.start.valueOf());
```
emrysal, line 34: Biggest win by far, before the performance showed >100ms for my small-ish event type, after it showed 0.1ms.

emrysal, line 34: Biggest win by far, before the performance showed >100ms for my small-ish event type, after it showed 0.1ms.

emrysal: TL;DR; don't use `.diff`

emrysal: Also sort is a mutable, so the returned assignment wasn't needed.

keithwillcode (approved):
Nice work!

PR #11227, opened 2023-09-08
fix: Handle removing user from org

### LINKED ISSUE #11204: [CAL-2415] Can't re-invite someone to an org
### Issue Summary

Deleting a pending invitation doesn't delete the user from the Users table causing an error when trying to invite again

### Steps to Reproduce

1. Create an org
2. Invite someone to the org
3. Delete the invitation
4. Try to invite the same person again (same email)

### Actual Results

Getting an error that that person already exists

### Expected Results

Should be able to re-invite the deleted invited person

<sub>[CAL-2415](https://linear.app/calcom/issue/CAL-2415/cant-re-invite-someone-to-a-sub-team)</sub>

sean-brydon:
Loom: https://www.loom.com/share/0d938a0f1ca04d8fbc3e2b8d3b629d8f

See issue for testing steps


packages/trpc/server/routers/viewer/teams/removeMember.handler.ts
```diff
@@ -54,6 +55,38 @@ export const removeMemberHandler = async ({ ctx, input }: RemoveMemberOptions) =
 
   if (input.isOrg) {
     // Deleting membership from all child teams
+    const foundUser = await ctx.prisma.user.findUnique({
+      where: { id: input.memberId },
+      select: {
+        email: true,
+      },
+    });
+
```
sean-brydon, line 64: Find the user by memberID so we can compare its email

packages/trpc/server/routers/viewer/teams/removeMember.handler.ts
```diff
@@ -54,6 +55,38 @@ export const removeMemberHandler = async ({ ctx, input }: RemoveMemberOptions) =
...
+    const orgInfo = await ctx.prisma.team.findUnique({
+      where: { id: input.teamId },
+      select: {
+        metadata: true,
+      },
+    });
+
+    if (!foundUser || !orgInfo) throw new TRPCError({ code: "NOT_FOUND" });
+
+    const parsedMetadata = teamMetadataSchema.parse(orgInfo.metadata);
```
sean-brydon, line 74: Get metadata

packages/trpc/server/routers/viewer/teams/removeMember.handler.ts
```diff
@@ -54,6 +55,38 @@ export const removeMemberHandler = async ({ ctx, input }: RemoveMemberOptions) =
...
+      parsedMetadata.orgAutoAcceptEmail
+    ) {
+      if (foundUser.email.endsWith(parsedMetadata.orgAutoAcceptEmail)) {
+        await ctx.prisma.user.delete({
+          where: { id: input.memberId },
+        });
+        // This should cascade delete all memberships and hosts etc
+        return;
+      }
+    }
```
sean-brydon, line 88: I put these into seperate IFs so its more read-able.

If the team is an organization & is verified & there is a autoAcceptDomain
AND
if the found users email ends with the autoAcceptDomain
Delete the user

leog (changes requested):
LGTM, as discussed we need to remove the user completely when password and/or username is not set

leog (approved):
Thanks for taking care of the last bit, LGTM

Udit-takkar (commented):
@sean-brydon awesome works fine for orgs but if i invite someone to team of an org then when i remove the user then the user is not deleted like in case of orgs

PR #11230, opened 2023-09-08
perf: Move query to get just the user email to the main get query

emrysal:
## What does this PR do?

Less queries = huge performance increase for large event types. (for an event type with 60 attendees this change reduces 60x async outbound calls))


sean-brydon (dismissed):
Good spot :P

keithwillcode (dismissed):
This will likely be a nice improvement for speed. Great stuff.

keithwillcode (approved):

PR #11231, opened 2023-09-08
fix: Fixed team image not showing on event-types page

pateldivyesh1323:
## What does this PR do?

Fixed team image not showing up on event-types page
- Before:
team-image-not-showing

- After:
team-image-showing

## Type of change

## How should this be tested?
Go on event-types page and check the image of teams.

## Checklist


joeauyeung (dismissed):
Thanks for your contribution @pateldivyesh1323. I think there's a way we can take this PR to the next level. In the `EventTypeListHeading` the profile that is being passed contains the proproty `image`.  I think we can use that property and pass it to the `Avatar` component.  This data comes from `packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts`.

The only bug I am seeing is that organization team avatar's are not being rendered now because they don't contain `/team` in the image URL.

pateldivyesh1323:
> Thanks for your contribution @pateldivyesh1323. I think there's a way we can take this PR to the next level. In the `EventTypeListHeading` the profile that is being passed contains the proproty `image`. I think we can use that property and pass it to the `Avatar` component. This data comes from `packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts`.
> 
> The only bug I am seeing is that organization team avatar's are not being rendered now because they don't contain `/team` in the image URL.

Updated as per suggested.

apps/web/pages/event-types/index.tsx
```diff
@@ -720,10 +720,7 @@ const EventTypeListHeading = ({
       <Avatar
         alt={profile?.name || ""}
         href={teamId ? `/settings/teams/${teamId}/profile` : "/settings/my-account/profile"}
-        imageSrc={
-          `${orgBranding?.fullDomain ?? WEBAPP_URL}${teamId ? "/team" : ""}/${profile.slug}/avatar.png` ||
-          undefined
-        }
+        imageSrc={profile?.image || undefined}
```
Udit-takkar, line 723: this isn't working in case of orgs

Udit-takkar: you can create org using http://localhost:3000/settings/organizations/new

pateldivyesh1323: It is asking to verify email. I dont know how to setup an org

Udit-takkar (approved):
LGTM. I fixed the organization avatar. @pateldivyesh1323 Thanks for looking into this

Screenshot 2023-09-09 at 1 55 34 AM

Issue #11234, opened 2023-09-08
[CAL-2428] better error message if user books seat from same slot again


CarinaWolli:
### Issue Summary

We should show a better message when a users tries to book a seat that they have already booked in a seated event.

Currently, we show this generic message:
<img width="394" alt="Screenshot 2023-09-08 at 11 18 01" src="https://github.com/calcom/cal.com/assets/30310907/ffc33d36-a9cf-4f09-a55f-8c90162cde1a">

We should show something like 'Could not book meeting. Already signed up for this booking.'

### Steps to Reproduce

1. Create an event with seats
2. Book a slot 
3. Book the same slot again with the same user 

<sub>[CAL-2428](https://linear.app/calcom/issue/CAL-2428/better-error-message-if-user-books-seat-from-same-slot-again)</sub>

TheNatkat:
hey i want to work on this and please tell me the message you want instead of this message

PR #11237, opened 2023-09-08
fix: Better error message if user books seat from same slot again

MeenuyD:
Better error message if the user books a seat from the same slot again. Which shows something like 'Could not book meeting. Already signed up for this booking.'
#11234


joeauyeung:
@MeenuyD thanks for your contribution, but this PR doesn't tackle the underlying issue. That message is being displayed for all booking errors. We should determine the error and communicate it appropriately to the booking user. I'm going to close this PR for now but feel free to reopen this PR with the new approach.

PR #11241, opened 2023-09-08
fix: Better error message if user books seat from same slot again 

### LINKED ISSUE #11234: [CAL-2428] better error message if user books seat from same slot again

m-Chetan:
Better error message if user books seat from same slot again in all languages.
It Displays:
"Could not book the meeting. You're already signed up for this booking"

Fixes #11234


joeauyeung:
@m-Chetan I don't think this PR tackles the underlying problem. The message that is displayed is a catch all for all booking errors. We should determine what error has occurred and communicate that to the attendee. I'm going to close this PR for now but feel free to reopen with the new approach.

m-Chetan:
@joeauyeung Thanks for letting me know I will try to come up with another approach

PR #11243, opened 2023-09-08
perf: pre-load booking data for all users

### LINKED ISSUE #11134: Loading calendar for large round robin is slow

### Issue Summary

When having round robins consisting of lots of users, we simply run too many queries and the performance is heavily degraded. We've talked about having a first version to fix this that loads availability for 10 users in the round robin upon page load. 

If the user who is doing the booking cannot find a slot that works for them, they click a "Show more" button and we load availability for 10 more users in the round robin. The user has the option to continue loading more availability until all users' availability is shown.

### Steps to Reproduce

1. Create a round robin event with lots of users (e.g. 50)
2. Load the calendar for that round robin

### Actual Results

- It's taking around 15 seconds to load the availability for 50 people

### Expected Results

- We should be able to load a round robin calendar for 10 users at a time in under a couple of seconds.

keithwillcode:
## What does this PR do?

Instead of loading booking data per user when determining availability, we now preload it for all users. The query currently loading booking data in the loop for all users is actually the slowest query in our system according to database load data.

## Type of change

## How should this be tested?

- What are the minimal test data to have? **A round robin event with 20 users minimum. Ideally more.**
- What is expected (happy path) to have (input and output)? **Booking calendar availability calculations don't change at all. They are just incredibly faster for large round robin events.**

## Mandatory Tasks


apps/web/test/utils/bookingScenario.ts
```diff
@@ -194,10 +194,17 @@ async function addBookings(bookings: InputBooking[], eventTypes: InputEventType[
             // eslint-disable-next-line @typescript-eslint/ban-ts-comment
             // @ts-ignore
             const statusIn = where.OR[0].status?.in || [];
+
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            const userIdIn = where.OR[0].userId?.in || [];
```
keithwillcode, line 200: This adds support for finding bookings by a list of users instead of just 1 particular user.

packages/core/getBusyTimes.ts
```diff
@@ -76,49 +88,56 @@ export async function getBusyTimes(params: {
...
-            some: {
-              email: userEmail,
+
+  // INFO: Refactored to allow this method to take in a list of current bookings for the user.
+  // Will keep support for retrieving a user's bookings if the caller does not already supply them.
+  // This function is called from multiple places but we aren't refactoring all of them at this moment
+  // to avoid potential side effects.
+  const bookings = params.currentBookings
+    ? params.currentBookings
+    : await prisma.booking.findMany({
```
keithwillcode, line 98: This query is one of the slowest we have in the system and most frequently called. This refactor pulls it out of being in a loop for round robin event types but still has support for other callers.

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -291,6 +292,73 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
     usersWithCredentials = eventType.hosts.map(({ isFixed, user }) => ({ isFixed, ...user }));
   }
 
+  const durationToUse = input.duration || 0;
+
+  const startTimeDate =
```
keithwillcode, line 297: This is the same logic used in getBusyTimes, just extracted to here.

emrysal (approved):
⚡ ⚡ ⚡ ⚡ ⚡ LGTM ⚡ ⚡ ⚡ ⚡ ⚡

PR #11250, opened 2023-09-09

### LINKED ISSUE #11106: Add placeholder design to the dates to override panel.

meet13552:
## What does this PR do?

Fixes #11106 

By default, for md size or greater DialogContent had w-full. I changed it to md:w-auto which solved the problem of extra blank space besides the date picker. But by doing this, the date picker got shrinked in md and lg size as there was not enough space available. So i made it flex-col for md and lg size in form which made conditional rendering below date picker instead of right side. And for xl and greater flex-row which made conditional rendering right side as expected. With this solution, it is working fine in all screen sizes and the issue is fixed.

## Type of change

### DIFF
--- a/packages/features/schedules/components/DateOverrideInputDialog.tsx
+++ b/packages/features/schedules/components/DateOverrideInputDialog.tsx
@@ -139,7 +139,7 @@ const DateOverrideForm = ({
         );
         onClose();
       }}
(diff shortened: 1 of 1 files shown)


PeerRich:
awesome, works!

PeerRich (approved):

PeerRich:
thank you a lot!

PR #11251, opened 2023-09-09
fix: placeholder when dates not selected #11106

### LINKED ISSUE #11106: Add placeholder design to the dates to override panel.

shreyas0924:
## What does this PR do?

Added placeholder for the date select dialog when dates are not selected instead of blank space.

Fixes #11106 

## Screenshot

image

## Type of change


PeerRich:
we're going with  a different solutiuon, thank you!

shreyas0924:
Thanks

PR #11266, opened 2023-09-10
fix: add hardcoded strings to locales so they can be translated

### LINKED ISSUE #11265: Hardcoded strings in english that will not get translated
### Issue Summary

While working on some other issue, I found few strings that were hardcoded in place and won't get translated.

- "Invite via email" (Shows up when we try to invite user to an organization and when we try to invite admins to an organization)
- "Upload a .csv file" (Shows up when we try bulk uploading user to invite to an organization)

### Steps to Reproduce

These labels are rendered when trying to invite users to an organization.

### Actual Results

- The mentioned labels won't get translated.

### Expected Results

- The mentioned labels would get translated when translations are added in the future.

### Evidence

- I did not test it but it is quite evident from the code that they won't get translated in other languages.

ujwalkumar1995:
## What does this PR do?

Add hardcoded strings to locales folder so they can be translated.

Fixes #11265 

## Type of change

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/public/static/locales/en/common.json
+++ b/apps/web/public/static/locales/en/common.json
@@ -592,6 +592,8 @@
   "invite_new_member": "Invite a new team member",
   "invite_new_member_description": "Note: This will <1>cost an extra seat ($15/m)</1> on your subscription.",
   "invite_new_team_member": "Invite someone to your team.",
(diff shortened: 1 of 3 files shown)


PeerRich (approved):
thank you!

PR #11267, opened 2023-09-10
feat: autolink org on provider signup

leog:
## What does this PR do?

When someone uses an external provider to sign-up, such as Google, and the `ORGANIZATIONS_AUTOLINK` env variable is used, then the signed up user will be autolinked to the organization automatically as a member. 

For example, if an organization "Acme" with domain `acme.com` is created, and the organization is verified, any user signing up with Google using a `acme.com` domain such as `john@acme.com`, they will automatically be linked to organization "Acme" as a member.

 Loom Video: Coming soon

## Type of change

## How should this be tested?

> **Note**
> You don't need to have organizations enabled for these steps, only env var `ORGANIZATIONS_AUTOLINK=1` 

1. Make sure you have organizations feature turned on by logging in as Admin and going to Features section in Settings
2. Configure Google integration using the steps [here](https://github.com/calcom/cal.com#obtaining-the-google-api-credentials)
3. Change env var `NEXT_PUBLIC_WEBSITE_URL=http://localhost:3001`
4. Run `yarn dev:website`
5. Create an org ([http://localhost:3000/settings/organizations/new](http://localhost:3000/settings/organizations/new)) with an email that matches the email address used in Google signup, for example, create an org using email john.doe@cal.com if you can signup using a cal.com Google account, or for testing purposes only, create an org using email john.doe@gmail.com if you have a gmail account to then signup
6. Go to [http://localhost:3001/signup](http://localhost:3001/signup) and input a username (which will not be used after all) and click the Google icon to sign up using a Google Account


packages/features/auth/lib/next-auth-options.ts
```diff
@@ -31,6 +31,8 @@ const { client_id: GOOGLE_CLIENT_ID, client_secret: GOOGLE_CLIENT_SECRET } =
   JSON.parse(GOOGLE_API_CREDENTIALS)?.web || {};
 const GOOGLE_LOGIN_ENABLED = process.env.GOOGLE_LOGIN_ENABLED === "true";
 const IS_GOOGLE_LOGIN_ENABLED = !!(GOOGLE_CLIENT_ID && GOOGLE_CLIENT_SECRET && GOOGLE_LOGIN_ENABLED);
+const ORGANIZATIONS_AUTOLINK =
+  process.env.ORGANIZATIONS_AUTOLINK === "1" || process.env.ORGANIZATIONS_AUTOLINK === "true";
```
leog, line 35: Making sure the new env var flag is turned on to allow autolinking for orgs

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -54,6 +56,35 @@ export const checkIfUserBelongsToActiveTeam = <T extends UserTeams>(user: T) =>
...
+            equals: apexDomain,
+          },
+        },
+      ],
+    },
+    select: {
+      id: true,
+    },
+  });
+  return [username, existingOrg?.id];
```
leog, line 85: Making sure we return the forced username from email address handle and orgId if exists

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -735,16 +766,25 @@ export const AUTH_OPTIONS: AuthOptions = {
...
 
+        // Associate with organization if enabled by flag
+        const [orgUsername, orgId] = await checkIfUserShouldBelongToOrg(user.email);
+
         const newUser = await prisma.user.create({
           data: {
             // Slugify the incoming name and append a few random characters to
             // prevent conflicts for users with the same name.
-            username: usernameSlug(user.name),
+            username: orgId ? orgUsername : usernameSlug(user.name),
```
leog, line 776: We use forced org username calculated in function if there is an orgId which in turn means autolink is turned on

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -735,16 +766,25 @@ export const AUTH_OPTIONS: AuthOptions = {
...
             name: user.name,
             email: user.email,
             identityProvider: idP,
             identityProviderId: account.providerAccountId,
+            ...(orgId && {
+              organization: { connect: { id: orgId } },
+              teams: {
+                create: { role: MembershipRole.MEMBER, accepted: true, team: { connect: { id: orgId } } },
+              },
+            }),
```
leog, line 787: We need to link the org by id to the newly created user as well as creating a membership for the user, only if orgId exists which in turn means autolink is turned on

packages/features/ee/organizations/components/OrganizationAvatar.tsx
```diff
@@ -15,7 +15,7 @@ const OrganizationAvatar = ({ size, imageSrc, alt, organizationSlug, ...rest }:
       indicator={
         organizationSlug ? (
           <div
-            className={classNames("absolute bottom-0 right-0 z-10", size === "lg" ? "h-3 w-3" : "h-10 w-10")}>
+            className={classNames("absolute bottom-0 right-0 z-10", size === "lg" ? "h-6 w-6" : "h-10 w-10")}>
```
leog, line 18: Found the sub-avatar for large regular avatar excessively small, increasing its size

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -148,7 +148,7 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
         organization: {
           create: {
             name,
-            ...(IS_TEAM_BILLING_ENABLED ? { slug } : {}),
+            ...(!IS_TEAM_BILLING_ENABLED ? { slug } : {}),
```
leog, line 151: Unintentional change that broke org branding locally (when `IS_TEAM_BILLING_ENABLED` is turned off), noted in the original PR after the fact [here](https://github.com/calcom/cal.com/pull/10967/files#r1320775851)

turbo.json
```diff
@@ -258,6 +258,7 @@
     "NEXTAUTH_URL",
     "NODE_ENV",
     "ORGANIZATIONS_ENABLED",
+    "ORGANIZATIONS_AUTOLINK",
```
leog, line 261: New env var to control whether org autolink for external providers should happen or not

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -54,6 +56,35 @@ export const checkIfUserBelongsToActiveTeam = <T extends UserTeams>(user: T) =>
     return metadata.success && metadata.data?.subscriptionId;
   });
 
+const checkIfUserShouldBelongToOrg = async (
+  email: string
```
emrysal, line 60: Email can't be null, why is there an if statement to check if email is null below?

packages/features/auth/lib/next-auth-options.ts
sean-brydon, line 85: NIT personally think this should return an object. 

If using this i have no clue what array 0,1 is

sean-brydon (commented):
Tested - seems to work well :)

packages/features/auth/lib/next-auth-options.ts
leog: Resolved, thanks

packages/trpc/server/routers/viewer/teams/inviteMember/utils.ts
```diff
@@ -148,6 +148,7 @@ export async function createNewUserConnectToOrgIfExists({
   const createdUser = await prisma.user.create({
     data: {
       email: usernameOrEmail,
+      verified: true,
```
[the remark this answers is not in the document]
emrysal: Makes sense, but I do think we should do the verification step on acceptance. - not on invite, but not a blocker

emrysal (approved):

PR #11281, opened 2023-09-11
fix: Selected calendar unhandled promise when removing calendars

joeauyeung:
## What does this PR do?

This PR fixes the unhandled promise when deleting selected calendars when removing a calendar integration.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Connect a calendar
- Select calendars to check for conflicts
- Delete that integration
- The error shouldn't appear in the logs and the selected calendars should be removed from the db

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
@@ -309,26 +309,33 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
 
   // If it's a calendar remove it from the SelectedCalendars
   if (credential.app?.categories.includes(AppCategories.calendar)) {
-    try {
-      const calendar = await getCalendar(credential);
+    const calendar = await getCalendar(credential);
 
-      const calendars = await calendar?.listCalendars();
+    const calendars = await calendar?.listCalendars();
 
-      if (calendars && calendars.length > 0) {
-        calendars.map(async (cal) => {
-          prisma.selectedCalendar.deleteMany({
+    if (calendars && calendars.length > 0) {
+      calendars.map(async (cal) => {
+        prisma.selectedCalendar
+          .delete({
             where: {
-              userId: user.id,
-              externalId: cal.externalId,
+              userId_integration_externalId: {
+                userId: user.id,
+                externalId: cal.externalId,
+                integration: cal.integration as string,
+              },
             },
+          })
+          .catch((error) => {
+            if (error instanceof Prisma.PrismaClientKnownRequestError && error.code === "P2025") {
+              console.log(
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -206,7 +206,6 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
                   bookingFields: true,
                   seatsPerTimeSlot: true,
                   seatsShowAttendees: true,
-                  seatsShowAvailabilityCount: true,
```
alannnc, line 209: This was deleted because we didn't use it?

joeauyeung: Searched through the file and this wasn't being used anywhere else.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -317,18 +315,18 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
...
-          await prisma.selectedCalendar.delete({
+          await prisma.selectedCalendar.deleteMany({
             where: {
-              userId_integration_externalId: {
-                userId: user.id,
-                externalId: cal.externalId,
-                integration: cal.integration as string,
-              },
+              userId: user.id,
+              externalId: cal.externalId,
```
alannnc, line 321: I think keeping .delete() but don't await prisma but with a .catch() on every promise should do the tric.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
alannnc: Yeah linting can do that

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -317,18 +315,18 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
 
       if (calendars && calendars.length > 0) {
         calendars.map(async (cal) => {
-          await prisma.selectedCalendar.delete({
+          await prisma.selectedCalendar.deleteMany({
```
alannnc, line 318: Something like this: 
```suggestion
         prisma.selectedCalendar.delete({
              userId_integration_externalId: {
                userId: user.id,
                externalId: cal.externalId,
                integration: cal.integration as string,
              },
            },
        }).catch(async catching in another thread);
```

alannnc (approved):

PR #11284, opened 2023-09-11
fix: Return all responses in booking api

kcrimi:
## What does this PR do?

 Because of validation, only the stock fields were returned in the `responses` in the booking API. This change allows all custom field responses to be returned as well.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Make a booking with custom questions in the registration form. 
- Call the `/bookings` endpoint
- Validate that all responses to custom questions come in in addition to the stock set (eg name, notes, email, etc)

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/api/lib/validations/booking.ts
+++ b/apps/api/lib/validations/booking.ts
@@ -58,6 +58,7 @@ export const schemaBookingReadPublic = Booking.extend({
       })
     )
     .optional(),
+  responses: z.record(z.any()),
 }).pick({
   id: true,
   userId: true,


apps/api/lib/validations/booking.ts
```diff
@@ -58,6 +58,7 @@ export const schemaBookingReadPublic = Booking.extend({
       })
     )
     .optional(),
+  responses: z.record(z.any()),
```
kcrimi, line 61: It seemed like this could have been added a level up at the  `Booking` which this is extending but it seemed safer to add it here since this is only used by the `/bookings` API

zomars (approved):
Awesome find @kcrimi 

PR #11288, opened 2023-09-11
fix: Dynamic events prevent fetching all users that share a username [CAL-2409]

### LINKED ISSUE #11186: [CAL-2409] orgs: duplicate invites for orgs and non-orgs with dynamic bookings
both [cal.com/peer](http://cal.com/peer) and [meet.cal.com/peer](http://meet.cal.com/peer) (see emails) are getting invites with dynamic links

same goes for [cal.com/bailey](http://cal.com/bailey) and [meet.cal.com/bailey](http://meet.cal.com/bailey)

i've only realised this since i renamed people today

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2409](https://linear.app/calcom/issue/CAL-2409/orgs-duplicate-invites-for-orgs-and-non-orgs-with-dynamic-bookings)</sub>

joeauyeung:
## What does this PR do?

Currently there can be multiple usernames if they belong to different organizations. Right now in `handleNewBooking` all users of the same username were being fetched in dynamic bookings, without any context to the org or not. The PR only fetches users that belong to the org or just normal users.

For example we have two users with the "admin" username "admin@example.com" and "org1member1@org1.com". We can see when we create a dynamic event with "admin+free" before it was grabbing both users but now it is querying for one. This works within an orgs context as well

## Before
image

## After
image

Fixes #11186 

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create two users with the same username. The second user should belong to an org
- Create a dynamic booking that includes the test username
- The created booking should only contain the user outside of the booking
- Go through this flow again but within an org context

## Mandatory Tasks

## Checklist


packages/features/ee/organizations/lib/orgDomains.ts
```diff
@@ -64,3 +64,8 @@ export function getSlugOrRequestedSlug(slug: string) {
     ],
   } satisfies Prisma.TeamWhereInput;
 }
+
+export function userOrgQuery(hostname: string, fallback?: string | string[]) {
+  const { currentOrgDomain, isValidOrgDomain } = orgDomainConfig(hostname, fallback);
+  return isValidOrgDomain && currentOrgDomain ? getSlugOrRequestedSlug(currentOrgDomain) : null;
+}
```
joeauyeung, line 71: This is being used in more than one place so let's abstract it.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -732,6 +733,7 @@ async function handler(
         const users = await prisma.user.findMany({
           where: {
             username: { in: dynamicUserList },
+            organization: userOrgQuery(req.headers.host.replace(/^https?:\/\//, "")),
```
joeauyeung, line 736: When loading users we should only grab users that belong to the org. This is similar to how we fetch users on the dynamic booking page

keithwillcode: This doesn't feel like the correct place to do this code. i.e. Parsing request headers inside of place where we create a new booking to determine the org.

Do we not have this on session or somewhere? If not, feels like this should be pulled out to some other class and called here.

keithwillcode (dismissed):

packages/features/bookings/lib/handleNewBooking.ts
alannnc: I think this can only come from the request headers sadly.

alannnc (dismissed):
Looks good to me, sadly that's the only way to tell if fetching users from an org.

packages/features/ee/organizations/lib/orgDomains.ts
```diff
@@ -64,3 +64,8 @@ export function getSlugOrRequestedSlug(slug: string) {
     ],
   } satisfies Prisma.TeamWhereInput;
 }
+
+export function userOrgQuery(hostname: string, fallback?: string | string[]) {
+  const { currentOrgDomain, isValidOrgDomain } = orgDomainConfig(hostname, fallback);
```
zomars, line 69: This is already being called in the parent: https://github.com/calcom/cal.com/pull/11288/files#diff-0a49c6d137ab5a519fee377425ff051a13bcf55ff728bca45daaa73a4c5f0458R144-R147

Can we either just pass the computed data or find another abstraction?

packages/features/bookings/lib/handleNewBooking.ts
emrysal: We don't have a session here (not guaranteed anyways) and we don't have an event type id. the only way really is parsed from the calling url. think that's the nicest way too

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -745,6 +746,7 @@ async function handler(
...
```
hariombalhara, line 749:
```suggestion
            organization: userOrgQuery(req.headers.host),
```

I don't think host has protocol in it. We seem to be using it as is without replace elsewhere as well.

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara: I am not sure if I am right but I am reluctant towards sending either the `req`(more inclined towards it) or `req.headers` to userOrgQuery so that it can be sure that it is receiving **host** only and not some other string. We will get better type safety that way.

hariombalhara: Also, I feel that prisma client extensions would really solve this problem once and for all. We would be able to enforce an explicit value for `organization` there in user queries. Problem arises mostly when we forget to use `organization` in where.

Though not sure if I should focus on that considering, we are looking for prisma alternatives.

hariombalhara (approved):
LGTM!!

hariombalhara (changes requested):
Typechecks are failing

packages/features/bookings/lib/handleNewBooking.ts
joeauyeung: I noticed that `req.headers.host` is being passed to `orgDomainConfig` which I think is similar to `userOrgQuery`.

packages/features/ee/organizations/lib/orgDomains.ts
joeauyeung: I noticed that `orgDomainConfig` and `getSlugOrRequestedSlug` were called together in multiple places throughout the app, so I created this function. Looking through the codebase I noticed this is happening more than just the file that is linked.

hariombalhara (approved):
Types issue resolved now. LGTM !!

alannnc:
@joeauyeung Should we merge? It was already approved by me but not sure if we waiting on something?

PR #11289, opened 2023-09-12
fix: Prefill not working if firstAndLastName variant enabled in name field and `name` doesn't have any space

hariombalhara:
## What does this PR do?

Fixes the following error and avoid prefilling failure if **name** prefill param is provided without lastname. It happens only when firstAndLastName variant is enabled.
image

## Type of change

## How should this be tested?
- Enable 'Split Full Name' toggle in booking questions  of an event-type
- Prefill name param during the booking as ?name=John&email=john@example.com. Notice that both email and name aren't prefilled. Because there is an error during zod validation, entire prefill is rejected and nothing is prefilled.

## Mandatory Tasks

### DIFF
--- a/packages/prisma/zod-utils.ts
+++ b/packages/prisma/zod-utils.ts
@@ -123,9 +123,7 @@ export const bookingResponses = z
       nonEmptyString(),
       z.object({
         firstName: nonEmptyString(),
-        lastName: nonEmptyString()
-          .refine((value: string) => value.trim().length > 0)
-          .optional(),
+        lastName: z.string().optional(),
       }),
     ]),
     guests: z.array(z.string()).optional(),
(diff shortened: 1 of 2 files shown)


packages/prisma/zod-utils.ts
```diff
@@ -123,9 +123,7 @@ export const bookingResponses = z
       nonEmptyString(),
       z.object({
         firstName: nonEmptyString(),
-        lastName: nonEmptyString()
-          .refine((value: string) => value.trim().length > 0)
-          .optional(),
+        lastName: z.string().optional(),
```
hariombalhara, line 126: `lastName` was not configured properly. It was marked optional as well as it was required to have 1 char at least.

sean-brydon (approved):
Great job @hariombalhara killing it 🔥

PR #11299, opened 2023-09-12
fix: multiple organizer address bug

Udit-takkar:
## What does this PR do?
Fixes https://linear.app/calcom/issue/CAL-2437/multiple-organizer-addresses-for-attendee-to-choose-from-doesnt-select
Added a unit test as well and modified existing e2e for multiple organizer to catch this bug.

### DIFF
--- a/apps/web/playwright/event-types.e2e.ts
+++ b/apps/web/playwright/event-types.e2e.ts
@@ -142,17 +142,25 @@ test.describe("Event Types tests", () => {
         .first()
         .getAttribute("href");
 
+      /**
+       * Verify first organizer address
+       */
       await page.goto(previewLink ?? "");
-
       await selectFirstAvailableTimeSlotNextMonth(page);
-
-      for (const location of locationData) {
-        await page.locator(`span:has-text("${location}")`).click();
-      }
-
+      await page.locator(`span:has-text("${locationData[0]}")`).click();
       await bookTimeSlot(page);
-
       await expect(page.locator("[data-testid=success-page]")).toBeVisible();
+      await expect(page.locator(`[data-testid="where"]`)).toHaveText(locationData[0]);
+
+      /**
+       * Verify second organizer address
+       */
+      await page.goto(previewLink ?? "");
+      await selectFirstAvailableTimeSlotNextMonth(page);
+      await page.locator(`span:has-text("${locationData[1]}")`).click();
+      await bookTimeSlot(page);
+      await expect(page.locator("[data-testid=success-page]")).toBeVisible();
+      await expect(page.locator(`[data-testid="where"]`)).toHaveText(locationData[1]);
     });
 
     test.describe("Different Locations Tests", () => {
--- a/packages/features/form-builder/Components.tsx
+++ b/packages/features/form-builder/Components.tsx
@@ -64,9 +64,10 @@ type Component =
         TProps extends SelectLikeComponentProps<{
           value: string;
           optionValue: string;
+          _type?: string;
         }> & {
           optionsInputs: NonNullable<z.infer<typeof fieldSchema>["optionsInputs"]>;
-          value: { value: string; optionValue: string };
+          value: { value: string; optionValue: string; _type?: string };
         } & {
           name?: string;
           required?: boolean;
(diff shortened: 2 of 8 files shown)


packages/features/form-builder/Components.tsx
```diff
@@ -392,7 +392,7 @@ export const Components: Record<FieldType, Component> = {
         if (!value) {
           setValue({
             value: options[0]?.value,
-            optionValue: "",
+            optionValue: options[0]?.label,
```
hariombalhara: `optionValue` is supposed to be added by the booker if applicable e.g. for Attendee Inputs. Also, it is a general purpose component that can be used by the booker to add more fields of this type.

Can we maybe tweak the options for locations so that there isn't a change in `radioInput` component and it receives the options[number].value as Label which would be practically unique.

Udit-takkar: wait this is outdated

Udit-takkar: can you pull again?

packages/features/form-builder/FormBuilderField.tsx
```diff
@@ -311,6 +311,10 @@ export const ComponentForField = ({
     if (!field.optionsInputs) {
       throw new Error("Field optionsInputs is not defined");
     }
+    const options = field.options.map((field) => {
+      return { ...field, value: field.value === "inPerson" ? field.label : field.value };
```
hariombalhara, line 315: Could you check if the organizer address value is correctly shown after creating booking. Also, in email.

I think with just this change, handleNewBooking wouldn't know which location was selected as it looks up by type

Udit-takkar: yeah i checked. i got the correct location

Screenshot 2023-09-12 at 5 49 05 PM

Udit-takkar: Screenshot 2023-09-12 at 5 49 59 PM

Udit-takkar: cc: @hariombalhara

hariombalhara: So, the problem is that `handleNewBooking` still doesn't have a clue which type of location was selected. It just knows that some text was given. That text can be an organizer address or organizer phone number or some other type of location. In case it can't detect which location type it is it just shows what was given to it.

So, even though it works, we don't logically know if it's an address, phone, or something else that was selected.

Also one thing to note is that there would be a change in the response that the webhooks receives for responses. Earlier it would have been **inPerson**, now it would be the actual value say 'New York'. I am not sure if people would be relying on it.

It seems to be good as a hotfix, but we need to fix this properly.

I am trying a few things before approving.

hariombalhara: I pushed a change to send _type along. Could you review that

hariombalhara: Leave it, I reverted my change. I think easiest would be for us to use just the slugified label as ID for location and it's practical to assume that locations would have different labels. If needed we can work on having an ID later.

I think an important followup would be to use label only for all types of locations and change everywhere to do lookup based on label instead of type of location.

apps/web/playwright/event-types.e2e.ts
```diff
@@ -142,17 +142,25 @@ test.describe("Event Types tests", () => {
...
 
+      /**
+       * Verify first organizer address
+       */
       await page.goto(previewLink ?? "");
-
       await selectFirstAvailableTimeSlotNextMonth(page);
-
-      for (const location of locationData) {
-        await page.locator(`span:has-text("${location}")`).click();
```
hariombalhara, line 150: The test was selecting all locations one by one and thus in the end only last location was selected which would work fine.

Now we are testing selection of two different locations one by one

hariombalhara (dismissed):
LGTM now !! Great work @Udit-takkar and thank you 🙏. A pretty simple temporary solution for now.

I have updated a test to ensure that it doesn't regres

hariombalhara (approved):

shivamklr (commented):
Tested Works.

Issue #11275, opened 2023-09-11
[CAL-2436] feat: make jitsi video ready for self-hosters


PeerRich:
right now the jitsi app only works with hosted jitsi links. it would be nice if self-hosters could change the "JITSI_URL" inside the admin key manager

<sub>[CAL-2436](https://linear.app/calcom/issue/CAL-2436/feat-make-jitsi-video-ready-for-self-hosters)</sub>

PeerRich:
go for it @neo773! 🥳

hariombalhara:
This is already supported through jitsiHost 
See it working on main(tested locally)
![image](https://github.com/calcom/cal.com/assets/1780212/567ba27c-1652-4727-8b25-f3533fba7cac)

PeerRich:
turns out we already have support for self-hosted jitsi, my bad!

PR #11326, opened 2023-09-13
fix: #11275 Jitsi self hosted URls fix

Hitesh-Sisara:
## What does this PR do?
Made self hosted jitsi links working, admin user can  change the "JITSI_URL" inside the admin key manager

Fixes issue #11275

/claim #11275

 Loom Video: https://www.loom.com/share/Apps-or-Calcom-13-September-2023-f620a0d2d00840e8992098155e9ce75c?sid=502da561-78de-401e-8644-c24945f06bd7


hariombalhara (changes requested):
The feature is already supported through `jitsiHost`. We shouldn't remove that variable. Existing configurations would break this way.
If we want to add validation to it that it must start with HTTP/HTTPS protocol that can be done.

Hitesh-sisara:
@hariombalhara variable names changed to same as the existing implementation

PR #11327, opened 2023-09-13
fix: Duplicate calendar events

joeauyeung:
## What does this PR do?

This PR aims to fix duplicate calendar events being created for team members and dynamic event participants. A full explanation of what happened can be found here: https://threads.com/34535947634?s=2qEeCrUqS5rpJPa2fo7kjU

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a team of users
- Connect different GCals to those users
- **IMPORTANT** those user's emails should be the emails being used to connect the calendar
- Now create one more user that is connected to Outlook
- Book a collective event type with all three users
    - There should be only one event created in GCal and one event in Outlook
- Repeat this again for round robin and a dynamic event with those three users

## Mandatory Tasks

## Checklist


packages/core/EventManager.ts
```diff
@@ -338,9 +338,18 @@ export default class EventManager {
...
     let createdEvents: EventResult<NewCalendarEventType>[] = [];
     if (event.destinationCalendar && event.destinationCalendar.length > 0) {
-      for (const destination of event.destinationCalendar) {
-        if (destination.credentialId) {
-          let credential = this.calendarCredentials.find((c) => c.id === destination.credentialId);
+      // Want to only create one event per calendar app as they sync across their own platforms
+      const destinationCalendarApps = event.destinationCalendar.reduce((apps, calendar) => {
+        if (!apps.includes(calendar.integration)) apps.push(calendar.integration);
+        return apps;
+      }, [] as string[]);
```
joeauyeung, line 345: We only want to create one event per calendar app platform. This is because if we add the correct calendar ids then those events will be pushed to other calendars on the same platform. 

I tested this on a team that contained GCal and Outlook integrations and events created on GCal did not get pushed to Outlook.

packages/core/EventManager.ts
```diff
@@ -338,9 +338,18 @@ export default class EventManager {
...
+      for (const destination of destinationCalendarApps) {
+        const destinationCalendar = event.destinationCalendar.find(
+          (calendar) => calendar.integration === destination
+        );
+        if (destinationCalendar?.credentialId) {
+          // Doesn't matter who creates the event for team events
+          let credential = this.calendarCredentials.find((c) => c.id === destinationCalendar?.credentialId);
```
joeauyeung, line 352: Since this one event was being created per calendar app I don't think it matters too much who's event this is being created on. 

Users shouldn't be changing calendar events on the calendar anyways, changes should be made on the Cal.com booking.

joeauyeung (commented):
Inital review. If everything looks good then I can translate this to the reschedule methods and other calendar apps.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -28,11 +29,13 @@ export default class GoogleCalendarService implements Calendar {
...
 
-  constructor(credential: CredentialPayload) {
+  constructor(credential: CredentialWithAppName) {
     this.integrationName = "google_calendar";
     this.auth = this.googleAuth(credential);
     this.log = logger.getChildLogger({ prefix: [`[[lib] ${this.integrationName}`] });
+    this.credential = credential;
   }
 
   private googleAuth = (credential: CredentialPayload) => {
```
emrysal, line 41: This credential is no longer needed as googleAuth can access this.credential

packages/core/EventManager.ts
```diff
@@ -338,9 +338,18 @@ export default class EventManager {
...
           if (!credential) {
             // Fetch credential from DB
             const credentialFromDB = await prisma.credential.findUnique({
```
emrysal, line 355: We need to deprecate this; as this is a non-bulk action and will be very slow if invoked

packages/app-store/googlecalendar/lib/CalendarService.ts
joeauyeung: We needed the email of the owner of the credential to not include their email in the list of team members to prevent duplicate events. `googleAuth` doesn't contain this. To get around it I've made `credentialUserEmail` a parameter in the create event method.

packages/core/EventManager.ts
alannnc: I think we should leave this as it is, we better support every host calendar app. And we just do better filtering with Google that has enhanced behaviour for existing users when being added on an event.

joeauyeung: Tested this with two Outlook calendars and it didn't duplicate events. There was still a duplicate event being created with two GCals so I'll change the logic to only create one event for only GCal.

packages/app-store/applecalendar/lib/CalendarService.ts
```diff
@@ -1,8 +1,8 @@
 import CalendarService from "@calcom/lib/CalendarService";
-import type { CredentialPayload } from "@calcom/types/Credential";
+import type { CredentialWithAppName } from "@calcom/types/Credential";
```
joeauyeung, line 2: Changing to type `CrednetialWithAppName` because the email is required but the id was always optional.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -194,19 +205,22 @@ export default class GoogleCalendarService implements Calendar {
 
   async updateEvent(uid: string, event: CalendarEvent, externalCalendarId: string): Promise<any> {
     return new Promise(async (resolve, reject) => {
-      const [mainHostDestinationCalendar] =
-        event?.destinationCalendar && event?.destinationCalendar.length > 0 ? event.destinationCalendar : [];
+      const mainHostDestinationCalendar = event.destinationCalendar?.find(
+        (cal) => cal.externalId === externalCalendarId
+      );
```
joeauyeung, line 210: When rescheduling events we know the external id of the calendar so we can just look for it in the destinationCalendar array.

packages/core/EventManager.ts
```diff
@@ -338,7 +341,25 @@ export default class EventManager {
...
+              destintaionCals.push(cal);
+            }
+          } else {
+            destintaionCals.push(cal);
+          }
+          return destinationCalendars;
+        },
+        [] as DestinationCalendar[]
+      );
+      for (const destination of destinationCalendars) {
```
joeauyeung, line 362: @alannnc @emrysal filtering to only create one GCal event among all destination calendars. Other calendar integrations will still get an event per calendar.

packages/app-store/googlecalendar/lib/CalendarService.ts
zomars: I do think this is a very destructive refactor. Feels like we're patching instead of solving for long term.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -118,9 +129,9 @@ export default class GoogleCalendarService implements Calendar {
             id: String(calEventRaw.organizer.id),
             responseStatus: "accepted",
             organizer: true,
-            email: selectedHostDestinationCalendar?.externalId
-              ? selectedHostDestinationCalendar.externalId
-              : calEventRaw.organizer.email,
+            displayName: calEventRaw.organizer.name,
```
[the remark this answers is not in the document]
zomars: This should be a comment in code

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -118,9 +129,9 @@ export default class GoogleCalendarService implements Calendar {
...
+            email: selectedHostDestinationCalendar?.externalId ?? calEventRaw.organizer.email,
+            self: true,
```
zomars, line 134: What does self means here?

packages/app-store/googlecalendar/lib/CalendarService.ts
zomars: This should be a comment in code

packages/lib/CalendarService.ts
```diff
@@ -140,7 +142,15 @@ export default abstract class BaseCalendarService implements Calendar {
...
           ...getAttendees(event.attendees),
-          ...(event.team?.members ? getAttendees(event.team.members) : []),
+          ...(event.team?.members
+            ? getAttendees(
+                event.team?.members
+                  ? getAttendees(
+                      event.team.members.filter((member) => member.email !== this.credentialUserEmail)
+                    )
+                  : []
+              )
```
zomars, line 152: Why 😭

zomars (changes requested):
I trust this PR works. But By merging this in this state I feel like we're making CalendarService harder to debug. Will have to block to make some improvements.

joeauyeung:
> I trust this PR works. But By merging this in this state I feel like we're making CalendarService harder to debug. Will have to block to make some improvements.

While I do agree this will make things harder, this bug has been causing a lot of headaches for users and touches a core component of our app. How would we feel about reverting the PR that caused the bug and starting from there?

packages/core/EventManager.ts
zomars: I still need to test this

packages/types/Credential.d.ts
```diff
@@ -6,20 +6,10 @@ import type { Prisma } from "@prisma/client";
...
-  };
+  select: typeof import("@calcom/prisma/selects/credential").credentialForCalendarServiceSelect;
 }>;
 
 export type CredentialFrontendPayload = Omit<CredentialPayload, "key"> & {
   /** We should type error if keys are leaked to the frontend */
   key?: never;
 };
-
-export type CredentialWithAppName = CredentialPayload & { appName: string };
```
zomars, line 25: appName was basically the credential appId

packages/app-store/larkcalendar/lib/CalendarService.ts
```diff
@@ -189,7 +194,9 @@ export default class LarkCalendarService implements Calendar {
   async updateEvent(uid: string, event: CalendarEvent, externalCalendarId?: string) {
     const eventId = uid;
     let eventRespData;
-    const [mainHostDestinationCalendar] = event.destinationCalendar ?? [];
+    const mainHostDestinationCalendar = event.destinationCalendar?.find(
+      (cal) => cal.externalId === externalCalendarId
+    );
```
zomars, line 199: I see more `const [mainHostDestinationCalendar]` in this file. Should we update those as well?

zomars:
> > I trust this PR works. But By merging this in this state I feel like we're making CalendarService harder to debug. Will have to block to make some improvements.
> 
> While I do agree this will make things harder, this bug has been causing a lot of headaches for users and touches a core component of our app. How would we feel about reverting the PR that caused the bug and starting from there?

Then we would be reintroducing the bug that on solved and so on and so on... I get the urgency but also I'm trying to keep the code as easy to debug as possible. This will help us find and prevent these kind of stuff in the first place.

Made a significant refactor, made sure types and builds are passing. Can you please verify the bug is still solved before merging?

zomars (dismissed):

alannnc (changes requested):
Find some issues, fixing them now..

alannnc (dismissed):
It's working now as expected. The main issue is solved now, but I think we can checkout all the other changes that are using the new email param when creating an event.
@zomars @joeauyeung

joeauyeung:
@zomars @alannnc thanks for helping make the PR more maintainable. I tested this out and the fix is still working.

keithwillcode:
@shivamklr This is a great example of a PR we need to follow up with automated tests for.

zomars (approved):

packages/core/EventManager.ts
```diff
@@ -338,33 +341,45 @@ export default class EventManager {
...
+      const destinationCalendars: DestinationCalendar[] = event.destinationCalendar.reduce(
+        (destinationCals, cal) => {
+          if (cal.integration === "google_calendar") {
+            if (gCalAdded) {
+              return destinationCals;
+            } else {
+              gCalAdded = true;
+              destinationCals.push(cal);
+            }
+          } else {
```
hariombalhara, line 355: This should be handled in google calendar App. Also, are we sure that no other calendar app has the same behaviour as Google Calendar as that might still be having duplicate events.
cc @joeauyeung @alannnc

joeauyeung: I was thinking about that but the GCal app is called on an individual calendar level so it wouldn't know if other calendars have already created the event.

I tested this with Outlook since it's our second biggest calendar and it wasn't having the same issue.

hariombalhara: We have got to change the abstraction for CalendarService then. We could start by having createEvent accept another argument(destinationCalendars) that can for now be used only by Google Calendar app.

Issue #11319, opened 2023-09-12
[CAL-2459] insights upgrade screen has no title and subtitle for upgrade screen


PeerRich:
insights upgrade screen:

![CleanShot 2023-09-12 at 23 05 37@2x](https://github.com/calcom/cal.com/assets/8019099/02b77a88-f35e-480e-925a-d45d8cbc0702)

misses the headline and subtitle, like routing forms:

![CleanShot 2023-09-12 at 23 05 52@2x](https://github.com/calcom/cal.com/assets/8019099/5dfd5f6c-a119-49de-947e-631ee7d8aad4)

<sub>[CAL-2459](https://linear.app/calcom/issue/CAL-2459/insights-upgrade-screen-has-no-title-and-subtitle-for-upgrade-screen)</sub>

void-main1812:
I can work on this issue can you assign this one to me ?

PeerRich:
sure

PeerRich:
tho #11328 has already been submitted

PR #11328, opened 2023-09-13
fix: insights upgrade screen has no title and subtitle for upgrade screen

MeenuyD:
Address the issue where the "Insights Upgrade" screen was missing a title and subtitle. Resolve this problem by adding the necessary headline and subtitle to the screen, ensuring consistency with routing forms and providing a more informative user experience.
#11319


MeenuyD:
Insights_screenshot

sean-brydon (approved):
Thank you

PR #11335, opened 2023-09-13
fix: Blank state wouldn't set 'schedule'; resulting toggle had no effect

emrysal:
## What does this PR do?

Fixes the common availability not having an effect if you didn't change the selected availability

Also minor performance improvement by requesting less data from availabilities. (just date,start,end,days)


CarinaWolli:
@emrysal  Can you describe in more detail what the current issue is, I am not sure how to reproduce it

CarinaWolli (approved):
Looks good 👏🏻

PR #11337, opened 2023-09-13
fix: updating workflow that is active on managed event type

CarinaWolli:
## What does this PR do?

Fixes issue that workflows that are active on managed event types could not be updated. 

## Type of change

## How should this be tested?

- Create a managed event type and assign team members to it 
- Create a team workflow and set it active for the managed event type 
- Save workflow
- Go back to workflow 
- Click save again => this used to fail and works now

## Mandatory Tasks

### DIFF
--- a/packages/trpc/server/routers/viewer/workflows/update.handler.ts
+++ b/packages/trpc/server/routers/viewer/workflows/update.handler.ts
@@ -92,6 +92,7 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
       id: {
         in: activeOn,
       },
+      parentId: null,
     },
     select: {
       id: true,


packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -92,6 +92,7 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
       id: {
         in: activeOn,
       },
+      parentId: null,
```
CarinaWolli, line 95: We are adding the children later in `activeOnWithChildren`. On the first update when the managed event type is added only the parentId is coming in

emrysal (approved):
Looks good.

shivamklr (commented):
Tested, works.

PR #11344, opened 2023-09-13
test: RFC: tests for EventTypeAppCardInterface (CALCOM-11005 )

gitstart-calcom:
## DEMO

- This test is testing only the `packages/app-store/_components/EventTypeAppCardInterface.tsx` component:

image


packages/app-store/_components/eventTypeAppCardInterface.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  LockedIcon: <div>MockedIcon</div>,
+  disabled: false,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+} as any;
+
+describe("Tests for EventTypeAppCard component", () => {
+  test("Should render DynamicComponent with correct slug", () => {
+    render(<EventTypeAppCard {...mockProps} />);
+
+    expect(DynamicComponent).toHaveBeenCalledWith(
```
hariombalhara, line 43: I feel we should test that `EventTypeAppCardInterface` of the appropriate app has been rendered here with correct props. It would test `DynamicComponent` as well in that way. But I think that's fine because that's what the purpose of the `EventTypeAppCardInterface` component is, it doesn't render anything of it's own in UI, it just delegates to the respective app's EventTypeAppCardInterface

hariombalhara: Also, we should test stripe app as one of the apps as that has a special handling in the code.
Screenshot 2023-09-15 at 3 40 03 PM

hariombalhara (dismissed):
Thanks @gitstart team 🙏 Left a comment to improve the tests.

packages/app-store/_components/eventTypeAppCardInterface.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+      }),
+      {}
+    );
+
+    expect(screen.getByText("MockedDynamicComponent")).toBeInTheDocument();
+  });
+
+  test("Should provide the correct values to EventTypeAppContext", () => {
+    render(
+      <EventTypeAppContext.Provider
```
hariombalhara, line 55: Why are we giving it a provider. Doesn't EventTypeAppCardInterface add the provider itself?

gitstart-calcom: You're right, Wrapping it in the test with another EventTypeAppContext.Provider is redundant. We will make the changes

gitstart-calcom:
We changed the test to fix this issue, please take a look @hariombalhara

packages/app-store/_components/eventTypeAppCardInterface.test.tsx
gitstart-calcom: Alright, we will implement it

keithwillcode (approved):

PR #11350, opened 2023-09-14
chore: Outlook `processBusyTimes` iterate busy times once

joeauyeung:
## What does this PR do?

This PR improves the performance of how we process the busy times from Outlook by only iterating through the times once instead of twice.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Connect an Outlook calendar with busy times
- Select those calendars to check availability on
- When booking an event ensure that those events are blocked

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/app-store/office365calendar/lib/CalendarService.ts
+++ b/packages/app-store/office365calendar/lib/CalendarService.ts
@@ -454,12 +454,13 @@ export default class Office365CalendarService implements Calendar {
       (acc: BufferedBusyTime[], subResponse: { body: { value?: BodyValue[]; error?: Error[] } }) => {
         if (!subResponse.body?.value) return acc;
         return acc.concat(
-          subResponse.body.value
-            .filter((evt) => evt.showAs !== "free" && evt.showAs !== "workingElsewhere")
-            .map((evt) => ({
+          subResponse.body.value.reduce((acc: BufferedBusyTime[], evt: BodyValue) => {
+            if (evt.showAs === "free" || evt.showAs === "workingElsewhere") return acc;
+            return acc.concat({
               start: evt.start.dateTime + "Z",
               end: evt.end.dateTime + "Z",
-            }))
+            });
+          }, [])
         );
       },
       []


packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -454,12 +454,13 @@ export default class Office365CalendarService implements Calendar {
       (acc: BufferedBusyTime[], subResponse: { body: { value?: BodyValue[]; error?: Error[] } }) => {
         if (!subResponse.body?.value) return acc;
         return acc.concat(
-          subResponse.body.value
-            .filter((evt) => evt.showAs !== "free" && evt.showAs !== "workingElsewhere")
-            .map((evt) => ({
```
joeauyeung, line 459: If we chain `.filter` and `.map` we are essentially iterating through the array twice. By replacing this with a `.reduce` we're able to filter through values while only iterating through the array once.

alannnc (approved):

packages/app-store/office365calendar/lib/CalendarService.ts
hariombalhara: `filter` and `map` usage keeps the code clean and simple to understand. Unless the array is huge it doesn't really matter performance wise how many times we iterate over it.

What's the performance improvement in here?

hariombalhara (approved):
LGTM! Left one comment to get clarity

packages/app-store/office365calendar/lib/CalendarService.ts
PeerRich: lets do this in a follow up PR

joeauyeung: @hariombalhara there was a user that was experiencing a long load time for availability from their Outlook calendar. From what I gathered the payload was very large. I think most of the time the busy times array is small enough that it's fine to do the `filter` and `map` but in this scenario, we're running into the theoretical O(2n) vs O(n).

hariombalhara: Thanks. Makes sense then.

PR #11009, opened 2023-08-29
fix: improve UX of team invite by copy link

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/10906

1) checked user session in getServersideprops to speed up redirecting to /login page if unauthenticated instead of showing unauthorised error on /teams and then redirecting which was very slow.

2) added alert to ask user to "sign in or sign up to see team invite".

https://github.com/calcom/cal.com/assets/53316345/7af9a602-6920-4910-8b01-a7b1af111cf3

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #11356, opened 2023-09-14
fix: org invite

### LINKED ISSUE #11311: [CAL-2457] copying an invite link doesn't account for orgs + regular team expiration issue
### Issue Summary

When the "Copy invite link" button in the invitation modal is used within a org context, it copies the the link which then doesn't account for an org context.

Also, aside from the org context, it seems like the created invite link for any regular team as well may always have an expired token as it gets created using `expires: new Date()` which refers to the exact moment of the token creation, not a future time.

### Steps to Reproduce

1. Create an org
3. In the add members modal for the org, select "Copy invite link"
4. Use that link to join the org in an incognito window

### Actual Results

Either get an error resulted because the token is expired or you get to see the login page with a message "You need to Sign in or Sign up to see team invitation." which is wrong.

### Expected Results

For an org context, you need to get redirected to `signup?token=` and never get an expiration issue.

<sub>[CAL-2457](https://linear.app/calcom/issue/CAL-2457/copying-an-invite-link-doesnt-account-for-orgs)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/11311

https://github.com/calcom/cal.com/assets/53316345/1d6eae9a-b9b2-421b-a2de-e43fc10f089d

How to test this ?

1) Create an organization
2) Go to http://app.cal.local:3000/settings/organizations/members and click on +Add
3) Click on "Copy Invite Link" and open the link in incognito mode or another browser
4) Sign up

**NOTE**: While testing locally you might have to comment the line await navigator.clipboard.writeText(inviteLink) and get the invite link by logging in on console.

## Type of change


apps/web/pages/api/auth/signup.ts
```diff
@@ -116,12 +117,18 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         });
       }
 
-      const membership = await prisma.membership.update({
+      const membership = await prisma.membership.upsert({
```
Udit-takkar, line 120: membership doesn't exist when inviting by "copy link".

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -87,7 +88,12 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
   });
 
   const copyInviteLinkToClipboard = async (token: string) => {
-    const inviteLink = `${WEBAPP_URL}/teams?token=${token}`;
+    const isOrgInvite = isOrg;
+    const teamInviteLink = `${WEBAPP_URL}/teams?token=${token}`;
+    const orgInviteLink = `${WEBAPP_URL}/signup?token=${token}&callbackUrl=/getting-started`;
+
+    const inviteLink = isOrgInvite ? orgInviteLink : teamInviteLink;
+
```
Udit-takkar, line 96: different invite link for org because it is not a team and won't show up in /teams page

apps/web/pages/teams/index.tsx
```diff
@@ -41,6 +42,20 @@ function Teams() {
...
+
+  if (!session) {
+    return {
+      redirect: {
+        destination: callbackUrl ? `/auth/login?callbackUrl=${callbackUrl}` : "/auth/login",
+        permanent: false,
+      },
+      props: {},
+    };
+  }
```
Udit-takkar, line 58: this was removed when this PR https://github.com/calcom/cal.com/pull/11009 was reverted

apps/web/pages/signup.tsx
```diff
@@ -278,15 +310,21 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
     username = available ? username : suggestion || username;
   }
 
+  const isValidEmail = checkValidEmail(verificationToken.identifier);
+  const isOrgInviteByLink = orgSlug && !isValidEmail;
+
```
Udit-takkar, line 315: when inviting user by copy link the identifier is something like "invite-link-for-teamId-2" which is not an email because we don't know the email yet

apps/web/pages/signup.tsx
```diff
@@ -118,21 +140,31 @@ export default function Signup({ prepopulateFormValues, token, orgSlug }: Signup
...
+                      addOnLeading={
+                        orgSlug
+                          ? `${getOrgFullDomain(orgSlug, { protocol: true })}/`
+                          : `${process.env.NEXT_PUBLIC_WEBSITE_URL}/`
+                      }
+                      {...register("username")}
+                      disabled={!!orgSlug}
+                      required
+                    />
+                  )}
```
Udit-takkar, line 167: hiding username field only in case of inviteByLink to org because we generate it from the email entered and it is disabled for user to change

leog (dismissed):
Code looks good, tested with org members and works great. I also tested with a subteam but it redirected me to login and it should behave the same way as org but as a result not only adding me to the subteam but also de org itself.

apps/web/pages/api/auth/signup.ts
hariombalhara: Is it worth adding a comment in the code itself?

apps/web/pages/api/auth/signup.ts
```diff
@@ -116,12 +117,18 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
         },
-        data: {
+        update: {
+          accepted: true,
+        },
+        create: {
+          userId: user.id,
+          teamId: team.id,
           accepted: true,
+          role: MembershipRole.MEMBER,
```
hariombalhara, line 131: Inviting by copy always adds a user as a **member**? Looks like we should support the role in it.

hariombalhara: Also, should we not create a membership for the parent team as well? I have seen in the codebase that when creating a membership we create a membership with parent team as well if it exists.
Screenshot 2023-09-15 at 4 00 03 PM

apps/web/pages/teams/index.tsx
hariombalhara: How is an invite to a team of org supposed to work? I copied the invite link and opened in a new incognito window to join by that link. But there was no session so it just redirected me to login page.

hariombalhara (dismissed):
Can't join a Team by copied link. Also, couldn't join the org by invite email(though that issue exists in main also)

apps/web/pages/api/auth/signup.ts
hariombalhara: Noticedthat inviting to a team link is different and doesn't bring to signup page. So, this might be fine.

Udit-takkar:
@hariombalhara @leog yeah thanks for review. i am fixing the invite to subteam

hariombalhara:
Not a blocker for this PR but I see an option to add existing team members to a team. Adding those who have accepted the request, throws error but adding those who are in Pending state silently ignores that it's already added.
https://www.loom.com/share/48bd8659a4424185b7e42ff3eff98af3

hariombalhara:
Not sure, if related to the changes in here but I see that members get duplicated in the list when inviting a new team member. 
https://www.loom.com/share/f4d4b8d821db4cd1bc77ab46fa08f5d0

hariombalhara (approved):
Looks mergeable now !!

Udit-takkar:
> Not a blocker for this PR but I see an option to add existing team members to a team. Adding those who have accepted the request, throws error but adding those who are in Pending state silently ignores that it's already added. https://www.loom.com/share/48bd8659a4424185b7e42ff3eff98af3

OK i'll open up issue for this

PR #11375, opened 2023-09-15
Main

### LINKED ISSUE #11113: There Should be a Loading icon when click on button events

When I am onboarding on the app. Every Time I click the Next or Continue button I am confused is that clicked or not. Because there is no loading effect or any loader showing after clicking on the button.

There should be a loader icon after clicking on the buttons.

We can also disable the button after click them once.

### Additional context

(Write your answer here.)

### Requirement/Document

(Share it here.)

RohitChauhan98:
## What does this PR do?

This PR add a top loading bar when the user click on "sign in" button in the login page and go "back to the login" Button on logout  page

Fixes # (issue)

Fixes #11113

## Type of change

## How should this be tested?

while Signing in notice a top loading bar

### DIFF
--- a/apps/web/pages/auth/login.tsx
+++ b/apps/web/pages/auth/login.tsx
@@ -9,6 +9,7 @@ import type { CSSProperties } from "react";
 import { useState } from "react";
 import { FormProvider, useForm } from "react-hook-form";
 import { FaGoogle } from "react-icons/fa";
+import LoadingBar from "react-top-loading-bar";
 import { z } from "zod";
 
 import { SAMLLogin } from "@calcom/features/auth/SAMLLogin";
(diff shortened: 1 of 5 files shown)


apps/web/pages/auth/login.tsx
```diff
@@ -174,6 +186,7 @@ export default function Login({
           "--cal-brand-subtle": "#9CA3AF",
         } as CSSProperties
       }>
+      <LoadingBar color="#f11946" progress={progress} onLoaderFinished={() => setProgress(0)} />
```
hariombalhara, line 189: I think we don't want to add a separate loading bar. How about we move the button to loading state(by setting loading=true)Once the navigation completes, the button itself would go away.

This would give a clear indication to the user that the action is being taken/.
cc @sean-brydon

hariombalhara (changes requested):
Thanks @RohitChauhan98 🙏
Left a comment suggesting an alternative approach. 

Also, please update the PR title

PeerRich:
artificial setTimeout is never a good idea. lets just add Loading to the button state. we already have a prop for this

PR #11390, opened 2023-09-15
fix: org username check for dynamic and signup

### LINKED ISSUE #11386: [CAL-2476] Availability from org account used for dynamic events
### Issue Summary

When org account and normal cal account have the same username the wrong availability is shown in dynamic events. A dynamic event on app.cal.com should show the availability from the normal cal account but it shows the availability from the org account. 

<sub>[CAL-2476](https://linear.app/calcom/issue/CAL-2476/availability-from-org-account-used-for-dynamic-events)</sub>

leog:
## What does this PR do?

Dynamic event types involving usernames shared with orgs was picking up the incorrect availability for calculation. Also, username existence check in an org context was broken now that we define username upon invitation, this fixes that too.

Fixes #11386

## Type of change

## How should this be tested?

Have two users with the same usernames, one belonging to an org and another as plain user, change availability to be different between them and try to book a dynamic event with the plain user with another person in the app, it was picking up availability for both usernames inside and out the org. Also, invite a person to an org and use the invitation link in the email, it was throwing an error as username already existed.

### DIFF
--- a/packages/app-store/salesforce/lib/CalendarService.ts
+++ b/packages/app-store/salesforce/lib/CalendarService.ts
@@ -196,7 +196,6 @@ export default class SalesforceCalendarService implements Calendar {
   }
 
   async createEvent(event: CalendarEvent): Promise<NewCalendarEventType> {
-    debugger;
     const contacts = await this.salesforceContactSearch(event);
     if (contacts.length) {
       if (contacts.length == event.attendees.length) {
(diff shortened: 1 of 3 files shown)


packages/app-store/salesforce/lib/CalendarService.ts
```diff
@@ -196,7 +196,6 @@ export default class SalesforceCalendarService implements Calendar {
   }
 
   async createEvent(event: CalendarEvent): Promise<NewCalendarEventType> {
-    debugger;
```
leog, line 199: Removing as I searched for "debugger" in the entire app to remove mine and found this one too

CarinaWolli (approved):
Tested and works 👍

PR #11391, opened 2023-09-15
fix: Removed resize option from bio editor

### LINKED ISSUE #11378: [CAL-3954] Hard to Focus on Text Area to enter User About description

It is very hard to focus on the text area to enter the description in About section. And also the formatting options are just above the text area. So we i can't able to find the focus of text area.

Try to increase the text area height so that we can easily focus to text area. And also shift the formatting options to right of text area .

<sub>[CAL-3954](https://linear.app/calcom/issue/CAL-3954/hard-to-focus-on-text-area-to-enter-user-about-description)</sub>

thepradipvc:
## What does this PR do?

I removed the resize option from the bio editor to avoid unexpected issues due to user interaction.

Fixes #11378

## Mandatory Tasks

### DIFF
--- a/packages/ui/components/editor/stylesEditor.css
+++ b/packages/ui/components/editor/stylesEditor.css
@@ -33,7 +33,6 @@
   border-bottom-left-radius: 6px;
   border-bottom-right-radius: 6px;
   overflow: scroll;
-  resize: vertical;
   height: auto;
   min-height: 40px;
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,46 +5,6 @@ __metadata:
   version: 6
   cacheKey: 8
 
-"@0no-co/graphql.web@npm:^1.0.1":
-  version: 1.0.4
-  resolution: "@0no-co/graphql.web@npm:1.0.4"
-  peerDependencies:
-    graphql: ^14.0.0 || ^15.0.0 || ^16.0.0
-  peerDependenciesMeta:
-    graphql:
-      optional: true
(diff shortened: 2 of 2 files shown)


up1512001:
You can add fix: prefix in your PR name so it can pass the Validate PRs check.

sean-brydon (changes requested):
I dont think we should remove the resize - we should just increase the default min-height for this component. I'd prefer to do this on a usecase basis also instead of all editors. 

Please also uncommit yarn.lock

git checkout origin/main -- yarn.lock

packages/ui/components/editor/stylesEditor.css
```diff
@@ -33,7 +33,6 @@
   border-bottom-left-radius: 6px;
   border-bottom-right-radius: 6px;
   overflow: scroll;
-  resize: vertical;
```
Udit-takkar, line 36: why did you remove this?

ChaudharyPradip: See, I don't think there is a need to increase the min-height for the editor because it automatically gets bigger when the content gets longer.

However, if a user uses the resize controller to resize the height of the editor. It will then stop adjusting to the size of the content. It will no longer adapt to the content length but if we remove the option to resize the editor it will always adapt to the content.

But, I'm not aware of any other place where the editor size might need to be adjusted by the user (I don't think it would be a necessity). So if that's the case then please ignore this change.

RavishankarMahto:
**I can solve this [issue.**

RavishankarMahto:
I can solve this issue.

Issue #11393, opened 2023-09-15
During the Event Edit It Should be Update


anisur2805:
When edit the screen the button label should be **Update** rather than **Save**
![image](https://github.com/calcom/cal.com/assets/11018986/5e3b3428-88fc-4757-950d-66cf1c6a77a6)

Udit-takkar:
"Save" is fine here

PR #11401, opened 2023-09-16
#11393-Changed-it-to-update

PriyamAgarwal999:
#11393 Fix the issue 
Replaced the Save with Update.


PriyamAgarwal999:
#11393 changed the save to update

PR #11402, opened 2023-09-16
change-to-update

PriyamAgarwal999:
#11393 - fixed -the -bug-for-During-the-edit-event-it-should-be-update

Changed the save to update


Udit-takkar (commented):
@PriyamAgarwal999 i don't know why you raised too many duplicate PRs https://github.com/calcom/cal.com/pull/11403#pullrequestreview-1639205827

PR #11403, opened 2023-09-16
Fix:#11393-Change-to-update

PriyamAgarwal999:
Fix:#11393-During-event-edit-it-should-be-update.


Udit-takkar (commented):
@PriyamAgarwal999 thanks for looking into this but using "save" is also fine here.

PriyamAgarwal999:
I am new to open source, so created duplicate PR's, Appologies for the same.

Issue #11394, opened 2023-09-15
Minimum Time is not correct.


anisur2805:
This time should not be less then 10 minutes during the update time.
![image](https://github.com/calcom/cal.com/assets/11018986/13204dde-187b-4c82-b8ac-b68115491874)

up1512001:
@anisur2805 can you further explain how to replicate this bug?

abhinaisai2002:
can i work on this issue

yavinash249:
@anisur2805  can you give the path?

abhinaisai2002:
which path?

adarshSrivastava01:
Is it a global configuration that this could not be less than 10 minutes?

PeerRich:
why not less than 10?

anisur2805:
@PeerRich because the minimum time is 10 minutes. And here i found its work less than 10.

PR #11407, opened 2023-09-17
fix: Minimum Duration Limit for event type creation

adarshSrivastava01:
## What does this PR do?

This PR fixes the issue where the minimum duration to create a event-type was set to be 1 minute, this PR includes the change where it updates this duration to be 10 minutes.

Fixes # (issue)

This PR fixes the issue: #11394

## Requirement/Documentation

## Type of change

- [ ] 
## Mandatory Tasks


PeerRich:
less than 10 is fine

PR #11418, opened 2023-09-18
fix: Prefill issue when name isn't present and add unit tests

### LINKED ISSUE #11290: [CAL-2444] Unit Test: `getBookingResponsesSchema`
Unit tests for the method would be quite important as that is used to validate the data when prefilling, or when submitting to backend. It uses quite some zod code with lots of conditions depending on field type, we need to exhaustively test it.

I am not sure at the moment what tests we can write but we need to have atleast a few tests so that we can quickly add more tests.

A bug that should have been caught by these unit tests would be that if someone gives it name="John", it should preprocess it successfully into {firstName:"John", lastName:""} without any validation error.

See [https://github.com/calcom/cal.com/pull/11289/files#diff-77cb3d643a600ea8f6417f61fe657647847ad5cb838f3cef04c65508a3b229a4R126](https://github.com/calcom/cal.com/pull/11289/files#diff-77cb3d643a600ea8f6417f61fe657647847ad5cb838f3cef04c65508a3b229a4R126) for the motivation/need for the tests.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2444](https://linear.app/calcom/issue/CAL-2444/unit-test-getbookingresponsesschema)</sub>

### LINKED ISSUE #11488: [CAL-2504] Prefilled needs name first to not ignore other query params

### Issue Summary

If you prefill data in the url you need to add the name as a query param first, else the res is ignored

### Steps to Reproduce

https://cal.com/askasp/samma?userid=samma  <- does not work
https://cal.com/askasp/samma?name=hei&userid=samma   <- does work.

### Actual Results

- What's happening right now that is different from what is expected
- right now userid is not prefilled without having name, im expecting it to be

<sub>[CAL-2504](https://linear.app/calcom/issue/CAL-2504/prefilled-needs-name-first-to-not-ignore-other-query-params)</sub>

hariombalhara:
## What does this PR do?

Fixes #11488
Fixes #11290

Fixes the bug where 'name' is required for prefilling the Booking Form

Also, adds tests for `getBookingResponsesSchema` including one test that would have caught the regression in #11289
Screenshot 2023-09-18 at 3 25 42 PM

## Type of change
- [x] Unit Tests

## How should this be tested?
- yarn test `getBookingResponsesSchema`

## Mandatory Tasks


packages/features/form-builder/schema.ts
```diff
@@ -289,7 +291,7 @@ export const fieldTypesSchemaMap: Partial<
       if (fields.length === 1) {
         const field = fields[0];
         if (variantSupportedFields.includes(field.type)) {
-          const schema = stringSchema;
+          const schema = field.required && !isPartialSchema ? nonEmptyString() : stringSchema;
```
hariombalhara, line 294: `required` field in name should be nonEmpty if it's marked required and it's not prefilling scenario(where it is okay to have no name or empty name)

packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -79,7 +79,7 @@ function preprocess<T extends z.ZodType>({
             isPartialSchema,
             field,
           });
-          return newResponses;
+          return;
```
hariombalhara, line 82: Unnecessary return value. superRefine has a void return type.

packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -103,7 +103,11 @@ function preprocess<T extends z.ZodType>({
           newResponses[field.name] = value;
         }
       });
-      return newResponses;
+
+      return {
+        ...parsedResponses,
```
hariombalhara, line 108: Make sure that any existing value is forwarded as is.

packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -137,8 +141,10 @@ function preprocess<T extends z.ZodType>({
           continue;
         }
 
-        if (isRequired && !isPartialSchema && !value)
+        if (isRequired && !isPartialSchema && !value) {
           ctx.addIssue({ code: z.ZodIssueCode.custom, message: m(`error_required_field`) });
+          return;
```
hariombalhara, line 146: There was an early return missing here.

packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -248,14 +255,17 @@ function preprocess<T extends z.ZodType>({
           continue;
         }
 
-        throw new Error(`Can't parse unknown booking field type: ${bookingField.type}`);
+        ctx.addIssue({
```
hariombalhara, line 258: Instead of crashing, return a zod error.

packages/features/bookings/lib/getBookingResponsesSchema.ts
```diff
@@ -248,14 +255,17 @@ function preprocess<T extends z.ZodType>({
...
+        });
       }
     })
   );
   if (isPartialSchema) {
     // Query Params can be completely invalid, try to preprocess as much of it in correct format but in worst case simply don't prefill instead of crashing
-    return preprocessed.catch(() => {
-      console.error("Failed to preprocess query params, prefilling will be skipped");
+    return preprocessed.catch(function (res?: { error?: unknown[] }) {
+      console.error("Failed to preprocess query params, prefilling will be skipped", res?.error);
```
hariombalhara, line 268: Makes debugging easy when prefilling fails due to some reason.

CarinaWolli (commented):
Prefilling a booking question without having the name param still doesn't work for me 🤔

hariombalhara:
Thanks for the quick review @CarinaWolli 🙏

I tried again locally and it works for me. See the loom https://www.loom.com/share/074d5e495808492cbfa8c7bcc18c1fce?from_recorder=1&focus_title=1

Could you share your prefill URL and the event-type setup? Also, you should see a console warning when prefilling is skipped due to some error, that could point out to an issue.

zomars (approved):
LGTM

Udit-takkar (approved):
LGTM

PR #11420, opened 2023-09-18
feat: support `firstName` and `lastName` query params

### LINKED ISSUE #11291: [CAL-2445] Support prefilling firstName and lastName using separate query params
It isn't possible to prefill `firstName` and `lastName` from Routing Form if user wants to ask first name and last name separately there. This is because Routing Form doesn't use Form Builder and thus user has to create two separate Short Text fields for First Name and Last Name and these individual fields aren't directly refillable in Booking Form.

Booking Form either needs a single `name` query param with spaces in it and it would automatically break it down in First Name and Last Name OR it needs it in format {"firstName": "John", "lastName": "John Doe"}

We need to support `firstName` and `lastName` separately as well for prefill. See if we can do `name.firstName` and `name.lastName` because that would be the best as it can be applied to any complex field(like name) that consists multiple fields

Workaround: is to ask for full name instead of split names and pass that as `name` and then booking form would automatically split that into first and lastname

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2445](https://linear.app/calcom/issue/CAL-2445/support-prefilling-firstname-and-lastname-using-separate-query-params)</sub>

hariombalhara:
## What does this PR do?

Closes #11291

[Demo Loom](https://www.loom.com/share/033fb175476a40c3a2412e22bac344a9)

## Type of change
- [x] Improvement

## How should this be tested?
- Create an event type and enable Split full name option for name field
- Create a Routing Form and add two short text fields with identifiers `firstName` and `lastName`. Configure that form to Route to the event type 
- Now submit the Routing Form with firstName and lastName prefilled and noticed that when you reach Booking Form page those would be prefilled in Booking Form as well.

## Mandatory Tasks

### DIFF
--- a/apps/web/playwright/manage-booking-questions.e2e.ts
+++ b/apps/web/playwright/manage-booking-questions.e2e.ts
@@ -153,6 +153,25 @@ test.describe("Manage Booking Questions", () => {
           });
         });
       });
+
+      await test.step("Verify that we can prefill name field with firstName,lastName query params", async () => {
+        const searchParams = new URLSearchParams();
+        searchParams.append("firstName", "John");
+        searchParams.append("lastName", "Doe");
+        await doOnFreshPreviewWithSearchParams(searchParams, page, context, async (page) => {
+          await selectFirstAvailableTimeSlotNextMonth(page);
+          await expectSystemFieldsToBeThereOnBookingPage({
+            page,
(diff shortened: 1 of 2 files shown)


sean-brydon (approved):
Sweet!

PR #11421, opened 2023-09-18
feat: signup refactor

sean-brydon:
Redesigns settings 

FIGMA: https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/New-Features?type=design&node-id=5237-479368&mode=design

How to test: 
Testing as a self hoster 
- run yarn dev 
- visit localhost:3000/signup
- Signup with premium username
- signup with normal username
- signup with team invite

Testing as a calcom hoster
- yarn dev
- run website - yarn dev in apps/website
- Signup with premium username -> See checkout flow
- signup with normal username
- signup with team invite

- ensure signup with google works correctly
- ensure sign up with saml works correctly
- ensure query params are filled from cal.com landing page button works correctly
- ensure email verification is sent 

Self hosted:
CleanShot 2023-09-20 at 10 57 41

Screenshots

Ipad pro 11
Sizzy-iPad Pro 11 localhost 21Sep 11 27
Iphone 14
Sizzy-iPhone 14 localhost 21Sep 11 27
Iphone 8
Sizzy-iPhone 8 localhost 21Sep 11 27
Macbook Air
Sizzy-MacBook Air localhost 21Sep 11 27
Ultrawide
CleanShot 2023-09-21 at 11 28 05


apps/web/pages/auth/login.tsx
```diff
@@ -5,6 +5,7 @@ import type { GetServerSidePropsContext } from "next";
 import { getCsrfToken, signIn } from "next-auth/react";
 import Link from "next/link";
 import { useRouter, useSearchParams } from "next/navigation";
+import { NextLink } from "next/navigation";
```
sean-brydon, line 8:
```suggestion
import { NextLink } from "next/link";
```

apps/web/pages/api/auth/signup.ts
sean-brydon, line 1: Moved to a CALCOM handler and a SELFHOSTED handler - to help SOC 

Both still use this endpoint for singing up

packages/lib/server/username.ts
sean-brydon, line 1: Move to a reusable file

apps/web/pages/signup.tsx
```diff
@@ -34,43 +43,130 @@ type FormValues = z.infer<typeof signupSchema>;
...
+    icon: Hexagon,
+  },
+  {
+    title: "scheduling_for_your_team",
+    description: "scheduling_for_your_team_description",
+    icon: Clock,
+  },
+];
+
+function UsernameField({
```
sean-brydon, line 64: Use a specific usename field and debounce the value to not call this on every key press like the zod schema approach is doing in website

apps/web/pages/signup.tsx
```diff
@@ -34,43 +43,130 @@ type FormValues = z.infer<typeof signupSchema>;
...
+}: React.ComponentProps<typeof TextField> & {
+  username: string;
+  setPremium: (value: boolean) => void;
+  premium: boolean;
+  usernameTaken: boolean;
+  setUsernameTaken: (value: boolean) => void;
+}) {
+  const { t } = useLocale();
+  const { register, formState } = useFormContext<FormValues>();
+  const debouncedUsername = useDebounce(username, 600);
```
sean-brydon, line 80: Debounce the username for every 600ms

apps/web/pages/signup.tsx
```diff
@@ -34,43 +43,130 @@ type FormValues = z.infer<typeof signupSchema>;
...
   const {
     register,
-    formState: { errors, isSubmitting },
-  } = methods;
+    watch,
+    formState: { isSubmitting, errors },
+  } = formMethods;
 
-  const handleErrors = async (resp: Response) => {
+  const handleErrorsAndStripe = async (resp: Response) => {
```
sean-brydon, line 155: Stolen from website approach - premium username throws a 402 if a premium username is required. 

We could probably handle this better but it outside the scope of this PR

apps/web/playwright/signup.e2e.ts
```diff
@@ -0,0 +1,193 @@
...
+
+import { test } from "./lib/fixtures";
+
+test.describe.configure({ mode: "parallel" });
+
+test.describe("Signup Flow Test", async () => {
+  test.afterAll(async ({ users }) => {
+    await users.deleteAll();
+    /**
+     * Delete specicially created users for this test
```
alannnc, line 14:
```suggestion
     * Delete specifically created users for this test
```

apps/web/pages/api/auth/signup.ts
```diff
@@ -1,218 +1,49 @@
...
+    throw new HttpError({
+      statusCode: 405,
+      message: "Method not allowed",
+    });
   }
+}
 
+function ensureSignupIsEnabled() {
   if (process.env.NEXT_PUBLIC_DISABLE_SIGNUP === "true") {
-    res.status(403).json({ message: "Signup is disabled" });
```
alannnc, line 23: What is the reason here for code moving/deleting?

apps/web/pages/signup.tsx
```diff
@@ -106,106 +202,203 @@ export default function Signup({ prepopulateFormValues, token, orgSlug, orgAutoA
...
+                type="email"
+                data-testid="signup-emailfield"
+              />
+
+              {/* Password */}
+              <PasswordField
+                data-testid="signup-passwordfield"
+                label={t("password")}
+                {...register("password")}
+                hintErrors={["caplow", "min", "num"]}
```
alannnc, line 277: This is flagging initially right and when password entered it doesn't change.

image

packages/features/auth/signup/handlers/selfHostedHandler.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (foundToken && foundToken?.teamId) {
+    const team = await prisma.team.findUnique({
+      where: {
+        id: foundToken.teamId,
+      },
+    });
+    if (team) {
+      const teamMetadata = teamMetadataSchema.parse(team?.metadata);
+
+      // if (IS_CALCOM && (!teamMetadata?.isOrganization || !!team.parentId)) {
```
alannnc, line 53: Should we remove?

packages/lib/constants.ts
```diff
@@ -33,7 +33,7 @@ export const SENDER_NAME = process.env.NEXT_PUBLIC_SENDGRID_SENDER_NAME || "Cal.
 export const CAL_URL = new URL(WEBAPP_URL).hostname.endsWith(".vercel.app") ? WEBAPP_URL : WEBSITE_URL;
 
 export const IS_CALCOM =
-  WEBAPP_URL &&
+  !!WEBAPP_URL &&
```
alannnc, line 36: This is intentional? Don't forget to remove it if it was made for testing.

packages/features/auth/signup/username.ts
```diff
@@ -0,0 +1,138 @@
...
+
+type CustomNextApiHandler<T = unknown> = (
+  req: RequestWithUsernameStatus,
+  res: NextApiResponse<T>
+) => void | Promise<void>;
+
+export const isPremiumUserName = (username: string): boolean =>
+  username.length <= 4 || Object.prototype.hasOwnProperty.call(wordlist, username);
+
+const generateUsernameSuggestion = async (users: string[], username: string) => {
```
alannnc, line 48: I'm seeing this duplicated, can we import it from lib?

packages/features/auth/signup/username.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    if (!check.available) {
+      req.usernameStatus.statusCode = 418;
+      req.usernameStatus.json.available = false;
+      req.usernameStatus.json.message = "A user exists with that username";
+    }
+
+    req.usernameStatus.json.suggestion = check.suggestedUsername;
+
+    console.log(req.usernameStatus);
```
alannnc, line 87:
```suggestion
```

alannnc (commented):
Left some nit comments, also I'm seeing error in signup page:

image

Overall pretty good job.

apps/web/pages/signup.tsx
alannnc: This works like this, when google autocompletes it marks everything green but when deleting input it doesn't return to original state.

apps/web/pages/signup.tsx
alannnc: This always considers premium logic, we should disable that if it is a self hosted.

alannnc: We are using this already:
```
export const getUsernameAvailabilityComponent = (isPremium: boolean) => {
  if (isPremium)
    return dynamic(() => import("./PremiumTextfield").then((m) => m.PremiumTextfield), { ssr: false });
  return dynamic(() => import("./UsernameTextfield").then((m) => m.UsernameTextfield), { ssr: false });
};
```

alannnc (commented):
Somehow I'm not able to connect website and app signup api/username, where you able to achieve this?

apps/web/pages/api/auth/signup.ts
sean-brydon: CleanShot 2023-09-22 at 08 49 05@2x
Throwing and error to catch internally - using it as a consitant way of handling errors across this route. Passing res around isnt ideal

apps/web/pages/signup.tsx
sean-brydon: Not related to this PR - will tackle in a follow up.

sean-brydon: https://github.com/calcom/cal.com/issues/11486 Created issue for a follow up

packages/features/auth/signup/handlers/selfHostedHandler.ts
sean-brydon: Yes - removed - on self hosted handler we dont care about premium

sean-brydon: yes on self hosted handler we dont care about premium checks

sean-brydon:
> Somehow I'm not able to connect website and app signup api/username, where you able to achieve this?
CleanShot 2023-09-22 at 08 56 03@2x
Ensure you have the .env set for website

also ensure you fake IS_CALCOM to be true in lib/constants

PeerRich:
CleanShot 2023-09-25 at 15 23 31@2x
missing i18n translation?

PeerRich:
quite a few things look better in figma:

figma:
CleanShot 2023-09-25 at 15 25 38@2x

PR: 
CleanShot 2023-09-25 at 15 26 16@2x

PeerRich (dismissed):
still some positioning, spacing and margins not like in figma. should be easy fixes

PeerRich (commented):
still some positioning, spacing and margins not like in figma. should be easy fixes

packages/lib/server/username.ts
```diff
@@ -0,0 +1,141 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import slugify from "@calcom/lib/slugify";
+import prisma from "@calcom/prisma";
+
+import notEmpty from "../../../apps/website/lib/utils/notEmpty";
+import { wordlist } from "../../../apps/website/lib/utils/wordlist/wordlist";
```
zomars, line 7: This will break cloning for al community contributors as they don't have access to these files.

zomars (dismissed):
So Farr looking good. Blocking to address the wordlist solution.

emrysal (changes requested):
Temporarily blocking to review redis

emrysal (dismissed):
Converting to draft as it needs a little more work - but it's basically there.

There was an infinite loop when IS_CALCOM was true, as the checkUsername was pointing to the old website check for premium usernames. updated in what I'm about to commit.

packages/features/auth/signup/handlers/calcomHandler.ts
```diff
@@ -7,14 +7,14 @@ import { sendEmailVerification } from "@calcom/features/auth/lib/verifyEmail";
 import { WEBAPP_URL } from "@calcom/lib/constants";
 import { getLocaleFromRequest } from "@calcom/lib/getLocaleFromRequest";
 import { HttpError } from "@calcom/lib/http-error";
+import { usernameHandler, type RequestWithUsernameStatus } from "@calcom/lib/server/username";
```
emrysal, line 10: Use the re-usable file, not entirely sure which file is the right one but this was the one I touched.

packages/lib/server/checkUsername.ts
```diff
@@ -1,6 +1,6 @@
-import { checkPremiumUsername } from "@calcom/features/ee/common/lib/checkPremiumUsername";
 import { IS_CALCOM } from "@calcom/lib/constants";
 
 import { checkRegularUsername } from "./checkRegularUsername";
+import { usernameCheck } from "./username";
 
-export const checkUsername = !IS_CALCOM ? checkRegularUsername : checkPremiumUsername;
+export const checkUsername = !IS_CALCOM ? checkRegularUsername : usernameCheck;
```
emrysal, line 6: checkPremiumUsername is a fetch function to /api/username.. was a bit loopy.

packages/lib/server/username.ts
```diff
@@ -7,7 +6,7 @@ import prisma from "@calcom/prisma";
 import { IS_CALCOM } from "../constants";
 import notEmpty from "../notEmpty";
 
-const WORDLIST_PATH = process.env.PREMIUM_WORDLIST_PATH_FROM_SERVER ?? "";
+const cachedData: Set<string> = new Set();
```
emrysal, line 9: Sets are way more efficient than arrays and less ugly than key/value objects with 1'd values. Note this is stored in lambda cache so provides a little perf boost (saves the network call)

packages/lib/server/username.ts
```diff
@@ -34,33 +33,23 @@ type CustomNextApiHandler<T = unknown> = (
...
-      ex: 60 * 60 * 24 * 7,
-    });
-
-    return data;
+export async function isBlacklisted(username: string) {
+  // NodeJS forEach is very, very fast (these days) so even though we only have to construct the Set
+  // once every few iterations, it doesn't add much overhead.
+  if (!cachedData.size && process.env.USERNAME_BLACKLIST_URL) {
+    await fetch(process.env.USERNAME_BLACKLIST_URL).then(async (resp) =>
+      (await resp.text()).split("\n").forEach(cachedData.add, cachedData)
```
emrysal, line 41: I contemplated plaintext vs json, but I prefer a plaintext file as that feels more natural; a simple list of usernames without. Needless to say split is also a lil faster than json.parse()

I'm also using then, this prevents having to temporarily assign a huge text file to a variable.

apps/web/pages/api/auth/signup.ts
```diff
@@ -1,218 +1,50 @@
...
-            },
-            accepted: false,
-          },
-          data: {
-            accepted: true,
-          },
-        });
-      }
+export default async function handler(req: RequestWithUsernameStatus, res: NextApiResponse) {
+  // Use a try catch instead of returning res every time
```
keithwillcode, line 28: Suggestion to be considered at a later time: try/catch structures ideally are used to catch unintended errors within the application, not to control the flow of logic. By using the structure this way, we are adding complexity to the catch, which tends to complicate the code. We now intentionally throw an HttpError and we have to clutter the catch with a check of that type.

An alternative to this is to create well-structured validation classes. In said validation class, you would have a function like `validate()` which would run through all of the rules that need to be passed. If validation fails, you return a `ValidationResult` object with more information on why. From there, we update the response and return from the handler before executing other code.

To make it slightly cleaner on the handler side of things (a bit more complex on this dependency injection side), we could actually wrap this handler with what's called a "validation proxy". Basically, it's using the Proxy pattern for validation. You end up building a handler that is injected with a validation proxy and before the true handler is called, the proxy to perform validation is called first. They are not in the same file. This keeps things separate and we avoid polluting true handler code with ugly validation code.

Again, just a suggestion for something to look into for later to organize our validation in a different way.

packages/features/auth/signup/handlers/calcomHandler.ts
```diff
@@ -0,0 +1,193 @@
...
+    });
+    if (team) {
+      const teamMetadata = teamMetadataSchema.parse(team?.metadata);
+
+      const user = await prisma.user.upsert({
+        where: { email },
+        update: {
+          username,
+          password: hashedPassword,
+          emailVerified: new Date(Date.now()),
```
keithwillcode, line 116: This properly uses UTC I assume?

packages/features/auth/signup/handlers/calcomHandler.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (foundToken && foundToken?.teamId) {
+    const team = await prisma.team.findUnique({
+      where: {
+        id: foundToken.teamId,
+      },
+    });
+    if (team) {
+      const teamMetadata = teamMetadataSchema.parse(team?.metadata);
+
+      const user = await prisma.user.upsert({
```
keithwillcode, line 111: This should go in a transaction since we are updating multiple records. If any fails, we'll have data in a bad state.

packages/features/auth/signup/handlers/selfHostedHandler.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (foundToken && foundToken?.teamId) {
+    const team = await prisma.team.findUnique({
+      where: {
+        id: foundToken.teamId,
+      },
+    });
+    if (team) {
+      const teamMetadata = teamMetadataSchema.parse(team?.metadata);
+
+      const user = await prisma.user.upsert({
```
keithwillcode, line 53: We should wrap these updates in a transaction so we don't leave data in a bad state if any of these fails.

packages/lib/slugify.ts
```diff
@@ -2,6 +2,9 @@
 // For eg:- "test-slug" is the slug user wants to set but while typing "test-" would get replace to "test" becauser of replace(/-+$/, "")
 
 export const slugify = (str: string, forDisplayingInput?: boolean) => {
+  if (!str) {
+    return "";
```
keithwillcode, line 6: This feels like it could result in letting bad data through. i.e. Instead of failing, this returns empty and stores empty slugs or something alike. Thoughts?

keithwillcode (commented):
Nice work, @sean-brydon! Just have a few comments.

packages/lib/slugify.ts
sean-brydon: Think this code was outside this PR - i just moved it

packages/features/auth/signup/handlers/selfHostedHandler.ts
sean-brydon: Will make these changes

PeerRich (dismissed):
- [ ] wrong spacing and widths
- [ ] password box is not working correctly (goes green immediately) 
- [x] illustration looks squishy unlike on figma
- [x] wrong line height of subtitle on mobile 
- [ ] missing social proof @ciaranha 
- [x] username and email parameters not working (https://dev-git-feat-signup-refactor-cal-staging.vercel.app/signup?username=peer)
- [ ] doesn't "wow" me. consider animations as discussed before. if we touch a component that is already working, it needs to be 5x to 10x better

PeerRich:
- [ ] wrong spacing and widths
- [ ] missing social proof @ciaranha 
- [ ] username and email parameters not working (https://dev-git-feat-signup-refactor-cal-staging.vercel.app/signup?username=peer)
- [ ] doesn't "wow" me. consider animations as discussed before. if we touch a component that is already working, it needs to be 5x to 10x better

apps/web/pages/signup.tsx
supalarry: > Not related to this PR - will tackle in a follow up.

@sean-brydon I saw you checked Loom I recorded but will post info also here.

The password validation checks work on main locally and also in prod. See Loom [here](https://www.loom.com/share/786f320e03d04aa991b9e4c40e27a9f8). So we need to find what fix is needed in this PR.

sean-brydon: > > Not related to this PR - will tackle in a follow up.
> 
> @sean-brydon I saw you checked Loom I recorded but will post info also here.
> 
> The password validation checks work on main locally and also in prod. See Loom [here](https://www.loom.com/share/786f320e03d04aa991b9e4c40e27a9f8). So we need to find what fix is needed in this PR.

Super weird - i copied and pasted it :P Will see if i can see anything obvious

sean-brydon:
Figma: https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/New-Features?type=design&node-id=5250-329737&mode=design&t=seR4xrWZHw9g0S1Y-4

Added: 

IS_CALCOM 
CleanShot 2023-10-24 at 10 14 56


!IS_CALCOM
CleanShot 2023-10-24 at 10 15 15

supalarry:
Looks amazing, love it!

ciaranha:
@sean-brydon is it possible the password instructions only appear when you're on the password field?

sean-brydon:
> @sean-brydon is it possible the password instructions only appear when you activate the password field?
> 
> CleanShot 2023-10-24 at 10 58 47@2x

As discussed this will be adressed in a follow up as it will effect other places in the app also

packages/lib/server/checkUsername.ts
```diff
@@ -1,6 +1,6 @@
-import { checkPremiumUsername } from "@calcom/features/ee/common/lib/checkPremiumUsername";
-import { IS_SELF_HOSTED } from "@calcom/lib/constants";
+import { IS_CALCOM } from "@calcom/lib/constants";
 
 import { checkRegularUsername } from "./checkRegularUsername";
+import { usernameCheck } from "./username";
 
-export const checkUsername = IS_SELF_HOSTED ? checkRegularUsername : checkPremiumUsername;
+export const checkUsername = !IS_CALCOM ? checkRegularUsername : usernameCheck;
```
ThyMinimalDev, line 6: naming is tough here 😆 , how can we improve this? this is what went in my head when I read this: checkUsername is either checkRegularUsername or usernameCheck, what's the difference ?

ThyMinimalDev (commented):
some global comments but overall seems good 👍

apps/web/pages/signup.tsx
supalarry: @sean-brydon I pushed commit [3e8470b](https://github.com/calcom/cal.com/pull/11421/commits/3e8470b8dd714f98d0f686f7830f01125c23fec7) to this branch that fixes the password hints not being dynamic. In [this commit](https://github.com/calcom/cal.com/pull/11421/commits/dc277a0afa2f70f6fdeb8871657a933032dd1442) `mode: "onChange"` was deleted from the `useForm` hook and it was not re-added later on.

apps/web/pages/api/auth/signup.ts
```diff
@@ -1,218 +1,50 @@
...
-import { validateUsernameInTeam, validateUsername } from "@calcom/lib/validateUsername";
-import prisma from "@calcom/prisma";
-import { IdentityProvider } from "@calcom/prisma/enums";
-import { MembershipRole } from "@calcom/prisma/enums";
-import { signupSchema } from "@calcom/prisma/zod-utils";
-import { teamMetadataSchema } from "@calcom/prisma/zod-utils";
+import { HttpError } from "@calcom/lib/http-error";
 
-export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+function ensureReqIsPost(req: RequestWithUsernameStatus) {
```
Udit-takkar, line 9: we already have `defaultHandler` in @calcom/lib/server which does this same thing

Udit-takkar (commented):
https://github.com/calcom/cal.com/assets/53316345/0e12561d-3d8c-448d-ad77-5fdf518c7d56

I can't sign up with google

Udit-takkar (commented):
Team invite 
1) Open Invite Team Member Modal
2) Click on copy invite link 

we would get url like
http://localhost:3000/teams?token=087fb287d607246acd44f4a88bcd27595221950fe8f353cc5587e4dc572bbcd4

Now open incognito or another browser and open this link. as no user is logged in right now we would be redirected to /login page with callback 

http://localhost:3000/auth/login?callbackUrl=/teams?token=087fb287d607246acd44f4a88bcd27595221950fe8f353cc5587e4dc572bbcd4

but if you click on "Don't have an account" to create an new account and then join the team the token and callbackurl is lost and you are redirected to http://localhost:3000/signup. so after signing up you won't be invited to team. 

IIRC previously we had some logic that doesn't loose this token and whenever user signed up it created team invite.

Udit-takkar (commented):
https://github.com/calcom/cal.com/assets/53316345/f89cffc6-7781-4f64-8fd0-bab2ba0ed233


we don't get any token in the email when inviting user by email in the invite member modal

sean-brydon:
> Team invite
> 
> 1. Open Invite Team Member Modal
> 2. Click on copy invite link
> 
> we would get url like http://localhost:3000/teams?token=087fb287d607246acd44f4a88bcd27595221950fe8f353cc5587e4dc572bbcd4
> 
> Now open incognito or another browser and open this link. as no user is logged in right now we would be redirected to /login page with callback
> 
> http://localhost:3000/auth/login?callbackUrl=/teams?token=087fb287d607246acd44f4a88bcd27595221950fe8f353cc5587e4dc572bbcd4
> 
> but if you click on "Don't have an account" to create an new account and then join the team the token and callbackurl is lost and you are redirected to http://localhost:3000/signup. so after signing up you won't be invited to team.
> 
> IIRC previously we had some logic that doesn't loose this token and whenever user signed up it created team invite.

I think both of these invite issues are not relevant to this PR : https://threads.com/thread/34546974164/34546974165?s=5Nf2N4SCgxDqu9W3xdsnLT

zomars (dismissed):
Blocking to fix some typing errors

apps/web/pages/api/auth/signup.ts
```diff
@@ -1,218 +1,50 @@
...
-              id: team.parentId,
-            },
-            accepted: false,
-          },
-          data: {
-            accepted: true,
-          },
-        });
+  }
+}
```
sean-brydon, line 25: @Udit-takkar  - After battle with TS types for default handler - i had to go back to the previous method 

@zomars im not sure if we can get default handler to work with generics that extends T,J (req,res). Would be ideal for situations like this where RequestWithUsernameStatus is NextReq * {....}

apps/web/playwright/signup.e2e.ts
```diff
@@ -0,0 +1,244 @@
...
+  });
+  test("Email verification sent if enabled", async ({ page, prisma, emails }) => {
+    const emailVerifyFlag = await prisma.feature.findUnique({
+      where: {
+        slug: "email-verification",
+      },
+    });
+
+    // eslint-disable-next-line playwright/no-skipped-test
+    test.skip(!emailVerifyFlag?.enabled);
```
sean-brydon, line 216: If email verification is disabled as a feature flag - dont run this test case

apps/web/pages/auth/sso/[provider].tsx
```diff
@@ -69,7 +70,7 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
 
   if (session) {
     // Validating if username is Premium, while this is true an email its required for stripe user confirmation
-    if (usernameParam && session.user.email) {
+    if (usernameParam && session.user.email && IS_CALCOM) {
```
Comment (sean-brydon) [inline apps/web/pages/auth/sso/[provider].tsx:73]: Since this is now checked in IS_CALCOM and also not calcom envs. we need to only check in calcom envs

emrysal (dismissed):
image

ThyMinimalDev (dismissed):

zomars:
Making some improvements to e2e before merging.

RN if e2e fails you cannot simply re-run since it fails to run the proper cleanups for the created users ("rick", etc.)

zomars:
> Making some improvements to e2e before merging.
> 
> RN if e2e fails you cannot simply re-run since it fails to run the proper cleanups for the created users ("rick", etc.)

Made some fixes. The last test needs to be refactored. It will never be run neither locally nor CI due to the IS_CALCOM condition.

image

apps/web/pages/signup.tsx
```diff
@@ -106,109 +230,262 @@ export default function Signup({ prepopulateFormValues, token, orgSlug, orgAutoA
...
+              className="flex flex-col gap-4"
+              form={formMethods}
+              handleSubmit={async (values) => {
+                let updatedValues = values;
+                if (!formMethods.getValues().username && isOrgInviteByLink && orgAutoAcceptEmail) {
+                  updatedValues = {
+                    ...values,
+                    username: getOrgUsernameFromEmail(values.email, orgAutoAcceptEmail),
+                  };
+                }
```
sean-brydon, line 281: @hariombalhara implemented your changes from #12566 

Please double check this is correct

zomars (approved):
Lfg Lets Fucking Go GIF

PR #11429, opened 2023-09-18
fix: unpublished org/team avatar and i18n

leog:
## What does this PR do?

When an unpublished org or (sub)team gets opened using their URL, the avatar was not shown correctly. Also following crowdin directives to separate i18n strings into their own for orgs/teams.

<kbd>image</kbd>

## Type of change

## How should this be tested?

Create an regular team, an org or org subteam and do not publish them (having env var `IS_TEAM_BILLING_ENABLED` set to true), then see the unpublished screen come up showing the correct avatar of the team/org previously set.


packages/ui/components/unpublished-entity/UnpublishedEntity.tsx
```diff
@@ -13,13 +13,17 @@ export function UnpublishedEntity(props: UnpublishedEntityProps) {
   return (
     <div className="m-8 flex items-center justify-center">
       <EmptyScreen
-        avatar={<Avatar alt={slug ?? ""} imageSrc={`/team/${slug}/avatar.png`} size="lg" />}
+        avatar={
+          <Avatar
+            alt={slug ?? ""}
+            imageSrc={props.orgSlug ? `/org/${slug}/avatar.png` : `/team/${slug}/avatar.png`}
+            size="lg"
```
Udit-takkar, line 20: both org and org subteams have orgSlug but both have different url for image

leog: Should be fixed now, thanks @Udit-takkar

Udit-takkar (approved):
LGTM

PR #11431, opened 2023-09-18
fix: update-stripe-presented-currency

### LINKED ISSUE #11374: [CAL-2471] Ability to change stripe currency to different from default.
This probably need some investigation but we should be able to change currency on stripe payment configuration.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2471](https://linear.app/calcom/issue/CAL-2471/ability-to-change-stripe-currency-to-different-from-default)</sub>

alannnc:
## What does this PR do?

-  Allows for stripe edit currency to be presented in booking page, this also changes the charge currency but owners should only receive payouts in their available account supported currency.

Fixes #11374 

## Requirement/Documentation

## How should this be tested?

- What are the minimal test data to have? Calcom account with installed stripe app and one event type linked.
- What is expected (happy path) to have (input and output)? Allow to edit stripe currency and only fallback to default_currency from account.

## Mandatory Tasks

## Checklist

- ~~I haven't added tests that prove my fix is effective or that my feature works~~


packages/app-store/stripepayment/lib/currencyConvertions.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Docs: https://stripe.com/docs/currencies#zero-decimal
+ * */
+const zeroDecimalCurrencies = [
```
alannnc, line 4: Special cases where we shouldn't convert to * 100.

zomars (dismissed):
Code LGTM. Can we provide testing evidence before merging?

alannnc:
@zomars Simple test added

zomars (approved):

PR #11451, opened 2023-09-19
fix: add 'data'

emrysal:
## What does this PR do?

🤦


packages/app-store/office365calendar/lib/CalendarService.ts
```diff
@@ -260,8 +260,11 @@ export default class Office365CalendarService implements Calendar {
         credential.userId
       );
       const responseJson = await handleErrorsJson(response);
-      const tokenResponse = parseRefreshTokenResponse(responseJson, refreshTokenResponseSchema);
-      o365AuthCredentials = { ...o365AuthCredentials, ...(tokenResponse.success && tokenResponse.data) };
```
joeauyeung, line 257: Notice how other apps handle the safe parse return in the app package

joeauyeung (commented):
In `parseRefreshTokenResponse` if credential sharing is enabled then we safe parse the credential response with the `minimumTokenResponseSchema` if we return the data from it will this cause any errors if we're typing the return to be the token schema in the app package?

packages/app-store/office365calendar/lib/CalendarService.ts
emrysal: parseRefreshTokenResponse has nothing to do with the type; auto-infer does that; so for o365calendar you can do the following to make it work similarly.

```typescript
const tokenResponse: z.infer<typeof refreshTokenResponseSchema> = refreshTokenResponseSchema(responseJson, refreshTokenResponseSchema);
```

PR #11099, opened 2023-09-02
fix: Handle payment flow webhooks in case of event requiring confirmation

hariombalhara:
## What does this PR do?

Fixes #7120

- Adds BOOKING_PAYMENT_INITIATED webhook
- Fixes the issue where, when a paid booking that require confirmation was submitted
	-  BOOKING_REQUESTED webhook didn't fire
	-  Booking Request related emails weren't sent to the organizer and booker.
- Also, correctly takes BOOKING_CREATED/BOOKING_REQUESTED flow for an event that requires confirmation based on duration notice threshold

Also, adds tests in that flow and few other booking flows.
Also, added tests for the stripe webhook flow which complements the booking flow.

Screenshot 2023-09-06 at 4 54 51 PM

There is a small refactor, cleanup and addition of more mocks in **bookingScenario** as well
- Made it easy to write tests by adding more reusable expects
- Mocked stripe payment flow

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?
You will need to setup stripe related env variables
- [x] Test regular event
	- Create an event type that requires payment. Also, enables requires confirmation for the event type.
	- Do a booking for that event type, reach the payment page. Expect an email to booker at the moment asking for completing the payment.
		- Also verify at this moment that BOOKING_PAYMENT_INITIATED webhook is fired. _This is new_
	- Complete the payment. Expect 2 more emails. One to booker that booking has been submited and other to organizer that booking is awaiting approval. _This wasn't working earlier. This has been fixed._
		- Also verify that BOOKING_REQUESTED webhook is fired.
- [x] Test the same for a Team Event - (Tested with collective)
- [x] Test the same for Seated Event
- [x] Test same Managed Event Type
	- Notable Observation: 2 Webhooks fires for each event when both team and user webhook are created.

## Mandatory Tasks
- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #11458, opened 2023-09-20
fix: Handle payment flow webhooks in case of event requiring confirmation

### LINKED ISSUE #7120: [CAL-1071] BOOKING_CREATED event is not triggered in case of "Paid" events and "Requires confirmation" events.
\-- Edit by Hariom
When payment is initiated, BOOKING_PAYMENT_INITIATED(or similar) event should be triggered
When payment is done, BOOKING_REQUESTED event should be triggered. Along with a BOOKING_PAYMENT_SUCCESSFUL event maybe

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1071](https://linear.app/calcom/issue/CAL-1071/booking-created-event-is-not-triggered-in-case-of-paid-events-and)</sub>

### LINKED ISSUE #9767: Webhooks to Zapier are not being triggered

### Issue Summary

I created a test event so that I could set up webhooks with Zapier. I have webhooks set up generally to trigger with booking requests and I have them set up for the test event to use all the triggers. My Zap is really simple. It receives the data from the trigger and shoots me a text message with the information. I've done the ping test for both webhooks and it works. However, when I submit a booking request on both my test event and one of my regular bookings, there is still nothing triggered. 

### Steps to Reproduce

1. Set up a webhook either on an event specifically or through settings
2. Add the Zapier webhook URL and do a ping test
3. After successful ping test, save and make sure webhook is enabled
4. Go to two separate events and create a booking request, and a booking without needing a request. 
5. Monitor Zapier history for any incoming data and refresh after 10 minutes.

I have tested it against different events, with bookings that require confirmation and those that don't, with bookings that require payment and those that don't, with webhooks created from account settings and webhooks created on an event.

EDIT: It seems that why an event requires payment and confirmation is when the trigger isn't firing. I was able to get the following triggers to fire "BOOKING_REJECTED" on the event requiring payment; "BOOKING_REQUESTED" and "BOOKING_REJECTED" on the event not requiring payment. I pulled up the dev tools on the booking requiring payment and I am attaching a screenshot

### Technical details

Chrome Browser (latest stable version), Zapier Pro Plan

- Node.js version
N/A

Screenshot 2023-06-26 at 10 27 27 AM

hariombalhara:
## What does this PR do?

Fixes #7120
Fixes #9767

It takes over the work from #11099 and add more test related fixes in it.

- Adds BOOKING_PAYMENT_INITIATED webhook
- Fixes the issue where, when a paid booking that require confirmation was submitted
	-  BOOKING_REQUESTED webhook didn't fire
	-  Booking Request related emails weren't sent to the organizer and booker.
- Also, correctly takes BOOKING_CREATED/BOOKING_REQUESTED flow for an event that requires confirmation based on duration notice threshold

## Tests
- Migrated to prismock(in memory DB) for **bookingScenario** that makes us avoid doing prisma queries mocking ourselves and thus saving a lot of headache. It is still in memory and thus super fast as earlier code.
	- The changes also makes it easy to replace prismock with some other adapter.
- Adds more tests for fresh booking
- Added a test for rescheduling
- Added tests for the stripe webhook flow which complements the booking flow.

Screenshot 2023-09-06 at 4 54 51 PM

There is a small refactor, cleanup and addition of more mocks in **bookingScenario** as well
- Made it easy to write tests by adding more reusable expects
- Mocked stripe payment flow

## Type of change

## How should this be tested?
You will need to setup stripe related env variables
- [x] Test regular event
	- Create an event type that requires payment. Also, enables requires confirmation for the event type.
	- Do a booking for that event type, reach the payment page. Expect an email to booker at the moment asking for completing the payment.
		- Also verify at this moment that BOOKING_PAYMENT_INITIATED webhook is fired. _This is new_
	- Complete the payment. Expect 2 more emails. One to booker that booking has been submited and other to organizer that booking is awaiting approval. _This wasn't working earlier. This has been fixed._
		- Also verify that BOOKING_REQUESTED webhook is fired.
- [x] Test the same for a Team Event - (Tested with collective)
- [x] Test the same for Seated Event
- [x] Test same Managed Event Type
	- Notable Observation: 2 Webhooks fires for each event when both team and user webhook are created.

## Mandatory Tasks


packages/core/EventManager.ts
```diff
@@ -399,13 +400,21 @@ export default class EventManager {
...
+        })
+      );
+
       /**
        *  Not ideal but, if we don't find a destination calendar,
-       * fallback to the first connected calendar
+       *  fallback to the first connected calendar - Shouldn't be a CRM calendar
        */
-      const [credential] = this.calendarCredentials.filter((cred) => cred.type === "calendar");
+      const [credential] = this.calendarCredentials.filter((cred) => !cred.type.endsWith("other_calendar"));
```
hariombalhara, line 414: It was causing us to default to no calendar actually. cred.type can never be equal to 'calendar' within calendarCredentials as that is prepared by filtering credential having _calendar in type

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1184,15 +1184,14 @@ async function handler(
     teamId,
   };
 
-  const subscribersMeetingEnded = await getWebhooks(subscriberOptionsMeetingEnded);
```
hariombalhara, line 1187: It was unused.

apps/web/test/lib/handleChildrenEventTypes.test.ts
```diff
@@ -1,4 +1,4 @@
-import prismaMock from "../../../../tests/libs/__mocks__/prisma";
+import prismaMock from "../../../../tests/libs/__mocks__/prismaMock";
```
hariombalhara, line 1: Let other users(other than bookingScenario) of prisma mock to keep on using prismaMock

tests/libs/__mocks__/prisma.ts
```diff
@@ -1,18 +1,90 @@
...
 
 vi.mock("@calcom/prisma", () => ({
   default: prisma,
   prisma,
-  availabilityUserSelect: vi.fn(),
-  userSelect: vi.fn(),
+  ...selects,
 }));
 
+const handlePrismockBugs = () => {
```
hariombalhara, line 13: I intend to create bugs in the repo for these. They should not be too hard to fix in there.

CarinaWolli:
@hariombalhara would this solve https://github.com/calcom/cal.com/issues/9767?

hariombalhara:
> @hariombalhara would this solve #9767?

As discussed in DM, it should be solved.

hariombalhara:
Resolved conflicts

packages/lib/payment/handlePaymentSuccess.ts
```diff
@@ -2,114 +2,18 @@ import type { Prisma } from "@prisma/client";
...
-    uid: booking.uid,
-    destinationCalendar: selectedDestinationCalendar ? [selectedDestinationCalendar] : [],
-    recurringEvent: parseRecurringEvent(eventTypeRaw?.recurringEvent),
-    paymentInfo: booking.payment?.[0] && {
-      amount: booking.payment[0].amount,
-      currency: booking.payment[0].currency,
-      paymentOption: booking.payment[0].paymentOption,
-    },
-  };
+  const { booking, user: userWithCredentials, evt, eventType } = await getBooking(bookingId);
```
hariombalhara, line 16: Reuse `getBooking` already in use by handleSetupSuccess

packages/lib/payment/handlePaymentSuccess.ts
```diff
@@ -125,10 +29,16 @@ export async function handlePaymentSuccess(paymentId: number, bookingId: number)
     bookingData.references = { create: scheduleResult.referencesToCreate };
   }
 
-  if (eventTypeRaw?.requiresConfirmation) {
+  const requiresConfirmation = doesBookingRequireConfirmation({
```
hariombalhara, line 32: Correctly determines requiresConfirmation for booking consider there can be threshold case as well.

packages/features/ee/payments/api/webhook.ts
```diff
@@ -5,131 +5,27 @@ import type Stripe from "stripe";
...
-    },
-    select: {
-      recurringEvent: true,
-      requiresConfirmation: true,
-      metadata: true,
-    },
-  });
-}
-
-async function getBooking(bookingId: number) {
```
hariombalhara, line 43: Moved to separate module for reuse

apps/web/test/lib/getSchedule.test.ts
```diff
@@ -1,20 +1,18 @@
 import CalendarManagerMock from "../../../../tests/libs/__mocks__/CalendarManager";
-import prismaMock from "../../../../tests/libs/__mocks__/prisma";
+import prismock from "../../../../tests/libs/__mocks__/prisma";
```
alannnc, line 2: What is the case here? Where below we import from prismaMock, should we be concise and rename this as the others? "prismaMock"

apps/web/test/utils/bookingScenario/MockPaymentService.ts
```diff
@@ -0,0 +1,88 @@
...
+    return "http://mock-payment.example.com/";
+  }
+  const paymentUid = uuidv4();
+  const externalId = uuidv4();
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  class MockPaymentService implements IAbstractPaymentService {
+    // TODO: We shouldn't need to implement adding a row to Payment table but that's a requirement right now.
+    // We should actually delegate table creation the the core app. Here, only the payment app specific logic should come
```
alannnc, line 23:
```suggestion
    // We should actually delegate table creation to the core app. Here, only the payment app specific logic should come
```

apps/web/test/utils/bookingScenario/MockPaymentService.ts
```diff
@@ -0,0 +1,88 @@
...
+    return "http://mock-payment.example.com/";
+  }
+  const paymentUid = uuidv4();
+  const externalId = uuidv4();
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  class MockPaymentService implements IAbstractPaymentService {
+    // TODO: We shouldn't need to implement adding a row to Payment table but that's a requirement right now.
+    // We should actually delegate table creation to the core app. Here, only the payment app specific logic should come
```
alannnc, line 23: Agree with this, we should have a payment core. Or maybe it will be too much abstraction.. :/

packages/features/bookings/lib/handleBookingRequested.ts
```diff
@@ -0,0 +1,69 @@
...
+    } | null;
+    eventTypeId: number | null;
+    userId: number | null;
+    id: number;
+  };
+}) {
+  const { evt, booking } = args;
+
+  await sendOrganizerRequestEmail({ ...evt });
+  await sendAttendeeRequestEmail({ ...evt }, evt.attendees[0]);
```
alannnc, line 35: Can we create something that if anything fails, don't stop anything else? Then we can add failed to retry system.

tests/libs/__mocks__/prisma.ts
```diff
@@ -1,18 +1,90 @@
...
+    // @ts-ignore
+    return __updateBooking(...rest);
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  prismock.webhook.findMany = (...rest: any[]) => {
+    // There is some bug in prismock where it can't handle complex where clauses
+    if (rest[0].where?.OR && rest[0].where.AND) {
+      rest[0].where = undefined;
+      logger.silly("Fixed Prismock bug-2");
```
alannnc, line 43: Should we keep these?

alannnc (approved):
Awesome job here! Everything looks good. Just as little nit I would like to have Failure test scenarios.

tests/libs/__mocks__/prisma.ts
hariombalhara: Yeah, I think so. My reasoning is that there might be bug in handling these scenarios and one should be aware of these. Ideally I would want these to be fixed in Prisma, 2 of them are actually not that hard to fix.

packages/features/bookings/lib/handleBookingRequested.ts
hariombalhara: I am not sure how that would look. Maybe we can do that when we implement retry system?

apps/web/test/utils/bookingScenario/MockPaymentService.ts
hariombalhara: I don't think that would be too much. It would make sense. We should really focus on such abstractions so that App authors need to right less and less code and thus can introduce less bugs.

apps/web/test/lib/getSchedule.test.ts
hariombalhara: If we want an In-memory DB implementation we have prismock otherwise we can use our mocked prisma which doesn't implement any method.

So, I believe we should use prismock for integration tests(like booking scenarios) and we can use prismaMock for unit tests(small functions)

hariombalhara:
Thanks for the review @alannnc 🙏 I have created a ticket for failure tests. I will add them soon.

PR #11463, opened 2023-09-20
chore: reducing team next data

leog:
## What does this PR do?

Org profile page included each member in each subteam to show in which subteam they belong too. Taking into account each subteam member is also a member of the org itself, this PR removes the members object by each children team to rely on the org members.

## Type of change

## How should this be tested?

Creating an org and subteams and inviting people to the subteam will continue to show the correct membership in the org profile for each subteam.

### DIFF
--- a/apps/web/pages/team/[slug].tsx
+++ b/apps/web/pages/team/[slug].tsx
@@ -112,9 +112,8 @@ function TeamPage({ team, isUnpublished, markdownStrippedBio, isValidOrgDomain }
     </ul>
   );
 
-  const SubTeams = () => {
-    console.log({ members: team.members });
-    return team.children.length ? (
+  const SubTeams = () =>
+    team.children.length ? (
       <ul className="divide-subtle border-subtle bg-default !static w-full divide-y rounded-md border">
         {team.children.map((ch, i) => (
           <li key={i} className="hover:bg-muted w-full">
@@ -162,7 +161,6 @@ function TeamPage({ team, isUnpublished, markdownStrippedBio, isValidOrgDomain }
         </div>
       </div>
     );
-  };
 
   return (
     <>
--- a/apps/web/pages/team/[slug].tsx
+++ b/apps/web/pages/team/[slug].tsx
@@ -112,8 +112,9 @@ function TeamPage({ team, isUnpublished, markdownStrippedBio, isValidOrgDomain }
     </ul>
   );
 
-  const SubTeams = () =>
-    team.children.length ? (
+  const SubTeams = () => {
+    console.log({ members: team.members });
+    return team.children.length ? (
       <ul className="divide-subtle border-subtle bg-default !static w-full divide-y rounded-md border">
         {team.children.map((ch, i) => (
           <li key={i} className="hover:bg-muted w-full">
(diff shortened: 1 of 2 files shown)


packages/lib/server/queries/teams/index.ts
```diff
@@ -68,16 +77,6 @@ export async function getTeamWithMembers(args: {
...
-        members: {
-          select: {
-            user: {
-              select: {
-                name: true,
-                username: true,
-              },
-            },
-          },
-        },
```
leog, line 80: No more members by each children team, which resulted in duplicated member info

packages/lib/server/queries/teams/index.ts
```diff
@@ -28,6 +28,15 @@ export async function getTeamWithMembers(args: {
...
     },
+    teams: {
+      select: {
+        team: {
+          select: {
+            slug: true,
+          },
+        },
+      },
+    },
```
leog, line 39: We now need the teams the user belongs to

packages/lib/server/queries/teams/index.ts
```diff
@@ -142,6 +141,9 @@ export async function getTeamWithMembers(args: {
       role: obj.role,
       accepted: obj.accepted,
       disableImpersonation: obj.disableImpersonation,
+      subteams: orgSlug
+        ? obj.user.teams.filter((obj) => obj.team.slug !== orgSlug).map((obj) => obj.team.slug)
+        : null,
```
leog, line 146: Only including subteams slugs on an org context

apps/web/pages/team/[slug].tsx
```diff
@@ -316,7 +319,9 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
       name: member.name,
       id: member.id,
       bio: member.bio,
+      subteams: member.subteams,
       username: member.username,
+      accepted: member.accepted,
```
Comment (leog) [inline apps/web/pages/team/[slug].tsx:324]: We now rely on accepted property to show members in the profile, as the username will be set even for invited people

apps/web/pages/team/[slug].tsx
```diff
@@ -129,7 +129,8 @@ function TeamPage({ team, isUnpublished, markdownStrippedBio, isValidOrgDomain }
                   <span className="text-default text-sm font-bold">{ch.name}</span>
                   <span className="text-subtle block text-xs">
                     {t("number_member", {
-                      count: ch.members.filter((mem) => mem.user.username !== null).length,
+                      count: team.members.filter((mem) => mem.subteams?.includes(ch.slug) && mem.accepted)
+                        .length,
```
Comment (leog) [inline apps/web/pages/team/[slug].tsx:133]: We now rely on subteam slug matching iterated subteam and accepted property to show members in the profile, as the username will be set even for invited people

apps/web/pages/team/[slug].tsx
```diff
@@ -138,9 +139,9 @@ function TeamPage({ team, isUnpublished, markdownStrippedBio, isValidOrgDomain }
                 className="mr-6"
                 size="sm"
                 truncateAfter={4}
-                items={ch.members
-                  .filter((mem) => mem.user.username !== null)
-                  .map(({ user: member }) => ({
+                items={team.members
+                  .filter((mem) => mem.subteams?.includes(ch.slug) && mem.accepted)
```
Comment (leog) [inline apps/web/pages/team/[slug].tsx:143]: We now rely on subteam slug matching iterated subteam and accepted property to show members in the profile, as the username will be set even for invited people

alannnc (approved):
Code looks good, it would be good idea in the future create views for desired values so we can avoid iterating on render.

PR #11464, opened 2023-09-20
feat: Add 300s accelerate cache to feature flags

emrysal:
## What does this PR do?

Adds 300s cache to feature flags, which does mean that feature flag changes take 5m to take effect but also means a lot less database calls.


zomars (approved):

keithwillcode (approved):
Nice

PR #11472, opened 2023-09-21
fix: incorrect client_id parameter assignment in generating access token for webex

Dhoni77:
## Type of change


Udit-takkar (approved):
LGTM. nice catch

PR #11474, opened 2023-09-21
fix: rescheduled value DB update on reschedule and insights view cancelleds

### LINKED ISSUE #10486: Several bugs in insights

### Issue Summary

I found bugs in the insights section. Today, I created two team meetings: one was scheduled, and the other was canceled. However, the insights section is showing that 2 events were created and 0 events were canceled.
Screenshot 2023-07-31 212150

Today is the 31st, but I noticed that the trace graph's end date varies when using different filters.
Screenshot 2023-07-31 211352

Screenshot 2023-07-31 211003

Screenshot 2023-07-31 210644

### Steps to Reproduce

1. Go to insights

alannnc:
## What does this PR do?

- When rescheduling a non-seats booking we flag DB field reschedule = true
- Update BookingTimeStatus view with correct value for cancelled bookings.
- Add many insights test!


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2068,6 +2068,7 @@ async function handler(
           id: originalRescheduledBooking.id,
         },
         data: {
+          rescheduled: true,
```
alannnc, line 2071: We here validate that the event type has no seats enabled.

packages/features/insights/filters/TeamAndSelfList.tsx
```diff
@@ -22,6 +22,11 @@ export const TeamAndSelfList = () => {
   const { data, isSuccess } = trpc.viewer.insights.teamListForUser.useQuery(undefined, {
     // Teams don't change that frequently
     refetchOnWindowFocus: false,
+    trpc: {
+      context: {
+        skipBatch: true,
+      },
+    },
```
alannnc, line 29: Skip batching since now we delay rendering everything waiting for this to end.

packages/prisma/migrations/20230921002822_fix_booking_time_status/migration.sql
```diff
@@ -0,0 +1,35 @@
...
+    "Booking".location,
+    "Booking".paid,
+    "Booking".status,
+    "Booking".rescheduled,
+    "Booking"."userId",
+    "et"."teamId",
+	  "et"."length" as "eventLength",
+      CASE
+          WHEN "Booking".rescheduled IS TRUE THEN 'rescheduled'::text
+          WHEN "Booking".status = 'cancelled'::"BookingStatus" AND "Booking".rescheduled IS NULL THEN 'cancelled'::text
```
alannnc, line 24: The main change here is we were looking for "Booking".rescheduled IS FALSE which was incorrect

alannnc: Views need to be recreated since cannot be updated

apps/web/playwright/insights.e2e.ts
```diff
@@ -0,0 +1,276 @@
...
+      .nth(1)
+      .click();
+    await page.waitForLoadState("networkidle");
+
+    // press escape button to close the filter
+    await page.keyboard.press("Escape");
+
+    await page.getByRole("button", { name: "Clear" }).click();
+
+    expect(page.url()).not.toContain("memberUserId=");
```
Udit-takkar, line 271: @alannnc tests are failing probably because it takes some time for the url to update after clicking "Clear"

packages/prisma/migrations/20230921002822_fix_booking_time_status/migration.sql
zomars: Nice SQL-fu in here

Noice

zomars (changes requested):
Blocking to fix tests first

zomars (approved):
Nice work @alannnc. Does the included test serves the purpose to prevent regression? And if not, can we add it in a follow up?

alannnc:
> Nice work @alannnc. Does the included test serves the purpose to prevent regression? And if not, can we add it in a follow up?

@zomars Tests expect the frontend to work like it's meant to be. If regression happens test will complain now, we didn't have them when I implemented org changes here so we couldn't tell if something broke.

PR #11475, opened 2023-09-21
fix: add a default timezone

Udit-takkar:
Fixes:- https://threads.com/34537862557?s=k5xPhyoGBtgnVwo78z6wgb


Abdel-Monaam-Aouini (approved):

emrysal (dismissed):
When all else fails, looks good to me.

emrysal (approved):

PR #11485, opened 2023-09-22
fix: server side error on /apps

Udit-takkar:
## What does this PR do?

Fixes # (issue)

I was getting this error locally 
Screenshot 2023-09-22 at 12 04 32 PM

AFTER:-

Screenshot 2023-09-22 at 12 08 12 PM

## Type of change

### DIFF
--- a/packages/features/ee/teams/lib/getUserAdminTeams.ts
+++ b/packages/features/ee/teams/lib/getUserAdminTeams.ts
@@ -82,7 +82,7 @@ const getUserAdminTeams = async ({
         name: user.name || "me",
         logo: user?.avatar === "" ? null : user?.avatar,
         isUser: true,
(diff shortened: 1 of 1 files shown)


packages/features/ee/teams/lib/getUserAdminTeams.ts
```diff
@@ -82,7 +82,7 @@ const getUserAdminTeams = async ({
         name: user.name || "me",
         logo: user?.avatar === "" ? null : user?.avatar,
         isUser: true,
-        credentials: includeCredentials ? [] : user.credentials,
+        credentials: includeCredentials ? user.credentials : [],
```
Udit-takkar, line 85: i think this should be the right condition here

hariombalhara (approved):
LGTM!!

PR #11493, opened 2023-09-22
fix: verify-rerender-checkout-stripe-session

alannnc:
## What does this PR do?

- Fixes a rare case when re render page when checkoutStripe props where undefined.
- Changes fetching strategy of trpc only call it when at least one param is available.


apps/web/pages/auth/verify.tsx
```diff
@@ -60,10 +60,16 @@ export default function Verify() {
...
-    checkoutSessionId: sessionId,
-  });
+  const { data } = trpc.viewer.public.stripeCheckoutSession.useQuery(
+    {
+      stripeCustomerId,
+      checkoutSessionId: sessionId,
+    },
+    {
+      enabled: !!stripeCustomerId || !!sessionId,
+      staleTime: Infinity,
```
alannnc, line 70: only call it once, no matter how many times it gets executed on every render.

apps/web/pages/auth/verify.tsx
```diff
@@ -123,12 +129,12 @@ export default function Verify() {
...
               <AlertTriangle className="text-inverted h-12 w-12 flex-shrink-0 p-0.5 font-extralight" />
             ) : sessionId ? (
-              <Check className="text-inverted h-12 w-12 flex-shrink-0 p-0.5 font-extralight" />
+              <Check className="text-inverted h-12 w-12 flex-shrink-0 p-0.5 font-extralight dark:text-white" />
             ) : (
               <MailOpen className="text-inverted h-12 w-12 flex-shrink-0 p-0.5 font-extralight" />
             )}
           </div>
-          <h3 className="font-cal my-6 text-3xl font-normal">
+          <h3 className="font-cal text-inverted my-6 text-3xl font-normal dark:text-white">
```
alannnc, line 137: style fixes for dark mode

apps/web/pages/auth/verify.tsx
```diff
@@ -158,7 +164,7 @@ export default function Verify() {
                 e.preventDefault();
                 setSecondsLeft(30);
                 // Update query params with t:timestamp, shallow: true doesn't re-render the page
-                const _searchParams = new URLSearchParams(searchParams);
+                const _searchParams = new URLSearchParams(searchParams.toString());
```
alannnc, line 167: it was type erroring here but you can to searchParams.toString and the new URLSearchParams should be able to pickup url encoded string.

zomars (approved):
Awesome detective work @alannnc 

LGTM

PR #11494, opened 2023-09-23
fix: profile slug added on create your first event type  dialog

Chiranjeev-droid:
## What does this PR do?
If a new user signs up so probably he has not created any event type or even an existing user also if he has no eventtype setup, then we show a create your first eventtype button at the center. When we click on that , Add a new event type dialog appears which ideally should have slug of the user but it shows empty . This pr removes this bug and shows the slug of the user.

Before making changes
 Loom Video: https://www.loom.com/share/36c09ead110144efa5b255e4fe8ad037?sid=60e2ba78-a34f-4548-9ec2-741dfd7e862d

After making changes :- 
 Loom Video:  https://www.loom.com/share/0a67d182d5a24223a71adc7fa028e5ab?sid=97389a96-dfd8-47a7-9227-894210869a31

## Type of change

## How should this be tested?
- Delete all the previous event types or create a new cal account.
- Click on the button with description create your first event type which is at the centre.
- you will see that slug does not appear in the url. Ideally the slug should appear here that you have setted  as username while creating account.

## Mandatory Tasks

### DIFF
--- a/apps/web/pages/event-types/index.tsx
+++ b/apps/web/pages/event-types/index.tsx
@@ -127,7 +127,7 @@ const MobileTeamsTab: FC<MobileTeamsTabProps> = (props) => {
           readOnly={events[0].metadata.readOnly}
         />
       ) : (
-        <CreateFirstEventTypeView />
+        <CreateFirstEventTypeView slug={eventTypeGroups[0].profile.slug ?? ""} />
       )}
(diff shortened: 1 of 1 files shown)


CarinaWolli (approved):
Looks good! Thank you for this fix :)

PR #11512, opened 2023-09-25
fix: Enhance the Robustness of Embed Snippets for Multiple Embeds on a Single Page

### LINKED ISSUE #10031: [CAL-2128] Embed Code Snippet Generator should use event slug as namespace for API
Using event slug as namespace for the generated code would allow users to have more than 1 embed on a single page without making any changes to the code snippet.

This is a problem faced by a user [as reported in Discord](https://discord.com/channels/1125390093386010654/1126084367119290438/1126122986106265640)

<sub>[CAL-2128](https://linear.app/calcom/issue/CAL-2128/embed-code-snippet-generator-should-use-event-slug-as-namespace-for)</sub>

hariombalhara:
## What does this PR do?

Embed Snippets are now using namespaces so that no change is required when installing multiple embed snippets on a single page.
Because event slug is unique and changes less frequently, it is being used as a namespace.

Fixes #10031

## Type of change

## How should this be tested?
- Create an event-type
- Click Embed
- Select any embed type
- Observe the uses of ns.${eventSlug} everywhere.
- Install the snippet as per instructions and test that they are working.

## Mandatory Tasks

## Checklist


zomars (approved):
Code LGTM. Rerunning checks to see if they pass

PR #11533, opened 2023-09-26
fix: timezone of default schedule should change if profile timezone is updated

### LINKED ISSUE #11452: [CAL-2496] Timezone updates should update availability schedules
**Problem**

Currently any timezone changes don't actually change availabillity schedules

**Solution**

1. Timezone changes also update availability schedules
2. The ability to override on a per schedule basis

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2496](https://linear.app/calcom/issue/CAL-2496/timezone-updates-should-update-availability-schedules)</sub>

CarinaWolli:
## What does this PR do?

What's currently happening? 
- When a schedule does not have timezone set in the db (which is the case if you never update your availability after creating it), then the schedule always has the timezone of the profile. 
- If a timezone is set, and the profile timezone is changed the schedule's timezones are not affected 

How do we want it? (discussed with @ciaranha) 
- When the timezone is changed in the profile, then the default timezone should also be changed to that timezone
- All other schedules should not be affected by the timezone change

Fixes #11452

## Type of change

## How should this be tested?

- Create at least two availbilities 
- Change profile timezone 
- See that the timezone of the default availability changes and other timezones stay 

## Mandatory Tasks

### DIFF
--- a/packages/trpc/server/routers/viewer/availability/schedule/create.handler.ts
+++ b/packages/trpc/server/routers/viewer/availability/schedule/create.handler.ts
@@ -54,6 +54,8 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
     },
   };
 
+  data.timeZone = user.timeZone;
+
   const schedule = await prisma.schedule.create({
     data,
   });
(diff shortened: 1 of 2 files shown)


packages/trpc/server/routers/viewer/availability/schedule/create.handler.ts
```diff
@@ -54,6 +54,8 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
     },
   };
 
+  data.timeZone = user.timeZone;
```
CarinaWolli, line 57: set the timezone already on creation

emrysal (approved):

Issue #11437, opened 2023-09-19
New i18n implementation due to the app-router migration


grzpab:
### Is your proposal related to a problem?

When cal.com moves to Next.js' App Router API, i18n will no longer be "natively" supported by that API, namely `.locale`, `.locales`, `.defaultLocale` and `.domainLocales` properties of the `router` object will not be accessible.

### Describe the solution you'd like
I thought of an architectural change that moves the resolving of translation functions from client-side components to server-side components.

This could be achieved by the following principles:

1. for any session, the language of the page will be decided based on:
    1. the language in the user settings, if the user is logged in; we can get that information using the session cookie,
    2. the accept-language header otherwise,
2. the language translations will be resolved only in the server-side code, based on the session language; the client-side code will be agnostic of the concept of translation.

The technical plan consists of the following steps:

1. the installation of the `i18next` module https://github.com/i18next/i18next, which contrary to the name, is agnostic of Next.js (the parent library of the `next-i18next` that [Cal.com](http://cal.com/) have been using),
2. loading of all supported languages during the booting of the Next.js server using https://www.i18next.com/overview/api#getfixedt, so all requests have access to these languages as global map of language/region codes mapped to the corresponding language objects (the exact way is to be decided, but the requirement is that this map is created once, is immutable and is available to every request),
3. every page uses a helper function that obtains the language/region pair from the request and matches it with the appropriate language object in the global language map. The selection is based on:
    1. the user language settings inferred from the session cookie in the request,
    2. the header called `accept-language`,
4. the language is then accessible to every server-side rendered component in a particular page,
5. the `react-i18next`, `next-i18next` libraries and associated middlewares and hooks will have to be cleaned up, each `t` and `Trans` usage will have to be changed into a usage of the language object mentioned beforehand (this will be codemodable):
    1. around 360 usages of `t` from `useLocale` from `@calcom/lib/hooks/useLocale` (this is the [Cal.com](http://cal.com/) hook),
    2. around 30 usages of `<Trans>` from `next-i18next` ,
    3. around 10 usages of `t` from `useTranslation` from `react-i18next` (in `AppConfiguration`).
6. trivia: there’s probably a significant number (possibly 1000+) of unused translations, e.g. `organization_banner_title` (removal can be codemoded as well).

### Describe alternatives you've considered
In theory, the translation could stay in the client-side components, but it has the following drawbacks:
[...]

PR #11534, opened 2023-09-26
refactor: implementation of locale calculated server-side

grzpab:
## What does this PR do?

This issue moves the calculation of the current user's locale into the backend and ensures there is just one way of calculating it.
The algorithm is as follows:
1. if the user token is present in the cookies, extract the locale by decrypting the JWT/JWE token,
2. if the user token is present in the authorization header, extract the locale by decrypting the JWT/JWE token,
3. if the user token is not present, extract the locale from the `accept-language` header,
4. otherwise the locale is `en`,

The application does not longer rely on `router.locale` or similar, instead we pass `newLocale` calculated in `_document` and made available to the frontend code by embedding it as `window.calNewLocale`.
We obtain the `newLocale` again in `getInitialProps` within `_app` using:
1. `window.calNewLocale` for the frontend
2. the request object available in the backend
and we pass it to the components using the PageWrapper / AppProviders utilities.

We removed the `I18nLanguageHandler`.

When you change the language in the settings, you will be logged out. The reason for this behavior is that without getting a new token, we would still read the old language from the current token. We are working on redeeming the token when changing the language so the UX is better.

Fixes # (issue)
No fixes.

## Requirement/Documentation
https://github.com/calcom/cal.com/issues/11437

## Type of change

## How should this be tested?
This change requires a whole test of the application, especially ensuring that:
* the locale is always available and correct translations are visible at all times
* the locale stays the same when moving between different pages
* there are no changes between different locale during the page loading (which happens currently if there are discrepancies between the server-side locale and client-side locale)

## Mandatory Tasks
I did review the current PR.

## Checklist


alex-from-intuita:
@grzpab

packages/features/auth/lib/getServerSession.ts
```diff
@@ -13,17 +13,6 @@ import prisma from "@calcom/prisma";
...
- * This is a slimmed down version of the `getServerSession` function from
- * `next-auth`.
- *
- * Instead of requiring the entire options object for NextAuth, we create
- * a compatible session using information from the incoming token.
- *
- * The downside to this is that we won't refresh sessions if the users
- * token has expired (30 days). This should be fine as we call `/auth/session`
- * frequently enough on the client-side to keep the session alive.
- */
```
zomars, line 26: Is this intended?

packages/trpc/server/routers/publicViewer/i18n.handler.ts
```diff
@@ -11,7 +11,7 @@ type I18nOptions = {
   input: I18nInputSchema;
 };
 
-export const i18nHandler = async ({ input }: I18nOptions) => {
+export const i18nHandler = async ({ ctx, input }: I18nOptions) => {
```
zomars, line 14: Is ctx being used here?

zomars (commented):
Looking good so far. Just noticing some typing errors. Will re-review once those are cleared out. 🙏🏽

packages/features/auth/lib/getServerSession.ts
grzpab: You are right, I restored the comment.

packages/trpc/server/routers/publicViewer/i18n.handler.ts
grzpab: It is not, I removed the usage of it and removed it from the `I18nOptions` type.

yarn.lock
zomars, line 1: Can we remove this change from the PR? Everything else is fine and I can see checks are passing now. 🙏🏽

grzpab: @zomars done, thanks for reviewing.

zomars (approved):

zomars:
Heads up @grzpab this PR is causing spacing issues on main since we're not setting the html `dir="ltr"` attribute anymore. We're working on a fix.

PR #11536, opened 2023-09-26
fix: Unresponsive Button

### LINKED ISSUE #11526: Button unresponsive

### Issue Summary

Button not navigating to next page when clicked for the first time.

### Steps to Reproduce

1. Go to the home(app.cal.com/getting-started) page.
2.Click on the "Next Step" button.
3. Observe that the page does not navigate to the next page.

### Actual Results

-The page doesn't navigate to next page

### Expected Results

-I expect the page should navigate to the next page when clicked for the first time

https://github.com/calcom/cal.com/assets/118684885/890e0db2-ae77-4013-a6cc-0efd34244ba9

suyash5053:
Fixes #11526

## Type of change

@sean-brydon Take a look, if everything works fine according to you.


apps/web/components/getting-started/components/ButtonLogic.tsx
```diff
@@ -0,0 +1,14 @@
+import { showToast } from "@calcom/ui";
+
+export const onButtonClick = async (nextStep: () => void, setButtonDisabled: (disabled: boolean) => void) => {
```
sean-brydon, line 3:
```suggestion
export const onNextstepLogic = async (nextStep: () => void, setButtonDisabled: (disabled: boolean) => void) => {
```

Maybe we can rename this so its clearer what it does

sean-brydon (changes requested):
Please make changes i have mentioned.

I cant see anywhere we are using the disabled value that onButtonClick is meant to be changing?

yarn.lock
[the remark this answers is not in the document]
suyash5053: Ohk

apps/web/components/getting-started/components/ButtonLogic.tsx
suyash5053: Sure it will be good for future references.

apps/web/components/getting-started/components/ButtonLogic.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+  try {
+    await nextStep();
+  } catch (e) {
+    showToast("An error occurred", "error");
+    setTimeout(() => {
+      setButtonDisabled(false);
+    }, 2000);
+  }
+};
```
PeerRich, line 14: this feels weird

suyash5053: To ensure that the button remains disabled after being clicked and is re-enabled only in the event of an error, we will implement a 2-second timeout (timeout can be increased), before allowing the button to be clicked again. This timeout period serves as a preventive measure to mitigate any rapid successive clicks that could potentially exacerbate the error or disrupt the intended functionality. By introducing this brief delay, we aim to maintain a smooth and controlled user experience while handling errors gracefully.

PeerRich:
i dont think this is a good approach of solving this

PR #11543, opened 2023-09-27
chore: Improved logging booking flow

hariombalhara:
## What does this PR do?

Adds detailed logging for booking flow that can be enabled for an eventType or user or team


packages/core/getBusyTimes.ts
```diff
@@ -89,56 +89,54 @@ export async function getBusyTimes(params: {
...
-            },
-            // The current user has a different booking at this time he/she attends
-            {
-              ...sharedQuery,
-              attendees: {
-                some: {
-                  email: userEmail,
-                },
-              },
+  const matchedBookings = await prisma.booking.findMany({
```
hariombalhara, line 92: Storing it in `matchedBookings` now

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -391,13 +392,16 @@ async function ensureAvailableUsers(
       }
     );
 
+    logger.silly(
```
hariombalhara, line 395: Moved above the early return so it's logged in that case as well.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2497,3 +2516,31 @@ const findBookingQuery = async (bookingId: number) => {
   // Don't leak any sensitive data
   return foundBooking;
 };
+
+function isLoggingEnabled({
```
hariombalhara, line 2520: Logging can be enabled either using `BOOKING_LOGGING_EVENT_IDS` by specifying comma separated list of event ids or by `BOOKING_LOGGING_USER_OR_TEAM_NAME` by specifying comma separated list of username or teamname

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2497,3 +2516,31 @@ const findBookingQuery = async (bookingId: number) => {
...
+  eventTypeId,
+  usernameOrTeamName,
+}: {
+  eventTypeId?: number | null;
+  usernameOrTeamName?: string | string[];
+}) {
+  const usernameOrTeamnamesList =
+    usernameOrTeamName instanceof Array ? usernameOrTeamName : [usernameOrTeamName];
+  // eslint-disable-next-line turbo/no-undeclared-env-vars
+  const bookingLoggingEventIds = process.env.BOOKING_LOGGING_EVENT_IDS || "";
```
hariombalhara, line 2530: Will add it to .env.example and turbo.json once we finalize this is what we need

PeerRich (dismissed):

packages/core/getBusyTimes.ts
keithwillcode: This is now always running the query which we don't want to do. This is what caused major slow downs for round-robin events and why we allowed the bookings to be passed into this function from the top-level `getSchedule` layer. This needs to be reverted.

keithwillcode (changes requested):
Nice work, @hariombalhara. There's just 1 change in here we need to revert because of major performance problems.

packages/core/getBusyTimes.ts
hariombalhara: Amazing spot @keithwillcode 🙏 Completely missed it.

keithwillcode (approved):

PR #11595, opened 2023-09-28

zomars:
Reverts calcom/cal.com#11594

Tagging @ciaranha @emrysal @keithwillcode to discuss the desired behavior to the case of 2 unconfirmed bookings fighting for the same slots.

Also this PR was blocking unconfirmed booking whenever the user had another event in their calendar. As it checks for actual user availability.

Also, if you get an unconfirmed bookings. The you add an event in that time to your calendar. Then you try to confirm. You won't be able to confirm it as well.

IMO we should either allow to confirm (allow double bookings) but with the user knowledge. Or give options to reschedule or "Ask for reschedule"


vijayraghav-io:
Hi @zomars , if multiple confirmation has to be checked within same event (or eventTypeId), the implementation could be quite simplified. We need to just check in booking table, if a booking with status-"Accepted" exists for input eventTypeId, startTime, endTime.

Please refer this #11613 .
Tested with same previously added testcase, its working fine.
I am working on adding new test case - for the above explained scenario by you, where user can confirm if booking request is for same time slot but a different event.

Kindly consider this as my proposed implementation and share your comments.


Also, in line with your suggestion we can introduce a new flag (toggle in ui) "allowMultiConfirm" on EventType. This would be active only if user chooses "Requires Confirmation". Thereby we are allowing user to choose if this check is required. Accordingly the if condition in "confirm.handler.ts" can be updated. 
Please confirm if this change can be introduced.

zomars:
Closing as not planned for now

PR #11605, opened 2023-09-29
feat: `Connect` component for Cal Atoms

Ryukemeister:
## What does this PR do?

Adds a `Connect` button component for Cal Atoms


packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+        if (err?.error.message) {
+          setErrMsg(err.error.message);
+        }
+      },
+    });
+  }
+
+  return (
+    <div>
+      <button className="" onClick={(event) => handleSubmit(event)} disabled={!isProcessing} type="button">
```
PeerRich, line 34: why empty className? do we have plans for a styling props API?

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+        {buttonText || "Install App"}
```
PeerRich, line 35: do we wanna do i18n for Cal Atoms?

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: I'm yet to add the default styling but I think if we use the `classNames` function from `@calcom/lib` that should work for any custom styles

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: Might have to ask @sean-brydon if handling translations inside our component is better or if we handle those outside? Imo we can pass in a translated string directly as a prop, eg `buttonText={t("Log in")}` and that should work just fine.

sean-brydon: IMHO button should be able to accept any children -> allowing us to customise it from where it is used. 

Translations inside the connectbutton is a bit weird - im not 100% sure how this would work unless we bundle translations with it

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return (
+    <div>
+      <button
+        className={classNames(
+          stylesClassname
+            ? `bg-default dark:bg-muted rounded-md ${stylesClassname}`
+            : `bg-default dark:bg-muted text-default dark:text-muted rounded-md`
+        )}
```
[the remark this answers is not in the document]
sean-brydon: Let not concat strings with classNames - if we let dynamic classNames be one of the last Items in `classNames` twMerge will correctly merge as expected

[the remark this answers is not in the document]
sean-brydon: Also - is the correct styling? Our normal button has way more features that surley this should support too? 

@PeerRich  @keithwillcode any spec on this?

sean-brydon (changes requested):
https://play.tailwindcss.com/iAVzm47vVY

This is what the button looks like. 

I think this needs some work before it is ready for review.

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+          setErrMsg(err.error.message);
+        }
+      },
+    });
+  }
+
+  return (
+    <div>
+      <Button
+        className={classNames(
```
ThyMinimalDev, line 40: Hey there, I think for cal atoms we should try to follow a similar approach as shown in shadcn/ui

meaning using the `cn` util and class variance authority to handle different variants of the component

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,55 @@
+import { useState } from "react";
+
+import { classNames } from "@calcom/lib";
+
+import { Button } from "../src/components/ui/button";
+import type { AtomsGlobalConfigProps } from "../types";
+import type { ConnectButtonProps } from "../types";
```
ThyMinimalDev, line 7: I think we should consider colocating the type of the button props with the component itself, either in the same file or in the same folder in a different file

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+// buttonText is for displaying buttonText
+// onButtonClick is a a method to handle onClick for button
+// it accepts another object which contains a callback in case of any error
+
+export function ConnectButton({
+  buttonText,
+  onButtonClick,
+  stylesClassname,
+  icon,
+}: ConnectButtonProps & AtomsGlobalConfigProps) {
```
ThyMinimalDev, line 19: here I think we should consider extending the props of a basic HTML button like in shadcn/ui component, 

this is kind of related to the Liskov substitution principle, meaning that since we have providing a wrapper around the basic HTML button, we should provide a similar interface 

this is how it's done in shadcn/ui ` React.forwardRef<HTMLButtonElement, ButtonProps>`

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [isProcessing, setIsProcessing] = useState<boolean>(false);
+  const [errMsg, setErrMsg] = useState<string>("");
+
+  function handleSubmit(e: React.MouseEvent<HTMLButtonElement, MouseEvent>) {
+    e.preventDefault();
+
+    setIsProcessing(true);
+    onButtonClick({
+      cb: (err) => {
```
ThyMinimalDev, line 28: I think this approach is too complicated, 

we should aim to provide simple interface to the atoms and one that people are used to, meaning props such as onChange, onSuccess, onError, onClick...

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+      },
+    });
+  }
+
+  return (
+    <div>
+      <Button
+        className={classNames(
+          "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed",
+          stylesClassname
```
ThyMinimalDev, line 42: I think this props should be just called className, and this is where the `cn` util would come into play and merge our classes with the provided classes

packages/atoms/connect/ConnectButton.tsx
ThyMinimalDev: for i18n here we would need to only bundle the translations related to an atom inside the atom to avoid providing unnecessary translations to users who only use few atoms. 

regarding accepting a ReactNode as a children is a good option to allow more control for the devs using our components, I think this should be on top of using basic props like label, errorMsg,... for less advanced usage

packages/atoms/connect/ConnectButton.tsx
sean-brydon: agreed - if we extend the HtmlButtonElement we can directly use the classname that gets put in.

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: @PeerRich @sean-brydon what do you think about this. Should we just do translations now or leave it for later?

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: And then we can pass in additional props inside of the button like this `<Button ...otherProps>` right?

ThyMinimalDev: yes exactly, like it is done inside shadcn/ui button component

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  return (
+    <div>
+      <Button
+        className={cn(
+          "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed",
+          className
+            ? className
+            : "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed"
```
ThyMinimalDev, line 63: no need to repeat the classnames here, `cn` util will merge everything, meaning you can just write 

```cn(
          "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed",
          className)```

File: packages/atoms/connect/ConnectButton.tsx [hunk 10]
```diff
@@ -0,0 +1,74 @@
...
+        )}
+        type="button"
+        disabled={!isProcessing}
+        onClick={(event) => handleSubmit(event)}>
+        {!!icon && icon}
```
ThyMinimalDev, line 68: prefer using `Boolean` util instead of `!!` in most cases

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onSuccess,
+  onError,
+  className,
+  icon,
+}: ConnectButtonProps & AtomsGlobalConfigProps) {
+  const [isProcessing, setIsProcessing] = useState<boolean>(false);
+  const [isSuccess, setIsSuccess] = useState<boolean>(false);
+  const [errMsg, setErrMsg] = useState<string>("");
+
+  function handleSubmit(e: React.MouseEvent<HTMLButtonElement, MouseEvent>) {
```
ThyMinimalDev, line 33: look at this again, I'm not sure we need handleSubmit at all here, not sure how we plan to use this component but for now we should just use onClick passed with button html props already

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+      setErrMsg(error?.message);
+    }
+
+    setIsProcessing(false);
+
+    // if user wants to handle onSuccess inside handleClick then it makes no sense to have a separate handler
+    // otherwise only if the user explicitly passes an onSuccess handler this gets triggered
+    if (isSuccess && onSuccess) {
+      onSuccess();
+    }
```
ThyMinimalDev, line 53: using state here is not necessary in this case, just call onSuccess at the end of the try statement

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+// we accept two props: buttonText & onButtonClick
+// buttonText is for displaying buttonText
+// onButtonClick is a a method to handle onClick for button
+// it accepts another object which contains a callback in case of any error
+
+interface ConnectButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  buttonText?: string;
+  icon?: JSX.Element;
+  handleClick: () => void;
```
ThyMinimalDev, line 16: onClick instead of handleClick, this props is already typed with HtmlButtonElement

packages/atoms/connect/ConnectButton.tsx
ThyMinimalDev: and then handle different variants of the button when the user passes props like disabled, loading, errMsg, children...

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: But  doesn't `!!` convert the value into it's boolean equivalent?

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -13,44 +13,40 @@ import type { AtomsGlobalConfigProps } from "../types";
...
 
     try {
       setIsProcessing(true);
-      setIsSuccess(true);
-      handleClick();
+      onClick && onClick(e);
+
+      // if user wants to handle onSuccess inside handleClick then it makes no sense to have a separate handler
+      // otherwise only if the user explicitly passes an onSuccess handler this gets triggered
+      if (onSuccess) {
```
Ryukemeister, line 40: Since onSuccess can be handled inside onClick if a user decides to do so that's why the if statement to check if the user passes a value for onSucess

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -13,44 +13,40 @@ import type { AtomsGlobalConfigProps } from "../types";
...
   const [errMsg, setErrMsg] = useState<string>("");
 
   function handleSubmit(e: React.MouseEvent<HTMLButtonElement, MouseEvent>) {
     e.preventDefault();
 
     try {
       setIsProcessing(true);
-      setIsSuccess(true);
-      handleClick();
+      onClick && onClick(e);
```
Ryukemeister, line 36: onClick can be possibly undefined so checking first to see if there's a method being passed

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -59,13 +55,11 @@ export function ConnectButton({
...
           "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed",
           className
-            ? className
-            : "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed"
         )}
         type="button"
         disabled={!isProcessing}
         onClick={(event) => handleSubmit(event)}>
-        {!!icon && icon}
+        {icon && icon}
```
Ryukemeister, line 62: @ThyMinimalDev would something like this work for icon?

packages/atoms/connect/ConnectButton.tsx
sean-brydon: Opinionated - Doesnt matter 
CleanShot 2023-10-13 at 12 00 53@2x

`{icon && icon}` also has the same effect as `{!!icon&&icon}`

emrysal: !! or Boolean doesn't matter, I'd advocate for !! actually as it is much simpler to error and write `new Boolean()` (which is objectively wrong) than `new !!`. 

Anyway we don't enforce little code style issues like this; that's up to the developer.

ThyMinimalDev: indeed it it's very up to the developer and the team, I'm fine with both since the result is the same, it's something I used to enforce in the past in less experience team in react using non boolean to test whether a component should be rendered or not.

anyway my bad I should have precised it was just a suggestion and not a required change

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { useState } from "react";
+import * as React from "react";
+
+import { Button } from "../src/components/ui/button";
+import type { AtomsGlobalConfigProps } from "../types";
+
+// we accept two props: buttonText & onButtonClick
+// buttonText is for displaying buttonText
+// onButtonClick is a a method to handle onClick for button
+// it accepts another object which contains a callback in case of any error
```
supalarry, line 11: 1. I don't see `onButtonClick` prop in `ConnectButtonProps`

2. For all atoms it's simpler to just always accept `onClick` handler instead of `onXClick`, but as Morgan suggested (great suggestion btw) we can just extend `React.ButtonHTMLAttributes<HTMLButtonElement>` which already includes `onClick` there. Here is the shadcn Button example: https://github.com/shadcn-ui/taxonomy/blob/main/components/ui/button.tsx

3. I would also remove all of the comments from line 8 - 11, because the prop names themselves should explain what is going on and comments in this case are redundant - they just restate what developers can see in code

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+          "bg-default text-default dark:text-muted dark:bg-muted relative inline-flex h-9 items-center whitespace-nowrap rounded-md px-4 py-2.5 text-sm font-medium !shadow-none transition-colors disabled:cursor-not-allowed",
+          className
+        )}
+        type="button"
+        disabled={!isProcessing}
+        onClick={(event) => handleSubmit(event)}>
+        {icon && icon}
+        {buttonText || "Install App"}
+      </Button>
+      {!!errMsg && <span>{errMsg}</span>}
```
supalarry, line 65: Cal handbook mentions to use ternaries over &&: https://handbook.cal.com/engineering/best-practices/prefer-ternaries-over-short-circuiting-and-and-for-react-components

Here is article: https://www.crocoder.dev/blog/react-conditional-rendering/

packages/atoms/src/lib/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { type ClassValue, clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+export function cn(...inputs: ClassValue[]) {
```
supalarry, line 4: Personally, i would call this function `classNames` as it's more descriptive, but others might prefer the shorthand 🤷

supalarry: Also, can we rename `lib` folder to just `utils` and then name the file `classNames.ts` or something?

If we have `lib/utils.ts` file then it will become a place where people just dump all utils together and after a while it will be hard to understand what kind of utils exactly are there.

supalarry (commented):
Well done Rajiv! Left some refactor comments.

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: Sorry about the comments, I kinda wrote them just at the start of the component for my own thought process and forgot to update them as I made changes, will remove them altogether. Nice catch for the `onClick` prop, I actually forgot to remove that from the props.

packages/atoms/connect/ConnectButton.tsx
Ryukemeister: @supalarry  I was trying to convert them into a boolean value at first using the `!!` operator just like on line 65, should I go back to that? It's also a possible way according to this article https://www.crocoder.dev/blog/react-conditional-rendering/

packages/atoms/src/lib/utils.ts
Ryukemeister: The classnames and folders are being generated hence I thought not to rename or tweak anything since it'll be uniform that way  for all the components now and also in the future

packages/atoms/connect/ConnectButton.tsx
supalarry: You are right, `!!` is fine

packages/atoms/src/lib/utils.ts
supalarry: Gotcha, what code generates them?

Ryukemeister: `npx shadcn-ui@latest add <component_name>` generated an src folder in the root https://github.com/calcom/cal.com/tree/connect-component/packages/atoms/src which contains the lib and components folder for all the shadcn components

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -5,21 +5,17 @@ import * as React from "react";
...
 
-// we accept two props: buttonText & onButtonClick
-// buttonText is for displaying buttonText
-// onButtonClick is a a method to handle onClick for button
-// it accepts another object which contains a callback in case of any error
-
 interface ConnectButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   buttonText?: string;
   icon?: JSX.Element;
+  handleClick: () => void;
```
supalarry, line 11: Why not call it `onClick`?

supalarry: discussed in threads DM

packages/atoms/connect/ConnectButton.tsx
```diff
@@ -8,14 +8,13 @@ import type { AtomsGlobalConfigProps } from "../types";
...
   icon?: JSX.Element;
-  handleClick: () => void;
   onSuccess?: () => void;
   onError: () => void;
 }
 
 export function ConnectButton({
   buttonText,
-  handleClick,
+  onClick,
```
Ryukemeister, line 17: @supalarry this is the correct way to do it right?

supalarry: Yeah i think this makes sense. Can you manually test it by using the component, passing onClick to it and seeing if it is fired when the button is clicked?

Ryukemeister: Should work IMHO but will test and let you know!

Ryukemeister: @supalarry I added a loader for the button can you review it again and make sure it's alright?

supalarry: Looks good!

PeerRich:
just to confirm, this has no functionality yet right? Its supposed to connect to Google Calendar, zoom, etc.

Ryukemeister:
> just to confirm, this has no functionality yet right? Its supposed to connect to Google Calendar, zoom, etc.

Correct, it doesn't handle any  functionality. The functionality can be passed in through the `onClick` handler and all button does is execute the method (connect to Google Calendar, zoom, etc) that will be passed onto it in the form of props

supalarry (changes requested):

PeerRich:
> Correct, it doesn't handle any functionality. The functionality can be passed in through the onClick handler and all button does is execute the method (connect to Google Calendar, zoom, etc) that will be passed onto it in the form of props.

ideally the Connect button has logic, especially the logic to connect to something. that's the whole point of this component.

`<Connect to="Google Calendar"/>` 

(not final props API) but somewhat like this

Ryukemeister:
> > Correct, it doesn't handle any functionality. The functionality can be passed in through the onClick handler and all button does is execute the method (connect to Google Calendar, zoom, etc) that will be passed onto it in the form of props.
> 
> ideally the Connect button has logic, especially the logic to connect to something. that's the whole point of this component.
> 
> `<Connect to="Google Calendar"/>`
> 
> (not final props API) but somewhat like this

Will keep this in mind

PR #11611, opened 2023-09-29
fix: availability for org team event 

CarinaWolli:
## What does this PR do?

There was a bug that sometimes a org team event didn't use the correct availability. This happened when there existed a non org event with the same event type slug. Then the availability of this event type was used instead. 

## Type of change

## How should this be tested?

- Create an org 
- Create a team event with the slug 'test'
- Create another account with a normal team 
- Create a team event with the slug 'test'
- Before the fix, the org team event used the availability of the wrong event type => check that both event types show the correct slots

### DIFF
--- a/packages/features/bookings/Booker/utils/event.ts
+++ b/packages/features/bookings/Booker/utils/event.ts
@@ -66,6 +66,8 @@ export const useScheduleForEvent = ({
 
   const pathname = usePathname();
 
+  const isTeam = !!event.data?.team?.parentId;
+
   return useSchedule({
     username: usernameFromStore ?? username,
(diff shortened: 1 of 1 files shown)


keithwillcode (approved):
Sneaky. 

We need to add a test around this (ideally unit test) to make sure it doesn’t regress. 

If complex, a follow up PR is ok.

leog (approved):
LGTM, thanks for fixing this.

CarinaWolli:
I will follow up with a test for it, not yet sure how to best test that. 

Issue: https://github.com/calcom/cal.com/issues/11617

PR #11613, opened 2023-09-29

vijayraghav-io:
refs #11595

### DIFF
--- a/packages/trpc/server/routers/viewer/bookings/confirm.handler.ts
+++ b/packages/trpc/server/routers/viewer/bookings/confirm.handler.ts
@@ -7,7 +7,6 @@ import { handleConfirmation } from "@calcom/features/bookings/lib/handleConfirma
 import { handleWebhookTrigger } from "@calcom/features/bookings/lib/handleWebhookTrigger";
 import type { EventTypeInfo } from "@calcom/features/webhooks/lib/sendPayload";
 import { isPrismaObjOrUndefined, parseRecurringEvent } from "@calcom/lib";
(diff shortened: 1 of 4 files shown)


packages/trpc/server/routers/viewer/bookings/confirm.handler.ts
```diff
@@ -116,23 +114,25 @@ export const confirmHandler = async ({ ctx, input }: ConfirmOptions) => {
     throw new TRPCError({ code: "BAD_REQUEST", message: "Booking already confirmed" });
   }
 
-  // Check for user's slot availability for current booking start and end time
+  // Check if user has already accepted/confirmed previously a booking request
+  // for the same event and same time slot
   if (confirmed) {
-    const slotsAvailable = await getAvailableSlots({
-      input: {
+    const acceptedBooking = await prisma.booking.findFirst({
```
moilejter, line 120: Don't you want to check _both_ of these conditions?  (if there is no accepted booking, we would still want to check if the slot is availalble -it could have been taken off the board due to changes in availability, right?)

vijayraghav-io: @moilejter , getAvailableSlots() checks the availability across event of a user. 
As per the latest requirement we have to avoid double confirmation of multiple requests for same time slot within a single event only (and not across different events).

And the latest user's availability is taken into consideration while selecting the slot itself.

zomars:
Hey @vijayraghav-io thank you for the patience and effort. Discussed this with the team and we decided that we're not going to pursue this further. There are some valid cases when one may allow conflicting bookings that are opt-in.

PR #11618, opened 2023-09-29
fix: profile loading fix in org

Udit-takkar:
## What does this PR do?

Fixes https://threads.com/34539679188?s=po6sYHCBaUoV1MzZqh55t1

Screenshot 2023-09-29 at 9 50 49 PM

E2E Test:-

Screenshot 2023-09-29 at 11 08 18 PM

### DIFF
--- a/apps/web/pages/settings/my-account/profile.tsx
+++ b/apps/web/pages/settings/my-account/profile.tsx
@@ -89,12 +89,15 @@ const ProfileView = () => {
   const [fetchedImgSrc, setFetchedImgSrc] = useState<string | undefined>(undefined);
 
   const { data: user, isLoading } = trpc.viewer.me.useQuery(undefined, {
-    onSuccess: (userData) => {
-      if (!userData.organization) {
-        fetch(userData.avatar).then((res) => {
+    onSuccess: async (userData) => {
+      try {
+        if (!userData.organization) {
+          const res = await fetch(userData.avatar);
           if (res.url) setFetchedImgSrc(res.url);
-        });
-      } else {
+        } else {
+          setFetchedImgSrc("");
+        }
+      } catch (err) {
         setFetchedImgSrc("");
       }
     },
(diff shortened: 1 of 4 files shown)


apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -90,9 +90,13 @@ const ProfileView = () => {
...
-      fetch(userData.avatar).then((res) => {
-        if (res.url) setFetchedImgSrc(res.url);
-      });
+      if (!userData.organization) {
+        fetch(userData.avatar).then((res) => {
+          if (res.url) setFetchedImgSrc(res.url);
+        });
+      } else {
+        setFetchedImgSrc("ORG_PROFILE_IMAGE");
+      }
```
Udit-takkar, line 99: don't need this for org

Udit-takkar: fetchedImgSrc was used to check if user has any set any image or not. url ending with /avatar.svg is allback image. (earlier it was gravatar url )

leog: NIT: feels weird to set a state to an unused string value, isn't there any other way? 🤔

leog (approved):
LGTM, left a NIT

Udit-takkar:
@leog i added a any random string to prevent loading state.  fetchedImgSrc is not used anywhere so we can just set it any value

https://github.com/calcom/cal.com/blob/60578a2a33d683dd1996840296fdb713eb729b88/apps/web/pages/settings/my-account/profile.tsx#L225

leog (changes requested):
@Udit-takkar understood about the random string, but it would be much better to be able to assign `undefined` to it.

Also, shifting to test-policeman mode, we need a test for this to comply with our new testing standards 🙏 

image

leog (approved):
Great job, LGTM

PeerRich:
CleanShot 2023-09-29 at 20 56 47@2x
what is going on here?

leog:
> what is going on here?

@PeerRich the active and hover states of those fields don't match it seems, but not related to this PR though, we could create a bug for it to follow up

PR #11621, opened 2023-09-29
fix: Special (or hidden) Chars in uid field

### LINKED ISSUE #11608: Special (or hidden) Chars in uid field
The uid column in the "Booking" table sometimes contains values that must contain hidden characters. 

We had complaints that sometimes bookings cannot be edited by our users - the booking page would return a 404. I checked the url and looked for the UID in the booking table. A simple SELECT against the uid did not return any results, but adding a wildcard to the lookup did:

image

image

When updating the value via SQL manually, the page will show. Therefore, there must be some way for special chars to be added to the uid field on generation (or possibly through other means - API or such).

Any help would be appreciated.

xonx4l:
fixes #11608


PeerRich (changes requested):
this looks dangerous to merge.

can you make a self-review explaining what you are doing here and why?

PeerRich:
no explanation for this. closing

PR #11625, opened 2023-09-29
fix: prisma import in extract user route in Cal AI

DexterStorey:
One line fix to extract users route. Previously prisma was not imported - but since it was tested locally, where prisma is globally declared, we missed this!


zomars (approved):
LGTM

zomars:
I'm hot fixing this to prod @keithwillcode @PeerRich

PR #11633, opened 2023-09-30
fix: inverted logos and new logo for cal video

### LINKED ISSUE #11631: [CAL-2557] new cal video logo
icon

<sub>[CAL-2557](https://linear.app/calcom/issue/CAL-2557/new-cal-video-logo)</sub>

PeerRich:
before:
image

after:

- invert if "-dark" included
- invert if NOT /app-store (so we invert the map.svg and link.svg icons) 

CleanShot 2023-09-30 at 09 54 02@2x


apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -301,7 +302,11 @@ export const EventSetupTab = (
                     <div className="flex items-center">
                       <img
                         src={eventLocationType.iconUrl}
-                        className="h-4 w-4 dark:invert-[.65]"
+                        className={classNames(
+                          "h-4 w-4",
+                          !eventLocationType.iconUrl?.startsWith("/app-store") && "dark:invert",
+                          eventLocationType.iconUrl?.includes("-dark") && "dark:invert"
```
PeerRich, line 308: maybe worth putting into a helper function since we do this to every "locationIcon"

hariombalhara: I really don't want an /app-store check here. That's very fragile.

Do we not follow the same -dark approach for app-store svgs?

keithwillcode: Yeah duplicated logic like this is more fragile. We should consolidate it

keithwillcode (dismissed):

apps/web/components/eventtype/EventSetupTab.tsx
PeerRich: the /app-store check was used before. essentially this excludes the hard coded .svgs which we are inverting (map.svg and link.svg)

PeerRich:
IMO we can merge this and refactor later, its not a crazy code duplication and /app-store is needed for a check

RahulSoni2532:
I think we have to correct the word cal from call.?

emrysal (approved):

PR #11636, opened 2023-09-30
fix: #11555

### LINKED ISSUE #11555: Block/Restrict e,-,+ for number input fields for Event Type page

At Event Type page while creating a new event or editing an event, we are able to enter e,-,+ for "duration" field. Duration field accepts only number so we should block/restrict user from entering these characters.

### Steps to Reproduce

1. go to the "Event Types" page
2. Create a new event or edit an existing event
3. enter e,-,+ for duration field and try saving those changes.

### Actual Results

https://github.com/calcom/cal.com/assets/86093112/82363877-8ed8-49bb-96eb-1fd93a183493

### Expected Results
We should be able to block/restrict user from entering those characters instead of showing them error message.

mehul963:
fix: #11555


mehul963:
I don't know why E2E test 2/5 is got failed but I sure it is not releted with this issue #11555

mehul963 (commented):
I change in this file and add custom input format called formatNumber.
and update onChange method that check input using regex.
please review it

PR #11639, opened 2023-10-01
fix: #11516 - Error message / Unable to remove the 'Check for Conflicts' google calendar.

### LINKED ISSUE #11516: [CAL-2817] Cal.com Error message / Unable to remove the 'Check for Conflicts' google calendar. (video linked)

### Issue Summary

Cal.com configuration setting throwing an error when attempting to remove the 'check for conflicts' Google Calendar. Ideally users are able to check for conflicts on the same **or different** calendar as the one where to add events. 

### Steps to Reproduce

1. Add a google calendar where to add events.
2. Attempt to 'toggle off' the check for conflicts, or swap to a different calendar.
3. Generic Error message in the bottom right corner

I would expect to be able to remove the 'check for conflicts calendar'

### Actual Results

- Error message 
- 
### Expected Results

- The ideal would being able to toggle off check for conflicts calendar. This has many practical reasons. 

### Technical details

Recording and audio explanation of the bug in action and the error message: https://www.loom.com/share/a840f0fdcfa147d59bd476ec93143a9f

- Browser version: Google Chrome Version 117.0.5938.92 

- Node.js version

### Evidence

- How was this tested? See linked Loom video above. 

<sub>[CAL-2817](https://linear.app/calcom/issue/CAL-2817/calcom-error-message-unable-to-remove-the-check-for-conflicts-google)</sub>

vikaspatil0021:
## What does this PR do?
Fixes #11516 

## Type of change

## How should this be tested?
This PR can be tested as follows:
1. Add a Google Calendar to your account.
2. Go to this url - "http://localhost:3000/settings/my-account/calendars"
3. By default, the primary calendar toggle will be turned on. Try to turn it off.
4. After turning off the primary calendar toggle, refresh the page to check if the toggle state changed without any reason. If the toggle state remains unchanged after the refresh, it means your primary calendar is still turned off.
   
By following these steps, you can effectively add a Google Calendar, toggle the primary calendar off, and verify if the toggle state remains as desired.

## Mandatory Tasks

## Checklist

##Evidence
https://www.loom.com/share/3d17a0797a994fb3bc094765d3409a0c?sid=ad15639d-1f59-4850-9394-16092a6bf45f


alishaz-polymath:
Hey @vikaspatil0021 
Thank you for raising this PR 🙌 
Can you please self-review the code explaining the changes and desired outcome from those changes? This speeds up the review process 🙏 
[Read this for reference](https://cal.com/docs/knowledgebase/contributors-guide/pull-requests#bug-fix-prs)

packages/app-store/googlecalendar/api/add.ts
```diff
@@ -9,6 +9,7 @@ import { encodeOAuthState } from "../../_utils/oauth/encodeOAuthState";
 const scopes = [
   "https://www.googleapis.com/auth/calendar.readonly",
   "https://www.googleapis.com/auth/calendar.events",
+  "https://www.googleapis.com/auth/userinfo.email",
```
vikaspatil0021, line 12: adding scope to get email access from google account

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -50,6 +50,34 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
 
+  // Fetch the user's email from the Google OAuth2 userinfo endpoint using the access token in 'key'.
+  const userEmail = await fetch("https://www.googleapis.com/oauth2/v3/userinfo", {
+    method: "GET",
+    headers: {
+      Authorization: `Bearer ${key?.access_token}`,
+    },
+  })
+    .then((res) => res.json())
+    .then((res) => res.email); // Extract the user's email from the response.
```
vikaspatil0021, line 61: Made a get request to google api to get users email

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -50,6 +50,34 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+        externalId: userEmail, // Match the user's email.
+      },
+    },
+    create: {
+      userId: req.session.user.id,
+      integration: credentials?.type, // Set the integration type from the credential.
+      externalId: userEmail, // Set the external ID to the user's email.
+      credentialId: credentials?.id, // Associate the credential with this record.
+    },
+    update: {},
```
vikaspatil0021, line 78: creating a record in database to turn the toggle on by default when google calendar app is installed.
userId: taken from session
integration: taken from credentials
externalId: taken from userEmail
credentialId: taken from credentials

vikaspatil0021 (commented):
**Bug Description:**
The issue at hand was causing the toggle to turn on every time the page loaded, even when there was no record in the "selectedCalendar" table in the database. Consequently, when users attempted to manually toggle it off, a request was sent to the database to delete a record that did not exist. This led to an error message stating, "Something went wrong when toggling 'x@gmail.com.'"

**Root Cause:**
I found that the root cause of this behavior lies in the following expression:
```javascript
cal.isSelected || cal.externalId === data?.destinationCalendar?.externalId
```
The logic in this expression did not account for the absence of a record in the "selectedCalendar" table, causing the toggle to always turn on.


**Resolution Update:**
While the adjustments made to the expression indeed resolved the initial bug, they also had an unintended consequence, affecting the default calendar toggle. As it stands, all calendar toggles, including the default calendar used for adding events, are now turned off by default.

**Solution Implementation:**
To rectify this situation and ensure that the default calendar toggle remains on by default, I've made further changes in two separate files:

1. **Scope Addition:** In one of the files, I've introduced a new scope within the scopes array to request permission for accessing the user's email.

2. **API Request:** In the second file, I've implemented a GET request to Google's API URL to access the user's email. Once we retrieve the email address, a new record is created in the selectedCalendar database, utilizing the credentials generated. This approach ensures that the toggle state remains on for the default Google calendar immediately after installed.

**Why we need email:**
The reason for requesting the user's email is rooted in how Google Calendar manages calendar identification. Google Calendar associates a user's email as the external ID for their primary calendar. This unique email-based identification system allows us to distinguish the user's default calendar from any other calendars they may have.

**Outcome:**
With this implementation, when a user installs our Google Calendar integration:

1. **Default Calendar Toggled On:** By default, the calendar associated with the user's email (the primary calendar) will remain toggled on.

2. **Retained Toggle State:** The toggle state of the default calendar will persist even if the user decides to switch to a different calendar. The default calendar will remain unaffected unless they choose to manually toggle it off.

**Note:**
Under this approach, the toggle for the user's primary Google Calendar will indeed remain turned on, and it will not automatically change to the calendar selected by the user if they switch to a different calendar for adding events. This behavior ensures that the user's primary calendar stays selected for convenience, and users can independently toggle other calendars as needed.

Lastly, As a newcomer to contributions, I wanted to take a moment to express my appreciation for the opportunity to contribute to this project. However, I'd like to apologize if my recent approach doesn't meet the project's standards or expectations.

packages/app-store/googlecalendar/api/add.ts
exception: `note`: AFAIK adding this scope does not change the permissions we require from Google, but worth checking, cc @keithwillcode @hariombalhara

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -50,6 +50,34 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+    .then((res) => res.json())
+    .then((res) => res.email); // Extract the user's email from the response.
+
+  // Upsert (Insert or Update) a record of primary calendar in the 'selectedCalendar' table in the Prisma database.
+  await prisma.selectedCalendar.upsert({
+    where: {
+      userId_integration_externalId: {
+        userId: req.session.user.id,
+        integration: credentials?.type, // Match the credential type.
+        externalId: userEmail, // Match the user's email.
```
exception, line 69: `question`: Would externalId of `primary` do the same as providing the user's email?

exception:
Aside from comments, PR looks good and works as intended.

packages/app-store/googlecalendar/api/add.ts
keithwillcode: Tagging @joeauyeung as well so we confirm this is what we want

packages/app-store/googlecalendar/api/callback.ts
joeauyeung: @vikaspatil0021 I'm going to back up @exception on this one. Not only would that reduce one more call we have to make to Google but that means we don't need to add any new scope to our OAuth request. If we add to the scope then any current access tokens won't be able to access the new scope and could cause some unintended side effects.

packages/app-store/googlecalendar/api/add.ts
joeauyeung: @exception see this comment about adding to our current scope
https://github.com/calcom/cal.com/pull/11639/files#r1362753492

packages/app-store/googlecalendar/api/callback.ts
vikaspatil0021: > `question`: Why is this operation being done as an upsert when there's no actual update happening? If in fact, this is only called when the integration is enabled, an insert would suffice.

[f0863a1](https://github.com/calcom/cal.com/pull/11639/commits/f0863a129bc9e7d76593db1d57c865b34f86d444)
I've updated the code to use the 'insert' method since there's no actual update happening.

packages/app-store/googlecalendar/api/callback.ts
vikaspatil0021: > `question`: Would externalId of `primary` do the same as providing the user's email?

Yes, 
For Google's primary calendar, the user's email is used as the external ID.
TRPC output for primary calendar
Screenshot 2023-10-18 154241
for other calendars
Screenshot 2023-10-18 154509

vikaspatil0021:
@exception
I'm uncertain if I should revert the changes made in the 'add.ts' and 'callback.ts' files to ensure a successful PR merge with the main branch. The result of this merge will be that all toggles will remain off by default, and users will have manual control of the toggle even when calendars are changed. Could you please confirm this understanding?

packages/app-store/googlecalendar/api/callback.ts
vikaspatil0021: > @vikaspatil0021 I'm going to back up @exception on this one. Not only would that reduce one more call we have to make to Google but that means we don't need to add any new scope to our OAuth request. If we add to the scope then any current access tokens won't be able to access the new scope and could cause some unintended side effects.

@joeauyeung Thank you for your feedback and support

About scopes, any modifications made in the "add.ts" and "callback.ts" files should not trigger for existing access tokens. These files run only during the initial installation of the Google Calendar integration.
Please correct me if I'm mistaken.

exception: @joeauyeung I believe @vikaspatil0021 is correct, however, I'm not sure if upon requesting a refresh token and the scopes having changed there will be any issues. I would assume not, but better be safe than sorry.

packages/app-store/googlecalendar/api/callback.ts
exception: Thank you!

vikaspatil0021:
Hey @exception 
Changes have been reverted in the 'add.ts' and 'callback.ts' files, which should resolve the issue. Please note that after Google Calendar installation, all toggles will be off by default. Users will need to manually enable them, even for the default calendar, to use the 'check for conflicts' feature.

I hope this PR gets merged.

keithwillcode:
> Hey @exception Changes have been reverted in the 'add.ts' and 'callback.ts' files, which should resolve the issue. Please note that after Google Calendar installation, all toggles will be off by default. Users will need to manually enable them, even for the default calendar, to use the 'check for conflicts' feature.

@ciaranha @PeerRich Before we merge this, I would like your input on UX for this.

vikaspatil0021:
@keithwillcode Any updates on this?

vikaspatil0021:
Hey @exception @keithwillcode @PeerRich @ciaranha 
any updates on ux review?

@ejgennrich-joinbetter

ejgennrich-joinbetter:
I am still getting this bug. Any update on the UX review? @exception @keithwillcode @PeerRich @ciaranha 

I believe the question is: 

> Please note that after Google Calendar installation, all toggles will be off by default. Users will need to manually enable them, even for the default calendar, to use the 'check for conflicts' feature.

FWIW as an end user, I'd rather have flexibility (resolving this bug that has been in production for 2+months now...) over the potential UX-concern (namely, that end users will be required to manually toggle calendars they want to check for conflicts)

joeauyeung:
We are closing this PR in favour of https://github.com/calcom/cal.com/pull/12388. 

While reviewing this PR, we realized there was different logic between `apps/installed/calendar` and `settings/my-account/calendars`. While this PR fixes the logic found in the settings page, it would be better for both pages to share the same logic.

If you are having trouble right now I would suggest going to `apps/installed/calendar` to make any changes.

ejgennrich-joinbetter:
@joeauyeung Does that mean these settings live in 2 places and could display different settings in both places? Which setting is primary, and is the secondary-setting ignored? Are you all going to remove these settings from one of the places? Thanks for the added context, this sounds like good news?

PR #11661, opened 2023-10-02
#11658 issue fixed 

### LINKED ISSUE #11658: Can Book the Meeting with Identical Host and User(Host to Host)
I have encountered an issue while attempting to book a meeting by copying the URL from event types and opening it in the side tab. It seems that we are able to book a meeting when the same host and user (host to Host). Actually, we are not allowed to any host to book the meeting itself.
### Issue Summary
book

### Steps to Reproduce

1. Go to  Event Types section 
2. Copy the URL
3. open a new tab with the same window where you have already logged it 
2. select time and enter the same data (as host data like email)
3. click  on confirm btn 
4. we will see the booking  successfully (HOST TO HOST)

### Expected Results
It is not possible for users to book a meeting with themselves or Booking a meeting with oneself is not permitted.

arunkumar201:
#11658

### DIFF
--- a/apps/web/pages/api/book/event.ts
+++ b/apps/web/pages/api/book/event.ts
@@ -4,10 +4,11 @@ import { getServerSession } from "@calcom/features/auth/lib/getServerSession";
 import handleNewBooking from "@calcom/features/bookings/lib/handleNewBooking";
 import { defaultResponder } from "@calcom/lib/server";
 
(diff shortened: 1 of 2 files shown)


PeerRich:
see https://github.com/calcom/cal.com/issues/11658#issuecomment-1743691712

PR #11668, opened 2023-10-03
fix: Use only the first connected calendar as fallback to create event

hariombalhara:
## What does this PR do?

Fixes a scenario where duplicate calendar events are created across calendars. Actual steps to replicate the scenario aren't clear, but it happens when credentialId isn't set on the **destinationCalendar**.

In this case we now use the first connected calendar as the fallback instead of all connected calendars.

Ideally, this shouldn't be the case that there is a destination calendar with no credentialId, but this is damage control.

## Type of change

## How should this be tested?
- Install Google Calendar twice with two different accounts
- Choose the first calendar as the destinationCalendar in an event type
- Go to DB studio and manually set the credentialId to null for that destinationCalendar
image
- Now, create a booking for the event and it would landup in both the Google calendars. This scenario is now fixed and it would go to only one calendar and we would log a warning in that scenario.

## Mandatory Tasks

## Checklist
- I haven't added tests that prove my fix is effective or that my feature works -> I will be adding test for this scenario in the PR https://github.com/calcom/cal.com/pull/11670/files as there is some groundwork there that's been done to make it easy.

### DIFF
--- a/packages/core/CalendarManager.ts
+++ b/packages/core/CalendarManager.ts
@@ -231,7 +231,7 @@ export const createEvent = async (
 
   log.debug(
     "Creating calendar event",
-    JSON.stringify({
+    safeStringify({
       calEvent: getPiiFreeCalendarEvent(calEvent),
     })
   );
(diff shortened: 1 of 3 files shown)


packages/core/CalendarManager.ts
```diff
@@ -267,11 +267,20 @@ export const createEvent = async (
         })
     : undefined;
   if (!creationResult) {
-    logger.silly("createEvent failed", { success, uid, creationResult, originalEvent: calEvent, calError });
+    logger.error(
```
hariombalhara, line 270: Switched to logging it as an error.

packages/lib/piiFreeData.ts
```diff
@@ -78,11 +78,11 @@ export function getPiiFreeDestinationCalendar(destinationCalendar: Partial<Desti
   return {
     integration: destinationCalendar.integration,
     userId: destinationCalendar.userId,
+    credentialId: destinationCalendar.credentialId,
```
hariombalhara, line 81: credentialId isn't PII

zomars (approved):
Awesome work @hariombalhara 🙏🏽

PR #11686, opened 2023-10-04
fix: improves the misleading text when using limit future booking settings

### LINKED ISSUE #11321: Improve Availability Copy when using Limit Future Bookings settings

I set up the "Limit future bookings" to be "3 calendar days" into the future.
Screenshot 2023-09-12 at 5 30 13 PM

When a user is trying to book, the availability feels a bit misleading, and I am worried that people will be discouraged from checking again (in my case) every three days.

Screenshot 2023-09-12 at 5 28 25 PM

There is indeed availability in the following months, just that it is too far into the future, so you must wait until the threshold is met if you care about a particular day.

Improve the copy text, making it clear that the user can not book that far into the future.

- Disable the configuration

khushalmajoka:
## What does this PR do?

This PR improves the misleading text when using limit future booking settings.

Fixes #11321

Before:
image

After:
image

## Type of change

## Mandatory Tasks


khushalmajoka:
please add "hacktoberfest-accepted" label if possible.

packages/prisma/.env
```diff
@@ -1 +1,259 @@
-../../.env
\ No newline at end of file
+# ********** INDEX **********
```
keithwillcode, line 1: This file should not be here

keithwillcode (changes requested):

khushalmajoka:
Hi @keithwillcode, I have made the changes. Please review it and also add the "hacktoberfest-accepted" label.

khushalmajoka:
@keithwillcode please review my changes, its ready to be merged.

khushalmajoka:
Hi there, please somebody lookout for this and merge this PR, if its okay. @PeerRich

packages/features/calendars/DatePicker.tsx
```diff
@@ -240,9 +230,7 @@ const Days = ({
         </div>
       ))}
 
-      {!props.isLoading && includedDates && includedDates?.length === 0 && (
-        <NoAvailabilityOverlay month={month} nextMonthButton={nextMonthButton} />
-      )}
```
PeerRich, line 245: why are we deleting this? We wanna keep this in case people don't have availabilities in a month

khushalmajoka: Because @PeerRich sir, this is related to the limit future booking setting:

image

For example if it allows booking 10 days in the future then it is sure that there will be no availability anytime in the future after 10 days.

So instead of showing this to the user which have a button to view next month which is misleading because of the same reason mentioned above (i.e. there will be no availability in the future too).

image

we should show this:

image

Because sir, view next month says that "go to next month there may be availabilities in the next month" but actually it is not true.

PeerRich: ok but would this change not remove this functionality all together, lets say if you have no free times in October, but you have them in November? would we still show the button? because we want to show the button then

PR #9815, opened 2023-06-27
feat: Column View Available Slots Fix

VinoV1999:
## What does this PR do?

Shows only available slots in a column view. Jumps forward and back to the next available 6 slots, on click of left and right arrow respectively.

Fixes #9593
/claim #9593

https://github.com/calcom/cal.com/assets/123581387/cc3c6306-5b62-4247-ad35-6322953702a9

## Type of change

- Increased the extraDaysConfig of column view from 4 to 6.
- Implemented the logic to calculate the no of days in between continuous slots. And passed the required arguments to child pages.
- Modified the format of header.
- Assorted the available slots.
- Added useSlotsForAvailableDates function to filter out the required data.

PR #11696, opened 2023-10-04
fix: google calendar busy times

### LINKED ISSUE #11688: [CAL-2575] [CAL-2574] Busy events from Google Calendar not reflected on the booking page
Screenshot 2023-10-04 at 10 39 15

When you try to book the slot that is not supposed to be there, the error shows
Screenshot 2023-10-04 at 10 44 52

https://www.loom.com/share/b3e5cd34f41b4fc5b95b06707df1bbf7?sid=866dd324-620d-4bc2-af62-96789daf44ef 

Created via Threads. See full discussion: [https://threads.com/34540603431](https://threads.com/34540603431)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2574](https://linear.app/calcom/issue/CAL-2574/busy-events-from-google-calendar-not-reflected-on-the-booking-page)</sub>

<sub>[CAL-2575](https://linear.app/calcom/issue/CAL-2575/[cal-2574]-busy-events-from-google-calendar-not-reflected-on-the)</sub>

CarinaWolli:
## What does this PR do?

This PR fixes the following issue:
Busy times from Google Calendar didn't block out slots anymore after Nov 4. I could reproduce that with having everything set to NY timezone, but this probably also happens in other timezones.

The reason for that, is that the request `https://www.googleapis.com/calendar/v3/freeBusy` was throwing `The requested time range is too long`. 

Fixes #11688

## Type of change

## How should this be tested?

- Connect Google calendar
- Add a busy event to calendar
- Check if the slots are blocked off in the availability (test after Nov 4)

### DIFF
--- a/packages/features/bookings/Booker/Booker.tsx
+++ b/packages/features/bookings/Booker/Booker.tsx
@@ -102,18 +102,21 @@ const BookerComponent = ({
   const bookerLayouts = event.data?.profile?.bookerLayouts || defaultBookerLayoutSettings;
   const animationScope = useBookerResizeAnimation(layout, bookerState);
   const totalWeekDays = 7;
-  const addonDays =
-    nonEmptyScheduleDays.length < extraDays
-      ? (extraDays - nonEmptyScheduleDays.length + 1) * totalWeekDays
-      : nonEmptyScheduleDays.length === extraDays
-      ? totalWeekDays
-      : 0;
+  const addonDays = nonEmptyScheduleDays.length < extraDays ? extraDays - nonEmptyScheduleDays.length + 2 : 0;
 
   // Taking one more available slot(extraDays + 1) to calculate the no of days in between, that next and prev button need to shift.
   const availableSlots = nonEmptyScheduleDays.slice(0, extraDays + 1);
   if (nonEmptyScheduleDays.length !== 0)
     columnViewExtraDays.current =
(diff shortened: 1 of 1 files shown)


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -115,6 +115,7 @@ const BookerComponent = ({
     columnViewExtraDays.current =
       Math.abs(dayjs(selectedDate).diff(availableSlots[availableSlots.length - 2], "day")) + addonDays;
   const prefetchNextMonth =
```
CarinaWolli, line 117: This change was made here: https://github.com/calcom/cal.com/pull/9815/files#diff-d8cdd443610a4483db45dcc4da6a02bcff302ad6c591aea6fa1ee8646a52aedbR116

We only need `prefetchNextMonth` for the COLUMN_VIEW. The error `The requested time range is too long.` can still happen in the column view if the timespan shown is too long

zomars (approved):
Nice find @CarinaWolli can we follow up with a test to prevent regression? 🙏

packages/features/bookings/Booker/Booker.tsx
hariombalhara: I think we can fix it for longer ranges completely by breaking down the range into multiple acceptable ranges and sending multiple requests? 

I saw someone did similar for Zoho Calendar

CarinaWolli: Yes, I will make a follow-up PR 👍

PR #11699, opened 2023-10-04
feat: env for disable self hosted mode

### LINKED ISSUE #11698: Running App locally considered SELF_HOSTED
### Issue Summary

While running the app locally it is run as a SELF HOSTED application. This means that not all the features run locally in development.
For development purposes as well as running E2E tests on all features we need to fix the developer experience.
### Steps to Reproduce

1. Copy .env.example
2. Follow installation steps
3.  Then `yarn dx`
4. Try any premium feature

### Expected Results
To be able to test things properly, we need an env variable to easily enable and disable features for SELF_HOSTED.

shivamklr:
## What does this PR do?
Fixes #11698

## Type of change

## How should this be tested?

- Set `DISABLE_SELF_HOSTED_MODE` to "1".


zomars (changes requested):
I think this opens up the possibility to bypass license checks very easily. Should we discuss more about the problem we're trying to solve and the best way to do it?

PR #11700, opened 2023-10-04
fix: Event Type Rescheduling disabled for events that are already cancelled

### LINKED ISSUE #11669: Cancelled event can be rescheduled via a reschedule link in a confirmation emal

### Issue Summary

A cancelled event can be rescheduled using a reschedule link which was sent in a confirmation email, when this calendar event was first accepted/created.

### Steps to Reproduce

1. Using a link for a calendar event choose a date and time for a meeting, and confirm it. The page will load a confirmation page that the event was confirmed. At the bottom you see buttons: Reschedule & Cancel event. DON'T CLOSE THIS PAGE.
2. The owner of the link will get a confirmation email, where at the bottom there are buttons: Reschedule & Cancel event
3. Go back to the confirmation page from point 1, and cancel the event. You will get a cancellation email, and the event will be removed from your calendars.
4. Go back to the confirmation email from point 2, and press reschedule, it will open a page where you can choose date/time, and you will be able to reschedule an already cancelled event.

Logically this rescheduled link should show that the event has been cancelled, and users should user an initial create link. 

Chiranjeev-droid:
## What does this PR do?
This PR aims to disable rescheduling of cancelled events.
---
Fixes #11669 
---
Detailed Description
---
- Included  booking status from database in the queries
- If status is cancelled, redirect the user to cancelled booking page.
---
Before Changes 
Loom Video: https://www.loom.com/share/b88a4881f1cf4002bba0f6faab8f3ff0?sid=e52ad725-c438-4273-a5f3-5cf69cdbf4e7

## After Changes 
 Loom Video:  https://www.loom.com/share/0554d0110f06467498d839465f8f90a3?sid=8aa061e4-d6d4-459f-8053-e339484b36f7

## Type of change
## How should this be tested?
- Create a booking
- Go to the booking page link where we see option of rescheduling or cancel
- Copy the link to reschedule in new browser tab
- Now cancel the event and reschedule the event
- event successfully rescheduled but ideally it should not be.
- This pr fixes the above bug.

## Mandatory Tasks


Chiranjeev-droid:
Instead of redirecting the users to cancelled booking page, we can also show a not found page from the server side itself.

PeerRich (changes requested):
why do we wanna prevent this? just today I rescheduled an event that was cancelled before

Chiranjeev-droid:
image
one guy opened a issue and i found it legitmate so fixed the issue, because if an event is cancelled then how it can be rescheduled. If cancelled event can be rescheduled then why we are not giving a reschedule button on cancelled page also. Anyways if it is not a bug ,then please close the pr 🙂

PR #11701, opened 2023-10-04
fix: adding member to "other team" in an org by admin/owner gives unexpected behaviour.

### LINKED ISSUE #11691: adding member to "other team" in an org by admin/owner gives unexpected behaviour
### Issue Summary
adding member to "other team" in an org by admin/owner gives unexpected behaviour
* the members that were alredy present in the team are repeated twice and the go away once we refresh the page

### Evidence

[Screencast from 04-10-23 03:43:45 PM IST.webm](https://github.com/calcom/cal.com/assets/32706411/8da333ff-816e-48c2-ba72-62da9ee12d34)

JeevanJoshi4434:
Fixes #11691 

The bug was occurred due to preventing the state whenever you adding the new user.
Every time when you Invites new user there is an api calls for new Updated data so the function was trying to prevent old data whenever fetching new data ( setState(state.concat(newData)) --> old data + updated data where updated data was containing old data ). What I do that I changed the state to set new data (setState(newData)) and Added A check with chain operator if new data have length greater than 0 then It will add that data to the state.


.env.example
```diff
@@ -1,259 +0,0 @@
-# ********** INDEX **********
```
CarinaWolli, line 1: This file should not be removed

yarn.lock
```diff
@@ -5,46 +5,6 @@ __metadata:
   version: 6
   cacheKey: 8
 
-"@0no-co/graphql.web@npm:^1.0.1":
```
CarinaWolli, line 8: We shouldn't commit the changes from the yarn.lock file

PR #8082, opened 2023-04-04
Send broken integration emails on broken calendar.

hariombalhara:
## What does this PR do?

See https://threads.com/34461148315?s=1QYhgXtfsUaQ5ufUDeWm4x
Fixes #8197

Fixes
- No booking confirmation email when there is a broken calendar connection.
**Environment**:  Production

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Chore (refactoring code, technical debt, workflow improvements)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist

- I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)
- My code doesn't follow the style guidelines of this project
- I haven't performed a self-review of my own code and corrected any misspellings
- I haven't commented my code, particularly in hard-to-understand areas
- I haven't checked if my PR needs changes to the documentation
- I haven't checked if my changes generate no new warnings
- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #11702, opened 2023-10-05
fix: booking failure when we don't find credential matching the destination calendar integration name

hariombalhara:
## What does this PR do?

image
Ideally this should't be the case that there is a destinationCalendar but it has no credential. It looks like some of the existing event-types might have this problem.
Regression from #11668 
## Type of change

## How should this be tested?
- 

## Mandatory Tasks

### DIFF
--- a/packages/core/EventManager.ts
+++ b/packages/core/EventManager.ts
@@ -400,16 +400,28 @@ export default class EventManager {
             (c) => c.type === destination.integration
           );
           // It might not be the first connected calendar as it seems that the order is not guaranteed to be ascending of credentialId.
-          const firstCalendarCredential = destinationCalendarCredentials[0];
-          log.warn(
-            "No credentialId found for destination calendar, falling back to first found calendar",
-            safeStringify({
-              destination: getPiiFreeDestinationCalendar(destination),
-              firstConnectedCalendar: getPiiFreeCredential(firstCalendarCredential),
-            })
-          );
+          const firstCalendarCredential: (typeof destinationCalendarCredentials)[number] | undefined =
+            destinationCalendarCredentials[0];
+          // It is seen in the production that there are `destinationCalendars` with no calendar credential at all.
+          if (firstCalendarCredential) {
+            log.warn(
+              "No credentialId found for destination calendar, falling back to first found calendar",
+              safeStringify({
+                destination: getPiiFreeDestinationCalendar(destination),
+                firstConnectedCalendar: getPiiFreeCredential(firstCalendarCredential),
+              })
+            );
 
-          createdEvents.push(await createEvent(firstCalendarCredential, event));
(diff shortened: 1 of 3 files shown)


packages/core/EventManager.ts
```diff
@@ -400,20 +404,32 @@ export default class EventManager {
...
-          log.warn(
-            "No credentialId found for destination calendar, falling back to first found calendar",
-            safeStringify({
-              destination: getPiiFreeDestinationCalendar(destination),
-              firstConnectedCalendar: getPiiFreeCredential(firstCalendarCredential),
-            })
-          );
-
-          createdEvents.push(await createEvent(firstCalendarCredential, event));
+          const firstCalendarCredential: (typeof destinationCalendarCredentials)[number] | undefined =
```
hariombalhara, line 407: With our current configuration, TypeScript doesn't consider it possibly undefined which it could be.

packages/core/EventManager.ts
```diff
@@ -400,20 +404,32 @@ export default class EventManager {
...
+              `No credentialId found for destination calendar, falling back to first found credential matching  destination.integration=${destination.integration}`,
+              safeStringify({
+                destination: getPiiFreeDestinationCalendar(destination),
+                firstConnectedCalendar: getPiiFreeCredential(firstCalendarCredential),
+              })
+            );
+
+            createdEvents.push(await createEvent(firstCalendarCredential, event));
+          } else {
+            loggerWithEventDetails.error(
```
hariombalhara, line 421: Log the erroneous situation which was earlier silently ignored.

CarinaWolli (approved):
LGTM 👍

keithwillcode (changes requested):
I’m concerned this will cause even more confusion if bookings start to go through but no calendar events are created. 

I think our approach to this should be different. 

- Find root cause of why/how this happened
- For event types that this would be triggered on if someone were to book, give the host a notification informing them the destination calendar is broken and they need to reconnect. 

Tagging @ciaranha here for UX suggestions.

hariombalhara:
Thanks for the review @keithwillcode @CarinaWolli .

Yeah I though about that @keithwillcode. I am just going back to the previous behaviour for now which was there before https://github.com/calcom/cal.com/pull/11668 was merged.

There is a flow to send a broken integration email which is currently disabled for a long time now. There  was [an attempt to enable it earlier ](https://github.com/calcom/cal.com/pull/8082/files), I would revive that again.

I would add that broken integration flow here as well.

I am creating a ticket to hook that broken integration flow here.

hariombalhara:
Closing this as we don't want booking to be successful when there is a calendar there and it's failing to create event in it.

keithwillcode:
CleanShot 2023-10-06 at 13 19 27@2x

We are going to hold off on merging this in favor of updating the booking page to something like the following for users who have no connected calendar/credentials.

PR #11710, opened 2023-10-05
fix: should allow org owners and admins to delete members from subteams

### LINKED ISSUE #11709: organizations - should allow org owners and admins to delete members from subteams
### Issue Summary
if we are allowing org owners and admins to add members to sub-teams then we should also allow them to remove users from sub-teams

### Actual Results
 - currently org owners and admins can add members to sub teams but not delete them

### Expected Results

- should allow org owners and admins to add as well as remove members from sub-teams

### Evidence
Screenshot from 2023-10-05 15-54-38

here the add button is visible but the options button in front of each member is missing

ibex088:
Fixes #11709

### user is not a org-admin/owner
Screenshot from 2023-10-05 16-18-16

### user is a org-admin/owner
Screenshot from 2023-10-05 16-18-09


packages/features/ee/teams/components/MemberListItem.tsx
```diff
@@ -109,7 +110,8 @@ export default function MemberListItem(props: Props) {
       (props.member.role !== MembershipRole.OWNER ||
         ownersInTeam() > 1 ||
         props.member.id !== currentUserId)) ||
-    (props.team.membership?.role === MembershipRole.ADMIN && props.member.role !== MembershipRole.OWNER);
+    (props.team.membership?.role === MembershipRole.ADMIN && props.member.role !== MembershipRole.OWNER) ||
```
CarinaWolli, line 113: We should do this for 'Other Teams' too (`organizations/pages/components/MemberListItem.tsx`)

CarinaWolli (changes requested):
Requesting changes as `removeMember.handler.ts` is throwing `UNAUTHORIZED` when an org owners tries to remove a team member from a team where they aren't admin/owner

packages/features/ee/teams/components/MemberListItem.tsx
SomayChauhan: hi @CarinaWolli
Screenshot from 2023-11-01 17-05-23
the other-teams tab is only visible to the organization admin and owners right?


and they should have all acceess (like deleting, editing) and (resending invitation for pending members) 
Screenshot from 2023-11-01 17-07-19
Screenshot from 2023-11-01 17-07-04
then why are we using a seperate component for other team why not just use /teams/pages/team-members-view

SomayChauhan: check this pr out #12196

CarinaWolli (approved):
Looks good now, we can revisit 'other Teams' in a separate PR. 

Thank you 🙏

PR #11714, opened 2023-10-05
feat: add parentid eventtype API along with children eventTypeIds

alishaz-polymath:
## What does this PR do?

This PR attempts to enable creation of and maintenance of a Managed EventType via API. For that, it needs the following:

- In response of a `GET` request
  - `parentId` in the child event type
  - children of a managed event type as an array containing their `id` and `userId` of the user associated with this event type
- As a body parameter for a `POST` request
  - `parentId` in the child event type

## Type of change

## How should this be tested?

Simply create a Managed Event Type by passing in a request of similar structure as a team event type, except the following:
    -  `schedulingType` should be `MANAGED`
    - `metadata` should contain something like

```
{
  "config": {
    "useHostSchedulesForTeamEvent": false
  },
  "managedEventConfig": {
    "unlockedFields": {
      "locations": true,
      "scheduleId": true,
      "destinationCalendar": true
    }
  }
}
```
- You should see a managed event created. Now, the way managed events work, to associate a member of the team you would need to create a single event for them with the exact same details (except not a team event) as the above, but include `parentId` as the `eventTypeId` of the newly created Manged Event Type. 
- You should now see the user you created the event type for is now assigned in this Managed Event Type.
- Finally, to test `GET` request, you can easily run `GET` queries against the Managed Event Type to see that it returns the children event type Ids and userIds.

## Mandatory Tasks


emrysal (dismissed):
We need ownership validation on the parentId - otherwise any event type can be marked as parent 😰

apps/api/pages/api/event-types/_post.ts
```diff
@@ -276,6 +281,11 @@ async function postHandler(req: NextApiRequest) {
 
   await checkPermissions(req);
 
+  if (isAdmin && parsedBody.parentId) {
+    await checkParentEventOwnership(parsedBody.parentId, userId);
+    await checkUserMembership(parsedBody.parentId, parsedBody.userId);
```
alishaz-polymath, line 286: We check:
1. If `parentId` provided is of the eventType that is a team event type, and the user making the request is either `Admin` or `Owner` of that team.
2. If `userId` provided is part of the team that the parentId belongs to.

emrysal (approved):
LGTM

zomars:
@emrysal @alishaz-polymath heads up. This is breaking API builds on main. Pushing a fix.

apps/api/pages/api/event-types/_post.ts
zomars: userId could be undefined. Causing a compile error.

zomars:
Pushed a fix to main

https://github.com/calcom/cal.com/commit/a53ea33168d75f377cc7841ea9a559b73a10a9ed

PR #11716, opened 2023-10-05
test: metadata for the event-types page

grzpab:
## What does this PR do?

This PR adds metadata tests for the event-types page.
It takes into account that title, `og:title` and `og:image` tags are unstable and their content changes.

## Type of change

## How should this be tested?
No additional testing needed.

## Checklist


PeerRich:
i dont think this is something we need to write an E2E test for. thank you tho!

PR #11703, opened 2023-10-05
fix: buffer dropdown text

CarinaWolli:
## What does this PR do?

Before: 
Screenshot 2023-10-05 at 09 40 52

After: 
Screenshot 2023-10-05 at 09 38 45

PR #11721, opened 2023-10-06
hotfix: buffer dropdown text

CarinaWolli:
## What does this PR do?

Hotfix for #11703


CarinaWolli:
let's make a release instead

PR #11722, opened 2023-10-06
fix: Function sendScheduledSeatsEmails ignore rule 'Disable default confirmation emails for attendees'

### LINKED ISSUE #11677: [CAL-2580] Function sendScheduledSeatsEmails ignore rule "Disable default confirmation emails for attendees"
### Issue Summary

When you have a event that offers seats, the confirmation email is sent to the 2nd attendees onwards even if the option _"Disable default confirmation emails for attendees"_ is enabled and you have active workflow to send emails for the event type.

The confirmation email is not sent for the first attendee because it uses the function `sendScheduledEmails` which have the props `hostEmailDisabled` and `attendeeEmailDisabled`, however the `sendScheduledSeatsEmails` doesn't have this props and ignore the rule to disable default confirmation emails for attendees.

### Steps to Reproduce

1. Create a Event that offers seats (Tab Advanced - Offer Seats)
2. Create and apply a custom workflow that sends email to the attendee after a event is booked
3. Go back to the event, Advanced Tab and check the option _Disable default confirmation emails for attendees_
4. Register for the event, you will receive only one e-mail, from the workflow.
5. Register again for the event with a different e-mail, you will receive two e-mails, one from the workflow and the other one from default email confirmation.

### Actual Results

- It doesn't respect the configuration _"Disable default confirmation emails for attendees"_

### Expected Results

- To not send the default confirmation when using the function `sendScheduledSeatsEmails` when the option _"Disable default confirmation emails for attendees"_ is enabled and you have active workflow to send emails for the event type.

### Technical details

Using production env.

### Evidence

Screenshot 2023-10-03 at 12 31 33

Screenshot 2023-10-03 at 12 32 45

Screenshot 2023-10-03 at 12 33 13

<sub>[CAL-2580](https://linear.app/calcom/issue/CAL-2580/function-sendscheduledseatsemails-ignore-rule-disable-default)</sub>

siddharth-movaliya:
## What does this PR do?

This PR addresses an issue where confirmation emails were incorrectly sent to all attendees, even when the "Disable default confirmation emails for attendees" option was enabled for an event. The problem specifically affected the first attendee, as the function `sendScheduledSeatsEmails` did not have the necessary props (`hostEmailDisabled` and `attendeeEmailDisabled`) to enforce this rule.

Fixes #11677 

## Requirement/Documentation
.
## Type of change

## How should this be tested?

1. Create a Event that offers seats (Tab Advanced - Offer Seats)
2. Create and apply a custom workflow that sends email to the attendee after a event is booked
3. Go back to the event, Advanced Tab and check the option Disable default confirmation emails for attendees
4. Register for the event from different accounts, you will receive only one e-mail, which will be from the workflow.

## Mandatory Tasks


CarinaWolli (approved):
Looks good! Thank you for your contribution 🙏

PR #11733, opened 2023-10-06
fix: Backwards compatibility on google reschedule/cancel

emrysal:
## What does this PR do?

Fixes issue with backwards compatibility of reschedule/cancel introduced in #10967


CarinaWolli (approved):

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -284,9 +284,10 @@ export default class GoogleCalendarService implements Calendar {
 
     const calendar = await this.authedCalendar();
```
hariombalhara, line 285: I think we can add a test for this in googleCalendar/CalendarService.test.ts

PR #11740, opened 2023-10-06
fix: API cold boot issues

### LINKED ISSUE #11739: API can have extremely slow cold boots due to tRPC router being imported

### Issue Summary

Around 5 months ago, we fixed most of the issues with cold boots for our web app. See https://cal.com/blog/cal-com-cold-start-resolution-blog. We are now seeing the same problem with certain API endpoints that use tRPC routers. Instead of importing the `viewer` router, which in turn loads 20+ other routers and all of their dependencies, we need to import the specific router needed to do the job.

Example: https://github.com/calcom/cal.com/blob/b4315b186a8808d062c298e8bb331f2f7720a798/apps/api/pages/api/slots/_get.ts#L6

### Steps to Reproduce

1. Let the API go cold (on a preview env)
2. Call the api/slots endpoint

### Actual Results

- It can take upwards of 40+ seconds to load whereas on a warm call it takes around 1.2-1.7s for most calls.

### Expected Results

- 1.2-1.7s for most calls.

keithwillcode:
## What does this PR do?

Fixes #11739

## Type of change

## How should this be tested?

- Run all of these endpoints and ensure the functionality has not regressed

### DIFF
--- a/apps/api/pages/api/slots/_get.ts
+++ b/apps/api/pages/api/slots/_get.ts
@@ -3,7 +3,7 @@ import type { NextApiRequest, NextApiResponse } from "next";
 import { HttpError } from "@calcom/lib/http-error";
 import { defaultResponder } from "@calcom/lib/server";
 import { createContext } from "@calcom/trpc/server/createContext";
-import { viewerRouter } from "@calcom/trpc/server/routers/viewer/_router";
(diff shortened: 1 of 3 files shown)


emrysal (approved):

keithwillcode:
@emrysal the api build was failing due to type errors

PR #11741, opened 2023-10-06
chore: ensure compat type definitions do not break existing types

grzpab:
## What does this PR do?

This PR introduces changes to handle nullable return value from Next.js hooks (`useSearchParams`, `useParams`, `usePathname` and `useRouter`).
If you add the `app` directory to the `web` directory and run `yarn build`, the types will be updated by Next.js builder/plugin and the projects will then include the compat type definitions provided by Next.js.

This mechanism is coded here: https://github.com/vercel/next.js/blob/524b31513a58e58e15862ac8aa3f27da8a47a267/packages/next/src/lib/typescript/writeAppTypeDeclarations.ts#L54

The example compat type definition is here: https://github.com/vercel/next.js/blob/524b31513a58e58e15862ac8aa3f27da8a47a267/packages/next/navigation-types/compat/navigation.d.ts

Fixes # (issue)
There is no reported issue, this problem becomes visible only when the `app` directory is added and someone runs `yarn type-check`.

## Type of change

## How should this be tested?
This requires the entire app check, because I touched
* `fromEntriesWithDuplicateKeys` in `packages/lib/hooks/useRouterQuery.ts` which is used in `useRouterQuery`, which in turn is used in 19 pages
* the metadata generating utilities (`packages/lib/next-seo.config.ts`)

## Checklist

### DIFF
--- a/packages/lib/hooks/useRouterQuery.ts
+++ b/packages/lib/hooks/useRouterQuery.ts
@@ -3,8 +3,13 @@ import { useSearchParams } from "next/navigation";
 /**
  * An alternative to Object.fromEntries that allows duplicate keys.
  */
-function fromEntriesWithDuplicateKeys(entries: ReturnType<ReturnType<typeof useSearchParams>["entries"]>) {
+function fromEntriesWithDuplicateKeys(entries: IterableIterator<[string, string]> | null) {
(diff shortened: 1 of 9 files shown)


zomars (approved):
LGTM!

PR #11747, opened 2023-10-07
feat: check availability only for first two occurrences while booking a recurring event

### LINKED ISSUE #11654: [CAL-2563] recurring events: only check for 2 recurrences for conflict, not all of them
lets only check 2 recurrences in advanced, not N

if you are booking an event for 12 weeks, check for 2 weeks, not all 12.

this will make more recurring bookings pass

downside: you recurring event may conflict with your calendar in 3 weeks, but this is fine because people offering recurring events typically clear their schedule anyway:

i.e.: personal tennis training is always Monday 5PM

<sub>[CAL-2563](https://linear.app/calcom/issue/CAL-2563/recurring-events-only-check-for-2-recurrences-for-conflict-not-all-of)</sub>

ujwalkumar1995:
## What does this PR do?

In recurring events, check for availability conflicts only for the first two occurrences.

Fixes #11654 

## Type of change

## How should this be tested?

Create recurring event.
Have multiple occurrences for the event (preferably greater than 2).
Make sure to have availability conflicts for the occurrences past the first two occurrences.
Event should be allowed to create.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -413,13 +413,14 @@ async function ensureAvailableUsers(
         recurringDatesInfo?.currentRecurringIndex === 0 &&
         recurringDatesInfo.allRecurringDates
       ) {
-        const allBookingDates = recurringDatesInfo.allRecurringDates.map((strDate) => new Date(strDate));
+        const firstTwoRecurringDates = recurringDatesInfo.allRecurringDates.slice(0, 2);
+        const firstTwoBookingDates = firstTwoRecurringDates.map((strDate) => new Date(strDate));
         // Go through each date for the recurring event and check if each one's availability
         // DONE: Decreased computational complexity from O(2^n) to O(n) by refactoring this loop to stop
         // running at the first unavailable time.
         let i = 0;
-        while (!foundConflict && i < allBookingDates.length) {
-          foundConflict = checkForConflicts(bufferedBusyTimes, allBookingDates[i++], duration);
+        while (!foundConflict && i < firstTwoBookingDates.length) {
+          foundConflict = checkForConflicts(bufferedBusyTimes, firstTwoBookingDates[i++], duration);
         }
       } else {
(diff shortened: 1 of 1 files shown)


hariombalhara (approved):

hariombalhara (changes requested):
Seeing issues in testing.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -413,13 +413,15 @@ async function ensureAvailableUsers(
...
       ) {
-        const allBookingDates = recurringDatesInfo.allRecurringDates.map((strDate) => new Date(strDate));
+        const firstTwoBookingDates = recurringDatesInfo.allRecurringDates
+          .slice(0, 2)
+          .map((strDate) => new Date(strDate));
         // Go through each date for the recurring event and check if each one's availability
         // DONE: Decreased computational complexity from O(2^n) to O(n) by refactoring this loop to stop
         // running at the first unavailable time.
         let i = 0;
-        while (!foundConflict && i < allBookingDates.length) {
```
hariombalhara, line 421: I don't think this would actually work.

Here is why:
`bufferedBusyTimes` contains busy times only for the DateRange for which `handleNewBooking` was called and we don't pass it the entire 12 weeks date range(for a 12 occurence recurring event), we actually call it 12 times with 12 different timeslots.

So, I believe this code is dead code as of now, maybe due to some refactor.

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara, line 1: To fix the issue we need to probably do `ensureAvailableUsers` for currentRecurringIndex = 0 or 1 and then skip where it is greater or equal to 2. Thus, it's going to freely allow bookings for those

PeerRich:
thank you for trying, but we'll go with @hariombalhara solution https://github.com/calcom/cal.com/pull/11774

hariombalhara (commented):
closing it in favour of other PR

Issue #11753, opened 2023-10-08
Typescript type error in `useRouterQuery` hook


rajpatelbot:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

At `useRouterQuery.ts` file `const _searchParams = new URLSearchParams(searchParams);` this line has a TS error about `searchParams` type. It the `URLSearchParams` requires `init?: string | string[][] | Record<string, string> | URLSearchParams | undefined` between this but the `searchParam` is `ReadonlyURLSearchParams`.

### Steps to Reproduce

1. Went to `lib/hooks/useRouterQuery.ts`
2. See TS error

- If its look relevent to maintainers i will happy to create a PR on it.

Thankyou..

sean-brydon:
Myself and @Udit-takkar do not see any type errors in this file.

<img width="680" alt="CleanShot 2023-10-10 at 11 58 55@2x" src="https://github.com/calcom/cal.com/assets/55134778/6810d861-7f0d-4b19-9054-d13e762db0e4">

rajpatelbot:
Don't know why this is happening. May be due to extention.

PR #11754, opened 2023-10-08
fix: Fixed The Typescript type error in useRouterQuery hook

AkashMondal0:
Fix Issue:  #11753


Udit-takkar (commented):
@AkashMondal0 I don't see any type error in useRouterQuery before.

Screenshot 2023-10-09 at 8 42 33 PM

AkashMondal0:
> @AkashMondal0 I don't see any type error in useRouterQuery before.
> 
> Screenshot 2023-10-09 at 8 42 33 PM

Screenshot (4)

PR #11761, opened 2023-10-09
refactor: only keep webhook and disable sent email

hoangbits:
## What does this PR do?

1. Disable sending email 
2. keep existing webhooking triggering code

## Type of change

### DIFF
--- a/apps/web/pages/api/recorded-daily-video.ts
+++ b/apps/web/pages/api/recorded-daily-video.ts
@@ -4,11 +4,9 @@ import { z } from "zod";
 
 import { DailyLocationType } from "@calcom/app-store/locations";
 import { getDownloadLinkOfCalVideoByRecordingId } from "@calcom/core/videoClient";
-import { sendDailyVideoRecordingEmails } from "@calcom/emails";
 import { getServerSession } from "@calcom/features/auth/lib/getServerSession";
 import getWebhooks from "@calcom/features/webhooks/lib/getWebhooks";
 import sendPayload from "@calcom/features/webhooks/lib/sendPayload";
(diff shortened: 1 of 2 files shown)


Udit-takkar (commented):
@hoangbits some people don't use recorded-daily-video webhook and just want the email with the recording download link sent to organiser and all attendees.

we can probably skip sending emails when there is a webhook present

PeerRich (changes requested):
many issues with this PR
- dont commit yarn.lock
- I don't think we should disable emails for everyone, even if a webhook is available
- disabling emails is dangerous and will always lead to edge cases where people expect an email
- why is it so bad if we also send an email?
- if anything, let people decide if they should send an email or not

I don't find an issue for this PR, please explain why you are doing this

hoangbits:
Sorry I create PR in wrong place.

hoangbits:
Sorry I create PR in wrong place.

hoangbits:
> many issues with this PR
> 
> * dont commit yarn.lock
> * I don't think we should disable emails for everyone, even if a webhook is available
> * disabling emails is dangerous and will always lead to edge cases where people expect an email
> * why is it so bad if we also send an email?
> * if anything, let people decide if they should send an email or not
> 
> I don't find an issue for this PR, please explain why you are doing this

Hello Guys.
I think not all the company will using SendGrid And if the code on this line stuck: https://github.com/calcom/cal.com/blob/3e08c668882ca24ec756438838699c7479d92dd2/apps/web/pages/api/recorded-daily-video.ts#L66
The subsequent line which trigger webhook will not be run.
I think We should handle it in a way that webhook don't be stop even if user don't adding config for SendGrid.

PR #11767, opened 2023-10-09
fix: update package.json to export typescript types

M-Nasser-M:
updated package to fix the problem of types not properly exported so it fails at build time if you are using typescript and also not type completions and erros while working with typescript

## What does this PR do?

Fixes#11301  # 11408

## Type of change

- [ *] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- Are there environment variables that should be set? no
- What are the minimal test data to have? a react-ts template with the react embed import as a dependecy 
- What is expected (happy path) to have (input and output)? the project builds correctly

## Mandatory Tasks

- [* ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

### DIFF
--- a/packages/embeds/embed-react/package.json
+++ b/packages/embeds/embed-react/package.json
@@ -39,7 +39,8 @@
   "exports": {
     ".": {
       "import": "./dist/Cal.es.mjs",
-      "require": "./dist/Cal.umd.js"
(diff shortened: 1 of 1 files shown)


emrysal (approved):
Thanks for this; looks good to me 🙏

Issue #9485, opened 2023-06-13
CalDAV integration with Fastmail is generating duplicate, erroneous invitation emails.


mtnowl:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

created from https://github.com/calcom/cal.com/issues/3457#issuecomment-1581465501, but I am also using Fastmail and seeing the same issues.

We also use Fastmail and have been seeing some oddities with the [cal.com](http://cal.com/) integration. Namely, that participants are receiving multiple calendar invites, which on it's own isn't that big of a deal. However, the Fastmail invites are always in UTC, not in the local time of the cal.com invite.

This is creating a lot of confusion with participants, they keep missing our meetings, not realizing one invite is UTC and one in local time. I spoke with the Fastmail team and they relayed the following notes about cal's specific integration.

1. there is a property (SCHEDULE-AGENT=CLIENT) a CalDAV client can add to the event that would prevent CalDAV servers from sending invitations. It appears cal.com is not setting this property even though they are sending an invitation.
2. the invitation being sent by [cal.com](http://cal.com/) is using a different UID. Basically every event has a UID that uniquely identifies the event. Since [cal.com](http://cal.com/) is sending the invitation of the event with a different UID, if you end up adding this event to your calendar [cal.com](http://cal.com/), it will create a duplicate event in your calendar as well. They need to fix it so that all emails associated with a single event use the same original UID.
3. the event added from [cal.com](http://cal.com/) doesn't have a timezone associated with it. Though it mentions the invitees timezone as "America/Chicago" in the notes, the event itself doesn't include a timezone and only specifies the time in UTC. Since the event time is specified in UTC, Fastmail's invitation email will also be sent in UTC.
4. There seems to be no way to stop Cal.com from creating the calendar events at all, and thus no real workaround.

### Steps to Reproduce

1. Connect to Fastmail using CalDav integration
2. Create a test booking
3. See issues above.

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?
1. SCHEDULE-AGENT=CLIENT should be set, so Fastmail knows not to send an invitation. **This is the main one.**
2. UID should be unique per event.
3. Event time should have a timezone associated with it.
4. An option could be added to make calendar event creation optional (nice-to-have, not essential).

### Technical details

None other than Fastmail/CalDAV.

mtnowl:
FYI @alishaz-polymath

alishaz-polymath:
[...]

PR #11771, opened 2023-10-09
fix: Mails are sent to the organizer twice

dasJ:
## What does this PR do?

The `ics` package doesn't support setting the `SCHEDULE-AGENT` parameter, there is a PR (https://github.com/adamgibbons/ics/pull/248), but it is not merged.

This is a workaround that relies on the fact that the `ics` package does not properly escape the `name` field of the organizer. In a perfect world, they would merge the PR and create a new release, but this is unfortunately not the world we live in.

Fixes # (issue)

This resolves one of the listed issues in #9485 which also affects NextCloud and SOGo users via WebDav.

## Requirement/Documentation

## Type of change

## How should this be tested?

- Pretty much the reproduce section of #9485

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/lib/CalendarService.ts
+++ b/packages/lib/CalendarService.ts
@@ -96,7 +96,13 @@ const getDuration = (start: string, end: string): DurationObject => ({
 });
 
 const mapAttendees = (attendees: Person[]): Attendee[] =>
-  attendees.map(({ email, name }) => ({ name, email, partstat: "NEEDS-ACTION" }));
+  // TODO Use `scheduleAgent` once https://github.com/adamgibbons/ics/pull/248 is merged rather
+  // than forcing the value into the partstat string.
+  attendees.map(({ email, name }) => ({
+    name,
+    email,
+    partstat: "NEEDS-ACTION;SCHEDULE-AGENT=CLIENT" as any,
+  }));
 
 export default abstract class BaseCalendarService implements Calendar {
   private url = "";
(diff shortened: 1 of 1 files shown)


neilj:
It looks like this just adds `SCHEDULE-AGENT=CLIENT` to the organizer. As per [RFC6638](https://datatracker.ietf.org/doc/html/rfc6638#section-7.1), you need to set this on each attendee that the client is handling the scheduling for.

dasJ:
Fair point, I wanted to start with this but it makes sense to also set this to every attendee.

dasJ:
Since the question may come up: I tried calling the ics package as-is and replacing the attendee into the resulting string. This did not work as well as I hoped because the ical event has a line limit of 75 so either my regex doesn't match or it may produce lines that are too long.

zomars (changes requested):

Can we come up with a simple test to prevent a future regression?

dasJ:
I'd love to but I don't see how unfortunately. My typescript is not the best (which you may see from the number of pushes I did ;) and the `createEvent` function seems to call a lot of things that don't seem to be currently mocked.
In fact, the `createEvent` function (that I modify) is usually mocked in tests.
Do you have any pointers for me how to get started?

dasJ:
bruh

alishaz-polymath:
@dasJ removed from stale state, that wasn't intended I'm sure 😅
But we do need tests for it still. 
@Udit-takkar can you please take a look?

Udit-takkar:
I am not sure about these changes @joeauyeung might be the best person to review this

joeauyeung (changes requested):
Tested this with Mailfence emails are still being sent from the CalDAV provider. Looking at the ics file, it seems like your changes aren't being applied to to the organizer and attendee fields.

ORGANIZER;CN=Free Example:mailto:free@example.com
ATTENDEE;RSVP=TRUE;ROLE=REQ-PARTICIPANT;PARTSTAT=ACCEPTED;CN=Test:mailto:*****@gmail.com

PR #11772, opened 2023-10-09
feat: `AvailabilityList` component

Ryukemeister:
## What does this PR do?

Adds an `AvailabilityList` component for Cal Atoms

## Type of change


packages/atoms/availabilitylist/AvailabilityList.tsx
```diff
@@ -0,0 +1,72 @@
...
+  deleteMutation,
+}: {
+  schedules: Schedule[] | [];
+  onCreateMutation: (values: {
+    onSucess: (schedule: Schedule) => void;
+    onError: (err: HttpError) => void;
+  }) => void;
+  updateMutation: ({ scheduleId, isDefault }: { scheduleId: number; isDefault: boolean }) => void;
+  duplicateMutation: ({ scheduleId }: { scheduleId: number }) => void;
+  deleteMutation: ({ scheduleId }: { scheduleId: number }) => void;
```
supalarry, line 40: Can we create `type AvailabilityListProps` that groups all of these props?

supalarry: The same goes for other components like `EmptyScreen`

packages/atoms/availabilitylist/Form.tsx
```diff
@@ -0,0 +1,41 @@
+import type { ReactElement, Ref } from "react";
+import React, { forwardRef } from "react";
+import type { FieldValues, SubmitHandler, UseFormReturn } from "react-hook-form";
+import { FormProvider } from "react-hook-form";
+
+// import { showToast } from "../../..";
```
supalarry, line 6: Comment needs to be removed

packages/atoms/availabilitylist/AvailabilityList.tsx
```diff
@@ -0,0 +1,72 @@
+import { NewScheduleButton } from "availabilitylist/NewScheduleButton";
```
supalarry, line 1: Can we rename `atoms/availabilitylist/` to `atoms/availability-list/` ?

supalarry: Then, I would re-name `AvailabilityList.tsx` to `index.tsx` so that it's clear it's the main file within `availability-list`

packages/atoms/availabilitylist/ScheduleListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+    userId: number | null;
+    eventTypeId: number | null;
+    date: Date | null;
+    days: number[];
+    scheduleId: number | null;
+  }[];
+  timezone: string | null;
+};
+
+export function Availability({
```
supalarry, line 33: This component's name is `Availability` but it's stored in a file named `ScheduleListItem`.

packages/atoms/availabilitylist/ScheduleListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+  deleteFunction: ({ scheduleId }: { scheduleId: number }) => void;
+  displayOptions?: {
+    timeZone?: string;
+    hour12?: boolean;
+  };
+}) {
+  const { toast } = useToast();
+
+  return (
+    <li key={schedule.id}>
```
supalarry, line 54: Can the returned JSX be split into smaller components that represent what we see in the UI, like <TimeZone> and for example 3 dots with Duplicate / Delete under <Controls> component etc ? It is hard to understand what's going on.

packages/atoms/availabilitylist/AvailabilityList.tsx
supalarry: Then, I would create `atoms/availability-list/components` folder where to put `EmptyScreen.tsx`, `Form.tsx` etc. and if some of those need to be split into smaller parts to improve readability then create a child folder, for example, `atoms/availability-list/components/form` where `Form.tsx` then becomes `index.tsx`

packages/atoms/availabilitylist/ScheduleListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+  DropdownMenuItem,
+} from "@/components/ui/dropdown-menu";
+import { Toaster } from "@/components/ui/toaster";
+import { useToast } from "@/components/ui/use-toast";
+import { Globe, MoreHorizontal, Star, Copy, Trash } from "lucide-react";
+import { Fragment } from "react";
+
+import { availabilityAsString } from "@calcom/lib/availability";
+
+type Schedule = {
```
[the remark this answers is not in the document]
supalarry: Also, I would use `userId?: number` as optional property instead of expecting it to be `null` and for other such properties too.

packages/atoms/availabilitylist/ScheduleListItem.tsx
Ryukemeister: My bad really need to focus on namings from next time on 😅

packages/atoms/availabilitylist/ScheduleListItem.tsx
Ryukemeister: @supalarry true the dropdown is a bit overwhelming so that can go into a separate `Controls` component but is there really a need to create a `TimeZone` component? I don't think that part contains enough code to go into a separate component for itself

packages/atoms/availability-list/index.tsx
```diff
@@ -14,13 +14,13 @@ export type Schedule = {
     id: number;
     startTime: Date;
     endTime: Date;
-    userId?: number;
-    eventTypeId?: number;
-    date?: Date;
+    userId: number | null;
```
Ryukemeister, line 17: @supalarry the availability object in type `Schedule` is expected to have a value and that's why setting it to optional gives a type error since it will be `value | undefined` that way instead of `value | null`

supalarry: Interesting, I checked `Availability` table locally and indeed `userId` can be `null` - i also thought that it must always be present.

Ryukemeister: > Interesting, I checked `Availability` table locally and indeed `userId` can be `null` - i also thought that it must always be present.

I tried passing the schedule down to a method inside a function and thats when I started getting type errors, I think maybe the components or methods down to which it is being passed expect its value to never be undefined

supalarry: In `schema.prisma` Availability userId type is `Int?` so if userId is not passed it will be null in database

Ryukemeister: Yeah that might be the case what I'm trying to say is the methods which take in `schedule.availability` always expect a value (null included) and as a result I get type errors when trying to make `availability` optional

supalarry: All good - understood!

supalarry (changes requested):
Love the clean folder structure - good job! I requested changes because we need to use `useEffect` hook but it's nothing major.

packages/atoms/availability-list/components/controls/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+        </DropdownMenuTrigger>
+        <DropdownMenuContent>
+          {!schedule.isDefault && (
+            <DropdownMenuItem
+              onClick={() => {
+                handleSetDefault();
+              }}
+              className="min-w-40 focus:ring-mute min-w-40 focus:ring-muted">
+              <Star />
+              Set as default
```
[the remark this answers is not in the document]
Ryukemeister: I don't think we've arrived at a solution on how to handle i18n for atoms just yet (still up for discussion in the `Connect` button [PR](https://github.com/calcom/cal.com/pull/11605)). So for the time being we only accept a translated string as props and then like you mentioned internationalization can be taken care of in a follow up PR.

PR #11783, opened 2023-10-10
feat: Add rate limting to API

sean-brydon:
Adds upstash rate limiting to API based on api key


zomars (approved):
LGTM. Do we have tests to check rate limiting works as expected? If not, can we follow up with some? 

Nice work @sean-brydon 🙏🏽

PR #11789, opened 2023-10-10
chore: Refactors Destination calendar API endpoint with a bunch of fixes

alishaz-polymath:
## What does this PR do?

- [x] Refactors/restructures destination calendar endpoint
- [x] Fixes the validation in destination calendar to allow optional fields of `eventTypeId` and `bookingId`.
- [x] Validation for eventTypeId 

## Requirement/Documentation

- NA

## Type of change

## How should this be tested?

- [ ] API call to destination calendars endpoint [REF](https://cal.com/docs/enterprise-features/api/api-reference/destination-calendars#create-a-new-destination-calendar)

## Mandatory Tasks


apps/api/lib/validations/destination-calendar.ts
```diff
@@ -14,8 +14,8 @@ const schemaDestinationCalendarCreateParams = z
   .object({
     integration: z.string(),
     externalId: z.string(),
-    eventTypeId: z.number(),
-    bookingId: z.number(),
+    eventTypeId: z.number().optional(),
+    bookingId: z.number().optional(),
```
alishaz-polymath, line 18: reflects actual schema
image

alishaz-polymath:
Also, probably a good idea to revisit this endpoint as a follow up for improvements so that it's consistent with the rest of the endpoints in structure.

sean-brydon (dismissed):

apps/api/lib/validations/destination-calendar.ts
```diff
@@ -14,8 +14,8 @@ const schemaDestinationCalendarCreateParams = z
...
     userId: z.number(),
```
alishaz-polymath, line 19:
```suggestion
    userId: z.number().optional(),
```

alishaz-polymath: This should also be optional I think, or we wouldn't be able to have event type only destination calendars

alishaz-polymath (commented):
⚠️ **DO NOT MERGE YET** ⚠️

apps/api/lib/validations/destination-calendar.ts
alishaz-polymath: Also, need to validate the eventType belongs to the user making the request

apps/api/pages/api/destination-calendars/[id]/_auth-middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+  const destinationCalendar = await prisma.destinationCalendar.findFirst({
+    where: {
+      AND: [
+        { id },
+        {
+          OR: [{ userId }, { eventTypeId: { in: userEventTypeIds } }],
+        },
+      ],
+    },
+  });
```
Comment (alishaz-polymath) [inline apps/api/pages/api/destination-calendars/[id]/_auth-middleware.ts:27]: Since destination calendar can be identified either by `eventTypeId` or `userId`, we check the ownership of the eventType for the user here as well. We can follow up with another PR to check for team event types as well in the near future

ThyMinimalDev:
> Also, probably a good idea to revisit this endpoint as a follow up for improvements so that it's consistent with the rest of the endpoints in structure.

super clean pr, easy to read and understand 👍

zomars (approved):
Loving the refactor @alishaz-polymath 

Will it make sense to follow up with some unit testing? Something basic like if this input the expect this output?

alishaz-polymath:
> Loving the refactor @alishaz-polymath
> 
> Will it make sense to follow up with some unit testing? Something basic like if this input the expect this output?

Definitely, I think it makes sense for our deliverability & stability aim 🙏 
I'll line up a follow up for this.

zomars:
API build is failing. Pushed a fix in here #11867

apps/api/pages/api/destination-calendars/[id]/_auth-middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { id } = schemaQueryIdParseInt.parse(req.query);
+  if (isAdmin) return;
+  const userEventTypes = await prisma.eventType.findMany({
+    where: { userId },
+    select: { id: true },
+  });
+
+  const userEventTypeIds = userEventTypes.map((eventType) => eventType.id);
+
+  const destinationCalendar = await prisma.destinationCalendar.findFirst({
```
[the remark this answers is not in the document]
Comment (alishaz-polymath) [reply apps/api/pages/api/destination-calendars/[id]/_auth-middleware.ts:18]: Yeah, I like that. Perhaps I can do that in the follow up where I add the tests? Definitely makes sense, nice catch @ThyMinimalDev 🙏

PR #11793, opened 2023-10-10
feat: Add & Edit event-type children via api to allow Managed Events

alishaz-polymath:
## What does this PR do?

This PR follows up on #11714 and allows add and edit of children via API for Managed Events.

## Requirement/Documentation

- NA

## Type of change

## How should this be tested?

- API calls to event-type endpoint with Children array of structure:
```
{
  id: <eventTypeId>,
  userId: <userId>,
}
```
Should now connect the Managed Event with the Child in `children` field

## Mandatory Tasks

### DIFF
--- a/apps/api/pages/api/event-types/[id]/_patch.ts
+++ b/apps/api/pages/api/event-types/[id]/_patch.ts
@@ -209,15 +209,23 @@ export async function patchHandler(req: NextApiRequest) {
     hosts = [],
     bookingLimits,
     durationLimits,
-    /** FIXME: Updating event-type children from API not supported for now  */
-    children: _,
+    children: parsedChildren,
     ...parsedBody
   } = schemaEventTypeEditBodyParams.parse(body);
+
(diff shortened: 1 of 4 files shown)


alishaz-polymath:
Upon further investigation, this isn't needed in the API as it doesn't fit in the flow.
**The current flow and how a managed event is created via API:**

1. Create a Team event type with `schedulingType` = `MANAGED` & metadata containing something along the lines of:
```
{
  "config": {
    "useHostSchedulesForTeamEvent": false
  },
  "managedEventConfig": {
    "unlockedFields": {
      "locations": true,
      "scheduleId": true,
      "destinationCalendar": true
    }
  }
}
```
2. Create a child event by sending an individual (user) event type query with the same details as the Managed event except without a `team_id`, or `scheduling_type`. Make sure to set `parentId` to the `id` of the event type created in step 1.

The Parent event type should now have the children connected to it. Thus, this PR is not needed.

PR #11794, opened 2023-10-10
fix: Remove credentials and some ts fixes

emrysal:
## What does this PR do?

Even though authenticated, we shouldn't add additional (sensitive) data we do not use, so this strips out the credentials bit from the locationOptions.


zomars (approved):
LGTM

PR #11795, opened 2023-10-10
fix: Update slug for zoom app

alishaz-polymath:
## What does this PR do?

Updates the slug as it is supposed to be
From `zoomvideo` -> `zoom`

## Requirement/Documentation

image

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
+++ b/packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
@@ -91,7 +91,7 @@ const zoomAuth = (credential: CredentialPayload) => {
             grant_type: "refresh_token",
           }),
         }),
-      "zoomvideo",
+      "zoom",
       credential.userId
     );
 


PeerRich (dismissed):
this will cause sideeffects.

Udit-takkar:
@PeerRich i think we should wait for @joeauyeung's reply before we close this.

refreshOAuthTokens requires appSlug in the second parameter which should have been "zoom" not "zoomvideo" which is the name of directory but not the appSlug

joeauyeung:
@Udit-takkar is right. This should be the app slug. I think we can improve on this PR by instead of hard coding "zoom" we can use the `app.slug` directly.

@PeerRich there won't be any side effects in our app. The app slug is being passed into `refreshOAuthToken` which only uses the slug when a self-hoster wants to sync app credentials.

PeerRich (approved):

PeerRich:
> I think we can improve on this PR by instead of hard coding "zoom" we can use the app.slug directly.

lets do a follow up PR

PR #11799, opened 2023-10-10
fix: fix position of arrow button and add animation

anikdhabal:
## What does this PR do?

 fix position of arrow button and add animation in the button inside advanced setting.

Before:-

Screenshot 2023-10-10 224715


sean-brydon (approved):
Great job thank you!

PR #11804, opened 2023-10-10
test: add orgs e2e tests

leog:
## What does this PR do?

Contributing to stability and reliability for orgs, this PR adds orgs e2e tests.

Also, by making them, found a bug on signing up with a token without a username, was fixed, 

## Type of change

- [X] Test (making the product more stable and reliable)

## How should this be tested?

Run `yarn playwright test organization-creation.e2e.ts` and `yarn playwright test organization-invitation.e2e.ts`.


apps/web/pages/signup.tsx
```diff
@@ -57,9 +57,10 @@ export default function Signup({ prepopulateFormValues, token, orgSlug, orgAutoA
   const telemetry = useTelemetry();
   const { t, i18n } = useLocale();
   const flags = useFlagMap();
+  const isOrgInviteByLink = orgSlug && !prepopulateFormValues?.username;
   const methods = useForm<FormValues>({
     mode: "onChange",
-    resolver: zodResolver(signupSchema),
+    resolver: zodResolver(isOrgInviteByLink ? signupSchema.omit({ username: true }) : signupSchema),
```
leog, line 63: Fixing bug that validated existence of a username in a org context when it's not being inputed by the user as it is inferred.

apps/web/pages/signup.tsx
```diff
@@ -74,9 +75,10 @@ export default function Signup({ prepopulateFormValues, token, orgSlug, orgAutoA
     }
   };
 
-  const isOrgInviteByLink = orgSlug && !prepopulateFormValues?.username;
-
   const signUp: SubmitHandler<FormValues> = async (data) => {
+    if (data.username === undefined && isOrgInviteByLink && orgAutoAcceptEmail) {
+      data = { ...data, username: getOrgUsernameFromEmail(methods.getValues().email, orgAutoAcceptEmail) };
+    }
```
leog, line 81: This username that is inferred was not being added to the form data correctly before sending to the server

apps/web/pages/signup.tsx
```diff
@@ -143,12 +145,6 @@ export default function Signup({ prepopulateFormValues, token, orgSlug, orgAutoA
                     methods.clearErrors("apiError");
                   }
 
-                  if (methods.getValues().username === undefined && isOrgInviteByLink && orgAutoAcceptEmail) {
-                    methods.setValue(
-                      "username",
-                      getOrgUsernameFromEmail(methods.getValues().email, orgAutoAcceptEmail)
-                    );
-                  }
```
leog, line 151: This code didn't work

apps/web/playwright/fixtures/users.ts
```diff
@@ -77,11 +77,13 @@ const createTeamAndAddUser = async (
...
     isUnpublished,
     isOrg,
+    isOrgVerified,
     hasSubteam,
   }: {
-    user: { id: number; username: string | null; role?: MembershipRole };
+    user: { id: number; email: string; username: string | null; role?: MembershipRole };
     isUnpublished?: boolean;
     isOrg?: boolean;
+    isOrgVerified?: boolean;
```
leog, line 86: Making sure you can create a verified org for a test user as well as being able to define its email address

apps/web/playwright/fixtures/users.ts
```diff
@@ -92,7 +94,14 @@ const createTeamAndAddUser = async (
...
     ...(isUnpublished ? { requestedSlug: slug } : {}),
-    ...(isOrg ? { isOrganization: true } : {}),
+    ...(isOrg
+      ? {
+          isOrganization: true,
+          isOrganizationVerified: !!isOrgVerified,
+          orgAutoAcceptEmail: user.email.split("@")[1],
+          isOrganizationConfigured: false,
+        }
+      : {}),
```
leog, line 104: Missing metadata for the org to be able to test functionality accordingly

apps/web/playwright/lib/testUtils.ts
```diff
@@ -274,3 +277,43 @@ export async function createUserWithSeatedEventAndAttendees(
...
+  await page.waitForTimeout(2000);
+  const receivedEmails = await getEmailsReceivedByUser({ emails, userEmail });
+  expect(receivedEmails?.total).toBe(1);
+  const [firstReceivedEmail] = (receivedEmails as Messages).items;
+  expect(firstReceivedEmail.subject).toBe(subject);
+  if (!returnLink) return;
+  const dom = new JSDOM(firstReceivedEmail.html);
+  const anchor = dom.window.document.querySelector(`a[href*="${returnLink}"]`);
+  return anchor?.getAttribute("href");
+}
```
leog, line 300: This function lets us check for email being sent when `E2E_TEST_MAILHOG_ENABLED` environment variable is set. Also returns a link within its body if required.

apps/web/playwright/lib/testUtils.ts
```diff
@@ -274,3 +277,43 @@ export async function createUserWithSeatedEventAndAttendees(
...
+
+export async function getInviteLinkFromConsole(page: Page): Promise<string> {
+  return new Promise((resolve) => {
+    page.on("console", (msg) => {
+      if (msg.text().indexOf("signup?token") > -1) {
+        resolve(msg.text());
+      }
+    });
+  });
+}
```
leog, line 310: When testing we needed a way to access the invite link that is copied to the clipboard, this extracts it from the console.

apps/web/playwright/lib/testUtils.ts
```diff
@@ -274,3 +277,43 @@ export async function createUserWithSeatedEventAndAttendees(
...
+
+export function generateTotpCode(email: string) {
+  const secret = createHash("md5")
+    .update(email + process.env.CALENDSO_ENCRYPTION_KEY)
+    .digest("hex");
+
+  totp.options = { step: 90 };
+  return totp.generate(secret);
+}
```
leog, line 319: To be able to use it in the org verification step, can help in other TOTP verification code situations that exists in the app.

packages/features/ee/teams/components/MemberInvitationModal.tsx
```diff
@@ -94,7 +94,14 @@ export default function MemberInvitationModal(props: MemberInvitationModalProps)
...
     const inviteLink =
       isOrgInvite || (props?.orgMembers && props.orgMembers?.length > 0) ? orgInviteLink : teamInviteLink;
-    await navigator.clipboard.writeText(inviteLink);
+
+    try {
+      await navigator.clipboard.writeText(inviteLink);
+      console.log(inviteLink);
+    } catch (e) {
+      console.log(inviteLink);
+    }
```
leog, line 103: Added try/catch to avoid the error we get when using a baseURL that is not localhost such as app.cal.local, and in either case logging the invite link to the console to be able to use it in tests

packages/features/users/components/UserTable/UserListTable.tsx
```diff
@@ -204,12 +204,15 @@ export function UserListTable() {
...
+          const { teams, accepted, email } = row.original;
           // TODO: Implement click to filter
           return (
             <div className="flex h-full flex-wrap items-center gap-2">
               {accepted ? null : (
-                <Badge variant="red" className="text-xs">
+                <Badge
+                  variant="red"
+                  className="text-xs"
+                  data-testid={`email-${email.replace("@", "")}-pending`}>
```
leog, line 215: To be able to check if the invited user is verified either after they sign up or because the org is verified in which case the user is automatically accepted.

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -175,7 +175,7 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
 
     return { user: { ...createOwnerOrg, password } };
   } else {
-    if (!IS_PRODUCTION) return { checked: true };
+    if (!IS_PRODUCTION && !process.env.NEXT_PUBLIC_IS_E2E) return { checked: true };
```
leog, line 178: Making sure we can test TOTP code in a E2E environment

package.json
```diff
@@ -78,6 +78,7 @@
     "@playwright/test": "^1.31.2",
     "@snaplet/copycat": "^0.3.0",
     "@testing-library/jest-dom": "^5.16.5",
+    "@types/jsdom": "^21.1.3",
```
zomars, line 81: Was this missing?

packages/features/ee/teams/components/MemberInvitationModal.tsx
zomars: We should make this explicit. By reading this alone, no one can infer that removing these console logs will cause tests to fail.

zomars (changes requested):
Appreciate the effort @leog saw some nits to address first. Also It seems like some tests were failing. Blocking to address those things first. 🙏🏽

package.json
leog: Yeah, it seems so

leog:
@zomars everything working fine locally, seems like the errors are related to emails not coming through and a verification code not being valid. The former might be something related to mailhog and the latter related to the `CALENDSO_ENCRYPTION_KEY`.

apps/web/pages/signup.tsx
hariombalhara: I couldn't replicate the bug. In my testing username was automatically computed and thus was available.

Tried with both an email that matches and not matches autoAcceptDomain

PR #11807, opened 2023-10-10
fix: fetching booking with null responses from API

### LINKED ISSUE #11626: Issue fetching bookings with the Cal API

### Issue Summary

### Steps to Reproduce

1. Send a get request to /bookings (https://api.cal.com/bookings?apiKey=X)
2. Should see: {"message":"invalid_type in 'responses': Expected object, received null"}

### Actual Results

- It takes a while and then errors.

### Expected Results

- Should return a list of bookings

### Notes

- We think it is likely an issue with zod schema or an upstream issue with prisma / prisma accelerate!

DexterStorey:
## What does this PR do?

Fixes [#11626](https://github.com/calcom/cal.com/issues/11626)

## Type of change

## How should this be tested?

- To reproduce the error, follow the steps in the issue (create a null response booking and fetch bookings)
- Without this patch you should get `invalid_type in 'responses': Expected object, received null`
- With this patch, you should get your bookings

## Mandatory Tasks

### DIFF
--- a/apps/api/lib/validations/booking.ts
+++ b/apps/api/lib/validations/booking.ts
@@ -58,7 +58,7 @@ export const schemaBookingReadPublic = Booking.extend({
       })
     )
     .optional(),
-  responses: z.record(z.any()),
+  responses: z.record(z.any()).nullable(),
 }).pick({
   id: true,
   userId: true,


zomars (approved):
Nice catch. Thank you for your contribution 🙏

PR #11808, opened 2023-10-10
fix: newLocale fallback to 'en' locale

emrysal:
## What does this PR do?

If newLocale is falsy we default to "en"

### DIFF
--- a/apps/web/pages/_document.tsx
+++ b/apps/web/pages/_document.tsx
@@ -46,7 +46,8 @@ class MyDocument extends Document<Props> {
   }
 
   render() {
-    const { isEmbed, embedColorScheme, newLocale } = this.props;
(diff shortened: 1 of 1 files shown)


zomars (approved):
Odd that all other fallbacks aren't being picked up. cc @grzpab for context

apps/web/pages/_document.tsx
```diff
@@ -46,7 +46,8 @@ class MyDocument extends Document<Props> {
   }
 
   render() {
-    const { isEmbed, embedColorScheme, newLocale } = this.props;
+    const { isEmbed, embedColorScheme } = this.props;
+    const newLocale = this.props.newLocale || "en";
```
emrysal, line 50:
```suggestion
    const newLocale = this.props.newLocale ?? "en";
```

PR #11832, opened 2023-10-11
fix: Change sidebar skeleton width to w-full, better for i18n

emrysal:
## What does this PR do?

Spotted & fixed. Before and after:

image
image


PeerRich:
i like the idea of making the skeleton as wide as the text, but lets just left align it no?

PeerRich (approved):
sad to see them go but i guess that's better

PR #11852, opened 2023-10-12
test: Tests the order of middleware in API

sean-brydon:
Adds a test to ensure the order is correct when using withMiddleware


apps/api/lib/helpers/verifyApiKey.ts
```diff
@@ -4,7 +4,7 @@ import { hashAPIKey } from "@calcom/features/ee/api-keys/lib/apiKeys";
 import checkLicense from "@calcom/features/ee/common/server/checkLicense";
 import { IS_PRODUCTION } from "@calcom/lib/constants";
 
-import { isAdminGuard } from "~/lib/utils/isAdmin";
+import { isAdminGuard } from "../utils/isAdmin";
```
sean-brydon, line 7: Vitest isnt configured to handle ~ aliases when in API - was outside the scope of this PR to fix

apps/api/test/lib/middleware/withMiddleware.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, vi, it, expect, afterEach } from "vitest";
+
+import { middlewareOrder } from "../../../lib/helpers/withMiddleware";
+
+afterEach(() => {
+  vi.resetAllMocks();
+});
+
+// Not sure if there is much point testing this order is actually applied via an integration test:
+// It is tested internally https://github.com/htunnicliff/next-api-middleware/blob/368b12aa30e79f4bd7cfe7aacc18da263cc3de2f/lib/label.spec.ts#L62
```
sean-brydon, line 10: I don think we should re-test that label(...) actually applies the middleware order correctly to any handler it calls

apps/api/lib/helpers/withMiddleware.ts
```diff
@@ -41,7 +44,8 @@ const withMiddleware = label(
...
     "rateLimitApiKey",
     "addRequestId",
-  ] // <-- Provide a list of middleware to call automatically
-);
+  ] as Middleware[]; // <-- Provide a list of middleware to call automatically
+
+const withMiddleware = label(middleware, middlewareOrder);
 
-export { withMiddleware };
+export { withMiddleware, middleware, middlewareOrder };
```
keithwillcode, line 51: Do we actually need to export the `middlewareOrder`? Are we not able to take the `withMiddleware` object and assert things about it in the tests to ensure the order is correct?

apps/api/lib/helpers/withMiddleware.ts
```diff
@@ -41,7 +44,8 @@ const withMiddleware = label(
     "verifyApiKey",
     "rateLimitApiKey",
     "addRequestId",
-  ] // <-- Provide a list of middleware to call automatically
-);
+  ] as Middleware[]; // <-- Provide a list of middleware to call automatically
```
keithwillcode, line 47: imho this comment is useless because the code explains itself

apps/api/lib/helpers/withMiddleware.ts
keithwillcode: @sean-brydon and I talked about this and it's not straightforward. We are sticking with this approach for now.

keithwillcode (approved):

PR #11831, opened 2023-10-11
fix: use session locale when dealing with SSG-ed pages

grzpab:
## What does this PR do?

For the booking page, which is statically generated, we don't run `getInitialProps` under `_document` or `_app`. This way we end up with the default `"en"` locale.

Fixes # (issue)
https://www.loom.com/share/188d9021286c427a8d789bf3644d2495
## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is ab UI/UX design document, please, share it here.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?
1. Go to any page (not bookings) and see if the locale are correct
2. Go to booking and see if the locale are correct. You will see flickering from English to your locale.
3. Go to page X and move to booking. Locale should remain the same.
4. Go to bookings and move to page X. Locale should remain the same.
5. Change the language and navigate to any page (booking or not). Locale should be changed.

## Mandatory Tasks

- [ ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)
- My code doesn't follow the style guidelines of this project
- I haven't commented my code, particularly in hard-to-understand areas
- I haven't checked if my PR needs changes to the documentation
- I haven't checked if my changes generate no new warnings
- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #11853, opened 2023-10-12
chore: add tests for locale and directionality; set html.dir when the session locale is available

grzpab:
## What does this PR do?

1. Adds tests for the locale and directionality inferring for signed-in and not-signed-in users
2. Loads `@calcom/features/auth/lib/getLocale` asynchronously in `_app` and `_document` files when in the server context; this should reduce the bundle size
3. The bugfix https://github.com/calcom/cal.com/pull/11831 did not change the `html.dir` and `html.lang` attributes, which is fixed and tested in this PR. This could be visible for users that use e.g. the Arabic language and would visit bookings, as their language uses RtL order instead of the default LtR.

Fixes # (issue)
The bugfix https://github.com/calcom/cal.com/pull/11831 did not change the `html.dir` and `html.lang` attributes, which is fixed and tested in this PR. 

## Type of change

## How should this be tested?
1. The users that are not signed in should see the language stemming from the locale in the accept-language header
2. The logged in users should see the locale as it is set in the language config in their profiles. Moving between any pages should not result in different translations or different directions of text.

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/lib/app-providers.tsx
+++ b/apps/web/lib/app-providers.tsx
@@ -8,6 +8,7 @@ import { ThemeProvider } from "next-themes";
 import type { AppProps as NextAppProps, AppProps as NextJsAppProps } from "next/app";
 import type { ParsedUrlQuery } from "querystring";
 import type { PropsWithChildren, ReactNode } from "react";
+import { useEffect } from "react";
 
 import { OrgBrandingProvider } from "@calcom/features/ee/organizations/context/provider";
 import DynamicHelpscoutProvider from "@calcom/features/ee/support/lib/helpscout/providerDynamic";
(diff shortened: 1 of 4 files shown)


zomars:
@PeerRich locale testing seems like high priority to me. Will update if that's fine.

zomars (approved):
Tested and LGTM. Thank you for your contribution 🙏

apps/web/lib/app-providers.tsx
```diff
@@ -75,13 +76,30 @@ const CustomI18nextProvider = (props: AppPropsWithoutNonce) => {
...
 
+  useEffect(() => {
+    window.document.documentElement.lang = locale;
+
+    let direction = window.document.dir || "ltr";
+
+    try {
+      const intlLocale = new Intl.Locale(locale);
+      // @ts-expect-error INFO: Typescript does not know about the Intl.Locale textInfo attribute
+      direction = intlLocale.textInfo?.direction;
```
zomars, line 87: This returns undefined on browser/mobile. It wasn't meant to be run on browser.

zomars: cc @grzpab

PR #11859, opened 2023-10-12
improve: hard coded zoom to app.slug

### LINKED ISSUE #11856: chore: use app.slug not hard coded "zoom"
> I think we can improve on this PR by instead of hard coding "zoom" we can use the app.slug directly.

lets do a follow up PR

_Originally posted by @PeerRich in https://github.com/calcom/cal.com/issues/11795#issuecomment-1759551870_

MeenuyD:
## What does this PR do?
Use app.slug dynamically instead of hardcoding 'zoom' for improved flexibility.

Fixes #11856


packages/app-store/zoomvideo/lib/VideoApiAdapter.ts
```diff
@@ -91,7 +91,8 @@ const zoomAuth = (credential: CredentialPayload) => {
             grant_type: "refresh_token",
           }),
         }),
-      "zoom",
+
+      app.slug,
```
PeerRich, line 95:
```suggestion
      app.slug,
```

PeerRich (dismissed):

joeauyeung (changes requested):
Thanks for your contribution @MeenuyD.

`app` does not exist in the `VideoApiAdapter` so this throws an error. We should find another way to dynamically get the app slug. We should also do this for other apps as well.

MeenuyD:
Ok I will try

MeenuyD:
Can you tell me how can we do

Siddharth-2382:
> Can you tell me how can we do

add `import metadata from "zoomvideo/_metadata";` at the top.
replace `app.slug` with `metadata.slug`

PeerRich:
already fixed

PR #11861, opened 2023-10-12
Updated the common.js of deutsch with english common.js

### LINKED ISSUE #11850: {{eventTypeTitle}} variable missing in translation
### Issue Summary

When saving and updating an event type, the {{eventTypeTitle}} variable is missing in the toast
Screenshot 2023-10-12 at 10 55 37

MeenuyD:
## What does this PR do?
fix the When saving and updating an event type, the {{eventTypeTitle}} variable is missing in the toast

Fixes #11850


PeerRich (changes requested):
see comment

PeerRich:
we should not remove the variable instead we should make the variable work

PR #11862, opened 2023-10-12
fix: Add `refresh_token` placeholder when parsing refresh token responses

joeauyeung:
## What does this PR do?

When fetching new access tokens from OAuth apps not every response included a new refresh token. As a part of credential syncing a refresh token placeholder was added assuming that normal access token responses would already contain a refresh token. This PR only adds the placeholder if credential syncing is enabled.

This issue was first noticed in Outlook but should apply to other apps.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Install Outlook
- In the DB set the expiry date to in the past to force a refresh
- In the token response ensure that a refresh token was not passed
- The refresh token placeholder should not have been added if a refresh token was not present in the MS token response

## Mandatory Tasks

## Checklist


zomars (approved):
Approving. But will leave some things that need to be addressed in a follow up.

This "feels" like constantly patching. We should strive to achieve syncing without having to fiddle with credentials. Avoid introducing more edge cases to be aware of.

If this is absolutely necessary. Then we need tests. Undoubtedly.

Udit-takkar (approved):
LGTM

Issue #11730, opened 2023-10-06
[CAL-2590] Show link location in /bookings


PeerRich:
show link to join a call in the /booking overview

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2590](https://linear.app/calcom/issue/CAL-2590/show-link-location-in-bookings)</sub>

PeerRich:
how about this? 
![CleanShot 2023-10-06 at 13 44 42@2x](https://github.com/calcom/cal.com/assets/8019099/745e637f-003a-48f6-bd48-e46336898a45)

PeerRich:
alternatively, show the name of the app: 
![CleanShot 2023-10-06 at 13 49 03@2x](https://github.com/calcom/cal.com/assets/8019099/bc75d80e-a1ee-439e-87ca-7104c40056c7)

Siddharth-2382:
Meeting Link seems to better as it is more of general term and conveys the message about what it does clearly. Also if you could finalise the design, I'd have liked to work on this issue but seems like you have self assigned it🙂

SeekerHub:
If the issue is opened I did like to work on it

Tan28bha:
pr

caioluis:
What is the state of this issue in fact? I've seen that multiple closing/reopening happened.

ptdatta:
Is this completed?

asadath1395:
@PeerRich I think this is done here https://github.com/calcom/cal.com/pull/11866, can we close this? Thanks

PR #11865, opened 2023-10-12
feat: show location link in /bookings

CntrlX:
Solves the below issue 
[CAL-2590] Show link location in /bookings #11730

Has the images as same as given in the latest design

/claim #11730


PeerRich (changes requested):
we should not hard code any logos or names. instead we should use the logo and names from the app store

PR #11877, opened 2023-10-13
fix: by-pass global unubsribe for from transactional emails

CarinaWolli:
## What does this PR do?

Adds headers to nodemailer to ignore Unsubscribes for Transactional Mails from Sendgrid.

### DIFF
--- a/packages/emails/templates/_base-email.ts
+++ b/packages/emails/templates/_base-email.ts
@@ -48,6 +48,17 @@ export default class BaseEmail {
     const payload = this.getNodeMailerPayload();
     const parseSubject = z.string().safeParse(payload?.subject);
     const payloadWithUnEscapedSubject = {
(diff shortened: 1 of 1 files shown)


PeerRich (approved):
nice!

packages/emails/templates/_base-email.ts
```diff
@@ -48,6 +48,17 @@ export default class BaseEmail {
     const payload = this.getNodeMailerPayload();
     const parseSubject = z.string().safeParse(payload?.subject);
     const payloadWithUnEscapedSubject = {
+      headers: {
```
keithwillcode, line 51: This implementation is specific to SendGrid, right?

If so, we should try to move this functionality to some sort of higher level function/layer that is specific to SendGrid and isn’t in the BaseEmail

PR #11883, opened 2023-10-13
fix: Remove Initial Credential Parsing For GCal

joeauyeung:
## What does this PR do?

This PR removes the initial Zod parsing of GCal credentials. This was causing problems for credentials created by 3rd party platforms. 

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Setup credential syncing by filling in the correct .env variables and setting a webhook that returns the access token for GCal
- Connect GCal to get the access token for the webhook
- Force a refresh of the token to call the webhook
- Calendars should be listed correctly
- When creating a new booking, the event should be created on the calendar

## Mandatory Tasks

## Checklist


packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -73,7 +74,9 @@ export default class GoogleCalendarService implements Calendar {
   }
 
   private googleAuth = (credential: CredentialPayload) => {
-    const googleCredentials = googleCredentialSchema.parse(credential.key);
```
joeauyeung, line 76: I believe we don't need this parsing step as we already parse the token response using this schema or the credential syncing schema.

zomars (changes requested):
Big nope from me. This is just errors waiting to happen in runtime.

PR #11884, opened 2023-10-13
fix: Overlaying of Override button by toast on mobile.

### LINKED ISSUE #11814: Override button hidden by toast on mobile
image

The following is what it looks like when adding 2 overrides in a short timeframe, the toast hides the button so i cant add more overrides until the toast goes away.

I'm on an iPhone 12 if screen size is of relevance.

---

Also, huge props to how this flow has been improved. It used to be quite the pain to add overrides for multiple days but now it's much less friction as the modal stays open and i can just click the dates i want much faster. Thanks for that 🙏

bhanu-io:
## What does this PR do?

Fixes #11814

Environment : Production

## Type of change

- Bug fix

## How should this be tested?

- This is applicable on mobile screen.
- Upon multiple override requests, toast used to hide the Add Override Button for too long.
- Now make an add override request, the  toasts now stays on for lesser duration.

https://github.com/calcom/cal.com/assets/122738012/55b29115-ce17-445a-80d5-1327f30ebc18

### DIFF
--- a/packages/features/schedules/components/DateOverrideInputDialog.tsx
+++ b/packages/features/schedules/components/DateOverrideInputDialog.tsx
@@ -179,7 +179,7 @@ const DateOverrideForm = ({
                 color="primary"
                 type="submit"
                 onClick={() => {
-                  showToast(t("date_successfully_added"), "success");
+                  showToast(t("date_successfully_added"), "success", 500);
                 }}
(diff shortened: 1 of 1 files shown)


emrysal (approved):
Feels like a workaround to me, but I guess it works 🙏

PR #11885, opened 2023-10-13
fix: RangeError: Incorrect locale information

Udit-takkar:
## What does this PR do?

Fixes  https://threads.com/34541951212?s=qxMqBH7J4vmD83KCfe7Qcf

Screenshot 2023-10-13 at 11 08 21 PM

https://stackoverflow.com/a/76548248

I tested this by turning on the debug mode locally and tried booking and rescheduling

## Type of change


zomars:
We're going with this approach @Udit-takkar #11833

PR #11890, opened 2023-10-14
fix: rescheduleReason not shown when booking status is pending

### LINKED ISSUE #11784: [CAL-2601] When a pending booking is rescheduled, it doesn't show the rescheduleReason on the booking success page - I think it should be shown
[https://github.com/calcom/cal.com/blob/2faf24fb986c7a323ee9aa1ff9387bcc14033439/apps/web/pages/booking/\[uid\].tsx#L401](https://github.com/calcom/cal.com/blob/2faf24fb986c7a323ee9aa1ff9387bcc14033439/apps/web/pages/booking/%5Buid%5D.tsx#L401)

This check is what prevents it from showing it for a rescheduled booking that's in pending state because `reschedule` is true only when booking status is accepted.

Steps would be

1. Create an event that requires confirmation
2. Book that event. The event would go to Pending state
3. Now, reschedule that event and add a reschedule reason
4. Booking would be rescheduled to the new time and status would remain pending
5. Notice that booking success page doesn't show the reschedule reason

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2601](https://linear.app/calcom/issue/CAL-2601/when-a-pending-booking-is-rescheduled-it-doesnt-show-the)</sub>

Vihaan-19:
## What does this PR do?
As mentioned in the issue ,`reschedule` was true only when the booking status was 'ACCEPTED', as required reschedule is now true for bookings with status 'PENDING' as well.

Fixes #11784

## Type of change

## Mandatory Tasks

## Checklist


PeerRich (dismissed):

CarinaWolli:
@Vihaan-19 (or @hariombalhara), I am unsure how to reproduce the initial bug. What are the steps?

apps/web/pages/booking/[uid].tsx
```diff
@@ -128,7 +128,8 @@ export default function Success(props: SuccessProps) {
   )?.videoCallUrl;
 
   const status = props.bookingInfo?.status;
-  const reschedule = props.bookingInfo.status === BookingStatus.ACCEPTED;
+  const reschedule =
```
Comment (hariombalhara) [inline apps/web/pages/booking/[uid].tsx:131]: I don't know why this variable's name is reschedule. Could we rename it better? or maybe see how it's used exactly and improve the logic.

hariombalhara (changes requested):
Thanks @Vihaan-19 🙏 Left a comment

hariombalhara (changes requested):
@Vihaan-19 You re-requested the review, but I see nothing changed in the PR.

apps/web/pages/booking/[uid].tsx
Comment (Vihaan-19) [reply apps/web/pages/booking/[uid].tsx:131]: reschedule is a bool variable which checks the status of reschedule. It is later user to check if a valid reschedule request was made. It is valid, if the booking status was either pending or accepted.
```const reschedule=props.bookingInfo.status === BookingStatus.ACCEPTED || props.bookingInfo.status === BookingStatus.PENDING;```
reschedule would be true if the booking status is ```Accepted``` or ```Pending```.
Based on this, is there any improvisation?

PR #11892, opened 2023-10-14
feat: Support moving a user and it's teams to an org as temporary approach

hariombalhara:
## What does this PR do?

Corresponding website PR: 
https://github.com/calcom/website/pull/638

## Type of change

## How should this be tested?
- Have a user within a team with many apps installed and many event-types across teams
- Migrate the user to an org
- Check that all the teams and event-types and apps are accessible.

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/lib/getTemporaryOrgRedirect.ts
+++ b/apps/web/lib/getTemporaryOrgRedirect.ts
@@ -1,0 +1,43 @@
+import logger from "@calcom/lib/logger";
+import { safeStringify } from "@calcom/lib/safeStringify";
+import type { RedirectType } from "@calcom/prisma/client";
+
+const log = logger.getChildLogger({ prefix: ["lib", "getTemporaryOrgRedirect"] });
+export const getTemporaryOrgRedirect = async ({
+  slug,
+  redirectType,
+  eventTypeSlug,
+}: {
+  slug: string;
+  redirectType: RedirectType;
+  eventTypeSlug: string | null;
+}) => {
+  const prisma = (await import("@calcom/prisma")).default;
+  log.debug(
+    `Looking for redirect for`,
+    safeStringify({
+      slug,
+      redirectType,
+      eventTypeSlug,
+    })
+  );
+  // Not supporting dynamic group for now
+  const redirect = await prisma.tempOrgRedirect.findFirst({
+    where: {
+      type: redirectType,
+      from: slug,
+      fromOrgId: 0,
+    },
+  });
+
+  if (redirect) {
+    log.debug(`Redirecting ${slug} to ${redirect.toUrl}`);
+    return {
+      redirect: {
+        permanent: false,
+        destination: eventTypeSlug ? `${redirect.toUrl}/${eventTypeSlug}` : redirect.toUrl,
+      },
+    } as const;
+  }
+  return null;
+};
--- a/apps/web/lib/getTemporaryOrgRedirect.ts
+++ b/apps/web/lib/getTemporaryOrgRedirect.ts
@@ -21,7 +21,6 @@ export const getTemporaryOrgRedirect = async ({
       eventTypeSlug,
     })
   );
-  // Not supporting dynamic group for now
   const redirect = await prisma.tempOrgRedirect.findFirst({
     where: {
       type: redirectType,
--- a/apps/web/lib/getTemporaryOrgRedirect.ts
+++ b/apps/web/lib/getTemporaryOrgRedirect.ts
@@ -22,11 +22,13 @@ export const getTemporaryOrgRedirect = async ({
     })
   );
   // Not supporting dynamic group for now
-  const redirect = await prisma.tempOrgRedirect.findFirst({
+  const redirect = await prisma.tempOrgRedirect.findUnique({
     where: {
-      type: redirectType,
-      from: slug,
-      fromOrgId: 0,
+      from_type_fromOrgId: {
+        type: redirectType,
+        from: slug,
+        fromOrgId: 0,
+      },
     },
   });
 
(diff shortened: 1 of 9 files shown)


apps/web/pages/[user].tsx
```diff
@@ -312,6 +316,18 @@ export const getServerSideProps: GetServerSideProps<UserPageProps> = async (cont
     avatar: `/${user.username}/avatar.png`,
   }));
 
+  if (!isOrgContext) {
+    const redirect = await getTemporaryOrgRedirect({
+      slug: usernameList[0],
```
Comment (hariombalhara) [inline apps/web/pages/[user].tsx:321]: Not supporting redirect for dynamic booking link for now as they would most probably be not hardcoded anywhere.

apps/web/pages/[user]/[type].tsx
```diff
@@ -74,6 +77,20 @@ async function getDynamicGroupPageProps(context: GetServerSidePropsContext) {
     context.params?.orgSlug
   );
 
+  const isOrgContext = currentOrgDomain && isValidOrgDomain;
+
+  if (!isOrgContext) {
+    const redirect = await getTemporaryOrgRedirect({
+      slug: usernames[0],
+      redirectType: RedirectType.User,
+      eventTypeSlug: slug,
```
Comment (hariombalhara) [inline apps/web/pages/[user]/[type].tsx:86]: Redirect for user even-type page.

apps/web/pages/team/[slug].tsx
```diff
@@ -272,13 +275,28 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
...
     slug: slugify(slug ?? ""),
     orgSlug: currentOrgDomain,
     isTeamView: true,
     isOrgView: isValidOrgDomain && context.resolvedUrl === "/",
   });
+
+  if (!isOrgContext && slug) {
+    const redirect = await getTemporaryOrgRedirect({
+      slug: slug,
+      redirectType: RedirectType.Team,
```
Comment (hariombalhara) [inline apps/web/pages/team/[slug].tsx:291]: Redirect for team profile page.

apps/web/pages/team/[slug]/[type].tsx
```diff
@@ -75,6 +78,19 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
     context.req.headers.host ?? "",
     context.params?.orgSlug
   );
+  const isOrgContext = currentOrgDomain && isValidOrgDomain;
+
+  if (!isOrgContext) {
+    const redirect = await getTemporaryOrgRedirect({
+      slug: teamSlug,
+      redirectType: RedirectType.Team,
+      eventTypeSlug: meetingSlug,
```
Comment (hariombalhara) [inline apps/web/pages/team/[slug]/[type].tsx:87]: Redirect for team event.

packages/trpc/server/routers/viewer/organizations/list.handler.ts
```diff
@@ -29,6 +29,13 @@ export const listHandler = async ({ ctx }: ListHandlerInput) => {
     },
   });
 
+  if (!membership) {
```
hariombalhara, line 32: Without membership wouldn't be able to see listing of an orgzniation

apps/web/lib/getTemporaryOrgRedirect.ts
```diff
@@ -0,0 +1,43 @@
...
+  const prisma = (await import("@calcom/prisma")).default;
+  log.debug(
+    `Looking for redirect for`,
+    safeStringify({
+      slug,
+      redirectType,
+      eventTypeSlug,
+    })
+  );
+  // Not supporting dynamic group for now
```
hariombalhara, line 24:
```suggestion
```

apps/web/lib/getTemporaryOrgRedirect.ts
```diff
@@ -0,0 +1,44 @@
...
+  const prisma = (await import("@calcom/prisma")).default;
+  log.debug(
+    `Looking for redirect for`,
+    safeStringify({
+      slug,
+      redirectType,
+      eventTypeSlug,
+    })
+  );
+  const redirect = await prisma.tempOrgRedirect.findUnique({
```
hariombalhara, line 24: It's a temporary table and would be really lightweight. So, It should be quick to query but there would still be DB latency that would be added in querying a booking page.

I feel it's okay to have it for now, till we actually see a significant issue with it(as we will remove it anyway soon when mono-user Org schema is ready)

If we see an issue, our option then is to query this in parallel, with other queries on Booking Page which is possible

apps/web/pages/org/[orgSlug]/[user]/[type]/embed.tsx
```diff
@@ -1,19 +1,7 @@
-import type { GetServerSidePropsContext } from "next";
+import withEmbedSsr from "@lib/withEmbedSsr";
 
 import { getServerSideProps as _getServerSideProps } from "../[type]";
 
 export { default } from "../[type]";
 
-export const getServerSideProps = async (context: GetServerSidePropsContext) => {
```
Comment (hariombalhara) [inline apps/web/pages/org/[orgSlug]/[user]/[type]/embed.tsx:7]: I am not sure why it wasn't using withEmbedSsr, all other embed pages were using that only.

zomars (commented):
Left some comments in the website PR. Also. Shouldn't this live in the web app admin panel instead of the website? (even if it's temporary)

hariombalhara:
Thanks for the review @zomars 🙏

There are 2 reasons why I didn't put into Admin Panel
- It doesn't concern our self-hosters at all because it is something only we are going to need and that too temporarily.
- More importantly, there is an endpoint in there that I don't want to be exposed to people. That endpoint that allow people to join an org by themselves(Though that logic is explicitly limited to isAdmin at the moment) and that could be abused.

zomars (approved):
Concerns were addressed 🙏🏽

PR #11898, opened 2023-10-15
feat: add rate limiting and more error handling to Cal.ai

### LINKED ISSUE #11916: [BUG]: "Undefined" Email Subject if left empty
### Issue Summary
The subject of the email is undefined if the user does not give the subject while chatting with the cal.com ai bot.

### Steps to Reproduce

1. Try to book a meeting using the cal.com AI bot
2. Leave the subject empty and send email
3. Wait for the reply from bot
4. The email subject would be "Re: undefined"

### Actual Results
Screenshot 2023-10-16 at 5 43 20 PM

### Expected Results

- Either the email must be rejected or the bot must set the subject to something meaningful

### Evidence
- Screenshot Added

DexterStorey:
## What does this PR do?

This PR adds:
- Rate Limiting (20 emails / day / user)
Screenshot 2023-10-14 at 10 07 19 PM

- A catch all error for OpenAI rate limit and any other funky action in agent loop that tells user to try again later
Screenshot 2023-10-14 at 10 06 39 PM

Fixes #11916

## Type of change

## How should this be tested?

To test rate limit: send more than 20 emails / day
To test agent loop 500 catch: overspend OpenAI tokens or use a bad OpenAI API key

- Tested rate limiting by sending too many requests
- Confirmed everything else still works with standard tests "Create a booking with...", "What meetings do I have...", etc.

## Mandatory Tasks

## Checklist

- ~I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)~
- ~My code doesn't follow the style guidelines of this project~
- ~I haven't commented my code, particularly in hard-to-understand areas~
- ~I haven't checked if my PR needs changes to the documentation~
- ~I haven't checked if my changes generate no new warnings~


packages/lib/checkRateLimitAndThrowError.ts
```diff
@@ -1,4 +1,4 @@
-import { TRPCError } from "@calcom/trpc";
+import { TRPCError } from "@calcom/trpc/server";
```
DexterStorey, line 1: !!!IMPORTANT!!!

We needed to swap this TRPC import from client side to server side because the AI app uses it server side.
Rate limiting in our use case **Does not work** without this.

We don't know if this causes breaking changes across the rest of the site.
We **need a senior engineer on Cal team** to verify that this change is okay.

We don't know if this has upstream repercussions.

Please don't merge this without checking this out. 🙏

DexterStorey: @zomars @emrysal @PeerRich 

Flagging this, otherwise, this PR should be good to 🚀

PeerRich (approved):

packages/lib/checkRateLimitAndThrowError.ts
sean-brydon: This looks safe to me <3 

This is only ever ran on the server in all of use cases.

sean-brydon (approved):
LGTM - always called on server side so that change is fine

NIT: we should move this to a HTTP error in future as now its not always called within trpc

DexterStorey:
Fixes: https://github.com/calcom/cal.com/issues/11916

PR #11906, opened 2023-10-15
feat: Adds missing bookingId to Cancel Webhook payload

### LINKED ISSUE #11849: Add bookingId to Cancel Webhook payload

Currently, i seem to only be able to follow a booking through webhooks going from `BOOKING_CREATED` to `BOOKING_RESCHEDULED` to `BOOKING_CANCELLED` through a booking's `uid` since the `BOOKING_CANCELLED` webhook has a property for `uid` but then no way for me to fetch the booking from that since the booking api uses booking `id` as a query param. There should be a single way for me to reference a [cal.com](http://cal.com/) booking. Using booking id makese the most sense as that's what the api exposes through urls, but that results in us not knowing which booking was canceled when we receive the webhook payload. We return booking id for both `BOOKING_CREATED` and `BOOKING_RESCHEDULED`, so adding it to `BOOKING_CANCELLED` makes sense.

(Write your answer here.)

Return booking id in payload of BOOKING_CANCELLED

manpoffc:
## What does this PR do?
This PR adds bookingId reference to the cancel webhook payload.

Fixes #11849

[https://www.loom.com/share/fbeb6c3a5016422bba8dd6809396d88e?sid=4c9cb1ad-f1b7-4323-8f57-92a8769315ec](url)

## Type of change

## How should this be tested?

- Create a webhook with triggers for booking created, booking rescheduled, booking cancelled.
- Create a booking and cancel it.
- The payload for the cancelled webhook will contain the bookingId.

## Mandatory Tasks


alishaz-polymath:
Thank you @manpoffc for the PR. I'll review this today 🙏

CarinaWolli (approved):
Looks good! Thank you 🙏

Issue #8197, opened 2023-04-11
[CAL-1458] Send Broken Integration Email when Calendar is broken


hariombalhara:
Currently if calendar is broken, meeting gets booked in Cal.com and no communication is sent through email.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-1458](https://linear.app/calcom/issue/CAL-1458/send-broken-integration-email-when-calendar-is-broken)</sub>

PR #11923, opened 2023-10-16
fix: `videoCallUrl` not updating when rescheduling with a broken Calendar integration

hariombalhara:
## What does this PR do?
Fixes, `videoCallUrl` not updating when rescheduling with a broken Calendar integration

- Added an asssertion that verifies that it works now.
- Added assertion to verify appsStatus also in email
- Started using node-html-parser to easily test more things in email HTML

More details https://threads.com/34542266165?s=iejN9Cku3XDgLFAaAXrKBo

## Type of change

## How should this be tested?
- Connect Google Calendar and setup Daily Video
- Book a meeting with an event-type that uses the Google Calendar and has Daily Video as location
- Go to Credential table and modify refresh_token and access_token for Google Calendar so that it can no longer fix itself
- Reschedule the previous booking and notice that it is giving the previous Cal Video URL instead of a new one corresponding to the new booking UID.

## Mandatory Tasks


apps/web/test/utils/bookingScenario/expects.ts
```diff
@@ -139,9 +139,10 @@ export function expectWebhookToHaveBeenCalledWith(
   const parsedBody = JSON.parse((body as string) || "{}");
 
   expect(parsedBody.triggerEvent).toBe(data.triggerEvent);
+
   if (parsedBody.payload.metadata?.videoCallUrl) {
     parsedBody.payload.metadata.videoCallUrl = parsedBody.payload.metadata.videoCallUrl
-      ? parsedBody.payload.metadata.videoCallUrl.replace(/\/video\/[a-zA-Z0-9]{22}/, "/video/DYNAMIC_UID")
```
hariombalhara, line 144: Instead of considering every UID as DYNAMIC_UID, actually test the value as there was a bug that could have found if we were checking that.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2134,7 +2133,8 @@ async function handler(
       }
       return prev;
     }, {} as { [key: string]: AppsStatus });
-    evt.appsStatus = Object.values(calcAppsStatus);
+    resultStatus = Object.values(calcAppsStatus);
+    return resultStatus;
```
hariombalhara, line 2137: Return `resultStatus` instead of setting it directly on evt in the function. evt isn't an argument passed to the function.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2174,44 +2174,43 @@ async function handler(
...
-        const copyEvent = cloneDeep(evt);
-        loggerWithEventDetails.debug("Emails: Sending rescheduled emails for booking confirmation");
-        await sendRescheduledEmails({
-          ...copyEvent,
-          additionalInformation: metadata,
-          additionalNotes, // Resets back to the additionalNote input and not the override value
-          cancellationReason: `$RCH$${rescheduleReason ? rescheduleReason : ""}`, // Removable code prefix to differentiate cancellation from rescheduling for email
-        });
-      }
+    const { metadata, videoCallUrl: _videoCallUrl } = getVideoCallDetails({
```
hariombalhara, line 2197: It's possible that only Calendar Integration may have failed, so we still have a new video URL.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2174,44 +2174,43 @@ async function handler(
...
+      results,
+    });
+
+    videoCallUrl = _videoCallUrl;
+    evt.appsStatus = handleAppsStatus(results, booking);
```
hariombalhara, line 2202: We should be able to call handleAppsStatus regardless of `updateEvent` presence. The function itself checks for `results`.

packages/features/bookings/lib/handleNewBooking/test/reschedule.test.ts
```diff
@@ -254,14 +259,18 @@ describe("handleNewBooking", () => {
...
+          ],
         });
 
         expectBookingRescheduledWebhookToHaveBeenFired({
           booker,
           organizer,
           location: BookingLocations.CalVideo,
           subscriberUrl: "http://my-webhook.example.com",
-          videoCallUrl: `${WEBAPP_URL}/video/DYNAMIC_UID`,
+          videoCallUrl: `${WEBAPP_URL}/video/${createdBooking.uid}`,
```
hariombalhara, line 273: Test the CalVideo URL completely.

packages/features/bookings/lib/handleNewBooking/test/reschedule.test.ts
```diff
@@ -577,16 +592,27 @@ describe("handleNewBooking", () => {
           },
           to: {
             description: "",
-            location: "New York",
+            location: "integrations:daily",
```
hariombalhara, line 595: Updated the test to use DailyVideo has location during reschedule

Udit-takkar (commented):
@hariombalhara so i got the correct meeting link on the booking page but i didn't got any email about rescheduling of meeting or that my integration in broken

hariombalhara:
Yeah that is a known issue @Udit-takkar. It needs to be tackled separately.
https://github.com/calcom/cal.com/issues/8197

In such case a broken integration email should be sent ideally. Also, we have considered even disabling the booking page itself if Google Calendar isn't connected correctly.
https://threads.com/34541026838?s=dAytZR5kvGHFyWgPAQ9Jv3

Udit-takkar (approved):
LGTM

PR #11936, opened 2023-10-17
fix: unlock limit tab for managed event types

### LINKED ISSUE #11882: managed event types: be able to have your own limits
CleanShot 2023-10-13 at 15 16 54@2x

as a non-admin member of a managed event-type it would be nice if people could still have their own buffers etc. for managed event-types

CarinaWolli:
## What does this PR do?

We want to have all fields in the limit tab unlocked for managed event types. In v2 admins will be able to define themselves which fields should be locked and unlocked. 

Fixes #11882

## Type of change

## How should this be tested?

- Create managed event type and assign to users
- Test if users can change the settings in the limit tab of their managed event type


PeerRich (approved):
looks good! only code removed afaik

PR #11940, opened 2023-10-17
chore: Remove users from getSchedule return

### LINKED ISSUE #11372: [CAL-2475] getSchedule payload size reduction

When calling /api/trpc/public/slots.getSchedule with ~150 users available for each time slot, the response includes an array with the emails of these 150 users for every time slot. This generates a ~1.5MB response payload with all that redundant data.

We could create an indexed array of users, and reference the position instead of the whole email. This reduces the response size by an order of magnitude. The downside is if the users tend to not repeat across time slots, then it would actually increase the payload.

e.g.:
```json
{
  "result": {
    "data": {
      "json": {
        "slots": {
          "2023-09-15": [
            {
              "time": "2023-09-15T04:00:00.000Z",
              "users": [
                "user1@example.com",
                "user2@example.com",
                "user3@example.com",
                "user4@example.com",
                "user5@example.com",
                "user6@example.com",
                "user7@example.com",
                "user8@example.com",
                "user9@example.com",
                "user10@example.com"
              ]
            },
            {
              "time": "2023-09-15T04:00:10.000Z",
              "users": [
                "user1@example.com",
                "user2@example.com",
                "user3@example.com",
                "user4@example.com",
                "user5@example.com",
                "user6@example.com",
                "user7@example.com",
                "user8@example.com",
                "user9@example.com",
                "user10@example.com"
              ]
            }
          ]
        }
      }
    }
  }
}
```
becomes
```json
{
  "result": {
    "data": {
      "json": {
        "slots": {
          "2023-09-15": [
            {
              "time": "2023-09-15T04:00:00.000Z",
              "users": [
                0,
                1,
                2,
                3,
                4,
                5,
                6,
                7,
                8,
                9
              ]
            },
            {
              "time": "2023-09-15T04:00:10.000Z",
              "users": [
                0,
                1,
                2,
                3,
                4,
                5,
                6,
                7,
                8,
                9
              ]
            }
          ]
        },
        "users": [
          "user1@example.com",
          "user2@example.com",
          "user3@example.com",
          "user4@example.com",
          "user5@example.com",
          "user6@example.com",
          "user7@example.com",
          "user8@example.com",
          "user9@example.com",
          "user10@example.com"
        ]
      }
    }
  }
}
```

<sub>[CAL-2475](https://linear.app/calcom/issue/CAL-2475/getschedule-payload-size-reduction)</sub>

emrysal:
## What does this PR do?

Heavily reduces the getSchedule payload.

### DIFF
--- a/packages/trpc/server/routers/viewer/slots/util.ts
+++ b/packages/trpc/server/routers/viewer/slots/util.ts
@@ -567,7 +567,7 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
 
   const computedAvailableSlots = availableTimeSlots.reduce(
     (
-      r: Record<string, { time: string; users: string[]; attendees?: number; bookingUid?: string }[]>,
+      r: Record<string, { time: string; attendees?: number; bookingUid?: string }[]>,
       { time, ...passThroughProps }
     ) => {
       // TODO: Adds unit tests to prevent regressions in getSchedule (try multiple timezones)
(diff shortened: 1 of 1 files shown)


CarinaWolli (approved):
LGTM!

keithwillcode:
@emrysal @CarinaWolli This tRPC procedure is called from our API. If we have clients using this data, their integration could break. I think we should add some sort of option specifying whether or not to include users. i.e. In the API where we call the procedure, make it still return the users (alternatively, actually add this to the API request if you want to include the data) and then in the normal tRPC call from our app, we tell it to not include this data so we reduce the payload size.

keithwillcode:
@emrysal and I chatted and decided that since this functionality is such a degraded experience, we will continue forward with this change.

PR #8506, opened 2023-04-24
Add feature to offset start times for event types

rjackson:
## What does this PR do?

This PR adds a new "Offset start time" option to event types, which can be used to push time slots back from aligning perfectly with calendar hours / intervals.

For example, you can set up an event to serve the slots:

- 9:10am to 10:00am
- 10:10am to 11:00am
- 11:10am to 12:00pm

By configuring it with:

- Event length: 50 minutes
- Offset start time: 10 minutes

This feature was requested through #8336

**New field in the Event Types edit screen**

image

I wasn't quite sure where to put this field. It doesn't seem like a common option to show prominently on the "Event Setup" tab. I ended up feeling like the "Limits" tab is the best place for it, as its conceptually similar to before & after buffer times – just aligned to the calendar slot frequency instead of prior or future events.

**An example appointment page**

image

**Environment**: Staging(main branch)

## Type of change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update
  - (if we want to add a page / note about it as a new feature?)

## How should this be tested?

- [ ] Unit tests (getSchedule.test.ts and slots.test.ts have been updated)
- [ ] Create a new event type and on its edit page
  - [ ] See the offset starts at 0
  - [ ] Change it and successfully save it to another value
  - [ ] Blank the field and save, see it reset back to 0
  - [ ] Try to set the field to a negative value, and be blocked by validation
  - [ ] Set it as required for your own testing
- [ ] View the appointment page, and see results appear as configured

## Checklist

- ~~I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)~~
- ~~My code doesn't follow the style guidelines of this project~~
- ~~I haven't performed a self-review of my own code and corrected any misspellings~~
- ~~I haven't commented my code, particularly in hard-to-understand areas~~
- ~~I haven't checked if my PR needs changes to the documentation~~
- ~~I haven't checked if my changes generate no new warnings~~
- ~~I haven't added tests that prove my fix is effective or that my feature works~~
- ~~I haven't checked if new and existing unit tests pass locally with my changes~~

Custom checklist additions:

- End-to-end tests: I couldn't run these locally, so I'm hoping the PR can verify these for me.

Bounty hunting: /claim #8336

PR #11943, opened 2023-10-17
fix: Removes iterative offset being added to each slot

alishaz-polymath:
## What does this PR do?

In the offset slot feature, currently, the offset is added to each slot iteratively, which means that each slot would not only start at the provided offset of 'X' minutes, but also add 'X' minutes in the slot interval.

Eg. With an offset of 10 mins, for a 30 min event, we expect:
9:10, 9:40, 10:10, 10:40, ...
However, what we get currently is:
9:10, 9:50, 10:30, 11:10, ...

So, an extra 10 min interval time is added to the slot interval, which appears to be unintended.

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/test/lib/getSchedule.test.ts
+++ b/apps/web/test/lib/getSchedule.test.ts
@@ -678,7 +678,7 @@ describe("getSchedule", () => {
         eventTypes: [
           {
             id: 1,
-            length: 25,
+            length: 30,
             offsetStart: 5,
             users: [
               {
--- a/packages/lib/slots.ts
+++ b/packages/lib/slots.ts
@@ -198,7 +198,7 @@ function buildSlotsWithDateRanges({
       slots.push({
         time: slotStartTime,
       });
-      slotStartTime = slotStartTime.add(frequency + (offsetStart ?? 0), "minutes");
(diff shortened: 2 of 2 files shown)


CarinaWolli:
This would break the initial idea of that feature, see the example in the PR description: https://github.com/calcom/cal.com/pull/8506

zomars:
Agreed with @CarinaWolli I think the offset is expected for each slot. In theory you "offset the whole schedule" by updating the schedule to start at 9:10 for example. Will close for now.

alishaz-polymath:
That PR info was helpful @CarinaWolli thank you 🙏 

> In theory you "offset the whole schedule" by updating the schedule to start at 9:10 for example
I think there's extra 'X' minutes coming in, because once you offset the start by adding let's say 10 mins in this case, the rest of the slots are automatically shifted by 10 mins as they are calculated by event lenght/slot interval length. However, in this case, we also add 10mins on top of that. Based on the PR description, this was intended, so it makes sense to close this PR 🙏 

I think in this case, one way to achieve what I was trying to do is to reverse-use the thought here. We could subtract the number of minutes added in offset from the event length, and then set that as slot interval, and that's probably good to go.

PR #11929, opened 2023-10-17
fix: scheduleEmailReminders cron job

CarinaWolli:
## What does this PR do?

Attempt to fix failing github action `Cron-scheduleEmailReminders `(https://github.com/calcom/cal.com/actions/workflows/cron-scheduleEmailReminders.yml) --> currently disabled

Axiom error: 
```
PrismaClientKnownRequestError: 
Invalid `prisma.workflowReminder.findMany()` invocation:

This request could not be understood by the server: {"type":"UnknownJsonError","body":{"code":"P6004","message":"The Query did not produce a result within the maximum allowed execution time of 10 seconds."}}
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #11960, opened 2023-10-18
fix: failing scheduleEmailReminder cron job

CarinaWolli:
## What does this PR do?

After https://github.com/calcom/cal.com/pull/11929 got merged, cron - scheduleEmailReminders is still failing. Now with a  new error message: 

```
PrismaClientKnownRequestError: 
Invalid `prisma.workflowReminder.findMany()` invocation:

This request could not be understood by the server: {"type":"UnknownJsonError","body":{"code":"P6009","message":"The response size of the query exceeded the the maximum of 5MB with 15.07MB. Consider refining the query by narrowing the selection set or applying appropriate filters. If that is not acceptable for your use case, please contact Prisma support."}} 
```

With this PR we now only select the needed properties when querying workflowReminders, which hopefully helps us to reduce the response size enough

### DIFF
--- a/packages/features/ee/workflows/api/scheduleEmailReminders.ts
+++ b/packages/features/ee/workflows/api/scheduleEmailReminders.ts
@@ -34,7 +34,11 @@ type Booking = Prisma.BookingGetPayload<{
   };
 }>;
 
-function getiCalEventAsString(booking: Booking) {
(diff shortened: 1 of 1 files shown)


Udit-takkar (approved):
LGTM

PR #11963, opened 2023-10-18
fix: use app.slug not hard coded zoom

manish-singh-bisht:
fixes chore: use app.slug not hard coded zoom #11856


joeauyeung (approved):
LGTM. Thank you for your contribution @manish-singh-bisht

PR #11969, opened 2023-10-18
chore: [app dir bootstrapping 1] generate nonce with native crypto API

grzpab:
## What does this PR do?

This PR replaces the usage of Node.js `crypto` module with native `crypto` API.
If the app router is used, the Node.js `crypto` module can no longer be used within middleware functions.
Otherwise, we run into the following error: https://nextjs.org/docs/messages/node-module-in-edge-runtime
You can also read: https://github.com/vercel/next.js/discussions/47314

I found a solution like this: https://gist.github.com/themikefuller/c1de46cbbdad02645b9dc006baedf88e but honestly, I find it hard to understand. Please take into account that `window.btoa` and `window.atob` operate on binary strings, as per https://developer.mozilla.org/en-US/docs/Web/API/btoa, which might be confusing in terms of conversion, code points, etc.

Since the nonce is just a string per documentation, https://html.spec.whatwg.org/multipage/urls-and-fetching.html#attr-nonce, not necessarily a base64 string, I came up with the following solution
* let's generate 22 random bytes using the native `crypto` API
* let's cut them all to 6 bits and replace these 6 bits with the base64 ASCII characters
* let's add `==` as padding
* we end up with a 24-ASCII-character string that is a valid base64 string (which means it contains the proper base64 characters and can be reverted into the initial non-encoded blob, but the latter step is not required unless YOU ACTUALLY DECODE THE NONCES).

Please note I did not convert any buffer value to base64, I generated a string that looks as if it was encoded from some value.

## Requirement/Documentation
1. https://html.spec.whatwg.org/multipage/urls-and-fetching.html#attr-nonce
2. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce
3. https://en.wikipedia.org/wiki/Base64#Output_padding
4. https://nextjs.org/docs/pages/api-reference/edge#unsupported-apis
5. https://vercel.com/templates/next.js/edge-functions-crypto

## Type of change

## How should this be tested?
The flow on Chrome:
1. Go to any page that uses nonces (e.g. login page, signup page)
2. Open the Developer Tools
3. Check the request to the HTML page that you are currently viewing (you'd probably have to reload the page to see it)
4. Go to Headers, Response Headers and look for `Content-Security-Policy-Report-Only:`
5. The header should contain the nonce value explicitly, 
6. Go to Response tab and look for `nonce` in the returned HTML. You should see the same value in around 46 places, usually in `<script>` tags.

Trivia:
You won't see `nonce` value in the HTML within the Elements tab. The reason is apparently security. 

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/lib/buildNonce.test.ts
+++ b/apps/web/lib/buildNonce.test.ts
@@ -1,0 +1,96 @@
+import { describe, it, expect } from "vitest";
+
+import { buildNonce } from "./buildNonce";
+
+describe("buildNonce", () => {
+  it("should return an empty string for an empty array", () => {
+    const nonce = buildNonce(new Uint8Array());
+
+    expect(nonce).toEqual("");
+    expect(atob(nonce).length).toEqual(0);
+  });
+
+  it("should return a base64 string for values from 0 to 63", () => {
+    const array = Array(22)
+      .fill(0)
+      .map((_, i) => i);
+    const nonce = buildNonce(new Uint8Array(array));
+
+    expect(nonce.length).toEqual(24);
+    expect(nonce).toEqual("ABCDEFGHIJKLMNOPQRSTQQ==");
+
+    expect(atob(nonce).length).toEqual(16);
+  });
+
+  it("should return a base64 string for values from 64 to 127", () => {
+    const array = Array(22)
+      .fill(0)
+      .map((_, i) => i + 64);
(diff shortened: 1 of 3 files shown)


dmytro068 (approved):

zomars (approved):
LGTM

PR #11976, opened 2023-10-18
fix: Add /embed route for booking/[uid]

hariombalhara:
## What does this PR do?
Support embedding /booking/[uid] page separately

## Type of change

## How should this be tested?
Try to open an existing booking link with /embed in the end e.g. http://app.cal.local:3000/booking/iHFhuzFdQz3xxCXjHUjSke/embed. It was not required to be supported earlier but now it's supported as requested. 

## Mandatory Tasks

## Checklist


alishaz-polymath (approved):
Looks good. Tested and works 🚀

PR #11978, opened 2023-10-18
feat: Allow hideBranding via public API

alishaz-polymath:
## What does this PR do?

Allows toggling `hideBranding` on or off for a user via public API calls (made by instance Admins, of course).

## Requirement/Documentation

## Type of change

## How should this be tested?

- [x] You should be able to update hideBranding in /users endpoint via `POST` or /users/[id] `PATCH` requests made by Admin.
- [x] You shouldn't be able to update hideBranding in /users/[id] `PATCH` request by a non-Admin user

## Mandatory Tasks


PeerRich (approved):

apps/api/pages/api/users/[userId]/_patch.ts
```diff
@@ -94,9 +97,10 @@ export async function patchHandler(req: NextApiRequest) {
   if (!isAdmin && query.userId !== req.userId) throw new HttpError({ statusCode: 403, message: "Forbidden" });
 
   const body = await schemaUserEditBodyParams.parseAsync(req.body);
-  // disable role changes unless admin.
-  if (!isAdmin && body.role) {
-    body.role = undefined;
+  // disable role or branding changes unless admin.
+  if (!isAdmin) {
+    if (body.role) body.role = undefined;
+    if (body.hideBranding) body.hideBranding = undefined;
```
Comment (keithwillcode) [inline apps/api/pages/api/users/[userId]/_patch.ts:103]: Am I incorrect in thinking this is going to actually override the hideBranding value that's already stored in the database instead of just ignoring what's on the body since we just use the full body object on line 199?

keithwillcode (changes requested):

apps/api/pages/api/users/[userId]/_patch.ts
Comment (alishaz-polymath) [reply apps/api/pages/api/users/[userId]/_patch.ts:103]: This would ignore "hideBranding" as the only acceptable values for hideBranding are boolean, and undefined is pretty much as if it hasn't been passed (since this is optional) in validation schema.

Comment (keithwillcode) [reply apps/api/pages/api/users/[userId]/_patch.ts:103]: Cool

keithwillcode (approved):
Looks good. Let's make sure to follow up with a PR with automated tests verifying this functionality.

PR #11980, opened 2023-10-18
chore: [app dir bootstrapping 3] check nullability in AppListCard

grzpab:
## What does this PR do?
After adding the app directory, the `next build` step will change the definitions of `usePathname` and `useSearchParams` hooks by creating new type definitions in `.next` directory. This will ensure that the return types contain null.
Actually, the return types can be nullable in the `pages` router under specific circumstances outlined here:
1. https://nextjs.org/docs/app/api-reference/functions/use-search-params
2. https://nextjs.org/docs/app/api-reference/functions/use-pathname

## Type of change

## How should this be tested?
This PR does no change in behavior.
1. Go to https://app.cal.com/apps/installed/conferencing?hl=daily-video
2. The Cal Video should be highlighted for 3 seconds
3. The highlight disappears and you get redirected to https://app.cal.com/apps/installed/conferencing?category=conferencing

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/components/AppListCard.tsx
+++ b/apps/web/components/AppListCard.tsx
@@ -60,14 +60,18 @@ export default function AppListCard(props: AppListCardProps) {
   const pathname = usePathname();
 
   useEffect(() => {
-    if (shouldHighlight && highlight) {
-      const timer = setTimeout(() => {
-        setHighlight(false);
+    if (shouldHighlight && highlight && searchParams !== null && pathname !== null) {
+      timeoutRef.current = setTimeout(() => {
         const _searchParams = new URLSearchParams(searchParams);
         _searchParams.delete("hl");
-        router.replace(`${pathname}?${_searchParams.toString()}`);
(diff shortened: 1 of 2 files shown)


zomars:
Shouldn't the new URL be `https://app.cal.com/apps/installed/conferencing` instead of `https://app.cal.com/apps/installed/conferencing?category=conferencing`

grzpab:
@zomars I investigated and the additional `category` search parameter comes from the dynamic params that are returned as part of `useSearchParams` in the current version of Next.js.

The solutions are as follows:
1. Remove the dynamic params from the new search params explicitly - that's what I did in this PR.
2. Upgrade Next.js to the newest version which does not bundle dynamic params in the search params.
3. Pass a callback from the page component that replaces the router URL to the `AppListCard`; the callback should contain the explicit set up of dynamic and search params.

zomars (approved):
Good enough to merge. Thanks for clarifying 🙏🏽

PR #11986, opened 2023-10-18
refactor: Falling Back to `FirstCalendarCredential`

joeauyeung:
## What does this PR do?

This PR aims to solve some error logs involving the `firstCalendarCredential` in the `EventManager`

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/core/EventManager.ts
```diff
@@ -489,6 +489,22 @@ export default class EventManager {
...
+       */
+      const [credential] = this.calendarCredentials.filter((cred) => !cred.type.endsWith("other_calendar"));
+      if (credential) {
+        const createdEvent = await createEvent(credential, event);
+        log.silly("Created Calendar event", safeStringify({ createdEvent }));
+        if (createdEvent) {
+          createdEvents.push(createdEvent);
+        }
+      }
+    };
```
joeauyeung, line 506: Reusing this logic so abstracted into it's own function within `createAllCalendarEvents`

packages/core/EventManager.ts
```diff
@@ -545,6 +561,14 @@ export default class EventManager {
           );
           // It might not be the first connected calendar as it seems that the order is not guaranteed to be ascending of credentialId.
           const firstCalendarCredential = destinationCalendarCredentials[0];
```
joeauyeung, line 563: This check is assuming that there is a second calendar credential of the same type as the original destination calendar.

packages/core/EventManager.ts
```diff
@@ -545,6 +561,14 @@ export default class EventManager {
...
+
+          if (!firstCalendarCredential) {
+            log.warn(
+              "No other credentials found of the same type as the destination calendar. Falling back to first connected calendar"
+            );
+            await fallbackToFirstConnectedCalendar();
+          }
```
joeauyeung, line 570: If there is no second credential of the same type as the destination calendar, then fallback to the first connected calendar.

packages/lib/piiFreeData.ts
```diff
@@ -59,18 +59,12 @@ export function getPiiFreeBooking(booking: {
 }
 
 export function getPiiFreeCredential(credential: Partial<Credential>) {
-  return {
-    id: credential.id,
```
joeauyeung, line 63: There were some logs with id being unidentified because we don't strictly type check to see if all these properties are in the original credential.

packages/lib/piiFreeData.ts
```diff
@@ -59,18 +59,12 @@ export function getPiiFreeBooking(booking: {
...
-     * Let's just get a boolean value for PII sensitive fields so that we atleast know if it's present or not
-     */
-    key: getBooleanStatus(credential.key),
-  };
+  /**
+   * Let's just get a boolean value for PII sensitive fields so that we atleast know if it's present or not
+   */
+  const booleanKeyStatus = getBooleanStatus(credential?.key);
+
+  return { ...credential, key: booleanKeyStatus };
```
joeauyeung, line 67: Let's just change the credential keys to a boolean and return the credential as we received it.

CarinaWolli (approved):
code looks good 👍

PR #11995, opened 2023-10-19
chore: [app dir bootstrapping 6] server-side translations

dmytro068:
## What does this PR do?
- Adds a patch for next-i18next library that makes it compatible with app dir
- Adds locale lookup to `getLocale` function to ensure locale fallback is always applied (e.g. es-419 will fallback to es). 
This fixes inconsistencies between html `lang` attribute and `i18n` selected locale.
- Updates tests
- Removes `i18n.fetch` from ssgInit/ssrInit functions. We dont need to make API call, because `serverSideTranslations` function can just read translation files directly from file system.  

## Requirement/Documentation

**This PR contains patch of `next-i18next`, so `yarn.lock` should be regenerated.** 
[Description of server-side translations issue ](https://www.notion.so/intuita/Server-side-translations-issue-next-i18next-28b295dbc45b4a53aee8b5cd9b73c88b)

## Type of change

## How should this be tested?

localization feature is covered by `apps/web/playwright/locale.e2e.ts`

## Mandatory Tasks


zomars (approved):
Tested and it's working. Merging to test on staging.

PR #11996, opened 2023-10-19
fix: Handle non-org team with same slug as the organization's requestedSlug

hariombalhara:
## What does this PR do?

The fix is in 2 parts
1.  Preventing the scenario where requestedSlug for an org is allowed to be same as a non-org team slug
2. Handle the scenario where even if there are conflicting slugs, it should retrieve the ORG only and not any team for Org Profile(i.e. https://calcom.cal.com would always serve an ORG, it can't serve a team)

[Demo before and After fix](https://www.loom.com/share/20c3d1fd570147f0baaa13987f90c90c)

More context in [internal thread](https://threads.com/thread/34543004525/34543097153?s=8eiMC4zWbUVBQoBeWWKne8)

## Type of change

## How should this be tested?
- [x] Create a team(T1) with a slug(Slug1) outside any organization.
- [x] Try creating an org(O1) with the same slug. Locally, you would have to set[ IS_TEAM_BILLING_ENABLED=1](https://github.com/calcom/cal.com/blob/987751beed3cfc736eed3f1474c02e6d83af0e67/packages/lib/constants.ts#L90) temporarily to create a scenario when instead of adding a slug to the Organization, we would add `requestedSlug` instead.
  With this branch, it won't be allowed but in main it is allowed.
- [x] Once the Org is created. Try accessing that org, e.g. http://calcom.cal.local:3000
- [x] Notice that you would see the team T1 instead of org O1

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/pages/team/[slug].tsx
+++ b/apps/web/pages/team/[slug].tsx
@@ -1,3 +1,9 @@
+// This route is reachable by
+// 1. /team/[slug]
+// 2. / (when on org domain e.g. http://calcom.cal.com/. This is through a rewrite from next.config.js)
+// Also the getServerSideProps and default export are reused by
+// 1. org/[orgSlug]/team/[slug]
+// 2. org/[orgSlug]/[user]/[type]
 import classNames from "classnames";
 import type { GetServerSidePropsContext } from "next";
 import Link from "next/link";
@@ -12,6 +18,7 @@ import { WEBAPP_URL } from "@calcom/lib/constants";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { useRouterQuery } from "@calcom/lib/hooks/useRouterQuery";
 import useTheme from "@calcom/lib/hooks/useTheme";
+import logger from "@calcom/lib/logger";
 import { markdownToSafeHTML } from "@calcom/lib/markdownToSafeHTML";
 import { getTeamWithMembers } from "@calcom/lib/server/queries/teams";
 import slugify from "@calcom/lib/slugify";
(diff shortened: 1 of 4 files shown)


apps/web/pages/team/[slug].tsx
```diff
@@ -277,12 +284,23 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
...
+    isOrganizationFeatureEnabled,
+    isValidOrgDomain,
+    currentOrgDomain,
+  });
+
   const team = await getTeamWithMembers({
     slug: slugify(slug ?? ""),
     orgSlug: currentOrgDomain,
     isTeamView: true,
-    isOrgView: isValidOrgDomain && context.resolvedUrl === "/",
```
Comment (hariombalhara) [inline apps/web/pages/team/[slug].tsx:285]: resolvedUrl is what we get after rewrite, so this should never be / AFAIU.
I am using a more reliable approach of passing the query param itself when we do the rewrite, because that's the logic that decides it's an org link and redirects accordingly.

packages/lib/server/queries/teams/index.ts
```diff
@@ -120,12 +124,30 @@ export async function getTeamWithMembers(args: {
...
+    };
+  }
 
-  const team = await prisma.team.findFirst({
+  const teams = await prisma.team.findMany({
     where,
     select: teamSelect,
   });
 
+  if (teams.length > 1) {
```
hariombalhara, line 143: Even though I am ensuring now that there should be one team, it could still happen in future and might be happening right now for some unknown reason.

So, I am logging it as an error but plan to convert it into an actual error, we shouldn't be serving accidentally different team here.

hariombalhara (commented):
Self reviewed

Udit-takkar (commented):
If an org(acme) and team(seeded-team) has same requestedSlug in metadata then i see the wrong avatar "ST" instead of "AC".

Screenshot 2023-10-19 at 11 00 22 PM

hariombalhara:
Thanks for review @Udit-takkar 🙏 Great testing.
I think this particular issue isn't related to the changes I have made here. I will fix it in a followup.

Also, practically, slug can be same as requestedSlug or null, unless someone plays with DB but still an important issue to fix.

Udit-takkar (approved):
LGTM. Everything else works fine 🚀

PR #12005, opened 2023-10-19
chore: [app dir bootstrapping 4] check nullability of navigation hook return values

grzpab:
## What does this PR do?
This PR enforces nullability checks for return values from `useParams` and `useSearchParams`.
If a page has neither `getServerSideProps` not `getInitialProps`, Next.js will optimize it statically. This makes the `usePathname` and `useSearchParams` equal `null` until the router becomes available during hydration.

If you add the app directory to the web directory and run yarn build, the types will be updated by Next.js builder/plugin and the projects will then include the compat type definitions provided by Next.js.

This mechanism is coded here: https://github.com/vercel/next.js/blob/524b31513a58e58e15862ac8aa3f27da8a47a267/packages/next/src/lib/typescript/writeAppTypeDeclarations.ts#L54

The example compat type definition is here: https://github.com/vercel/next.js/blob/524b31513a58e58e15862ac8aa3f27da8a47a267/packages/next/navigation-types/compat/navigation.d.ts

## Requirement/Documentation
https://nextjs.org/docs/app/api-reference/functions/use-pathname
https://nextjs.org/docs/app/api-reference/functions/use-search-params
https://nextjs.org/docs/pages/building-your-application/rendering/automatic-static-optimization

## Type of change

## How should this be tested?
The app should work exactly as worked before.

Since nullability issues (TypeErrors) have not been observed before, this change should only satisfy the TS compiler.

Caveats:
1. The prerendered pages doesn't have the `useSearchParams` and `usePathname` return values on the prerender. Other pages don't have these values on the first render. The code should not allow:
    a) running write operation on first render (with e.g. `useEffect`) that use values from these hooks.
    b) running read operations unconditionally based on the values from these hooks.

One idea was to wrap each component with a HOC that:
1) returns `null` when `useSearchParams` or `usePathname` are `null`
2) otherwise, it passes non-empty `searchParams` and `pathname` as props to the wrapped component.
One problem is possible flickering stemming from a lot of components appearing when these two values become available.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/features/ee/organizations/pages/settings/other-team-members-view.tsx
+++ b/packages/features/ee/organizations/pages/settings/other-team-members-view.tsx
@@ -58,7 +58,7 @@ const MembersView = () => {
   const { t, i18n } = useLocale();
   const router = useRouter();
   const searchParams = useSearchParams();
-  const teamId = Number(searchParams.get("id"));
+  const teamId = Number(searchParams?.get("id"));
   const session = useSession();
   const utils = trpc.useContext();
   const [offset, setOffset] = useState<number>(1);
--- a/packages/features/ee/organizations/pages/settings/other-team-members-view.tsx
+++ b/packages/features/ee/organizations/pages/settings/other-team-members-view.tsx
@@ -74,6 +74,7 @@ const MembersView = () => {
   const { data: team, isLoading: isTeamLoading } = trpc.viewer.organizations.getOtherTeam.useQuery(
     { teamId },
     {
+      enabled: !Number.isNaN(teamId),
       onError: () => {
         router.push("/settings");
       },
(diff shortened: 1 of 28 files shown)


packages/features/ee/organizations/pages/settings/other-team-members-view.tsx
```diff
@@ -101,12 +103,8 @@ const MembersView = () => {
 
   useEffect(() => {
     if (membersFetch) {
-      if (membersFetch.length < limit) {
-        setLoadMore(false);
-      } else {
-        setLoadMore(true);
-      }
-      setMembers(members.concat(membersFetch));
+      setLoadMore(membersFetch.length >= limit);
```
zomars, line 106: nice

zomars (commented):
Nice. Big change. Will test locally before approving. Thanks again @grzpab 🙏🏽

zomars (approved):
Code LGTM. Tested locally and found no issues. Checks are passing. I think it's safe to merge and test on dev and QA.

PR #12017, opened 2023-10-20
fix: get correct count for team members in slider

sean-brydon:
fixes issue where team membercount was never correct when in a team context. 

Org context is uneffected for this

### DIFF
--- a/packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
+++ b/packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
@@ -1,3 +1,5 @@
+import { Prisma } from "@prisma/client";
+
 import type { Dayjs } from "@calcom/dayjs";
 import dayjs from "@calcom/dayjs";
 import type { DateRange } from "@calcom/lib/date-ranges";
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -120,9 +122,16 @@ async function getInfoForAllTeams({ ctx, input }: GetOptions) {
     limit,
   });
 
+  // Get total team count across all teams the user is in (for pagination)
+
+  const totalTeamMembers =
+    await prisma.$queryRaw<number>`SELECT COUNT(DISTINCT "userId")::integer from "Membership" WHERE "teamId" IN (${Prisma.join(
+      teamIds
+    )})`;
```
sean-brydon, line 130: You cant count distinct when using IN in prisma client for some reason. Had to go with the raw approach

packages/trpc/server/routers/viewer/availability/team/listTeamAvailability.handler.ts
```diff
@@ -120,9 +122,16 @@ async function getInfoForAllTeams({ ctx, input }: GetOptions) {
...
+
   return {
     teamMembers,
-    totalTeamMembers: teamMembers.length,
```
sean-brydon, line 125: No wonder it wasnt working for teams - teamMembers is always the length of the limit via pagination

alishaz-polymath (approved):
LGTM 🚀

keithwillcode (approved):

PR #12019, opened 2023-10-20
fix: Date add 1 day adds 24 hours, not 1 day

emrysal:
causing the last date to be lost on dst change

### DIFF
--- a/packages/features/calendars/DatePicker.tsx
+++ b/packages/features/calendars/DatePicker.tsx
@@ -122,20 +122,19 @@ const Days = ({
   // Create placeholder elements for empty days in first week
   const weekdayOfFirst = browsingDate.date(1).day();
   const currentDate = minDate.utcOffset(browsingDate.utcOffset());
+
   const availableDates = (includedDates: string[] | undefined) => {
     const dates = [];
     const lastDateOfMonth = browsingDate.date(daysInMonth(browsingDate));
-    for (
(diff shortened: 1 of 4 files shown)


CarinaWolli:
I am have some missing slots now on October 31. 

#### System tz: Melbourne

Last timeslot in London tz: 12:45
Screenshot 2023-10-20 at 22 28 51


#### System tz: Vienna

Last timeslot in London tz: 16:45
Screenshot 2023-10-20 at 13 29 46

packages/features/calendars/DatePicker.tsx
```diff
@@ -122,20 +122,19 @@ const Days = ({
...
     const dates = [];
     const lastDateOfMonth = browsingDate.date(daysInMonth(browsingDate));
-    for (
-      let date = currentDate;
-      date.isBefore(lastDateOfMonth) || date.isSame(lastDateOfMonth, "day");
-      date = date.add(1, "day")
-    ) {
+    let date = currentDate;
+    do {
+      date = date.date(date.date() + 1);
```
keithwillcode, line 131: In the existing code, the date was updated after dates were pushed to the array. Am I reading this correctly that the first iteration through this do/while will actually update the date before pushing any dates to the array. Will we miss the first day here somehow?

zomars (approved):
Nice find!

PR #12031, opened 2023-10-23
fix: Avatar slug and cal links for cross org users

hariombalhara:
## What does this PR do?

- Fix wrong avatar and booking link for non-org member in an org team(Scenario-1)
	- When migrating a user to an org we end up in this scenario where team is migrated to Org but some of it's users aren't
		- It's important to do it because we should infer the avatar and link from the user and not from the booking domain he is on.This completely avoids the issues where we see different user's avatar for a user.
- Fix wrong avatar shown on Team Availability

## Type of change

## How should this be tested?
- [x] Scenario-1: Non org member in an org team
	- 	Create an org(O1) and a team(T1) inside it.
	-  Add a member to the team T1
	- Now go to db-studio and set organizationId to null for the user and also remove membership of the user with the Org. It creates a scenario where non-org member is member of a team.
	- Now, go to org booking page(o1.cal.local:3000) and then go to the team page in the org. You will still see that member there. It should correctly show it's avatar, and when you click on it, it should correctly take you to it's booking link which would be on app.cal.local:3000 and not o1.cal.local:3000
- [x] Scenario-2: Two users with same username(one in org and other outside org)
	- Login to org O1 and invite a new user teampro@o1.com which would have username teampro(same as seeded user)
	- Go to availability and switch to Team Availability, notice that the avatar for teampro is not of Org user `teampro` but of non-org user `teampro`. This bug is fixed here.

## Mandatory Tasks

## Checklist


apps/web/components/getting-started/steps-views/UserProfile.tsx
```diff
@@ -96,16 +97,19 @@ const UserProfile = () => {
     },
   ];
 
+  const organization =
+    user.organization && user.organization.id
+      ? {
+          ...(user.organization as Ensure<typeof user.organization, "id">),
+          slug: user.organization.slug || null,
+          requestedSlug: user.organization.metadata?.requestedSlug || null,
+        }
```
hariombalhara, line 106: Due to the way user.organization is set, all it's props are optional. So, handling it like this so that we can have all it's props set correctly

apps/web/pages/api/user/avatar.ts
```diff
@@ -30,39 +35,28 @@ async function getIdentityData(req: NextApiRequest) {
...
+    const user = await prisma.user.findFirst({
       where: {
         username,
         organization: orgQuery,
       },
       select: { avatar: true, email: true },
     });
 
-    /**
-     * TEMPORARY CODE STARTS - TO BE REMOVED after mono-user schema is implemented
```
hariombalhara, line 46: We don't need this hack now as the avatar itself has the orgId being passed from all places on booking pages.|
Goal is to do that from everywhere.

packages/features/ee/organizations/lib/orgDomains.ts
```diff
@@ -48,10 +56,14 @@ export function subdomainSuffix() {
   return urlSplit.length === 3 ? urlSplit.slice(1).join(".") : urlSplit.join(".");
 }
 
-export function getOrgFullDomain(slug: string, options: { protocol: boolean } = { protocol: true }) {
+export function getOrgFullOrigin(slug: string, options: { protocol: boolean } = { protocol: true }) {
```
hariombalhara, line 59: Renamed to correctly convey what it returns.

packages/features/ee/organizations/components/OrganizationMemberAvatar.tsx
```diff
@@ -0,0 +1,47 @@
+import classNames from "@calcom/lib/classNames";
```
hariombalhara, line 1: Renamed `OrganizationAvatar` to `OrganizationMemberAvatar` to more correctly convey what it is
Also, it started using UserAvatar instead of Avatar

packages/features/bookings/components/event-meta/Members.tsx
```diff
@@ -38,40 +25,22 @@ export const EventMembers = ({ schedulingType, users, profile, entity }: EventMe
...
-    alt: profile.name || undefined,
-    href: profile.username
-      ? `${CAL_URL}${pathname?.indexOf("/team/") !== -1 ? "/team" : ""}/${profile.username}`
-      : undefined,
-  });
-
-  const uniqueAvatars = avatars
-    .filter((item): item is AvatarWithRequiredImage => !!item.image)
-    .filter((item, index, self) => self.findIndex((t) => t.image === item.image) === index);
-
```
hariombalhara, line 71: I am not sure we were comparing images, it didn't make sense. We need to show all the users avatars and users should be unique already.

packages/features/bookings/components/event-meta/Members.tsx
```diff
@@ -38,40 +25,22 @@ export const EventMembers = ({ schedulingType, users, profile, entity }: EventMe
...
-  // Add organization avatar
-  if (entity.orgSlug) {
-    avatars.unshift({
-      title: `${entity.name}`,
-      image: `${WEBAPP_URL}/team/${entity.orgSlug}/avatar.png`,
-      alt: entity.name || undefined,
-      href: getOrgFullDomain(entity.orgSlug),
-    });
-  }
```
hariombalhara, line 56: Handled by `UserAvatarGroupWithOrg`

hariombalhara (commented):
Self reviewed

apps/web/components/ui/avatar/UserAvatar.tsx
```diff
@@ -0,0 +1,19 @@
+import { getUserAvatarUrl } from "@calcom/lib/getAvatarUrl";
+import type { User } from "@calcom/prisma/client";
+import { Avatar } from "@calcom/ui";
+
+type UserAvatarProps = Omit<React.ComponentProps<typeof Avatar>, "alt" | "imageSrc"> & {
+  user: Pick<User, "organizationId" | "name" | "username">;
+  /**
+   * Useful when allowing the user to upload their own avatar and showing the avatar before it's uploaded
+   */
+  imageSrc?: string | null;
```
ThyMinimalDev, line 10: I find keeping imageSrc a little bit confusing, is there a case where we don't want to display the user's avatar when using UserAvatar component ?

wouldn't we want to just use the Avatar component if it is to pass an imageSrc props ?

ThyMinimalDev, line 10: maybe we should rename this props to previewSrc or previewImgSrc to make it more clear that it is supposed to be used to display avatar while it's uploading, so that there is no mistake when using this component.

apps/web/pages/api/user/avatar.ts
```diff
@@ -1,15 +1,23 @@
...
 
+const log = logger.getSubLogger({ prefix: ["team/[slug]"] });
 const querySchema = z
   .object({
     username: z.string(),
     teamname: z.string(),
+    /**
+     * Passed when we want to fetch avatar of a particular organization
+     */
     orgSlug: z.string(),
```
ThyMinimalDev, line 21: the comment above orgSlug makes it sound like this can be optional ?

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
```diff
@@ -78,10 +81,17 @@ export function AvailabilitySliderTable() {
...
             <div className="max-w-64 flex flex-shrink-0 items-center gap-2 overflow-hidden">
-              <Avatar size="sm" alt={username || email} imageSrc={`/${username}/avatar.png`} />
+              <UserAvatar
+                size="sm"
+                user={{
+                  username,
+                  name,
+                  organizationId,
+                }}
+              />
```
ThyMinimalDev, line 94: it seems like we used to display the email as the image alt when the username was not defined, `UserAvatar` doesn't do that anymore, is it a wanted change ?

apps/web/components/ui/avatar/UserAvatar.tsx
hariombalhara: Agreed. Couldn't find a better name, but this sounds better.

apps/web/pages/api/user/avatar.ts
hariombalhara: Yeah actually all are optional. See .partial() below. We should improve this schema to instead use unions to better represent the structure.

packages/features/timezone-buddy/components/AvailabilitySliderTable.tsx
hariombalhara: We were doing different things at different places I believe. e.g. I saw that actually user.name is used at most places as `alt` value and if not set empty string is used.

So, I am not sure what we should finalize but to me looks like this is what we could do

§user.name || user.username || user.email§

apps/web/components/ui/avatar/UserAvatar.tsx
hariombalhara: Done

hariombalhara:
Great review @ThyMinimalDev !! Thanks 🙏

Resolved all the comments now.

ThyMinimalDev (approved):

PR #12041, opened 2023-10-23
fix: [app dir bootstrapping 8] `useParamsWithFallback` hook and add tests

hbjORbj:
## What does this PR do?

- Bug: `Error: NextRouter was not mounted` is thrown if `useRouter` from `next/router` is called in App router. Since `useParamsWithFallback` is meant to be used in both page router and app router, this shouldn't happen. 
- How the bug was found: In order to migrate `pages/workflows/[workflow].tsx` to App router, I created a file at `app/workflows/[workflow]/page.tsx` that imports component from `@calcom/features/ee/workflows/pages/workflow`. The component at `@calcom/features/ee/workflows/pages/workflow` uses `useParamsWithFallback` hook, in which `useRouter` from `next/router` is called.
- Solution: Use next.js compat router (this doesn't throw error but rather returns null in app router)
- Extra Note: We added tests to ensure the functionality of the hook.

## Requirement/Documentation

- [Next.js doc stating that `useParams` returns `null` in Pages Router](https://nextjs.org/docs/app/api-reference/functions/use-params#returns)
- [next.js compat router](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/compat/router.ts)

## Type of change

## How should this be tested?

1. We added tests to ensure the functionality of the hook. Run `yarn vitest run packages/lib/hooks/useParamsWithFallback.test.ts`.
2. To test in real browser, try to call the hook `useParamsWithFallback` in any client component in app router page. You can go to this [branch](https://github.com/intuita-inc/cal.com-demo/tree/intuita/app-router-migration), create any dummy component in app router and call the hook from there.

## Mandatory Tasks

### DIFF
--- a/packages/lib/hooks/useParamsWithFallback.test.ts
+++ b/packages/lib/hooks/useParamsWithFallback.test.ts
@@ -1,0 +1,49 @@
+import { renderHook } from "@testing-library/react-hooks";
+import { vi } from "vitest";
+import { describe, expect, it } from "vitest";
+
+import { useParamsWithFallback } from "./useParamsWithFallback";
+
+describe("useParamsWithFallback hook", () => {
+  it("should return router.query when router.query exists", () => {
+    vi.mock("next/navigation", () => ({
+      useParams: vi.fn().mockReturnValue(null),
+    }));
+
+    vi.mock("react", () => ({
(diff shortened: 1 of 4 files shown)


emrysal (approved):
Looks good to me, lgtm

Issue #10805, opened 2023-08-16
Symmetric encryption keys are weaker than they should be


nicktrn:
### Issue Summary

https://github.com/calcom/cal.com/blob/7c237736365c188e1762691fa4376a5e71daeee5/.env.example#L88-L91

Currently, 24 random bytes are expanded via base64-encoding to match the required AES-256 key length of 32 bytes.

This is effectively AES-192 security (with two additional rounds).

### Recommendations

```diff
diff --git a/.env.example b/.env.example
index 20bef8f8a..209740148 100644
--- a/.env.example
+++ b/.env.example
@@ -87,7 +87,7 @@ CRON_ENABLE_APP_SYNC=false
 
 # Application Key for symmetric encryption and decryption
 # must be 32 bytes for AES256 encryption algorithm
-# You can use: `openssl rand -base64 24` to generate one
+# You can use: `openssl rand -base64 32` to generate one
 CALENDSO_ENCRYPTION_KEY=
 
 # Intercom Config
```

```diff
diff --git a/packages/lib/crypto.ts b/packages/lib/crypto.ts
index d33f0c8f4..f662f22cc 100644
--- a/packages/lib/crypto.ts
+++ b/packages/lib/crypto.ts
@@ -13,7 +13,7 @@ const IV_LENGTH = 16; // AES blocksize
  * @returns Encrypted value using key
  */
 export const symmetricEncrypt = function (text: string, key: string) {
-  const _key = Buffer.from(key, "latin1");
+  const _key = Buffer.from(key, "base64");
   const iv = crypto.randomBytes(IV_LENGTH);
 
   const cipher = crypto.createCipheriv(ALGORITHM, _key, iv);
@@ -30,7 +30,7 @@ export const symmetricEncrypt = function (text: string, key: string) {
  * @param key Key used to decrypt value must be 32 bytes for AES256 encryption algorithm
  */
 export const symmetricDecrypt = function (text: string, key: string) {
-  const _key = Buffer.from(key, "latin1");
+  const _key = Buffer.from(key, "base64");
 
   const components = text.split(":");
   const iv_from_ciphertext = Buffer.from(components.shift() || "", OUTPUT_ENCODING);
```

The key encoding could also be chosen dynamically based on its length and an additional env var `OLD_ENCRYPTION_KEY` provided to migrate old secrets on-demand.

[...]

PR #12043, opened 2023-10-23
fix: env.example requesting 24 bytes instead of 32 bytes encryption key

mattinannt:
## What does this PR do?

The current `.env.example` file is mentioning:
```
# Application Key for symmetric encryption and decryption
# must be 32 bytes for AES256 encryption algorithm
# You can use: `openssl rand -base64 24` to generate one
```
But the command `openssl rand -base64 24` is generating a 24 bytes key (32 characters) instead of a 32 bytes string. To follow the instructions the command should be: `openssl rand -base64 32`

## Requirement/Documentation

## Type of change

## Mandatory Tasks

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -87,7 +87,7 @@ CRON_ENABLE_APP_SYNC=false
 
 # Application Key for symmetric encryption and decryption
 # must be 32 bytes for AES256 encryption algorithm
-# You can use: `openssl rand -base64 24` to generate one
+# You can use: `openssl rand -base64 32` to generate one
 CALENDSO_ENCRYPTION_KEY=
 
 # Intercom Config


PeerRich (approved):

mattinannt:
@PeerRich Can someone with knowledge of how cal.com uses the encryption key check on this? I'm not sure anymore as it also seems to depend on how the key is processed and used in different encryption algorithms.

nicktrn:
> also seems to depend on how the key is processed and used in different encryption algorithms

Quite right. Anyone following the new advice of generating 32 byte keys will actually break the encryption feature.

This would have to be done in conjunction with the proposed changes in #10805.

PR #12044, opened 2023-10-23
chore: [app dir bootstrapping 7] update middleware

dmytro068:
## What does this PR do?

- Updates middleware matcher config to be executed on each page request
- Adds headers required by RootLayout (x-csp, x-url)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

### DIFF
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -5,9 +5,13 @@ import { NextResponse } from "next/server";
 
 import { extendEventData, nextCollectBasicSettings } from "@calcom/lib/telemetry";
 
+import { csp } from "@lib/csp";
+
 const middleware: NextMiddleware = async (req) => {
   const url = req.nextUrl;
   const requestHeaders = new Headers(req.headers);
+
+  requestHeaders.set("x-url", req.url);
 
   if (!url.pathname.startsWith("/api")) {
     //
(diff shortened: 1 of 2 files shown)


apps/web/middleware.ts
```diff
@@ -63,18 +79,8 @@ const routingForms = {
...
-  // https://github.com/vercel/next.js/discussions/42458
-  matcher: [
-    "/:path*/embed",
-    "/api/trpc/:path*",
-    "/login",
-    "/auth/login",
-    /**
-     * Paths required by routingForms.handle
-     */
-    "/apps/routing_forms/:path*",
```
zomars, line 76: Are these removals intended? tagging @hariombalhara who might have more context on this.

DmytroHryshyn: This change was required to ensure that middleware is executed on each page request. This is needed to set the `x-url` and `x-csp` headers that are used in RootLayout. With previous matcher settings we executed middleware only on some specific routes, which resulted in a bug where some app dir pages missed `x-url` header. 
We are open to  discuss alternative solutions. 
Also, we can add app dir pages paths to the matcher one by one as we migrate pages. So for current PR matcher will remain the same.

emrysal (approved):
Looks good to me, but ideally consider changing the req.headers map (when it is one) to a Headers object so you don't have to conditional as much.

emrysal:
@hbjORbj Looks like the embeds tests are consistently failing, could you have look at that?

hbjORbj:
> @hbjORbj Looks like the embeds tests are consistently failing, could you have look at that?

Thank you for catching that.

@DmytroHryshyn is taking care of this issue. Can you provide an ETA?

DmytroHryshyn:
> > @hbjORbj Looks like the embeds tests are consistently failing, could you have look at that?
> 
> Thank you for catching that.
> 
> @DmytroHryshyn is taking care of this issue. Can you provide an ETA?

After rebasing against main, all CI steps are passing 
@emrysal

PR #11959, opened 2023-10-18
refactor: Abstract `createBooking` in `handleNewBooking` [CAL-2619]

joeauyeung:
## What does this PR do?

In anticipation of refactoring `handleSeats` we need to abstract `createBooking` in order to get the return type. This PR is purposely aiming to do one thing so nothing is missed while refactoring `handleNewBooking`

Fixes # (issue)

## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is ab UI/UX design document, please, share it here.

## Type of change

- [x] Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- Are there environment variables that should be set?
- What are the minimal test data to have?
- What is expected (happy path) to have (input and output)?
- Any other important info that could help to test that PR

## Mandatory Tasks

- [ ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)
- My code doesn't follow the style guidelines of this project
- I haven't commented my code, particularly in hard-to-understand areas
- I haven't checked if my PR needs changes to the documentation
- I haven't checked if my changes generate no new warnings
- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #12062, opened 2023-10-24
refactor: Abstract `handleSeats` from `handleNewBooking`

joeauyeung:
## What does this PR do?

This PR is stacked on #11959.

I was going to wait since I wanted to refactor the logic in `handleSeats` to make it more maintainable but because so many things are being changed here I'll leave this to review now.

Since `handleSeats` was declared inside of `handleNewSeats` it relied on a lot of what was already present inside of the handler. This PR aims to abstract `handleSeats` into it's own file so it can be refactored even more and allows us to add unit tests to function. 

This is stacked on 

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- The E2E tests should still continue to pass

## Mandatory Tasks

## Checklist


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -92,23 +93,49 @@ import type { EventResult, PartialReference } from "@calcom/types/EventManager";
...
+};
+export type OriginalRescheduledBooking = Awaited<ReturnType<typeof getOriginalRescheduledBooking>>;
+export type RescheduleReason = Awaited<ReturnType<typeof getBookingData>>["rescheduleReason"];
+export type NoEmail = Awaited<ReturnType<typeof getBookingData>>["noEmail"];
+export type IsConfirmedByDefault = ReturnType<typeof getRequiresConfirmationFlags>["isConfirmedByDefault"];
+export type AdditionalNotes = Awaited<ReturnType<typeof getBookingData>>["notes"];
+export type ReqAppsStatus = Awaited<ReturnType<typeof getBookingData>>["appsStatus"];
+export type PaymentAppData = ReturnType<typeof getPaymentAppData>;
+export type SmsReminderNumber = Awaited<ReturnType<typeof getBookingData>>["smsReminderNumber"];
+export type EventTypeId = Awaited<ReturnType<typeof getBookingData>>["eventTypeId"];
```
sean-brydon, line 135: NIT

Lets save TS a little bit and so something like
```
type AwaitedBookingData = Awaited<ReturnType<typeof getBookingData>>`
export type RescheduleReason = AwaitedBookingData["rescheduleReason"];

Comment (joeauyeung) [reply packages/features/bookings/lib/handleNewBooking.ts:135]: I love this. Thanks for the suggestion

**Review by hariombalhara — DISMISSED** [review]
There are merge conflicts now @joeauyeung 

Btw I am not sure if you are using graphite cli but it would be great for you workflow as you use stacked Prs a lot.

**Review by CarinaWolli — APPROVED** [review]
Looks good 👍

File: packages/features/bookings/lib/handleNewBooking.ts [hunk 2]
```diff
@@ -822,6 +854,76 @@ export function getCustomInputsResponses(
...
 }
 
+/** Updates the evt object with video call data found from booking references
+ *
+ * @param bookingReferences
+ * @param evt
+ *
+ * @returns updated evt with video call data
+ */
+export const addVideoCallDataToEvt = (bookingReferences: BookingReference[], evt: CalendarEvent) => {
```
exception, line 864: Considering other functions already have the `event` name, could we rename this to `addVideoCallDataToEvent`? Two extra characters won't hurt.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -822,6 +854,76 @@ export function getCustomInputsResponses(
...
+      prev[curr.type].success += curr.success;
+      prev[curr.type].errors = prev[curr.type].errors.concat(curr.errors);
+      prev[curr.type].warnings = prev[curr.type].warnings?.concat(curr.warnings || []);
+    } else {
+      prev[curr.type] = curr;
+    }
+    return prev;
+  }, {} as { [key: string]: AppsStatus });
+  resultStatus = Object.values(calcAppsStatus);
+  return resultStatus;
```
exception, line 924: What do you think about something like:

```suggestion
  const calcAppsStatus: { [key: string]: AppsStatus } = reqAppsStatus.concat(resultStatus)
    .reduce((prev, curr) => {
      if (prev[curr.type]) {
        prev[curr.type].success += curr.success;
        prev[curr.type].errors.push(...curr.errors);
        prev[curr.type].warnings?.push(...(curr.warnings || []));
      } else {
        prev[curr.type] = curr;
      }
      return prev;
    }, {});

  return Object.values(calcAppsStatus);
```

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -837,6 +939,52 @@ function getICalSequence(originalRescheduledBooking: BookingType | null) {
   return originalRescheduledBooking.iCalSequence + 1;
 }
 
+export const findBookingQuery = async (bookingId: number) => {
+  const foundBooking = await prisma.booking.findUnique({
```
exception, line 943: Could this not be simplified to a `findUniqueOrThrow`?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2219,7 +2314,7 @@ async function handler(
       );
     }
 
-    addVideoCallDataToEvt(originalRescheduledBooking.references);
+    addVideoCallDataToEvt(originalRescheduledBooking.references, evt);
```
exception, line 2317: Shouldn't this be assigned to a variable? Like the previous use of `addVideoCallDataToEvt`

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+
+  let resultBooking:
+    | (Partial<Booking> & {
+        appsStatus?: AppsStatus[];
+        seatReferenceUid?: string;
+        paymentUid?: string;
+        message?: string;
+        paymentId?: number;
+      })
+    | null = null;
```
exception, line 178: Could we assign a proper type to this and re-use it across the file? Rather than enforcing a type on the function call.

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+
+  const booking = await prisma.booking.findFirst({
```
exception, line 180: findFirstOrThrow?

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+        attendees: {
+          include: {
+            bookingSeat: true,
+          },
+        },
+      },
+    });
+
+    const credentials = await refreshCredentials(allCredentials);
+    const eventManager = new EventManager({ ...organizerUser, credentials });
```
exception, line 241: newTimeSlotBooking should be a findFirstOrThrow and these two constants should be called after.

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+                      bookingId: newTimeSlotBooking.id,
+                    },
+                  },
+                },
+              },
+            })
+          );
+        }
+
+        await Promise.all([
```
exception, line 433: I wonder if we could use prisma transactions here...

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+          where: {
+            id: newTimeSlotBooking.id,
+          },
+          include: {
+            attendees: true,
+            references: true,
+          },
+        });
+
+        if (!updatedNewBooking) {
```
exception, line 455: findUniqueOrThrow, i'll stop now but there's a few recurrences of this.

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+
+        const results = updateManager.results;
+
+        const calendarResult = results.find((result) => result.type.includes("_calendar"));
+
+        evt.iCalUID = Array.isArray(calendarResult?.updatedEvent)
+          ? calendarResult?.updatedEvent[0]?.iCalUID
+          : calendarResult?.updatedEvent?.iCalUID || undefined;
+
+        if (noEmail !== true && isConfirmedByDefault) {
```
exception, line 484: What's this `noEmail !== true` check, can it be explicitly undefined|null?

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+        originalRescheduledBooking = null;
+
+        return null;
+      }
+
+      // Need to change the new seat reference and attendee record to remove it from the old booking and add it to the new booking
+      // https://stackoverflow.com/questions/4980963/database-insert-new-rows-or-update-existing-ones
+      if (seatAttendee?.id && bookingSeat?.id) {
+        await Promise.all([
+          await prisma.attendee.update({
```
exception, line 550: No need to await these if they're already promises :)

packages/features/bookings/lib/handleSeats.ts
```diff
@@ -0,0 +1,796 @@
...
+      include: {
+        attendees: true,
+      },
+      data: {
+        attendees: {
+          create: {
+            email: invitee[0].email,
+            name: invitee[0].name,
+            timeZone: invitee[0].timeZone,
+            locale: invitee[0].language.locale,
```
exception, line 629: Access `invitee[0]` only once outside the update call, it'll be cleaner.

exception (changes requested):

exception (changes requested):
First Review, once these are handled I can do another pass!

Good work!

joeauyeung:
@exception thanks for the review, I think taking the time to refactor these functions also opens up an opportunity to improve our code base.

I know in a number of comments you mention using `findFindOrThrow`. I'm noticing right now we're returning specific error messages when the resource isn't found. If we transitioned to `findFirstOrThrow` wouldn't we have to add a catch statement anyways to send the right error message?

packages/features/bookings/lib/handleNewBooking.ts
joeauyeung: Addressed in https://github.com/calcom/cal.com/pull/12905/commits/649dd1fd0d6df98dfe77424b7713b7bf2aa178e0

packages/features/bookings/lib/handleNewBooking.ts
joeauyeung: Addressed in https://github.com/calcom/cal.com/pull/12905/commits/9bd94947ece8a8f7278d57424e76c0c635a239eb

packages/features/bookings/lib/handleSeats.ts
joeauyeung: This is handled in #12905

packages/features/bookings/lib/handleSeats.ts
joeauyeung: Changing these to prisma transactions

packages/features/bookings/lib/handleSeats.ts
joeauyeung: Fixed in https://github.com/calcom/cal.com/pull/12905/commits/adddd1bb77c5134bca19ae244bcfc9b49c025a15

packages/features/bookings/lib/handleSeats.ts
joeauyeung: Addressed in c6e5853e2cf91b9f3adb8dd3da859b2513b86a27

exception (approved):

PR #12063, opened 2023-10-24
fix: Infinite loop in timezones on the negative side of UTC

emrysal:
## What does this PR do?

Test by:
* Run new tests against old code; observe there's an infinite loop
* Then run tests against updated code.


packages/features/calendars/lib/getAvailableDatesInMonth.test.ts
```diff
@@ -35,5 +35,30 @@ describe("Test Suite: Date Picker", () => {
...
+        const currentDate = new Date();
+        const result = getAvailableDatesInMonth({
+          browsingDate: currentDate,
+        });
+
+        expect(result).toHaveLength(daysInMonth(currentDate) - currentDate.getDate() + 1);
+      }
+      {
+        // test a date in negative UTC offset
+        vi.useFakeTimers().setSystemTime(new Date("2023-10-24T13:27:00.000+07:00"));
```
keithwillcode, line 53: Since `currentDate` is `new Date()`, this won't start failing tomorrow/next month/etc. will it? 😅

keithwillcode (dismissed):

packages/features/calendars/lib/getAvailableDatesInMonth.test.ts
```diff
@@ -35,5 +35,30 @@ describe("Test Suite: Date Picker", () => {
...
+
+        const currentDate = new Date();
+        const result = getAvailableDatesInMonth({
+          browsingDate: currentDate,
+        });
+
+        expect(result).toHaveLength(daysInMonth(currentDate) - currentDate.getDate() + 1);
+      }
+      {
+        // test a date in negative UTC offset
```
keithwillcode, line 52: This should say positive

emrysal, line 52:
```suggestion
        // test a date in positive UTC offset
```

packages/features/calendars/lib/getAvailableDatesInMonth.test.ts
emrysal: If you pass new Date() a full ISO string; it'll always stay equal to ""2023-10-24T13:27:00.000+07:00""

zomars:
Can we check with e2e are failing?

zomars:
Maybe unrelated but I'm getting this test failure locally. cc @CarinaWolli I think I've shared this with you before

image

CarinaWolli:
> Maybe unrelated but I'm getting this test failure locally. cc @CarinaWolli I think I've shared this with you before
> 
> image

@zomars  That's unrelated, investigating right now. Somehow dayjs has some different results depending what tz your system is running on

emrysal:
@zomars not entirely unrelated, turns out after the test the system time wasn't reverted, and some of our e2e tests are failing on the new system time - I'm pushing a fix to revert to real system time for now but it's a problem that that test fails in different system times.

CarinaWolli (dismissed):
Not getting any available slots in October, which I think is also the reason for e2e tests to fail

alishaz-polymath:
The new test passes for me for both old and updated code 😮

hariombalhara:
Same here, but the bug is actually fixed. Trying to see if I can fix the test.

hariombalhara:
@emrysal I found that we need to fake the timezone using TZ env variable in Vitest.

So, I added the concept of timeZone dependent tests and marked these tests as timezone dependent by having an extension `test.timezone.ts`

Doing this I was able to replicate that infinite loop scenario in main with the tests. But even with the fix, I see some expectations mismatch here.

Passing with Asia/Kokata
image

Failing with America/Los_Angeles
Screenshot 2023-10-25 at 1 01 57 PM


Also, feel free to revert the commit, if it's not the right approach. I was debugging and just went ahead with this approach.

emrysal:
@hariombalhara thanks! Turned out the test was broken and casting to UTC - fixed the cast and pushed.

keithwillcode (dismissed):
Updated code changes look good. Will wait for tests to pass to merge.

alishaz-polymath (approved):
Tested Locally, the new test spins in infinite loop with the old code. With the new code, test works fine 🚀

PR #10687, opened 2023-08-10
perf: bookings query improvement

hariombalhara:
## bookings.get handler
In a DB with ~15000 bookings and ~150K even-types, for a user having 900 bookings it takes on average 100ms locally for the bookings.get handler to respond with `main` branch.
In this branch, it takes on average 30ms for the same
I expect improvements to be much better in production with more data and latency b/w DB and servers.

Also added a seeder for performance testing locally(seed-performance-testing.ts). Courtesy of @emrysal 
## Type of change
- [x] Perf Improvement

## How should this be tested?
- Go to /bookings/upcoming and measure the response time of bookings/get endpoint
- Regression Test
	- Test filters with different event-types and users
	- Test that when we do load more of bookings, there are no duplicate bookings

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist
- I haven't checked if my changes generate no new warnings

PR #12068, opened 2023-10-24
fix: prevents prisma idle connections

zomars:
## What does this PR do?

Seems like in PR #10687 the global guard that prevented a lot of idle instances got replaced. Left a comment in [the specific line in here](https://github.com/calcom/cal.com/pull/10687/files#r1370735278).

This PR Makes sure that both the extended and unextended prisma clients are reused a much as possible.

## Type of change

## How should this be tested?

- yarn build
- yarn e2e
- yarn test
- On staging/QA measure the idle connections over time. It should be under the limit.

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/prisma/index.ts
+++ b/packages/prisma/index.ts
@@ -6,9 +6,16 @@ import { bookingReferenceMiddleware } from "./middleware";
 
 const prismaOptions: Prisma.PrismaClientOptions = {};
 
+const globalForPrisma = global as unknown as {
+  prismaWithoutClientExtensions: PrismaClientWithoutExtension;
+  prismaWithClientExtensions: PrismaClientWithExtensions;
+};
+
 if (!!process.env.NEXT_PUBLIC_DEBUG) prismaOptions.log = ["query", "error", "warn"];
 
-const prismaWithoutClientExtensions = new PrismaClientWithoutExtension(prismaOptions);
+// Prevents flooding with idle connections
+const prismaWithoutClientExtensions =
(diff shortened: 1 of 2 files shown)


packages/prisma/index.ts
```diff
@@ -6,9 +6,16 @@ import { bookingReferenceMiddleware } from "./middleware";
...
 const prismaOptions: Prisma.PrismaClientOptions = {};
 
+const globalForPrisma = global as unknown as {
+  prismaWithoutClientExtensions: PrismaClientWithoutExtension;
+  prismaWithClientExtensions: PrismaClientWithExtensions;
+};
+
 if (!!process.env.NEXT_PUBLIC_DEBUG) prismaOptions.log = ["query", "error", "warn"];
 
-const prismaWithoutClientExtensions = new PrismaClientWithoutExtension(prismaOptions);
```
zomars, line 11: This was invoking a new client each time this file is imported.

Introduced here:

https://github.com/calcom/cal.com/pull/10687/files#r1370735278

zomars (commented):
Self review done

keithwillcode (approved):
Wow nice find

hariombalhara:
This is a nice find @zomars. 🙏 looks like I missed it during code cleanup 

But I was thinking, how does it have any impact in production as there a module is imported just once anyway. This should have been an issue on local only right?

zomars:
> This is a nice find @zomars. 🙏 looks like I missed it during code cleanup 
> 
> But I was thinking, how does it have any impact in production as there a module is imported just once anyway. This should have been an issue on local only right?

AFAIK each time the file is imported it would create a new client. The hot reload issue is another one.

PR #12070, opened 2023-10-25
fix: Teams listing and wrong avatars - Cross org boundary scenario

hariombalhara:
## What does this PR do?
- Fix avatar at more places(Scenario-1) Follows up on #12031
- For an org on /teams page, show teams that aren't part of org if the logged in user is part of them.(Scenario-2)

## Type of change

## How should this be tested?
Scenario-1
- Create two users with same username(one in org and one outside). Notice that the avatar for the org user uses non-org user's avatar on Team Assignment page, event-types list. This bug is now fixed.

Scenario-2
- As an org user, make yourself part of a non-org team by directly creating a membership with that team
- Test that you can see the team in /teams when logged in as org member.

## Mandatory Tasks

## Checklist


ThyMinimalDev (approved):
LGTM

PR #12075, opened 2023-10-25
fix: Embed: data-cal-origin not used by modal(which is used by both element click popup and floating button popup)

### LINKED ISSUE #12073: [CAL-2644] Not able to configure origin for element click popup with React snippet.
This is because `data-cal-origin` is unused for element-click-popup

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2644](https://linear.app/calcom/issue/CAL-2644/not-able-to-configure-origin-for-element-click-popup-with-react)</sub>

hariombalhara:
## What does this PR do?

Fixes #12073

## Type of change

## How should this be tested?
Set `data-cal-origin` on the element click popup button and see that the cal link is still being opened from default app.cal.local:3000 and not the provided value

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/embeds/embed-core/src/embed.ts
+++ b/packages/embeds/embed-core/src/embed.ts
@@ -236,7 +236,7 @@ export class Cal {
   }: {
     calLink: string;
     queryObject?: PrefillAndIframeAttrsConfig & { guest?: string | string[] };
-    calOrigin?: string;
+    calOrigin: string | null;
   }) {
     const iframe = (this.iframe = document.createElement("iframe"));
     iframe.className = "cal-embed";
@@ -473,10 +473,12 @@ class CalApi {
     }
 
     config.embedType = "inline";
+    const calConfig = this.cal.getConfig();
 
     const iframe = this.cal.createIframe({
       calLink,
       queryObject: withColorScheme(Cal.getQueryObject(config), containerEl),
+      calOrigin: calConfig.calOrigin,
     });
 
     iframe.style.height = "100%";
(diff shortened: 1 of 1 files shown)


packages/embeds/embed-core/src/embed.ts
```diff
@@ -236,7 +236,7 @@ export class Cal {
   }: {
     calLink: string;
     queryObject?: PrefillAndIframeAttrsConfig & { guest?: string | string[] };
-    calOrigin?: string;
+    calOrigin: string | null;
```
hariombalhara, line 239: Make it required so that it's never missed

packages/embeds/embed-core/src/embed.ts
```diff
@@ -473,10 +473,12 @@ class CalApi {
     }
 
     config.embedType = "inline";
+    const calConfig = this.cal.getConfig();
 
     const iframe = this.cal.createIframe({
       calLink,
       queryObject: withColorScheme(Cal.getQueryObject(config), containerEl),
+      calOrigin: calConfig.calOrigin,
```
hariombalhara, line 481: Even though createIframe fallbacks to it, but it's better to be explcit.

packages/embeds/embed-core/src/embed.ts
```diff
@@ -607,6 +610,7 @@ class CalApi {
       iframe = this.cal.createIframe({
         calLink,
         queryObject,
+        calOrigin: calOrigin || null,
```
hariombalhara, line 613: This is what actually fixes the bug.

PeerRich (approved):

PR #12086, opened 2023-10-25
fix: Attach `credentialId` to location option

joeauyeung:
## What does this PR do?

This PR adds a missing field `credentialId` to event type location options. This credentialId is expected in `handleNewBooking` under `conferenceCredentialId`. For team events, the default host is not always present but it was still falling back to their location option.

Fixes # (issue)

## Type of change

## How should this be tested?

- On the event type setup tab, the location options should contain the `credentialId`
- Install a video app that is scoped to a team
- Set that team installed location to a team event type
- When creating a new booking the `conferenceCredentialId` in `handleNewBooking` should be defined

## Mandatory Tasks

## Checklist


apps/web/components/eventtype/EventSetupTab.tsx
```diff
@@ -382,7 +382,7 @@ export const EventSetupTab = (
                       !validLocations.find((location) => location.type === newLocationType);
 
                     if (canAppendLocation) {
-                      append({ type: newLocationType });
+                      append({ type: newLocationType, credentialId: e.credentialId });
```
Udit-takkar, line 385: I think we should use something like

  ```
 append({
                        type: newLocationType,
                        ...(e.credential && {
                          credentialId: e.credential.id,
                          teamName: e.credential.team?.name,
                        }),
                      });
```

Udit-takkar (changes requested):
@joeauyeung we should also pass teamName like in https://github.com/calcom/cal.com/blob/51fd4102ae15af67e973257e996017d20fc309bb/apps/web/components/dialog/EditLocationDialog.tsx#L360


and we should probably use type credentialId and teamName instead of only credentialId as both credential.id and credential.team.name would be required in EditLocationDialog.tsx and EventSetup.

Udit-takkar (approved):
LGTM. I Fixed everything now

joeauyeung:
> @joeauyeung we should also pass teamName like in
> 
> https://github.com/calcom/cal.com/blob/51fd4102ae15af67e973257e996017d20fc309bb/apps/web/components/dialog/EditLocationDialog.tsx#L360
> 
> and we should probably use type credentialId and teamName instead of only credentialId as both credential.id and credential.team.name would be required in EditLocationDialog.tsx and EventSetup.

Good call thank you!

CarinaWolli (approved):
LGTM!

PR #12101, opened 2023-10-26
fix: padding in footer in profile

Udit-takkar:
Before:-

Screenshot 2023-10-26 at 5 27 25 PM

After:-

Screenshot 2023-10-26 at 5 27 44 PM

### DIFF
--- a/apps/web/public/static/locales/en/common.json
+++ b/apps/web/public/static/locales/en/common.json
@@ -605,7 +605,7 @@
   "hide_book_a_team_member": "Hide Book a Team Member Button",
   "hide_book_a_team_member_description": "Hide Book a Team Member Button from your public pages.",
   "danger_zone": "Danger zone",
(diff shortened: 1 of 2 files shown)


apps/web/public/static/locales/en/common.json
```diff
@@ -605,7 +605,7 @@
   "hide_book_a_team_member": "Hide Book a Team Member Button",
   "hide_book_a_team_member_description": "Hide Book a Team Member Button from your public pages.",
   "danger_zone": "Danger zone",
-  "account_deletion_cannot_be_undone":"Careful. Account deletion cannot be undone.",
+  "account_deletion_cannot_be_undone":"Be Careful. account deletion cannot be undone.",
```
supalarry, line 608: Shouldn't "account" be "Account" to match other locales like `disband_org_confirmation_message` ?

Udit-takkar: @supalarry  Done

PeerRich (approved):

PR #12103, opened 2023-10-26
feat: Stripe paid apps flow

exception:
## Introduce Paid apps to the App Store

Example: https://www.loom.com/share/c27381ddcb1645f7b6c8432f5da12b80?sid=8b0baa67-4e66-4b3b-9446-4f41da33003a

## Important:

> Before this is merged, the API route `/integrations/subscriptions/webhook` needs to be added to Stripe webhook handlers.

### How to add a Paid App:

- Go to the desired app's package in the `app-store` package, eg `cal-ai`.
- Add the following to the `config.json`:
```json
"paid": {
    "price": 25,
    "priceId": "price_1OACLADVtJmy8ddXXgEPHXPo",
    "mode": "subscription" // can be subscription or one_time
 }
```
- Replace the `GET /add` handler with:
```ts

import { defaultResponder } from "@calcom/lib/server";

import checkSession from "../../_utils/auth";
import { withPaidAppRedirect } from "../../_utils/paid-apps";
import appConfig from "../config.json";

export async function getHandler(req: NextApiRequest, res: NextApiResponse) {
  const session = checkSession(req);

  const redirectUrl = await withPaidAppRedirect({
    appPaidMode: appConfig.paid.mode,
    appSlug: appConfig.slug,
    userId: session.user.id,
    priceId: appConfig.paid.priceId,
  });

  if (!redirectUrl) {
    return res.status(500).json({ message: "Failed to create Stripe checkout session" });
  }

  return { url: redirectUrl };
}

export default defaultResponder(getHandler);

```

- Add a `GET /callback` handler:
```ts
export async function getHandler(req: NextApiRequest, res: NextApiResponse) {
  const session = checkSession(req);
  const slug = appConfig.slug;
  const appType = appConfig.type;

  const { checkoutId } = req.query as { checkoutId: string };
  if (!checkoutId) {
    return { url: `/apps/installed?error=${JSON.stringify({ message: "No Stripe Checkout Session ID" })}` };
  }

  const { url } = await withStripeCallback(checkoutId, slug, async ({ checkoutSession }) => {
    const ctx = await createContext({ req, res });
    const caller = apiKeysRouter.createCaller(ctx);

    const apiKey = await caller.create({
      note: "Cal.ai",
      expiresAt: null,
      appId: "cal-ai",
    });

    await checkInstalled(slug, session.user.id);
    await createDefaultInstallation({
      appType,
      userId: session.user.id,
      slug,
      key: {
        apiKey,
      },
      subscriptionId: checkoutSession.subscription?.toString(),
      billingCycleStart: new Date().getDate(),
      paymentStatus: "active",
    });

    await fetch(
      `${process.env.NODE_ENV === "development" ? "http://localhost:3005" : "https://cal.ai"}/api/onboard`,
      {
        method: "POST",
        headers: {
          "Content-Type": "application/json",
        },
        body: JSON.stringify({
          userId: session.user.id,
        }),
      }
    );

    return { url: getInstalledAppPath({ variant: appConfig.variant, slug: "cal-ai" }) };
  });

  return res.redirect(url);
}
```

> Note that within the `withStripeCallback` you get access to the `checkoutSession`, anything within the callback function will be called after the session is retrieved from Stripe's API, this is where app installs should be done. Any errors are handled by the `withStripeCallback` function.


apps/web/components/apps/AppPage.tsx
```diff
@@ -163,6 +165,19 @@ export const AppPage = ({
...
+              {paid && (
+                <>
+                  <Badge className="mr-1">
+                    {Intl.NumberFormat(i18n.language, {
+                      style: "currency",
+                      currency: "USD",
+                      useGrouping: false,
+                      maximumFractionDigits: 0,
+                    }).format(paid.price)}
+                    /mo
```
PeerRich, line 177: i18n needed

apps/web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@calcom/web",
-  "version": "3.4.7",
```
PeerRich, line 3: why?

exception: caching issues for local translations.

keithwillcode: This shouldn't be removed though. If we want to look at modifying https://github.com/calcom/cal.com/blob/main/packages/trpc/server/createNextApiHandler.ts#L67 for local dev, that's a potential solution.

exception: That's fine by me.

packages/app-store/cal-ai/config.json
```diff
@@ -14,5 +13,10 @@
   "isTemplate": false,
   "__createdUsingCli": true,
   "__template": "basic",
-  "dirName": "cal-ai"
+  "dirName": "cal-ai",
+  "paid": {
+    "price": 25,
+    "priceId": "price_1O1ziDH8UDiwIftkDHp3MCTP",
+    "mode": "subscription"
```
hariombalhara, line 20: It is not clear to me here whether it's per month or per year or something else.

packages/ui/components/apps/AppCard.tsx
```diff
@@ -201,6 +205,19 @@ const InstallAppButtonChild = ({
   });
 
   if (!userAdminTeams?.length || !doesAppSupportTeamInstall(appCategories, concurrentMeetings)) {
```
hariombalhara, line 207: We need to fix this check. Currently I am logged in as an admin/owner of an org and `useAdminTeams` has certain items in it. So, first condition is false.

Second condition though I was expecting it to be true is false because by `doesAppSupportTeamInstall`'s logic any app that is not a calendar or video app supports team installation.


I suggest disabling the team installation for any paid app because we haven't implemented that yet(I guess). So, we could just move line 208 - 219 out of this if condition

hariombalhara:
nit and a side note:

Got this error when I didn't have ai service running. Testing it for the first time, it would help to improve the message to say that the service isn't running
image

hariombalhara:
Another side note:

Once I had a failure connecting with ai service, I couldn't reinstall the app because it was marked installed already.

Also, I was shown this useless error message, it should have been clearly informed that the app is already installed.
image

packages/ui/components/apps/AppCard.tsx
hariombalhara: I have fixed it as suggested after verifying that cal.ai doesn't support team install anyways.
https://github.com/calcom/cal.com/blob/5ef737f22ecb9a2fdbc5c11f9883f6b241dda6a4/apps/ai/src/app/api/onboard/route.ts#L11

packages/app-store/_utils/stripe.ts
```diff
@@ -0,0 +1,74 @@
...
+          ...(user.metadata as Prisma.JsonObject),
+          stripeCustomerId: customerId,
+        },
+      },
+    });
+  }
+
+  return customerId;
+}
+
```
ThyMinimalDev, line 70: just a suggestion: what do you think about rewriting this to return early and avoid nesting to make it more readable

```suggestion
export async function getStripeCustomerId(user: UserType): Promise<string> {
  if (
    user?.metadata &&
    typeof user.metadata === "object" &&
    "stripeCustomerId" in user.metadata.
  ) {
    return (user?.metadata as Prisma.JsonObject).stripeCustomerId as string;
  }
  /* We fallback to finding the customer by email (which is not optimal) */
  const customersResponse = await stripe.customers.list({
    email: user.email,
    limit: 1,
  });
  if (customersResponse.data[0]?.id) {
    return customersResponse.data[0].id;
  }
  /* Creating customer on Stripe and saving it on prisma */
  const customer = await stripe.customers.create({ email: user.email });
  await prisma.user.update({
    where: {
      email: user.email,
    },
    data: {
      metadata: {
        ...(user.metadata as Prisma.JsonObject),
        stripeCustomerId: customerId,
      },
    },
  });
  return customer.id;
}

```

packages/app-store/cal-ai/lib/stripe.ts
```diff
@@ -0,0 +1,74 @@
...
+        metadata: {
+          ...(user.metadata as Prisma.JsonObject),
+          stripeCustomerId: customerId,
+        },
+      },
+    });
+  }
+
+  return customerId;
+}
```
ThyMinimalDev, line 69: just noticed this seems like repeated code from the `packages/app-store/_utils/stripe.ts` is it wanted ?

packages/app-store/cal-ai/config.json
```diff
@@ -14,5 +13,10 @@
   "isTemplate": false,
   "__createdUsingCli": true,
   "__template": "basic",
-  "dirName": "cal-ai"
+  "dirName": "cal-ai",
+  "paid": {
+    "price": 25,
+    "priceId": "price_1O1ziDH8UDiwIftkDHp3MCTP",
```
hariombalhara, line 19: Not a blocker: This should be configurable. Need to figure out where it should be read from exactly.

packages/app-store/cal-ai/config.json
```diff
@@ -14,5 +13,10 @@
   "isTemplate": false,
   "__createdUsingCli": true,
   "__template": "basic",
-  "dirName": "cal-ai"
+  "dirName": "cal-ai",
+  "paid": {
+    "price": 25,
```
hariombalhara, line 18: nit: Because it is assumed in USD everywhere, better name could be priceInUsd

packages/app-store/cal-ai/lib/stripe.ts
hariombalhara: I think this module is unused, can be deleted.

packages/app-store/cal-ai/package.json
```diff
@@ -5,7 +5,9 @@
   "version": "0.0.0",
   "main": "./index.ts",
   "dependencies": {
-    "@calcom/lib": "*"
+    "@calcom/lib": "*",
+    "@calcom/prisma": "workspace:^",
```
hariombalhara, line 9: Once we remove the above cal-ai/lib/stripe.ts dependency, we can remove these deps I think.

packages/app-store/utils.ts
```diff
@@ -142,10 +142,19 @@ export function getAppFromLocationValue(type: string): AppMeta | undefined {
...
+  appCategories,
+  concurrentMeetings = undefined,
+  isPaid,
+}: {
+  appCategories: string[];
+  concurrentMeetings: boolean | undefined;
+  isPaid: boolean;
+}) {
+  // Paid apps can't be installed on team level - That isn't supported
+  if (isPaid) {
```
hariombalhara, line 155: Without this some states were getting inconsistent, e.g . Repeat installation of Cal.ai was allowed.

packages/lib/server/getServerErrorFromUnknown.ts
```diff
@@ -55,6 +55,7 @@ export function getServerErrorFromUnknown(cause: unknown): HttpError {
     const redactedCause = redactError(cause);
     return {
       ...redactedCause,
+      message: redactedCause.message,
```
hariombalhara, line 58: `message` was also not enumerable. Fixed that.

packages/lib/piiFreeData.ts
```diff
@@ -105,7 +105,7 @@ export function getPiiFreeUser(user: {
   allowDynamicBooking?: boolean | null;
   defaultScheduleId?: number | null;
   organizationId?: number | null;
-  credentials?: Credential[];
+  credentials?: Partial<Credential>[];
```
hariombalhara, line 108: Fixes Type Error.

apps/web/components/apps/InstallAppButtonChild.tsx
```diff
@@ -46,15 +48,35 @@ export const InstallAppButtonChild = ({
...
+    !doesAppSupportTeamInstall({ appCategories, concurrentMeetings, isPaid: !!paid })
+  ) {
     return (
       <Button
         data-testid="install-app-button"
         {...props}
         // @TODO: Overriding color and size prevent us from
         // having to duplicate InstallAppButton for now.
         color="primary"
+        disabled={shouldDisableInstallation}
```
hariombalhara, line 79: It is an existing bug in production where the button to install is enabled even if the app is installed already. Though backend still rejects the installation.

packages/app-store/cal-ai/config.json
exception: wdym? Peer asked for it to be in the config

apps/web/pages/api/integrations/subscriptions/webhook.ts
hariombalhara, line 1: I am not sure when this endpoint is hit. I deleted this file and tried a complete install and uninstall and faced no error.

hariombalhara:
Screenshot 2023-11-10 at 5 05 49 PM
On refresh on apps/cal-ai page I see this hydration error. Not a blocker as such, but would be good if it can be fixed.

apps/web/pages/api/integrations/subscriptions/webhook.ts
exception: It's a Stripe webhook, to test it you need to install the stripe cli, enable forwarding and try again.

exception: It's in the PR description.

hariombalhara:
Tested:
- Installation  - Only works after payment after tweaking the trialEndDate. Within the trial range, it can be installed without payment. [Pass]
- Uninstallation works [Pass]
- Repeat installation of app is restricted [Pass]
- App can't be installed on a team/org [Pass]
- Testing failed payment - Stripe doesn't let go further the payment page if there is some issue[Pass]


Great work @exception 🙏 I have left a few comments that need to be addressed though. Feel free to followup on non blocking comments

packages/app-store/cal-ai/config.json
hariombalhara: I mean that in general while supporting paid apps, self hosters might want to configure the price id as it would be different for them. So, not an immediate requirement.

hariombalhara: Even when I was testing locally, I had to change it as per my stripe account and I have to remember to not commit that.

hariombalhara (changes requested):
There are 1-2 comments that need to be addressed before merge

apps/web/pages/api/integrations/subscriptions/webhook.ts
hariombalhara: https://www.loom.com/share/65c1fb811d3c4d17bfe4437de88bea06

I think webhook flow isn't being used as I tested even after deleting the module and it works. Maybe stripe checkout doesn't need webhooks at all.

Also, I haven't setup stripe webhook on my new machine(that I got a few days back :) ) still it works

exception: Yeah, it's a webhook, it's not sync. It's forwarded from Stripe to our API to make sure we know the current state of a subscription, if someone cancels their subscription, it uninstalls the app, and makes sure we keep a record of that.

hariombalhara (changes requested):
1. [Blocking] I was able to test the webhook now but I found that it's not able to do what it's supposed to do.
This is because similar webhook handling code in main access the subscription as [event.data.object](https://github.com/calcom/cal.com/blob/main/packages/features/ee/payments/api/webhook.ts#L50) but we are accessing it as [`event.data`](https://github.com/calcom/cal.com/blob/feat/paid-apps/apps/web/pages/api/integrations/subscriptions/webhook.ts#L20). So, `subscription.id` comes as undefined and it ends up targeting wrong credential in the table.

2. As we are doing type assertion there, we should validate that it has subscription.id atleast so that it doesn't endup targeting wrong credential accidentally.

3. I am not sure how we would be able to distinguish b/w an App Subscription update/delete and a Booking Payment update/delete. AFAIU, both the webhook endpoints would be hit when one of the subscription is updated in Stripe. I think, it is not an issue at the moment because` features/ee/payments/api/webhook` and  `api/integrations/subscriptions/webhook` are using mutually exclusive events but it could potentially cause problems when there is some overlap in that.

exception:
@hariombalhara Addressed.

apps/web/pages/api/integrations/subscriptions/webhook.ts
```diff
@@ -0,0 +1,116 @@
...
+    where: {
+      subscriptionId: subscription.id,
+    },
+  });
+
+  if (!app) {
+    throw new HttpCode({ statusCode: 202, message: "Received and discarded" });
+  }
+
+  // should we delete the credential here rather than marking as inactive?
```
hariombalhara, line 57: It could be drastic to uninstall the app in this case. I think we should mark it inactive only but prevent using the app if it's inactive.

hariombalhara (approved):
LGTM now !! 

Feel free to merge. Also, there are some comments that haven't been acknowledged yet. It would be great if those can be acknowledged and ticket can be created if needed.

PR #12106, opened 2023-10-26
fix: pull managed event type bookings in zapier

CarinaWolli:
## What does this PR do?

Managed event type bookings weren't pulled when testing zapier trigger (listBookings endpoint). This worked with API keys but not with OAuth


alishaz-polymath (approved):
🚀

PR #12108, opened 2023-10-26
fix: cron scheduleEmailRemider time out

CarinaWolli:
## What does this PR do?

Use Promise.allSettled instead of await each request to sendgrid


packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -26,6 +26,7 @@ const sendgridAPIKey = process.env.SENDGRID_API_KEY as string;
 const senderEmail = process.env.SENDGRID_EMAIL as string;
 
 sgMail.setApiKey(sendgridAPIKey);
+client.setApiKey(sendgridAPIKey);
```
CarinaWolli, line 29: API key for SendGrid client wasn't set and I am seeing `Unauthorized` errors in production. However, locally this doesn't seem to be an issue

keithwillcode (approved):

PR #12111, opened 2023-10-26
fix: adding managed event type to workflow

CarinaWolli:
## What does this PR do?

Fixes that managed event type couldn't be added to a user workflow. It didn't throw any error, but the managed event type just got ignored. 

## Type of change

## How should this be tested?

- Create managed event type and assign yourself
- Create a new user workflow
- Add managed event type in the dropdown
- Click save
- Go back to workflow and check if managed event type is checked in dropdown

### DIFF
--- a/packages/trpc/server/routers/viewer/workflows/update.handler.ts
+++ b/packages/trpc/server/routers/viewer/workflows/update.handler.ts
@@ -86,13 +86,16 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
 
   const hasOrgsPlan = IS_SELF_HOSTED || ctx.user.organizationId;
 
+  const where: Prisma.EventTypeWhereInput = {};
+  where.id = {
+    in: activeOn,
+  };
+  if (userWorkflow.teamId) {
+    //all children managed event types are added after
+    where.parentId = null;
+  }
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -86,13 +86,16 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
+  if (userWorkflow.teamId) {
+    //all children managed event types are added after
+    where.parentId = null;
+  }
   const activeOnEventTypes = await ctx.prisma.eventType.findMany({
-    where: {
-      id: {
-        in: activeOn,
-      },
-      parentId: null,
```
CarinaWolli, line 94: here we excluded all managed event type children, we only want to do that for team workflows

supalarry:
I wanted to test this locally but don't see `Click save` button. Is there some setting I need to enable?

Screenshot 2023-10-27 at 11 22 55

Udit-takkar (approved):
LGTM.

Udit-takkar:
> I wanted to test this locally but don't see `Click save` button after adding managed event type to workflow. Is there some setting I need to enable?
> 

@supalarry looks like you need a dev license key in .env file.

PR #12113, opened 2023-10-27
fix: impersonation for orgs

sean-brydon:
Fixes impersonation in an org context:

How to test:
Org admins/owners can impersonate anyone in the org if they have impersonation enabeld
Org Team owners/admins can impersonate anyone in THAT org team if they have impersonation enabled 

Normal team/admin impersonation should remain the same

### DIFF
--- a/packages/features/ee/impersonation/lib/ImpersonationProvider.ts
+++ b/packages/features/ee/impersonation/lib/ImpersonationProvider.ts
@@ -72,6 +72,67 @@ export function checkPermission(session: Session | null) {
   }
 }
 
+async function getImpersonatedUser({
+  session,
+  teamId,
+  creds,
+}: {
+  session: Session | null;
+  teamId: number | undefined;
+  creds: Credentials | null;
+}) {
+  let TeamWhereClause: Prisma.MembershipWhereInput = {
+    disableImpersonation: false, // Ensure they have impersonation enabled
+    accepted: true, // Ensure they are apart of the team and not just invited.
+    team: {
+      id: teamId, // Bring back only the right team
+    },
+  };
+
+  // If you are an admin we dont need to follow this flow -> We can just follow the usual flow
+  // If orgId and teamId are the same we can follow the same flow
+  if (session?.user.org?.id && session.user.org.id !== teamId && session?.user.role !== "ADMIN") {
+    TeamWhereClause = {
+      disableImpersonation: false,
+      accepted: true,
+      team: {
+        id: session.user.org.id,
+      },
+    };
+  }
+
+  // Get user who is being impersonated
+  const impersonatedUser = await prisma.user.findFirst({
+    where: {
+      OR: [{ username: creds?.username }, { email: creds?.username }],
+    },
+    select: {
+      id: true,
+      username: true,
+      role: true,
+      name: true,
+      email: true,
+      organizationId: true,
+      disableImpersonation: true,
+      locale: true,
+      teams: {
+        where: TeamWhereClause,
+        select: {
+          teamId: true,
+          disableImpersonation: true,
+          role: true,
+        },
+      },
+    },
+  });
+
+  if (!impersonatedUser) {
+    throw new Error("This user does not exist");
+  }
+
+  return impersonatedUser;
+}
+
 const ImpersonationProvider = CredentialsProvider({
   id: "impersonation-auth",
   name: "Impersonation",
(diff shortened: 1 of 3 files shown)


sean-brydon:
* main
  * **PR #12113** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12113" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12113?utm_source=stack-comment).

packages/features/ee/teams/components/MemberListItem.tsx
```diff
@@ -321,7 +321,7 @@ export default function MemberListItem(props: Props) {
               onSubmit={async (e) => {
                 e.preventDefault();
                 await signIn("impersonation-auth", {
-                  username: props.member.username || props.member.email,
+                  username: props.member.email,
```
sean-brydon, line 324: Always use emails since usernames are now none unique

ThyMinimalDev: we are still using username here: [https://github.com/calcom/cal.com/blob/10-27-fix_impersonation-orgs/packages/features/users/components/UserTable/ImpersonationMemberModal.tsx#L30](https://github.com/calcom/cal.com/blob/10-27-fix_impersonation-orgs/packages/features/users/components/UserTable/ImpersonationMemberModal.tsx#L30
)
is it wanted?

Udit-takkar (commented):
~~how to impersonate as an org owner ? I don't see impersonation tab item in sidebar~~
Nevermind got it

packages/features/ee/teams/components/MemberListItem.tsx
sean-brydon: Good spot

sean-brydon:
> @sean-brydon how to impersonate as an org owner ? I don't see impersonation tab item in sidebar
> 
> Screenshot 2023-10-27 at 5 35 37 PM

its located in the org members table -> three ∂ots and impersonate

packages/features/ee/impersonation/lib/ImpersonationProvider.ts
```diff
@@ -85,6 +86,26 @@ const ImpersonationProvider = CredentialsProvider({
...
+        accepted: true,
+        team: {
+          id: session.user.org.id,
+        },
+      };
+    }
+
     // Get user who is being impersonated
     const impersonatedUser = await prisma.user.findFirst({
       where: {
```
ThyMinimalDev, line 111: do we still need the or statement to check the username here if we only pass the email now ?

sean-brydon: I would say yes - you can still pass a username from the admin side of things. This only doesnt work if there is a conflict with user names.

packages/features/ee/impersonation/lib/ImpersonationProvider.ts
```diff
@@ -85,6 +86,26 @@ const ImpersonationProvider = CredentialsProvider({
...
+        disableImpersonation: false,
+        accepted: true,
+        team: {
+          id: session.user.org.id,
+        },
+      };
+    }
+
     // Get user who is being impersonated
     const impersonatedUser = await prisma.user.findFirst({
```
ThyMinimalDev, line 110: what do you think about putting this logic in a function to scope it out of the main handler ? I see something like this 

```
const getImpersonatedUser = async ({
  user,
  creds,
  teamId,
}: {
  user?: Session["user"];
  teamId?: number;
  creds: Record<"teamId" | "username", string> | undefined;
}) => { ... }

....

const impersonatedUser = getImpersonatedUser(...)
 ```

File: packages/features/ee/impersonation/lib/ImpersonationProvider.ts [hunk 4]
```diff
@@ -85,6 +86,26 @@ const ImpersonationProvider = CredentialsProvider({
     checkUserIdentifier(creds);
     checkPermission(session);
```
ThyMinimalDev, line 88: In ImpersonationProvider I don't see any  error thrown  if `session` is undefined but the typing says it can be ? can it actually happen a user reaches this point without having a session?

sean-brydon: It would get caught down the trea when looking for the session user but yeah deffo worth adding an a guard

packages/features/ee/impersonation/lib/ImpersonationProvider.ts
```diff
@@ -63,11 +64,75 @@ export function checkUserIdentifier(creds: Partial<Credentials>) {
 }
 
 export function checkPermission(session: Session | null) {
-  if (session?.user.role !== "ADMIN" && process.env.NEXT_PUBLIC_TEAM_IMPERSONATION === "false") {
+  if (
+    (session?.user.role !== "ADMIN" && process.env.NEXT_PUBLIC_TEAM_IMPERSONATION === "false") ||
```
ThyMinimalDev, line 68: probably a good spot to use `!Boolean(process.env.NEXT_PUBLIC_TEAM_IMPERSONATION)` instead of `process.env.NEXT_PUBLIC_TEAM_IMPERSONATION === "false"`

because: Boolean("true") => true and  Boolean("false") = false and Boolean(undefined) = false

ThyMinimalDev (approved):

PR #12122, opened 2023-10-27
feat: Add consistent iCalUID

joeauyeung:
## What does this PR do?

This is a part of stacked PRs that aim to refactor how we handle `.ics` files.

This PR adds the `iCalUID` and the `iCalSequence` to the booking record. This is a part of a stack PRs that aims to unify the `.ics` file that we sent.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Set GCal as your destination calendar
- Create a booking
    - The `iCalUID` should be `${uid}@cal.com` in the DB and the .ics file sent
    - The `iCalSequence` should be set to 0
- Reschedule the booking
    - In the new booking, the `iCalUID` should be the same as the original booking
    - The `iCalSequence` should be set to 1
- Reschedule that booking a second time
    - The sequence should be set to 2
- Cancel that booking
    - The sequence should be 3
-  Now set Outlook as your destination calendar and create a booking
    - The `iCalUID` should come from Outlook as you can not set the `icalUID` through the API

## Mandatory Tasks

## Checklist


packages/prisma/migrations/20231027144522_add_ical_columns_to_booking/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - Added the required column `iCalUID` to the `Booking` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "Booking" ADD COLUMN     "iCalSequence" INTEGER NOT NULL DEFAULT 0,
```
keithwillcode, line 8: These adds are out of order compared to the code. Intentional?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2189,6 +2203,7 @@ async function handler(
       );
     } else {
       const calendarResult = results.find((result) => result.type.includes("_calendar"));
+      console.log("🚀 ~ file: handleNewBooking.ts:2195 ~ calendarResult:", calendarResult);
```
keithwillcode, line 2206: Was this intentionally left here?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1099,6 +1099,16 @@ async function handler(
...
 
+  const iCalUID = originalRescheduledBooking?.iCalUID ?? `${uid}@cal.com`;
+  // For bookings made before introducing iCalSequence, assume that the sequence should start at 1. For new bookings start at 0.
+  const iCalSequence = originalRescheduledBooking?.iCalSequence
+    ? originalRescheduledBooking.iCalSequence + 1
+    : originalRescheduledBooking
+    ? 1
+    : 0;
+
+  console.log("🚀 ~ file: handleNewBooking.ts:1105 ~ iCalSequence:", iCalSequence);
```
keithwillcode, line 1110: Was this intentionally left here?

keithwillcode (changes requested):
Left some comments. But we should also add unit tests to this to confirm the logic for calculating the UID and sequence.

packages/prisma/migrations/20231027144522_add_ical_columns_to_booking/migration.sql
keithwillcode: To clarify - The order specified throughout the code is `iCalUID` and then `iCalSequence`.  See prisma schema.

This alter table statement has them in the reverse order.

packages/prisma/migrations/20231027144522_add_ical_columns_to_booking/migration.sql
```diff
@@ -0,0 +1,9 @@
...
+ADD COLUMN     "iCalUID" TEXT NOT NULL;
```
keithwillcode, line 9: Why is this a TEXT column and not VARCHAR?

packages/prisma/migrations/20231027144522_add_ical_columns_to_booking/migration.sql
joeauyeung: This file was generated using `prisma migrate`. I can fix this and moving forward ensure that the adds line up with the schema.

packages/prisma/migrations/20231027144522_add_ical_columns_to_booking/migration.sql
joeauyeung: This file was generated using `prisma migrate`. I can change this and moving forward I'll change the column type to `VARCHAR` when appropriate.

packages/prisma/schema.prisma
```diff
@@ -417,6 +417,8 @@ model Booking {
   /// @zod.custom(imports.bookingMetadataSchema)
   metadata              Json?
   isRecorded            Boolean              @default(false)
+  iCalUID               String               @db.VarChar()
```
joeauyeung, line 420: @keithwillcode add `@db.VarChar()` to the schema maps the column to `VARCHAR` since Prisma will default map the column to `TEXT` for Postgres.

https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#default-type-mappings

packages/emails/templates/attendee-was-requested-to-reschedule-email.ts
```diff
@@ -42,6 +43,8 @@ export default class AttendeeWasRequestedToRescheduleEmail extends OrganizerSche
   // @OVERRIDE
   protected getiCalEventAsString(): string | undefined {
     const icsEvent = createEvent({
+      uid: this.calEvent.iCalUID || this.calEvent.uid!,
```
exception, line 46: `nitpick`: I'm tending to start telling people to use nullish coalescing, eg:
```suggestion
      uid: this.calEvent.iCalUID ?? this.calEvent.uid!,
```

exception: Also, do we have no guarantee of `this.calEvent.uid`?

packages/emails/templates/attendee-was-requested-to-reschedule-email.ts
```diff
@@ -42,6 +43,8 @@ export default class AttendeeWasRequestedToRescheduleEmail extends OrganizerSche
...
+      sequence: 100,
```
exception, line 47: `question-non-blocking`: What's sequence 100?

packages/emails/templates/attendee-was-requested-to-reschedule-email.ts
joeauyeung: It's interesting. I've seen this in our codebase before. Looking at the schema, `uid` is required when saving bookings. Although the `CalendarEvent` type we often use in the code base sets `uid` as possibly undefined.

packages/emails/lib/getICalUID.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (event?.iCalUID) return event.iCalUID;
+
+  if (defaultToEventUid && event?.uid) return `${event.uid}@${APP_NAME}`;
+
+  if (uid) return `${uid}@${APP_NAME}`;
+
+  const translator = short();
+
+  uid = translator.fromUUID(uuidv5(APP_NAME, uuidv5.URL));
+  return `${uid}@${APP_NAME}`;
```
emrysal, line 33: I'm a big fan of this PR; but this may break badly (I believe an iCAL uid HAS to be a 'valid' email address, whether it exists or not. But I think it's not a given that `APP_NAME` is actually a domain like `Cal.com`.

emrysal (dismissed):

packages/emails/lib/getICalUID.ts
joeauyeung: It doesn't have to be a valid email address. It is recommended and I think it would be cool to have some of our branding for people that dig into this stuff.

Outlook's iCalUID it generates does not have a domain.

https://icalendar.org/New-Properties-for-iCalendar-RFC-7986/5-3-uid-property.html

packages/prisma/migrations/20231113202947_add_ical_columns_to_booking/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "Booking" ADD COLUMN     "iCalSequence" INTEGER NOT NULL DEFAULT 0,
+ADD COLUMN     "iCalUID" VARCHAR DEFAULT '';
```
keithwillcode, line 3: @joeauyeung In the case where you want this field to be unlimited, I don't think we need to go with VARCHAR since TEXT has already been the standard. My suggestion was to consider putting a max length on this field since it's not just free text.

joeauyeung: I'll change this back to TEXT. Although there's no limit on the length of the iCalUID I don't want to set an upper limit just in case there's a calendar provider that sets long ids.

packages/prisma/schema.prisma
```diff
@@ -431,6 +431,8 @@ model Booking {
   /// @zod.custom(imports.bookingMetadataSchema)
   metadata              Json?
   isRecorded            Boolean              @default(false)
+  iCalUID               String?              @default("")
```
hariombalhara, line 434: nit: Why are we giving it a default of empty string? It is marked to accept NULL values already by having ?.

Semantically, it makes more sense to not set the value or keep it NULL(because an empty iCalUID isn't a valid iCalUID anyway). AFAIU, a booking should have iCalUID set going forward always.

hariombalhara:
Test Results
1. Do a fresh booking -> ics file received by organizer has an untranslated string `emailed_you_and_any_other_attendees`
image

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1101,6 +1102,17 @@ async function handler(
...
+    event: { iCalUID: originalRescheduledBooking?.iCalUID, uid: originalRescheduledBooking?.uid },
+    uid,
+  });
+  // For bookings made before introducing iCalSequence, assume that the sequence should start at 1. For new bookings start at 0.
+  const iCalSequence = originalRescheduledBooking?.iCalSequence
+    ? originalRescheduledBooking.iCalSequence + 1
+    : originalRescheduledBooking
+    ? 1
+    : 0;
+
```
hariombalhara, line 1115: nit: It deserves it's own function IMHO.

function getIcalSequence(originalRescheduledBooking) {
      // Fresh booking
	if (!originalRescheduleBooking) {
	   return 0;
	}
	// Reschedule but iCalSequence not available. So, it should atleast be 1, assume 1.
	if (!originalRescheduleBooking.iCalSequence) { 
	  return 1;
	}
	
	return originalRescheduleBooking.iCalSequence + 1;
}

hariombalhara (dismissed):
Tested and I am seeing some failures in testing.

packages/prisma/schema.prisma
joeauyeung: I also agree that with the way this PR is setup all bookings will have a iCalUID. There were some instances where if we were rescheduling or working with bookings without an iCalUID, it would throw a Prisma error.

joeauyeung:
> **Test Results**
> 
> * **[Fail]** Do a fresh booking  -> ics file received by **organizer** has an untranslated string `emailed_you_and_any_other_attendees`
>   image
> * **[Pass]** Rescheduling a booking correctly updates it in Apple Calendar when I open the ics file. Also, ics file correctly increases the **sequence**.
> 
> ## Requires Confirmation
> * **[Fail]** Do a booking for an event **that requires confirmation**. On confirming the booking from dashboard, the ics files sent in confirmation emails don't have @Cal.com UIDs.
> * **[Fail]** Also, rescheduling such a booking endup having sending different UID in ics file and thus **duplicate events** get created.
>   image
> 
> ## Recurring Bookings
> * **[Pass]** A fresh Recurring booking has correct ics uid
> * **[Fail, On production too?]** Cancelling a single booking for a recurring event, cancels all the events in Apple Calendar when I open the ics file.
>   -[ Found weird bug with rescheduling unrelated to this change](https://github.com/calcom/cal.com/issues/12461)
> 
> **Compatiblity Tests**
> 
> * **[Fail]** Request Reschedule for an existing booking(that has default iCalUID of empty string in table), sends the ics file with UID without @Cal.com - _Is it expected ?_ I think there shouldn't be any functionality issue though it's just that we would still be seeing ics files without @Cal.com UIDs in production for some time.
> 
> @joeauyeung Can we run some manual tests for seated events as well and post results here if they are working fine.

- Fixed the requires confirmation event types. 
- Recurring events when creating, rescheduling, and canceling the whole series
    - @CarinaWolli is following up on rescheduling individual events in a series so we can build off of that
- For seated events it works when attendees are hidden but needs some rework when attendees are displayed for everyone. Since I'm working on refactoring the handleSeats function can this wait for that PR?

hariombalhara:
@joeauyeung If we are not breaking anything in seated bookings, it can certainly be done later.

keithwillcode (approved):
Talked with @joeauyeung and this has been tested thoroughly.

hariombalhara (approved):

PR #12018, opened 2023-10-20
perf: Adding 'stage 1' of avatar refactor, wip

emrysal:
## What does this PR do?

Needs tests.

PR #12080, opened 2023-10-25
fix/profile-dont-wait-for-avatar

sean-brydon:
Fixes: https://threads.com/thread/34544006411/34544006409?s=kFN7CaE1EA6bZRgApH3KNY

Loading avatar was blocking - instead we just show the fallback avatar until it loaded instead of blocking the whole profile form

PR #12143, opened 2023-10-30
fix: Prevent possible reason behind avatar infinite redirect

### LINKED ISSUE #12102: [CAL-2655] Avatars not showing correctly
### Issue Summary

The profile avatar and organisation avatars are not showing correctly on the left sidebar even after multiple attempts of saving it. Once you upload and save your avatar the left sidebar just has some broken image placeholder rather than the recently uploaded image. I don't have any plan currently active and I don't think not having a plan should be doing these.

### Steps to Reproduce

1. Go to settings
2. Go to profile(your account or organisation)
3. Upload a new image for the avatar and save it

### Actual Results

- The image is uploaded and saved successfully but the left sidebar shows a broken image icon.

### Expected Results

- Should be showing the latest uploaded image as the avatar.

### Technical details

- Tested on different accounts
- Tested on safari and brave browser (latest versions)
- Happens on live version of cal.com 

### Evidence

https://github.com/calcom/cal.com/assets/94699055/a0fa8228-e778-4a54-b261-286a1cbb09b0

<sub>[CAL-2655](https://linear.app/calcom/issue/CAL-2655/avatars-not-showing-correctly)</sub>

hariombalhara:
## What does this PR do?

Fixes #12102

[Before and After fix loom](https://www.loom.com/share/abb4cc151916423e850a32bbbf4c4326)

**NOTE:** Because avatar is stored wrongly in DB, a re-upload is required.

**More important note:** As avatar endpoint is cached by Cloudflare/Vercel, even though it would fix the avatar in DB, it would probably not fix the issue immediately. After the cache expires, which is probably a day, the problem would automatically solve for users who have re-uploaded the avatar.

The caching thing is supposed to be solved in a better way in #12018 or a followup of it.

## Type of change

## How should this be tested?

See loom

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
@@ -238,6 +238,7 @@ async function getAvatarToSet(avatar: string | null | undefined) {
     } else {
       // Non Base64 avatar currently could only be the dynamic avatar URL(i.e. /{USER}/avatar.png). If we allow setting that URL, we would get infinite redirects on /user/avatar.ts endpoint
       log.warn("Non Base64 avatar, ignored it", { avatar });
+      // `undefined` would not ignore the avatar, but `null` would remove it. So, we return `undefined` here.
       return undefined;
     }
   }
--- a/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
@@ -22,6 +22,7 @@ import { TRPCError } from "@trpc/server";
 import { getDefaultScheduleId } from "../viewer/availability/util";
 import { updateUserMetadataAllowedKeys, type TUpdateProfileInputSchema } from "./updateProfile.schema";
 
+const log = logger.getSubLogger({ prefix: ["updateProfile"] });
 type UpdateProfileOptions = {
   ctx: {
     user: NonNullable<TrpcSessionUser>;
@@ -35,6 +36,7 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
   const userMetadata = handleUserMetadata({ ctx, input });
   const data: Prisma.UserUpdateInput = {
     ...input,
+    avatar: await getAvatarToSet(input.avatar),
     metadata: userMetadata,
   };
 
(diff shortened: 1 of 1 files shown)


hariombalhara:
* main
  * **PR #12143** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12143" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12143?utm_source=stack-comment).

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -61,12 +63,6 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
       }
     }
   }
-  if (input.avatar) {
-    data.avatar = await resizeBase64Image(input.avatar);
-  }
-  if (input.avatar === null) {
-    data.avatar = null;
-  }
```
hariombalhara, line 69: It was confusing because data initially gets avatar from input as is and then we try to modify the data.

Instead of modifying we should set the correct value initially.

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -234,3 +230,21 @@ const handleUserMetadata = ({ ctx, input }: UpdateProfileOptions) => {
...
+      // `undefined` would not ignore the avatar, but `null` would remove it. So, we return `undefined` here.
+      return undefined;
+    }
+  }
+
+  if (avatar === null) {
+    return null;
+  }
+  return avatar;
+}
```
ThyMinimalDev, line 250: I think we can rewrite this function to be more clear by returning early and avoiding nesting

```
if avatar is null => return null

if avatar does not start with data:image =>  return undefined

return await resizeBase64Image(avatar);

```

ThyMinimalDev (changes requested):

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
hariombalhara: Done

Udit-takkar (commented):
I don't see "remove" button now when we have avatar set.  maybe not related to this PR

Screenshot 2023-10-30 at 2 06 24 PM

hariombalhara:
> I don't see "remove" button now when we have avatar set. maybe not related to this PR
> 
> Screenshot 2023-10-30 at 2 06 24 PM

Yeah I noticed that too. I though it was like that only. It's in prod as well.

Udit-takkar (approved):
LGTM

Udit-takkar:
> Yeah I noticed that too. I though it was like that only. It's in prod as well.

Ok I'll fix this in a follow up PR

ThyMinimalDev (approved):

hariombalhara:
Possible regression from here https://github.com/calcom/cal.com/pull/12080/files

PeerRich (approved):

PR #12154, opened 2023-10-30
fix: improve error message in add location

Udit-takkar:
Fixes: 

1) Add Placeholder

Screenshot 2023-10-30 at 8 51 17 PM

2) Better Error message Hint

Screenshot 2023-10-30 at 8 51 32 PM


sean-brydon (dismissed):
Nice job - tested and works well.

NIT: we could add i18n to the error but i dont mind this being in a follow up as its not too common we do it with error messages

Udit-takkar:
@sean-brydon done

PeerRich (approved):
awesome!

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -321,6 +322,47 @@ const EventTypePage = (props: EventTypeSetupProps) => {
...
+                    if (!valid) {
+                      ctx.addIssue({
+                        code: z.ZodIssueCode.custom,
+                        path: [eventLocationType?.defaultValueVariable ?? "link"],
+                        message: `Invalid URL`,
+                      });
+                    }
+                  }
+                  return;
+                })
```
Comment (PeerRich) [inline apps/web/pages/event-types/[type]/index.tsx:365]: can you explain what all of this is? looks pretty funky

Comment (Udit-takkar) [reply apps/web/pages/event-types/[type]/index.tsx:365]: @PeerRich this code is from editLocationDialog.tsx. we are just validating the url here . We already have regex of some location options that vaildates if url is valid and if it not then displaying error message with the sampleUrl that we are expecting

Comment (PeerRich) [reply apps/web/pages/event-types/[type]/index.tsx:365]: do we wanna extract this into a helper function if we use it in two places?

Comment (PeerRich) [reply apps/web/pages/event-types/[type]/index.tsx:365]: or is this the code we removed but are re-adding?

Comment (Udit-takkar) [reply apps/web/pages/event-types/[type]/index.tsx:365]: actually There are some differences like this line isn't required in EditLocationDialog

` const eventLocationType = getEventLocationType(val.type);
` 
 and path ( path: [eventLocationType?.defaultValueVariable ?? "link"]) is also not required in EditLocationDialog.

Comment (PeerRich) [reply apps/web/pages/event-types/[type]/index.tsx:365]: so we are ok merging this?

Comment (Udit-takkar) [reply apps/web/pages/event-types/[type]/index.tsx:365]: yeah

sean-brydon (approved):
Tested locally - works well. Nice job!

PR #12155, opened 2023-10-30
feat: Add Sentry

keithwillcode:
## What does this PR do?

Adds support to send server exceptions to Sentry. No edge nor client support is added.

## Type of change

## How should this be tested?

- Trigger any error from the server-side and ensure it lands in Sentry (this has already been tested in our own Sentry account)

## Mandatory Tasks


apps/web/sentry.server.config.ts
```diff
@@ -0,0 +1,6 @@
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+  tracesSampleRate: 0,
```
keithwillcode, line 5: We will not be adding support for performance tracing right now.

packages/trpc/server/middlewares/captureErrorsMiddleware.ts
```diff
@@ -9,6 +11,7 @@ const captureErrorsMiddleware = middleware(async ({ next }) => {
     if (!cause) {
       return result;
     }
+    Sentry.captureException(cause);
```
keithwillcode, line 14: If the NEXT_PUBLIC_SENTRY_DSN is not set, this is a no-op. This is the same strategy the API project currently has so I decided to follow suit.

I considered adding a logger factory that returns a logger specific to Sentry when NEXT_PUBLIC_SENTRY_DSN is set, but decided against it because it introduces more classes and layering that seems necessary since this operation is harmless.

Happy to reconsider if someone feels strongly about it.

apps/web/sentry.edge.config.ts
```diff
@@ -0,0 +1 @@
+export {};
```
keithwillcode, line 1: This is only required for type checks

apps/web/sentry.client.config.ts
```diff
@@ -0,0 +1 @@
+export {};
```
keithwillcode, line 1: This is only required for type checks

apps/web/sentry.server.config.ts
exception: Worth this being an env var as well? In case we need to change in the future.

exception (dismissed):
LGTM.

hariombalhara (changes requested):
LGTM!! Except that yarn.lock seems to have a lot of deleted lines(most probably because of missing auth, console, website modules checked out locally) so that needs to be fixed probably.

keithwillcode:
> LGTM!! Except that yarn.lock seems to have a lot of deleted lines(most probably because of missing auth, console, website modules checked out locally) so that needs to be fixed probably.

Oh yikes nice spot, @hariombalhara

apps/web/sentry.server.config.ts
keithwillcode: I was thinking we'd add it in the future when we want to enable it. This was meant to intentionally disable it

yarn.lock
```diff
@@ -3545,12 +3545,12 @@ __metadata:
     "@types/react-dom": ^18.0.9
     eslint: ^8.34.0
     eslint-config-next: ^13.2.1
-    next: ^13.5.4
+    next: ^13.4.6
```
keithwillcode, line 3548: Hmm, looking into why this went down

keithwillcode: Resolved

hariombalhara (approved):

PR #12156, opened 2023-10-30
fix: two failing test cases that were incorrect

emrysal:
## What does this PR do?

* Fixes the failing test cases due to DST aware code. The error was in the test cases so not a big issue, but it broke on DST change.

### DIFF
--- a/packages/lib/date-ranges.test.ts
+++ b/packages/lib/date-ranges.test.ts
@@ -5,12 +5,11 @@ import dayjs from "@calcom/dayjs";
 import { buildDateRanges, processDateOverride, processWorkingHours, subtract } from "./date-ranges";
 
 describe("processWorkingHours", () => {
-  // TEMPORAIRLY SKIPPING THIS TEST - Started failing after 29th Oct
-  it.skip("should return the correct working hours given a specific availability, timezone, and date range", () => {
+  it("should return the correct working hours given a specific availability, timezone, and date range", () => {
     const item = {
       days: [1, 2, 3, 4, 5], // Monday to Friday
-      startTime: new Date(Date.UTC(2023, 5, 12, 8, 0)), // 8 AM
-      endTime: new Date(Date.UTC(2023, 5, 12, 17, 0)), // 5 PM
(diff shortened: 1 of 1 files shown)


packages/lib/date-ranges.test.ts
```diff
@@ -5,12 +5,11 @@ import dayjs from "@calcom/dayjs";
 import { buildDateRanges, processDateOverride, processWorkingHours, subtract } from "./date-ranges";
 
 describe("processWorkingHours", () => {
-  // TEMPORAIRLY SKIPPING THIS TEST - Started failing after 29th Oct
-  it.skip("should return the correct working hours given a specific availability, timezone, and date range", () => {
+  it("should return the correct working hours given a specific availability, timezone, and date range", () => {
     const item = {
       days: [1, 2, 3, 4, 5], // Monday to Friday
-      startTime: new Date(Date.UTC(2023, 5, 12, 8, 0)), // 8 AM
-      endTime: new Date(Date.UTC(2023, 5, 12, 17, 0)), // 5 PM
```
emrysal, line 13: Date.UTC is cast to local time upon passing to new Date()

packages/lib/date-ranges.test.ts
```diff
@@ -48,33 +47,32 @@ describe("processWorkingHours", () => {
...
-    const dateTo = dayjs().month(10).endOf("month");
+    const dateFrom = dayjs.utc().month(10).date(1).startOf("day");
+    const dateTo = dayjs.utc().month(10).endOf("month");
 
     const results = processWorkingHours({ item, timeZone, dateFrom, dateTo });
 
     const allDSTStartAt12 = results
       .filter((res) => res.start.isBefore(firstSundayOfNovember))
-      .every((result) => result.start.utc().hour() === 12);
+      .every((result) => result.start.utc().hour() === 11);
```
emrysal, line 72: Here the problem was that the hour would change based on the system time, fixed it by ensuring the firstSundayOfNovember variable is defined in UTC. Technically 11,12 would work right now as well - but this would break on the next DST change.

PR #12168, opened 2023-10-31
fix(e2e): failsafe if no availabilities

ThyMinimalDev:
## What does this PR do?

This pr fixes e2e tests that are trying to select slots on the schedule the last day of the month, if there are no availabilities at all left during the month, we should click on the View Next Month button.

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/playwright/fixtures/regularBookings.ts
+++ b/apps/web/playwright/fixtures/regularBookings.ts
@@ -1,5 +1,7 @@
 import { expect, type Page } from "@playwright/test";
 
+import dayjs from "@calcom/dayjs";
+
 import type { createUsersFixture } from "./users";
 
 const reschedulePlaceholderText = "Let others know why you need to reschedule";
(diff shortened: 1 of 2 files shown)


emrysal:
No longer needed as we fixed the underlying issue :)

keithwillcode:
@emrysal @ThyMinimalDev Should we keep this fix around just for the case of running the test suite at, let's say, 11:58PM? Or cross that bridge when it happens?

ThyMinimalDev:
> @emrysal @ThyMinimalDev Should we keep this fix around just for the case of running the test suite at, let's say, 11:58PM? Or cross that bridge when it happens?

we could, we have unit tests to ensure that the core logic is working, we should probably make specific e2e tests as well where there is actually no slots at all and assess that

ThyMinimalDev:
we actually need a check for the end of the month on the e2e tests, otherwise tests will fail every end of month after 5pm since our seeded users don't have availabilities after that time

keithwillcode (approved):

PR #12170, opened 2023-10-31
refactor: Generating `.ics` files

joeauyeung:
## What does this PR do?

This PR refactors how our `.ics` files are generated for a more cohesive experience for our attendees.

https://www.loom.com/share/4b1526e8ae9f42899bfc443ef4e4c2fc

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/prisma/migrations/20231027144522_add_ical_columns_to_booking/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - Added the required column `iCalUID` to the `Booking` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "Booking" ADD COLUMN     "iCalSequence" INTEGER NOT NULL DEFAULT 0,
+ADD COLUMN     "iCalUID" TEXT NOT NULL;
```
keithwillcode, line 9: We should use VARCHAR here because TEXT columns are historically known to be worse in terms of storage and perf. Maybe there's a max we know this will be? 255?

packages/emails/lib/generateIcsString.ts
```diff
@@ -0,0 +1,142 @@
...
+      break;
+  }
+
+  const getTextBody = (title = "", subtitle = "emailed_you_and_any_other_attendees"): string => {
+    let body: string;
+    if (BookingAction.RequestReschedule && role === "attendee") {
+      body = `
+      ${t(title)}
+      ${getWhen(event, t)}
+      ${t(subtitle)}`;
```
exception, line 88: `question-non-blocking`: Since `subtitle` can sometimes already be a translated string, would calling `t()` again on it cause any issues?

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -391,7 +391,8 @@ async function handler(req: CustomRequest) {
       data: {
         status: BookingStatus.CANCELLED,
         cancellationReason: cancellationReason,
-        iCalSequence: evt.iCalSequence,
+        // Assume that canceling the booking is the last action
+        iCalSequence: evt.iCalSequence || 100,
```
exception, line 395: `nitpick`: Nullish coalescing, could 0 be explicitly set?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1099,7 +1100,7 @@ async function handler(
   const calEventUserFieldsResponses =
     "calEventUserFieldsResponses" in reqBody ? reqBody.calEventUserFieldsResponses : null;
 
-  const iCalUID = originalRescheduledBooking?.iCalUID ?? `${uid}@cal.com`;
+  const iCalUID = originalRescheduledBooking?.iCalUID ?? `${uid}@${ICAL_UID_DOMAIN}`;
```
exception, line 1103: `suggestion`: I've seen this string being computed a few times throughout the PR, thoughts on abstracting it to a function w/ documentation so it can be easily reused and documented?

packages/features/bookings/lib/handleNewBooking/test/fresh-booking.test.ts
```diff
@@ -192,6 +191,7 @@ describe("handleNewBooking", () => {
               meetingUrl: "https://UNUSED_URL",
             },
           ],
+          iCalUID: `${createdBooking.uid}@${ICAL_UID_DOMAIN}`,
```
exception, line 194: `nitpick`: Same as before.

packages/emails/lib/generateIcsString.ts
joeauyeung: Before this PR I see that in `organizer-requested-to-reschedule-email` on line 38, a `title` that is a translated string is passed into `getTextBody` which passes it into another `t()`. Testing it we see that the text output is expected.

CleanShot 2023-11-01 at 20 06 37

packages/features/bookings/lib/handleCancelBooking.ts
joeauyeung: When calendars parse `.ics` files they look at the sequence for the latest update. This sequence starts at 0. In the case of cancelling bookings, it should be the last update for the event, so I wanted to fall back to a sequence number that would most likely be the highest.

packages/features/bookings/lib/handleNewBooking.ts
joeauyeung: @exception thoughts on this commit? f4a7377d283208f1c46519b7c14f786f0ef302e9

.env.example
```diff
@@ -257,3 +257,5 @@ E2E_TEST_OIDC_USER_EMAIL=
 E2E_TEST_OIDC_USER_PASSWORD=
 
 # ***********************************************************************************************************
+
+ICAL_UID_DOMAIN="cal.com"
```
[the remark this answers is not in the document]
hariombalhara: We can probably reuse one of these https://github.com/calcom/cal.com/blob/0091ecf572c063e899ec664e2e3f6aa6e1c6499c/.env.example#L192 instead of adding a new env variable.

apps/web/test/utils/bookingScenario/expects.ts
```diff
@@ -160,14 +161,14 @@ expect.extend({
...
-      };
-    }
+    // if (!expectedEmail.noIcs && !isIcsUIDExpected) {
+    //   return {
+    //     pass: false,
+    //     actual: JSON.stringify(icsObject),
+    //     expected: expectedEmail.ics?.iCalUID,
+    //     message: () => `Expected ICS UID ${isNot ? "is" : "isn't"} present in actual`,
+    //   };
+    // }
```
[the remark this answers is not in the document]
hariombalhara: I don't think we should.  Even if `generateIcsString` has no bug there could be bug in passing the arguments to it.

So, this verifies more of an integration/e2e scenario. So, it should be there IMHO and infact we should add more tests like `reschedule` should have the same iCalUID as in original booking

packages/features/bookings/lib/handleNewBooking.ts
exception: Looks good!

exception (approved):

hariombalhara:
@joeauyeung Does this PR fix any known issue at the moment? or we plan to stack the fixes over it?
I want to take this change slow if it's just a refactor, adding as many unit tests/integration tests getting ~100% confidence in the changes.

**Integration Tests** - BookingScenario tests
- Have we added all ics related tests that could be impacted by this flow ?

** Unit Tests **
- Should we add unit tests for more functions. I see `getICalUid` and `generateIcsString` tests already there

**E2E**
- Should we add some e2e tests that might help in testing some scenario not reliably covered by integration tests?

packages/core/EventManager.ts
```diff
@@ -179,7 +179,7 @@ export default class EventManager {
       }
       const isCalendarType = isCalendarResult(result);
       if (isCalendarType) {
-        evt.iCalUID = result.iCalUID || undefined;
+        evt.iCalUID = result.iCalUID || event.iCalUID || undefined;
```
hariombalhara, line 182: What's the reasoning behind this logic? 
I am not a fan of adding fallbacks like this as in some scenarios where we expect iCalUID to to be returned(and if it doesn't return) it can wrongly do the fallback. I would prefer to fail in such cases so that the issue can be detected.

1. When do we expect result.iCalUID to be set? Should it be set by a calendar app? If yes, may be we can error if it's not set by Calendar App.
2. What is the scenario when we expect result.iCalUID to be not set but event.iCalUID would be set?
3. What is the scenario when both of them will not be set for a calendar app? Shouldn't we error out in that case?

hariombalhara: If we know that a particular scenario shouldn't happen but might be happening in production, it might be better to log it as an error so that we can investigate how many times it's happening and in which case exactly.

packages/emails/lib/generateIcsString.ts
```diff
@@ -0,0 +1,142 @@
...
+  const icsRole: ParticipationRole = "REQ-PARTICIPANT";
+  if (event.recurringEvent?.count) {
+    // ics appends "RRULE:" already, so removing it from RRule generated string
+    recurrenceRule = new RRule(event.recurringEvent).toString().replace("RRULE:", "");
+  }
+
+  switch (bookingAction) {
+    case BookingAction.Create:
+      if (role === "organizer") {
+        title = event.recurringEvent?.count ? "new_event_scheduled_recurring" : "new_event_scheduled";
```
hariombalhara, line 45: Currently bookingScenario tests don't check the title in ics file but they do test it in Email content, we can quickly add the support to verify this in ics file. We would just need to modify `toHaveEmail` function as it already has all the appropriate data.

hariombalhara: We could do it as a followup as well if it takes more time just in case as there are unit tests for this function which adds some confidence already.

packages/emails/templates/attendee-location-change-email.ts
```diff
@@ -1,12 +1,19 @@
...
+import generateIcsString, { BookingAction } from "../lib/generateIcsString";
 import AttendeeScheduledEmail from "./attendee-scheduled-email";
 
 export default class AttendeeLocationChangeEmail extends AttendeeScheduledEmail {
   protected getNodeMailerPayload(): Record<string, unknown> {
     return {
       icalEvent: {
         filename: "event.ics",
-        content: this.getiCalEventAsString(),
+        content: generateIcsString({
```
hariombalhara, line 10: With the current logic of `generateIcsString`, it's adding similar logic as in it's caller to derive title, subtitle, and status.
e.g. 


How about we change `generateIcsString` signature to accept a `title`, `subtitle` and `status` as well. 

That way that logic stays together in the caller. `generateIcsString` wouldn't need to do switches over bookingAction and would be simpler.

packages/emails/lib/generateIcsString.ts
```diff
@@ -0,0 +1,142 @@
...
+        });
+      }
+      status = "CANCELLED";
+      break;
+    case BookingAction.LocationChange:
+      title = "event_location_changed";
+      break;
+  }
+
+  const getTextBody = (title = "", subtitle = "emailed_you_and_any_other_attendees"): string => {
```
hariombalhara, line 82: We shouldn't have a default value here. The parameters should be explicit and required, as forgetting to pass them would end up send email with wrong content.

packages/emails/lib/generateIcsString.ts
```diff
@@ -0,0 +1,142 @@
...
+    let body: string;
+    if (BookingAction.RequestReschedule && role === "attendee") {
```
hariombalhara, line 84: It shouldn't be it's responsibility to manage content based on action. We should just pass the content.
If we need to pass a function then we can do that too. We could just let `generateIcsString` accept a fn called getTextBody.

packages/lib/CalEventParser.ts
```diff
@@ -176,7 +176,11 @@ export const getRescheduleLink = (calEvent: CalendarEvent): string => {
   return `${calEvent.bookerUrl ?? WEBAPP_URL}/reschedule/${seatUid ? seatUid : Uid}`;
 };
 
-export const getRichDescription = (calEvent: CalendarEvent, t_?: TFunction /*, attendee?: Person*/) => {
+export const getRichDescription = (
```
hariombalhara, line 179: Can we make it accept an object with these props to improve readability at the place where we are calling it.

packages/features/bookings/lib/handleNewBooking/test/fresh-booking.test.ts
```diff
@@ -161,6 +161,7 @@ describe("handleNewBooking", () => {
         });
 
         const createdBooking = await handleNewBooking(req);
+        const iCalUID = getICalUID({ uid: createdBooking.uid });
```
hariombalhara, line 164: I don't like the idea of using `getICalUID` in the test. It is the part of code being tested, so it shouldn't be used in test ideally. It could cause a situation like 'not-working' === 'not-working' is true so it's working.

Because we are controlling the scenario here, we should be knowing exactly what id was chosen to be iCalUID among `event.iCalUID`, `event.uid`, `createdBooking.uid`, or a brand new ID. So, we should use that value here directly to verify.

packages/features/bookings/lib/handleNewBooking/test/fresh-booking.test.ts
```diff
@@ -140,7 +141,6 @@ describe("handleNewBooking", () => {
         const calendarMock = mockCalendarToHaveNoBusySlots("googlecalendar", {
           create: {
             id: "MOCKED_GOOGLE_CALENDAR_EVENT_ID",
-            iCalUID: "MOCKED_GOOGLE_CALENDAR_ICS_ID",
```
hariombalhara, line 143: We shouldn't remove this. By removing this, AFAIU, we are mocking a scenario where iCalUID isn't returned by Google Calendar which it normally does.

It could be a different test where Google Calendar doesn't return iCalUID, so we could copy this test.

hariombalhara (changes requested):
Thanks @joeauyeung for the PR !! It is a bold attempt to do such a refactor 🙏

I have left some comments to be addressed. As I commented earlier, I would like it to be done the right way with ~100% confidence in the changes that they aren't introducing any bugs or unwanted behaviour.

packages/emails/templates/attendee-location-change-email.ts
joeauyeung: Implemented this feedback and updated tests in 63936a0b4726b38acbcb4d46d2045b3f937f11cf and 0c91dc1e0ae07a0c4d274212b9092cdfd3a6d6b4

packages/features/bookings/lib/handleNewBooking/test/fresh-booking.test.ts
joeauyeung: When passing a iCalUID to Google Calendar, we already know what it is. I've made some changes to the code where it won't return the iCalUID from Google Calendar because we were the ones that set it.

packages/emails/lib/generateIcsString.ts
joeauyeung: With `getTextBody` I remember making changes to emails and having to make the change in every email class. Requesting reschedule is the only email where it differs. Could we have a default `getTextBody` we can use and if a function is passed then use that to generate the text body?

packages/features/bookings/lib/handleNewBooking/test/fresh-booking.test.ts
hariombalhara: I don't see those changes in this PR. But quickly checked and I guess they are part of the target branch in PR https://github.com/calcom/cal.com/pull/12122/files#diff-6bfbaf63c841dc9343dba0c7487e30a546ffe2a63c9d9418379d0dbb88644d27.
So, it should be fine then.

packages/emails/lib/getICalUID.ts
```diff
@@ -0,0 +1,36 @@
...
+  uid?: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  event?: { iCalUID?: string | null; uid?: string | null; [key: string]: any };
+  defaultToEventUid?: boolean;
+}) => {
+  if (event?.iCalUID) return event.iCalUID;
+
+  if (defaultToEventUid && event?.uid) return `${event.uid}@${APP_NAME}`;
+
+  if (uid) return `${uid}@${APP_NAME}`;
```
hariombalhara, line 28: I think we should lowercase APP_NAME so that it is abc@cal.com instead of abc@cal.com

packages/features/bookings/lib/handleNewBooking/test/fresh-booking.test.ts
hariombalhara: Fixed it cc @joeauyeung

packages/features/bookings/lib/handleCancelBooking.ts
hariombalhara: THis comment is important enough to go inline in the code.

packages/core/EventManager.ts
hariombalhara: In that case I think we should have a clear way to configure what such apps are and we can do `doesCalendarSupportStoringCustomCalUid()`. So, if a calendar supports that, we  don't use result.iCalUID and if it does we use that. That makes the logic very clear and it would be easier to discover bugs. 

I dread a scenario when iCalUID is still being passed for Google Calendar but it isn't what we expect, so we should have strict checks everywhere.

packages/emails/lib/generateIcsString.ts
hariombalhara: For long term I think no, because with which params `generateIcsString` is called and whether it's called aren't tested. That can only be tested by integration tests or e2e tests.

But for now it's fine, we can work on #12214 later.

packages/emails/lib/generateIcsString.ts
hariombalhara: Yeah maybe we can define a utility `getTextBody` and we can pass it from all places to `generateIcsString` except the rescheduling case which would have different `getTextBody`. In this way `generateIcsString` doesn't care about rescheduling and it is just concerned about generating an consistent ics string.

hariombalhara:
Thanks @joeauyeung for quickly considering and quickly addressing the feedback 🙏

There are some comments still open, but I have merged it as it would make more sense to review all the changes together before going into main.  Let's address that feedback in #12122

PR #12172, opened 2023-10-31
fix: booking limits

CarinaWolli:
## What does this PR do?

Fixes availability with booking limits when booking falls on a different day in UTC. 

Fixes the following example: 

- Availability (Australia/Brisbane):
Screenshot 2023-11-02 at 15 36 06

- Create event type with booking limit 1 per day

- Book any day at 16:00 (Australia/Brisbane)

- Check availability: 
Before: Availability on that day + next day partially blocked 
After: Full day of the day of the booking is shown as unavailable, next day is not effected


packages/lib/server/checkBookingLimits.ts
```diff
@@ -11,14 +11,15 @@ import { parseBookingLimit } from "../isBookingLimits";
 export async function checkBookingLimits(
   bookingLimits: IntervalLimit,
   eventStartDate: Date,
-  eventId: number
+  eventId: number,
+  timezone?: string | null
```
emrysal, line 15: nit: should be timeZone

Abdel-Monaam-Aouini (approved):

exception (approved):
Looks ok to me, works as intended locally following PR description.

PR #12174, opened 2023-11-01
feat: configure availability time picker interval by env variable

### LINKED ISSUE #12189: Allow configurable time intervals on Availability for a Schedule
- Currently,  you can set an event to have a meeting length of any time.  In our case, 10 minutes.
- But availability time options in the Schedule can only be selected in 15 minute intervals. 
- In this scenario,  availability is desired to start to coincide with these 10 minute intervals.  
  - ie- booking availability is desired to start from 9.10am to match to event booking slots.
  - setting a available time of 9am allows a booking time too early (ie-9am) to be chosen , and an available time of 9.15 misses the first possible appointment of 9.10am, so first booking can only be made at 9.20.

- Add an optional environment variable or global settings option for changing the availability time options.    So starting times can be more precise than every 15 mintues.
  - This allows availability to match the event-type length settings 
- Falls back to default 15 minutes if no env set.
- Potential changes that worked locally: https://github.com/calcom/cal.com/compare/main...Shpadoinkle:cal.com:feat/schedule-10-min-interval

- we haven't found a way around this with the current event/schedule options.

### Requirement/Document

(Share it here.)

---
##### House rules

Shpadoinkle:
## What does this PR do?
- Add an optional environment variable to changing the availability time options.    So starting times can be more precise than every 15 mintues.
- Falls back to default 15 minutes if no env set.
- Add the env variable `NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL`
  - Just followed the format of some of the other variables listed, feel free to change.

Fixes #12189
https://github.com/calcom/cal.com/issues/12189

## Issue
- Currently,  you can set an event to have a meeting length of any time.  In our case, 10 minutes.
- But availability time options in the Schedule can only be selected in 15 minute intervals. 
- In this scenario,  availability is desired to start to coincide with these 10 minute intervals.  
  - ie- bookings can start from 9.10am
  - setting a available time of 9am allows a booking time too early (ie-9am), and an availabile time of 9.15 misses the first  possible appointment of 9.10am.

## Type of change

## How should this be tested?
- set your env variable to the desired amount,  in our case we just wanted 10 minutes.
- A users availability options should now be able to be set using 10 minute intervals instead of 15.
With a booking
- setting an event-type with 10 minute intervals
- then setting a user with a starting time of 9.10 or any combination of start times that go off 10 mins instead of 15/

## Mandatory Tasks

## Checklist


packages/lib/slots.ts
```diff
@@ -169,7 +169,7 @@ function buildSlotsWithDateRanges({
       ? range.start
       : startTimeWithMinNotice;
 
-    let interval = 15;
+    let interval = Number(process.env.NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL) || 15;
```
Shpadoinkle, line 172: I assumed this 15 minute interval likely relates to the 15 minutes on availability.

apps/web/next.config.js
```diff
@@ -30,6 +30,10 @@ if (!process.env.NEXTAUTH_URL && process.env.NEXT_PUBLIC_WEBAPP_URL) {
 if (!process.env.NEXT_PUBLIC_WEBSITE_URL) {
   process.env.NEXT_PUBLIC_WEBSITE_URL = process.env.NEXT_PUBLIC_WEBAPP_URL;
 }
+if (!process.env.NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL) {
```
emrysal, line 33: This isn't needed 🙏

packages/lib/slots.ts
emrysal: I'm not sure this interval does anything tbh - just a good default; @CarinaWolli could you have a look at this line?

emrysal (dismissed):
few changes needed, waiting for a bit of feedback from @CarinaWolli

packages/lib/slots.ts
CarinaWolli: The interval setting ensures that available slots start only at specific times: on the hour, half-hour, or at 20-minute, 10-minute, or the default 15-minute intervals, depending on the event length.

Example: 
- NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL = 1
- Event length: 8 mins
- Availability starts at 9:37 => Yes, we only allowed 15-minute intervals before, but the incoming date range could still start at a random time because of a calendar busy time or the current day's availability is also usually random.

Before: 
First slot: 9:45 => starting by default at the next 15 minute slot

With these changes: 
First slot: 9:37

Shpadoinkle: Ok, @CarinaWolli   if our desired config is to use NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL = 10. 

And our desired first availability slot is to start at 9.10am.   How would this behave?  
Or would not changing this line at all still result in a 9.10 first booking?

apps/web/next.config.js
Shpadoinkle: removed

packages/lib/slots.ts
Shpadoinkle: or are you saying that leaving this at 15 would have no impact?      ie-  event length of 10mins, and a users availability start time of 9.10 would show up correctly with 9.10 as the first available booking time?

Shpadoinkle: @CarinaWolli  or are you saying that leaving this at 15 would have no impact? ie- event length of 10mins, and a users availability allows 10 minute intervals and starts their day at 9.10am. it would show up correctly with 9.10am as the first available booking time?

CarinaWolli: > Ok, @CarinaWolli if our desired config is to use NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL = 10.
> And our desired first availability slot is to start at 9.10am. How would this behave?
> Or would not changing this line at all still result in a 9.10 first booking?

In this case, we could leave it as 15. Because 10 is part of the `intervalsWithDefinedStartTimes` array. But if your NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL=1 and you have an availability starting at 8:13, then the first slot would not be 8:13 but 8:15.

I think we should keep your changes like that, it makes sense to me

CarinaWolli (approved):
Looks good to me, @emrysal's feedback got addressed as well 🙏

PR #12183, opened 2023-11-01
fix: unhandled promise error in scheduleEmailReminders endpoint

CarinaWolli:
## What does this PR do?

- Fixes 'Unhandled Promise Rejection'
- Add more comments to improve readability 
- Move getiCalEventAsString to separate file for better readability of endpoint

## Requirement/Documentation

## Type of change

### DIFF
--- a/packages/features/ee/workflows/api/scheduleEmailReminders.ts
+++ b/packages/features/ee/workflows/api/scheduleEmailReminders.ts
@@ -107,7 +107,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
 
   const pageSize = 90;
   let pageNumber = 0;
-  const deletePromises = [];
+  const deletePromises: Promise<any>[] = [];
 
   //delete batch_ids with already past scheduled date from scheduled_sends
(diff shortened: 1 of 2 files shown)


packages/features/ee/workflows/api/scheduleEmailReminders.ts
```diff
@@ -130,14 +62,14 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
       break;
     }
 
-    deletePromises.push(
-      remindersToDelete.map((reminder) =>
```
CarinaWolli, line 134: Resulted in a two-dimensional array and this caused issues with Promise.allSettled not working correctly

Udit-takkar (approved):
NIT:  use logger.error instead of console.log

```
 Promise.allSettled(allPromisesCancelReminders).then((results) => {
    results.forEach((result) => {
      if (result.status === "rejected") {
        console.log(`Error cancelling scheduled_sends: ${result.reason}`);
      }
    });
  });
```
  
code LGTM

PR #12194, opened 2023-11-02
fix: Make sure if header is set already, it isn't set again

hariombalhara:
## What does this PR do?

Fixes the following error. Though as far as I tested, the error arises only when there was already an error in the callback endpoint

Screenshot 2023-11-02 at 10 26 12 AM

In case of zohocalendar, the error was coming when this line was executed https://github.com/calcom/cal.com/blob/0091ecf572c063e899ec664e2e3f6aa6e1c6499c/packages/app-store/zohocalendar/api/callback.ts#L58. It returned a non nullish value causing the bug in `defaultResponder`.

## Type of change

## How should this be tested?
- Go to http://app.cal.local:3000/api/integrations/zohocalendar/callback?state=%7B%22returnTo%22%3A%22http%3A%2F%2Fapp.cal.local%3A3000%2Fapps%2Finstalled%22%7D&code=1000.86ee71b55ddb790a73d28b386032e781.1190f4c17b3b5fce09eae6121c8fc4b9&location=us and observe the header already set error. Shouldn't come with this branch

## Mandatory Tasks

## Checklist


hariombalhara:
* main
  * **PR #12194** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12194" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12194?utm_source=stack-comment).

packages/lib/server/defaultResponder.ts
```diff
@@ -13,7 +13,9 @@ export function defaultResponder<T>(f: Handle<T>) {
       performance.mark("Start");
       const result = await f(req, res);
       ok = true;
-      if (result) res.json(result);
+      if (result && !res.writableEnded) {
```
hariombalhara, line 16: There could be cases where header is already set and result is also set. e.g.` res.redirect` unlike `res.json` returns a non nullish value and thus we would endup setting `json` in that case as well which would cause 'header already set error'

Udit-takkar (commented):
Wait is the endpoint api/integrations/zoho_calendar or  api/integrations/zohocalendar ?

I see api/integrations/zoho_calendar when installing app

Screenshot 2023-11-02 at 5 42 27 PM

packages/lib/server/defaultResponder.ts
keithwillcode: Nice find. Unit testable?

keithwillcode (approved):

PR #12213, opened 2023-11-03
fix: team booking page having same slug as the org

hariombalhara:
## What does this PR do?

Fixes the bug where depending on which team would be fetched first by prisma(which is usually last-modified first), sometimes Org is served for a team page because of slug confusion

Also added a repository for team to kickoff moving to repository pattern for querying. 

## Motivation to move to the pattern along with this bug fix
- The issue arose because we can't query the team and org by slug reliably because slugs aren't unique in table, so we need well tested code to handle that querying thing
- Going forward we plan to move `metadata.isOrganization` to a column so that querying is faster and more reliable and then we would just need to modify the repository code.

## Type of change

## How should this be tested?

## Mandatory Tasks

### DIFF
--- a/packages/lib/server/queries/teams/index.ts
+++ b/packages/lib/server/queries/teams/index.ts
@@ -1,7 +1,12 @@
+import type { Membership } from "@prisma/client";
 import { Prisma } from "@prisma/client";
+import type { z } from "zod";
 
 import { getAppFromSlug } from "@calcom/app-store/utils";
-import { getOrgFullOrigin, getSlugOrRequestedSlug } from "@calcom/ee/organizations/lib/orgDomains";
+import {
+  getOrgFullOrigin,
+  whereClauseForOrgWithSlugOrRequestedSlug,
+} from "@calcom/ee/organizations/lib/orgDomains";
 import prisma, { baseEventTypeSelect } from "@calcom/prisma";
 import { SchedulingType } from "@calcom/prisma/enums";
 import { EventTypeMetaDataSchema, teamMetadataSchema } from "@calcom/prisma/zod-utils";
@@ -9,6 +14,7 @@ import { EventTypeMetaDataSchema, teamMetadataSchema } from "@calcom/prisma/zod-
 import { WEBAPP_URL } from "../../../constants";
 import logger from "../../../logger";
 
+const log = logger.getSubLogger({ prefix: ["queries", "teams"] });
 export type TeamWithMembers = Awaited<ReturnType<typeof getTeamWithMembers>>;
 
 export async function getTeamWithMembers(args: {
@@ -21,9 +27,12 @@ export async function getTeamWithMembers(args: {
   /**
    * If true, means that you are fetching an organization and not a team
    */
-  isOrgView?: boolean;
+  getOrgOnly?: boolean;
 }) {
-  const { id, slug, userId, orgSlug, isTeamView, isOrgView, includeTeamLogo } = args;
+  const { id, slug, userId, orgSlug, isTeamView, getOrgOnly: isOrgView, includeTeamLogo } = args;
+
+  // This should improve performance saving already app data found.
+  const appDataMap = new Map();
   const userSelect = Prisma.validator<Prisma.UserSelect>()({
     username: true,
     email: true,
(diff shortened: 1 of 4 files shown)


hariombalhara:
* main
  * **PR #12213** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12213" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12213?utm_source=stack-comment).

packages/lib/server/repository/team.ts
```diff
@@ -0,0 +1,149 @@
...
+  isOrg: boolean;
+  teamSelect: TeamSelect;
+};
+
+const log = logger.getSubLogger({ prefix: ["repository", "team"] });
+
+/**
+ * Get's the team or organization with the given slug or id reliably along with parsed metadata.
+ */
+async function getTeamOrOrg<TeamSelect extends Prisma.TeamSelect>({
```
hariombalhara, line 42: Prisma doesn't make it easy to wrap generics over it.

packages/lib/server/repository/team.ts
```diff
@@ -0,0 +1,149 @@
...
+        };
+      }),
+    });
+  }
+
+  const team = teamsWithParsedMetadata[0];
+  if (!team) return null;
+  // HACK: I am not sure how to make Prisma in peace with TypeScript with this repository pattern
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return team as any;
```
hariombalhara, line 122: There is a return type on function already that should ensure that the function callers would get typesafey as long as this function is well tested.

packages/lib/server/repository/team.ts
```diff
@@ -0,0 +1,148 @@
...
+  const teamsWithParsedMetadata = teams
+    .map((team) => ({
+      ...team,
+      // Using Type assertion here because we know that the metadata is present and Prisma and TypeScript aren't playing well together
+      metadata: teamMetadataSchema.parse((team as { metadata: z.infer<typeof teamMetadataSchema> }).metadata),
+    }))
+    // In cases where there are many teams with the same slug, we need to find out the one and only one that matches our criteria
+    .filter((team) => {
+      // We need an org if isOrgView otherwise we need a team
+      return isOrg ? team.metadata?.isOrganization : !team.metadata?.isOrganization;
```
hariombalhara, line 100: This is what fixes the bug. So, even if we get wrong result, we validate it here and then reject it.

packages/lib/server/queries/teams/index.ts
```diff
@@ -61,115 +64,100 @@ export async function getTeamWithMembers(args: {
...
-        ? obj.user.teams.filter((obj) => obj.team.slug !== orgSlug).map((obj) => obj.team.slug)
+        ? m.user.teams
+            .filter((membership) => membership.team.slug !== orgSlug)
+            .map((membership) => membership.team.slug)
         : null,
-      avatar: `${WEBAPP_URL}/${obj.user.username}/avatar.png`,
-      orgOrigin: getOrgFullOrigin(obj.user.organization?.slug || ""),
+      avatar: `${WEBAPP_URL}/${m.user.username}/avatar.png`,
+      orgOrigin: getOrgFullOrigin(m.user.organization?.slug || ""),
       connectedApps: !isTeamView
```
Udit-takkar, line 161: What is the use of isTeamView ? Do we need both variable isOrgView and isTeamView

Udit-takkar (commented):
Edit: can be fixed in another PR

https://github.com/calcom/cal.com/assets/53316345/64a012d4-6b34-42a8-8101-40374ffa5a33

I tried creating a team with same slug as my org and it was successfull but on the next page (onboard-members) I get an error. teams.create route returned already existing org instead of creating a team

packages/lib/server/repository/team.ts
sean-brydon: Do we actually need a generic here? Or can we we just use Prisma.TeamSelect as the type of team select

I made an example that seems to work - see your DMs
https://threads.com/messages/ef1928b9-ab1a-4903-bad2-670138a519e6?messageID=d40dff2c-de71-47ae-ab38-7f3770452777

packages/lib/server/queries/teams/index.ts
hariombalhara: `isTeamView` is actually meant to decide if credentials are supposed to be available in the function return value or not.

It is true on team booking pages. The name is quite confusing though, I fully agree

hariombalhara:
> Edit: can be fixed in another PR
> 
>  Screen.Recording.2023-11-07.at.9.06.46.PM.mov 
> I tried creating a team with same slug as my org and it was successfull but on the next page (onboard-members) I get an error. teams.create route returned already existing org instead of creating a team

Thanks I will take it up later.

packages/lib/server/repository/team.ts
hariombalhara: Discussed with Sean. It won't work

Udit-takkar (dismissed):
LGTM. 
Great work again in using the repository pattern here!

Udit-takkar (approved):

PR #12223, opened 2023-11-03
fix(stripePaymentCallback): better errors

ThyMinimalDev:
## What does this PR do?

This Pr attempts to improve errors message on stripepayment/paymentCallback endpoint to help us debug this endpoint. Also try to avoid ERR_HTTP_HEADERS_SENT.

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/app-store/stripepayment/api/paymentCallback.ts
```diff
@@ -43,10 +45,10 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
     });
   }
 
+  if (!user) throw new HttpError({ statusCode: 404, message: "User not found" });
+
   if (checkoutSession.payment_status === "paid" && stripeCustomer.metadata.username) {
     try {
-      if (!user) return { message: "User not found" };
-
```
ThyMinimalDev, line 49: did this make sense ? why did we check for user only if payment_status was paid and stripeCustomer.metadata.username was set ?

ThyMinimalDev: does it make sense to redirect a user that is not in the database ?

apps/web/pages/api/integrations/[...args].ts
```diff
@@ -73,7 +73,8 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
       redirectUrl = handler.redirect?.url || getInstalledAppPath(handler);
       res.json({ url: redirectUrl, newTab: handler.redirect?.newTab });
     }
-    return res.status(200);
+    if (!res.writableEnded) return res.status(200);
```
Comment (ThyMinimalDev) [inline apps/web/pages/api/integrations/[...args].ts:76]: try to prevent ERR_HTTP_HEADERS_SENT

packages/app-store/stripepayment/api/paymentCallback.ts
joeauyeung: This could have just been to silence typescript. When commenting it out we get typescript errors. The chances of no user being found are very very slim. Before sending the user to the checkout page we get the userId from Stripe. We then get the query again by the Stripe customer. I highly doubt that that information would change during the time the user is checking out.

packages/app-store/stripepayment/api/paymentCallback.ts
```diff
@@ -61,16 +64,15 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
...
     }
   }
   callbackUrl.searchParams.set("paymentStatus", checkoutSession.payment_status);
   return res.redirect(callbackUrl.toString()).end();
 }
 
-export default defaultHandler({
-  GET: Promise.resolve({ default: defaultResponder(getHandler) }),
-});
+export default handler;
```
ThyMinimalDev, line 78: removed defaultHandler and defaultResponder, everything is already handled in `apps/web/pages/api/integrations/[...args].ts`

hariombalhara: I think we shouldn't remove it. It takes care of some other things that has to be manually handled now and could potentially handle more things.
e.g. it was automatically erroring out when some other method than GET was used.

packages/app-store/stripepayment/api/paymentCallback.ts
hariombalhara: Looks good to me.

hariombalhara (dismissed):
Requesting to add back defaultHandler and defaultResponder otherwise it's good.

packages/app-store/stripepayment/api/paymentCallback.ts
ThyMinimalDev: I'll run some test trying to keep defaultHandler, but right now we are handling the response in too many places so probably don't need default responder

packages/lib/server/getServerErrorFromUnknown.ts
```diff
@@ -52,9 +52,11 @@ export function getServerErrorFromUnknown(cause: unknown): HttpError {
     return getHttpError({ statusCode: 400, cause });
   }
   if (cause instanceof HttpError) {
-    const redactedCause = redactError(cause);
+    const _ = redactError(cause);
     return {
-      ...redactedCause,
```
ThyMinimalDev, line 57: spreading an Error doesn't work, it doesn't have any enumerable property

packages/lib/server/getServerErrorFromUnknown.ts
```diff
@@ -52,9 +52,11 @@ export function getServerErrorFromUnknown(cause: unknown): HttpError {
...
+      name: cause.name,
+      message: cause.message ?? "",
+      stack: cause.stack,
```
ThyMinimalDev, line 59: extract interesting fields from error

packages/lib/server/defaultResponder.ts
```diff
@@ -14,13 +14,14 @@ export function defaultResponder<T>(f: Handle<T>) {
...
+        return res.json(result);
       }
     } catch (err) {
       console.error(err);
       const error = getServerErrorFromUnknown(err);
-      res.statusCode = error.statusCode;
-      res.json({ message: error.message });
+      return res
+        .status(error.statusCode)
+        .json({ message: error.message, url: error.url, method: error.method });
```
ThyMinimalDev, line 24: add some information to the error, we now properly get error message when using http error, spreading redactedClause didn't work

packages/lib/server/getServerErrorFromUnknown.ts
keithwillcode, line 59: The `redactError` method was intentionally introduced to avoid a situation where the Prisma connection information was being passed to the client. Are we sure these changes won't reintroduce that?

keithwillcode (changes requested):

packages/lib/server/getServerErrorFromUnknown.ts
ThyMinimalDev: I don't think so, here the function is just there to do some logging and is still being used. it's returning an error which we  used to spread with the spread operator. but js Errors don't have any enumerable properties so spread has no effect

ThyMinimalDev: added redactedCuse back and removed stack just in case

keithwillcode: ok cool. @hariombalhara can you please take a look here as you had originally introduced this?

packages/lib/server/getServerErrorFromUnknown.ts
ThyMinimalDev: this only works when redactedCause is an Object

keithwillcode (dismissed):

keithwillcode (approved):

PR #12249, opened 2023-11-06
fix: avatar remove button

### LINKED ISSUE #12164: [CAL-2665] Remove button not displayed now in avatar upload ( Regression)
https://github.com/calcom/cal.com/pull/12143#pullrequestreview-1703557065

<sub>[CAL-2665](https://linear.app/calcom/issue/CAL-2665/remove-button-not-displayed-now-in-avatar-upload-regression)</sub>

MehulZR:
## What does this PR do?

Fixes #12164  (issue)

Sometimes there is a weird caching behavior(requires tab reload) which I have not been able to pin-down the root cause.
If someone is familiar with this and can point me to the right direction I would like to look into it more.
Apart from that it is working as expected.
 
https://github.com/calcom/cal.com/assets/86384652/d2d999d7-7250-4e35-a58b-449136ffaac1

## Type of change

## Mandatory Tasks

### DIFF
--- a/apps/web/pages/settings/my-account/profile.tsx
+++ b/apps/web/pages/settings/my-account/profile.tsx
@@ -9,8 +9,8 @@ import { ErrorCode } from "@calcom/features/auth/lib/ErrorCode";
 import OrganizationMemberAvatar from "@calcom/features/ee/organizations/components/OrganizationMemberAvatar";
 import SectionBottomActions from "@calcom/features/settings/SectionBottomActions";
 import { getLayout } from "@calcom/features/settings/layouts/SettingsLayout";
+import checkIfItFallbackImage from "@calcom/lib/checkIfItFallbackImage";
 import { APP_NAME, FULL_NAME_LENGTH_MAX_LIMIT } from "@calcom/lib/constants";
-import { AVATAR_FALLBACK } from "@calcom/lib/constants";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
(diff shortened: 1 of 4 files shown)


Udit-takkar (changes requested):
~There are some issues in organization settings profile~

apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -72,32 +72,24 @@ interface DeleteAccountValues {
 
 type FormValues = {
   username: string;
-  avatar: string | null;
```
Udit-takkar, line 75: Why did you remove null ? Avatar would be null if we want to remove it

Udit-takkar: Okay you added this in updateProfile 
"input.avatar ? await getAvatarToSet(input.avatar) : null"

Udit-takkar (approved):
LGTM.

Tested on normal account and org account.

apps/web/pages/settings/my-account/profile.tsx
MehulZR: I removed it because in case it is null
image
The *previewSrc* prop for OrganizationMemberAvatar component will be null, and it uses Avatar component under the hood.
image

Which if provided null value for previewSrc, will just fetch the current avatar of the user. 

But on the profile settings page we would like to show the default avatar svg. 
image

So using an empty string for that purpose work as a trick. I could do it to support null but dont want regression at some other place.

PR #12251, opened 2023-11-06
fix: saving credential id for payment apps

joeauyeung:
## What does this PR do?

This PR adds the `credentialId` to payment app data. This fixes a bug where team installed payment apps were not working with team events.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Install Stripe for the individual user
- Enable it for the individual's event type
- The `credentialId` should be saved to the metadata
- Install Stripe to the user's team
- Enable it in the team's event type
- The `credentialId` should be saved to the metadata

## Mandatory Tasks

## Checklist


packages/app-store/alby/components/EventTypeAppCardInterface.tsx
```diff
@@ -33,6 +33,15 @@ const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({
   const { t } = useLocale();
   const recurringEventDefined = eventType.recurringEvent?.count !== undefined;
 
+  let credentialId;
+
+  if (eventType.teamId) {
+    const teamCredential = app.teams.find((team) => team.teamId === eventType.teamId);
+    credentialId = teamCredential?.credentialId ?? app.userCredentialIds[0];
```
joeauyeung, line 40: For event type apps, I found that `app.userCredentialIds` always has one entry.

ThyMinimalDev:
can't we do this by default for all apps in `packages/app-store/_components/EventTypeAppCardInterface.tsx` ?

ThyMinimalDev:
> can't we do this by default for all apps in `packages/app-store/_components/EventTypeAppCardInterface.tsx` ?
> 
> I mean do this by default ? api endpoints can choose to use it or not in the end `setAppData("credentialId", credentialId);`

@hariombalhara what do you think of this ?

Udit-takkar:
@joeauyeung there are some type errors and one E2E test is failing

apps/web/playwright/integrations-stripe.e2e.ts
```diff
@@ -34,6 +37,91 @@ test.describe("Stripe integration", () => {
...
+      page.click('[data-testid="anything else"]'),
+    ]);
+
+    await Promise.all([
+      page.waitForURL("/apps/installed/payment?hl=stripe"),
+      /** We skip filling Stripe forms (testing mode only) */
+      page.click('[id="skip-account-app"]'),
+    ]);
+
+    await owner.setupEventWithPrice(teamEvent);
```
Udit-takkar, line 104: Screenshot 2023-11-07 at 2 53 13 PM

This Line is causing issue when stripe is installed for a team
await page.locator("div > .ml-auto").first().click();

you can add test-id to the switch

Screenshot 2023-11-07 at 2 54 49 PM

Udit-takkar (dismissed):
LGTM. I have tested with stripe. 

How do I set up alby?


https://github.com/calcom/cal.com/assets/53316345/03082d1a-7d58-419e-8611-2ce36aeb0492

hariombalhara:
> > can't we do this by default for all apps in `packages/app-store/_components/EventTypeAppCardInterface.tsx` ?
> > I mean do this by default ? api endpoints can choose to use it or not in the end `setAppData("credentialId", credentialId);`
> 
> @hariombalhara what do you think of this ?

@ThyMinimalDev That certainly makes more sense. Maybe we should modify setData to automatically set `credentialId` so that whenever something is set `credentialId` is always set.
 
Right now, there is a different kind of abstraction in place which requires us to call `updateEnabled` which then takes care of setting `credentialId`. But, payment apps, don't call `updateEnabled` which seems to be the reason for this issue in the first place.

But if we can just add credentialId in the `setAppData` flow we could avoid the issue from happening ever again.

packages/app-store/stripepayment/zod.ts
```diff
@@ -18,6 +18,7 @@ export const appDataSchema = eventTypeAppCardZod.merge(
     currency: z.string(),
     paymentOption: paymentOptionEnum.optional(),
     enabled: z.boolean().optional(),
+    credentialId: z.number().optional(),
```
hariombalhara, line 21: https://github.com/calcom/cal.com/blob/1610d083428ddda277ffbfdbf4f11fc4537eba26/packages/app-store/eventTypeAppCardZod.ts#L6

eventTypeAppCardZod already has `credentialId`, so we shouldn't be needing it.

packages/app-store/alby/components/EventTypeAppCardInterface.tsx
zomars: Should this data come from the server directly?

packages/app-store/paypal/components/EventTypeAppCardInterface.tsx
```diff
@@ -29,6 +29,15 @@
     isAcceptedCurrencyCode(currency) ? currencySymbols[currency] : ""
   );
 
+  let credentialId: number;
```
zomars, line 32: Same here

packages/app-store/stripepayment/components/EventTypeAppCardInterface.tsx
```diff
@@ -33,6 +33,15 @@ const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({
   const paymentOptionSelectValue = paymentOptions.find((option) => paymentOption === option.value);
   const [requirePayment, setRequirePayment] = useState(getAppData("enabled"));
 
+  let credentialId: number;
+
+  if (eventType.teamId) {
```
zomars, line 38: I'm noticing a pattern here

zomars: How can we be more DRY in here?

zomars (dismissed):

joeauyeung:
> > > can't we do this by default for all apps in `packages/app-store/_components/EventTypeAppCardInterface.tsx` ?
> > > I mean do this by default ? api endpoints can choose to use it or not in the end `setAppData("credentialId", credentialId);`
> > 
> > 
> > @hariombalhara what do you think of this ?
> 
> @ThyMinimalDev That certainly makes more sense. Maybe we should modify setData to automatically set `credentialId` so that whenever something is set `credentialId` is always set.
> 
> Right now, there is a different kind of abstraction in place which requires us to call `updateEnabled` which then takes care of setting `credentialId`. But, payment apps, don't call `updateEnabled` which seems to be the reason for this issue in the first place.
> 
> But if we can just add credentialId in the `setAppData` flow we could avoid the issue from happening ever again.
> 
> @joeauyeung We can fix it in a followup or in this PR. I am happy either way.

Implemented suggestion in this commit

packages/app-store/paypal/components/EventTypeAppCardInterface.tsx
joeauyeung: Fixed in this commit https://github.com/calcom/cal.com/commit/b88334b0cd6c8b92ba7d2870c61e8797eb7ecbd3

packages/app-store/stripepayment/components/EventTypeAppCardInterface.tsx
joeauyeung: Fixed in this commit https://github.com/calcom/cal.com/commit/b88334b0cd6c8b92ba7d2870c61e8797eb7ecbd3

packages/app-store/alby/zod.ts
```diff
@@ -29,6 +29,7 @@ export const appDataSchema = eventTypeAppCardZod.merge(
     currency: z.string(),
     paymentOption: z.string().optional(),
     enabled: z.boolean().optional(),
+    credentialId: z.number().optional(),
```
hariombalhara, line 32: I think this too isn't needed.

hariombalhara (approved):

zomars:
### Merge activity

* **Nov 15, 9:59 PM**: @zomars merged this pull request with [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12251).

PR #12270, opened 2023-11-07
chore: Reduce slots API serverless memory configuration

keithwillcode:
## What does this PR do?

Vercel serverless function pricing is based on the memory allocated of serverless functions, not usage of memory. After analyzing the memory usage of the slots endpoint, it looks like we can confidently lower its memory allocation to reduce cost by half.

## Type of change

## How should this be tested?

- [ ] Ensure the memory allocation configuration is working properly for the slots route in the API

## Mandatory Tasks


keithwillcode:
Have been testing this and for whatever reason, the memory value won't take when deployed on Vercel. Still investigating.

apps/api/vercel.json
```diff
@@ -0,0 +1,7 @@
+{
+  "functions": {
+    "pages/api/slots/*.ts": {
+      "memory": 512
```
keithwillcode, line 4: This route on the API takes less memory than in the web app.

apps/web/vercel.json
```diff
@@ -4,5 +4,13 @@
...
       "schedule": "0 5 * * *"
     }
-  ]
+  ],
+  "functions": {
+    "pages/api/trpc/public/[trpc].ts": {
+      "memory": 768
+    },
+    "pages/api/trpc/slots/[trpc].ts": {
+      "memory": 768
```
keithwillcode, line 13: We currently don't use this route from the UI but preparing this because we have a TODO to switch it from `/api/trpc/public/slots.getSchedule` to `/api/trpc/slots/getSchedule`.

emrysal (approved):

PR #12272, opened 2023-11-07
fix: Update Event Type Pricing For Multiple Installed Payment Apps

joeauyeung:
## What does this PR do?

This PR fixes a bug when a user had multiple payment apps installed, the pricing was not being updated on the event type when the payment apps settings were changed.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Install Stripe and Paypal
- Enable one payment app on the event type and save
- Now enable the second payment app with a new price and save
    - The price should be updated
- Try enabling both payment apps
    - An error should be displayed 

## Mandatory Tasks

## Checklist


apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -473,6 +473,19 @@ const EventTypePage = (props: EventTypeSetupProps) => {
...
+    for (const appKey in metadata?.apps) {
+      const app = metadata?.apps[appKey];
+      if (app.price && app.enabled) {
+        enabledPaymentApps++;
+      }
+
+      if (enabledPaymentApps > 1) {
+        throw new Error(t("event_setup_multiple_payment_apps_error"));
+      }
+    }
```
Comment (exception) [inline apps/web/pages/event-types/[type]/index.tsx:487]: `suggestion`: Something like this could be cleaner.

```suggestion
    const enabledPaymentApps = Object.values(metadata?.apps || {}).filter(app => app.price && app.enabled).length;
if (enabledPaymentApps > 1) {
  throw new Error(t("event_setup_multiple_payment_apps_error"));
}
```

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
```diff
@@ -245,26 +245,21 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
...
-    input?.metadata?.apps?.stripe?.currency
-  ) {
-    data.currency =
-      input.metadata?.apps?.alby?.currency ||
-      input.metadata.apps.paypal?.currency ||
-      input.metadata.apps.stripe?.currency;
+  if (enabledPaymentAppKey) {
+    const paymentApp = input.metadata?.apps[enabledPaymentAppKey];
+    data.price = paymentApp.price;
+    data.currency = paymentApp.currency;
```
ThyMinimalDev, line 262: this will create typescripts errors

apps/web/pages/event-types/[type]/index.tsx
Comment (ThyMinimalDev) [reply apps/web/pages/event-types/[type]/index.tsx:487]: typescript issue 
Screenshot 2023-11-08 at 10 46 48

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -473,6 +473,19 @@ const EventTypePage = (props: EventTypeSetupProps) => {
...
 
+    // Prevent two payment apps to be enabled
+    let enabledPaymentApps = 0;
+    for (const appKey in metadata?.apps) {
+      const app = metadata?.apps[appKey];
+      if (app.price && app.enabled) {
+        enabledPaymentApps++;
+      }
+
+      if (enabledPaymentApps > 1) {
```
Comment (ThyMinimalDev) [inline apps/web/pages/event-types/[type]/index.tsx:484]: we should extract this logic in the appStore I think, create properly typed utils to get enabledApps and enabledPaymentApps

ThyMinimalDev (dismissed):

packages/trpc/server/routers/viewer/eventTypes/update.handler.ts
ThyMinimalDev: we should have utils in appstore to make this easier, something like this that we could generate 

```ts
 const PAYMENT_APP_KEYS: (keyof Pick<typeof appDataSchemas, "stripe" | "paypal" | "alby">)[] = [
    "paypal",
    "alby",
    "stripe",
  ];
  const eventTypePaymentApps = PAYMENT_APP_KEYS.map((appKey) => input.metadata?.apps?.[appKey]);
  const paymentApp = eventTypePaymentApps.find((app) => app?.enabled && app?.price);

  if (paymentApp) {
    data.price = paymentApp.price;
    data.currency = paymentApp.currency;
  }
```

ThyMinimalDev: maybe @hariombalhara you have an idea to handle this ?

joeauyeung: There's a function called `getPaymentAppData` that exists. The thing is, it expects the entire return of `getEventTypeById` but it only looks at the event type metadata. Right now in the update handler we're not querying for the whole event type, just what we need.

apps/web/pages/event-types/[type]/index.tsx
Comment (ThyMinimalDev) [reply apps/web/pages/event-types/[type]/index.tsx:487]: yes it would indeed iterate over all the apps inside the metadata, both approach are fine on such small dataset, both work just a matter of preference if you like the shorter version or not up to you.

I would extract this logic in a small function in the same file with clear naming and just the required parameters 

preventMultiplePaymentApps(...)

also there is a type error you can fix the same way you did in the other place ` const app = metadata?.apps[appKey as keyof AppMetadata];`

ThyMinimalDev (changes requested):

ThyMinimalDev (dismissed):

ThyMinimalDev:
looks good to me, can we check with product if this new behaviour is ok?

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -570,6 +577,20 @@ const EventTypePage = (props: EventTypeSetupProps) => {
...
+            for (const appKey in metadata?.apps) {
+              const app = metadata?.apps[appKey as keyof typeof appDataSchemas];
+              if (app.price && app.enabled) {
+                enabledPaymentApps++;
+              }
+
+              if (enabledPaymentApps > 1) {
+                throw new Error(t("event_setup_multiple_payment_apps_error"));
+              }
+            }
```
Comment (zomars) [inline apps/web/pages/event-types/[type]/index.tsx:592]: Why not use checkForMultiplePaymentApps in here as well? We already extracted the logic.

zomars (commented):
Code looks mostly fine. Left a comment in there

ThyMinimalDev:
Would you mind creating a Loom video to showcase the expected behavior after this pull request? This will be helpful for maintaining context when we later explore handling multiple payment apps. Thank you!

joeauyeung:
@ThyMinimalDev here's a Loom with what the user will from #12414
https://www.loom.com/share/dbec216876c24dac998cc12db7595814

I want to get this PR merged since this will fix some payment app issues that users are facing right now.

ThyMinimalDev (dismissed):
LGTM, let's check in db if there are any event-types with 2 payments app enabled already, in order to know if we need to run a migration script

exception (dismissed):

ThyMinimalDev (dismissed):

exception (dismissed):

ThyMinimalDev:
updating branch to see if E2E tests are passing, I think the e2e failures were not related to the code in this pr

zomars (approved):
LGTM

zomars (changes requested):
My bad. Seems like the e2e failures are Stripe related. Which this PR touches code from.

joeauyeung:
> My bad. Seems like the e2e failures are Stripe related. Which this PR touches code from.

I noticed this happening in other PRs as well. The tests are passing locally.

zomars (dismissed):
Still seeing some legit errors that aren't main related.

joeauyeung:
Tests related to this PR are now passing

keithwillcode (approved):

PR #12275, opened 2023-11-07
fix: handle reschedule request for dynamic meetings

### LINKED ISSUE #12263: Bug: Can't do 'Request Reschedule' for a dynamic group booking
Screenshot 2023-11-07 at 6 00 36 PM

manpoffc:
## What does this PR do?
This PR fixes the reschedule request handler for the dynamic meetings.

Fixes #12263   

## Type of change

## Mandatory Tasks


packages/trpc/server/routers/viewer/bookings/requestReschedule.handler.ts
```diff
@@ -78,7 +78,7 @@ export const requestRescheduleHandler = async ({ ctx, input }: RequestReschedule
     throw new TRPCError({ code: "FORBIDDEN", message: "Booking to reschedule doesn't have an owner" });
   }
 
-  if (!bookingToReschedule.eventType) {
+  if (!bookingToReschedule.eventType && !bookingToReschedule.dynamicEventSlugRef) {
```
supalarry, line 81: Just brainstorming - maybe the way to fix the issue is understanding why doesn't `eventType` exist on the `bookingToReschedule` object?

I am thinking that resolving that would lead to `requestReschedule.handler.ts` being untouched. The TRPCError has message `"EventType not found for current booking."` but this `&&` conditional also checks for slug, so the error message does not reflect the check in code.

I am not familiar with this part of codebase, but to me it feels the problem is somewhere else.

Nonetheless, thank you for your contributions! ❤️

manpoffc: Yes certainly you're correct, but whenever a dynamic meeting is created there is no eventType created for that meeting. I will further investigate more into this.

CarinaWolli (approved):
Looks good, it makes sense that a dynamic booking doesn't have an event type 👍
Thank you 🙏

PR #12011, opened 2023-10-20
test: Integration Test GCal Primary Calendar

joeauyeung:
## What does this PR do?

This PR aims to add E2E testing to our GCal integration. The tests use a predefined QA user that is already connected to Google. When going through this PR is to think about what will happen when 10 of these tests are running at the same time.

Next step is to add tests for:
- Ensure that events on external calendars are blocked on Cal.com
- Tests for non primary calendars
- Expand this to Outlook

Fixes # (issue)

## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is ab UI/UX design document, please, share it here.

## Type of change

- [x] Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- Set the `.env` variables `E2E_TEST_CALCOM_QA_EMAIL` and `E2E_TEST_CALCOM_QA_PASSWORD`
- Create a QA user account with the above .env variables
- Connect a Google account to this QA user
- Run the tests found in `packages/app-store/googlecalendar/tests/google-calendar.e2e.ts`
- If you can, add breakpoints to certain steps of the tests to ensure that the events are being created on GCal

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #12291, opened 2023-11-09
fix: When Installing Calendar App, Set Primary Calendar As Selected Calendar

joeauyeung:
## What does this PR do?

This PR sets the primary/default calendar of a calendar integration as the first selected calendar on install. 

The calendars that have this ability are:
- Google Calendar
- Outlook
- Zoho
- Lark

Once #12011 is merged we can add a test for this

CleanShot 2023-11-08 at 22 17 13

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Install the calendars that are included in this PR
- On the installed page, the primary calendar should already be selected

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/app-store/larkcalendar/api/callback.ts
+++ b/packages/app-store/larkcalendar/api/callback.ts
@@ -89,6 +89,32 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
       });
     }
 
+    const primaryCalendarResponse = await fetch(
+      `https://${LARK_HOST}/open-apis/calendar/v4/calendars/primary`,
+      {
+        method: "GET",
+        headers: {
+          Authorization: `Bearer ${key.access_token}`,
+          "Content-Type": "application/json",
+        },
+      }
+    );
+
+    if (primaryCalendarResponse.status === 200) {
+      const primaryCalendar = await primaryCalendarResponse.json();
+
+      if (primaryCalendar.data.calendars.calendar.calendar_id && req.session?.user?.id) {
+        await prisma.selectedCalendar.create({
+          data: {
+            userId: req.session?.user.id,
+            integration: "lark_calendar",
+            externalId: primaryCalendar.data.calendars.calendar.calendar_id as string,
+            credentialId: currentCredential?.id,
+          },
+        });
+      }
+    }
+
     res.redirect(
       getSafeRedirectUrl(state?.returnTo) ??
         getInstalledAppPath({ variant: "calendar", slug: "lark-calendar" })
--- a/packages/app-store/office365calendar/api/callback.ts
+++ b/packages/app-store/office365calendar/api/callback.ts
@@ -1,6 +1,8 @@
+import type { Calendar as OfficeCalendar } from "@microsoft/microsoft-graph-types-beta";
 import type { NextApiRequest, NextApiResponse } from "next";
 
 import { WEBAPP_URL } from "@calcom/lib/constants";
+import { handleErrorsJson } from "@calcom/lib/errors";
 import { getSafeRedirectUrl } from "@calcom/lib/getSafeRedirectUrl";
 import prisma from "@calcom/prisma";
 
@@ -65,7 +67,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   responseBody.expiry_date = Math.round(+new Date() / 1000 + responseBody.expires_in); // set expiry date in seconds
   delete responseBody.expires_in;
 
-  await prisma.credential.create({
+  const credential = await prisma.credential.create({
     data: {
       type: "office365_calendar",
       key: responseBody,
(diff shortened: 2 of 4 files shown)


packages/app-store/office365calendar/api/callback.ts
```diff
@@ -74,6 +76,50 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     },
   });
 
+  // Set the isDefaultCalendar as selectedCalendar
+  // If a user has multiple calendars, keep on making calls until we find the default calendar
+  let defaultCalendar: OfficeCalendar | undefined = undefined;
+  let requestUrl = "https://graph.microsoft.com/v1.0/me/calendars?$select=id,isDefaultCalendar";
+  let finishedParsingCalendars = false;
+
+  while (!finishedParsingCalendars) {
```
joeauyeung, line 85: MS Graph only returns 10 calendars at a time. If for any reason the primary calendar is not a part of the first call then continue making calendar calls.

supalarry:
Tested locally - can confirm that after installing Google Calendar and going to "Installed apps" the toggle for my e-mail for Google Calendar was on.

packages/app-store/office365calendar/api/callback.ts
```diff
@@ -74,6 +76,50 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+  if (defaultCalendar?.id && req.session?.user?.id) {
+    await prisma.selectedCalendar.create({
+      data: {
+        userId: req.session?.user.id,
+        integration: "office365_calendar",
+        externalId: defaultCalendar.id,
+        credentialId: credential.id,
+      },
+    });
+  }
```
exception, line 121: `suggestion-non-blocking`: I'd prefer encapsulating this logic into its own function, and only running it if the actual `req.session.user.id` exists, if not running it is useless no?

```ts
const findAndStoreDefaultCalendar = async (accessToken: string, credentialId: string, userId: string) => {
  let requestUrl = "https://graph.microsoft.com/v1.0/me/calendars?$select=id,isDefaultCalendar";
  
  try {
    while (true) {
      const response = await fetch(requestUrl, {
        method: "GET",
        headers: {
          Authorization: `Bearer ${accessToken}`,
          "Content-Type": "application/json",
        },
      });

      const { value: calendars, "@odata.nextLink": nextLink } = await handleErrorsJson<{ value: OfficeCalendar[]; "@odata.nextLink"?: string }>(response);
      const defaultCalendar = calendars.find(calendar => calendar.isDefaultCalendar);

      if (defaultCalendar) {
        await prisma.selectedCalendar.create({
          data: {
            userId,
            integration: "office365_calendar",
            externalId: defaultCalendar.id,
            credentialId,
          },
        });

        break;
      }

      if (!nextLink) {
        break;
      }

      requestUrl = nextLink;
    }
  } catch (error) {
    throw error;
  }
}

if (req.session?.user?.id) {
  await findAndStoreDefaultCalendar(responseBody.access_token, credential.id, req.session.user.id);
}
```

packages/app-store/zohocalendar/api/callback.ts
```diff
@@ -64,7 +64,36 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
...
+  const calendarResponse = await fetch("https://calendar.zoho.com/api/v1/calendars", {
+    method: "GET",
+    headers: {
+      Authorization: `Bearer ${key.access_token}`,
+      "Content-Type": "application/json",
+    },
+  });
+  const data = await calendarResponse.json();
+
+  const primaryCalendar = data.calendars.find((calendar: any) => calendar.isdefault);
```
exception, line 85: `nitpick`: Please ensure a type, even if `{ isdefault: boolean }`

emrysal, line 85: Valid eslint warning, non-blocking tho.

emrysal (approved):

packages/app-store/larkcalendar/api/callback.ts
```diff
@@ -89,6 +89,32 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
...
+        await prisma.selectedCalendar.create({
+          data: {
+            userId: req.session?.user.id,
+            integration: "lark_calendar",
+            externalId: primaryCalendar.data.calendars.calendar.calendar_id as string,
+            credentialId: currentCredential?.id,
+          },
+        });
+      }
+    }
```
exception, line 116: `suggestion`: Please check for `req.session.user.id` first, no need to run the logic if we're not gonna store it.

exception (changes requested):

packages/app-store/zohocalendar/api/callback.ts
emrysal: Uncertain types aren't a good idea, `any` is actually better as it conveys something is really "unknown" - I did mark this as a nitpick as this is a potential issue.

The right fix for an issue like this is to parse the JSON response and validate the isdefault key is there using zod, then you don't need to type any - then throw a descriptive error when this fails the parsing.

packages/app-store/larkcalendar/api/callback.ts
emrysal: See L106 - the extra ? isn't necessary here though @joeauyeung, req.session.user.id is sure to be defined here.

packages/app-store/office365calendar/api/callback.ts
emrysal: Even better, looking at the entire code it looks like the credential create will also fail if there is no session user, so the ideal fix is to return early out of the entire callback when the session user id isn't truthy.

PR #12300, opened 2023-11-09
fix: Sentry warnings for startTransaction

keithwillcode:
## What does this PR do?

Removes the tracing variable completely for Sentry transactions.

## Type of change

## How should this be tested?

- Ensure errors are still sent to Sentry

## Mandatory Tasks

### DIFF
--- a/apps/web/sentry.server.config.ts
+++ b/apps/web/sentry.server.config.ts
@@ -2,5 +2,4 @@ import * as Sentry from "@sentry/nextjs";
 
 Sentry.init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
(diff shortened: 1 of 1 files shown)


emrysal (approved):
Strange. Oh well if it works (you'd think 0 === <none>)

apps/web/sentry.server.config.ts
```diff
@@ -2,5 +2,4 @@ import * as Sentry from "@sentry/nextjs";
 
 Sentry.init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
-  tracesSampleRate: 0,
```
keithwillcode, line 5: Removing this option completely for now since we don't use transactions

PR #12308, opened 2023-11-09
fix: meeting url variable in workflow notifitcations

### LINKED ISSUE #10758: [CAL-2518] {MEETING_URL} intermittently missing from some custom emails
### Issue Summary

I've noticed on some custom email workflows, the {MEETING_URL} doesn't seem to be populating

When enabling multiple custom workflows triggered by a new booking creation, I received 5 copies of each workflow email. This was an unrelated bug that I've been unable to reproduce but it did help me confirm that 4 copies of the emails came in without the meeting_url but the 5th one **_was correctly populated_**. This makes me think there's a race condition of the emails getting sent out before the meeting url is generated

### Steps to Reproduce

1. Create a new workflow for when a booking is created
2. Copy the html from [this gist](https://gist.github.com/kcrimi/0106b86992e78ea86ab6928e5bdcf69e) as the custom body of the email
3. Save the workflow to function on an event type
4. Create a booking for that event type

### Actual Results

- The email is delivered as expected but intermittently does not include the meeting url while all other variables are included
Screenshot 2023-08-14 at 2 40 58 PM

### Expected Results

- All variables are filled in, including the meeting url (below picture is from the same workflow)
Screenshot 2023-08-14 at 2 41 07 PM

### Technical details

- Using the cloud hosted application

### Evidence

- See above

<sub>[CAL-2518](https://linear.app/calcom/issue/CAL-2518/meeting-url-intermittently-missing-from-some-custom-emails)</sub>

CarinaWolli:
## What does this PR do?

Fixes that the {MEETING_URL} from custom workflow templates wasn't populated for event types that require confirmation or payment.

Also fixes that confirmation emails were still sent after they were confirmed even when the default confirmation emails were disabled. 

Fixes #10758

## Requirement/Documentation

Have Sendgrid set up to send workflow emails.

## Type of change

## How should this be tested?

- Create a workflow that triggers when new event is booked & sends email to attendees
- Add a custom emails template with {MEETING_URL} variable
- Set active on an event type that requires confrimation 
- Book event type 
- Confirm booking 
- Check if email has meeting url

## Mandatory Tasks

### DIFF
--- a/packages/features/bookings/lib/handleConfirmation.ts
+++ b/packages/features/bookings/lib/handleConfirmation.ts
@@ -39,7 +39,7 @@ export async function handleConfirmation(args: {
       title: string;
       teamId?: number | null;
       parentId?: number | null;
-      metadata: Prisma.JsonValue;
+      metadata?: Prisma.JsonValue;
       workflows?: {
         workflow: Workflow & {
           steps: WorkflowStep[];
@@ -134,6 +134,8 @@ export async function handleConfirmation(args: {
     } | null;
   }[] = [];
 
+  const videoCallUrl = metadata.hangoutLink ? metadata.hangoutLink : evt.videoCallData?.url || "";
+
   if (recurringEventId) {
     // The booking to confirm is a recurring event and comes from /booking/recurring, proceeding to mark all related
     // bookings as confirmed. Prisma updateMany does not support relations, so doing this in two steps for now.
@@ -155,6 +157,7 @@ export async function handleConfirmation(args: {
             create: scheduleResult.referencesToCreate,
           },
           paid,
+          metadata: { ...recurringBooking.metadata, videoCallUrl },
         },
         select: {
           eventType: {
@@ -206,6 +209,7 @@ export async function handleConfirmation(args: {
         references: {
           create: scheduleResult.referencesToCreate,
         },
+        metadata: { ...booking.metadata, videoCallUrl },
       },
       select: {
         eventType: {
@@ -246,8 +250,6 @@ export async function handleConfirmation(args: {
 
   //Workflows - set reminders for confirmed events
   try {
-    const videoCallUrl = metadata.hangoutLink ? metadata.hangoutLink : evt.videoCallData?.url || "";
-
     for (let index = 0; index < updatedBookings.length; index++) {
       if (updatedBookings[index].eventType?.workflows) {
         const evtOfBooking = evt;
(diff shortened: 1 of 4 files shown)


packages/features/bookings/lib/handleConfirmation.ts
```diff
@@ -54,16 +67,41 @@ export async function handleConfirmation(args: {
...
+        if (isHostConfirmationEmailsDisabled) {
+          isHostConfirmationEmailsDisabled = allowDisablingHostConfirmationEmails(workflows);
+        }
+
+        if (isAttendeeConfirmationEmailDisabled) {
+          isAttendeeConfirmationEmailDisabled = allowDisablingAttendeeConfirmationEmails(workflows);
+        }
+      }
+
+      await sendScheduledEmails(
```
CarinaWolli, line 99: Make sure confirmation emails are not sent when disabled because of an existing workflow

packages/features/bookings/lib/handleConfirmation.ts
```diff
@@ -169,6 +213,7 @@ export async function handleConfirmation(args: {
         references: {
           create: scheduleResult.referencesToCreate,
         },
+        metadata: { ...(typeof booking.metadata === "object" ? booking.metadata : {}), videoCallUrl },
```
CarinaWolli, line 216: bookings that require confirmation never had the videoCallUrl added to metadata

packages/features/bookings/lib/handleConfirmation.ts
```diff
@@ -218,8 +263,6 @@ export async function handleConfirmation(args: {
         const eventTypeSlug = updatedBookings[index].eventType?.slug || "";
 
         const isFirstBooking = index === 0;
-        const videoCallUrl =
```
CarinaWolli, line 221: The original booking didn't have videoCallUrl in metadata, that's why it was missing in the workflow emails that trigger when a meeting was confirmed

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2524,7 +2524,7 @@ async function handler(
         ...evt,
         ...{ metadata: metadataFromEvent, eventType: { slug: eventType.slug } },
       },
-      isNotConfirmed: evt.requiresConfirmation || false,
+      isNotConfirmed: !isConfirmedByDefault,
```
CarinaWolli, line 2527: also includes bookings with payments not just event types that require confirmation

emrysal (approved):
Looks good to me, but a add-on is needed soon to add tests to ensure emails aren't send when e.g. host confirmation emails are disabled.

kemilbeltre (approved):

PR #12302, opened 2023-11-09
fix: Org: Any team URL with slug starting with `d` is 404

hariombalhara:
## What does this PR do?

Fixes #12334

The reason behind the issue was that /d is a existing route for private links. The same issue would come if team slug starts with any of the registered routes names e.g /apps-something, /auth-something

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?
- Create a team with slug `dsomething` in an org
- Visit the team profile link i.e. app.cal.local:3000/dsomething. Notice that it is 404 in main but not in this branch

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #12311, opened 2023-11-10
fix: Broken embed in incognito of chrome

### LINKED ISSUE #12298: [CAL-2687] Embed is broken when third party cookies are disabled
Had two reports of this today. Seems like a recent change

This is the crash

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2687](https://linear.app/calcom/issue/CAL-2687/embed-is-broken-when-third-party-cookies-are-disabled)</sub>

hariombalhara:
## What does this PR do?

Fixes #12298
- Use localStorage from `@calcom/webStorage` instead of native `localStorage`, so that uncaught errors are handled.
- Newly added localStorage usage of getting `overlayCalendarSwitchDefault` is the reason behind it breaking.

**Note:** Latest chrome atleast 119(latest version) as this behaviour fixed automatically where accessing `localStorage` throws error, so as people update to latest versions of chrome, this problem should be solved for good anyway. 

To be able to replicate this behaviour make sure that Third party cookies are blocked. If that doesn't work, you can explicitly disable [chrome://flags/#third-party-storage-partitioning](chrome://flags/#third-party-storage-partitioning) flag
Screenshot 2023-11-10 at 12 38 37 PM

**Side Note:**  There is an ESLint rule that errors when using `localStorage` but it currently only detects `window.localStorage` usage and thus couldn't detect these issues as `localStorage` was used without using `window`.
One way to fix it could be to start using `webStorage.localStorage` everywhere and then we can prevent `window.localStorage` and `localStorage` usages.

## Type of change

## Mandatory Tasks

## Important Note
- Couldn't add Checkly test because it doesn't allow controlling the browser version and it is fixed in latest chrome.

### DIFF
--- a/packages/features/bookings/Booker/components/LargeCalendar.tsx
+++ b/packages/features/bookings/Booker/components/LargeCalendar.tsx
@@ -4,6 +4,7 @@ import dayjs from "@calcom/dayjs";
 import { Calendar } from "@calcom/features/calendars/weeklyview";
 import type { CalendarEvent } from "@calcom/features/calendars/weeklyview/types/events";
 import type { CalendarAvailableTimeslots } from "@calcom/features/calendars/weeklyview/types/state";
+import { localStorage } from "@calcom/lib/webstorage";
 
 import { useBookerStore } from "../store";
 import { useEvent, useScheduleForEvent } from "../utils/event";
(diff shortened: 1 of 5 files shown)


hariombalhara:
* main
  * **PR #12302** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #12311** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12311?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12311?utm_source=stack-comment).

PeerRich (approved):

PR #12316, opened 2023-11-10
feat: pipedrive crm app on cal

### LINKED ISSUE #12197: [CAL-2668] App: Pipedrive CRM
same as the other CRMs

<sub>[CAL-2668](https://linear.app/calcom/issue/CAL-2668/app-pipedrive-crm)</sub>

jatinsandilya:
## What does this PR do?

Closes #12197 

Loom demo: https://www.loom.com/share/5d5cf2091c2f4a80983b550938e81d5a 

## Requirement/Documentation

- Mentioned in the issue #12197 (same behaviour expected as other CRM apps)

## Type of change

## How should this be tested?

**Scenarios tested:** 

**When a meeting is booked with the Pipedrive CRM app installed,** 
* a meeting gets created inside Pipedrive 
* if the contact is not already present in Pipedrive, a new contact with the attendee email gets created inside Pipedrive. 
* When multiple attendees are present, only the ones that are not already present in the CRM get created in Pipedrive. 
* When a meeting gets deleted in Cal, it gets deleted in Pipedrive. 
* When a meeting gets updated in Cal, it gets updated in Pipedrive. 

## Mandatory Tasks


packages/app-store/pipedrive-crm/lib/CalendarService.ts
```diff
@@ -0,0 +1,313 @@
...
+
+export default class PipedriveCalendarService implements Calendar {
+  private log: typeof logger;
+  private tenantId: string;
+  private revertApiKey: string;
+  private revertApiUrl: string;
+  constructor(credential: CredentialPayload) {
+    this.revertApiKey = process.env.REVERT_API_KEY || "";
+    this.revertApiUrl = process.env.REVERT_API_URL || "https://api.revert.dev/";
+    this.tenantId = String(credential.teamId ? credential.teamId : credential.userId); // Question: Is this a reasonable assumption to be made? Get confirmation on the exact field to be used here.
```
jatinsandilya, line 45: **Question:** Is this a reasonable assumption to be made? 

Revert api expects a unique `tenantId` for each CRM/app. Since a CRM/app can be used by multiple users that are part of a team, can `teamId` be used here to uniquely identify a CRM?

PeerRich:
when using revert, is any customer data going through your systems?

jatinsandilya:
> when using revert, is any customer data going through your systems?

We don't store any customer data (deals, contact information etc). Its only transformed during transit when an API call is made. We only store the OAuth tokens for each connection similar to the way Cal stores them inside the `credentials` table. 

Does that make sense?

PeerRich:
added revert keys. havent added pipedrive keys yet

PeerRich (dismissed):
added all credentials

PeerRich:
building now on staging

PeerRich:
submitted app for review to pipedrive

PeerRich (dismissed):

.env.appStore.example
```diff
@@ -126,5 +126,16 @@ SALESFORCE_CONSUMER_SECRET=""
 ZOHOCRM_CLIENT_ID=""
 ZOHOCRM_CLIENT_SECRET=""
 
-
+# Used for the Pipedrive CRM integration
+PIPEDRIVE_CLIENT_ID=""
+PIPEDRIVE_CLIENT_SECRET=""
```
hariombalhara, line 131: I don't think we need these. These should be stored in DB and configurable via settings/admin/apps

hariombalhara: Just tested that the values are available there, so we can remove these safely. Should update the README as well to update the values in admin instead of env variables.
image

packages/app-store/pipedrive-crm/api/add.ts
```diff
@@ -0,0 +1,30 @@
...
+  await createDefaultInstallation({
+    appType: `${appConfig.slug}_other_calendar`,
+    userId: userId,
+    slug: appConfig.slug,
+    key: {},
+    teamId: Number(teamId),
+  });
+  const tenantId = teamId ? teamId : userId;
+  res.status(200).json({
+    url: `https://oauth.pipedrive.com/oauth/authorize?client_id=${process.env.PIPEDRIVE_CLIENT_ID}&redirect_uri=https://app.revert.dev/oauth-callback/pipedrive&state={%22tenantId%22:%22${tenantId}%22,%22revertPublicToken%22:%22${process.env.REVERT_PUBLIC_TOKEN}%22}`,
```
hariombalhara, line 27: Screenshot 2023-12-19 at 12 40 30 PM
You should read the pipedrive client_id and secret like this

packages/app-store/pipedrive-crm/lib/CalendarService.ts
hariombalhara: Looks right to me

turbo.json
```diff
@@ -344,6 +344,10 @@
     "ZOHOCRM_CLIENT_SECRET",
     "ZOOM_CLIENT_ID",
     "ZOOM_CLIENT_SECRET",
-    "RESEND_API_KEY"
+    "REVERT_API_KEY",
+    "REVERT_API_URL",
+    "REVERT_PUBLIC_TOKEN",
+    "PIPEDRIVE_CLIENT_ID",
+    "PIPEDRIVE_CLIENT_SECRET"
```
hariombalhara, line 351: Can be removed.

hariombalhara (changes requested):
Thanks @jatinsandilya!! Code looks good to me except that we should remove Pipedrive specific env variables.

.env.appStore.example
jatinsandilya: @hariombalhara I'm looking into this. Would you be able to tell me which action triggers this popup? 

Also after the lates pull from `main` I keep getting a `404` on the event-types link. 

Any idea why I might be facing this or how I could resolve this locally? 

Screenshot 2023-12-19 at 1 59 08 PM

jatinsandilya:
> Thanks @jatinsandilya!! Code looks good to me except that we should remove Pipedrive specific env variables.

Addressed these changes! Let me know if there's anything else. Thanks!

hariombalhara (dismissed):

hariombalhara (approved):

PR #12319, opened 2023-11-10
fix: Remove insights from nav map

sean-brydon:


emrysal (approved):
Let's go. 🙈

exception (approved):

PeerRich:
CleanShot 2023-11-10 at 12 22 07@2x
not needed

PR #11958, opened 2023-10-17
feat: managed event types v2

leog:
## What does this PR do?

Implements Managed Event Types v2 where the event owner/team admin can discretely select which field to lock or unlock in a managed event type.

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

@ciaranha just FYI, went for the badge with title and the switch as it was the one that can be universally applied to any type of field, whether is a text or a setting toggle.

Talked to @PeerRich and he instructed me to tag you @CarinaWolli for this in order to have you take over MET.

Pending tasks: 
 - [ ] E2E tests

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How should this be tested?

Create a team and then a managed event type. Upon getting to the event type settings page, see the different toggles to lock/unlock them.

PR #12320, opened 2023-11-10
feat: Managed Events V2

### LINKED ISSUE #9307: [CAL-2513] Unlocked Pre and Post Buffer Times for Managed Events

Our platform hosts live one on one soft skills trainings in the workplace. Our trainers are not technical and need as many things automated as possible. Managed events are perfect for this. They allow us to set all the properties that we want to be held consistent and then distribute them to our trainers. Our issue is that trainers need a widely varying amount of time in between trainings. Right now the times between events for a managed event type are locked at the admin level and cannot be set by each trainer. This either requires us to set a window that is comfortable for some but inefficient for others or set a window that is efficient for some and uncomfortable for others.

It would be very helpful for us if the managed event types allowed the individual team members to change the currently locked fields. Specifically for us that is the amount of time in between scheduled events (buffer times).

We are currently in Beta of our scheduling feature and have implemented Calendly. Calendly supports the feature described here. However, Calendly has presented other problems for us with their embed and how it fires (or doesn't fire) events related to event reschedule and cancel that require brittle technical workarounds that we are less than comfortable with. 

### Additional context

I recognize that this feature is in progress. My hope is that this request can provide context for why a feature like this matters in a real world use case.

Thanks for your consideration.

<sub>[CAL-2513](https://linear.app/calcom/issue/CAL-2513/unlocked-pre-and-post-buffer-times-for-managed-events)</sub>

### LINKED ISSUE #11941: [CAL-3185] Managed event type v2
Let admin decide which fields should be locked and unlocked for managed event types. 

<sub>[CAL-3185](https://linear.app/calcom/issue/CAL-3185/managed-event-type-v2)</sub>

alishaz-polymath:
## What does this PR do?

Continues the work done by @leog in #11958 👏 

Fixes #9307 #11941 

### Screenshots

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

[Loom](https://www.loom.com/share/483902d4c4154eea97dbbf2688feac3a)

- [x] This PR allows the Managed Events to be configurable enough that the Admin can lock/unlock fields. 
- [x] The values of the unlocked fields can be modified by the members. 
- [x] The field values updated by members persist if the Admin makes changes to any other fields in the event type, which is the desired outcome

Current quirks remaining and being fixed (checked once fixed)

- [x] The locked/unlocked changed state not persisting when moving from one tab to another within the event type as an Admin
- [x] Private URLs are currently excluded from being locked/unlocked and have thus been put as disabled. Currently working to sort this out
- [x] Fields contains "unlocked" labels in non-managed events as well, so they need to be removed to avoid any confusion

## Requirement/Documentation

- [REF](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/New-Features?type=design&node-id=8320-40915)
- For the locked/unlocked UI, went ahead with the badge with title and the switch as it was the one that can be universally applied to any type of field, whether is a text or a setting toggle.

## Type of change

## How should this be tested?

## Mandatory Tasks


CarinaWolli (commented):
Nice work, looks pretty good already 👏🏻 

Some things I found during testing: 

- When locking/unlocking fields and switching tabs before saving the changes are lost
- Any change in the parent resets everything from the children to the default values of the parent
  - Not sure about the actual field that is changed if it is unlocked ->  If an unlocked field is changed in the parent, should this change the value of all children?  
- When locking workflows and a user already has a personal workflow assigned (as it was unlocked before), it stays active. For all other field we change back to the default settings if locked, so I think we should also disable all personal workflows once locked

packages/features/eventtypes/lib/types.ts
```diff
@@ -0,0 +1,77 @@
...
+export type CustomInputParsed = typeof customInputSchema._output;
+
+export type AvailabilityOption = {
+  label: string;
+  value: number;
+  isDefault: boolean;
+  isManaged?: boolean;
+};
+
+export type FormValues = {
```
CarinaWolli, line 20: misses the following fields: 

```
isInstantEvent: boolean; 
lockTimeZoneToggleOnBookingPage: boolean;
onlyShowFirstAvailableSlot: boolean;
```

apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -84,68 +77,6 @@ const EventWebhooksTab = dynamic(() =>
 
 const ManagedEventTypeDialog = dynamic(() => import("@components/eventtype/ManagedEventDialog"));
 
-export type FormValues = {
```
Comment (CarinaWolli) [inline apps/web/pages/event-types/[type]/index.tsx:87]: import for `FormValues` needs to be fixed in `InstantEventController.tsx`

alishaz-polymath:
> Nice work, looks pretty good already 👏🏻
> 
> Some things I found during testing:
> 
>     * When locking/unlocking fields and switching tabs before saving the changes are lost

Thank you for this, will take a look 🙏 

> 
>     * Any change in the parent resets everything from the children to the default values of the parent
>       
>       * Not sure about the actual field that is changed if it is unlocked ->  If an unlocked field is changed in the parent, should this change the value of all children?

I think it should. If the Admin/Parent event type field is changed, it should apply to the children, locked or unlocked. Although, I see the problem where the Admin wishes to change the default (for future users) but that might just be over-engineering 🤔 

> 
>     * When locking workflows and a user already has a personal workflow assigned (as it was unlocked before), it stays active. For all other field we change back to the default settings if locked, so I think we should also disable all personal workflows once locked

Ah, right. I had a similar issue with Private URL as well. I'll check if the fix is straightforward for this. I have currently locked private url, for example. If the fix seems like a more complex thing to fix, probably better to do a follow up and fix exactly this kind of issue in v2.1. I'll check and report back 🙏 

Thank you for these shouts @CarinaWolli 👏

CarinaWolli:
> I think it should. If the Admin/Parent event type field is changed, it should apply to the children, locked or unlocked. Although, I see the problem where the Admin wishes to change the default (for future users) but that might just be over-engineering 🤔

I think the actual change is fine to be applied in all children. We can revisit if users run into problems or would expect different 👍

> Any change in the parent resets everything from the children to the default values of the parent 

This definitely shouldn't happen, only the field changed should be affected

alishaz-polymath:
> > I think it should. If the Admin/Parent event type field is changed, it should apply to the children, locked or unlocked. Although, I see the problem where the Admin wishes to change the default (for future users) but that might just be over-engineering 🤔
> 
> I think the actual change is fine to be applied in all children. We can revisit if users run into problems or would expect different 👍
> 
> > Any change in the parent resets everything from the children to the default values of the parent
> 
> This definitely shouldn't happen, only the field changed should be affected

Oh right, yeah this is what I meant. Any specific field change should update the value of that field in child event types

leog:
@alishaz-polymath is there anything I can help with? Sorry I didn't catch this up before.

alishaz-polymath:
> @alishaz-polymath is there anything I can help with? Sorry I didn't catch this before.

All good! Thank you for asking @leog 🙌

alishaz-polymath (commented):
Most things are self-explanatory, comments added to clarify the potentially confusing bits. Happy to provide more context, if needed.

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -106,23 +106,24 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps,
...
 
   const successRedirectUrlLocked = shouldLockDisableProps("successRedirectUrl");
   const seatsLocked = shouldLockDisableProps("seatsPerTimeSlotEnabled");
+  const requiresBookerEmailVerificationProps = shouldLockDisableProps("requiresBookerEmailVerification");
+  const hideCalendarNotesLocked = shouldLockDisableProps("hideCalendarNotes");
+  const lockTimeZoneToggleOnBookingPageLocked = shouldLockDisableProps("lockTimeZoneToggleOnBookingPage");
 
   const closeEventNameTip = () => setShowEventNameTip(false);
   const displayDestinationCalendarSelector =
     !!connectedCalendarsQuery.data?.connectedCalendars.length && !team;
```
joeauyeung, line 124: Would this also be a good opprotunity to allow children event types to select their own destination calendar?

joeauyeung (dismissed):
Great PR @alishaz-polymath I've gone through and tested each fields except for "Apps" and "Workflows" since I think those would need a bit more time. I've left a few comments so far.

### Blocking
- When an admin locks a previously unlocked field, I think we should push the parent field down to the children. Ex. if a child event type has a different title from the parent, if a field is going to be locked again there isn't a way to change the title of the child unless the field is unlocked, changed back, and then locked again
- I noticed when the parent makes changes and saves it's propagated to children event types even if that field is unlocked and there's a different value set from the parent
- When multiple durations are set, the toggle to lock/unlock the default value isn't changing
image
- Member's default location isn't being set in the field unless we choose something else
image
- When selecting offset start times, there's two toggles for lock/unlock but they're tied together. I noticed other options that are similar to this don't have the same toggles for the sub options. I think we should decide if we want to offer this level of granularity or not
image
- When toggling the "Enable Private URL", on and off we'll sometimes run into the prisma error `An operation failed because it depends on one or more records that were required but not found. No 'HashedLink' record was found for a nested delete on relation 'EventTypeToHashedLink'`. The second time I hit the save button it goes through.


### Non-blocking
- I noticed that for the single line text fields, when toggling the switch it focuses on the text field. I think we can make this looked more polished if we had these separated
- Likewise I'm not a huge fan of the changing sizes of the field lock toggle. Could it be a set width? 
V2 managed field
- Should we also lock/unlock the number of seats offered?
image

apps/web/components/eventtype/EventAdvancedTab.tsx
alishaz-polymath: Yeah, good spot. I think this makes sense!

alishaz-polymath:
> Great PR @alishaz-polymath I've gone through and tested each fields except for "Apps" and "Workflows" since I think those would need a bit more time. I've left a few comments so far.
> ### Blocking
> 
>     * When an admin locks a previously unlocked field, I think we should push the parent field down to the children. Ex. if a child event type has a different title from the parent, if a field is going to be locked again there isn't a way to change the title of the child unless the field is unlocked, changed back, and then locked again
  Asked @ciaranha for his inputs here ^, once we have finalized the ideal flow, I'll implement the changes
_________________________________________________________________________________
>     * I noticed when the parent makes changes and saves it's propagated to children event types even if that field is unlocked and there's a different value set from the parent
> 

Asked @ciaranha for his inputs here ^, once we have finalized the ideal flow, I'll implement the changes
_________________________________________________________________________________
>     * When multiple durations are set, the toggle to lock/unlock the default value isn't changing

Resolved ✅ 
_________________________________________________________________________________
>     * Member's default location isn't being set in the field unless we choose something else

Resolved ✅ 
 _________________________________________________________________________________
>     * When selecting offset start times, there's two toggles for lock/unlock but they're tied together. I noticed other options that are similar to this don't have the same toggles for the sub options. I think we should decide if we want to offer this level of granularity or not

This was unintended, fixed and Resolved ✅ 
_________________________________________________________________________________
>     * When toggling the "Enable Private URL", on and off we'll sometimes run into the prisma error `An operation failed because it depends on one or more records that were required but not found. No 'HashedLink' record was found for a nested delete on relation 'EventTypeToHashedLink'`. The second time I hit the save button it goes through.
 
I couldn't re-create this at all on my end, so perhaps I'd need more information on how to recreate this.
_________________________________________________________________________________
> 
> ### Non-blocking
> 
>     * I noticed that for the single line text fields, when toggling the switch it focuses on the text field. I think we can make this looked more polished if we had these separated
> 

Not entirely sure I follow here, when I click on the toggle, it doesn't switch focus to the text fields for me. Tested on Title
_________________________________________________________________________________
>     * Likewise I'm not a huge fan of the changing sizes of the field lock toggle. Could it be a set width?

I agree, Done & Resolved ✅ 
_________________________________________________________________________________
>     * Should we also lock/unlock the number of seats offered?

I think that's overkill tbh. Just the parent field being locked/unlocked should be enough IMO.

packages/features/ee/workflows/components/EventWorkflowsTab.tsx
```diff
@@ -223,13 +232,17 @@ function EventWorkflowsTab(props: Props) {
...
         (workflow) =>
+          (!workflow.teamId || eventType.teamId === workflow.teamId) &&
           !workflows
             .map((workflow) => {
               return workflow.id;
             })
             .includes(workflow.id)
       );
-      setSortedWorkflows(activeWorkflows.concat(disabledWorkflows));
+      const allSortedWorkflows = workflowsDisableProps.isLocked
```
CarinaWolli, line 242: for the parent managed event type we want to list all team workflows here even when locked

apps/web/components/eventtype/EventAppsTab.tsx
```diff
@@ -133,12 +136,28 @@ export const EventAppsTab = ({ eventType }: { eventType: EventType }) => {
...
               className="mb-2"
-              title={t("locked_for_members")}
-              message={t("locked_apps_description")}
+              title={
+                <Trans i18nKey={`${lockedText}_${isManagedEventType ? "for_members" : "by_team_admins"}`}>
+                  {lockedText[0].toUpperCase()}
+                  {lockedText.slice(1)} {isManagedEventType ? "for members" : "by team admins"}
+                </Trans>
+              }
+              actions={<div className="flex h-full items-center">{appsDisableProps.LockedIcon}</div>}
```
CarinaWolli, line 149: If enabling/disabling one toggle on this page all toggles changes
Screenshot 2024-03-01 at 3 31 50 PM

CarinaWolli (dismissed):
Awesome work 👏🏻 I tested everything and found a few small things that need to be addressed 🙏

joeauyeung:
@alishaz-polymath added some clarifying points

- When toggling the "Enable Private URL", on and off we'll sometimes run into the prisma error `An operation failed because it depends on one or more records that were required but not found. No 'HashedLink' record was found for a nested delete on relation 'EventTypeToHashedLink'`. The second time I hit the save button it goes through.
	- This might be related to propagating changes from the parent down to the child. Since the private URL is regenerated when the option is toggled, when the parent toggles the option the child private URL becomes broken. Which is causing the URL to become invalid or causes the DB issue above

- I noticed that for the single line text fields, when toggling the switch it focuses on the text field. I think we can make this looked more polished if we had these separated
	- In that GIF when we click the toggle, the field turns white and we can start typing. For other fields we don't do that so I think we should keep it consistent

apps/web/components/eventtype/EventAppsTab.tsx
alishaz-polymath: This is too granular for now, so only added the locked/unlocked for All apps.

apps/web/components/eventtype/EventWebhooksTab.tsx
```diff
@@ -4,8 +4,10 @@ import { Trans } from "next-i18next";
 import Link from "next/link";
```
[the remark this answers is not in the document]
alishaz-polymath: We can line it up for a follow up

CarinaWolli:
Is this ready for review again or still something missing?

alishaz-polymath:
> Is this ready for review again or still something missing?

Just waiting for the checks to pass 😅

joeauyeung (dismissed):
Retested my review comments
## Passing
- When an admin locks a previously unlocked field, I think we should push the parent field down to the children ✅ 

## Blocking
- Still getting errors around the private URL field. When the field is unlocked and the child has the option enabled. When toggling the option on the parent I think it's trying to generate new URLs for the children. If it's using Prisma's `createMany` I would look at the `skipDuplicates` option.
CleanShot 2024-03-08 at 11 39 31@2x

alishaz-polymath:
> Retested my review comments
> ## Passing
> 
>     * When an admin locks a previously unlocked field, I think we should push the parent field down to the children ✅
> 
> 
> ## Blocking
> 
>     * Still getting errors around the private URL field. When the field is unlocked and the child has the option enabled. When toggling the option on the parent I think it's trying to generate new URLs for the children. If it's using Prisma's `createMany` I would look at the `skipDuplicates` option.
>       CleanShot 2024-03-08 at 11 39 31@2x

Just managed to re-create the issue. Working on a fix 🙏

CarinaWolli (commented):
Now a user can enable a workflow on a managed event type even when it is locked. All my other comments are fixed 🙏

joeauyeung (commented):
All my blocking comments have been resolved. Great work @alishaz-polymath! I'll wait until @CarinaWolli's comments have been addressed before approving.

packages/features/ee/workflows/components/WorkflowDetailsPage.tsx
```diff
@@ -54,12 +54,18 @@ export default function WorkflowDetailsPage(props: Props) {
         if (teamId && teamId !== group.teamId) return options;
         return [
           ...options,
-          ...group.eventTypes.map((eventType) => ({
-            value: String(eventType.id),
-            label: `${eventType.title} ${
-              eventType.children && eventType.children.length ? `(+${eventType.children.length})` : ``
-            }`,
-          })),
+          ...group.eventTypes
```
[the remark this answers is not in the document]
alishaz-polymath: In this case as well, we do not want to allow users to create workflow when workflows are locked, correct? That should only be doable on Managed Event Level and not Child Event level, unless workflows are unlocked?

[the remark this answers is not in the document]
CarinaWolli: Yes. When locked, the parent managed event type can still set the event type active on a team workflow. But the user can't set a locked child managed event active on one of their workflows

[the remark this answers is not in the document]
alishaz-polymath: I think the latest changes sorts this. Can you test and confirm unless I missed something? 🙏

CarinaWolli (dismissed):
> All my blocking comments have been resolved. Great work @alishaz-polymath! I'll wait until @CarinaWolli's comments have been addressed before approving.

All my comments are also fixed. Looking good now 🙌🏻

alishaz-polymath:
Since this is Core, let's wait for Foundations team to give a green signal before we merge this 🙏

packages/features/ee/managed-event-types/hooks/useLockedFieldsManager.tsx
```diff
@@ -1,27 +1,102 @@
 // eslint-disable-next-line no-restricted-imports
```
emrysal, line 1: Just feedback on the naming, as this file doesn't only include the hook file this may be confusing.

packages/ui/components/form/switch/SettingsToggle.tsx
```diff
@@ -55,7 +55,8 @@ function SettingsToggle({
               <div>
                 <div className="flex items-center">
                   <Label
-                    className={classNames("mt-0.5 text-base font-semibold leading-none", labelClassName)}>
+                    className={classNames("mt-0.5 text-base font-semibold leading-none", labelClassName)}
+                    htmlFor="">
```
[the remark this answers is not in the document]
emrysal: ❓ I assume this was discussed elsewhere 😅  - OK

emrysal (dismissed):
Pushed some changes, but overall really happy with how this works. Happy to approve 👍  Well done @alishaz-polymath

emrysal (approved):

packages/features/ee/managed-event-types/hooks/useLockedFieldsManager.tsx
alishaz-polymath: Oops, a remnant from earlier changes 🤦 
Nice catch

PR #12322, opened 2023-11-10
fix: Fix recurring booking failure due to rate limiting

### LINKED ISSUE #10673: [CAL-2517] Recurring Event - Error while booking a meeting.
### Issue Summary
I get the Couldn't Book Meeting Error while trying to book a recurring meeting. 

### Steps to Reproduce

1. Book a recurring event type

Screenshot

[Chrome logs](https://paste.c-net.org/FelicityComrades)

<sub>[CAL-2517](https://linear.app/calcom/issue/CAL-2517/error-while-booking-a-meeting)</sub>

hariombalhara:
## What does this PR do?

"core" has a limit of 10 requests every 60s. Due to `handleNewBooking` called from `recurring-event` handler multiple times depending on the recurrence, it can fail.

Moved the rateLimiting logic to **actual route** to guarantee that it's called once per route.

Note: I think we should have some eslint rule to disallow rateLimiting in non api route files because it can accidentally cause issues. cc @sean-brydon @keithwillcode 

image

## Type of change

## How should this be tested?
- Create a stash account and setup env variables
- Reduce the "core" limit here to a small number say 2(https://github.com/calcom/cal.com/blob/15734921d34dec3af2b5a37ffa0e95fbde680541/packages/lib/rateLimit.ts#L58)
- Try booking a recurring event in main and it will fail(num intervals should alteast be 3)
- In this branch, it will pass
- Also, test that if in less than a minute 3 normals bookings are done. Third one fails.

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/api/pages/api/bookings/_post.ts
+++ b/apps/api/pages/api/bookings/_post.ts
@@ -1,8 +1,6 @@
 import type { NextApiRequest } from "next";
 
 import handleNewBooking from "@calcom/features/bookings/lib/handleNewBooking";
-import { checkRateLimitAndThrowError } from "@calcom/lib/checkRateLimitAndThrowError";
-import getIP from "@calcom/lib/getIP";
 import { defaultResponder } from "@calcom/lib/server";
 
 /**
@@ -207,15 +205,8 @@ import { defaultResponder } from "@calcom/lib/server";
 async function handler(req: NextApiRequest) {
   const { userId, isAdmin } = req;
   if (isAdmin) req.userId = req.body.userId || userId;
-  const userIp = getIP(req);
 
-  await checkRateLimitAndThrowError({
-    rateLimitingType: "core",
-    identifier: userIp,
-  });
-
-  const booking = await handleNewBooking(req);
-  return booking;
+  return await handleNewBooking(req);
 }
 
 export default defaultResponder(handler);
--- a/apps/api/pages/api/bookings/_post.ts
+++ b/apps/api/pages/api/bookings/_post.ts
@@ -1,6 +1,8 @@
 import type { NextApiRequest } from "next";
 
 import handleNewBooking from "@calcom/features/bookings/lib/handleNewBooking";
+import { checkRateLimitAndThrowError } from "@calcom/lib/checkRateLimitAndThrowError";
+import getIP from "@calcom/lib/getIP";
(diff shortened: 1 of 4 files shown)


hariombalhara:
* main
  * **PR #12322** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12322?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12322?utm_source=stack-comment).

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -630,12 +628,6 @@ async function handler(
   }
 ) {
   const { userId } = req;
-  const userIp = getIP(req);
-
-  await checkRateLimitAndThrowError({
```
hariombalhara, line 635: We shouldn't call rate limit here as this function is called by recurring-event route many times.

apps/api/pages/api/bookings/_post.ts
```diff
@@ -205,6 +207,13 @@ import { defaultResponder } from "@calcom/lib/server";
 async function handler(req: NextApiRequest) {
   const { userId, isAdmin } = req;
   if (isAdmin) req.userId = req.body.userId || userId;
+  const userIp = getIP(req);
+
+  await checkRateLimitAndThrowError({
```
keithwillcode, line 212: Why is this using "core" instead of "api"?

keithwillcode: We already have a global rate limit check by the API key

keithwillcode (changes requested):
Requesting changes for clarification on API route rate limiting

apps/api/pages/api/bookings/_post.ts
hariombalhara: `handleNewBooking` which was called from here was using core only. This is the reason.

hariombalhara: I think it makes sense to remove it then.

It also means that there were two different rate limitings in place for this route

keithwillcode (approved):

PR #12337, opened 2023-11-13
fix: getTeamOrThrow ran on every request

emrysal:
This creates unnecessary strain on the DB when someone is hitting the endpoint a lot.

### DIFF
--- a/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
+++ b/packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
@@ -32,8 +32,6 @@ type InviteMemberOptions = {
 };
 
 export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) => {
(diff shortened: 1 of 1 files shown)


sean-brydon (dismissed):

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
```diff
@@ -32,8 +33,9 @@ type InviteMemberOptions = {
 };
 
 export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) => {
-  const team = await getTeamOrThrow(input.teamId, input.isOrg);
-  const { autoAcceptEmailDomain, orgVerified } = getIsOrgVerified(input.isOrg, team);
+  await checkRateLimitAndThrowError({
+    identifier: `userId:${ctx.user.id}`,
```
sean-brydon, line 37:
```suggestion
    identifier: `invitedBy:${ctx.user.id}`,
```

keithwillcode (approved):

PR #12352, opened 2023-11-13
chore: variables names

Udit-takkar:


sean-brydon (approved):
LGTM - much easier to read now

PR #12353, opened 2023-11-13
chore: Clean Up Delete Credential Selected Calendar Error Message

joeauyeung:
## What does this PR do?

This PR aims to clean up an error message when deleting a calendar credential. Since last year, selected calendars have been deleted on cascade when the credential is deleted. There are still a lot of records that don't have this yet, so we have to keep these lines of code.

This is the error message that will be cleaned up.
CleanShot 2023-11-13 at 13 47 34

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Install two calendars of the same app (GCal)
- Select some calendars to check for conflicts for both credentials
- Delete one of the calendar apps
    - No error message should appear

## Mandatory Tasks

## Checklist


packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -328,52 +327,30 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
     }
   }
 
-  // Validated that credential is user's above
-  await prisma.credential.delete({
-    where: {
-      id: id,
-    },
-  });
```
joeauyeung, line 336: While valid while here because we still had the credential data in memory. I felt better moving this to the end.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -328,52 +327,30 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
...
-
   // Backwards compatibility. Selected calendars cascade on delete when deleting a credential
   // If it's a calendar remove it from the SelectedCalendars
   if (credential.app?.categories.includes(AppCategories.calendar)) {
-    const selectedCalendars = await prisma.selectedCalendar.findMany({
```
joeauyeung, line 341: We were searching for selected calendars that belonged to the same type of integration as the one being deleted. So if there was a second instance of a calendar credential with selected calendars, this query would return those records.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -328,52 +327,30 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
...
-                console.log(
-                  `Error deleting selected calendars for user ${user.id} and calendar ${credential.appId}. Could not find selected calendar.`
-                );
-              }
-              console.log(
-                `Error deleting selected calendars for user ${user.id} and calendar ${credential.appId} with error: ${error}`
-              );
-            });
-        });
-      }
```
joeauyeung, line 376: Because of the above reason, if these credentials were deleted on cascade, this would throw an error.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -328,52 +327,30 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
...
+
+    const calendarIds = calendars?.map((cal) => cal.externalId);
+
+    await prisma.selectedCalendar.deleteMany({
       where: {
         userId: user.id,
         integration: credential.type as string,
+        externalId: {
+          in: calendarIds,
+        },
```
joeauyeung, line 345: `deleteMany` will not throw an error if no records are found.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -328,52 +327,37 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
...
   if (credential.app?.categories.includes(AppCategories.calendar)) {
-    const selectedCalendars = await prisma.selectedCalendar.findMany({
-      where: {
-        userId: user.id,
-        integration: credential.type as string,
-      },
-    });
-
-    if (selectedCalendars.length) {
+    try {
```
joeauyeung, line 333: Wrapping in a try/catch block since if there is an error here. It shouldn't stop the deletion of the credential.

keithwillcode:
Tested on preview deploy. Deleting the calendar credential working fine still.

keithwillcode (approved):

PR #12354, opened 2023-11-13
feat: `AtomsProvider` component

Ryukemeister:
## What does this PR do?

Adds an `AtomsProvider` component which acts as a context provider for all other component wrapped around it (PostHog does something similar [here](https://posthog.com/docs/libraries/react))


supalarry (dismissed):
Looks good, but I would remove comments
`// context for the api key value` and `// custom hook to simplify using the context` because they restate what the code already does. You should only use comments if it's not possible to understand what is going on based on code itself, and even then before adding comments need to think if you can re-factor code without making comments.

supalarry (approved):

PR #12358, opened 2023-11-14
fix: Across Org Scenarios - Wrong links for event and team

### LINKED ISSUE #12002: [CAL-2629] Across Org scenarios
When moving existing users to an org, there would be some users who are part of a Team that shouldn't go to the organization.

We can fix migration script to handle not migrating those teams but such a user would see the event-types of those teams still

There are a few issues with it

* [X] Non Org Team Event-types use the org domain in dashboard. Though it would work on regular [cal.com/team/non-org-team](http://cal.com/team/non-org-team)
* [X] /teams route doesn't list those non-org teams.
* [X] In the team members listing, for a team that is part of the org but the member that is not part of the org, we use the org link instead of non-org link and it obviously takes to a 404 page.

  
* [X] A team event that has it's organizer from outside the org , would end-up sending non-org links in email to manage that event. Those would not be  404 because they would redirect appropriately. But we should still fix it as we don't intend to keep redirects forever. This scenario is currently possible due to migration to Org but in future it could be possible for a team to have across org team members. So, it's more logical to derive the domain from the event(instead of the organizer) as that domain is used to access the event
* [X] Consider a team that has been migrated to an Org with it's members staying outside the org. On the dashboard of non-org member of the team, we see wrong link for team event. It should have been calcom.cal.local:3000, but we see app.cal.local

We are working on implementing new OrgUserProfile schema(i.e. mono-user schema), which is going to allow a user to be part of multiple organizations, it should be taken care of with that as well.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2629](https://linear.app/calcom/issue/CAL-2629/across-org-scenarios)</sub>

### LINKED ISSUE #12317: [CAL-2692] A non-migrated(to Org) user is seeing wrong link for team events that he is part of.
More details on internal link
https://threads.com/thread/34546443956/34546443957?s=nmAyPyKBUMERjXfasrMQX5

<sub>[CAL-2692](https://linear.app/calcom/issue/CAL-2692/a-non-migratedto-org-user-is-seeing-wrong-link-for-team-events-that-he)</sub>

hariombalhara:
## What does this PR do?

This PR makes it more reliable to do partial migration of users or teams to an org so that during the migration there is no disruption of services even if the migration is really gradual.

Fixes #12002 
- [x] Wrong Avatar in Managed Event Type Assignment if the user isn't part of org - Started using non org link for avatar with orgId as param here also.
- [x] Wrong Avatar in Create EventType, Workflow, RoutingForm dropdowns - Similar to the fix for the previous point.
- [x] Booking a non-org team event that has an org member as the organizer was sending the wrong link in email for reschedule and cancel - Started using event to identify the domain instead of using the organizer.
	- [x] [Test] Added bookingScenario test for this.
- [x] Team Profile in dashboard was showing wrong URL for a team that was part of Org but the user seeing the page wasn't part of the org - Started using the team's parent to determine the domain instead of the logged In user.
	- [x] [Test] Added an e2e for this.
- [x] On /event-types, a non-org member(part of a team in an org) was seeing wrong link to the team events  Fixes #12317
	- [x] [Test] Added an e2e for this 	
- [x] Other team members page was showing wrong avatar and links. Fixed that

## Type of change

## Test Results
https://github.com/calcom/cal.com/actions/runs/7258635346

## How should this be tested?
See [this loom ](https://www.loom.com/share/f8ffb981613c408c81d0bf2a2744787c)

## Mandatory Tasks


hariombalhara:
* main
  * **PR #12358** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12358?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12358?utm_source=stack-comment).

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -291,7 +303,6 @@ export const getByViewerHandler = async ({ ctx, input }: GetByViewerOptions) =>
       ...group.metadata,
       teamId: group.teamId,
       membershipRole: group.membershipRole,
-      image: `${bookerUrl}/${group.profile.slug}/avatar.png`,
```
hariombalhara, line 294: No need to set it here, we are already setting it in the above code

apps/web/components/eventtype/EventTypeSingleLayout.tsx
```diff
@@ -235,10 +235,8 @@ function EventTypeSingleLayout({
     formMethods,
   ]);
 
-  const orgBranding = useOrgBranding();
-  const isOrgEvent = orgBranding?.fullDomain;
-  const permalink = `${orgBranding?.fullDomain ?? CAL_URL}/${
-    team ? `${!isOrgEvent ? "team/" : ""}${team.slug}` : eventType.users[0].username
+  const permalink = `${bookerUrl}/${
```
hariombalhara, line 238: An event-type, depending on if it belongs to a team/user in org or a team/user outside org, can have different booking links.
So, ensure that we use that event's booking URL instead of relying on the logged in user's organization url

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -44,6 +44,7 @@ export const mapMemberToChildrenOption = (
       username: member.username ?? "",
       membership: member.membership,
       eventTypeSlugs: member.eventTypes ?? [],
+      avatar: member.avatar,
```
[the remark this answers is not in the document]
emrysal: This is going to be deprecated nowadays, avatar field shouldn't be used in new code.

packages/lib/getBookerUrl/client.ts
```diff
@@ -0,0 +1,30 @@
...
+  }: {
+    orgSlug: string | null;
+    teamSlug: string | null;
+  },
+  options?: {
+    protocol: boolean;
+  }
+) => {
+  const bookerUrl = getBookerBaseUrlSync(orgSlug, options);
+  teamSlug = teamSlug ? teamSlug : "";
```
sean-brydon, line 25:
```suggestion
  teamSlug = teamSlug ||  "";
```

apps/web/components/eventtype/EventTeamTab.tsx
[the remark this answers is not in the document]
hariombalhara: What's the recommended way now ?

sean-brydon (dismissed):
Tested locally and can confirm all links for eventTypes and avatars for users in a team not in an org appears to be correct. 

Great work this must not have been easy to get your head around. Code looks great. Amazing coverage with the tests too!

sean-brydon (dismissed):

sean-brydon (approved):

PR #12363, opened 2023-11-14
fix: getting shimmer app keys

joeauyeung:
## What does this PR do?

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/app-store/shimmervideo/lib/getShimmerAppKeys.ts
```diff
@@ -8,6 +8,6 @@ const shimmerAppKeysSchema = z.object({
 });
 
 export const getShimmerAppKeys = async () => {
-  const appKeys = await getAppKeysFromSlug("shimmer-video");
+  const appKeys = await getAppKeysFromSlug("shimmervideo");
```
joeauyeung, line 11: Lines up with the slug in the config file

hariombalhara: We should read this from config

hariombalhara (approved):

Udit-takkar (approved):

PR #12369, opened 2023-11-14
feat: Base implementation of v2 of avatars

emrysal:
## What does this PR do?

Team ID - without user ID:
This is a team logo.

Team ID - with user ID:
This is a user avatar in the context of an organization.

User ID - without team ID:
This is just a user avatar (no context)


packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -31,12 +32,36 @@ type UpdateProfileOptions = {
   input: TUpdateProfileInputSchema;
 };
 
+const uploadAvatar = async ({ userId, avatar: data }: { userId: number; avatar: string }) => {
+  const objectKey = uuidv4();
+
+  await prisma.avatar.upsert({
+    where: {
+      teamId_userId: {
```
emrysal, line 40: Ideally we'd be able to use `null` for teamId in this situation but this is something Prisma does not support. Union types are non-nullable. In this situation that is OK as we can use `0` instead.

apps/web/components/ui/avatar/UserAvatar.tsx
```diff
@@ -14,6 +14,6 @@ type UserAvatarProps = Omit<React.ComponentProps<typeof Avatar>, "alt" | "imageS
  * It is aware of the user's organization to correctly show the avatar from the correct URL
  */
 export function UserAvatar(props: UserAvatarProps) {
-  const { user, previewSrc, ...rest } = props;
-  return <Avatar {...rest} alt={user.name || ""} imageSrc={previewSrc ?? getUserAvatarUrl(user)} />;
+  const { user, previewSrc = getUserAvatarUrl(user), ...rest } = props;
+  return <Avatar {...rest} alt={user.name || ""} imageSrc={previewSrc} />;
```
exception, line 18: `question-non-blocking`: Would it not be better for the alt to be something like `Unknown Name` rather than empty?

apps/web/pages/api/avatar/[uuid].ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const decoded = img.toString().replace("data:image/png;base64,", "").replace("data:image/jpeg;base64,", "");
+  const imageResp = Buffer.from(decoded, "base64");
+
+  res.writeHead(200, {
+    "Content-Type": "image/png",
+    "Content-Length": imageResp.length,
+  });
+
+  res.end(imageResp);
```
Comment (exception) [inline apps/web/pages/api/avatar/[uuid].ts:60]:
```suggestion
   try {
    // Fetch the avatar image data
    const { data: img } = await prisma.avatar.findUniqueOrThrow({
      where: { objectKey },
      select: { data: true },
    });

    // Decode the image data
    const decoded = img.toString().replace(/^data:image\/(png|jpeg);base64,/, "");
    const imageBuffer = Buffer.from(decoded, "base64");

    // Send the image response
    res.writeHead(200, {
      "Content-Type": "image/png",
      "Content-Length": imageBuffer.length,
    });
    res.end(imageBuffer);
  } catch (e) {
    // Handle errors (e.g., avatar not found) by redirecting to the default avatar
    res.writeHead(302, { Location: AVATAR_FALLBACK });
    res.end();
  }
```

apps/web/pages/settings/my-account/profile.tsx
```diff
@@ -432,7 +428,7 @@ const ProfileForm = ({
             control={formMethods.control}
             name="avatar"
             render={({ field: { value } }) => {
-              const showRemoveAvatarButton = !checkIfItFallbackImage(value);
+              const showRemoveAvatarButton = value === null ? false : !isFallbackImg;
```
exception, line 431: `nitpick`: `const showRemoveAvatarButton = value !== null && !isFallbackImg;`

apps/web/components/ui/avatar/UserAvatar.tsx
emrysal: Not a bad shout, will get that sorted.

apps/web/pages/api/avatar/[uuid].ts
Comment (emrysal) [reply apps/web/pages/api/avatar/[uuid].ts:60]: This was mentioned before but there is a good reason I want to have the try/catch as short as possible, any error thrown - however unlikely, will cause really weird 302 behaviour, this is why the catch block is so small and the img data is assigned to a variable instead.

Comment (sean-brydon) [reply apps/web/pages/api/avatar/[uuid].ts:60]: I suggested the same - But i agree with alex here i think its best to keep this how it is

sean-brydon (approved):
Tested - works perfectly. 
- OG Image 
- Booking page
- teams page
- orgs 

Everywhere i have tested seems to work 🙏

PeerRich (approved):

PR #12373, opened 2023-11-15
chore: Insights readonly DB client

exception:
## What does this PR do?

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Add a new `INSIGHTS_DATABASE_URL` env var to your `.env`, it can point to the same database if you only have one.
- Go to the Insights tab and validate that the required data is coming through.

## Mandatory Tasks


packages/prisma/index.ts
```diff
@@ -59,6 +59,10 @@ const prismaWithClientExtensions = prismaWithoutClientExtensions
 
 export const prisma = globalForPrisma.prismaWithClientExtensions || prismaWithClientExtensions;
 
+export const readonlyPrisma = customPrisma({
+  datasources: { db: { url: process.env.INSIGHTS_DATABASE_URL } },
```
keithwillcode, line 63: We need to check and make sure this env variable is present and has data, otherwise, it should fall back to the DATABASE_URL. This helps self-hosters upgrade and not run into problems and also not have to install a separate DB.

exception: Done!

keithwillcode:
@exception Can you update the description to clear up bullet points that need to be there + add testing steps?

zomars (approved):
Code LGTM

 LGTM

PR #12384, opened 2023-11-16
fix: Update yarn.lock

hariombalhara:


emrysal (approved):
Let's gooo

PeerRich:
empty yarn.lock. someone must've updated it already

PR #12403, opened 2023-11-17
fix: better errors for googleCalendar integration CAL-2721

ThyMinimalDev:
## What does this PR do?

Fixes CAL-2721

app.cal.com/api/integrations/googlecalendar/... api endpoints are not redacting errors before sending to frontend

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks


joeauyeung (commented):
The code looks like it works but what is the purpose of this PR? I still see the error message on the frontend. 

Ex. 
CleanShot 2023-11-20 at 18 01 20

ThyMinimalDev:
> The code looks like it works but what is the purpose of this PR? I still see the error message on the frontend.
> 
> Ex. CleanShot 2023-11-20 at 18 01 20

the goal is not to hide errors, the goal is to avoid cases where prisma full error stack is displayed on the frontend

joeauyeung (approved):
Thanks for the explanation @ThyMinimalDev. I can't get the full Prisma stack to appear on the client and everything works as normal so approving.

PR #12414, opened 2023-11-17
chore: Enable One Payment App Per Event Type

joeauyeung:
## What does this PR do?

This PR is stacked on #12272

Until we create the UI to allow bookers to choose which payment app to use, we should only enable one payment app per event type.

https://www.loom.com/share/dbec216876c24dac998cc12db7595814

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/app-store/_utils/payments/checkForMultiplePaymentApps.ts
```diff
@@ -0,0 +1,40 @@
...
+      }
+    } else if ("price" in app && app.enabled) {
+      enabledPaymentApps++;
+    }
+
+    if (inclusive && enabledPaymentApps >= 1) {
+      return true;
+    } else if (!inclusive && enabledPaymentApps > 1) {
+      return true;
+    }
```
joeauyeung, line 34: This function is being used in two places. One where we want the logic to be one or more and the other being more than one. Rather than repeating the same function pass a parameter to see if this check is inclusive of one.

ThyMinimalDev (commented):
LGTM, but why is this pr pointing to fix/update-eventype-price branch and not main ?

joeauyeung:
> LGTM, but why is this pr pointing to fix/update-eventype-price branch and not main ?

This is a stacked PR on top of https://github.com/calcom/cal.com/pull/12272

I like taking this approach because we can get the bug fix in #12272 merged while this is still under review. I also like the idea that each PR should have one change, making reviewing PRs easier as there are fewer files to look at.

packages/app-store/_utils/payments/checkForMultiplePaymentApps.ts
ThyMinimalDev: I think we can directly return the booleans here

```return (inclusive && enabledPaymentApps >= 1) || (!inclusive && enabledPaymentApps > 1);
```

ThyMinimalDev (approved):
LGTM

PR #12424, opened 2023-11-18
feat: More admin options for organizations

hariombalhara:
## What does this PR do?
Stacked over https://github.com/calcom/cal.com/pull/12358. Please merge that first.
### Admin Options for Organizations
- Ability to publish an org 
- Edit
	- Name
	- orgAutoAcceptEmail
	- Slug
		-  Delete previous domain and record on Vercel and Cloudflare
		-  Create new domain and record on Vercel and Cloudflare
- Ability to delete an organization.
	- Delete domain on Vercel
	- Delete DNS record on cloudflare

[Loom Demo](https://www.loom.com/share/3af28427fb7b41ed944551316166e42b)

### Other fixes
- Fixes booker URLs shown in Other Team Profile View and Create New Team Views
- Disable Error Redaction locally.

## Type of change

## How should this be tested?

## Mandatory Tasks


hariombalhara:
* main
  * **PR #12358** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12358?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #12424** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/12424?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/12424?utm_source=stack-comment).

packages/lib/redactError.ts
```diff
@@ -18,7 +20,7 @@ export const redactError = <T extends Error | unknown>(error: T) => {
     return error;
   }
   log.debug("Type of Error: ", error.constructor);
-  if (shouldRedact(error)) {
+  if (shouldRedact(error) && IS_PRODUCTION) {
```
hariombalhara, line 23: Let's redact on production only. Keeping things convenient locally.

packages/trpc/server/routers/viewer/organizations/_router.tsx
```diff
@@ -114,7 +109,7 @@ export const viewerOrganizationsRouter = router({
     const handler = await importHandler(namespaced("getOtherTeam"), () => import("./getOtherTeam.handler"));
     return handler(opts);
   }),
-  listOtherTeams: authedOrgAdminProcedure.query(async (opts) => {
+  listOtherTeams: authedProcedure.query(async (opts) => {
```
hariombalhara, line 112: It was unnecessarily causing 403 for all non admin members. Instead, let the route itself handle it and return empty array for non admin members,

packages/trpc/server/routers/viewer/organizations/_router.tsx
```diff
@@ -64,14 +67,6 @@ export const viewerOrganizationsRouter = router({
...
   }),
-  adminGetAll: authedAdminProcedure.query(async (opts) => {
-    const handler = await importHandler(namespaced("adminGetAll"), () => import("./adminGetAll.handler"));
-    return handler(opts);
-  }),
-  adminVerify: authedAdminProcedure.input(ZAdminVerifyInput).mutation(async (opts) => {
-    const handler = await importHandler(namespaced("adminVerify"), () => import("./adminVerify.handler"));
-    return handler(opts);
-  }),
   listMembers: authedProcedure.input(ZListMembersSchema).query(async (opts) => {
```
hariombalhara, line 70: Moved below to keep all admin endpoints together.

packages/trpc/server/routers/viewer/teams/delete.handler.ts
```diff
@@ -73,11 +38,7 @@ export const deleteHandler = async ({ ctx, input }: DeleteOptions) => {
 
   const deletedTeamMetadata = teamMetadataSchema.parse(deletedTeam.metadata);
 
-  if (IS_PRODUCTION)
```
hariombalhara, line 76: Let us allow testing domain deletion locally as well.

packages/features/ee/organizations/pages/settings/admin/AdminOrgEditPage.tsx
```diff
@@ -0,0 +1,107 @@
+import type { Team } from "@prisma/client";
```
hariombalhara, line 1: Borrowed some of the stuff from AdminOrgPage

packages/features/ee/organizations/pages/settings/admin/AdminOrgPage.tsx
```diff
@@ -54,7 +92,7 @@ function AdminOrgTable() {
                   <span className="text-default">{org.name}</span>
                   <br />
                   <span className="text-muted">
-                    {org.slug}.{extractDomainFromWebsiteUrl}
+                    {org.slug}.{subdomainSuffix()}
```
hariombalhara, line 95: Fixes the domain shown

packages/trpc/server/routers/viewer/organizations/adminVerify.schema.ts
```diff
@@ -1,10 +1,7 @@
 import { z } from "zod";
 
-const statusSchema = z.enum(["ACCEPT", "DENY"] as const);
-
 export const ZAdminVerifyInput = z.object({
   orgId: z.number(),
-  status: statusSchema,
```
hariombalhara, line 7: Wasn't being used.

PeerRich:
great work, super pumped about it

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
```diff
@@ -0,0 +1,70 @@
...
+export default adminDeleteHandler;
+
+async function renameUsersToAvoidUsernameConflicts(users: { id: number; username: string | null }[]) {
+  for (const user of users) {
+    await prisma.user.update({
+      where: {
+        id: user.id,
+      },
+      data: {
+        username: `${user.username || ""}-${user.id}`,
```
sean-brydon, line 66: I think we should actually specific this -> You can very easily get a cal account with no username set and having a user with `-519` isnt something we normally allow i dont think 

```suggestion
        username: `${user.username || "nameless_user"}-${user.id}`,
```

packages/trpc/server/routers/viewer/organizations/adminGet.handler.ts
```diff
@@ -0,0 +1,60 @@
...
+    user: NonNullable<TrpcSessionUser>;
+  };
+  input: TAdminGet;
+};
+
+export const adminGetHandler = async ({ input }: AdminGetOptions) => {
+  const org = await prisma.team.findUnique({
+    where: {
+      id: input.id,
+      AND: [
```
sean-brydon, line 20: Just a little idea i had when reading this - i wonder if it would be easier to query by  just team ID -> parse metadata -> check its set and throw from there. 

(Probably ms saved no biggy)

packages/lib/domainManager/deploymentServices/cloudflare.ts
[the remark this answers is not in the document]
hariombalhara: Yeah I thought about using classes but decided against it initially as it would add some complexity. IMHO, when we add another service we could refactor then.

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  return {
+    ok: true,
+    message: `Organization ${foundOrg.name} deleted.`,
+  };
+};
+
+export default adminDeleteHandler;
+
+async function renameUsersToAvoidUsernameConflicts(users: { id: number; username: string | null }[]) {
```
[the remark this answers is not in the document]
hariombalhara: I doubt that it would happen because X is user.id which is always incrementing even when users are deleted.

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
hariombalhara: What's the case when there is no username ? Does it happen when we invite a user?

I actually wanted to avoid `nameless` but I think it is a better approach when we know for sure that having null username is possible.

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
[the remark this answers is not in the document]
hariombalhara: Added an inline comment for this as well.

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
sean-brydon: Its very rare but you can replicate it by
1) invite a new user
2) new user doesnt want to join a team 
3) resets password
4) logs in 

We should probably work on re-directing to onboarding if no username is set. Or ensure user.username exists bfore we delete them? Not sure the best approach around that

packages/features/ee/organizations/pages/settings/admin/AdminOrgPage.tsx
```diff
@@ -145,3 +228,54 @@ const AdminOrgList = () => {
...
+        <Trans
+          i18nKey="admin_delete_organization_description"
+          components={{ li: <li />, ul: <ul className="ml-4 list-disc" /> }}>
+          <ul>
+            <li>
+              Teams that are member of this organization will also be deleted along with their event-types
+            </li>
+            <li>
+              Users that were part of the organization will not be deleted and their event-types will also
+              remain intact.
```
Udit-takkar, line 262: NIT: use i18n

hariombalhara: It already is. Notice `i18nKey`

packages/trpc/server/routers/viewer/organizations/adminGet.handler.ts
hariombalhara: Yeah, makes a lot of sense actually.

hariombalhara: Fixed

packages/lib/domainManager/deploymentServices/vercel.ts
```diff
@@ -0,0 +1,147 @@
...
+      })
+    );
+    throw new HttpError({
+      message: errorMessage,
+      statusCode: 400,
+    });
+  }
+  if (error.code === "domain_already_in_use") {
+    return true;
+  }
```
ThyMinimalDev, line 92: should we have extra information here on the reason for this error.code if statement to just return true, maybe some comments or logs

packages/lib/domainManager/deploymentServices/vercel.ts
```diff
@@ -0,0 +1,147 @@
...
+  TEAM_ID_VERCEL: string;
+  AUTH_BEARER_TOKEN_VERCEL: string;
+} & typeof process.env {
+  if (!env.PROJECT_ID_VERCEL) {
+    throw new Error("Missing env var: PROJECT_ID_VERCEL");
+  }
+
+  if (!env.TEAM_ID_VERCEL) {
+    // TEAM_ID_VERCEL is optional
+  }
```
ThyMinimalDev, line 142: are we planing to add something in this if statement ?

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
```diff
@@ -0,0 +1,80 @@
...
```
ThyMinimalDev, line 61: no need to have any custom check for premium username here I suppose ?

packages/trpc/server/routers/viewer/organizations/adminUpdate.handler.ts
```diff
@@ -0,0 +1,93 @@
...
+    throw new HttpError({
+      message: "There can only be one organization with a given slug",
+      statusCode: 400,
+    });
+  }
+
+  const foundOrg = organizationsWithSameSlug[0];
+  if (!foundOrg) {
+    return;
+  }
```
ThyMinimalDev, line 84: shouldn't we return an error here ? or some information to display on the frontend ?

packages/trpc/server/routers/viewer/organizations/adminDelete.handler.ts
hariombalhara: Yeah I don't think so. I don't think premium usernames are anyways applicable for an organization.

packages/trpc/server/routers/viewer/organizations/adminUpdate.handler.ts
hariombalhara: This is actually the happy case as there is no org that has the same slug. So, no conflict and no need to throw an error.

hariombalhara:
Thanks for the review @ThyMinimalDev. Addressed all comments.

Also, FYI it's a stacked PR. So, I would really appreciate if you could [review the target branch PR ](https://github.com/calcom/cal.com/pull/12358) as well.

hariombalhara:
@ThyMinimalDev No longer a stacked PR now. It's targeted to main now. If you could do a final review that would be great.

ThyMinimalDev (approved):
looking good

PR #12436, opened 2023-11-20
feat: validate api keys in cal provider

Ryukemeister:
## What does this PR do?

Validates api keys in `CalProvider` component before sending it down to child components


packages/atoms/cal-provider/index.tsx
```diff
@@ -11,5 +11,30 @@ const ApiKeyContext = createContext("");
...
 export function CalProvider({ apiKey, children }: CalProviderProps) {
-  return <ApiKeyContext.Provider value={apiKey}>{children}</ApiKeyContext.Provider>;
+  const [key, setKey] = useState("");
+
+  useEffect(() => {
+    async function verifyKey(key: string) {
+      try {
+        // here we'll call the /me endpoint in v2 to get user profile
+        // v2 is not ready yet thats why calling localhost for now
+        const response = await fetch(`http://localhost:3002/v1/me?apiKey=${key}`);
```
Udit-takkar, line 21: you shouldn't hard code localhost:3002

Ryukemeister: The original endpoint to be called would be something like `https://api.cal.com/v2/me` but since the v2 api is not ready, for the time being I was calling localhost

supalarry (changes requested):
Can you please move `verifyKey` as separate a `useCallback` function within the component and then within `useEffect` simply invoke it in the else block like it's now already?

Ryukemeister:
Closing this in favour of #12672

PR #12439, opened 2023-11-20
fix: Better error reporting & some fixes to deleteUser API

emrysal:
## What does this PR do?

Started out as a bit of a NIT to remove user errors from our application errors; but ended up fixing some inconsistencies in the deleteUser flow.


sean-brydon (dismissed):
LGTM - tested works. Good spot on the inconsistencies

Udit-takkar (dismissed):
LGTM

sean-brydon (approved):
Thanks for the rename ;) 

LGTM

Issue #12286, opened 2023-11-08
Documented fields are missing from some webhook events


nicolls1:
### Issue Summary

Documented fields are missing from some webhook events. Additionally, these missing fields will cause a custom payload template to be ignored and the default one to be used instead.

### Steps to Reproduce

1. Create new webhook and set `Payload Template` to `Custom` and enter:
```
{
  "triggerEvent": "{{triggerEvent}}",
  "type": "{{type}}"
}
```
2. Point to testing site (https://webhook.site/)
3. Create new meeting
4. Look at webhook event with triggerEvent `BOOKING_CREATED` and notice the webhook correctly shows only the 2 fields in the template
5. Wait for meeting to end
6. Look at the webhook event with triggerEvent `MEETING_ENDED` and notice that the full event is returned and there is no `type` field on that event

I also noticed that the "metadata" field was missing from the triggerEvent: BOOKING_CANCELLED. I did not go through every triggerEvent and look at every field. These are the only ones I have noticed up to this point.

### Actual Results

Not all documented webhook variables are included in all triggerEvents.

### Expected Results

All documented webhook variables are included in all triggerEvents.

### Technical details

Create brand new account at cal.com and test on live version following steps above.

### Evidence

Create brand new account at cal.com and test on live version following steps above.

PR #12445, opened 2023-11-20
fix: include eventTypeId in BOOKING_CANCELLED event

nicolls1:
## What does this PR do?

This addresses one missing field in webhook events as mentioned here: #12286. My integration is breaking because this field is missing.

## Requirement/Documentation

This updates the actual payload to be in line with the current documentation

## Type of change

## How should this be tested?

I could not find any unit tests to update for this, happy to do so if I missed them.

## Mandatory Tasks

### DIFF
--- a/packages/features/bookings/lib/handleCancelBooking.ts
+++ b/packages/features/bookings/lib/handleCancelBooking.ts
@@ -227,6 +227,7 @@ async function handler(req: CustomRequest) {
     type: bookingToDelete?.eventType?.slug as string,
     description: bookingToDelete?.description || "",
     customInputs: isPrismaObjOrUndefined(bookingToDelete.customInputs),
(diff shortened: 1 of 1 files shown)


jkdowdle:
@Udit-takkar @CarinaWolli If this were to merge, when could we expect it to be deployed and available for use on the cal.com services?

swain:
@CarinaWolli Worth noting: the failed status check here is only present because of a fluke Docker failure on one of the E2E test runs. It just needs to be re-triggered.

PeerRich (approved):

PeerRich:
LGTM

aecorredor:
@PeerRich thanks for getting this merged so quickly!

One more question: how does the deployment cycle work for `main`? At what point does it end up in your production environment for customers who use the cloud solution?

PR #12460, opened 2023-11-21
fix: alby payment could not be created

ThyMinimalDev:
## What does this PR do?

Fixes "Alby: Payment could not be created"

alby invoice now contains a field called isPaid, this field is not a boolean but a promise returning a boolean, we were trying to serialize isPaid which caused an error.

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

### DIFF
--- a/packages/app-store/alby/lib/PaymentService.ts
+++ b/packages/app-store/alby/lib/PaymentService.ts
@@ -87,8 +87,7 @@ export class PaymentService implements IAbstractPaymentService {
       }
       return paymentData;
     } catch (error) {
-      console.error(error);
-      log.error("Alby: Payment could not be created", bookingId);
+      log.error("Alby: Payment could not be created", bookingId, JSON.stringify(error));
       throw new Error(ErrorCode.PaymentCreationFailure);
     }
   }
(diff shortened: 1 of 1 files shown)


packages/app-store/alby/lib/PaymentService.ts
```diff
@@ -84,6 +87,7 @@ export class PaymentService implements IAbstractPaymentService {
       }
       return paymentData;
     } catch (error) {
+      console.error(error);
       log.error("Alby: Payment could not be created", bookingId);
```
Udit-takkar, line 91: @ThyMinimalDev how about this instead?

`log.error("Alby: Payment could not be created", bookingId,  JSON.stringify(error));
`

PeerRich (dismissed):

PeerRich (approved):

PeerRich (approved):

packages/app-store/alby/lib/PaymentService.ts
```diff
@@ -84,7 +87,7 @@ export class PaymentService implements IAbstractPaymentService {
       }
       return paymentData;
     } catch (error) {
-      log.error("Alby: Payment could not be created", bookingId);
+      log.error("Alby: Payment could not be created", bookingId, JSON.stringify(error));
```
hariombalhara, line 90: We could use `safeStringify` here. It's shorter plus prevents accidental crash in case we stringify a cyclic object

hariombalhara (approved):

Issue #12470, opened 2023-11-21
MEETING_ENDED webhook has incorrect content type


aecorredor:
### Issue Summary

Our webhook endpoint checks for `content-type` to be `application/json` for `POST` requests. It caught an error for the cal.com `MEETING_ENDED` event where the POST request is incorrectly setting  `content-type = text/plain;charset=UTF-8` when in reality it is sending JSON data.

### Steps to Reproduce

1. Configure a `MEETING_ENDED` webhook.
2. Consume the webhook from an API.
3. Schedule an appointment and let it finish so the `MEETING_ENDED` event triggers your API.
4. See the incorrect `content-type` come in.

### Actual Results

- The `content-type` is `text/plain;charset=UTF-8` in the `MEETING_ENDED` event.

### Expected Results

- The `content-type` is `application/json` in the `MEETING_ENDED` event.

### Technical details

N/A

### Evidence

- This was tested through our own webhook integration.

PR #12484, opened 2023-11-22
fix: meeting ended incorrect content type

Amit91848:
## What does this PR do?

Fixes #12470 
/claim #12470 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #12488, opened 2023-11-22
fix: MEETING_ENDED webhook has incorrect content type

ibex088:
## What does this PR do?

 claim #12470

### DIFF
--- a/packages/features/webhooks/lib/cron.ts
+++ b/packages/features/webhooks/lib/cron.ts
@@ -2,6 +2,7 @@
 import type { NextApiRequest, NextApiResponse } from "next";
 
 import dayjs from "@calcom/dayjs";
+import sendPayload from "@calcom/features/webhooks/lib/sendPayload";
 import { defaultHandler } from "@calcom/lib/server";
 import prisma from "@calcom/prisma";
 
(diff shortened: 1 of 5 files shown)


packages/features/webhooks/lib/cron.ts
```diff
@@ -19,15 +20,32 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
   // run jobs
   for (const job of jobsToRun) {
     try {
-      await fetch(job.subscriberUrl, {
-        method: "POST",
-        body: job.payload,
-      });
+      if (job.webhook) {
+        const payload = JSON.parse(job.payload);
+        await sendPayload(
```
SomayChauhan, line 33: I opt for using sendPayload because it takes care of all the logic involving the secretKey, which is used to send various other webhooks.
believe we should also utilize it for scheduledWebhooks, which we haven't been doing until now.

packages/features/webhooks/lib/cron.ts
```diff
@@ -19,15 +20,32 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+          job?.webhook?.secret,
+          payload?.triggerEvent,
+          new Date().toISOString(),
+          job.webhook,
+          payload
+        );
+      } else {
+        await fetch(job.subscriberUrl, {
```
SomayChauhan, line 41: this is for backwards compatibility

CarinaWolli:
Already fixed with https://github.com/calcom/cal.com/pull/12484

PR #12490, opened 2023-11-22
feat: API for getting an event type by slug

### LINKED ISSUE #12411: [CAL-2745] API for getting an event type by slug

### Additional context

### Requirement/Document

N/A

jemiluv8:
## What does this PR do?
Fixes #12411
/claim #12411

Fixes # (issue)

## Requirement/Documentation

Issue #12411

## Type of change

## How should this be tested?

- [x] use postman to test the api for various cases

TEST CASES
- [x] unauthorized user - without credentials
- [x] slug not found
- [x] slug exists but not for user
- [x] slug exists for user  

## Mandatory Tasks

## Checklist

- ~~I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)~~
- ~~My code doesn't follow the style guidelines of this project~~
- ~~I haven't commented my code, particularly in hard-to-understand areas~~
- ~~I haven't checked if my PR needs changes to the documentation~~
- ~~I haven't checked if my changes generate no new warnings~~
- ~~I haven't added tests that prove my fix is effective or that my feature works~~
- ~~I haven't checked if new and existing unit tests pass locally with my changes~~


apps/web/pages/api/event-types/[slug].ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const parsedQuery = querySchema.safeParse(req.query);
+  if (!parsedQuery.success) throw new HttpError({ statusCode: 400, message: parsedQuery.error.message });
+
+  const {
+    data: { slug },
+  } = parsedQuery;
+  if (!slug) return res.status(400).json({ message: "event-type slug is needed" });
+
+  const eventType = await prisma.eventType.findFirst({
```
Comment (jemiluv8) [inline apps/web/pages/api/event-types/[slug].ts:33]: This only applies to event-types "owned" by whoever is accessing the api.
I saw multiple users may be allowed access to the same event-type but not sure
if that should be enforced in this endpoint

For instance I initially had something to check if user is allowed access to the matched event type
```ts
  const canAccessEventType = eventType.users.some((user) => user.id === session.user.id);
  if (!canAccessEventType) return res.status(409).json({ message: "Unauthorized" });
```

Had to remove it.

Also I just returned a minimal payload but that can be updated as needed.

PR #12491, opened 2023-11-22
feat: support slug and id on GET event-types

### LINKED ISSUE #12411: [CAL-2745] API for getting an event type by slug

ouwargui:
## What does this PR do?

Fixes #12411
/claim #12411

## Type of change

## How should this be tested?

  - No
  - `yarn db-seed`
  - `/api/event-types/:id` should support id or slug as path param

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/api/pages/api/event-types/[id]/_auth-middleware.ts
+++ b/apps/api/pages/api/event-types/[id]/_auth-middleware.ts
@@ -2,19 +2,14 @@ import type { NextApiRequest } from "next";
 
 import { HttpError } from "@calcom/lib/http-error";
 
-import { schemaQueryIdAsString } from "~/lib/validations/shared/queryIdString";
 import { schemaQueryIdParseInt } from "~/lib/validations/shared/queryIdTransformParseInt";
 
 async function authMiddleware(req: NextApiRequest) {
   const { userId, isAdmin, prisma } = req;
-  const idParse = schemaQueryIdParseInt.safeParse(req.query);
-  const { id: slug } = schemaQueryIdAsString.parse(req.query);
-
-  const id = idParse.success ? idParse.data.id : undefined;
-
+  const { id } = schemaQueryIdParseInt.parse(req.query);
   if (isAdmin) return;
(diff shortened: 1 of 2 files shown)


emrysal (dismissed):
Thank you, excellent!

emrysal (dismissed):

apps/api/pages/api/event-types/[id]/_auth-middleware.ts
Comment (hariombalhara) [inline apps/api/pages/api/event-types/[id]/_auth-middleware.ts:1]: We should rename the path to [idOrSlug]

apps/api/pages/api/event-types/[id]/_auth-middleware.ts
```diff
@@ -2,14 +2,19 @@ import type { NextApiRequest } from "next";
...
-  const { id } = schemaQueryIdParseInt.parse(req.query);
+  const idParse = schemaQueryIdParseInt.safeParse(req.query);
+  const { id: slug } = schemaQueryIdAsString.parse(req.query);
+
+  const id = idParse.success ? idParse.data.id : undefined;
+
   if (isAdmin) return;
   const eventType = await prisma.eventType.findFirst({
-    where: { id, users: { some: { id: userId } } },
+    where: { OR: [{ id }, { slug }], AND: { users: { some: { id: userId } } } },
```
Comment (hariombalhara) [inline apps/api/pages/api/event-types/[id]/_auth-middleware.ts:17]: Because it would either be id or slug and we want to be clear if we are fetching by slug or id so using OR doesn't make sense.

Also, there could possibly be a scenario where an event-type has a numeric slug(say 123). In this case querying for event-types/123 could potentially fetch the eventType with that numeric slug. So, we need to be able to distinguish it clearly in the API request as to what the user wants.

Maybe we could go for /event-types?slug=something or a new endpoint maybe /event-types/bySlug/[slug] ?

hariombalhara (changes requested):
I think we should go for a change in logic a bit.

keithwillcode:
Closing as this PR now has 0 changes

PR #12493, opened 2023-11-22
fix: Cookie Not Expiring After Logout

### LINKED ISSUE #12483: Issue Report - Cookie Not Expiring After Logout
Domain: [https://app.cal.com/](https://app.cal.com/)

**Overview:**
I am writing to bring attention to a concerning issue regarding the expiration of cookies after a user logs out from our platform. It has come to our notice that despite logging out, the session cookie associated with the user's account does not expire as expected.

**Issue Description:**
Upon thorough testing and investigation, we have observed that the session cookie persists even after the user initiates a logout. This behavior poses a potential security risk as it allows unauthorized access to the user's account in the absence of cookie expiration after logout.

**Steps to Reproduce:**
1. Log in to the platform using valid credentials.
2. Perform various actions within the platform to establish an active session.
3. Copy the Cookie in Cookie-editor extension.
5. Click on the logout button to log out of the account.
6. Paste the Cookie in Cookie-editor extension.
7. Observe that the session cookie remains valid and active.

**POC: Video Attached**

**Expected Behavior:**
After the user logs out, the session cookie should be invalidated or set to expire, ensuring that subsequent attempts to access the platform using the same cookie are unsuccessful.

**Impact:**
Security Risk: The persistence of session cookies post-logout exposes users to potential unauthorized access.
User Experience: Users might assume that logging out terminates their session, leading to a misunderstanding of the platform's security measures.

Thank you.

Sincerely,

Mayank Kumar
mayankbgp2004@gmail.com

https://github.com/calcom/cal.com/assets/108946605/e5accbea-ec20-4b70-9a12-d305ce56df29

mayankthinks:
### fix: Cookie Not Expiring After Logout #12483

Fixes #12483

**What does this PR do?**
In this updated code:

1. I replaced the manual cookie deletion with the destroy method provided by NextAuth.
2. The destroy method will clear the session and remove the session cookie.
3. I also made sure to call destroy only when the session is authenticated (status === "authenticated").
4. I updated the import for the useRouter hook to be next/router instead of next/navigation as the latter is not a valid import in recent Next.js versions.

### DIFF
--- a/apps/web/pages/auth/logout.tsx
+++ b/apps/web/pages/auth/logout.tsx
@@ -1,6 +1,6 @@
 import type { GetServerSidePropsContext } from "next";
-import { signOut, useSession } from "next-auth/react";
-import { useRouter } from "next/navigation";
+import { destroy, useSession } from "next-auth/react";
+import { useRouter } from "next/router";
 import { useEffect, useState } from "react";
 
 import { WEBSITE_URL } from "@calcom/lib/constants";
@@ -20,14 +20,19 @@ type Props = inferSSRProps<typeof getServerSideProps>;
 export function Logout(props: Props) {
   const [btnLoading, setBtnLoading] = useState<boolean>(false);
   const { status } = useSession();
-  if (status === "authenticated") signOut({ redirect: false });
   const router = useRouter();
+
   useEffect(() => {
+    if (status === "authenticated") {
+      // Destroy the session and remove the session cookie
+      destroy();
+    }
+
     if (props.query?.survey === "true") {
       router.push(`${WEBSITE_URL}/cancellation`);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.query?.survey]);
+  }, [props.query?.survey, status]);
+
   const { t } = useLocale();
 
   const message = () => {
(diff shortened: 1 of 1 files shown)


PiyushChandra17 (commented):
Looks good 👍

zomars:
> I updated the import for the useRouter hook to be next/router instead of next/navigation as the latter is not a valid import in recent Next.js versions.

Can you post the docs about this issue? We're deprecating importing from "next/router" as we're planning to migrate to the app router soon.

apps/web/pages/auth/logout.tsx
```diff
@@ -1,6 +1,6 @@
 import type { GetServerSidePropsContext } from "next";
-import { signOut, useSession } from "next-auth/react";
-import { useRouter } from "next/navigation";
+import { destroy, useSession } from "next-auth/react";
```
zomars, line 2: As with our current version we get this error:

Module '"next-auth/react"' has no exported member 'destroy'.

apps/web/pages/auth/logout.tsx
```diff
@@ -1,6 +1,6 @@
...
+import { useRouter } from "next/router";
```
zomars, line 3: We're deprecating importing from "next/router" as we're planning to migrate to app router and it is incompatible.

zomars (changes requested):
Left some comments. The PR is not achieving the intended result. Also checks seems to be failing.

PR #12501, opened 2023-11-22
refactor: Team Creation Flow [CAL-2751]

joeauyeung:
## What does this PR do?

This PR changes the team creation flow. Currently you would have the option to create a team and then go through the Stripe checkout flow to publish the team.

This PR changes the flow so now, to create a team you need an active subscription before continuing.

Fixes # (issue)

https://www.loom.com/share/47e1a588a85d4e6fa47a989dee55282b

## Requirement/Documentation

## Type of change

## How should this be tested?

- Before checking out the branch, create a team but do not publish
- Now checkout the branch this PR is on
- Fill in the first step of creating a team
- You should be brought to the checkout page. Complete this step
- After you should be brought back to the add members page with the team now created in the DB

## Mandatory Tasks

## Checklist


apps/web/pages/api/teams/create.ts
joeauyeung, line 1: Since we're now creating a team after the checkout session. I think this should belong in it's own endpoint.

apps/web/pages/api/teams/create.ts
```diff
@@ -0,0 +1,90 @@
...
+        message: "Can't publish team/org without userId",
+      });
+    }
+  }
+
+  const checkoutSessionMetadata = parseCheckoutSessionMetadata.data ?? {
+    teamName: checkoutSession?.metadata?.teamName ?? generateRandomString(),
+    teamSlug: checkoutSession?.metadata?.teamSlug ?? generateRandomString(),
+    userId: checkoutSession.metadata.userId,
+  };
```
joeauyeung, line 60: Very unlikely that the metadata from creating the checkout session won't be carried over but I wanted to think about this edge case. Do we think this is worth covering?

apps/web/pages/api/teams/create.ts
```diff
@@ -0,0 +1,90 @@
...
+      parseCheckoutSessionMetadata.error,
+      checkoutSession.id
+    );
+
+    if (!checkoutSession.metadata.userId) {
+      throw new HttpError({
+        statusCode: 400,
+        message: "Can't publish team/org without userId",
+      });
+    }
```
joeauyeung, line 53: Absolutely need the userId of the owner to create the first member.

packages/features/ee/teams/components/AddNewTeamMembers.tsx
```diff
@@ -175,13 +175,9 @@ export const AddNewTeamMembersForm = ({
         className="w-full justify-center"
         disabled={publishTeamMutation.isLoading}
         onClick={() => {
-          if (orgBranding) {
-            router.push("/settings/teams");
-          } else {
-            publishTeamMutation.mutate({ teamId });
-          }
+          router.push(`/settings/teams/${teamId}/profile`);
```
joeauyeung, line 178: When finishing on this page, just navigate to the new team's profile page.

packages/features/ee/teams/components/CreateANewTeamForm.tsx
```diff
@@ -128,35 +129,14 @@ export const CreateANewTeamForm = () => {
...
-                      handleAvatarChange={(newAvatar: string) => {
-                        newTeamFormMethods.setValue("logo", newAvatar);
-                        createTeamMutation.reset();
-                      }}
-                      imageSrc={value}
-                    />
-                  </div>
-                </div>
-              </>
-            )}
```
joeauyeung, line 159: Since users can leave this part of the form without the data being saved, uploading an avatar and losing it might be too much friction. Let's leave this part out until the user has paid for the team and is directed to the profile page.

packages/features/ee/teams/lib/payments.ts
```diff
@@ -29,6 +29,51 @@ export const checkIfTeamPaymentRequired = async ({ teamId = -1 }) => {
   return { url: `${WEBAPP_URL}/api/teams/${teamId}/upgrade?session_id=${metadata.paymentId}` };
 };
 
+/**
+ * Used to generate a checkout session when trying to create a team
+ */
+export const generateTeamCheckoutSession = async ({
```
joeauyeung, line 35: Created a new function to generate a checkout session for this scenario. I want to keep the previous function that was used `purchaseTeamSubscription` for backwards compatibility for unpublished teams and orgs.

packages/trpc/server/routers/viewer/teams/create.handler.ts
```diff
@@ -45,6 +72,17 @@ export const createHandler = async ({ ctx, input }: CreateOptions) => {
...
+  // If the user is not a part of an org, then make them pay before creating the team
+  if (!isOrgChildTeam) {
+    const checkoutSession = await generateCheckoutSession({
+      teamSlug: slug,
+      teamName: name,
+      userId: user.id,
+    });
+
+    return checkoutSession;
+  }
```
joeauyeung, line 84: If this team is not created under an org, then direct user to checkout session.

packages/features/ee/teams/components/AddNewTeamMembers.tsx
```diff
@@ -33,11 +34,21 @@ type FormValues = {
...
     { teamId },
     { enabled: session.status === "authenticated" }
   );
+
+  useEffect(() => {
+    const { event } = router.query;
+    if (event === "team_created") {
+      telemetry.event(telemetryEventTypes.team_created);
+    }
+  }, []);
```
joeauyeung, line 50: When coming to this page after the checkout session, we should trigger the created team event.

packages/features/ee/teams/components/CreateANewTeamForm.tsx
```diff
@@ -42,8 +43,13 @@ export const CreateANewTeamForm = () => {
 
   const createTeamMutation = trpc.viewer.teams.create.useMutation({
     onSuccess: (data) => {
-      telemetry.event(telemetryEventTypes.team_created);
-      router.push(`/settings/teams/${data.id}/onboard-members`);
+      telemetry.event(telemetryEventTypes.team_checkout_session_created);
```
joeauyeung, line 46: Since a team isn't really being created at this stage let's create a new event. That way, we can see if people are abandoning the process.

joeauyeung (commented):
Changes to telemetry

packages/features/ee/teams/components/AddNewTeamMembers.tsx
```diff
@@ -33,11 +34,21 @@ type FormValues = {
...
   const session = useSession();
+  const telemetry = useTelemetry();
+
   const teamId = searchParams?.get("id") ? Number(searchParams.get("id")) : -1;
   const teamQuery = trpc.viewer.teams.get.useQuery(
     { teamId },
     { enabled: session.status === "authenticated" }
   );
+
+  useEffect(() => {
```
zomars, line 45: Can we track directly from server side? I don't think this should be a client side event.

apps/web/pages/api/teams/create.ts
```diff
@@ -0,0 +1,92 @@
...
+        subscriptionItemId: subscription.items.data[0].id || null,
+      },
+    },
+  });
+
+  // Sync Services: Close.com
+  // closeComUpdateTeam(prevTeam, team);
+
+  // redirect to team screen
+  res.redirect(302, `/settings/teams/${team.id}/onboard-members?event=team_created`);
```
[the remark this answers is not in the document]
zomars: Should this be a trpc mutation?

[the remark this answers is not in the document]
zomars: Nevermind.  It's good for now

packages/features/ee/teams/components/AddNewTeamMembers.tsx
zomars: Update: we can but we need to upgrade next-collect. Will do in a follow up.

zomars (changes requested):
We need to update tests to accommodate to the new team creating flow 
image

joeauyeung:
> We need to update tests to accommodate to the new team creating flow image

Looking at this test. Much of it is commented out and ends at the team creation process. Since that part is already tested, can we skip this test until it's fully flushed out?

apps/web/pages/api/teams/create.ts
zomars: I'd rather let it fail than generating a random string.

zomars:
> Looking at this test. Much of it is commented out and ends at the team creation process. Since that part is already tested, can we skip this test until it's fully flushed out?

If it was passing before, and it isn't now. Then we must fix it or update the test. I'll try to get this fixed and merged ASAP due it's urgency to mitigate recent abuse.

apps/web/pages/api/teams/create.ts
joeauyeung: I was thinking about this as well. At this point the team is already paid for. The user is redirected back to the profile page, where they can see if they need to change the team name.

If we fail, then we would have to make another call to Stripe to cancel the subscription. I think that would be more friction than making the user change the team name.

zomars (dismissed):
New team creation flow broke many e2e tests. Made some fixes to prevent flakyness.

Udit-takkar (commented):
@zomars some tests are still failing

zomars:
> @zomars some tests are still failing

One more to go. If no one beats me to it I'll take another stab later today

zomars (approved):

PR #12508, opened 2023-11-23
feat: add booking date range filters and freeze GPT-4 version

### LINKED ISSUE #11957: Issues with bad data -> gaps in the consumer api exposed by cal.ai

### Issue Summary

This issue is an aggregate of several reports of cal.ai not being able to fetch bookings and in rare cases, event types.

This is likely because those accounts have at least one "bugged" meeting in the past. The way the GET /bookings consumer api works - is it returns ALL meetings ever - possibly paginated? But at the very least - a ton of meetings from the past.

If there is a single booking with malformed data -> it can error. Not sure but it's possible it times out if you have booked a ton of meetings.

Note: This is likely not a bug with cal.ai - but since cal.ai is many users' first interaction with the consumer api - it is exposing historically bad data in the DB / gaps in the api that do not get flagged when just using the frontend.

Note 2: This issue seems to disproportionately affect power users, early adopters and cal team, investors, etc. Since those are the accounts that have tested beta features or been around long enough to have experienced migrations.

We cannot reproduce this naturally with new accounts.

### Steps to Reproduce

*TO TEST IF YOU HAVE BUGGED BOOKINGS*

1. [Create an API KEY here](https://app.cal.com/settings/developer/api-keys)
2. Navigate to https://api.cal.com/bookings?apiKey=[YOUR_API_KEY]
4. If the page does not load or returns an error, you have at least 1 bad historical meeting or (possibly) too many meetings

*TO CREATE A BUGGED BOOKING FOR TESTING*

- On Local, POST to the /bookings route with a number in the userId section of responses (supposed to be a string). This api call won't fail but should since you are supposed to pass userId as a string.
Screenshot 2023-10-17 at 4 30 22 PM

- Watch as the data is inserted into the db improperly in prisma studio with `yarn db-studio`
Screenshot 2023-10-17 at 4 30 31 PM

*UPDATE* This method no longer works to break getBookings locally, so I'm not sure exactly how to reproduce breaking getBookings locally anymore.

This video is additional proof of a bugged booking in prod:
https://github.com/calcom/cal.com/assets/36115192/9c1951c8-05a7-4a12-87cd-6f9df925bc77

@PeerRich, @DexterStorey, @tedspare and others have this bug in prod.

### Actual Results

In Cal.ai:
Screenshot 2023-10-17 at 3 35 39 PM

With API directly:

`{"message":"An error occured while querying the database."}`

OR

Screenshot 2023-10-17 at 3 51 32 PM

### Expected Results

- You have 2 meetings today...

### Ideas

- It seems too many users are effected to manually clean the data - unless a util is built.
- Loosening zod types on the consumer api could work but isn't super clean (although the zod types should be looked at since you can currently pass bad data to createBookings without error / validation)
- Adding "dateFrom and dateTo" to the getBookings api could be the way. [We already filter the response in cal.ai](https://github.com/calcom/cal.com/blob/59fa713549e38aa92161499e171b74cceb9f578b/apps/ai/src/tools/getBookings.ts#L42) (because if we gave the AI all the meetings it would be too much text). So we could just implement something similar in the consumer API (that way past meetings wouldn't be fetched 99% of the time)

(just some ideas! - let's brainstorm!)

### Evidence

- Tested with cal.ai "What meetings do I have today" and using the consumer api at api.cal.com

DexterStorey:
## What does this PR do?

This PR adds optional filtering by date to the Cal.com consumer API, and integration with cal.ai.

It fixes a problem where many accounts can't access the getBookings route since they have historical buggy data - and the consumer API previously fetched all booking ever, leading to errors.

We now add the option to fetch booking within a range.

Partially Fixes #11957 (issue)

Additionally, this pr freezes the GPT-4 model (gpt-4-0613) version in an attempt to prevent model degradation.

## Type of change

## Mandatory Tasks

## Checklist

- ~I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)~
- ~My code doesn't follow the style guidelines of this project~


DexterStorey:
Tested and fixed!
This seems to work now.

When I go to: `http://localhost:3002/bookings?dateFrom=2023-12-20T16:00:00Z&dateTo=2023-12-20T16:15:00Z&apiKey=API_KEY`

I can see the booking:

Screenshot 2023-12-19 at 10 33 47 PM

But with: `http://localhost:3002/bookings?dateFrom=2023-12-20T16:00:00Z&dateTo=2023-12-20T16:14:00Z&apiKey=API_KEY`

It is hidden

Screenshot 2023-12-19 at 10 34 57 PM

With no `dateFrom` or `dateTo` passed, the api query responds normally, so in theory this is purely additive / optional.

`http://localhost:3002/bookings?apiKey=API_KEY`

Screenshot 2023-12-19 at 10 36 14 PM

DexterStorey:
LGTM 🚀
This should fix #11957
Docs should probably be updated as we now have a new feature for the getBookings API (`dateFrom`, `dateTo`) which filters bookings by date.

apps/api/pages/api/bookings/_get.ts
```diff
@@ -203,6 +207,25 @@ async function handler(req: NextApiRequest) {
...
+    args.where = {
+      ...args.where,
+      startTime: { gte: dateFrom },
+    };
+  } else if (dateTo) {
+    args.where = {
+      ...args.where,
+      endTime: { lte: dateTo },
+    };
+  }
```
alishaz-polymath, line 227: Could probably do something like this to keep it clean, more readable and remove the need for extra `else-ifs` , thoughts @emrysal 🤔?
```suggestion
  if (dateFrom) {
    args.where.startTime = { gte: dateFrom };
  }
  if (dateTo) {
    args.where.endTime = { lte: dateTo };
  }
```

alishaz-polymath (changes requested):
Temporal filters are quite handy, thank you for the PR 🙏 
Left a change request which I think should improve code quality

apps/api/pages/api/bookings/_get.ts
DexterStorey: 🔥

apps/api/pages/api/bookings/_get.ts
```diff
@@ -208,18 +208,13 @@ async function handler(req: NextApiRequest) {
...
     args.where = {
       ...args.where,
       startTime: { gte: dateFrom },
     };
-  } else if (dateTo) {
+  }
+  if (dateTo) {
     args.where = {
       ...args.where,
       endTime: { lte: dateTo },
```
DexterStorey, line 220: @alishaz-polymath I was able to simplify a bit but I can't directly set `args.where.startTime` or `args.where.endTime` because `args.where` is possibly undefined - so we need to create `args.where` if it doesn't exit. This feels pretty clean to me but LMK!

apps/api/pages/api/bookings/_get.ts
```diff
@@ -162,6 +162,10 @@ function buildWhereClause(
 
 async function handler(req: NextApiRequest) {
   const { userId, isAdmin, prisma } = req;
+
+  const dateFrom = req.query.dateFrom ? new Date(req.query.dateFrom as string) : undefined;
+  const dateTo = req.query.dateTo ? new Date(req.query.dateTo as string) : undefined;
+
```
alishaz-polymath, line 168: I wonder if we can do something like this:
```suggestion
const args: Prisma.BookingFindManyArgs = {
  include: {
    attendees: true,
    user: true,
    payment: true,
  },
  where: {}, // Initialize 'where' here
};
```
from current L169 to L174
That would initialize where inside args and then we could have 
```
if (dateFrom) {
  args.where.startTime = { gte: dateFrom };
}
if (dateTo) {
  args.where.endTime = { lte: dateTo };
}
```
later on without the issue you mention

alishaz-polymath (changes requested):
Just a NIT, although your changes should are already better, this would just be much cleaner and easier to read. 🙌

apps/api/pages/api/bookings/_get.ts
DexterStorey: Hey, thx for the feedback the support here is killer!

Unfortunately, even when initializing args with `where`, I still get a type issue: `'args.where' is possibly 'undefined'.`

Screenshot 2023-12-20 at 5 46 11 PM

This is because the `BookingFindManyArgs` type sets `where` as optional and args is cast as `BookingFindManyArgs`.

Screenshot 2023-12-20 at 5 46 38 PM

There are two ways around this: 
- update the type
- re-cast the type

However, I want this PR to be as minimal as possible as I don't have context on any downstream implications or any new potential errors introduced by an empty `where` object `{}`.

For these reasons, I recommend sticking with the proposed implementation which sets and updates `where` non-confrontationally and only when necessary.

```
if (dateFrom) {
    args.where = {
      ...args.where,
      startTime: { gte: dateFrom },
    };
  }
```

@alishaz-polymath → Do you think we can look into approving / merging this as is?

exception (dismissed):

apps/api/pages/api/bookings/_get.ts
alishaz-polymath: Sure 👍 
I think we can optimize the code as a follow up. 🙌 Thanks for the PR 🎉

alishaz-polymath (dismissed):
LGTM

apps/api/pages/api/bookings/_get.ts
```diff
@@ -162,6 +162,10 @@ function buildWhereClause(
 
 async function handler(req: NextApiRequest) {
   const { userId, isAdmin, prisma } = req;
+
+  const dateFrom = req.query.dateFrom ? new Date(req.query.dateFrom as string) : undefined;
+  const dateTo = req.query.dateTo ? new Date(req.query.dateTo as string) : undefined;
```
alishaz-polymath, line 167: Apologies for missing this earlier, I think we need to put these in try catch and return a viable error message instead of the generic 
```
{
    "message": "An error occured while querying the database."
}
```
that you'd get right now if you pass invalid dateFrom and/or dateTo. Wrapping them up in a try catch, or better yet, pull out a reusable function that validates while parsing these inputs would be awesome 🙌

alishaz-polymath (changes requested):

apps/api/pages/api/bookings/_get.ts
alishaz-polymath: Once these changes are in, it should be good to go 🙏

alishaz-polymath: Hey @DexterStorey did you manage to get a chance to have a look ^?

apps/api/lib/validations/booking.ts
```diff
@@ -18,6 +18,11 @@ const schemaBookingBaseBodyParams = Booking.pick({
 
 export const schemaBookingCreateBodyParams = extendedBookingCreateBody.merge(schemaQueryUserId.partial());
 
+export const schemaBookingGetParams = z.object({
+  dateFrom: iso8601.optional(),
+  dateTo: iso8601.optional(),
+});
+
```
DexterStorey, line 25: support schema parsing for the bookings get api

apps/api/pages/api/bookings/_get.ts
```diff
@@ -162,6 +162,9 @@ function buildWhereClause(
 
 async function handler(req: NextApiRequest) {
   const { userId, isAdmin, prisma } = req;
+
+  const { dateFrom, dateTo } = schemaBookingGetParams.parse(req.query);
```
DexterStorey, line 166: @alishaz-polymath can you take a look at this?

DexterStorey: Screenshot 2024-01-08 at 3 35 00 PM

apps/ai/src/utils/agent.ts
```diff
@@ -13,7 +13,7 @@ import type { User, UserList } from "../types/user";
 import type { WorkingHours } from "../types/workingHours";
 import now from "./now";
 
-const gptModel = "gpt-4";
+const gptModel = "gpt-4-0613";
```
alishaz-polymath, line 16: Is there any specific reason why we picked this Model exaclty? Additional context would be really helpful

alishaz-polymath (approved):
Looks good, thank you @DexterStorey 🚀

Issue #12503, opened 2023-11-23
Internal Server Error (500) on visiting cal.com/getting-started


Aamir-Dev-6600:
### Issue Summary

Created a new account and was trying to setup the information, I was automatically routed to the following mentioned route. Now every time when I go to cal.com and open my dashboard, it shows the 'Black Screen' on getting started page, and the shows the 500 error (screenshot attached).

### Steps to Reproduce

1. Create a new account on cal.com
2. It will route you to the getting-started page, if not do 3
3. Simply go to Cal.com/getting-started (it shows 500 internal server error)

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

### Evidence
![image](https://github.com/calcom/cal.com/assets/132020600/5cd47929-3a92-4849-8b79-b1009a5784c8)

Aamir-Dev-6600:
UPDATE

If I'm already logged in and load the cal.com, on the top right corner I see "Go To App" button, a click leads me to cal.com/getting-started page with black interface. 

When I reload cal.com/getting-started, it shows the 500 error.

PR #12510, opened 2023-11-23
fix: set nextjs version to 13.4.6

dmytro068:
## What does this PR do?
Sets the nextjs version to 13.4.6 (strictly) so it cannot be implicitly changed when other packages (ai package) have higher version of nextjs.

Fixes # (issue)
This PR will resolve the [issue](https://github.com/calcom/cal.com/issues/12503) with crash on `getting-started` page that was caused by new `useParams` behavior in next 13.5.4. 
In next 13.4.6 useParams always returns null and in next 13.5.4 it will return an object that has `undefined` value for optiona catch-all route. 
 

## Type of change

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -33,7 +33,7 @@
     "@sentry/nextjs": "^7.73.0",
     "bcryptjs": "^2.4.3",
     "memory-cache": "^0.2.0",
-    "next": "^13.4.6",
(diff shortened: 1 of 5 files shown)


PeerRich (approved):

PeerRich (approved):

zomars:
Closing as we're planning to upgrade soon

PR #12520, opened 2023-11-24
fix:  MEETING_ENDED webhook has incorrect content type

rahul3v:
/claim #12470


CarinaWolli:
already fixed with https://github.com/calcom/cal.com/pull/12484

PR #12530, opened 2023-11-24
fix: OAuth URL account for dev environment

joeauyeung:
## What does this PR do?

This PR allows the `WEBAPP_URL_FOR_OAUTH` to account for our dev environment to use the `WEBPP_URL` instead of defaulting back to localhost.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/lib/constants.ts
```diff
@@ -17,7 +18,7 @@ export const WEBAPP_URL =
 
 // OAuth needs to have HTTPS(which is not generally setup locally) and a valid tld(*.local isn't a valid tld)
 // So for development purpose, we would stick to localhost only
-export const WEBAPP_URL_FOR_OAUTH = IS_PRODUCTION ? WEBAPP_URL : "http://localhost:3000";
+export const WEBAPP_URL_FOR_OAUTH = IS_PRODUCTION || IS_DEV ? WEBAPP_URL : "http://localhost:3000";
```
joeauyeung, line 21: Confirmed in a replt that this will return true if either `IS_PRODUCTION` or `IS_DEV` is true.

hariombalhara (approved):

PR #12540, opened 2023-11-25
chore: removed last web3 references

### LINKED ISSUE #12534: remove web3 from codebase
remove any legacy web3 related stuff

(context: cal.com/blog/sunsetting-web3-app-rainbow)

PeerRich:
good bye my lover


packages/prisma/schema.prisma
```diff
@@ -660,7 +660,6 @@ enum AppCategories {
   other
   payment
   video // deprecated, please use 'conferencing' instead
-  web3 // deprecated, we should no longer have any web3 apps
```
Udit-takkar, line 663: Getting this build error: The values [web3] on the enum `AppCategories` will be removed. If these variants are still used in the database, this will fail.

Udit-takkar: one easy solution is just marking it as depreceated and not removing it from schema.prisma file.

PeerRich: done

Udit-takkar (approved):

Issue #12415, opened 2023-11-17
[CAL-2904] Request to add 2 Fields to Webhook Payloads


henrym9345:
### Is your proposal related to a problem?

I'm using the webhooks to trigger appointment reminders (internal and external) from our marketing automation tool rather than use built in reminders. 

Problem 1: I want to include cancel and reschedule links in these messages which require the url slug, which is not present in the webhook. The current workaround is to maintain a collection that contains the url slug and event type ID. This is called upon in the messaging workflow and a slug is obtained that matches the event type ID in the webhook. This is suboptimal though because this collection must be maintained as new events are added, slugs can be somewhat easily changed from cal.com and it just adds additional steps.

Problem 2:Timestamps for the appointments must be in the correct timezone on all messages. Since the payload only provides UTC time, a conversion must be performed[ using Liquid tags](https://customer.io/docs/journeys/liquid-tag-list/#timezone). This requires a UTC offset (ie -7), so the timezone string (ie America/Los Angeles) must be matched to provide this. This is done with a similar method as described in Problem 1. To make matters more complicated our local timezone of America/Phoenix does not observe daylight savings, so the offset will switch between -7 and -8. This adds the need for additional logic to check the current date before performing the timezone conversion. 

### Describe the solution you'd like
Solution for Problem 1: Add a field for url slug to the webhook payload

Solution for Problem 2: 
Option 1: Add the attendee UTC offset (ie -7) as a field, rather than the string of the timezone. 
Option 2: Add a timestamp field where the timestamp is already matches the timezone setting of the appointment.

(Describe your proposed solution here.)

### Describe alternatives you've considered

Possible alternative for Problem 1: Use url parameters+ a hidden field to pass through the slug. Not particularly stable though.

### Additional context

NA

### Requirement/Document

NA

---
##### House rules
- If this issue has a `🚨 needs approval` label, don't start coding yet. Wait until a core member approves feature request by removing this label, then you can start coding.
  - For clarity: Non-core member issues automatically get the `🚨 needs approval` label.
  - Your feature ideas are invaluable to us! However, they undergo review to ensure alignment with the product's direction.

<sub>[CAL-2904](https://linear.app/calcom/issue/CAL-2904/request-to-add-2-fields-to-webhook-payloads)</sub>

PeerRich:
i think this is reasonable. looking for people to help 🙏

TwistingTwists:
Would like to take this up. Is there a timeline for this ?

[...]

PR #12542, opened 2023-11-26
feat: added urlSlug and utcOffset to Webhook

mallela32:
## What does this PR do?
added URL slug and UTC offset in the payload 

/claim #12415


packages/features/webhooks/components/WebhookForm.tsx
```diff
@@ -16,15 +16,19 @@ export type TWebhook = RouterOutputs["viewer"]["webhook"]["list"][number];
 export type WebhookFormData = {
   id?: string;
   subscriberUrl: string;
+  urlSlug: string;
   active: boolean;
   eventTriggers: WebhookTriggerEvents[];
-  secret: string | null;
+  secret: string;
```
alishaz-polymath, line 22: Why are we doing this exactly? A webhook secret is an optional field. 🤔

packages/features/webhooks/components/WebhookForm.tsx
```diff
@@ -74,13 +78,16 @@ const WebhookForm = (props: {
         : props.webhook.eventTriggers,
       secret: props?.webhook?.secret || "",
       payloadTemplate: props?.webhook?.payloadTemplate || undefined,
+      utcOffset: props?.webhook?.utcOffset || "-7",
```
alishaz-polymath, line 81: Any particular reason for a fallback to "-7" here?

alishaz-polymath (changes requested):
We already have "type" in the payload which is the same as slug, so we **do not** need a urlSlug field. 
As for the utcOffset field, we only need to add it next to the attendee timezone string.

PR #12543, opened 2023-11-26
fix: remove duplicate key in common.json

Udit-takkar:


PeerRich (approved):
always good to have some i18n hygene, thanks!

PR #12547, opened 2023-11-27
fix: seat enabled booking and rescheduling

### LINKED ISSUE #12188: [CAL-2906] Event with seating: The custom question answers of the first user to book are unintentionally exposed to subsequent attendees when rescheduling or canceling.
### Issue Summary

When you host an event that provides seats and includes custom booking questions, the answers of the first user to book are inadvertently revealed to subsequent attendees of the event when they attempt to reschedule or cancel.

### Steps to Reproduce

1. Create an event that offers seats, uncheck both checkboxes 'Share attendee information between guests' and 'Show the number of available seats.'
2. Add custom questions like phone number, zip code, and so on.
3. Register for this event.
4. Register again for the same event on the same day and time, but provide different answers and use a different email.
5. Open the second registrant's email inbox for the confirmation email and click on the button to cancel or reschedule the meeting.
6. Notice that the first person to register for this event appears in the field instead of the user's own questions.

### Actual Results

* The answers of the first user to book are inadvertently revealed to subsequent attendees of the event when they attempt to reschedule or cancel.

### Expected Results

* To display the answers of the correct attendee instead of the first attendee, or have them appear blank instead.

### Technical details

* Production env.

### Evidence

image

Screenshot 2023-11-01 at 17 02 33

Screenshot 2023-11-01 at 17 07 53

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2906](https://linear.app/calcom/issue/CAL-2906/event-with-seating-the-custom-question-answers-of-the-first-user-to)</sub>

TaduJR:
## What does this PR do?

This change Fixes #12188

## Requirement/Documentation

## Type of change

## How should this be tested?

As mentioned on the issue booking and rescheduling does not add new booking after one user booked at the same time and date

## Mandatory Tasks


CarinaWolli (changes requested):
Tested locally and issue is not fixed. 

What I did to test: 
1. Create seated event with custom questions 
2. Book seated event with a non-user email address
3. Book same slot with another non-user email address 
    - Booking success page showed answers to custom questions from first user 
4. Reschedule second booking
    - Custom questions were prefilled with the ones from the first user

packages/core/getUserAvailability.ts
```diff
@@ -99,15 +99,41 @@ const getUser = (where: Prisma.UserWhereInput) =>
 
 type User = Awaited<ReturnType<typeof getUser>>;
 
-export const getCurrentSeats = (eventTypeId: number, dateFrom: Dayjs, dateTo: Dayjs) =>
-  prisma.booking.findMany({
+const getUserBookings = async (userId: number) => {
```
CarinaWolli, line 102: Why do we need these changes in getUserAvailability? 

I saw an issue if I booked a seat, it kept showing that all seats area available for this slot. Might be because of these changes here

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -591,10 +592,11 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
             currentSeats[
               currentSeats.findIndex((booking) => booking.startTime.toISOString() === time.toISOString())
             ]._count.attendees,
-          bookingUid:
-            currentSeats[
-              currentSeats.findIndex((booking) => booking.startTime.toISOString() === time.toISOString())
-            ].uid,
+          bookingUid: input?.bookerUserId
```
CarinaWolli, line 595: Why do we need that?

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1120,6 +1121,36 @@ async function handler(
     users = [...availableUsers.filter((user) => user.isFixed), ...luckyUsers];
   }
 
+  const attachNonUserBookUid = async (bookerEmail: string, dateFrom: Dayjs, dateTo: Dayjs) => {
```
CarinaWolli, line 1124: Why is that needed? Shouldn't `reqBody.bookingUid` be already set

CarinaWolli (changes requested):
Please make sure to only make changes where needed. All we want is to show the correct answers of custom questions. It also shouldn't matter if the attendee is a registered user or not. 

Can you please add some comments to explain your changes 🙏

TaduJR:
@CarinaWolli The bug is a lot deeper than it looks. Let me explain if you booked a seat enabled event and your are trying to book again on the same time you are added to the attendee list only. The previous code implements this without getting the user email and checking if it was booked as the same time. As a result, it will give access to the first user always. This will also give edit access of the first user custom questions during rescheduling for all other bookers at the same time. In addition, the bookers will not be add to the booking table as a result of the leaked bookingUid for the wrong user. This all because of not attaching the correct bookingUid the the logged in user. So if user booked at 12:00 and trying to book again. Then it will attach bookingUid to add the person in attendee table only. Check handleNewBooking line 2052 for more. If it was not the pervious user who booked then it must not attach bookingUid. But in the main code this is not happening.

TaduJR:
> Please make sure to only make changes where needed. All we want is to show the correct answers of custom questions. It also shouldn't matter if the attendee is a registered user or not.
> 
> Can you please add some comments to explain your changes 🙏

@CarinaWolli The bug is a lot deeper than it looks. Let me explain if you booked a seat enabled event and your are trying to book again on the same time you are added to the attendee list only. The previous code implements this without getting the user email and checking if it was booked as the same time. As a result, it will give access to the first user always. This will also give edit access of the first user custom questions during rescheduling for all other bookers at the same time. In addition, the bookers will not be add to the booking table as a result of the leaked bookingUid for the wrong user. This all because of not attaching the correct bookingUid the the logged in user. So if user booked at 12:00 and trying to book again. Then it will attach bookingUid to add the person in attendee table only. Check handleNewBooking line 2052 for more. If it was not the pervious user who booked then it must not attach bookingUid. But in the main code this is not happening.

packages/core/getUserAvailability.ts
TaduJR: During getAvailableSlot function call getUserAvailablity is also called which it self calls the getCurrentSeat function. The getCurrentSeat function is responsible for fetching the reserved seats. getUserAvailability function is resposible for fetching all the free time and free seats also reserved seat which it got from getCurrentSeat. So the change in getUserAvailability in short is responsible for fetching the correct bookingUid for logged in or registering booker which is used for rescheduling and editing user info.

packages/trpc/server/routers/viewer/slots/util.ts
TaduJR: This is needed if a logged in user exists then it will return bookingUid for the same date time reservation. Non logged in user is another case which i will expain.

packages/features/bookings/lib/handleNewBooking.ts
TaduJR: attachNonUserBookUid function is responsible for attaching bookingUid if non user booker is trying to book on the same time slot.

packages/core/getUserAvailability.ts
TaduJR: Also the issue you got i am not able to reproduce it. I can upload loom vid if you would like. 😊

CarinaWolli:
I see several issues with this PR, so I will close it. Feel free to open a new one if you can address the issues:

- Your changes in `getUserAvailability` break the display of reserved seats in a slot (they no longer show correctly)
- The logged-in user should not be relevant. A user can be logged in and still book a seat for another email address. Also, seats can be booked by users without an account, so all we need to identify the seat is the email address
- With your changes, for every seat a new booking is now created. We don't want that and should instead make use of `seatReferences`
- It appears that the main issue is that we don't save custom answers for additional seats
   - We should store these answers in the `data` field of `bookingSeat`, where we are already saving additional notes. This should also enable us to display the correct custom answers upon rescheduling.

MehulZR:
Hey @CarinaWolli, can you take a lot at https://github.com/calcom/cal.com/issues/12188#issuecomment-1809405538. If you approve that I can start working on it.

PR #12525, opened 2023-11-24
chore: sync platform branch with main

supalarry:
It was decided to kickstart v2 API on the newly created `platform` branch.

This PR adds an action that will sync `platform` branch with `main` each time code is merged in the `main`.

PR #12548, opened 2023-11-27
revert: sync platform branch with main

supalarry:
After 2 attempts to sync platform branch with main in #12525 and #12526 there still was an error
Screenshot 2023-11-27 at 09 15 35

and it was running red in main. Since this is not high priority i will remove the workflow file since its causing trouble and focus on more important stuff.


Ryukemeister (approved):
LGTM, @supalarry let's manually merge the `platform` branch with `main` for the time being, won't be much of a hassle anyway!

PR #12554, opened 2023-11-27
feat: scaffold v2 API

supalarry:
## base nest project
As a base is used nest.js project created by @exception (Erik):
https://github.com/calcom/nestjs-poc

## prisma
To make prisma run, within the new `apps/api-v2`
1. Create `prisma` folder and copy over `schema.prisma` from our `packages/prisma` but remove all generators except `generator client`.
2. install `prisma` and `@client/prisma`
3. import `PrismaClient` from `@client/prisma` instead of `@calcom/prisma` which also allows us creating `customPrisma` but we will roll with vanilla `PrismaClient` for now.
4. Add `generate-schemas` to `package.json` to generate prisma client for `api-v2`.

## prettier
Setup `.prettierrc.js` file that extends same prettier config as the root `.prettierrc.js` with `typescript` and `decorators-legacy` plugins to work with Nest decorators.

## Kysely

Kysely: created a new stand-alone `@calcom/kysely` package generating types in `packages/kysely/types.ts` based on prisma schema. Notably, kysely types are generated based on `packages/prisma` schema not the one in `apps/api-v2/prisma`.

## turbo
Added entry for `api-v2` in `turbo.json` informing about the expected environment variables.

## ee
Created `apps/api-v2/src/ee` folder with empty LICENSE file

## lock file
commit lock file to have `@calcom/platform-api` (apps/api-v2) in lock file


PeerRich:
exciting!

exception (approved):

ThyMinimalDev:
I think we can ignore the preview branch failure for now, we will fix it before merging to main

PR #12568, opened 2023-11-28
Fix:use typed query initially fill default paramaters 

sean-brydon:
## What does this PR do?
Default query params not being filled when loading a page with useTypedQuery.

Comparing prod vs this PR
https://github.com/calcom/cal.com/assets/55134778/ac2e6624-ff07-49c3-acb6-5f5110cbabbd

## Type of change

## How should this be tested?

- Load event types settings on dev and notice the default tab is already selected

## Mandatory Tasks


packages/lib/hooks/useTypedQuery.ts
```diff
@@ -46,6 +46,17 @@ export function useTypedQuery<T extends z.AnyZodObject>(schema: T) {
     return {} as Output;
   }, []);
 
+  useEffect(() => {
+    if (parsedQuerySchema.success && parsedQuerySchema.data) {
+      Object.entries(parsedQuerySchema.data).forEach(([key, value]) => {
+        if (key in unparsedQuery || !value) return;
```
sean-brydon, line 52: If the value isnt in unparsedQuery (from router) or there is no value - skip it

packages/lib/hooks/useTypedQuery.ts
```diff
@@ -46,6 +46,17 @@ export function useTypedQuery<T extends z.AnyZodObject>(schema: T) {
...
+        const search = new URLSearchParams(parsedQuery);
+        search.set(String(key), String(value));
+        router.replace(`${pathname}?${search.toString()}`);
```
sean-brydon, line 55: Set the default parameter

supalarry (approved):
Tested locally all good! Also confirmed that if instead proper query param `?tabName=setup` we enter random `?a=b` that it gets replaced with default `?tabName=setup` and that event setup tab is selected.

PR #12569, opened 2023-11-28
fix: premium username to premium username

sean-brydon:
https://www.loom.com/share/5b4abf304027499f966822d3dd1e2215

Fixes premium username -> premium username changes

### DIFF
--- a/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
@@ -3,7 +3,7 @@ import type { GetServerSidePropsContext, NextApiResponse } from "next";
 import { v4 as uuidv4 } from "uuid";
 
 import stripe from "@calcom/app-store/stripepayment/lib/server";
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -110,7 +110,7 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
     // @TODO: iterate if stripeSubscriptions.hasMore is true
     const isPremiumUsernameSubscriptionActive = stripeSubscriptions.data.some(
       (subscription) =>
-        subscription.items.data[0].price.product === getPremiumPlanProductId() &&
+        subscription.items.data[0].price.id === getPremiumMonthlyPlanPriceId() &&
```
sean-brydon, line 113: cal.com/apps/website/pages/api/signup.ts

In signup we create the sub on the priceId and not the productId. 

So change this check to check for the right priceId

exception (approved):

PR #12570, opened 2023-11-28
feat: rate limit removeMember

sean-brydon:
adds sourceIP to ctx + rate limit remove member


packages/trpc/server/createContext.ts
```diff
@@ -12,6 +13,7 @@ import type { CreateNextContextOptions } from "@trpc/server/adapters/next";
 type CreateContextOptions = CreateNextContextOptions | GetServerSidePropsContext;
 
 export type CreateInnerContextOptions = {
+  sourceIp?: string;
```
sean-brydon, line 16: Maybe @zomars @keithwillcode can figure out a better way to have this not be optional 

It needs to be optional due to the fact CreateSsgHelpers needs the full context - however we have no way of getting the IP into that. they also use the infered types so we can't omit in one place and not the other.

packages/trpc/server/routers/viewer/teams/removeMember.handler.ts
```diff
@@ -14,11 +15,16 @@ type RemoveMemberOptions = {
   ctx: {
     user: NonNullable<TrpcSessionUser>;
     prisma: PrismaClient;
+    sourceIp?: string;
```
sean-brydon, line 18: This is always there - check getIP return signature. Type issue within context i cant seem to figure out

zomars (commented):
Are we experiencing abuse on this endpoint?

ThyMinimalDev (approved):
looking good, can't figure out the type issue, approving as is

PR #12580, opened 2023-11-28
fix: Toggling really fast throws internal errors

emrysal:
## What does this PR do?

Fixes UI issue that causes "An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist."


joeauyeung (approved):
LGTM

PR #12584, opened 2023-11-29
chore: add check for recording

Udit-takkar:
Screenshot 2023-11-29 at 2 01 54 PM

view recordings -> we have isRecorded property marked as true

in case user wasn't logged in while recording or any other failure which couldn't update the isRecorded to true. user is unable to get the recording link so added a button

check for recordings -> isRecorded is marked as false but you can check if any recording exist for this daily vide room.


PeerRich (approved):
makes sense, good edge case!

PR #12585, opened 2023-11-29
fix: signup nit

sean-brydon:
Thread: https://threads.com/thread/34549086570/34549086572?s=modBCstRmXYUHzZvYsQfrL

Disable form when no username or email is present 
Fix allignment in error message
add loading state to google button
CleanShot 2023-11-29 at 09 54 37


emrysal (dismissed):
lgtm

PeerRich (dismissed):

PeerRich (dismissed):

apps/web/pages/signup.tsx
```diff
@@ -360,13 +379,9 @@ export default function Signup({
                         router.push(`${GOOGLE_AUTH_URL}?${searchQueryParams.toString()}`);
                         return;
                       }
+                      setIsGoogleLoading(false);
```
ThyMinimalDev, line 382: I don't think it is necessary to change the loading back to false,  if changing page with router.push is taking time, the loading spinner will stop before we are redirected

ThyMinimalDev (approved):
Looking good !

PR #12592, opened 2023-11-29
fix: updating workflow with new step and new active event type

CarinaWolli:
## What does this PR do?

Fixes issue that a workflow with an added step and a new active event types couldn't be saved because `booking.findMany()` failed. 

## Type of change

## How should this be tested?

- Create Workflow with before event trigger 
- Add a step 
- add an event type in the dropdown 
- Test if saving the workflow is successful

### DIFF
--- a/packages/trpc/server/routers/viewer/workflows/update.handler.ts
+++ b/packages/trpc/server/routers/viewer/workflows/update.handler.ts
@@ -629,11 +629,10 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
   });
 
   if (addedSteps) {
-    const eventTypesToCreateReminders = activeOn.map((activeEventType) => {
-      if (activeEventType && !newEventTypes.includes(activeEventType)) {
-        return activeEventType;
-      }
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -629,11 +621,9 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
   });
 
   if (addedSteps) {
-    const eventTypesToCreateReminders = activeOn.map((activeEventType) => {
```
CarinaWolli, line 632: this resulted in having null and undefined values in `eventTypesToCreateReminders` which made `booking.findMany()` from [Line 644](https://github.com/calcom/cal.com/pull/12592/files#diff-e434eb7fdd98d2062613cf2c8f5a69bd969b932d5cd5b0ca329d434b118f6fa6R644) fail

Udit-takkar (approved):
LGTM. works as expected

PR #12603, opened 2023-11-30
chore: init platform packages

ThyMinimalDev:
## What does this PR do?

Init platform internal packages

### DIFF
--- a/packages/platform/atoms/index.ts
+++ b/packages/platform/atoms/index.ts
@@ -1,5 +1,3 @@
 export { Booker } from "./booker/Booker";
 export { CalProvider } from "./cal-provider/index";
 export { ConnectToCal } from "./connect-to-cal-button/index";
-export * from "./src/components/ui/button";
-export * from "./src/lib/utils";
--- a/packages/platform/atoms/index.ts
+++ b/packages/platform/atoms/index.ts
@@ -1,3 +1,5 @@
 export { Booker } from "./booker/Booker";
 export { CalProvider } from "./cal-provider/index";
 export { ConnectToCal } from "./connect-to-cal-button/index";
+export * from "./src/components/ui/button";
+export * from "./src/lib/utils";
--- a/packages/platform/atoms/package.json
+++ b/packages/platform/atoms/package.json
@@ -19,7 +19,7 @@
     "typescript": "^4.9.4",
     "vite": "^4.1.2"
   },
-  "main": "./index,ts",
+  "main": "./index.ts",
   "types": "./index.ts",
   "dependencies": {
     "@calcom/ui": "*",
--- a/packages/platform/atoms/package.json
+++ b/packages/platform/atoms/package.json
@@ -19,7 +19,8 @@
     "typescript": "^4.9.4",
     "vite": "^4.1.2"
   },
-  "main": "./index",
+  "main": "./index,ts",
+  "types": "./index.ts",
   "dependencies": {
     "@calcom/ui": "*",
     "@radix-ui/react-slot": "^1.0.2",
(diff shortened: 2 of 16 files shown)


packages/platform/atoms/index.ts
```diff
@@ -1,3 +1,5 @@
 export { Booker } from "./booker/Booker";
 export { CalProvider } from "./cal-provider/index";
 export { ConnectToCal } from "./connect-to-cal-button/index";
+export * from "./src/components/ui/button";
```
supalarry, line 4: Just brainstorming: Is there a specific reason to export components used to create cal atoms and lib/utils?  If we export more and more stuff then it might be harder to understand which of the exported things are actually "real atoms" aka file is named `atoms/index.ts` but we are pushing upwards UI components used to create atoms which is just shadcn. It could be said that maybe whoever uses platform API wants to use similar button to atoms but then we are creating UI library.

supalarry: Nit picking: Also, if we export stuff here probably cleaner would be having `index.ts` file in `src/components/ui` that exports all components in UI folder, so here we would have only:
```
export * from "./src/components/ui";
```

Ryukemeister: > Just brainstorming: Is there a specific reason to export components used to create cal atoms and lib/utils? If we export more and more stuff then it might be harder to understand which of the exported things are actually "real atoms" aka file is named `atoms/index.ts` but we are pushing upwards UI components used to create atoms which is just shadcn. It could be said that maybe whoever uses platform API wants to use similar button to atoms but then we are creating UI library.

Agreed with this point, all the UI components can easily be found from shadcn so wouldn't it better not to export the components from shadcn?

packages/platform/types/permissions.ts
```diff
@@ -0,0 +1,3 @@
+import type { PERMISSIONS } from "@calcom/platform-const";
+
+export type PLATFORM_PERMISSIONS = (typeof PERMISSIONS)[number];
```
supalarry, line 3: Shall we call it singular `type PLATFORM_PERMISSION` because we select specific one using [number], so it can be only 1 specific permission?

Screenshot 2023-11-30 at 10 11 25

supalarry: Below we use `PLATFORM_PERMISSIONS[]` so it would read nicer `PLATFORM_PERMISSION[]`

ThyMinimalDev: indeed good catch

packages/platform/constants/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "@calcom/tsconfig/base.json",
+  "compilerOptions": {
+    "target": "es5",
+    "jsx": "preserve",
```
supalarry, line 5: For constants packages tsconfig and one of types and utils packages - do we need `"jsx": "preserve"` ?  We would need it if we would use `.tsx` files which probably atoms would be the only place for such files.

supalarry (approved):
After my last 2 comments above are addressed - looks good to me!

packages/platform/atoms/package.json
```diff
@@ -19,7 +19,8 @@
     "typescript": "^4.9.4",
     "vite": "^4.1.2"
   },
-  "main": "./index",
+  "main": "./index.ts",
+  "types": "./index.ts",
```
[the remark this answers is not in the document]
ThyMinimalDev: the point is just to tell where to look for exported types, does not have a big impact it's there if we ever export type from the atoms for people to use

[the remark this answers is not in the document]
ThyMinimalDev: yes I often find it useful to have exported types when wanting to build props for a component, it's a good practice to export the component props type for example

Ryukemeister (approved):

PR #12604, opened 2023-11-30
feat: oauth clients management

supalarry:
Fixes PRO-139, v2 API related

dto, repository and controller to manage manage cal oAuth clients for platform users.


apps/api/v2/src/modules/oauth/oauth-client.controller.ts
```diff
@@ -0,0 +1,88 @@
...
+import { AuthGuard } from "@nestjs/passport";
+
+@Controller("oauth-clients")
+export class OAuthClientController {
+  private readonly logger = new Logger("OAuthClientController");
+
+  constructor(private readonly oauthClientRepository: OAuthClientRepository) {}
+
+  @Post("/")
+  @Version(VERSION_NEUTRAL)
```
exception, line 30: These should all be v2, no?

supalarry: Well spotted - will update it

exception (approved):

PR #12610, opened 2023-11-30
fix: Signup options are not disabled

### LINKED ISSUE #12586: Signup options are not disabled after clicking on signup button and some other minor issues
## Issue Summary

Main Issue:

After clicking on signup button, you can still click on signup with google or SAML SSO which is not a good thing and can bring unexpected results. when you click on signup first and then signup with google it will take you to the google signup page. and when you try to go back you'll see that your account is created in a result of your first signup click

[signup-issue-calcom.webm](https://github.com/calcom/cal.com/assets/131521332/f819667a-52d5-4624-9e81-71e26f497486)

### Other Minor Issues:

email-body-issue-calcom

#### 1st Issue 
on verify email section there is a text "verify_email_page_body"  which appears before the main text is loaded. which is not good for user experience.

#### 2nd Issue
on verify email section when you click on back you can see I'm navigated to the previous account deletion survey page which is the result of my prior account deletion. this is unexpected and should not happen. (I also have experienced after signing up you'll be redirected to the verify email section and when you click back 1 or 2 times you'll be redirected to the user dashboard)

#### Suggestion:
user should directly be navigated to dashboard on signup because without verifying email you can still access the user dashboard so why not directly redirect the user to the main page and the main topbar will guide the user that their email is not verified so they have to verify the email.

### Steps to Reproduce the Main Issue:

1. Go to signup page and fill out the required details.
2. Click on signup and then click on signup with google.
3. You'll be redirected to google signup page and your account will be created because of first signup click.
4. for other issues watch full video.

Pratik-Kumar-621:
## What does this PR do?

Fixes #12586 (issue)

## Type of change

-  Bug fix (non-breaking change which fixes an issue)
  <b>Major Issue:</b> While clicking on signup button there is no actions provided to disable the other signup options resulting in the main issue. Fixed it by creating a boolean state clickObserv which helps in disabling the other signup buttons when one is clicked.
--
<b>Minor Issue 1:</b> There are 2 options to improve the user experience in this case. Either by loading the verification page only after the verify_email_page_body is loaded or by changing the verify_email_page_body with a spin loader.
--
<b>Minor Issue 2:</b> This is happening because those routes are not protected. It's not like you are navigated to delete survey page on back click. If you  open any other page before signup you will be redirected to that particular page on clicking back as those routes are not protected and there is no conditions for visiting any route. From my opinion this is not considered as a minor issue but it's a major issue.
<b>Regarding the Suggestion:</b> As per my thought, Instead of navigating them directly to dashboard page we can add a skip or "do it later" button and let the user decide that he need to verify the email or access the dashboard.

## How should this be tested?
[fixed_disabling.webm](https://github.com/calcom/cal.com/assets/70286186/8c8afc1c-e4e2-4f15-a4b0-07c29e95c665)

As you see in the video that the other buttons got disabled when click on create account. Same happens if user clicked on any other signup option

### DIFF
--- a/apps/web/pages/signup.tsx
+++ b/apps/web/pages/signup.tsx
@@ -323,6 +323,7 @@ export default function Signup({
                   !!formMethods.formState.errors.email ||
                   !formMethods.getValues("email") ||
                   !formMethods.getValues("password") ||
+                  isSubmitting ||
                   usernameTaken
                 }>
                 {premiumUsername && !usernameTaken
@@ -388,7 +389,8 @@ export default function Signup({
                       !!formMethods.formState.errors.username ||
                       !!formMethods.formState.errors.email ||
                       premiumUsername ||
-                      isSubmitting
+                      isSubmitting ||
+                      isGoogleLoading
                     }
                     className={classNames(
                       "w-full justify-center rounded-md text-center",
--- a/apps/web/pages/signup.tsx
+++ b/apps/web/pages/signup.tsx
@@ -204,7 +204,6 @@ export default function Signup({
   const isOrgInviteByLink = orgSlug && !prepopulateFormValues?.username;
 
   const signUp: SubmitHandler<FormValues> = async (data) => {
-    setClickObserve(true);
     await fetch("/api/auth/signup", {
       body: JSON.stringify({
         ...data,
@@ -235,7 +234,6 @@ export default function Signup({
       })
       .catch((err) => {
         formMethods.setError("apiError", { message: err.message });
-        setClickObserve(false);
       });
   };
 
@@ -379,7 +377,6 @@ export default function Signup({
                         return;
                       }
                       router.push(GOOGLE_AUTH_URL);
-                      setClickObserve(false);
                     }}>
                     Google
                   </Button>
@@ -391,7 +388,7 @@ export default function Signup({
                       !!formMethods.formState.errors.username ||
                       !!formMethods.formState.errors.email ||
                       premiumUsername ||
-                      clickObserve
+                      isSubmitting
                     }
                     className={classNames(
                       "w-full justify-center rounded-md text-center",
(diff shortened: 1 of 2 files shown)


apps/web/pages/signup.tsx
```diff
@@ -161,6 +161,7 @@ export default function Signup({
 }: SignupProps) {
   const [premiumUsername, setPremiumUsername] = useState(false);
   const [usernameTaken, setUsernameTaken] = useState(false);
+  const [clickObserve, setClickObserve] = useState(false);
```
Udit-takkar, line 164: We don't need this. you can use isSubmitting variable

apps/web/pages/signup.tsx
```diff
@@ -318,7 +321,8 @@ export default function Signup({
                 disabled={
                   !!formMethods.formState.errors.username ||
                   !!formMethods.formState.errors.email ||
-                  usernameTaken
+                  usernameTaken ||
+                  clickObserve
```
Udit-takkar, line 325:
```suggestion
                  usernameTaken ||
                  isSubmitting
```

apps/web/pages/signup.tsx
```diff
@@ -343,12 +347,13 @@ export default function Signup({
                 {isGoogleLoginEnabled ? (
                   <Button
                     color="secondary"
-                    disabled={!!formMethods.formState.errors.username || premiumUsername}
+                    disabled={!!formMethods.formState.errors.username || premiumUsername || clickObserve}
```
Udit-takkar, line 350:
```suggestion
                    disabled={!!formMethods.formState.errors.username || premiumUsername || isSubmitting}
```

Pratik-Kumar-621:
@Udit-takkar Suggested changes are done. Is there anything else which i need to do?

shazzx:
@Pratik-Kumar-621 @Udit-takkar 
I think the signup issue is already fixed by someone or the team. but the same issue is also on the login page. (after clicking on login other login options are not disabled) fix that one.

I could have fixed all of those issues but the project is so large that my laptop is not capable of running it lol.

Gambatte udasai!

Udit-takkar (approved):

PR #12611, opened 2023-11-30
feat: Instant meeting

### LINKED ISSUE #12034: [CAL-2637] Instant Meeting
# problem:

* sales teams want to do a call immediately if possible

# solution:

* add "Instant Meeting" as an option for Round Robin events
* if a person is available, show a link: "Join Meeting" immediately
* use webhook and zapier for notifications
* notify host
  * v0: send a slack message to a group "A lead is looking for a call"
    * have a dynamic link that expires so only one person can join
    * i.e.: [http://app.cal.com/instant-meeting/6YsEEOIom8HbXrzQOWdo](http://app.cal.com/instant-meeting/6YsEEOIom8HbXrzQOWdo)
    * later: add more notification options (SMS, Email, Phone Call)
  * require logged in user for the "host"
  * track who joins (need logged in [cal.com](http://cal.com) account)
  * trigger a full booking flow (as if someone booked a time in the future)
* add that event to their account (for analytics purposes)
* fastest to join gets the call
* requires team to be on enterprise (Upgrade button)

[https://github.com/calcom/cal.com/assets/8019099/c0baff7e-889f-4e75-90ba-6aa6a25f9845](https://github.com/calcom/cal.com/assets/8019099/c0baff7e-889f-4e75-90ba-6aa6a25f9845)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2637](https://linear.app/calcom/issue/CAL-2637/instant-meeting)</sub>

Udit-takkar:
Instant Meeting v1

- Instant Meetings are only available for org team event types.
- Only Cal video room is supported.
- Booker books the event and a slack message is sent with url to connect and join to meeting link as configured in zapier by team admin/team member.
- only team member part of the same team as event type would be able to accept the booking

//TODO: Add new demo video
Demo:-

Flow

Create team event type for organization and enable "Instant Booking" switch in event type settings

1) Booker goes to org team event type and sees a bottom banner "Don't want to wait?".
2) Click on "connect now" should directly open BookEventForm with slot start time as current time and duration as default duration.
3) Booker books the event and sees a modal with text "we are connecting you to someone".
4) Now check the DB and get the instant meeting token table and copy the token and go to url http://app.cal.local:3000/connect-and-join?token={token}. 
5) Click on Join Meeting to accept the meeting and redirect to the meeting url

(To test wehooks ) Go to zapier and  create a zap and select "webhooks by zapier" and get the webhook url from Test section and then create a webhook in cal.com in that event type with this url and "instant meeting" as webhook trigger)

Screenshot 2023-12-13 at 9 14 22 PM


apps/web/pages/connect-and-join.tsx
PeerRich, line 1: can we rename this to /pages/instant-meeting.tsx

apps/web/pages/api/book/instant-event.ts
```diff
@@ -0,0 +1,22 @@
...
+import { checkRateLimitAndThrowError } from "@calcom/lib/checkRateLimitAndThrowError";
+import getIP from "@calcom/lib/getIP";
+import { defaultResponder } from "@calcom/lib/server";
+
+async function handler(req: NextApiRequest & { userId?: number }, res: NextApiResponse) {
+  const userIp = getIP(req);
+
+  await checkRateLimitAndThrowError({
+    rateLimitingType: "core",
+    identifier: userIp,
```
sean-brydon, line 14: Lets make this more specific so we can get a detailed breakdown in upstash of where things could be potentially abused
```suggestion
    identifier: `instant.event-${userIp}`,
```

packages/features/bookings/Booker/store.ts
```diff
@@ -272,6 +276,19 @@ export const useBookerStore = create<BookerStore>((set, get) => ({
...
+      const setSelectedTimeslot = get().setSelectedTimeslot;
+      const setMonth = get().setMonth;
+      const setSelectedDate = get().setSelectedDate;
+
+      setMonth(dayjs().format("YYYY-MM"));
+      setSelectedDate(dayjs().format("YYYY-MM-DD"));
+
+      const currentTime = new Date().toISOString();
+      setSelectedTimeslot(currentTime);
+      set({ isInstantMeeting });
```
sean-brydon, line 290: I think we should be using set for this - Using get get fn to get the setters can have issues when we use shallow IIRC 

```suggestion
    set({
        month: dayjs().format("YYYY-MM"),
        selectedDate: dayjs().format("YYYY-MM-DD"),
        selectedTimeslot: new Date().toISOString(),
        isInstantMeeting,
      });
```

packages/trpc/server/routers/loggedInViewer/connectAndJoin.handler.ts
```diff
@@ -0,0 +1,120 @@
...
+            },
+          }),
+    },
+    select: {
+      location: true,
+      metadata: true,
+    },
+  });
+
+  const locationVideoCallUrl: string | undefined = bookingMetadataSchema.parse(
```
sean-brydon, line 108:
```suggestion
  const locationVideoCallUrl = bookingMetadataSchema.parse(
```

sean-brydon (commented):
I know its early but heres a few NITs i found when scanning. 

Once thing i noticed on the FE side is that meetingURL is always possible if a user has already joined the meeting. 

Do we want to pass this from the BE even if that is the case?

Udit-takkar:
> Once thing i noticed on the FE side is that meetingURL is always possible if a user has already joined the meeting.
> 
> Do we want to pass this from the BE even if that is the case?

Yes if the logged in user is part of the same team then 

1) If meeting has not been accepted by anyone then we redirect to the meeting url.
2) If some else already accepted the meeting then we need to display this and give option to still join the meeting if they want to.

Screenshot 2023-12-09 at 2 18 40 PM

I write all the requirements in the description.

PR #12613, opened 2023-12-01
chore: platform more permissions utils and tests

ThyMinimalDev:
## What does this PR do?

Adds more platform permissions utils and tests

### DIFF
--- a/packages/platform/utils/permissions.ts
+++ b/packages/platform/utils/permissions.ts
@@ -1,3 +1,12 @@
+import {
+  BOOKING_READ,
+  BOOKING_WRITE,
+  EVENT_TYPE_READ,
+  EVENT_TYPE_WRITE,
+  PERMISSIONS,
+  SCHEDULE_READ,
+  SCHEDULE_WRITE,
+} from "@calcom/platform-constants";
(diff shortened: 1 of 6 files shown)


packages/platform/utils/permissions.ts
```diff
@@ -9,3 +18,36 @@ export const hasPermissions = (userPermissions: number, permissions: PLATFORM_PE
   // use bitwise AND to check if each required permission is present
   return permissions.every((permission) => hasPermission(userPermissions, permission));
 };
+
+export const hasEventTypeReadPermission = (userPermissions: number): boolean => {
```
supalarry, line 22: Shall we have `utils/permissions` as a folder and then sub-folders for booking, event-type and schedule e.g. `utils/permissions/booking` ?

It's okaish now to keep track of 3 permission groups but if we would add more stuff than booking, event-type and schedule then its hard to navigate everything in 1 file.

What do you think?

supalarry: Then we would also have smaller test file as each test group would be in its own folder

ThyMinimalDev: let's see how it goes, we can do it when the need appears

supalarry (approved):

PR #12625, opened 2023-12-01
fix: ignore original rescheduled booking for booking limits

CarinaWolli:
## What does this PR do?

Fixes that when a booking is rescheduled it will not be counted as an existing booking for the booking limits. 

So if there is a limit of 1 booking a day, the user can still reschedule to the same day. 

Fixes #12347

## Type of change

## How should this be tested?

- Create event type with booking limit 1 per day 
- Make a booking 
- Reschedule booking and see if you have available slots on the days of the original booking


Udit-takkar (dismissed):
LGTM. 

works fine

Screenshot 2023-12-02 at 6 25 07 PM

Udit-takkar:
Some unit tests are failing

Udit-takkar (approved):

PR #12626, opened 2023-12-01
feat: invites bypass disabled signup

sean-brydon:
When disable signup is enabled - people can still be invited via team invites


apps/web/pages/api/auth/signup.ts
```diff
@@ -29,7 +39,8 @@ export default async function handler(req: RequestWithUsernameStatus, res: NextA
   // Use a try catch instead of returning res every time
   try {
     ensureReqIsPost(req);
-    ensureSignupIsEnabled();
+
```
sean-brydon, line 42:
```suggestion
```

Udit-takkar (approved):
LGTM. Thanks for adding E2E test.

PR #12637, opened 2023-12-02
feat: hide default booking success screen

### LINKED ISSUE #12499: Embed: Allow for scheduled confirmation screen to be disabled

We tap into the `bookingSuccessful` event and it's `data` property to get the booking details and show them with our custom UI. We'd like to use the Cal.com embed for our scheduling needs, but would also like to customize the confirmation screen to blend well with our content.

This is the default confirmation screen:

Screenshot 2023-11-22 at 19 13 23

Our need would be solved by allowing the default confirmation screen to be disabled, and instead we would be able to handle the `bookingSuccessful` event without the confirmation screen popping up. Sadly, redirecting immediately after the `bookingSuccessful` event fires is not a great experience because there is some flickering involved.

I have not found any native solution with the current implementation of the React embed. I have considered forking the package and implementing some custom event or property that would allow for the confirmation screen to be disabled.

thehanimo:
## What does this PR do?
Fixes #12499

Adds a hideSuccessScreen event that lives under the advanced section of the eventType.

`/api/book/event/` returns a boolean hideSuccessScreen which is then used by useBookingSuccessRedirect hook to add it as a query param to `/booking/*`. This field is used to add a `hidden` className on the wrapper component. This should prevent flicker as mentioned in the bug.

In case the user toggles this field but does not handle the `bookingSuccessful` event to show a custom UI, the user will be stuck on an empty page. Something worth considering!

TODO:
- Add translations for `disable_success_screen ` and `disable_success_screen_description`.

## Type of change

## How should this be tested?

- Toggle the `Hide default success screen` field to true in the advanced section.
- Booking an event should show nothing on the confirmation screen.

## Mandatory Tasks


exception:
Please ensure e2e tests are passing!

emrysal (changes requested):
Hello, I pushed a small commit to fix your type errors; please review. Regarding:

> "In case the user toggles this field but does not handle the bookingSuccessful event to show a custom UI, the user will be stuck on an empty page. Something worth considering!"

There needs to be a failsafe for this; because a user being able to get the app into a broken state (or their booking UI in a broken state) is never an acceptable merge.

Please request my review again when these changes are done 🙏

EnriqCG:
I agree with @thehanimo; I understand that getting the app into a broken state is not desirable but it can also happen when setting the wrong URL using "Redirect on booking". We could add a warning on the settings page to indicate this behavior.

PR #12638, opened 2023-12-02
fix: type error in `settings/admin` in app router

hbjORbj:
## What does this PR do?

Fixes the type errors occurring in four pages under `settings/admin` in app router directory

## Type of change

## How should this be tested?

run type-check command in apps/web

## Mandatory Tasks


zomars (approved):
Interesting fix. Do you have some docs about the expected bundling differences when using "use client"?

hbjORbj:
@zomars Hey Omar. Sure thing. In this [React doc on 'use client'](https://react.dev/reference/react/use-client#use-client), it says:
> When a file marked with 'use client' is imported from a Server Component, [compatible bundlers](https://react.dev/learn/start-a-new-react-project#bleeding-edge-react-frameworks) will treat the module import as a boundary between server-run and client-run code.

Issue #11655, opened 2023-10-02
[CAL-2564] recurring events: use proper "recurring" API from calendar APIs


PeerRich:
right now people select how many events they wanna book (i.e. 12 Mondays)

this shortcut was taken to ship this feature faster.

lets use the proper "recurring" APIs from google/outlook (maybe caldav?)

we could fall back to individual events for other calendars but I think all of them should support recurring events 

<sub>[CAL-2564](https://linear.app/calcom/issue/CAL-2564/recurring-events-use-proper-recurring-api-from-calendar-apis)</sub>

shakthi2003-ice:
Hey @PeerRich 
It would be grateful if you could assign me this issue.

PeerRich:
no need to ask, just open a PR

keithwillcode:
Moving to v3.6 with high priority because this will help fix many issues with recurring events

PR #12639, opened 2023-12-02
feat: Adjusted Recurring Event Booking Menu to Follow Google Calendar-like Structure

cvz5:
…Format

## What does this PR do?
Adjusted the menu when setting a recurring event to be more similar to GCal.
Specifically, set the default values to be yearly, monthly, weekly, daily, weekdays and custom.
Toggling custom allows selection of intervals of x days/weeks/months/years

Added an option to denote meetings should continue indefinitely-sets count behind the scenes to max value.
Otherwise, allows both selection of max meeting numbers and/or a definite end date.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

Tested this by running development and making sure that all the updates saved correctly. We checked all references to changed variables to make sure nothing else was affected.

## Mandatory Tasks

## Checklist


CarinaWolli:
Closing this as it doesn't fix issue #11655. As described in the issue description, we are looking to implement proper recurring events with the calendar's API. This approach should replace the current method of creating individual events for each booking.

PR #12642, opened 2023-12-03
chore: Sentry Wrapper with Performance and Error Tracing

bwoody13:
## What does this PR do?

Creates a wapper for callbacks using Sentry to monitor performance and log errors directly into the dashboard. We updated all calls in `getUserAvailability` as a demonstration of how to use the wrapper and confirmed that calls are being traced in the dashboard by making a booking. Adds tracking to an existing parent transaction if it can locate one. See photo below for this.
image

Fixes # (issue)
https://osshack.com/bounty/29

Updated `tracesSampleRate` in `apps/web/sentry.server.config.ts`  to 1.0, so that transactions can be traced.

## Type of change

## How should this be tested?

Create bookings and check that the dashboard is monitoring performance.

## Mandatory Tasks

## Performance Results
Running a script that performs a simple mathematical calculation within a loop of 1000000 iterations. See comments for the script. Our results were: `Plain execution time: 441, Monitored execution time: 8094`. This suggests that using these wrappers within large loops can incur significant overhead and is thus not recommended.

For smaller loops, the cost incurred may not be very significant on an absolute scale
considering that a million monitored iterations only took roughly 8 seconds when monitored.


packages/core/getUserAvailability.ts
```diff
@@ -122,8 +135,39 @@ export const getCurrentSeats = (eventTypeId: number, dateFrom: Dayjs, dateTo: Da
...
+      eventType: Pick<
+        PrismaEventType,
+        "id" | "beforeEventBuffer" | "afterEventBuffer" | "seatsPerTimeSlot"
+      > | null;
+      _count?: {
+        seatsReferences: number;
+      };
+    })[];
+  }
+) => {
```
ThyMinimalDev, line 165: when writing the wrapper functions, no need to respecify all the parameters, you can use typescript `Parameters<typeof _getUserAvailability>` and the same approach used in sentry wrapper with `...args`, example:

```ts
const getEventType = async (...args: Parameters<typeof _getEventType>) => {
  return monitorCallback(_getEventType, args);
};
```

bwoody13: In the newest commit. Thanks for the suggestion!

keithwillcode (approved):
Nice work!

packages/core/sentryWrapper.ts
```diff
@@ -0,0 +1,86 @@
...
+  // Check if there's an existing transaction, if not, start a new one
+  if (!transaction) {
+    transaction = Sentry.startTransaction({
+      op: name,
+      name: name,
+    });
+  }
+
+  // Start a new span in the current transaction
+  const span = transaction.startChild({
```
zomars, line 27: Seems like this is causing this https://github.com/calcom/cal.com/issues/12753

Issue #12289, opened 2023-11-08
Serve an official OpenAPI spec?


another-guy:
### Is your proposal related to a problem?

Cal.com service API is unclear. The documentation under https://cal.com/docs/enterprise-features/api is lacking necessary details for reliable integration.
E.g. what's the shape of EventType object? Which fields are required, which are optional, and so on.
<img width="750" alt="image" src="https://github.com/calcom/cal.com/assets/5762340/309a502e-7bdf-43ec-b451-a633e9200531">

### Describe the solution you'd like

Post an official automatically generated documentation in two forms:
1. openapi.json/yml for machines and humans
2. the interactive HTML version of the above (swagger or similar)

### Describe alternatives you've considered

Not sure

---
##### House rules
- If this issue has a `🚨 needs approval` label, don't start coding yet. Wait until a core member approves feature request by removing this label, then you can start coding.
  - For clarity: Non-core member issues automatically get the `🚨 needs approval` label.
  - Your feature ideas are invaluable to us! However, they undergo review to ensure alignment with the product's direction.

alishaz-polymath:
Thank you for your suggestion. We're in the middle of reworking the API docs, and this suggestion will definitely be taken into consideration 🙏

cmgriffing:
I recently did some discovery around this and found some interesting info.

The public URL is: https://api.cal.com/docs
However, that spec is missing all of the paths for each tag.

Curiously, if you pull down the repo and run the API app directly, you will get a full spec with the paths:
```
cd apps/api
yarn dev
curl http://localhost:3002/docs
```

I'm uncertain of where the script for deploying the API lives, but my first guess would be to look at that and try and figure out why the production URL is missing the paths from the OpenAPI spec.

`next-swagger-doc` seems to be the library responsible for generating the spec. https://github.com/jellydn/next-swagger-doc
Updating to the latest version did not seem to have any effect in my local testing.

cmgriffing:
I just had a random thought. Is it possible that the deploy script is stripping the comments from the code before deploy? This would mean that next-swagger-doc wouldn't be able to see the `@swagger` JSDocTag since it got removed.

Not sure, but a random thought I had that might help someone with higher access privileges troubleshoot this.

CodeFromAnywhere:
Curious on the status of this.

CarinaWolli:
Closing as our API docs had some major improvements since then: https://cal.com/docs/api-reference/v2/introduction

PR #12644, opened 2023-12-03
feat: Validate and test OpenAPI spec

cmgriffing:
## What does this PR do?

This PR attempts to address testing the OpenAPI spec for consistency with the API by using Dredd.js. Dredd tests all listed responses including errors. Some of the low-hanging fruit to get some of the tests to pass were as simple as changing status codes or specifying the proper response body for a 401.

It does not solve all outstanding issues. I would be happy to work with someone from your team to help them understand how to fix the remaining issues. Some tests were skipped because testing them proves harder without modifying the seed script even more.

Current status:
`complete: 126 passing, 35 failing, 0 errors, 38 skipped, 199 total`

There is an issue with the production docs endpoint. I have commented about it here: https://github.com/calcom/cal.com/issues/12289#issuecomment-1837329444
This PR exposes a script for generating the openapi json and yaml files and putting them in the public directory.

One more note: The seed scripts in the prisma package generate a dredd-data.json file. This is used for values that were already seeded but needed to be exposed to the dredd testing script. eg: apiKeys and user Ids associated with them.

## Type of change

## How should this be tested?

To run the tests make sure the API server is running on `localhost:3002`. Then run the following command in the API directory:

```
yarn test:dredd
```

The generated dredd-output.html file can be easier to navigate than the CLI.

## Mandatory Tasks

## Other Notes about the API

The largest issue with the spec seems to be a lack of example responses and requests. These would likely be essential to anyone trying to create an integration.

------ AUTH

In many places, a 401 is thrown when a user does not have access to an item. A 403 or a 404 would be preferable. 403 would leak that the requested entity exists but user just does not have access. 401 does this as well, because we can still fetch other things with our valid apiKey. 404 would obfuscate the existence of the entities.

------ RESPONSE CASING

Most of the responses have a snake cased entity prop. Some have camel-cased. Making this consistent would be ideal.

------ Availability

        The response is not what the @swagger section says it will be.

------ Destination Calendars

      ALL:
        - No credential is seeded for any users and no API exists to make it as part of Dredd script

      POST:
        - credentialId is required but is not in list of params. Seems like credentialId does not actually need to be passed on the body since it gets looked up in the credential table

------ Teams

      POST:
        - Spec does not call out required or optional body fields properly
          - slug is not required
          - isPrivate, hideBookATeamMember, brandColor, darkBrandColor, timeZone, and weekStart are required by the Zod validator, but not listed in spec


cmgriffing (commented):
I have done a full review of files I changed.

cmgriffing:
I ended up needing to rebase. This should be good now. I would love to chat with anyone from your team if there are questions. They could even be a guest on stream if they want. :)

PeerRich:
i believe we are in the middle of deprecating the old URL and the new nestjs implementation comes with OpenAPI specs in-house

cmgriffing:
Cool, looking forward to how that works. I would still probably recommend a Dredd test if language specific SDKs were planned to be generated from the OpenAPI spec.

keithwillcode:
This is cool. Thanks for the contribution @cmgriffing. Will close for now based on what you and Peer discussed but will definitely keep as a reference. 🙏

PR #12651, opened 2023-12-03
fix: Use Google recurring events protocol

### LINKED ISSUE #11655: [CAL-2564] recurring events: use proper "recurring" API from calendar APIs

haranrk:
## What does this PR do?

Uses the native recurring cal API of google calendar to make recurring bookings.
Features - 
- Cancel all remaining events - will cancel the entire recurring event set with one API call 
- Modifying a booking will only change that instance of the recurring event
- Creating a new event will create the recurring events with one API call

Partially fixes #11655 as only the support for Google Calendar is added. Though it should be quick to support other Calendars.

https://www.loom.com/share/e66dad98c4a54d229b480bebf15b074f
### Previous recurring event flow 
- `recurring-event` -> 
   - For every event
       - run `handleNewBooking` -> `eventManager` -> `calendarManager` -> `googleCalendarService`
       - check if the user is available during the proposed event time (this would mean that even if the user is unavailable for a particular instance of the recurring event, only that instance will not be created, but the other instances will)
       - a new standalone event is created

### New recurring event flow
- `recurring-event` -> 
   - Check if the user is available for every instance of the proposed recurring event
   - For every event
       - run `handleNewBooking` -> `eventManager` -> `calendarManager` -> `googleCalendarService`
       - in the first iteration of the loop, a recurring event on gcal is created. this would create all the instances on gcal. 
       - subsequent iterations of the loop do all the additional `cal.com` housekeeping necessary like creating videoconferencing links, making new `cal.com` db records etc.
         
      
       

## Type of change

## How should this be tested?

## Mandatory Tasks


CarinaWolli:
For every event of the recurring booking a new recurring google calendar event is created for me. 

So I end up with a ton of calendar events in my google calendar:
Screenshot 2023-12-05 at 5 18 40 PM

@haranrk I don't see this happening in your loom. Any idea why this is happening?

haranrk:
@CarinaWolli, I just pulled the latest changes and tested it on my end and it worked correctly. Can you give the details of the event you created? 

In the interest of speed, would it be possible to get on a video conferencing call with you to identify the bug?

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +54,25 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
       isNotAnApiCall: true,
-      skipAvailabilityCheck: key >= numSlotsToCheckForAvailability,
+      skipAvailabilityCheck: key >= numSlotsToCheckForAvailability || thirdPartyRecurringEventId !== "",
     });
-
     createdBookings.push(eachRecurringBooking);
+
+    if (thirdPartyRecurringEventId == "") {
+      if (eachRecurringBooking.references!.length > 0) {
+        const bookingReference = eachRecurringBooking.references![0];
```
CarinaWolli, line 70: I found the issue: google_calendar is not always the first reference. And in my case daily_video is. That's why it never got the `recurringMeetingId`

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -27,6 +27,7 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
   req.userId = session?.user?.id || -1;
   const numSlotsToCheckForAvailability = 2;
 
+  let thirdPartyRecurringEventId = "";
```
hariombalhara, line 30: Why not use null for for initial value? null makes more sense for that.

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +54,25 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
       allRecurringDates,
+      thirdPartyRecurringEventId,
       currentRecurringIndex: key,
       noEmail: key !== 0,
     };
 
     const eachRecurringBooking = await handleNewBooking(recurringEventReq, {
       isNotAnApiCall: true,
-      skipAvailabilityCheck: key >= numSlotsToCheckForAvailability,
+      skipAvailabilityCheck: key >= numSlotsToCheckForAvailability || thirdPartyRecurringEventId !== "",
```
hariombalhara, line 64: Don't we want to check for availability for the number of slots equal to `numSlotsToCheckForAvailability` now?

`thirdPartyRecurringEventId` could be non empty after the very first booking right? and then it would stop checking for availability

apps/web/tsconfig.json
```diff
@@ -17,6 +17,7 @@
     ],
     "strictNullChecks": true
   },
+  "sourceMap": true,
```
hariombalhara, line 20: Seems unrelated change. Let's add it in a separate PR if needed.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -219,38 +239,73 @@ export default class GoogleCalendarService implements Calendar {
...
         });
       }
 
       return {
         uid: "",
-        ...event.data,
-        id: event.data.id || "",
+        ...event,
+        id: event.id || "",
+        recurringMeetingId: recurringEventId,
```
hariombalhara, line 301: Better name would be `recurringEventId` only instead of `recurringMeetingId`

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -219,38 +239,73 @@ export default class GoogleCalendarService implements Calendar {
...
+              calendarId: selectedCalendar,
+              eventId: recurringEventId || "",
+            });
+
+            if (recurringEventInstances.data.items) {
+              event = recurringEventInstances.data.items[0];
+            }
+          }
+        }
+      }
```
hariombalhara, line 281: It would be best to abstract it out in getFirstEventInRecurrence()

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -219,38 +239,73 @@ export default class GoogleCalendarService implements Calendar {
...
-      });
+      let event;
+      let recurringEventId = null;
+      if (calEventRaw.existingRecurringEvent) {
+        recurringEventId = calEventRaw.existingRecurringEvent.recurringEventId;
+        const recurringEventInstances = await calendar.events.instances({
+          calendarId: selectedCalendar,
+          eventId: calEventRaw.existingRecurringEvent.recurringEventId,
+        });
+        event = recurringEventInstances.data.items![calEventRaw.existingRecurringEvent.eventIndex];
```
hariombalhara, line 250: This seems like a very risky approach to me. We are using the index of the occurence as per the number of occurences we have in our system. That index is used to get the event in Google Calendar. Google Calendar and Cal.com can easily be out of sync and we can endup accidentally referring to wrong event in Google Calendar.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -219,38 +239,73 @@ export default class GoogleCalendarService implements Calendar {
...
+        if (event.recurrence) {
+          if (event.recurrence.length > 0) {
+            recurringEventId = event.id;
+            const recurringEventInstances = await calendar.events.instances({
+              calendarId: selectedCalendar,
+              eventId: recurringEventId || "",
+            });
+
+            if (recurringEventInstances.data.items) {
+              event = recurringEventInstances.data.items[0];
```
hariombalhara, line 277: Instead of just storing the first occurence we should also plan to store the recurring event ID, so that it can be cleanly deleted from Google Calendar using that ID. It could be a later improvement though.

packages/prisma/schema.prisma
```diff
@@ -359,6 +359,7 @@ model BookingReference {
   /// @zod.min(1)
   uid                String
   meetingId          String?
+  recurringMeetingId String?
```
hariombalhara, line 362:
```suggestion
  recurringEventId String?
```

It's better name for this.

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -452,24 +453,31 @@ async function handler(req: CustomRequest) {
...
-                for (const updBooking of updatedBookings) {
-                  const bookingRef = updBooking.references.find((ref) => ref.type.includes("_calendar"));
-                  if (bookingRef) {
-                    const { uid, externalCalendarId } = bookingRef;
-                    const deletedEvent = await calendar?.deleteEvent(uid, evt, externalCalendarId);
-                    apiDeletes.push(deletedEvent);
+            if (bookingToDelete.references.length > 0 && bookingToDelete?.references[0].recurringMeetingId) {
+              const recurringMeetingId = bookingToDelete?.references[0].recurringMeetingId;
+              apiDeletes.push(
+                calendar?.deleteEvent(recurringMeetingId, evt, externalCalendarId) as Promise<unknown>
```
hariombalhara, line 459: What if we just want to delete the first occurence and not all the occurences?

hariombalhara (changes requested):
Great work @haranrk 🙏 Left some comments

packages/features/bookings/lib/handleCancelBooking.ts
haranrk: Then the else condition at line 483 would be triggered. For recurring events, there are two IDs, the recurring event ID and the instance ID. I've set the instance ID for the existing `UID`, `meetingID` fields in the `BookingReference` database and created a new field for the `recurring ID`. Therefore the existing flows for modifying bookings will affect only that particular instance and not recurring event as a whole.

apps/web/pages/api/book/recurring-event.ts
haranrk: Thanks for the catch, I now check if there are any conflicts in any of the instances in the first iteration of this loop.

apps/web/pages/api/book/recurring-event.ts
haranrk: @CarinaWolli I've fixed it, can you check if it works now?

packages/app-store/googlecalendar/lib/CalendarService.ts
haranrk: Changed so that a match is done by comparing start times

packages/app-store/googlecalendar/lib/CalendarService.ts
haranrk: Renamed it to `thirdPartyRecurringEventId` because there already exists a `recurringEventId` that `calcom` uses

haranrk:
@CarinaWolli @hariombalhara made the changes, PTAL

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -452,24 +453,34 @@ async function handler(req: CustomRequest) {
...
-                for (const updBooking of updatedBookings) {
-                  const bookingRef = updBooking.references.find((ref) => ref.type.includes("_calendar"));
-                  if (bookingRef) {
-                    const { uid, externalCalendarId } = bookingRef;
-                    const deletedEvent = await calendar?.deleteEvent(uid, evt, externalCalendarId);
-                    apiDeletes.push(deletedEvent);
+            if (
+              bookingToDelete.references.length > 0 &&
+              bookingToDelete?.references[0].thirdPartyRecurringEventId
+            ) {
```
CarinaWolli, line 459: same here `google_calendar` is not always the first reference

CarinaWolli (changes requested):
Calendar events are now created correctly when booking a recurring event 👍

I tested rescheduling but calendar events weren't updated correctly. Also cancelling throws an error (same issue with the reference, I added a comment below)

**My test for rescheduling:**
Original dates booked: 13. Dez, 20. Dez, 27. Dez 

Rescheduled 27. Dez to 28. Dez 

Resulted in: event from 20. Dez got moved to 28. Dez

haranrk:
@CarinaWolli I booked a meeting on those dates and was able to move the booking from 27 to 28 successfully. Can you give me more context on how you booked the meeting?

CarinaWolli:
> @CarinaWolli I booked a meeting on those dates and was able to move the booking from 27 to 28 successfully. Can you give me more context on how you updated the meeting?

https://www.loom.com/share/039edd39dbeb42c384d2352f87155f00

CarinaWolli:
There are some type errors, can you take a look at that? @haranrk

CarinaWolli:
Type errors are fixed now. Everything else also looks good to me as well 👍

- Rescheduling one event works 
- Canceling one event works 
- Cancelling the whole series works

Great work @haranrk 👏🏻 @hariombalhara could you take a look again as well?

CarinaWolli (dismissed):
Good to go from my side. Waiting for @hariombalhara before merging 🙏

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +56,34 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
+      }
+    );
+
+    const eachRecurringBooking = await promiseEachRecurringBooking;
 
     createdBookings.push(eachRecurringBooking);
+
+    if (!thirdPartyRecurringEventId) {
+      if (eachRecurringBooking.references && eachRecurringBooking.references.length > 0) {
+        for (const reference of eachRecurringBooking.references!) {
```
hariombalhara, line 79: Let's avoid non-null assertion

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +56,34 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
-    });
+    const promiseEachRecurringBooking: ReturnType<typeof handleNewBooking> = handleNewBooking(
+      recurringEventReq,
+      {
+        isNotAnApiCall: true,
+        skipAvailabilityCheck: key >= numSlotsToCheckForAvailability || thirdPartyRecurringEventId !== null,
+      }
+    );
+
+    const eachRecurringBooking = await promiseEachRecurringBooking;
```
hariombalhara, line 73: Don't understand why `await` is being done in a separate statement. We could simply do await on line 65 and avoid creating the temp variable `promiseEachRecurringBooking`

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +56,34 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
 
     createdBookings.push(eachRecurringBooking);
+
+    if (!thirdPartyRecurringEventId) {
+      if (eachRecurringBooking.references && eachRecurringBooking.references.length > 0) {
+        for (const reference of eachRecurringBooking.references!) {
+          if (reference.thirdPartyRecurringEventId) {
+            thirdPartyRecurringEventId = reference.thirdPartyRecurringEventId;
```
hariombalhara, line 81: It's more concise and readable
`thirdPartyRecurringEventId = eachRecurringBooking.references.find((reference)=>reference.thirdPartyRecurringEventId)?.thirdPartyRecurringEventId`

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1086,6 +1089,26 @@ async function handler(
...
+          {
+            dateFrom: dayjs(start).tz(reqBody.timeZone).format(),
+            dateTo: dayjs(end).tz(reqBody.timeZone).format(),
+            timeZone: reqBody.timeZone,
+            originalRescheduledBooking,
+          },
+          loggerWithEventDetails
+        );
+      }
+    }
```
hariombalhara, line 1111: Now that we are checking for the availability of all slots(as per numSlotsToCheckForAvailability) on first slot iteration itself, what's the need of doing `ensureAvailableUsers` on line 1082 for recurring event.

We could make the code easier to understand by having `isFirstSlot` property in req.body and do away with `skipAvailabilityCheck`.

```javascript
if(req.body.allRecurringDates) {
	if (req.body.isFirstSlot) {
	  // This code to check for availability for all required slots(as per `numSlotsToCheckForAvailability`)
	}
} else {
 // Line 1082 Code
}
```

hariombalhara: This logic is faulty in current state. If we set numSlotsToCheckForAvailability to 3 or more, a recurring event can never be booked because for 3rd slot we would check for 2nd slot availability as well which will always be booked because we reached 3 after iterating 2.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -208,6 +209,16 @@ export default class GoogleCalendarService implements Calendar {
...
 
+    if (calEventRaw.recurringEvent) {
+      const rule = new RRule({
+        freq: calEventRaw.recurringEvent.freq,
+        interval: calEventRaw.recurringEvent.interval,
+        count: calEventRaw.recurringEvent.count,
+      });
+
+      payload["recurrence"] = [rule.toString()];
+    }
```
hariombalhara, line 220: I am not sure if every calendar follows a different format for recurring event creation, but if it doesn't we should really move this code to core so that it's available consistently to all calendars.

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -659,7 +673,8 @@ async function handler(req: CustomRequest) {
   const prismaPromises: Promise<unknown>[] = [bookingReferenceDeletes];
 
   try {
-    const settled = await Promise.allSettled(prismaPromises.concat(apiDeletes));
+    const temp = prismaPromises.concat(apiDeletes);
+    const settled = await Promise.allSettled(temp);
```
hariombalhara, line 677: I think this change can be reverted.

packages/features/bookings/types.ts
```diff
@@ -25,7 +25,7 @@ export type RecurringBookingCreateBody = BookingCreateBody & {
   noEmail?: boolean;
   recurringCount?: number;
   appsStatus?: AppsStatus[] | undefined;
-  allRecurringDates?: string[];
+  allRecurringDates?: Record<string, string>[];
```
hariombalhara, line 28: nit: `{start:string, end:string}`

packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -452,24 +453,37 @@ async function handler(req: CustomRequest) {
...
-                    const { uid, externalCalendarId } = bookingRef;
-                    const deletedEvent = await calendar?.deleteEvent(uid, evt, externalCalendarId);
-                    apiDeletes.push(deletedEvent);
+            let thirdPartyRecurringEventId;
+            for (const reference of bookingToDelete.references) {
+              if (reference.thirdPartyRecurringEventId) {
+                thirdPartyRecurringEventId = reference.thirdPartyRecurringEventId;
+                break;
+              }
+            }
```
hariombalhara, line 462: Not a fan of adding inline code in an already huge function. How about abstracting it out in 2 fns.

getThirdPartyRecurringEventId() -> Scans all references to find it

hariombalhara (changes requested):
Great work @haranrk 🙏 Almost there.

There is  just [this comment]( https://github.com/calcom/cal.com/pull/12651/files#r1429775451) that I would like to be addressed before this PR can be merged


Feel free to open followup PR for other suggestions

Tests Run
- Outlook Calendar(Recurring API not used) - Recurring Event works
                - Cancellation works
                    - Individual and all
            **- numSlotsToCheckForAvailability - Setting it to 3. Booking fails** 
            - Side feature of the PR that if conflict is there on 2nd day, no booking is created at all
- Google Calendar(Recurring API used) - Recurring Event booking works
    - If there is a conflict in [Cal.com](http://cal.com/) booking fails.
    - Cancellation works - Individual and all.
- No Calendar - Recurring Event booking works

packages/features/bookings/lib/handleNewBooking.ts
haranrk: @hariombalhara 

> Blocker: This logic is faulty in current state. If we set numSlotsToCheckForAvailability to 3 or more, a recurring event can never be booked because for 3rd slot we would check for 2nd slot availability as well which will always be booked because we reached 3 after iterating 2.

After the first call to `handleNewBooking` in `recurring-event.ts`, `thirdPartyRecurringEventId` will be set. For every subsequent call to `handleNewBooking`, `skipAvailabilityCheck` will be set to `True`. Therefore, in this example, for 3rd slot, 2nd slot availability will not be checked.

haranrk: > Now that we are checking for the availability of all slots(as per numSlotsToCheckForAvailability) on first slot iteration itself, what's the need of doing `ensureAvailableUsers` on line 1082 for recurring event.
> 
> We could make the code easier to understand by having `isFirstRecurringSlot` property in req.body, and do away with `skipAvailabilityCheck`.
> 
> ```js
> if(req.body.allRecurringDates) {
> 	if (req.body.isFirstRecurringSlot) {
> 	  // This code to check for availability for all required slots(as per `numSlotsToCheckForAvailability`)
> 	}
> } else {
>  // Line 1082 Code
> }
> ```
> 
> In this way, I believe, we would be able to also remove the newly added thirdPartyRecurringEventId related logic in recurring-event.ts.

`thirdPartyRecurringEventId` would still be necessary because for every event instance within `cal.com` this value would need to be stored in the database.

`skipAvailabilityCheck` would also still be necessary for the logic that uses `numSlotsToCheckForAvailability`.

I can rewrite this code in the format you had suggested with the addition of the `isFirstRecurringSlot`. Let me know if you'd still like me to do that.

haranrk:
@hariombalhara 

PTAL at these and let me know how you would like me to resolve them

https://github.com/calcom/cal.com/pull/12651/files#r1442416872
https://github.com/calcom/cal.com/pull/12651/files#r1442419477

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara: > @hariombalhara
> 
> > Blocker: This logic is faulty in current state. If we set numSlotsToCheckForAvailability to 3 or more, a recurring event can never be booked because for 3rd slot we would check for 2nd slot availability as well which will always be booked because we reached 3 after iterating 2.
> 
> After the first call to `handleNewBooking` in `recurring-event.ts`, `thirdPartyRecurringEventId` will be set. For every subsequent call to `handleNewBooking`, `skipAvailabilityCheck` will be set to `True`. Therefore, in this example, for 3rd slot, 2nd slot availability will not be checked.

The issue comes when there are no calendars and conferencing apps connected. In that case, skipAvailability isn't correctly set. See https://www.loom.com/share/4b17b060bcab4ba0a1ec35dea8f973d7

hariombalhara: > thirdPartyRecurringEventId would still be necessary because for every event instance within cal.com this value would need to be stored in the database.

I meant to remove it's usage in recurring-event.ts only, the goal being simplifying the code and making it easy to understand. I still like the `isFirstRecurringSlot` approach if that's fine by you.

hariombalhara:
> @hariombalhara
> 
> PTAL at these and let me know how you would like me to resolve them
> 
> [#12651 (files)](https://github.com/calcom/cal.com/pull/12651/files#r1442416872) [#12651 (files)](https://github.com/calcom/cal.com/pull/12651/files#r1442419477)

Replied inline

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +56,34 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
 
-    const eachRecurringBooking = await handleNewBooking(recurringEventReq, {
-      isNotAnApiCall: true,
-      skipAvailabilityCheck: key >= numSlotsToCheckForAvailability,
-    });
+    const promiseEachRecurringBooking: ReturnType<typeof handleNewBooking> = handleNewBooking(
+      recurringEventReq,
+      {
+        isNotAnApiCall: true,
+        skipAvailabilityCheck: key >= numSlotsToCheckForAvailability || thirdPartyRecurringEventId !== null,
```
hariombalhara, line 69: Also, I wanted to point out that for the calendar apps that don't implement recurring api(all except Google Calendar) at the moment, `thirdPartyRecurringEventId` comes as undefined which is not really a value but `skipAvailabilityCheck` is correctly calculated because `null !== undefined`. 

We should explicitly handle this.

apps/web/pages/api/book/recurring-event.ts
```diff
@@ -53,16 +56,34 @@ async function handler(req: NextApiRequest & { userId?: number }, res: NextApiRe
...
+      }
+    );
+
+    const eachRecurringBooking = await promiseEachRecurringBooking;
 
     createdBookings.push(eachRecurringBooking);
+
+    if (!thirdPartyRecurringEventId) {
```
hariombalhara, line 77: It would be `undefined` here as well for every iteration of non-recurring event supported calendars

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1081,58 +1081,85 @@ async function handler(
...
+        }
       }
-      luckyUsers.push(newLuckyUser);
     }
-    // ALL fixed users must be available
-    if (
-      availableUsers.filter((user) => user.isFixed).length !== users.filter((user) => user.isFixed).length
-    ) {
-      throw new Error(ErrorCode.HostsUnavailableForBooking);
+    if (!req.body.allRecurringDates || req.body.isFirstRecurringSlot) {
```
haranrk, line 1115: @hariombalhara PTAL

haranrk (commented):
@hariombalhara I fixed the bug and rewrote the code to make it more easy to read. Since, the `skipAvailabilityCheck` is not being used anywhere else, I removed it.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1230,58 +1226,85 @@ async function handler(
...
+      users: users as IsFixedAwareUser[],
+      ...(eventType.recurringEvent && {
+        recurringEvent: {
+          ...eventType.recurringEvent,
+          count: recurringCount || eventType.recurringEvent.count,
+        },
+      }),
+    };
+    if (req.body.allRecurringDates) {
+      if (req.body.isFirstRecurringSlot) {
```
hariombalhara, line 1243:
```suggestion
      if (req.body.allRecurringDates && req.body.isFirstRecurringSlot) {
```

hariombalhara, line 1243:
```suggestion
      if (req.body.allRecurringDates && req.body.isFirstRecurringSlot) {
```

hariombalhara (approved):

hariombalhara:
It's merged 🎉Thanks @haranrk 🙏

PR #12653, opened 2023-12-03
feat: OOO booking-forwarding

### LINKED ISSUE #11031: [CAL-2900] RFC: handoff redirection (OOO, Vacation)
same as email, where you can redirect incoming emails to a new inbox for a predefined period, we wanna auto-redirect cal.com links to someone elses link who is temporarily replacing you

scope of how this will look like will follow

open questions:
- how will round robin/collective be affected?

<sub>[CAL-2900](https://linear.app/calcom/issue/CAL-2900/rfc-handoff-redirection-ooo-vacation)</sub>

alannnc:
## What does this PR do?

New feature that should handle profile redirection or away status if setup by user.

Fixes #11031

https://www.loom.com/share/1774045f98ed4759bf57322e62bf63f4

## Type of change

## How should this be tested?

- What are the minimal test data to have? Seed data, teampro, pro users
- What is expected (happy path) to have (input and output)? When setup we handle redirect on user out of office to selected user.

## Mandatory Tasks


apps/web/pages/[user].tsx
```diff
@@ -99,6 +105,24 @@ export function UserPage(props: InferGetServerSidePropsType<typeof getServerSide
...
+                })}{" "}
+                🏝️
+              </h2>
+              <p className="text-default text-sm">
+                {t("user_redirect_description", {
+                  profile: {
+                    username: user.username,
+                  },
+                })}{" "}
+                😄
```
Comment (emrysal) [inline apps/web/pages/[user].tsx:122]: 😄

CarinaWolli (changes requested):
Awesome work @alannnc 👏🏻  I've reviewed the code and tested it locally and found a few (mainly small) issues that should be addressed.

In addition to my comments. Here is one more thing I found: 
- If the user has only one event type then you will get redirected to the event-type link without any information that the original user is away. That's because of the code in [line 425](https://github.com/calcom/cal.com/pull/12653/files#diff-46c8cd3bb294c9d77189134b00657b7e2c4a41e2798899beefdeaf3ea186814aR425). I would suggest that in that case we should not directly redirect to the event-type

alannnc:
Thanks a lot @CarinaWolli ! Suggested changes are fixed now :)

packages/trpc/server/routers/loggedInViewer/bookingForwarding.handler.ts
```diff
@@ -0,0 +1,266 @@
...
+    throw new Error("booking_redirect_limit_reached");
+  }
+
+  const createdForwarding = await prisma.bookingForwarding.create({
+    data: {
+      uuid: uuidv4(),
+      start: dayjs(input.startDate).startOf("day").toISOString(),
+      end: dayjs(input.endDate).endOf("day").toISOString(),
+      userId: ctx.user.id,
+      ...{ ...(input.teamId && { toTeamId: Number(input.teamId) }) },
```
CarinaWolli, line 130: Do we allow passing a teamdId? 

`BookingForwardingForm` only includes  
```
{ 
  startDate: string;
  endDate: string;
  toUsernameOrEmail: string;
}
```
so I don't think we ever have a teamId coming in, maybe I am missing something

packages/trpc/server/routers/loggedInViewer/bookingForwarding.schema.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from "zod";
+
+export const ZBookingForwardingInputSchema = z.object({
+  startDate: z.string(),
+  endDate: z.string(),
+  toUsernameOrEmail: z.string().optional(),
+  teamId: z.string().optional(),
```
CarinaWolli, line 7: When do we use that?

CarinaWolli (changes requested):
There are still some unresolved comments and I just re-reviewed your code and added new comments. @alannnc could you please take a look at all unresolved comments?

PeerRich (changes requested):
we also have the "Set yourself as away" feature in the dropdown, i wonder if we should combine the two

image

--

this UI looks very technical. i dont think its clear what "Username or email" means.

image

--
gendered language, should be neutral "their profile"


image

--

this could get the emptyscreen component with some icons etc. and an action button (go back to home) or so

image

PeerRich:
i am not sure if we should have a generic username input field (that anyone could potentially abuse and send spam emails) vs a dropdown of all your team member colleagues.

I dont mind making this feature teams only behind an upgrade screen (`<UpgradeTip/>`)

packages/trpc/server/routers/loggedInViewer/bookingForwarding.handler.ts
alannnc: Yeah, since there wasn't much details in the description I kinda middle ground it just in case. From peer's comment will make this only for teams and update inputs. :)

alannnc:
> we also have the "Set yourself as away" feature in the dropdown, i wonder if we should combine the two

Probably changing that actionable button to instead navigate to the out-of-office page when you can simply mark yourself as away or forward your bookings to team profile.
 
> this UI looks very technical. i dont think its clear what "Username or email" means.

Yeah definitely will improve with team member selector and probably using same DateRange from insights.

 
> -- gendered language, should be neutral "their profile"

done

> this could get the emptyscreen component with some icons etc. and an action button (go back to home) or so

Taking a look here 

@PeerRich

apps/web/pages/settings/my-account/out-of-office/index.tsx
```diff
@@ -0,0 +1,367 @@
...
+import { useLocale } from "@calcom/lib/hooks/useLocale";
+import { trpc } from "@calcom/trpc/react";
+import useMeQuery from "@calcom/trpc/react/hooks/useMeQuery";
+import { Button, ButtonGroup, Meta, showToast, Badge, Select, SkeletonText } from "@calcom/ui";
+import { TableNew, TableBody, TableCell, TableHead, TableHeader, TableRow } from "@calcom/ui";
+import { FastForward, Link, Moon, Send, Trash2 } from "@calcom/ui/components/icon";
+
+import PageWrapper from "@components/PageWrapper";
+
+import "./DateSelect.css";
```
CarinaWolli, line 20: this breaks the page for me: 

Screenshot 2023-12-18 at 4 24 06 PM

packages/trpc/server/routers/loggedInViewer/bookingForwarding.handler.ts
```diff
@@ -0,0 +1,266 @@
...
+      toUserId: toUserId,
+    },
+  });
+
+  if (bookingForwarding) {
+    throw new Error("booking_forwarding_already_exists");
+  }
+
+  // Prevent infinite redirects
+  const existingBookingForwarding = await prisma.bookingForwarding.findFirst({
```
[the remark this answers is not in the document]
CarinaWolli: We should only check if a forwarding request exists in the same time range

Example: 
User1 can have a forwarding to User2 on Dec 20 but User2 can have forwarding to User1 on Dez 22.

packages/trpc/server/routers/loggedInViewer/bookingForwarding.handler.ts
```diff
@@ -0,0 +1,266 @@
...
+      status: true,
+    },
+    where: {
+      userId: toUserId,
+      toUserId: ctx.user.id,
+    },
+  });
+
+  if (existingBookingForwarding) {
+    throw new Error("booking_forwarding_infinite_not_allowed");
```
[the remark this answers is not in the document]
CarinaWolli: We should make sure to throw an error when there is already a forwarding at that time as two forwarding at the same time shouldn't be possible

packages/emails/src/templates/AcceptBookingForwardingEmail.tsx
```diff
@@ -0,0 +1,55 @@
...
+        })}
+        {props.dates}
+        <br />
+        <div
+          style={{
+            display: "flex",
+            justifyContent: "center",
+            marginTop: "16px",
+          }}>
+          <a href={props.acceptLink} target="_blank">
```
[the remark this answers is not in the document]
CarinaWolli: Buttons still have the wrong format, we should use the existing component
Screenshot 2023-12-18 at 4 34 47 PM

CarinaWolli (dismissed):
There are still some open comments to address 🙏

apps/web/pages/settings/my-account/out-of-office/index.tsx
alannnc: I think for this I will go for the .module.css route.
https://nextjs.org/docs/pages/building-your-application/styling/css-modules#example

alannnc:
@CarinaWolli 
Using current style for buttons ✅ 
image

For the existingBookingForwarding error, I think I have added to the query 2 scenarios that should prevent collisions or overlap. 

```
OR: [
        // Outside of range
        {
          AND: [
            { start: { lte: inputEndTime.toISOString() } },
            { end: { gte: inputStartTime.toISOString() } },
          ],
        },
        // Inside of range
        {
          AND: [
            { start: { gte: inputStartTime.toISOString() } },
            { end: { lte: inputEndTime.toISOString() } },
          ],
        },
      ],
```
This should prevent forwarding with the same dates and also avoid existing forwarding outside/inside of date range from the input.

PeerRich (commented):
is it also possible to go OOO without a redirect?

alannnc:
> is it also possible to go OOO without a redirect?

That would be just going away currently functionality. But it's on the new page as well.

PeerRich:
Couple things: 

- I would entirely remove the "Set as away" functionality and replace it and link to the OOO page
- OOO always needs start and end
- A forwarder is optional (requires team plan)
- Use the `<UpgradeTeamsBadge>` button and link to /teams


we don't need this page:

CleanShot 2024-01-03 at 17 20 01@2x

PeerRich (dismissed):
looks great! the rest we will do in a follow up PR 🙏

PeerRich (dismissed):
great work 🙏

CarinaWolli (approved):
Awesome work @alannnc and great new feature 🙌🏻

PR #12672, opened 2023-12-04
feat: validate `CalProvider` api keys

Ryukemeister:
Validates api keys in the `CalProvider` component before sending it down to child components


packages/atoms/cal-provider/index.tsx
```diff
@@ -1,15 +1,45 @@
...
-import { createContext, useContext } from "react";
+import { createContext, useContext, useState, useCallback, useEffect } from "react";
 
 type CalProviderProps = {
   apiKey: string;
   children: ReactNode;
 };
 
-const ApiKeyContext = createContext("");
+const ApiKeyContext = createContext({ key: "", error: "" });
```
Ryukemeister: Created two different states, one for api key values in case of error messages based on which every atom can do proper error handling

packages/atoms/cal-provider/index.tsx
```diff
@@ -1,15 +1,45 @@
...
+      try {
+        // here we'll call the /me endpoint in v2 to get user profile
+        const response = await fetch(`/v2/me?apiKey=${key}`);
+
+        if (response.ok) {
+          setKey(apiKey);
+        }
+      } catch (error) {
+        console.error(error);
+        setErrorMessage("invalid api key");
```
supalarry: It's a good practice to group error messages in constants or enums. Given that we won't think about i18n now it would be good if you create a file `errors.ts` or alike to store constants:
```
const INVALID_API_KEY = "invalid api key"
```

and do the same for the other error. Then, import them and use in your code. Having them as constants helps with:
1. If you want to change error message then you need to change it in 1 place
2. Makes it easier to test - in tests when you write negative (tests with errors) then you can re-use the constant for expected error.

supalarry (changes requested):
Looks good but last note is on grouping error messages - see comment above. We could also do it in an enum but for sake of simplicity we can also do simple constants.

supalarry (approved):

PR #12677, opened 2023-12-04
chore: create authenticated user guard

ThyMinimalDev:
## What does this PR do?

1. Create custom auth [@GetUser](https://github.com/calcom/cal.com/pull/12677/files#diff-87102523513e8ee02c650f87a9710ac822c97fc2b9746a6b2c54f33824a0b47a) decorator that can be used in endpoint parameters to extract user from the request and also specific properties. [Example usage](https://github.com/calcom/cal.com/pull/12677/files#diff-8f38f05a67bc098c7af467f4663afd74cb81a6cccd9ee5b4d20b6a7de45efb36R32).
2. Create custom guard [NextAuthGuard](https://github.com/calcom/cal.com/pull/12677/files#diff-f7b9cb0256412d2fb4264ec74139a083d13498a710c4152bbbca0fb753faec36) that links to [NextAuthStrategy](https://github.com/calcom/cal.com/pull/12677/files#diff-1e2d7dfe789e019c89926dc04f51a985554488e2509b4529c9d3c0c59b8a7f4a). The custom guard was created for more reusability - instead of hardcoding AuthGuard("next-auth") we can now just re-use NextAuthGuard in UseGuards.
3. NextAuthStrategy has `authenticate` method that uses `getToken` from the `Cookie` of the request. Given a valid user cookie, NextAuthStrategy `authenticate` returns the associated user and that leads to it being attached to the Request object. Then, the custom `@GetUser` decorator is used to extract `id` of the user, which is then used in `createOAuthClient` and `getOAuthClients` endpoints of the `OAuthClientController`.

## Type of change


ThyMinimalDev:
@exception @supalarry please double check, not familiar with nest.js but basically this works to get the user inside response.locals using a guard

apps/api/v2/src/modules/repositories/user/user-repository.service.ts
```diff
@@ -12,4 +13,11 @@ export class UserRepository {
       },
     });
   }
+  async findByEmail(email: string) {
+    return this.dbRead.prisma.user.findUnique({
+      where: {
+        email,
+      },
+    });
+  }
```
ThyMinimalDev, line 22: how do we make sure with nest.js to not expose the password ?

apps/api/v2/src/modules/oauth/oauth-client.controller.ts
```diff
@@ -47,7 +51,10 @@ export class OAuthClientController {
 
   @Get("/")
   @HttpCode(HttpStatus.OK)
-  async getOAuthClients(@Res({ passthrough: true }) res: Response) {
+  @UseGuards(NextAuthGuard)
```
supalarry, line 54: We can move `@UseGuards(NextAuthGuard)` below the
```
@Controller({
  path: "oauth-clients",
  version: "2",
})
@UseGuards(NextAuthGuard)
```

to apply it to all routes.

apps/api/v2/src/types/index.ts
```diff
@@ -7,6 +7,9 @@ export interface PlatformApiLocals extends Record<string, unknown> {
     userId: number;
     teamId: number;
   };
+  user?: {
```
supalarry, line 10: I am thinkign that we need to split `PlatformApiLocals` into separate interfaces and then also have different `type Response` because some routes, like oAuth` will not have `apiKey` and vice versa and if we keep adding to `locals` then it will be hard to make sense of it.

What are your thoughts?

ThyMinimalDev: yes, let's come up with a discriminative union

apps/api/v2/src/modules/repositories/user/user-repository.service.ts
sean-brydon: We could use middleware/priosma extentions for this 
``` 
async function excludePasswordMiddleware(params, next) {
  const result = await next(params)
  if (params?.model === 'User' && params?.args?.select?.password !== true) {
    delete result.password
  }
  return result
}

const prisma = new PrismaClient()
``` 

SRC: https://github.com/prisma/prisma-client-extensions/blob/main/obfuscated-fields/script.ts && some random stack overflow i cant find anymore

ThyMinimalDev: nice one 👍

apps/api/v2/src/modules/oauth/oauth-client.controller.ts
```diff
@@ -13,25 +14,22 @@ import {
...
   constructor(private readonly oauthClientRepository: OAuthClientRepository) {}
 
   @Post("/")
   @HttpCode(HttpStatus.CREATED)
-  async createOAuthClient(
-    @Res({ passthrough: true }) res: Response,
-    @Body() createOAuthClientDto: CreateOAuthClientDto
-  ) {
-    const userId = res.locals.apiKey?.userId;
+  async createOAuthClient(@GetUser("id") userId: number, @Body() createOAuthClientDto: CreateOAuthClientDto) {
```
ThyMinimalDev, line 32: @exception if we used the same approach for the apiKey, would we still need to set the apiKey in response.locals in ApiKeyService ?

it seems like we can completely avoid using response.locals following the approach in this pr?

apps/api/v2/src/modules/oauth/oauth-client.controller.ts
```diff
@@ -47,8 +45,7 @@ export class OAuthClientController {
 
   @Get("/")
   @HttpCode(HttpStatus.OK)
-  async getOAuthClients(@Res({ passthrough: true }) res: Response) {
-    const userId = res.locals.apiKey?.userId;
+  async getOAuthClients(@GetUser("id") userId: number) {
     return this.oauthClientRepository.getUserOAuthClients(userId);
```
ThyMinimalDev, line 49: @exception where can we define in nest.js a generic response handler ? making sure that we always return some thing to the user in the same format ?

like 

```ts
{
status: "OK",
data: []
}
```

{
status: "error",
error: ...
}
```

File: apps/api/v2/src/modules/oauth/oauth-client.controller.ts (hunk 4 above)
Comment (exception) [reply apps/api/v2/src/modules/oauth/oauth-client.controller.ts:32]: Yes that should be fine!

File: apps/api/v2/src/modules/repositories/user/user-repository.service.ts (hunk 1 above)
Comment (supalarry) [reply apps/api/v2/src/modules/repositories/user/user-repository.service.ts:22]: Great idea Sean! Implemented in commit [exclude password from prisma read service](https://github.com/calcom/cal.com/pull/12677/commits/24b1f8ed59f6d99f0ecc589c691efb0ac1f2ac3b)

Comment (supalarry) [reply apps/api/v2/src/modules/repositories/user/user-repository.service.ts:22]: Oof, the code is not working because if we have following query:
```
async findByEmail(email: string) {
    return this.dbRead.prisma.user.findUnique({
      where: {
        email,
      },
    });
  }
  ```
  
  then `select` of it is empty `select: {}` aka select all, but if we add `password: false` to it, then it becomes {password: false` and in prisma queries at least 1 key needs to be as true aka it can't be just false.

supalarry: PS Sean recommended to use prisma client extension to obfuscate the field, but its docs say `This extension is provided as an example only. It is not intended to be used in production environments.`

supalarry: Morgan's implementation: [fix: userRepo sanitize password](https://github.com/calcom/cal.com/pull/12677/commits/1f286e99e2c8e1aca4d9f74c5615a08b47fa7ce8)

sean-brydon: Awesome job @ThyMinimalDev 

```
import { PrismaReadService } from "@/modules/prisma/prisma-read.service";
import { Injectable } from "@nestjs/common";
import { User } from "@prisma/client";

// TODO make sure we dont expose sensitive data like password
@Injectable()
export class UserRepository {
  private readonly sensitiveFields: (keyof User)[] = ["password"];

  constructor(private readonly dbRead: PrismaReadService) {}

  private sanitize(user: User): Partial<User> {
    return Object.fromEntries(
      Object.entries(user).filter(([key]) => !this.sensitiveFields.includes(key as keyof User))
    );
  }

  async findById(userId: number) {
    const user = await this.dbRead.prisma.user.findUnique({
      where: {
        id: userId,
      },
    });
    return this.sanitize(user);
  }

  async findByEmail(email: string) {
    const user = await this.dbRead.prisma.user.findUnique({
      where: {
        email,
      },
    });
    return this.sanitize(user);
  }
}
```

IMHO we should refactor this as above and even consider having this extend a baseReposity that allows us to just define sensitiveKeys. Let me know if you want me to push these changes

ThyMinimalDev: extending a base repo exposing functions like sanitize makes sense I think, let's do it

apps/api/v2/src/modules/auth/strategy/next-auth/index.ts
```diff
@@ -0,0 +1,38 @@
...
+      const user = await this.userRepository.findByEmail(payload.email);
+      if (!user) {
+        throw new UnauthorizedException();
+      }
+
+      return this.success(user);
+    } catch (error) {
+      if (error instanceof Error) return this.error(error);
+    }
+  }
```
ThyMinimalDev, line 37: do we need to try catch anywhere if we use global exception filters ? @exception @supalarry

apps/api/v2/src/app.ts
```diff
@@ -17,10 +19,25 @@ export const bootstrap = (app: NestExpressApplication): NestExpressApplication =
...
+      },
+    })
+  );
+
   app.useGlobalFilters(new PrismaExceptionFilter());
   app.setGlobalPrefix("api", {
     exclude: [{ path: "health", method: RequestMethod.GET }],
   });
 
+  app.use(cookieParser());
```
exception, line 41: `suggestion-non-blocking`: Could consider adding `helmet`?

apps/api/v2/src/config/app.ts
```diff
@@ -1,16 +1,21 @@
+import { getEnv } from "@/env";
+
 import type { AppConfig } from "./type";
 
 const loadConfig = (): AppConfig => {
   return {
     env: {
-      type: (process.env.NODE_ENV as "production" | "development") ?? "development",
+      type: getEnv("NODE_ENV") ?? "development",
```
exception, line 8: `suggestion-non-blocking`: Fallback could be part of the `getEnv` function. eg: `getEnv("NODE_ENV", "development")`

apps/api/v2/src/modules/api-key/api-key.service.ts
```diff
@@ -1,38 +1,27 @@
...
     const apiKey = request.get("Authorization")?.replace("Bearer ", "");
+
+    if (!apiKey) {
+      return null;
+    }
+
     const hashedKey = hashAPIKey(apiKey.replace("cal_", ""));
 
-    const apiKeyResult = await this.dbRead.prisma.apiKey.findUnique({
+    const apiKeyResult = await this.dbRead.prisma.apiKey.findUniqueOrThrow({
```
exception, line 19: `nitpick`: Return directly, no need to await.

packages/prisma/schema.prisma
```diff
@@ -318,6 +318,7 @@ model Team {
   apiKeys             ApiKey[]
   credentials         Credential[]
   accessCodes         AccessCode[]
+  PlatformOAuthClient PlatformOAuthClient[]
```
exception, line 321: `nitpick`: Naming.

exception (changes requested):

apps/api/v2/src/config/app.ts
supalarry: Great idea! Done in [feat: fallback value for getEnv](https://github.com/calcom/cal.com/pull/12677/commits/8a0916668a705df2943231dce39d3713a345a26b)

apps/api/v2/src/app.ts
supalarry: Good idea, done in [feat: add helmet for security](https://github.com/calcom/cal.com/pull/12677/commits/57731a6ae07730394846da05b12632b64612245f)

apps/api/v2/src/modules/oauth/oauth-client.controller.ts
```diff
@@ -13,65 +16,77 @@ import {
...
+import { MembershipRole, PlatformOAuthClient } from "@prisma/client";
+
+import { SUCCESS_STATUS } from "@calcom/platform-constants";
+import { ApiResponse } from "@calcom/platform-types";
 
 @Controller({
   path: "oauth-clients",
   version: "2",
 })
+@UseGuards(NextAuthGuard, OrganizationRolesGuard)
```
ThyMinimalDev, line 30: doesn't this mean that the organizationRolesGuard is applied to all the endpoints ?

supalarry: It is used together with "@Roles" decorator applied to each endpoint e.g. create has `@Roles([MembershipRole.ADMIN, MembershipRole.OWNER])` but read does not, so then the roles are passed to the `OrganizationsRolesGuard`, which checks if any roles have been passed to Roles guard or if it has been applied, and if none of them is true it just returns `true` meaning the request passes through.

PR #12673, opened 2023-12-04
chore: upgrade storybook 7.6.3 + use @storybook/nextjs

ThyMinimalDev:
## What does this PR do?

Upgrade storybook to latest version + use  @storybook/nextjs + fix breaking changes

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- run storybook in local

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #12628, opened 2023-12-01
fix: upgrade Storybook et. al. to enable Railway build

henrikklee:
## What does this PR do?

This PR includes the changes I had to make in my quest to build and deploy Cal.com on Railway.

Most notably:
- Upgrade `apps/storybook` to Storybook 7 (a major breaking change)
- Fix a plethora of stories in `packages/ui`, `packages/atoms`, and `packages/features`
- Upgrade `packages/atoms` to Vite 5
- Fix some typing issues int `apps/swagger`

Overall, this turned out to be quite the ride, taking A LOT of time to get right. But with the implemented changes the Railway build and deployment work flawlessly. The environment variables from `apps/ai` are missing in the Railway template, though.

## Storybook Upgrade

Many of the stories in Storybook are still not perfect. However, if you take a look at https://ui.cal.com, you will find that many stories aren't perfect, or even rendering, there, as well. Overall, the stories in this PR work better or equally well as those currently deployed.

A major change from Storybook 6 to Storybook 7 is the new separation of stories and documentation. Previously, stories and docs would be defined together in one `.mdx` file. With SB7, stories are defined in `.stories.tsx` files and the corresponding docs are defined in a `.mdx` file. For the sake of clarity I chose the suffix `.docs.mdx`.

While the old format of defining stories within `.mdx` files is still supported, it's marked as deprecated. I migrated some of the stories to the new format, to fix some issues, and for reference.

The app router (`useRouter` from `next/navigation`) is supported natively with Storybook 7. However, it didn't work for me with Next 14, so the dependency is set to the latest version I found to work, `^13.5.6`.

## Vite Upgrade

When building and deploying on/to Railway, Vite wasn't able to parse the tsx files of `packages/atoms` correctly, especially `ogImages.tsx` from `packages/lib`. Upgrading to Vite 5 fixed this issue.

## Radix UI Tooltip Upgrade

I had some issues with Radix UI Tooltip: `Tooltip must be inside TooltipProvider`, despite `AppLayout` containing TooltipProvider. Upgrading to `^1.0.7` fixed this for me.

## Updates to `apps/swagger`

Fixed some unmaintained types

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- Test visually, especially Storybook, and compare to https://ui.cal.com
- Run automated tests (passing for me)

### Please describe the tests that you ran to verify your changes

[...]

PR #12681, opened 2023-12-05
fix: types in apps/swagger (fixes #10023) and chore: upgrade Vite to v5 in packages/atoms (fixes Railway build)

henrikklee:
## What does this PR do?

Originally part of https://github.com/calcom/cal.com/pull/12628

These changes are necessary to fix builds on Railway. 

- Upgrade `packages/atoms` to Vite 5
- Fix some typing issues in `apps/swagger`

The original Vite 4 config and swagger types produced build errors on Railway.

This should also fix https://github.com/calcom/cal.com/issues/10023

## Notes

- Renamed snippets.ts to snippets.js as it doesn't contain the necessary types which leads to build errors on Railway. The types are not available.
- Removed two non-existing props from index.tsx
- Crypto alias apparently not necessary anymore (nor working)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Chore (refactoring code, technical debt, workflow improvements)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

PR #12683, opened 2023-12-05
fix: fixes multiple storybook stories and updates to the new sb7 format

henrikklee:
## What does this PR do?

Replacement of https://github.com/calcom/cal.com/pull/12628 to account for updates made in https://github.com/calcom/cal.com/pull/12673 by @ThyMinimalDev (cc @emrysal)

> A major change from Storybook 6 to Storybook 7 is the new separation of stories and documentation. Previously, stories and docs would be defined together in one `.mdx` file. With SB7, stories are defined in `*.stories.tsx` files and the corresponding docs are defined in a `.mdx` file. For the sake of clarity I chose the suffix `.docs.mdx`.

- Converts the following stories to the new format (splits component into .stories.tsx and .docs.mdx)
  - Tooltip
  - Dialog
  - Booker
  - EventMeta
  - VerifyCodeDialog
- Fixes the following stories
  - Badge (typo)
  - Breadcrumb (missing import)
  - Vertical Tabs (wrong import)
  - Color picker (redundant import `TooltipProvider`, see above)
  - Wizard story (unused import `useRouter`)
- Adds TooltipProvider and StorybookTrpcProvider as global decorators to make them wrap all stories automatically
  - StorybookTrpcProvider mocks TRPC, esp. for the Booker
- Migrates main.js and preview.jsx to the new typescript config format
- Excludes storybook cover image and favicon from gitignore
- Removes unnecessary GFM (Github flavored markdown) plugin
- Comments out lazy build option (didn't work for me)
- Fixes color picker test (await)

## Type of change


packages/ui/components/form/color-picker/colorpicker.test.tsx
```diff
@@ -48,7 +48,7 @@ describe("Tests for ColorPicker component", () => {
     const colorInput = screen.getByRole("textbox");
     await act(async () => userEvent.clear(colorInput));
     const newColorValue = "#00FF00";
-    await act(async () => userEvent.type(colorInput, newColorValue));
+    await act(async () => await userEvent.type(colorInput, newColorValue));
```
ThyMinimalDev, line 51: this seems like a recurring issue, this test is quite flaky, do we event need act ?

apps/storybook/.storybook/preview.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    nextjs: {
+      appDirectory: true,
+    },
+  },
+
+  decorators: [
+    (Story) => (
+      <StorybookTrpcProvider>
+        <TooltipProvider>
```
ThyMinimalDev, line 44: awesome 🔥

packages/ui/components/form/color-picker/colorpicker.test.tsx
stabildev: I had an issue with this in my previous PR https://github.com/calcom/cal.com/pull/12628 where I updated many more packages, including Vite from 4 to 5 in `atoms`, which I now moved to https://github.com/calcom/cal.com/pull/12681 . I added this fix for completeness as I think this will pop up again. `await` is the correct way to handle this

apps/storybook/.storybook/main.ts
```diff
@@ -85,6 +89,8 @@ module.exports = {
   },
```
ThyMinimalDev, line 89: what's you opinion on autodocs ?

stabildev: IMO not needed for Cal.com since all stories have an explicit `mdx`

stabildev: I didn't see any difference when adding or omitting this option so I don't think it matters.

What was your reasoning behind lazy builds?

packages/ui/components/form/color-picker/colorpicker.test.tsx
ThyMinimalDev: @testing-library doc is quite explicit that act should not be necessary, need to double check why we need to chain await like that

ThyMinimalDev (approved):
great work

PR #12684, opened 2023-12-05
fix: Edit Location modal fields truncates on mobile view

### LINKED ISSUE #12640: Edit Location modal truncates fields/cannot scroll on mobile
### Issue Summary

When I receive a new booking and try to edit the location on my phone, I can choose a Meeting URL but then the successive field for Meeting URL is truncated.

It's visible once I switch to landscape mode, but that also seems quite broken since the previous fields aren't visible.

Screenshot at bottom

I did see #5987 and #3476 but those seem resolved already, and am unable to find others that are the same issue

### Steps to Reproduce

1. On a mobile device with a small screen
    - I'm using a Sony Xperia XZ2 Compact, resolution 1080x2160
    - There's a similar issue in 2160x1080 but it's different in that the fields pop out at the bottom
3. Navigate to an existing listing in https://app.cal.com/bookings/upcoming
4. Click on "Edit -> Edit Location"
5. Select "Link meeting"
6. See that new "Provide a Meeting Link" field is not visible and cannot be scrolled to

### Actual Results

- Field is getting cut off and not accessible via scrolling

### Expected Results

- All fields should be visible or scrollable

### Technical details

- Browser version: Chrome on Android, 119.0.6045.163
- Node.js version: N/A, using https://app.cal.com/

### Evidence

Screenshots:

Screenshot_20231202-134150

Screenshot_20231202-134211

ujjwal2505:
fixes #12640


apps/web/components/dialog/EditLocationDialog.tsx
```diff
@@ -382,7 +382,7 @@ export const EditLocationDialog = (props: ISetLocationDialog) => {
                 }}
               />
               {selectedLocation && LocationOptions}
-              <DialogFooter className="mt-4">
+              <DialogFooter className="!static mt-4">
```
Udit-takkar, line 385: is there any other way we can fix the error? this looks like a hacky solution

ujjwal2505: We can add overflow: auto property on the form tag which can help us fix the button overflow on the text field. This way we wont need to pass !static on the DialogFooter component.

Screenshot 2023-12-06 175640

apps/web/components/dialog/EditLocationDialog.tsx
```diff
@@ -382,7 +382,7 @@ export const EditLocationDialog = (props: ISetLocationDialog) => {
                 }}
               />
               {selectedLocation && LocationOptions}
-              <DialogFooter className="mt-4">
+              <DialogFooter className="relative">
```
Udit-takkar, line 385: @ujjwal2505 this seem to be working. can you pull the changes and test it once

ujjwal2505: I tested the code and it is working. We can close the issue.

Udit-takkar (approved):

PR #12693, opened 2023-12-06
fix: '/slots' endpoint API fixes

alishaz-polymath:
## What does this PR do?

Fixes the slots endpoint issues:
- [x] usernameList requiring more than 1 usernameList entries (event for single user) and thus ignoring the eventTypeSlug and instead picking dynamic event type all the time
- [x] timeZone handling to return timeZone specific slots instead of slots in UTC which is currently happening. We need to do this on the API front alone so as to not disrupt the slot fetching flow in the UI.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/api/pages/api/slots/_get.ts
+++ b/apps/api/pages/api/slots/_get.ts
@@ -18,6 +18,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     }
     const input = getScheduleSchema.parse({ usernameList: slugs, ...rest });
     const availableSlots = await getAvailableSlots({ ctx: await createContext({ req, res }), input });
+    // TODO: use input.timeZone to convert the availableSlots into timeZone specific slots
     return availableSlots;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   } catch (cause) {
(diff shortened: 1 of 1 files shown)


apps/api/pages/api/slots/_get.ts
```diff
@@ -9,10 +13,34 @@ import { getAvailableSlots } from "@calcom/trpc/server/routers/viewer/slots/util
...
+
 async function handler(req: NextApiRequest, res: NextApiResponse) {
   try {
-    const input = getScheduleSchema.parse(req.query);
-    return await getAvailableSlots({ ctx: await createContext({ req, res }), input });
+    const { usernameList, ...rest } = req.query;
+    let slugs = usernameList;
+    if (!Array.isArray(usernameList)) {
+      slugs = usernameList ? [usernameList] : [];
+    }
```
alishaz-polymath, line 26: handles usernameList when only one user slug is provided. It was essential otherwise to provide an extra usernameList query parameter (empty) to satisfy the validation, as well as then had a side-effect of it being considered as a dynamic event

zomars (approved):
Nice find @alishaz-polymath 🙏🏽

PR #12696, opened 2023-12-06
feat: hide password validation no password

sean-brydon:
Adds an option to hide hint errors 
CleanShot 2023-12-07 at 07 45 42

### DIFF
--- a/packages/ui/components/form/inputs/HintOrErrors.tsx
+++ b/packages/ui/components/form/inputs/HintOrErrors.tsx
@@ -3,7 +3,7 @@ import { useFormContext } from "react-hook-form";
 
 import { classNames } from "@calcom/lib";
 
-import { Check, Circle, Info, X } from "../../icon";
+import { Check, Circle, X } from "../../icon";
 
 type hintsOrErrorsProps = {
   hintErrors?: string[];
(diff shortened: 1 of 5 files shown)


packages/ui/components/form/inputs/HintOrErrors.tsx
```diff
@@ -1,106 +1,95 @@
...
   fieldName,
   t,
+  hideIfEmpty = false,
 }: hintsOrErrorsProps) {
   const methods = useFormContext() as ReturnType<typeof useFormContext> | null;
   /* If there's no methods it means we're using these components outside a React Hook Form context */
   if (!methods) return null;
-  const { formState } = methods;
+
+  if (!methods && hintErrors?.length) {
```
hariombalhara, line 46: methods can't be falsy here.

packages/ui/components/form/inputs/HintOrErrors.tsx
```diff
@@ -1,106 +1,95 @@
...
+    throw new Error("InputField is expecting hint errors but is not use within a RHF context");
+  }
+
+  const { formState, watch } = methods;
+  const hasBeenTouched = watch(fieldName);
```
hariombalhara, line 51: I think it's not identifying if the input has been changed/touched. It is just returning the value.

hariombalhara (changes requested):
Left some comments.

PR #12704, opened 2023-12-07
fix: rr-host booked outside of availability

CarinaWolli:
## What does this PR do?

Fixes the follow issue: 

I have two rr-hosts (teampro, teamfree). Got these two personal bookings for teampro in the db (see screenshot, first two bookings). If teamfree doesn't have any availability, the 10AM slot on dez 12 (Melbourne) is not showing (that's correct). If teamfree is available the slot shows (as it should) BUT if it is teampro's round robin turn, it books teampro

Screenshot 2023-12-07 at 5 06 53 PM

The issue was the getBusyTimes didn't return the bookings because we were filtering the bookings like that: 
{
   "startTime":{"gte":"2023-12-11T23:00:00.000Z"},
   "endTime":{"lte":"2023-12-11T23:15:00.000Z"},
}

## Type of change


packages/core/getBusyTimes.ts
```diff
@@ -256,6 +257,8 @@ export async function getBusyTimes(params: {
       allBusyTimes: busyTimes,
     })
   );
+  console.log(`user ${username}`);
+  console.log(`busy times ${JSON.stringify(busyTimes)}`);
```
alishaz-polymath, line 261: Do we want to keep console logs here?

alishaz-polymath (dismissed):
LGTM, waiting on Alex to review before merging in. His context knowledge is essential here.

packages/core/getBusyTimes.ts
```diff
@@ -83,9 +83,10 @@ export async function getBusyTimes(params: {
   const endTimeDate =
     rescheduleUid && duration ? dayjs(endTime).add(duration, "minute").toDate() : new Date(endTime);
 
+  // startTime is less than endTimeDate and endTime grater than startTimeDate
   const sharedQuery = {
-    startTime: { gte: startTimeDate },
-    endTime: { lte: endTimeDate },
+    startTime: { lte: endTimeDate },
```
emrysal, line 88: Watch out, this isn't the only place this happens, https://github.com/calcom/cal.com/blob/969411041b75833ad6625d341a9d0b7ea535de43/packages/trpc/server/routers/viewer/slots/util.ts#L330

But such a good spot 💯

emrysal (changes requested):

packages/core/getBusyTimes.ts
CarinaWolli: Thank you, fixed it there as well 🙏

emrysal (dismissed):
Good to go 🔥

emrysal (approved):

PR #12709, opened 2023-12-08
feat: add utcOffset in webhook payload

### LINKED ISSUE #12415: [CAL-2904] Request to add 2 Fields to Webhook Payloads

jkcs:
## What does this PR do?

Fixes #12415
/claim #12415

## Requirement/Documentation

## Type of change

## Mandatory Tasks


packages/lib/date-fns/index.ts
```diff
@@ -226,3 +226,18 @@ export const isInDST = (date: Dayjs) => {
...
+/**
+ * Get UTC offset of given time zone
+ * @param timeZone Time Zone Name (Ex. America/Mazatlan)
+ * @returns
+ */
+export function getUTCOffsetByTimezone(timeZone: string) {
+  if (!timeZone) return null;
+
+  if (timeZoneWithDST(timeZone)) {
+    return getUTCOffsetInDST(timeZone);
```
CarinaWolli, line 239: Wouldn't this now always return DST offset of this timezone, even when we are not in DST like now in December?

CarinaWolli (changes requested):
Thank you for contributing 🙏 Left you one comment, everything else looks good

packages/lib/date-fns/index.ts
jkcs: Wow, you reminded me so well. I was overthinking it. This is not necessary.

CarinaWolli (changes requested):

packages/lib/date-fns/index.ts
```diff
@@ -226,3 +226,14 @@ export const isInDST = (date: Dayjs) => {
...
+
+/**
+ * Get UTC offset of given time zone
+ * @param timeZone Time Zone Name (Ex. America/Mazatlan)
+ * @returns
+ */
+export function getUTCOffsetByTimezone(timeZone: string) {
+  if (!timeZone) return null;
+
+  return dayjs().tz(timeZone).utcOffset();
```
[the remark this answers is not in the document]
jkcs: done

Shpadoinkle:
Is there a way to also include the timeZone that was used on the bookings page on the main webhook body?   ie 
image

If the person who booked was outside the timezone of the eventType organiser, then the organisers timezone isn't relevant.  

In our case our booking pages are public so don't have "attendees" on the webhook responses.    So none of the additions here would be useful.    
Is it more useful passing the timezone string itself instead of a number offset? ie- 'America/New_York'

jkcs:
Hello, @Shpadoinkle . I suggest that you open a new issue. The changes in this pull request only include the modification mentioned in #12415 "Add the attendee UTC offset as a field, rather than the string of the timezone." As for the issue you mentioned, I think it's worth addressing as well.

Shpadoinkle:
@jkcs ah fair enough, didnt notice that detail in the write up. 
All good, i have raised other request this with main team

packages/features/webhooks/lib/sendPayload.ts
```diff
@@ -32,14 +45,31 @@ export type WebhookDataType = CalendarEvent &
     paymentId?: number;
   };
 
+function addUTCOffset(
+  data: Omit<WebhookDataType, "createdAt" | "triggerEvent">
+): WithUTCOffsetType<WebhookDataType> {
+  if (data.organizer?.timeZone) {
+    (data.organizer as Person & UTCOffset).utcOffset = getUTCOffsetByTimezone(data.organizer.timeZone);
```
CarinaWolli, line 52: we are missing the `date` parameter here (`data.startTime`), otherwise, we will always get the utc offset from the current date (important for DST)

packages/features/webhooks/lib/sendPayload.ts
```diff
@@ -32,14 +45,31 @@ export type WebhookDataType = CalendarEvent &
...
+  }
+
+  if (data?.attendees?.length) {
+    (data.attendees as (Person & UTCOffset)[]).forEach((attendee) => {
+      attendee.utcOffset = getUTCOffsetByTimezone(attendee.timeZone);
```
CarinaWolli, line 57: same here

CarinaWolli (changes requested):

CarinaWolli (approved):
Looks good now 👍 Thank you for your contribution 🙏

PR #12723, opened 2023-12-11
fix: Hide expires on message if never expires is selected

patrick91:
I don't think it makes sense to show "This will expire on ..." when selecting "Never expires":

CleanShot 2023-12-11 at 11 21 54@2x

Another option would be to change the message, happy to do that if you prefer 😊


emrysal (approved):
Agree! Thanks

PR #12726, opened 2023-12-11
chore: Platform rate limiting

exception:
## What does this PR do?

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/api/v2/src/modules/api-key/api-key.service.ts
+++ b/apps/api/v2/src/modules/api-key/api-key.service.ts
@@ -19,6 +19,11 @@ export class ApiKeyService {
     response.locals.apiKey = key;
   };
 
+  getApiKeyFromRequest(request: Request): string | null {
+    const apiKey = request.get("Authorization")?.replace("Bearer ", "");
+    return apiKey;
+  }
+
   async retrieveApiKey(request: Request, response?: Response) {
     const apiKey = request.get("Authorization")?.replace("Bearer ", "");
     const hashedKey = hashAPIKey(apiKey.replace("cal_", ""));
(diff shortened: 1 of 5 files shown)


PeerRich (approved):
just rate limiting right? looks good

apps/api/v2/src/modules/api-key/api-key.service.ts
```diff
@@ -19,6 +19,11 @@ export class ApiKeyService {
     response.locals.apiKey = key;
   };
 
+  getApiKeyFromRequest(request: Request): string | null {
+    const apiKey = request.get("Authorization")?.replace("Bearer ", "");
+    return apiKey;
+  }
+
   async retrieveApiKey(request: Request, response?: Response) {
```
supalarry, line 27: Since we now have `getApiKeyFromRequest`, it sounds similar to the `retrieveApiKey` below. Probably best to rename `retrieveApiKey` -> `retrieveCalApiKey` so that there is no confusion with the newly added `getApiKeyFromRequest`.

supalarry (changes requested):
Since "AppConfig" in "apps/api/v2/src/config/type.ts" is updated we also need to update "loadConfig" in "apps/api/v2/src/config/app.ts"

sean-brydon (commented):
Is there any reason we are not using `@nestjs/throttler` from nestjs? Is this just a more flexible approach? 

We used this at accenture and just created a ThrottlerProxy to handle the edgecases for ip/api keys etc etc

PR #12134, opened 2023-10-29
fix: change booking page filter ui to match Figma

asdotdev:
## What does this PR do?

Fixes #11586

[Loom Link](https://www.loom.com/share/f9dd91f18dba4a15ab0fbb64dd314cf8?sid=4e4ec04e-c2b0-4509-acce-23255442da1b)

## Requirement/Documentation

[Figma Link](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/New-Features?type=design&node-id=7869%3A93929&mode=dev)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

Open the booking page and match the UI with Figma.

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent-sized PR without self-review might be rejected.

## Checklist

- I haven't added tests that prove my fix is effective or that my feature works

## Note

I have added "filters" in common.json for only en
```diff
+ "filters": "Filters",
```

PR #12742, opened 2023-12-12
fix: checkbox color

Udit-takkar:
Regression from https://github.com/calcom/cal.com/pull/12134

After:-

Screenshot 2023-12-12 at 5 22 12 PM

Screenshot 2023-12-12 at 5 22 24 PM


CarinaWolli (commented):
Is this fix about it being blue? Because this is happening for all checkboxes:
Screenshot 2023-12-12 at 3 53 01 PM

sean-brydon:
> Is this fix about it being blue? Because this is happening for all checkboxes: Screenshot 2023-12-12 at 3 53 01 PM

Doesnt look like it - it corrects it in the teams filter as we had to use the none component version IIRC. I will fix that blue issue in another PR

PeerRich (approved):

Issue #12688, opened 2023-12-06
Impersonation and Login on users with uppercase email addresses bug


Shpadoinkle:
#### UPDATE
- not an API only issue.
  - the front end dashboard also doesnt enforce or expect lower case email addresses, emails are saved to DB with capitals.

#### Impersonation Issue
- Saw an issue when impersonating a user,  every page returns a generic unauthorized response. 
- Turns out again it was due to the impersonated users email address containing an uppercase letter.
- Changing this users email address in the db directly solved the issue,  could not access their own profile to do so.

#### Login issue
- Users that existing in the database with with Captial letters in their email address
  - Login does not work when providing email in either lowercase or uppercase format. 
  - logic on this input to the database does not work

### Steps to Reproduce

1. Create a user via the API OR front end with capital letters in email address
2. Try to login or impersonate user,  dashboard will return unauthorized errors.

### Actual Results
- a bulk of unauthorized errors.

### Expected Results
- Normal login or impersonation.

### Technical details
v3.5.5

sarthakgarg814:
Please Assign it to me

zomars:
Hey @Shpadoinkle Thanks for reporting this. Just wanted to add our two cents since we stumbled with this before. Also here's a workaround you can use to sanitize your DB data:

```sql
-- Find users with uppercase characters in email
SELECT * FROM users u WHERE u.email ~ '[[:upper:]]';

-- Lowercase all user emails
UPDATE users SET email = lower(email) WHERE email ~ '[[:upper:]]';
```

I recommend running the first query first to see how many users are affected. Then run the UPDATE query to lowercase them all.

We didn't put this in a migration since there could be an edge case where there's conflicting emails like `user@example.com` and `User@example.com` would cause the UPDATE query to fail due to unique constrains.

If you do find a duplicate emails you can query their duplicate with this:

```sql
SELECT * FROM users u WHERE LOWER(u.email)=LOWER('user@example.com');
```

And then decide which one to rename/delete/etc.

Hope this helps 🙏🏽

PR #12743, opened 2023-12-12
fix: email validation in user post method API

TaduJR:
## What does this PR do?

ref #12688

## Type of change


Shpadoinkle:
Noticed this isnt isolated to the API,  as Creating a user through the front end also doesnt lower case the email address in the DB.   so i dont think this is a useful change as Calcom has never insisted on lowercase emails.  

not sure of the best solution to this.  It's weird that it expects a lower case email on the impersonation despite it never enforcing lower case anywhere else. 


maybe its best to change the impersonation logic.  Change how it handles the email checking as who knows how many users exist with capital letters in their email addresses.

Shpadoinkle:
Noticed this isnt isolated to the API @hariombalhara ,   as Creating a user through the front end also doesnt lower case the email address in the DB either.   so i dont think this is a useful change as Calcom has never insisted on lowercase emails.  

not sure of the best solution to this.  It's weird that it expects a lower case email on the impersonation logic despite it never enforcing lower case anywhere else. 


maybe its best to change the impersonation logic.  Change how it handles the email checking as who knows how many users exist with capital letters in their email addresses.

Shpadoinkle:
This PR also wont solve the issue for existing users with Capital email addresses that try to login.    You'll need to compare a lowercased input to a lowercased version of the db entry.

TaduJR:
> This PR also wont solve the issue for existing users with Capital email addresses that try to login. You'll need to change the logic that is comparing a lowercased input to a **_non_**-lowercased version of the db entry.

@Shpadoinkle Note that this issue exists on API version only. So If you created a user via API and tried to login via dashboard it doesn't work because API creation of user is not converting email address to lowercase but not in tRPC backend server.

Shpadoinkle:
> > This PR also wont solve the issue for existing users with Capital email addresses that try to login. You'll need to change the logic that is comparing a lowercased input to a **_non_**-lowercased version of the db entry.
> 
> @Shpadoinkle Note that this issue exists on API version only. So If you created a user via API and tried to login via dashboard it doesn't work because API creation of user is not converting email address to lowercase but not in tRPC backend server.

Has this changed since version 3.5.5?    I just created a user in the front end  and it also saves the user with capital letters to the database.

Shpadoinkle:
> > This PR also wont solve the issue for existing users with Capital email addresses that try to login. You'll need to change the logic that is comparing a lowercased input to a **_non_**-lowercased version of the db entry.
> 
> @Shpadoinkle Note that this issue exists on API version only. So If you created a user via API and tried to login via dashboard it doesn't work because API creation of user is not converting email address to lowercase but not in tRPC backend server.

@TaduJR  Has this changed since version 3.5.5?    I just created a user in the front end  and it also saves the user with capital letters to the database.

TaduJR:
Hmmm. Ok @Shpadoinkle. I will look it to that. Thanks.

Shpadoinkle:
this PR should also handle existing users somehow.  

As lowercasing emails for new users only doesn't solve the login or impersonation issue for existing users.

TaduJR:
@Shpadoinkle I created a user on the frontend using case-sensitive email but it is stored in lower-case email. I think it is fixed on later versions. Also I made my test email case-sensitive and it returned error.

Yes you are right.  The best solution is making lowercase their input behind the scenes to avoid throwing breaking API errors then also change the login/impersonation logic to read a lowercased version of the db entry.

But I tried to checkout Prisma Docs case-sensitive querying. It has some errors and lack of usage documentation in some prisma users report. https://www.prisma.io/docs/orm/prisma-client/queries/case-sensitivity

So the best would be converting all tables who has email attribute to lowercase in the production database.

Then This fix handles the API side by user creation email to lowercase.
And on the Backend I think it is already implemented. Because I created case-sensitive email and the backend stored it as lowercase.

zomars:
Agreed with @Shpadoinkle that this PR helps but if doesn't close #12688  as we need to make sure all user creates/updates are lowercasing emails properly. Merging without closing the issue for now.

zomars (approved):

PR #12757, opened 2023-12-12
fix: datatable toolbar

sean-brydon:
Position of the toolbar for tables was always at the bottom of the table and not fixed to the UI. 

Adds an animation of this opening and closing

Fixes the selection model when you bulk delete users -> We unselect them indexes so you dont select the wrong user 
CleanShot 2023-12-13 at 07 14 34

Fixes: CAL-2795


PeerRich (dismissed):
looks good! havent tested mobile

PeerRich (approved):

PR #12762, opened 2023-12-13
chore: OAuth clients tests and fixes

ThyMinimalDev:
## What does this PR do?

Tests for platform oAuth clients api v2 endpoints

## Type of change

### DIFF
--- a/apps/api/v2/src/app.ts
+++ b/apps/api/v2/src/app.ts
@@ -41,6 +41,9 @@ export const bootstrap = (app: NestExpressApplication): NestExpressApplication =
     })
   );
 
+  const { httpAdapter } = app.get(HttpAdapterHost);
+  app.useGlobalFilters(new SentryFilter(httpAdapter));
+
   app.useGlobalFilters(new PrismaExceptionFilter());
 
   if (process.env.SENTRY_DNS) {
@@ -48,9 +51,7 @@ export const bootstrap = (app: NestExpressApplication): NestExpressApplication =
       dsn: getEnv("SENTRY_DNS"),
     });
   }
-  const { httpAdapter } = app.get(HttpAdapterHost);
   // Sentry Exception Filter only logs if Sentry.init has not been called
-  app.useGlobalFilters(new SentryFilter(httpAdapter));
   app.useGlobalFilters(new HttpExceptionFilter());
 
   app.setGlobalPrefix("api", {
--- a/apps/api/v2/src/app.ts
+++ b/apps/api/v2/src/app.ts
@@ -41,17 +41,16 @@ export const bootstrap = (app: NestExpressApplication): NestExpressApplication =
     })
   );
 
-  const { httpAdapter } = app.get(HttpAdapterHost);
-  app.useGlobalFilters(new SentryFilter(httpAdapter));
-
-  app.useGlobalFilters(new PrismaExceptionFilter());
-
   if (process.env.SENTRY_DNS) {
     Sentry.init({
       dsn: getEnv("SENTRY_DNS"),
     });
   }
-  // Sentry Exception Filter only logs if Sentry.init has not been called
+
+  // Exception filters, new filters go at the bottom, keep the order
+  const { httpAdapter } = app.get(HttpAdapterHost);
+  app.useGlobalFilters(new SentryFilter(httpAdapter));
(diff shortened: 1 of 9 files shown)


apps/api/v2/src/app.ts
```diff
@@ -41,16 +41,17 @@ export const bootstrap = (app: NestExpressApplication): NestExpressApplication =
     })
   );
 
+  const { httpAdapter } = app.get(HttpAdapterHost);
+  app.useGlobalFilters(new SentryFilter(httpAdapter));
```
ThyMinimalDev, line 45: according to nest.js doc: `When combining an exception filter that catches everything with a filter that is bound to a specific type, the "Catch anything" filter should be declared first to allow the specific filter to correctly handle the bound type.`

supalarry: I think that this deserves abstracting filter setup into functions, so that in the future developers do not add specific filters before the catch all Sentry one:

setupCatchAllExceptionFilter(app);
setupSpecificExceptionFilters(app);
setupHttpExceptionFilter(app);

apps/api/v2/src/app.ts
```diff
@@ -41,16 +41,17 @@ export const bootstrap = (app: NestExpressApplication): NestExpressApplication =
...
+
   app.useGlobalFilters(new PrismaExceptionFilter());
 
   if (process.env.SENTRY_DNS) {
```
supalarry, line 49: We don't want to start our app if "SENTRY_DNS" is missing, right? So we can remove the if check and have only
```
Sentry.init({
      dsn: getEnv("SENTRY_DNS"),
    });
```

apps/api/v2/src/modules/oauth/oauth-client.controller.ts
```diff
@@ -66,8 +70,12 @@ export class OAuthClientController {
   @Get("/:clientId")
   @HttpCode(HttpStatus.OK)
   @Roles([MembershipRole.ADMIN, MembershipRole.OWNER, MembershipRole.MEMBER])
+  @UseFilters(HttpExceptionFilter)
```
supalarry, line 73: Do we need to add this if we have HttpEceptionFilter applied globally in app.ts ?

app.useGlobalFilters(new HttpExceptionFilter());

apps/api/v2/src/app.ts
ThyMinimalDev: I think it should not prevent starting the app, especially if we end up having people self hosting apiv2

supalarry: Ah thats true - agreed!

apps/api/v2/test/oauth.e2e-spec.ts
```diff
@@ -1,81 +1,312 @@
...
+    it(`/GET/:id`, () => {
+      return request(appWithoutAuth.getHttpServer()).get("/api/v2/oauth-clients/1234").expect(401);
+    });
+    it(`/POST`, () => {
+      return request(appWithoutAuth.getHttpServer()).post("/api/v2/oauth-clients").expect(401);
+    });
+    it(`/PUT/:id`, () => {
+      return request(appWithoutAuth.getHttpServer()).put("/api/v2/oauth-clients/1234").expect(401);
+    });
+    it(`/delete/:id`, () => {
```
supalarry, line 53: Nit pick - in all tests we have `/delete` in lower case instead of upper case like `/PUT` so would be good to have `/DELETE`

apps/api/v2/test/oauth.e2e-spec.ts
```diff
@@ -1,81 +1,312 @@
...
-        imports: [AppModule, OAuthClientModule, UserModule, AuthModule],
-      })
-    ).compile();
-    const strategy = moduleRef.get(NextAuthStrategy);
-    expect(strategy).toBeInstanceOf(NextAuthMockStrategy);
-    app = moduleRef.createNestApplication();
-    await app.init();
-  });
+    beforeAll(async () => {
+      const moduleRef = await withNextAuth(
```
supalarry, line 72: Love the `withNextAuth` util ❤️

supalarry (approved):
Amazing work on setting up all the fixtures and setting a standard for us how to test endpoints! 🚀 Tests are looking good and ran locally without problems!

Issue #12728, opened 2023-12-11
Webhook on meeting starts


raelix:
### Is your proposal related to a problem?
No, it isn't but would be nice to have the ability of calling a webhook at the start time of a meeting.

### Describe the solution you'd like
The solution aims to have one more available option for webooks. It would allow to call a webhook at the start time of a meeting.

##### House rules
- If this issue has a `🚨 needs approval` label, don't start coding yet. Wait until a core member approves feature request by removing this label, then you can start coding.
  - For clarity: Non-core member issues automatically get the `🚨 needs approval` label.
  - Your feature ideas are invaluable to us! However, they undergo review to ensure alignment with the product's direction.

sakshambp:
can you give more context on what it means?

raelix:
When this will be available on the cloud cal.com?

hrshkshri:
@PeerRich 
I think this issue is resolved ...

Arif3028:
is the issue still exits?

TusharBhatt1:
The PR has been merged , bounty awarded , its seems the issue is resolved.

PR #12764, opened 2023-12-13
feat: meeting started webhook

ibex088:
/claim #12728

right now the output format of the webhook is similar to meeting-ended, but different from all other webhook.
 Additionally, it doesn't currently sign the header with the webhook secret.


packages/features/webhooks/lib/scheduleTrigger.ts
```diff
@@ -45,7 +45,10 @@ export async function addSubscription({
       },
     });
 
-    if (triggerEvent === WebhookTriggerEvents.MEETING_ENDED) {
+    if (
+      triggerEvent === WebhookTriggerEvents.MEETING_ENDED
+      // || triggerEvent === WebhookTriggerEvents.MEETING_STARTED
```
SomayChauhan, line 50: not sure about this

CarinaWolli: I think this is needed for zapier and make.com. As we mostly likely add that trigger there too one day, we can already add this here as well 👍

CarinaWolli (commented):
Nice work 👏🏻 Code looks good, can you uncomment this line of code: https://github.com/calcom/cal.com/pull/12764/files#diff-9ca4b590fb0eb3a0c0be9708bfa9afea092aec81cacac8187fc63864d426fdf9R50

Then it's good to go :)

apps/web/public/static/locales/en/common.json
```diff
@@ -2160,5 +2160,6 @@
   "need_help": "Need help?",
   "troubleshooter": "Troubleshooter",
   "please_install_a_calendar": "Please install a calendar",
+  "meeting_started": "Meeting Started",
```
Aravindh09, line 2163: Suggestion : "meeting_started": "Meeting started", - "started" to start with lower case "s"

CarinaWolli: It should be uppercase like all the other webhooks as well: 
Screenshot 2023-12-19 at 8 47 24 AM

CarinaWolli (approved):
Great work @SomayChauhan 👏🏻 Thank you 🙏

PR #12798, opened 2023-12-14
chore(platform): OAuth Flow

exception:
## What does this PR do?

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


apps/api/v2/src/modules/auth/guard/oauth/refresh-token.guard.ts
```diff
@@ -0,0 +1,37 @@
+import { PrismaReadService } from "@/modules/prisma/prisma-read.service";
+import { CanActivate, ExecutionContext, Injectable, UnauthorizedException } from "@nestjs/common";
+
+@Injectable()
+export class RefreshTokenGuard implements CanActivate {
```
ThyMinimalDev, line 5: this guard would need to be changed a little, the refresh-token would be sent in the body of the request, just need to check if it's valid

checking the access-token in headers is doable for added security, basically just need to make sure the access_token in the headers matches the user associated with the refresh token, doesn't matter if the access token is expired

apps/api/v2/src/modules/oauth/flow/oauth-flow.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+  version: "2",
+})
+export class OAuthFlowController {
+  private logger = new Logger("OauthFlowController");
+
+  constructor(private readonly oauthClientRepository: OAuthClientRepository) {}
+
+  @Post("/authorize")
+  @HttpCode(HttpStatus.OK)
+  @UseGuards(OAuthClientGuard, NextAuthGuard)
```
ThyMinimalDev, line 33: this endpoint is called from the front-end cal.com app, I think the oAuthGard is not needed, frontend can send client_id in the body

apps/api/v2/src/modules/oauth/flow/oauth-flow.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+      status: SUCCESS_STATUS,
+      data: {
+        code: id,
+      },
+    };
+  }
+
+  @Post("/exchange")
+  @HttpCode(HttpStatus.OK)
+  async exchange(
```
ThyMinimalDev, line 62: I think this endpoint most likely uses the oauth client guard, it will be called from the api of the customer passing the client_id and secret and headers

exception: It will be passing the authorization token, exchanging it for the access and refresh ones.

apps/api/v2/src/modules/oauth/flow/oauth-flow.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+      throw new BadRequestException();
+    }
+
+    if (!oauthClient?.redirect_uris.includes(body.redirect_uri)) {
+      throw new BadRequestException("Invalid 'redirect_uri' value.");
+    }
+
+    const { id } = await this.oauthClientRepository.createAuthorizationToken(clientId, userId);
+
+    // ! Redirecting should be taken care of by the client?
```
ThyMinimalDev, line 50: I think this endpoint usually redirects directly to the redirect_uri provided by the user, with the code in the url query parameters

apps/api/v2/src/modules/auth/guard/oauth/refresh-token.guard.ts
exception: The `/refresh` endpoint will need some sort of guard as well, what do you think we should do there?

apps/api/v2/src/modules/oauth/flow/oauth-flow.controller.ts
exception: Hmm, fair. The issue i see is we can't guarantee redirect-uri whitelisting if an error happens, eg we need to redirect to the `body.redirect_uri` on an error, and if that error happens before the check, we can't guarantee it is whitelisted.

apps/api/v2/src/modules/auth/guard/oauth/refresh-token.guard.ts
ThyMinimalDev: I think the guard is just checking if the refresh token is valid

ThyMinimalDev: and if it is requested with the right client_secret

ThyMinimalDev: basically this guard will be tied to the endpoint which should be called fron the customer api, so I think it makes sense they should pass the client_secret as headers

apps/api/v2/src/modules/oauth/flow/oauth-flow.controller.ts
ThyMinimalDev: in case of an error I don't think it matters since the code will not be in the query params

apps/api/v2/src/modules/oauth/oauth-client.repository.ts
```diff
@@ -54,4 +58,143 @@ export class OAuthClientRepository {
...
+        id: clientId,
+        secret: clientSecret,
+      },
+    });
+
+    if (!oauthClient) {
+      throw new BadRequestException("Invalid OAuthClient credentials.");
+    }
+
+    const expiring = await this.dbRead.prisma.refreshToken.findUnique({
```
ThyMinimalDev, line 156: should we rename this to `refreshToken`

apps/api/v2/src/modules/oauth/oauth-client.repository.ts
```diff
@@ -54,4 +58,143 @@ export class OAuthClientRepository {
...
+      throw new BadRequestException("Invalid refresh token");
+    }
+
+    const accessExpiry = DateTime.now().plus({ days: 1 }).startOf("day").toJSDate();
+    const refreshExpiry = DateTime.now().plus({ year: 1 }).startOf("day").toJSDate();
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [_, _refresh, accessToken, refreshToken] = await this.dbWrite.prisma.$transaction([
+      this.dbWrite.prisma.accessToken.deleteMany({
+        where: { client: { id: oauthClient.id }, expiresAt: { lte: new Date() } },
```
ThyMinimalDev, line 178: not sure we want to delete all the expired access token of an oAuth client, we won't be able to differentiate expired tokens still trying to be used from the ones that never existed in the first place

exception: Why would that be something we care about?

apps/api/v2/src/modules/oauth/flow/oauth-flow.service.ts
```diff
@@ -0,0 +1,49 @@
...
+  async validateAccessToken(secret: string) {
+    // status can be "CACHE_HIT" or "CACHE_MISS", MISS will most likely mean the token has expired
+    // but we need to check the SQL db for it anyways.
+    const { status } = await this.readFromCache(secret);
+
+    if (status === "CACHE_HIT") {
+      return true;
+    }
+
+    const token = await this.prismaReadService.prisma.accessToken.findFirst({
```
ThyMinimalDev, line 26: I think we need to create repositories for tokens, to avoid mixing prisma with the business logic

apps/api/v2/src/modules/oauth/oauth-client.repository.ts
ThyMinimalDev: I'm thinking it could be useful to have specific error messages, so that the Atoms know when it should bother trying to refresh the token or just display unauthorized error

apps/api/v2/src/modules/oauth/oauth-client.repository.ts
```diff
@@ -90,24 +90,27 @@ export class OAuthClientRepository {
...
-      },
-      include: {
-        owner: {
-          select: {
-            id: true,
-          },
-        },
-      },
-    });
+    const authorizationToken = oauthClient.authorizationTokens[0];
```
ThyMinimalDev, line 113: now that I think about it, it doesn't change anything since prisma will do multiple queries anyways, sorry my bad xD

I think it was better before and we just need to add a check if the authorizationToken exists and throw error if not 

sorry 🙏

apps/api/v2/src/modules/oauth/flow/oauth-flow.controller.ts
ThyMinimalDev: but true on error the frontend can deal with it so that we display proper error message

apps/api/v2/src/modules/oauth/flow/oauth-flow.service.ts
ThyMinimalDev, line 26: should this use the tokens.repository ?

ThyMinimalDev (approved):

Issue #12806, opened 2023-12-15
Documentation and Script Key Generation Mismatch


guptadeepak8:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

The deploy/install.sh script instructs users to use openssl rand -base64 32 to generate a key and add it under CALENDSO_ENCRYPTION_KEY in the .env file. However, the documented command in the README is openssl rand -base64 24. This may lead to confusion or unexpected behavior.

### Steps to Reproduce

1.Navigate to the deploy directory.
2.Open the install.sh script.
3.Observe the line containing the CALENDSO_ENCRYPTION_KEY .

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

### Actual Results

The install.sh script uses openssl rand -base64 32 to generate a key for CALENDSO_ENCRYPTION_KEY, instead of documented openssl rand -base64 24.

### Expected Results

- The documentation and the install.sh script should align in specifying the correct command for generating the key, avoiding confusion.

rijuoddusual1999:
hii @guptadeepak8 I have solved this issue locally, pls assign me

emrysal:
Hi @guptadeepak8 - this was actually a documentation error in the README.md - I pushed a fix.

PR #12810, opened 2023-12-15
fix: updated the openssl rand -base64 32 with openssl rand -base64 24…

ghost:
As per Document the correct way of adding CALENDSO_ENCRYPTION_KEY in .env file is through generating a secret key through openssl rand -base64 24. But in current head in the ./deploy/install.sh it is written as openssl rand -base64 32, so to avoid confusion I have fixed it in my PR.

* Fixes issue #12806

Screenshot from 2023-12-15 18-28-00


emrysal:
Fixed README.md - thanks for this (just the other way around)

PR #12814, opened 2023-12-15
feat: oauth client guard

supalarry:
## What does this PR do?

1. Create "OAuthClientGuard" that checks id (x-cal-client-id) and secret (x-cal-secret-key) headers sent by the frontend. It will be used to protect cal oAuth clients when new users are about to be added to them.
2. Split out jest config from package.json and update it to resolve import paths correctly in tests
3. Create a new "OAuthClientRepositoryFixture"
4. Tests the "OAuthClientGuard" in "oauth-client.guard.spec.ts"
5. Adds cal headers e.g. "x-cal-client-id" as constants to the platform constants package


apps/api/v2/test/fixtures/oauth-client.repository.fixture.ts
```diff
@@ -0,0 +1,37 @@
+import { CreateOAuthClientInput } from "@/modules/oauth/input/create-oauth-client";
```
supalarry, line 1: I called this "OAuthClientRepositoryFixture" instead of "OAuthClientFixture" just like other fixtures, because during the test I need to instantiate OAuthClient:

let oauthClient: PlatformOAuthClient;

and if we would call fixture "PlatformOAuthClientFixture" then it is confusing.

supalarry (commented):
Self review

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts
```diff
@@ -0,0 +1,29 @@
+import { OAuthClientRepository } from "@/modules/oauth/oauth-client.repository";
```
supalarry, line 1: I tried placing this guard in "modules/auth/guard" but ran into circular dependency issue - this "OAuthClientGuard" guard requires "OAuthClientRepository". If we moved the guard to "AuthModule", then we need to import the "OAuthClientModule" that has the repository to inject in the guard, but the "OAuthClientModule" already imports "AuthModule".

supalarry (commented):
Self review pt. 2

apps/api/v2/test/fixtures/oauth-client.repository.fixture.ts
supalarry: @ThyMinimalDev  thoughts?

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    await expect(guard.canActivate(mockContext)).resolves.toBe(false);
+  });
+
+  afterAll(async () => {
+    await oauthClientRepositoryFixture.delete(oauthClient.id);
+  });
+
+  function createMockExecutionContext(headers: Record<string, string>): ExecutionContext {
+    return createMock<ExecutionContext>({
```
supalarry, line 75: FYI: "createMock" makes mocking ExecutionContext super easy!

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { AppModule } from "@/app.module";
```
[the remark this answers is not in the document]
ThyMinimalDev: I think it's fine to put unit tests next to the related code

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts
supalarry: I created a branch of this branch to continue the work, and now tried importing the new guard into the "user.module.ts" to use the guard on user controller, and run into circular dependency again:

user.module imports OAuthClientModule to use the guard, but OAuthClientModule already imports user.module :/

apps/api/v2/package.json
```diff
@@ -25,6 +25,7 @@
     "@calcom/platform-types": "*",
     "@calcom/platform-utils": "*",
     "@calcom/prisma": "*",
+    "@golevelup/ts-jest": "^0.4.0",
```
ThyMinimalDev, line 28: great find

apps/api/v2/test/fixtures/oauth-client.repository.fixture.ts
ThyMinimalDev: not sure to see the issue with naming the fixture PlatformOAuthClientFixture

ThyMinimalDev: I would just name it like the other fixtures, and import with alias if needed

"import OAuthClientFixture as OAuthClientRepositoryFixture from ..."

apps/api/v2/tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "strict": true,
     "noImplicitAny": true,
     "forceConsistentCasingInFileNames": false,
-    "noFallthroughCasesInSwitch": false
+    "noFallthroughCasesInSwitch": false,
+    "allowJs": true
```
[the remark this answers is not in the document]
supalarry: Because we are using platform constants package which is compiled into .js, and we are importing the .js files in the end. Jest was giving a warning about it.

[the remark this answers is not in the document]
supalarry: Could ignore but then would always have that warning during tests.

apps/api/v2/test/fixtures/oauth-client.repository.fixture.ts
supalarry: But why are we not calling them repository fixtures if they perform CRUD operations just like real repositories?

apps/api/v2/tsconfig.json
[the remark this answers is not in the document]
ThyMinimalDev: let's remove allowJs for now please, I'd rather find a more local solution than impact the tsconfig of the whole project with this flag

apps/api/v2/test/fixtures/oauth-client.repository.fixture.ts
ThyMinimalDev: good point, if it makes more sense we can rename all the other fixtures

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts
ThyMinimalDev: does it make sense to import whole module when we only need the repository for example ?

apps/api/v2/test/fixtures/oauth-client.repository.fixture.ts
supalarry: Done in [refactor: test repository fixtures](https://github.com/calcom/cal.com/pull/12814/commits/665f1d68c0e093aef7502cbd4923f5ab86dd6845).

confirmed that "yarn test" and "yarn test:e2e" work.

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts
ThyMinimalDev: maybe it makes sense to declare oAuth as a global module since most of the endpoints will use it

apps/api/v2/tsconfig.json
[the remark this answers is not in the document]
supalarry: Agreed. Instead, I made "ts-jest" to ignore .js files in [ignore js files in ts-jest](https://github.com/calcom/cal.com/pull/12814/commits/b759275a9c86745e659a3629698ef2750628cf07)

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts
supalarry: Made oAuth client module with its new guard global and made it export oauth repository and the new guard - now other modules like user module can simply use @UseGuards(OAuthClientGuard).

https://github.com/calcom/cal.com/pull/12814/files#diff-b17562a14071a985176d16db25acbe2c3fdf94013c3fd68a08923ed536ca25f6

ThyMinimalDev (approved):

PR #12819, opened 2023-12-15
chore: Moved e2e suite to run on pushes to main

keithwillcode:
## What does this PR do?

Instead of running the full e2e test suite on every PR update, we are now going to run it on pushes to main. This will reduce PR wait times and costs. A known side effect is that we can end up with broken e2e tests in main but we will ensure this suite is passing before new releases go out. This PR precedes other changes to the release process to run this e2e job.

## Type of change

## How should this be tested?

- Manually trigger this to ensure all of the E2E tests run correctly


emrysal (approved):
LGTM

zomars (approved):
LGTM

PR #12828, opened 2023-12-16
chore: Chore team metadata table + isOrganization migration from metadata

sean-brydon:
## Type of change

- [X] Tests (Unit/Integration/E2E or any other test)

## How should this be tested?

1. Seed the database 
2. Connect to DB 
3.  UPDATE "Team" SET "metadata" = '{"isOrganization": "true", "isOrganizationConfigured": "true", "orgAutoAcceptEmail": "example@example.com", "isOrganizationVerified": "false"}'
WHERE "id" in (1,2,3);
UPDATE "Team" SET "isOrganization" = true WHERE "id" in (1,2,3);
4. Run SQL above to set metadata 
5. Run migrations - ensure migrations have ran successfully and each team in the above query has a OrganizationSettings  table associated with it 

## Mandatory Tasks

## Checklist


packages/prisma/migrations/20231215054122_organization_metadata/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+	(t.metadata ->> 'isOrganizationVerified')::BOOLEAN AS "isOrganizationVerified"
+FROM (
+	SELECT
+		id,
+		metadata
+	FROM
+		"Team"
+	WHERE
+		metadata ->> 'isOrganization' = 'true'
+) AS t;
```
sean-brydon, line 45: important: 

Here we are just updating a table - we are not removing or deleting this data from metadata FOR NOW. 

Once we have ran this custom migration we can consider removing the existing fields in prod db to ensure we dont have duplicated data.

packages/trpc/server/routers/viewer/organizations/adminVerify.handler.ts
```diff
@@ -42,18 +38,13 @@ export const adminVerifyHandler = async ({ input }: AdminVerifyOptions) => {
...
     where: {
-      id: input.orgId,
+      teamId: input.orgId,
     },
     data: {
-      metadata: {
-        ...metaDataParsed,
-        isOrganizationVerified: true,
-        orgAutoAcceptEmail: acceptedEmailDomain,
-      },
```
sean-brydon, line 56: We no longer need to update metadata for this as it is handled in organizationSettings

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
```diff
@@ -33,15 +33,17 @@ export const inviteMemberHandler = async ({ ctx, input }: InviteMemberOptions) =
   await checkRateLimitAndThrowError({
     identifier: `invitedBy:${ctx.user.id}`,
   });
+  const team = await getTeamOrThrow(input.teamId);
+
+  const isOrg = team.isOrganization;
```
sean-brydon, line 38: Switched this order around so we can get if it is an organization based on teamId 

@ThyMinimalDev can you confirm this is the correct logic here?

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMemberUtils.test.ts
sean-brydon, line 1: @ThyMinimalDev given the changes in [inviteHandler](https://github.com/calcom/cal.com/pull/12828/files#diff-95177a5fa5351ed5e907bdd95108677a14b23c59f5b214fd059d15750445b921) do these changes look fine to you?

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
ThyMinimalDev: makes sense, probably better to check if the team is an org from database rathe than relying on input parameter

PeerRich:
oh nice! metadata as json was a bad idea from the get go. great job refactoring!

sean-brydon:
Note : e2e suite has some failing test but it isn’t shown with the new changes

apps/web/pages/signup.tsx
```diff
@@ -665,8 +667,8 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   };
 
   const isATeamInOrganization = tokenTeam?.parentId !== null;
-  const isOrganization = tokenTeam.metadata?.isOrganization;
   // Detect if the team is an org by either the metadata flag or if it has a parent team
```
hariombalhara, line 670: This comment was stale. Should have been  'Detect if the team is an org by checking metadata'

packages/prisma/schema.prisma
```diff
@@ -313,9 +313,21 @@ model Team {
...
   instantMeetingTokens InstantMeetingToken[]
 
   @@unique([slug, parentId])
+  @@unique([slug, isOrganization])
+}
+
+model OrganizationSettings {
+  id                       Int     @id @default(autoincrement())
+  team                     Team    @relation(fields: [teamId], references: [id], onDelete: Cascade)
+  teamId                   Int     @unique
```
hariombalhara, line 327: let's call it organizationId

packages/trpc/server/routers/viewer/organizations/adminUpdate.handler.ts
```diff
@@ -49,13 +50,33 @@ export const adminUpdateHandler = async ({ input }: AdminUpdateOptions) => {
...
       requestedSlug: undefined,
-      ...input.metadata,
     });
   }
 
-  const updatedOrganisation = await prisma.team.update({
-    where: { id },
-    data,
+  const updatedOrganisation = await prisma.$transaction(async (tx) => {
+    const updatedOrganisation = await prisma.team.update({
```
hariombalhara, line 57: For it to be considered in a transaction I think it needs to use `tx`
```suggestion
    const updatedOrganisation = await tx.team.update({
```

packages/trpc/server/routers/viewer/organizations/adminUpdate.handler.ts
```diff
@@ -49,13 +50,33 @@ export const adminUpdateHandler = async ({ input }: AdminUpdateOptions) => {
...
+      where: { id },
+      data,
+    });
+
+    if (organizationSettings || existingOrg.organizationSettings) {
+      await prisma.organizationSettings.update({
```
hariombalhara, line 63:
```suggestion
      await tx.organizationSettings.update({
```

packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -123,12 +123,16 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
...
             metadata: {
               ...(IS_TEAM_BILLING_ENABLED ? { requestedSlug: slug } : {}),
-              isOrganization: true,
-              isOrganizationVerified: true,
-              isOrganizationConfigured,
-              orgAutoAcceptEmail: emailDomain,
+            },
+            organizationSettings: {
+              create: {
+                isOrganizationVerified: false,
```
hariombalhara, line 132:
```suggestion
                isOrganizationVerified: true,
```

All orgs are now automatically verified

hariombalhara (dismissed):
Awesome work @sean-brydon 🙏 Thanks for the PR

hariombalhara (approved):

hariombalhara (approved):

zomars:
Heads up. the migration in this PR is failing on prod. Blocking the latest release.

packages/prisma/migrations/20231215054122_organization_metadata/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+    "orgAutoAcceptEmail" TEXT NOT NULL,
+
+    CONSTRAINT "OrganizationSettings_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "OrganizationSettings_organizationId_key" ON "OrganizationSettings"("organizationId");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "Team_slug_isOrganization_key" ON "Team"("slug", "isOrganization");
```
zomars, line 18:
```
ERROR: could not create unique index "Team_slug_isOrganization_key"
DETAIL: Key (slug, "isOrganization")=(advisors, f) is duplicated.
```

zomars:
@calcom/foundation @sean-brydon had to revert this PR due to migration not being compatible with actual prod data. Causing it to fail.

zomars:
From internal chat:

> ... parentId isn't being taking into account for the new unique constrain
> just slug and isOrganization
> So one of those is a subteam and another one is a stantalone team (non-org) with slug "advisors". Both have metadata→isOrganization as null/false. The former because is a subteam in an org and the latter because  is not on an org. Hence the conflict.

Issue #12836, opened 2023-12-16
[Feature] Removing white borders after toggling buttons


H9660:
While toggling buttons in the `event-types` page, I found that when we toggle them and their is a white border that comes around them which does not go until we click somewhere. 

What I propose it that we could either remove that white border that comes every time the button is toggled or make it transparent.

Here is a short demonstration:

https://github.com/calcom/cal.com/assets/73359258/87d0f709-0cfc-4795-b87c-0f9bd2e6a1ac

Steps to reproduce:

- Visit https://app.cal.com/
- Click login and create an account
- You will then be redirected to the event types page. 
- There you can find the above demonstration.

ameeetgaikwad:
Would like to help! Can someone assign me?

H9660:
@ameeetgaikwad This feature is no longer needed. Thanks!

PR #12840, opened 2023-12-17
style: removed white borders in Switch component.

H9660:
Added new style for buttons: [#12836 ](url)

### DIFF
--- a/packages/ui/components/form/switch/Switch.tsx
+++ b/packages/ui/components/form/switch/Switch.tsx
@@ -44,7 +44,7 @@ const Switch = (
           className={cx(
             isChecked ? "bg-brand-default dark:bg-emphasis" : "bg-emphasis",
             primitiveProps.disabled && "cursor-not-allowed",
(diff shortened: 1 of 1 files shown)


packages/ui/components/form/switch/Switch.tsx
```diff
@@ -44,7 +44,7 @@ const Switch = (
           className={cx(
             isChecked ? "bg-brand-default dark:bg-emphasis" : "bg-emphasis",
             primitiveProps.disabled && "cursor-not-allowed",
-            "focus:ring-brand-default h-5 w-[34px] rounded-full shadow-none focus:border-neutral-300 focus:outline-none focus:ring-2 focus:ring-neutral-800 focus:ring-offset-1",
```
Udit-takkar, line 47: @H9660 this is intentional

Udit-takkar: we don't want to remove this

Issue #12773, opened 2023-12-13
Fix the Bg color of the sign In button after clicking


anikdhabal:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

![Screenshot 2023-12-13 180213](https://github.com/calcom/cal.com/assets/81948346/1dce5a27-5443-4962-9f01-d2f89c82abef)

### Steps to Reproduce

1. Go to Sign In
2. Clicked on Sign in button

sebzz2k2:
Can i work on it ?

PeerRich:
![CleanShot 2023-12-13 at 09 53 23@2x](https://github.com/calcom/cal.com/assets/8019099/9718efe8-f620-4eef-ba2d-b18a791c14cc)
thats weird, i cant reproduce this

anikdhabal:
[...]

PR #12851, opened 2023-12-17
fix: bg color of the sign In button after clicking

Nitin034:
Fix Issue: #12773


Udit-takkar (changes requested):
@Nitin034 the issue is not fixed

PR #12853, opened 2023-12-18
feat: CRUD for oAuth users

supalarry:
## What does this PR do?

Allows creating users tied to a specific internal Cal oAuth, and then to fetch or update them. Since all managed users will be tied to an oAuth client, the routes are:
1. POST `/api/v2/oauth-clients/:clientId/users` - create new user tied to oAuth client
2. GET `/api/v2/oauth-clients/:clientId/users/:userId` - access oAuth client user
3. PUT `/api/v2/oauth-clients/:clientId/users/:userId` - update oAuth client user

The new routes can be found in the new [OAuthUserController](https://github.com/calcom/cal.com/pull/12853/files#diff-e9b3a49ec089b2b99e95994b4fd9b7dab220b9026ed020e42f6ffc658ec5ec2d).

Because of the decision to have `clientId` in the URL, the [OAuthClientGuard](apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts) was changed to access `clientId` from the URL and also it's tests [have been updated](https://github.com/calcom/cal.com/pull/12853/files#diff-73f42b57f202ac41d02143a48b59b30865e3c9d6f0819b41750fe6d1c5a75530).

Linear issue: [PRO-155](https://linear.app/calcom/issue/PRO-155/crud-platform-users-using-oauth-client)

## How should this be tested?

The new [OAuthUserController](https://github.com/calcom/cal.com/pull/12853/files#diff-e9b3a49ec089b2b99e95994b4fd9b7dab220b9026ed020e42f6ffc658ec5ec2d) has [e2e tests](https://github.com/calcom/cal.com/pull/12853/files#diff-aad793ae3f067850994d6a81afe4c9faf0c1dcbb2d29e0c368b1a4cac7e8defb). To test manually check test structure and make postman requests.


apps/api/v2/src/modules/oauth/user/oauth-user.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+  ): Promise<ApiResponse<CreateUserResponse>> {
+    this.logger.log(
+      `Creating user with data: ${JSON.stringify(body, null, 2)} for OAuth Client with ID ${oAuthClientId}`
+    );
+
+    const exists = await this.userRepository.findByEmail(body.email);
+
+    if (exists) {
+      throw new BadRequestException(DUPLICATE_RESOURCE);
+    }
```
ThyMinimalDev, line 49:
```suggestion
    const existingUser = await this.userRepository.findByEmail(body.email);

    if (Boolean(existingUser)) {
      throw new BadRequestException(DUPLICATE_RESOURCE);
    }
```
suggestion for more semanticity

apps/api/v2/src/modules/user/user.repository.ts
```diff
@@ -1,29 +1,69 @@
...
       where: {
         email,
       },
     });
+
+    if (!user) {
+      return user;
+    }
+
     return this.sanitize(user);
```
ThyMinimalDev, line 50:
```suggestion

    if (user) {
          return this.sanitize(user);
    }

    return null;
```

ThyMinimalDev: suggestion for more semanticity

apps/api/v2/src/modules/oauth/user/oauth-user.e2e-spec.ts
```diff
@@ -0,0 +1,180 @@
+import { bootstrap } from "@/app";
```
ThyMinimalDev, line 1: should we add .controller in this file name ?

ThyMinimalDev: oauth-user.controller.e2e-spec.ts

apps/api/v2/src/modules/oauth/user/oauth-user.controller.ts
supalarry: I agree with renaming "exists" to "existingUser" but I did not add "Boolean" in [refactor: variable naming](https://github.com/calcom/cal.com/pull/12853/commits/3ca77178df3f7e70a056c419b60b483d2fce2c5d), because in my opinion its redundant - if "findByEmail" returns an object or null, then anything other than null in this case is true. It would be similar to explicitly writing types in typescript IMO.

apps/api/v2/src/modules/user/user.repository.ts
supalarry: Good suggestion! Done.

apps/api/v2/src/modules/oauth/user/oauth-user.e2e-spec.ts
supalarry: I like the idea of this convention! Done in [rename test file](https://github.com/calcom/cal.com/pull/12853/commits/cd318586bbe87d8a074e6724f64dac906e9822f1)

apps/api/v2/src/modules/oauth/guard/oauth-client/oauth-client.guard.ts
```diff
@@ -1,29 +1,40 @@
...
+
+    const client = await this.oauthRepository.getOAuthClient(oauthClientId);
 
-    if (!oauthClient || oauthClient.secret !== oauthClientSecret) {
-      return false;
+    if (!client || client.secret !== oauthClientSecret) {
+      throw new UnauthorizedException();
     }
 
+    request.oAuthClient = client;
```
ThyMinimalDev, line 36: we dont need to type the request or create a decorator for this ?

supalarry: Removed in [remove oauth client from request](https://github.com/calcom/cal.com/pull/12853/commits/5817d57f01f5b36d5dae951aad6d5f59f1999e4d)

PR #12858, opened 2023-12-18
chore: User page crash on bad metadata

emrysal:
## What does this PR do?

When the metadata is bad the [user].tsx page shows a 500 error. This should just exclude the bad event type and shows the correct ones.


apps/web/pages/[user].tsx
```diff
@@ -1,3 +1,4 @@
+import * as Sentry from "@sentry/node";
```
Comment (keithwillcode) [inline apps/web/pages/[user].tsx:1]: Ideally we don't import Sentry all across the app but instead we have a generic logger/exception handler that we can shield this dependency behind which would allow us to quickly swap out Sentry in the future if desired. Let's quickly add this, please. 🙏

keithwillcode (changes requested):

apps/web/pages/[user].tsx
```diff
@@ -20,6 +20,7 @@ import { getUsernameList } from "@calcom/lib/defaultEvents";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { useRouterQuery } from "@calcom/lib/hooks/useRouterQuery";
 import useTheme from "@calcom/lib/hooks/useTheme";
+import logger from "@calcom/lib/logger";
```
Comment (keithwillcode) [inline apps/web/pages/[user].tsx:23]: It seems like we are going to need to wire up something in the logger itself to actually send the exception to Sentry. Were you able to find docs stating that when Sentry is wired up in the Next config, tslog is compatible and will send the exception out?

keithwillcode (approved):

PR #12878, opened 2023-12-19
fix: creating workflow reminders for existing bookings when new step is added

CarinaWolli:
## What does this PR do?

This PR fixes the error in `workflows/update.handler.ts` when adding a new step to a workflow to and workflow that has active event types.

Error:
Screenshot 2023-12-19 at 1 54 34 PM

## Type of change

## How should this be tested?

- Create new workflow with before event trigger
- In dropdown, set active on an event type 
- add a new step, 'send email to host'
- save workflow

### DIFF
--- a/packages/trpc/server/routers/viewer/workflows/update.handler.ts
+++ b/packages/trpc/server/routers/viewer/workflows/update.handler.ts
@@ -262,7 +262,7 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
         },
       });
 
-      const promiseSteps = steps.map(async (step) => {
+      const promiseSteps = userWorkflow.steps.map(async (step) => {
         if (
           step.action !== WorkflowActions.SMS_ATTENDEE &&
           step.action !== WorkflowActions.WHATSAPP_ATTENDEE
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -262,7 +262,7 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
         },
       });
 
-      const promiseSteps = steps.map(async (step) => {
```
CarinaWolli, line 265: `steps` includes new steps that are not yet created in the DB. That's why it was throwing the foreign key error

packages/trpc/server/routers/viewer/workflows/update.handler.ts
```diff
@@ -619,11 +619,7 @@ export const updateHandler = async ({ ctx, input }: UpdateOptions) => {
       return stepToAdd;
     }
   });
-
   if (addedSteps) {
-    const eventTypesToCreateReminders = activeOn.filter(
-      (activeEventType) => activeEventType && !newEventTypes.includes(activeEventType)
```
CarinaWolli, line 625: As we don't schedule email reminders for new event types anymore on the new steps, we want to schedule them here, so no need to filter them out

Udit-takkar (approved):
LGTM

PR #12894, opened 2023-12-20
fix: only use event type's webhooks

Udit-takkar:
- Only get the event type's webhook with trigger as WebhookTriggerEvents.INSTANT_MEETING.

As we don't want to execute webhooks of all event types when instant meeting is booked.

### DIFF
--- a/packages/features/instant-meeting/handleInstantMeeting.ts
+++ b/packages/features/instant-meeting/handleInstantMeeting.ts
@@ -31,15 +31,11 @@ const handleInstantMeetingWebhookTrigger = async (args: {
     const subscribers = await prisma.webhook.findMany({
       where: {
         AND: {
-          eventTypeId: {
-            equals: args.eventTypeId,
-          },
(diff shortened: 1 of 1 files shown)


packages/features/instant-meeting/handleInstantMeeting.ts
```diff
@@ -24,12 +22,35 @@ import prisma from "@calcom/prisma";
...
+        AND: {
+          eventTypeId: {
+            equals: args.eventTypeId,
+          },
+          eventTriggers: {
+            has: eventTrigger,
+          },
+          active: {
+            equals: true,
+          },
```
hariombalhara, line 42: nit: ```suggestion
          eventTypeId: args.eventTypeId,
          eventTriggers: {
            has: eventTrigger,
          },
          active: true,
```

hariombalhara (dismissed):
Codewise looks good

hariombalhara (approved):

PR #12900, opened 2023-12-20
fix: fix-animations on event-types/[type] page

dmytro068:
## What does this PR do?

Updates the version of autoAnimate lib, current version had a bug in the useAutoAnimate hook (see [doc](https://www.notion.so/intuita/Animations-regression-not-formatted-0880517d286842c6979795706f0aa23d))

Adds prefetching of dynamically imported tabs. Prefetching was adding because dynamic loading of components caused broken animation when clicking tab for the first time. (We can consider disabling prefetching on mobile devices to save traffic)

## Requirement/Documentation

[Document](https://www.notion.so/intuita/Animations-regression-not-formatted-0880517d286842c6979795706f0aa23d)

## Type of change

## How should this be tested?

Navigate to `/event-types/[type]` click tabs, see if animation is not broken. 
[Screencast from 20.12.23 18:00:30.webm](https://github.com/calcom/cal.com/assets/125881252/ff38ed2b-ade0-435c-9148-fa59ceed7cff)

## Mandatory Tasks

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -39,7 +39,7 @@
     "@calcom/tsconfig": "*",
     "@calcom/ui": "*",
     "@daily-co/daily-js": "^0.37.0",
-    "@formkit/auto-animate": "^1.0.0-beta.5",
+    "@formkit/auto-animate": "^0.8.1",
     "@glidejs/glide": "^3.5.2",
     "@hookform/error-message": "^2.0.0",
     "@hookform/resolvers": "^2.9.7",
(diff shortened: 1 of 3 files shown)


sean-brydon (dismissed):
Changes LGTM 

Would you be able to also update the other package that uses formkit auto animate so we dont have a mistmatch in versions?

DmytroHryshyn:
> Changes LGTM
> 
> Would you be able to also update the other package that uses formkit auto animate so we dont have a mistmatch in versions?

I see that only 2 packages use `@formkit/auto-animate` and they are already updated to the latest version
image

PeerRich (approved):

PR #12907, opened 2023-12-21
fix: check for isEmbed

sean-brydon:
fixes: https://discord.com/channels/1125390093386010654/1182022772000555039/1182022772000555039

Hides "needs help" button on embed.


packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -32,7 +32,7 @@ export function Header({
 }) {
   const { t, i18n } = useLocale();
   const session = useSession();
-
+  const isEmbed = typeof window !== "undefined" && window?.isEmbed?.();
```
hariombalhara, line 35: How about using `useIsEmbed`?

hariombalhara (dismissed):
I think we can use the hook here.

packages/features/bookings/Booker/components/Header.tsx
PeerRich: yeh lets use the useIsEmbed hook her

PeerRich:
also, the need help button only shows up in your own embed if you are the owner of the link

packages/features/bookings/Booker/components/Header.tsx
```diff
@@ -32,7 +33,7 @@ export function Header({
...
```
PeerRich, line 36:
```suggestion
  const isEmbed = useIsEmbed();
```

PeerRich: this is better, no @hariombalhara ?

keithwillcode (approved):

hariombalhara (approved):

Issue #12888, opened 2023-12-20
[CAL-2834] Colour referenced by bg-emphasis in dark mode (presumably stone-700) needs updating


ciaranha:
Tiny update which affects our bg-emphasis token in dark mode. I'm assuming the issue is coming from the `stone-700` value and..

* `stone-700` is currently potentially set to `#525252`
* should be `#404040`

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2834](https://linear.app/calcom/issue/CAL-2834/colour-referenced-by-bg-emphasis-in-dark-mode-presumably-stone-700)</sub>

PR #12909, opened 2023-12-21
fix: bg emphasis darkmdoe

sean-brydon:
fixes: CAL-2834 #12888

### DIFF
--- a/apps/web/styles/globals.css
+++ b/apps/web/styles/globals.css
@@ -51,7 +51,7 @@
 .dark {
   /* background */
 
-  --cal-bg-emphasis: hsla(0, 0%, 32%, 1);
+  --cal-bg-emphasis: hsla(0, 0%, 32%,1);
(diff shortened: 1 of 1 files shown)


apps/web/styles/globals.css
```diff
@@ -51,7 +51,7 @@
 .dark {
   /* background */
 
-  --cal-bg-emphasis: hsla(0, 0%, 32%, 1);
+  --cal-bg-emphasis: hsla(0, 0%, 32%,1);
```
Udit-takkar, line 54: @sean-brydon isn't this same as before? just without space between comman and 1

sean-brydon: Ermmm yes - very confused as this was wrong for me locally - just merged and it’s this like someone beat me to it

emrysal:
Closing as it's already done (with correct space :))

PR #12912, opened 2023-12-21
feat: endpoint for deleting oAuth users & oAuth users returned data

supalarry:
1. Endpoint to delete oAuth users that do not have username or password set. Tests have been created too.
2. In [driveby: return only user id and email in oauth users endpoints](https://github.com/calcom/cal.com/pull/12912/commits/eb98f5e67e9d90b09fd83e3b332dfc42c2bf5094) made oAuth client users endpoint only return user id and email as the user data.


apps/api/v2/src/modules/oauth/user/oauth-user.controller.ts
```diff
@@ -91,6 +92,31 @@ export class OAuthUserController {
...
     return { status: SUCCESS_STATUS, data: user };
   }
+
+  @Delete("/:userId")
+  @HttpCode(HttpStatus.OK)
+  @UseGuards(AccessTokenGuard)
+  async deleteUser(@Param("userId") userId: number): Promise<ApiResponse<Partial<User>>> {
+    this.logger.log(`Deleting user with ID: ${userId}`);
+
+    const existingUser = await this.userRepository.findById(userId);
```
ThyMinimalDev, line 102: I think we should get the user through the oAuth client, make sure the user is associated to the client_id

ThyMinimalDev: or just make sure this user has this oAuth client

ThyMinimalDev: actually we should also check if the Access Token belongs to the  user you are trying to delete

ThyMinimalDev: we should create a guard called, AccessTokenOwnershipGuard, this guard checks that the access token is tied to the userId

and we should update the AccessTokenGuard, it should check if the access_token belongs to the oauth client

supalarry: Done in https://github.com/calcom/cal.com/pull/12913

ThyMinimalDev (approved):

PR #12916, opened 2023-12-21
chore: Revert app dir (2)

emrysal:
## What does this PR do?

Reverts appDir changes so e2e is closer to passing.


keithwillcode:
We’re using scripts to ignore building this until we fix build issues.

PR #12923, opened 2023-12-22
fix: Creating New Bookings From Already Rescheduled Links

### LINKED ISSUE #12922: [CAL-2983] Can make multiple reschedules against an already rescheduled Booking: this creates multiple new active bookings
### Issue Summary

* a user has made a booking, and is provided a rescheduled link.
  * ie.. the url for the event-type booking page + `?rescheduleUid={{uid}}` in the url params for example.
  * or via the `/reschedule/{{uid}}`  url
* They use this link to reschedule their original booking,   the original booking is marked 'cancelled' in calcom db and a new booking is made.
* Then,  If they later use that same original reschedule link, they are able to make another new booking despite the original being marked 'cancelled',  the new booking from the previous reschedule is ignored.
* This effectively turns one booking into two.   and can be used unlimited times.

### Steps to Reproduce

* make a booking
* reschedule that booking using the event-type booking page plus url params ie- `?rescheduleUid={{Original_bookgin_uuid}}`
  * or via the `/reschedule/{{uid}}` url
* reschedule the appointment.
* now use that same reschedule url. and reschedule the original appointment again.
* You should now have two active bookings and one cancelled booking in the DB.

### Actual Results

* two reschedules were made again the one original booking.  Now customer has two active bookings instead of one.

### Expected Results

* Option 1
  * the original reschedule uid is no longer valid due to status = cancelled,  customer must find the new reschedule link from the new booking.
* Option 2
  * The second reschedule attempt checks for all previous reschedules against this original booking and cancels them too.
    * this could get complex if dealing with already duplicated reschedules
  * The end result should be only one active booking and two cancelled in the database.

### Technical details

v3.5.5

### Evidence

Rows from the "Bookings" table after following steps above.

image

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2983](https://linear.app/calcom/issue/CAL-2983/can-make-multiple-reschedules-against-an-already-rescheduled-booking)</sub>

joeauyeung:
## What does this PR do?

This PR prevents new bookings from being created when the same reschedule link is used multiple times.

Fixes #12922

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a booking
- Reschedule a booking with link #1
- Try rescheduling with the same booking link
    - It should bring you back to the success page of the most recent booking

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/pages/booking/[uid].tsx
+++ b/apps/web/pages/booking/[uid].tsx
@@ -92,6 +92,7 @@ const querySchema = z.object({
   isSuccessBookingPage: stringToBoolean,
   formerTime: z.string().optional(),
   seatReferenceUid: z.string().optional(),
+  alreadyRescheduled: stringToBoolean,
 });
 
 export default function Success(props: SuccessProps) {
@@ -107,6 +108,7 @@ export default function Success(props: SuccessProps) {
     formerTime,
     email,
     seatReferenceUid,
+    alreadyRescheduled,
   } = querySchema.parse(routerQuery);
 
   const attendeeTimeZone = props?.bookingInfo?.attendees.find(
@@ -283,6 +285,9 @@ export default function Success(props: SuccessProps) {
       }
       return t(`needs_to_be_confirmed_or_rejected${titleSuffix}`);
     }
+    if (alreadyRescheduled) {
+      return t("below_are_the_details_of_your_rescheduled_meeting");
+    }
     return t(`emailed_you_and_attendees${titleSuffix}`);
   }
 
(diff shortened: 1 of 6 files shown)


joeauyeung (commented):
The main mechanism of finding the last booking in the reschedule chain is to see if there is a booking in the `rescheduleTo` field. We then keep querying until we find the end. Once we find the right bookingUid then we query for the required booking data.

I was worried about making multiple Prisma queries but since we are only getting 3 string fields, I didn't notice too much of a slow down locally. Also most users will have the most recent reschedule link but I think it's worth making the multiple queries to prevent branching bookings.

keithwillcode:
Closing this until we have the real need to come back and address it permanently. Currently, the bug is not possible to reproduce because we now return a 404. Feel free to reopen this PR when/if the issue is taken back up.

PR #12932, opened 2023-12-22
refactor: access token errors

supalarry:
1. Throw a newly created `TokenExpiredException` in the `validateAccessToken` of `oauth-flow.service.ts` if access token is expired. It will throw 498 using newly created constants `ACCESS_TOKEN_EXPIRED` and `HTTP_CODE_TOKEN_EXPIRED` stored in `@calcom/platform-constants`.
2. Throw `InternalServerErrorException` if access token was provided and is valid, but we could not find access token in our database using the secret (access token from perspective of user) - it should not happen and is not users fault but ours.


apps/api/v2/src/modules/auth/strategies/access-token/access-token.strategy.ts
```diff
@@ -25,19 +25,15 @@ export class AccessTokenStrategy extends PassportStrategy(BaseStrategy, "access-
...
-      if (!valid) {
-        throw new UnauthorizedException("Access token is missing or invalid.");
-      }
+      this.oauthFlowService.validateAccessToken(accessToken);
 
       const ownerId = await this.tokensRepository.getAccessTokenOwnerId(accessToken);
 
       if (!ownerId) {
-        throw new UnauthorizedException("Invalid access token");
+        throw new InternalServerErrorException("Access token not found in the database");
```
Ryukemeister, line 36: NIT: what if we modify the error message as `Access token not found` instead of `Access token not found in the database`? But I don't understand why you can't use previous message `Invalid access token` because this sounds pretty clear and concise too so why change it?
Cc @ThyMinimalDev

Ryukemeister: Now that the access token expiry has a different message `ACCESS_TOKEN_IS_EXPIRED` I think just returning `Invalid access token` should be good!

PR #12950, opened 2023-12-27
feat: `/me` endpoint for v2 api

Ryukemeister:
## What does this PR do?

Adds a `/me` endpoint in the v2 api that returns some general info about a user


apps/api/v2/src/modules/user/controllers/user.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+export class UserController {
+  private readonly logger = new Logger("UserController");
+
+  constructor(private readonly userRepository: UserRepository) {}
+
+  @Get("/")
+  @HttpCode(HttpStatus.OK)
+  @UseGuards(AccessTokenGuard)
+  async getUser(
+    @Param("clientId") clientId: string,
```
ThyMinimalDev, line 24: client_id is not needed

apps/api/v2/src/modules/user/controllers/user.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+    @Body() body: GetUserInput,
+    @GetUser() user: any
```
ThyMinimalDev, line 26: avoid using any, use the type return by getUser decorator

apps/api/v2/src/modules/user/controllers/user.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+  ): Promise<ApiResponse<UserReturned>> {
+    if (!user) {
+      throw new BadRequestException("This user does not exist.");
+    }
```
ThyMinimalDev, line 30: not needed, the access token guard should fail if there are no associated user

Ryukemeister:
Closing this since everything here is included in #13006

PR #12954, opened 2023-12-28
feat: external gcal

ThyMinimalDev:
## What does this PR do?

connect external calendar

## Type of change


apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+  ) {}
+
+  @Get("/oauth/redirect")
+  @Redirect(undefined, 301)
+  @UseGuards(AccessTokenGuard)
+  async redirect(@Req() req: Request): Promise<ApiRedirectResponseType> {
+    const scopes = [
+      "https://www.googleapis.com/auth/calendar.readonly",
+      "https://www.googleapis.com/auth/calendar.events",
+    ];
```
exception, line 50: Define scopes as a constant out of the function call. Avoid allocations.

apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+    const app = await this.appRepository.getAppBySlug("google-calendar");
+    if (!app) {
+      throw new NotFoundException();
+    }
+    const keys = app.keys as { client_id?: string; client_secret?: string };
+    if (!keys.client_id) throw new BadRequestException("Missing google calendar client_id.");
+    if (!keys.client_secret) throw new BadRequestException("Missing google calendar secret.");
```
exception, line 57: Could use Zod safeParse for this.

apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+  @Redirect(undefined, 301)
+  @UseGuards(AccessTokenGuard)
+  async redirect(@Req() req: Request): Promise<ApiRedirectResponseType> {
+    const scopes = [
+      "https://www.googleapis.com/auth/calendar.readonly",
+      "https://www.googleapis.com/auth/calendar.events",
+    ];
+    const app = await this.appRepository.getAppBySlug("google-calendar");
+    if (!app) {
+      throw new NotFoundException();
```
exception, line 53: Could throw the exception directly on the repository but should be ok.

apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+      state: `accessToken=${accessToken}&origin=${origin}`,
+    });
+    return { url: authUrl };
+  }
+
+  @Get("/oauth/save")
+  @Redirect(undefined, 301)
+  @HttpCode(HttpStatus.OK)
+  async save(@Query("state") state: string, @Query("code") code: string): Promise<ApiRedirectResponseType> {
+    const stateParams = new URLSearchParams(state);
```
exception, line 75: Personally I would use Zod to parse the url params and ensure types.

exception (changes requested):

apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    if (!gcalCredentials) {
+      throw new BadRequestException();
+    }
+
+    if (gcalCredentials.invalid) {
+      throw new BadRequestException("Invalid google oauth credentials.");
+    }
+
+    return { status: SUCCESS_STATUS, data: undefined };
```
exception, line 156: imo, we should be able to return `{ status: SUCCESS_STATUS }` without any data.

apps/api/v2/src/modules/credentials/credentials.repository.ts
```diff
@@ -0,0 +1,23 @@
...
+  createAppCredential(type: string, key: Prisma.InputJsonValue, userId: number, appId: string) {
+    return this.dbWrite.prisma.credential.create({
+      data: {
+        type,
+        key,
+        userId,
+        appId,
+      },
+    });
+  }
```
exception, line 19:
```suggestion
  }
  
```

apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
ThyMinimalDev: not sure repository's should be handling HTTP exceptions

apps/api/v2/src/modules/credentials/credentials.repository.ts
```diff
@@ -0,0 +1,23 @@
...
+  getByTypeAndUserId(type: string, userId: number) {
```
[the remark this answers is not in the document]
supalarry: The same for app slugs "google-calendar" and maybe create a map where both of them are tied together

apps/api/v2/src/modules/apps/controllers/gcal-oauth/gcal-oauth.controller.ts
```diff
@@ -0,0 +1,150 @@
...
+    const calendar = google.calendar({
+      version: "v3",
+      auth: oAuth2Client,
+    });
+
+    const cals = await calendar.calendarList.list({ fields: "items(id,summary,primary,accessRole)" });
+
+    const primaryCal = cals.data.items?.find((cal) => cal.primary);
+
+    if (primaryCal?.id) {
```
supalarry, line 122: Question: why don't we want to only set selected calendar if its primary aka if none of them are primary, why ignore them all - shouldn't we then throw an error or something, like whats the use case without creating a selected calendar?

ThyMinimalDev: not 100% sure on this, this logic comes from the main app

PR #12982, opened 2024-01-02
fix: use same timezone to compare

### LINKED ISSUE #12971: [CAL-2850] After reschedule Cal Video link is broken
Consistenly replicable after a reschedule

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2850](https://linear.app/calcom/issue/CAL-2850/after-reschedule-cal-video-link-is-broken)</sub>

Udit-takkar:
Fixes: https://github.com/calcom/cal.com/issues/12971

Although i wasn't able to replicate the bug but this seems https://github.com/calcom/cal.com/pull/12982#discussion_r1439338340 the main problem.


apps/web/pages/video/[uid].tsx
```diff
@@ -305,11 +305,12 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
   }
 
   //daily.co calls have a 60 minute exit buffer when a user enters a call when it's not available it will trigger the modals
-  const now = new Date();
-  const exitDate = new Date(now.getTime() - 60 * 60 * 1000);
+  const exitDate = dayjs.utc().subtract(1, "hour");
+  const endTime = dayjs.utc(booking?.endTime);
 
   //find out if the meeting is in the past
-  const isPast = booking?.endTime <= exitDate;
```
Comment (Udit-takkar) [inline apps/web/pages/video/[uid].tsx:312]: This comparison might now work properly because booking.endTime is always in UTC but exitDate in browser's local time

Comment (emrysal) [reply apps/web/pages/video/[uid].tsx:312]: I don't think this is the problem because during this comparison there is a valueOf cast (and it compares the numeric offset from UTC 197..)

Comment (Udit-takkar) [reply apps/web/pages/video/[uid].tsx:312]: exitDate is already in UTC format "2024-01-03T05:49:44.703Z". I thought it's in local time format

PR #13004, opened 2024-01-03
feat: implement Flagsmith feature flags

### LINKED ISSUE #12598: [CAL-2775] Add ability to flight features

Our feature flags currently only support on/off. It would be great to have a feature flighting system where we can roll out new features to small subsets of users and/or particular users.

### Desired features:

* **Controlled Rollouts**: Develop a system that allows us to control the rollout of new features to specific user segments or a percentage of the user base.
* **Granular Targeting**: Provide the ability to target features to specific user attributes, such as user roles, geography, or any custom criteria relevant to our application.
* **Configuration Management**: Hook into our current configuration management system that enables easy toggling of features on and off without requiring code changes.
* **Metrics and Monitoring**: Integrate metrics and monitoring capabilities to track the performance and user engagement of features during flighting.
* **Rollback Mechanism**: Include a rollback mechanism to quickly revert the rollout of a feature in case of unexpected issues or negative user feedback.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2775](https://linear.app/calcom/issue/CAL-2775/add-ability-to-flight-features)</sub>

ibex088:
## What does this PR do?

Fixes #12598

used flagsmith which is an open-source Feature Flag Service that we can self-host as well, (currently using their hosted service) 

https://www.loom.com/share/d91d727549f74f2984a3327e00265878

Take a look and let me know if it aligns with the requirements, and i'll be happy to work on it further.


.env.example
```diff
@@ -329,4 +329,5 @@ APP_ROUTER_APPS_ENABLED=0
 APP_ROUTER_VIDEO_ENABLED=0
 
 # disable setry server source maps
-SENTRY_DISABLE_SERVER_WEBPACK_PLUGIN=1
\ No newline at end of file
+SENTRY_DISABLE_SERVER_WEBPACK_PLUGIN=1
+FLAGSMITH_ENVIRONMENT_ID=2KQbbhDBpJL6jxHhFYqbfa
```
PeerRich, line 333: you sure this should be in here?

SomayChauhan: > you sure this should be in here?

this would probably change in future, it is here just for the testing purpose
is it wrong? not very sure. :sweat_smile:

keithwillcode (commented):
This is looking cool, @SomayChauhan. Realize it's still a draft but I think a huge thing for us to get this integrated would be to not force self-hosters to use this. A big advantage to using something like Flagsmith is it gives us the ability to flight features, which we really desire but self-hosters may not need it. 

In those cases, I'd love for our current system to continue working as is. We should build this in such a way that we use something like the Factory pattern to return the correct provider based on whether or not the env has been configured to run Flagsmith. i.e. If not, use our existing feature flag system. If so, use Flagsmith.

SomayChauhan:
understood,
will start working on it further then!!!

zomars:
> @zomars do we wanna use this instead of our own?

As @keithwillcode  mentioned. We don't want to rely on third parties for this. It would be nice to opt-in if a hoster wants it. But it's not quite there in the current state.

SomayChauhan:
@keithwillcode @PeerRich @zomars 
can you take a look
to quickly setup and test:
* set `FLAGSMITH_ENVIRONMENT_ID=QLCiG9qqCJRfN34e6BAwjW` in .env
* go to http://13.233.201.155:8000/
* login with `caladmin@example.com` and `PasswordPassword@111` 
* and test on maybe insights page

if the FLAGSMITH_ENVIRONMENT_ID is not set it fetches flags from our Features table, otherwise it will fetch flags form flagsmith and save them in our existing `FeatureProvider`

SomayChauhan:
closing in favour of #13420

PR #13005, opened 2024-01-03
chore: e2e tests for oAuth flow

supalarry:
## What does this PR do?

1. Adds [e2e tests](https://github.com/calcom/cal.com/pull/13005/files#diff-422165fb6312839fa18ca19a56694ed0d5403d1c30660471b9973bcec145c1c3) for oAuth flow "oauth/authorize", "oauth/exchange" and "oauth/refresh" endpoints.
2. Refactors [oauth-flow.controller.ts](https://github.com/calcom/cal.com/pull/13005/files#diff-4ffbb95eedb8475fde14ad22e1b89de33d1cfc1a24f727ab63132792a017fef8) "/authorize" endpoint - if userId was already authorized for an oAuth client, then throw "BadRequestException" error explaining that. Alternatively, the request could proceed and return "id" aka authorize "code", but then it would be prone to getting existing codes using brute forcing. Next, in the "/exchange" endpoint rename "const bearerToken" to "const authorizeEndpointCode" to better explain the relationship that the code in the Authorization Bearer token has to come from the "/authorize" endpoint.
3. Refactors [tokens.repository.ts](https://github.com/calcom/cal.com/pull/13005/files#diff-f8a987668cfc396b4d9e47fb3de02e037f85bfb27b4e03cb2d37aae634abb4bb) - prior to this PR when creating access and refresh tokens and then re-creating them, "secret" of the keys were generated the same, but in prisma schema "AccessToken.secret" and "RefreshToken.secret" are unique fields, meaning that when re-creating access tokens using the "oauth/refresh" endpoint resulted in unique key constraint. To accommodate that add "iat" (issued at time in seconds, not milliseconds to match JWT specification) to the payload resulting in a unique secret each time.
4. Refactors [schema.prisma](https://github.com/calcom/cal.com/pull/13005/files#diff-56fa4d384ba6fb94b130e9eb42c5cad00734b025ebd6c197f76873469725ae87) "PlatformAuthorizationToken" model - if  the owner or client of the authorization token were to be deleted, the authorization token was not deleted, so cascade delete it.
5. Adds null checks for [get-user.decorator](https://github.com/calcom/cal.com/pull/13005/files#diff-7de2273ac638e08a574833d70cc18124b98e04b2eaf8946bdc4281b065cb8a27), [organization-roles.guard](https://github.com/calcom/cal.com/pull/13005/files#diff-e8d2c7f0be80ca4c67a2af4cec5c591055c1e9081753d3d2381fb49a5cb951ef).

### DIFF
--- a/apps/api/v2/src/modules/tokens/tokens.repository.ts
+++ b/apps/api/v2/src/modules/tokens/tokens.repository.ts
@@ -30,6 +30,15 @@ export class TokensRepository {
     });
   }
 
+  async getAuthorizationTokenByClientUserIds(clientId: string, userId: number) {
+    return this.dbRead.prisma.platformAuthorizationToken.findFirst({
+      where: {
+        platformOAuthClientId: clientId,
+        userId: userId,
+      },
+    });
+  }
+
   async createOAuthTokens(clientId: string, ownerId: number) {
     const accessExpiry = DateTime.now().plus({ days: 1 }).startOf("day").toJSDate();
     const refreshExpiry = DateTime.now().plus({ year: 1 }).startOf("day").toJSDate();
--- a/apps/api/v2/src/modules/tokens/tokens.repository.ts
+++ b/apps/api/v2/src/modules/tokens/tokens.repository.ts
@@ -43,10 +43,14 @@ export class TokensRepository {
     const accessExpiry = DateTime.now().plus({ days: 1 }).startOf("day").toJSDate();
     const refreshExpiry = DateTime.now().plus({ year: 1 }).startOf("day").toJSDate();
 
+    const issuedAtTime = Math.floor(Date.now() / 1000);
+
     const [accessToken, refreshToken] = await this.dbWrite.prisma.$transaction([
       this.dbWrite.prisma.accessToken.create({
         data: {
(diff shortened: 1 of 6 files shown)


apps/api/v2/src/modules/tokens/tokens.repository.ts
```diff
@@ -30,22 +30,37 @@ export class TokensRepository {
...
+        userId: userId,
+      },
+    });
+  }
+
   async createOAuthTokens(clientId: string, ownerId: number) {
     const accessExpiry = DateTime.now().plus({ days: 1 }).startOf("day").toJSDate();
     const refreshExpiry = DateTime.now().plus({ year: 1 }).startOf("day").toJSDate();
 
+    const issuedAtTime = Math.floor(Date.now() / 1000);
```
ThyMinimalDev, line 46: great Idea, why not just use Date.now() directly ?

supalarry: / 1000 to have it in seconds as required by jwt specification

supalarry: will create global jwt service in follow up PR

ThyMinimalDev (approved):

PR #13006, opened 2024-01-03
Basic atoms in barebone example platform apps

ThyMinimalDev:
## What does this PR do?

Platform Example app with first atoms


packages/platform/atoms/gcal-connect/GcalConnect.tsx
```diff
@@ -0,0 +1,37 @@
+import { cn } from "@/lib/utils";
+import type { FC } from "react";
+
+import { Button } from "@calcom/ui";
+import { CalendarDays } from "@calcom/ui/components/icon";
+
+import { useAtomsContext } from "../hooks/useAtomsContext";
+import { useGcal } from "../hooks/useGcal";
+
+interface GcalConnectProps {
```
supalarry, line 10: I think maybe it would be nice to add another prop - "alreadyConnectedClassName", so that connected Gcal stands out more instead of having the same style.

packages/platform/atoms/hooks/useGcal.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export interface useGcalProps {
+  isAuth: boolean;
+}
+
+export const useGcal = ({ isAuth }: useGcalProps) => {
+  const [allowConnect, setAllowConnect] = useState<boolean>(false);
+  const [checked, setChecked] = useState<boolean>(false);
+
+  const redirectToGcalOAuth = () => {
```
supalarry, line 13: All component functions should use `useEffect` to not be re-instantiated on each re-render.

PR #13008, opened 2024-01-03
fix: bad request in webhook route

### LINKED ISSUE #13007: [CAL-2857] Bad request on webhooks
there were some bad requests in webhooks logs in daily video

image

It's because of "max_participants" property is not sent in the request payload by daily video which used to be present before.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2857](https://linear.app/calcom/issue/CAL-2857/bad-request-on-webhooks)</sub>

Udit-takkar:
Fixes: https://github.com/calcom/cal.com/issues/13007

Made some keys optional that are not being used

### DIFF
--- a/apps/web/pages/api/recorded-daily-video.ts
+++ b/apps/web/pages/api/recorded-daily-video.ts
@@ -21,16 +21,16 @@ const schema = z
     id: z.string(),
     payload: z.object({
       recording_id: z.string(),
-      end_ts: z.number(),
+      end_ts: z.number().optional(),
       room_name: z.string(),
-      start_ts: z.number(),
+      start_ts: z.number().optional(),
       status: z.string(),
 
-      max_participants: z.number(),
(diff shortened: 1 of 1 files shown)


apps/web/pages/api/recorded-daily-video.ts
```diff
@@ -21,16 +21,16 @@ const schema = z
...
     payload: z.object({
       recording_id: z.string(),
-      end_ts: z.number(),
+      end_ts: z.number().optional(),
       room_name: z.string(),
-      start_ts: z.number(),
+      start_ts: z.number().optional(),
       status: z.string(),
 
-      max_participants: z.number(),
```
Udit-takkar, line 29: earlier "max_participants" was also present in the request payload but not now.

emrysal (approved):
👍

PR #13016, opened 2024-01-03
feat: abstracted jwt service

supalarry:
## What does this PR do?

Instead of registering JwtModule of "@nestjs/jwt" in module that wants to use JwtService of "@nestjs/jwt", we now have globally injectable internal service [JwtService](https://github.com/calcom/cal.com/pull/13016/files#diff-ff54dd7218b6dd045ff50dd3f73fa4a4eb7fbfad5e53fef0d3fa039b1974fdc5) with 3 methods:
1. sign - creates jwt token using payload
2. signAccessToken - attaches type: access_token to payload
3. signRefreshToken - attaches type: refresh_token to payload

All of them attach iat (issuedAtTime) to the tokens. The JwtService under the hood uses logic of JwtModule and JwtService of "@nestjs/jwt" and all of the tokens are signed with the same JWT_SECRET secret. This prevents scattering secrets used for signing tokens into using only JWT_SECRET to sign and also standardizes payload using issuetAtTime and ensuring that tokens are always unique.

[tokens.repository.ts](https://github.com/calcom/cal.com/pull/13016/files#diff-f8a987668cfc396b4d9e47fb3de02e037f85bfb27b4e03cb2d37aae634abb4bb) and [oauth-client.repository.ts](https://github.com/calcom/cal.com/pull/13016/files#diff-804bc052b99345b2ed8d149e97bd1b1005d08d475212b80aa21f767ba2e8cda3) have been updated to use the new service.

### DIFF
--- a/apps/api/v2/src/modules/jwt/jwt.service.ts
+++ b/apps/api/v2/src/modules/jwt/jwt.service.ts
@@ -1,0 +1,40 @@
+import { Injectable } from "@nestjs/common";
+import { JwtService as NestJwtService } from "@nestjs/jwt";
+
+@Injectable()
+export class JwtService {
+  constructor(private readonly nestJwtService: NestJwtService) {}
+
+  sign(payload: Record<any, any>) {
+    const issuedAtTime = this.getIssuedAtTime();
+
+    const token = this.nestJwtService.sign(JSON.stringify({ ...payload, iat: issuedAtTime }));
+
+    return token;
+  }
+
+  signAccessToken(payload: Record<any, any>) {
+    const issuedAtTime = this.getIssuedAtTime();
+
+    const accessToken = this.nestJwtService.sign(
+      JSON.stringify({ type: "access_token", ...payload, iat: issuedAtTime })
+    );
+
+    return accessToken;
+  }
+
+  signRefreshToken(payload: Record<any, any>) {
+    const issuedAtTime = this.getIssuedAtTime();
+
+    const accessToken = this.nestJwtService.sign(
+      JSON.stringify({ type: "refresh_token", ...payload, iat: issuedAtTime })
+    );
+
+    return accessToken;
+  }
+
+  getIssuedAtTime() {
+    // divided by 1000 because iat (issued at time) is in seconds (not milliseconds) as informed by JWT speficication
+    return Math.floor(Date.now() / 1000);
+  }
+}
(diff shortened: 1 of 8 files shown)


apps/api/v2/src/modules/jwt/jwt.service.ts
```diff
@@ -0,0 +1,40 @@
+import { Injectable } from "@nestjs/common";
+import { JwtService as NestJwtService } from "@nestjs/jwt";
+
+@Injectable()
+export class JwtService {
+  constructor(private readonly nestJwtService: NestJwtService) {}
+
+  sign(payload: Record<any, any>) {
```
ThyMinimalDev, line 8: is it possible to use unknown instead of any ?

apps/api/v2/src/modules/jwt/jwt.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  signAccessToken(payload: Record<any, any>) {
+    const issuedAtTime = this.getIssuedAtTime();
+
+    const accessToken = this.nestJwtService.sign(
+      JSON.stringify({ type: "access_token", ...payload, iat: issuedAtTime })
+    );
+
+    return accessToken;
```
ThyMinimalDev, line 23: I think you can directly use `return this.sign({ type: "access_token", ...payload})`

apps/api/v2/src/modules/jwt/jwt.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  signRefreshToken(payload: Record<any, any>) {
+    const issuedAtTime = this.getIssuedAtTime();
+
+    const accessToken = this.nestJwtService.sign(
+      JSON.stringify({ type: "refresh_token", ...payload, iat: issuedAtTime })
+    );
+
+    return accessToken;
```
ThyMinimalDev, line 33: `return this.sign({ type: "refresh_token", ...payload})`

apps/api/v2/src/modules/jwt/jwt.service.ts
supalarry: Great idea! Done in [implement Morgans requests](https://github.com/calcom/cal.com/pull/13016/commits/4c20f7dfcde72243a985522e923670b553af668c) + also stopped using JSON.stringify since nest's jwt service sign accepts object directly.

apps/api/v2/src/modules/jwt/jwt.service.ts
supalarry: Great idea! Done in [implement Morgans requests](https://github.com/calcom/cal.com/pull/13016/commits/4c20f7dfcde72243a985522e923670b553af668c) + also stopped using JSON.stringify since nest's jwt service sign accepts object directly.

ThyMinimalDev (approved):

PR #13039, opened 2024-01-04
chore: [app-router-migration 17]: migrate payments page

dmytro068:
## What does this PR do?

migrate payments page

## Requirement/Documentation

## Type of change

## Mandatory Tasks


apps/web/middleware.ts
```diff
@@ -118,6 +118,7 @@ export const config = {
     "/apps/routing_forms/:path*",
     "/event-types",
     "/future/event-types/",
+<<<<<<< HEAD
```
keithwillcode, line 121: This git data needs to be removed

packages/app-store/apps.keys-schemas.generated.ts
```diff
@@ -55,6 +56,7 @@ export const appKeysSchemas = {
   make: make_zod_ts,
   matomo: matomo_zod_ts,
   metapixel: metapixel_zod_ts,
+  "mock-payment-app": mock_payment_app_zod_ts,
```
keithwillcode, line 59: I don't think we should have this in the actual list of apps in the app store. We need to put this in a place that's specific to the E2E tests

keithwillcode (changes requested):

apps/web/middleware.ts
hbjORbj: Addressed

packages/app-store/apps.keys-schemas.generated.ts
hbjORbj: Addressed

keithwillcode (changes requested):
Type checks are now failing

keithwillcode (approved):

PR #13042, opened 2024-01-04
chore: [app-router-migration 18] Migrate "/settings/organizations/*" pages

hbjORbj:
## What does this PR do?

- This PR migrates the `/settings/organizations/*` page group to the app directory (which runs under the App Router).
- The migrated page exists under `/future` and is accessible at all times.

## Requirement/Documentation

## Type of change

## Mandatory Tasks


keithwillcode (dismissed):
Changes generally look good but type checks are failing and the build fails as well.

keithwillcode (approved):

PR #13048, opened 2024-01-05
fix: Avoid duplicate inline embeds

### LINKED ISSUE #12983: [CAL-3366] Avoid duplication of inline embed for same namespace
### Issue Summary

On my website (https://disigner.nl/) I've implemented the Cal.com inline embed in a popup. The website is build with Wordpress and Elementor. In Elementor Popup builder the embed looks fine https://gyazo.com/0e6d4325c996ee662eef5d9e4bf21d96. But when I try it on the frontend on top of the embed there is a loading animation https://gyazo.com/3ec607bab36272e305750c44bd677f8c. When I go to the code it looks like the <cal-inline> loads 2 times. 

### Steps to Reproduce

Here is a video: https://gyazo.com/069619a002c97612ee4be32e2d2a8f2d

1. Pressed the yellow phone button in the header
2. A popup shows with 2 loading animations
3. The bottom animation loads but the top does not, which should not be there in the first place.

### Actual Results

- 2 loading animations. It seems that somehow the <cal-inline> is implemented 2 times when looking at it in the front-end.

### Expected Results

- When pressing the popup the inline embed should load only 1 time.

### Technical details

- Wordpress 6.4.2
- Elementor 3.18.2
- Elementor Pro 3.18.1
- WP Rocket 3.15.6
- PHP version: 8.1.24

### Evidence
https://gyazo.com/069619a002c97612ee4be32e2d2a8f2d

<sub>[CAL-3366](https://linear.app/calcom/issue/CAL-3366/avoid-duplication-of-inline-embed-for-same-namespace)</sub>

hariombalhara:
## What does this PR do?

Fixes #12983

## Type of change

## How should this be tested?
- Simply execute inline embed twice. See the issue in production in Codesandbox [here](https://codesandbox.io/p/sandbox/vanillajs-week-view-forked-hp3j2c?file=%2Findex.html%3A54%2C39)

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/embeds/embed-core/src/embed.ts
+++ b/packages/embeds/embed-core/src/embed.ts
@@ -439,6 +439,10 @@ class CalApi {
     elementOrSelector: string | HTMLElement;
     config?: PrefillAndIframeAttrsConfig;
   }) {
+    if (this.cal.inlineEl) {
+      console.warn("Inline embed already exists. Ignoring this call");
+      return;
+    }
     // eslint-disable-next-line prefer-rest-params
     validate(arguments[0], {
       required: true,


hariombalhara:
* `main`
  * **PR #13048** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/13048?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/13048?utm_source=stack-comment).

packages/embeds/embed-core/src/embed.ts
```diff
@@ -439,6 +439,10 @@ class CalApi {
     elementOrSelector: string | HTMLElement;
     config?: PrefillAndIframeAttrsConfig;
   }) {
+    if (this.cal.inlineEl) {
```
hariombalhara, line 442: Every namespace has his own cal instance, so we just check if the instance has inlineEl already defined.

emrysal (approved):
Nice, simple fix.

xta:
This is breaking my usage of the embed. After opening the embed with a modal, closing the modal, and re-opening the modal, I'm getting the "fix" or error `Inline embed already exists. Ignoring this call`.

When I look at the DOM, I don't see the embed in my re-opened modal body.

xta:
I was able to work around this, but I think there is a bug. The `Cal.instance.inlineEl` keeps a reference to the `template.content.children[0]` even though that part of the DOM no longer exists (since React removed it in my app). https://github.com/calcom/cal.com/blob/main/packages/embeds/embed-core/src/embed.ts#L494C25-L494C53

This prevents the component from creating the iframe in the future when it should be able to.

cc @hariombalhara

itsfaraaz:
> I was able to work around this, but I think there is a bug. The `Cal.instance.inlineEl` keeps a reference to the `template.content.children[0]` even though that part of the DOM no longer exists (since React removed it in my app). https://github.com/calcom/cal.com/blob/main/packages/embeds/embed-core/src/embed.ts#L494C25-L494C53
> 
> This prevents the component from creating the iframe in the future when it should be able to.
> 
> cc @hariombalhara

How'd you work around this? Can't figure this out

xta:
with my setup, I can access global `window.Cal.instance.inlineEl`

PR #13053, opened 2024-01-05
feat: user schedule management

supalarry:
This PR adds [schedules.controller.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-831f89f78a1bca4514e7a14a19c32fd673be76fdb8d1da9ec0fa5128a60627f4) to manage user schedules.

Controller's inputs are [create-schedule.input.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-e9f0720f808424995ac8329b33eabf4028111c42ed7b9b0c140ec3d8d04aed97) and [update-schedule.input.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-dbfa53be84a36125723c4ae75ca7882fbb93d063d2625648f00a1f0f0dcbd9fe). Since schedule consists of availabilities, the schedule inputs have a nested [create-availability.input.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-11fdeed6bf8243c3e4ee96956947584e0eb30a02470d380a8bb1189eb81f3ca2). Unlike in the v1 API, there are no availabilities endpoints but everything is managed via schedules endpoints that accept availabilities as endpoints. Notably, none of the schedule inputs have "isDefault" as a parameter - when a new schedule is created it is set as the user's default one in "SchedulesService.createUserSchedule". I omitted it to simplify incoming props and I am assuming that for now people will need just 1 schedule. Finally, [timezones.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-6d8d1a4eb5605751efabfad17f7ef434693f10db34ce06bb725e567ed2ea0160) define type for allowed TimeZone by the schedule inputs. The timezones are a list of the possible timezones when using dropdown in our Call app to select timezone, so they are the same.

The controller invokes [schedules.service.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-14c77b9997084c683358013e17747bf88a650455a2c5e35243b060a952c309c2). The service uses [availabilities.service.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-1dc70ee3a5b39d33af3d9dc1513ea7d505d553e77b52eaf2d55300c02170041a) and [schedules.repository.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-1d22143416e92a38d7ed9ba36d04f19b8e58cd2307fd42b54106591bbb0f11f7).

Since schedules repository returns nested availability objects, we use [response.ts](https://github.com/calcom/cal.com/pull/13053/files#diff-17434bf0ccb55ed32a5a7cfe226b2521c1e5e62599da98eb524036797130342d) zod schema to parse response - only keep properties required.


apps/api/v2/src/ee/schedules/schedules.repository.ts
```diff
@@ -0,0 +1,133 @@
...
+              startTime: availability.startTime,
+              endTime: availability.endTime,
+              userId,
+            };
+          }),
+        },
+      };
+    }
+
+    const createdSchedule = await this.dbWrite.prisma.schedule.create({
```
exception, line 43: Could you return the Promise here rather than awaiting, and having the function call be the one awaiting?

packages/platform/constants/timezones.ts
```diff
@@ -0,0 +1,337 @@
+export type TimeZone = (typeof TIMEZONES)[number];
+
+const TIMEZONES = [
```
exception, line 3: Was this taken from the main app?

apps/api/v2/src/ee/schedules/inputs/create-schedule.input.ts
```diff
@@ -0,0 +1,18 @@
...
+  @IsString()
+  name!: string;
+
+  @IsString()
+  timeZone!: TimeZone;
+
+  @ValidateNested({ each: true })
+  @Type(() => CreateAvailabilityInput)
+  @IsOptional()
+  availabilities?: CreateAvailabilityInput[];
```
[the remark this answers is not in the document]
supalarry: Good catch! Done in [schedule inputs availabilities as array](https://github.com/calcom/cal.com/pull/13053/commits/3ed296e741aafd8ac9b35d05b09fc2d02308b358)

apps/api/v2/src/ee/schedules/schedules.repository.ts
supalarry: What is the benefit of doing that?

As far as I understand there is no need - "this.dbWrite.prisma.schedule.create" is within an "async createScheduleWithAvailabilities" which is an async function, which means it always returns a promise, so it's already up to the caller of "createScheduleWithAvailabilities" whether or not to await it.

apps/api/v2/src/ee/schedules/services/schedules.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    private readonly availabilitiesService: AvailabilitiesService,
+    private readonly usersRepository: UsersRepository
+  ) {}
+
+  async createUserSchedule(userId: number, schedule: CreateScheduleInput) {
+    const availabilities = schedule.availabilities?.length
+      ? schedule.availabilities
+      : [this.availabilitiesService.getDefaultAvailabilityInput()];
+
+    const createdSchedule = await this.schedulesRepository.createScheduleWithAvailabilities(
```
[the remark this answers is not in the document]
supalarry: Replied in previous comment

apps/api/v2/src/ee/schedules/schedules.repository.ts
supalarry: So its essentially the same thing

packages/platform/constants/timezones.ts
supalarry: Yes. “packages/ui/components/form/timezone-select/TimezoneSelect.tsx” invokes "cityTimezones.useQuery" which returns "data" which is then passed to "addCitiesToDropdown" which returns "cityTimezones" which are passed to "timezones" of the dropdown. I logged "cityTimezones" and these are values that are possible in the dropdown when selecting / writing timezones.

PR #13061, opened 2024-01-05
chore: [app-router-migration 27] migrate the "signup" page

grzpab:
## What does this PR do?
- This PR migrates the `/signup` page to the app directory (which runs under the App Router).
- The migrated page exists under `/future/signup` and is accessible at all times.
- We support a redirect from the former to the latter with the middleware and environmental variables, as done previously.

## Type of change

## How should this be tested?
There should be no changes between the `/signup` and `/future/signup` in terms of performance and content (on Lighthouse both pages get near 100 scores).

If the ```APP_ROUTER_SIGNUP_ENABLED``` is set to `1` and ```AB_TEST_BUCKET_PROBABILITY``` is set to `100`, then the future signup page should be rendered in place of the `/signup` page WITHOUT a visible redirect. We can see it by inspecting the existence of the `x-calcom-future-routes-enabled` cookie.

## Mandatory Tasks

### DIFF
--- a/apps/api/pages/api/slots/_get.ts
+++ b/apps/api/pages/api/slots/_get.ts
@@ -22,7 +22,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     const { usernameList, ...rest } = req.query;
     let slugs = usernameList;
     if (!Array.isArray(usernameList)) {
-      slugs = usernameList ? [usernameList] : [];
+      slugs = usernameList ? [usernameList] : undefined;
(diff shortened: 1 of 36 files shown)


PeerRich:
whats needed here to leave draft @grzpab

hbjORbj:
@PeerRich Sorry for the delay. We got it ready now. @grzpab will turn it ready for review soon.

zomars (approved):
LGTM

PR #13071, opened 2024-01-07
fix: don't allow uid update in bookings update api

### LINKED ISSUE #13070: [CAL-2919] Update bookings api allow to update uid
### Issue Summary

Update bookings api allow to update uid. I don’t see any usecase where this would be required. And as it is uid, it must be immutable.

### Steps to Reproduce

Issue 4 1

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2919](https://linear.app/calcom/issue/CAL-2919/update-bookings-api-allow-to-update-uid)</sub>

riddhesh-mahajan:
## What does this PR do?

This PR prevents uid update in booking api

Fixes #13070 

Issue 4 1 Sol 1


PeerRich (dismissed):

apps/api/lib/validations/booking.ts
```diff
@@ -6,7 +6,6 @@ import { extendedBookingCreateBody, iso8601 } from "@calcom/prisma/zod-utils";
 import { schemaQueryUserId } from "./shared/queryUserId";
 
 const schemaBookingBaseBodyParams = Booking.pick({
```
keithwillcode, line 8: Instead of updating this object which is the base object, let's instead update the `schemaBookingEditBodyParams` to omit the `uid` field when merging with this.

keithwillcode (changes requested):

keithwillcode (approved):

PR #13074, opened 2024-01-07
fix: improve slug filtering to prevent same slug with diff case

### LINKED ISSUE #13073: Create team api is accepting invalid slug
### Issue Summary

Create team API allows a slug to be passed as input. If the exact same slug already exists in the database, it returns an error. However, it accepts a slug with the same spelling but different case.

For example, I have created two teams: one using a free user's API key and one with a pro user's API key.

The free user created the team with the slug "myteam," and the pro user created "Myteam." Both were accepted by the API. However, when the pro user tries to open the team's event page, it actually opens the free user's page.

We should restrict this because, for the pro user, the team is created but is useless. They cannot open the events page, and any booking link will return a 404 error.

### Steps to Reproduce
Issue 5-1 1
Issue 5-1 2
Issue 5-1 3Free user events
Issue 5-1 4 Pro user events

riddhesh-mahajan:
## What does this PR do?

This PR adds insensitive match for team slug for better slug availability logic.

Fixes #13073 

## Type of change

## How should this be tested?
Create 2 teams using api call. Both must have same slug with different case.


alishaz-polymath (approved):
Thank you for the contribution! 🙌

PR #13078, opened 2024-01-07
fix: add check for parentId

### LINKED ISSUE #13077: Team API permits the updating of the parentId
### Issue Summary

The Update Team API permits the updating of the parentId without any permission checks. In this instance, I updated a team belonging to a free user and used the team ID of a pro user as the new parentId. It's essential to note that the free user and their team were not previously associated with the pro user.

As a result of this update, the team now possesses the subdomain of the pro user's team.

For instance, if we consider a team with 10 subteams and a slug of "deel," all other 10 subteams' URLs will now feature "[deel.cal.com](http://deel.cal.com/)." However, if an unrelated user updates their team's parentId with the ID of the "deel" team, they will also receive "deel" as their subdomain.

### Steps to Reproduce

Issue 5-3 1
Issue 5-3 2

riddhesh-mahajan:
## What does this PR do?

This PR adds parentId check to prevent use of non related parentId in update team api. 

Fixes #13077 

Issue 5-3 sol1
Issue 5-3 sol2

## Type of change


emrysal (approved):
Thank you for this 🙏  - approved.

PR #13106, opened 2024-01-08
fix: Button re-enables on sluggish networks whilst redirecting

emrysal:
## What does this PR do?

* Fixes a potential window for spam clicking on slow networks between the createBooking mutation finishing and the redirect happening.
* Fixes a potential window for spam clicking also on slow networks between the button click and form validation success - only when the phone number is enabled as this is the point the libphonenumber library is downloaded.


packages/features/bookings/Booker/components/BookEventForm/BookEventForm.tsx
```diff
@@ -200,7 +200,7 @@ export const BookEventFormChild = ({
       const { uid, paymentUid } = responseData;
       const fullName = getFullName(bookingForm.getValues("responses.name"));
       if (paymentUid) {
-        return router.push(
+        router.push(
```
emrysal, line 203: Be very clear that router.push returns a void value. It's possible this wasn't historically and this has changed with more recent Next.js versions.

joeauyeung (approved):
LGTM, talked about this already.

keithwillcode (approved):

PR #13113, opened 2024-01-09
fix: Added disabled prop in save button

soham2k06:
EventTypeSingleLayout.tsx
If no input field is changed then 'save' button will be disabled.

## What does this PR do?

If no input field is changed then 'save' button will be disabled.

Fixes # (issue)

It fixes the bug where used could click save button even if no field is updated in event. It reduces fetch requests.

## Type of change

## Mandatory Tasks


PeerRich (approved):

emrysal (changes requested):
Got some failing e2e tests; looks like the button does not enable when making changes to Stripe (Price, in this case)

Udit-takkar (commented):
@soham2k06 i don't see any change. I haven't changed anything but i see save button as enabled

https://github.com/calcom/cal.com/assets/53316345/72988d82-73cd-482c-b7d5-b175b74581f4

Udit-takkar (changes requested):
Sometimes the button is disabled and sometimes it isn't.

Uploading Screen Recording 2024-01-09 at 8.35.50 PM.mov…

Ryukemeister (commented):
Tested this locally and works as expected for me. However the issue that @Udit-takkar you're facing about sometimes the button being disabled and sometimes not could arise from the fact that we might have missed providing some of our inputs a default values at the time of initializing our useForm, which is what the docs suggest [here](https://www.react-hook-form.com/api/useform/formstate/) . @soham2k06 can you just check once if we're providing all of our useForm inputs some default values?

Udit-takkar:
This PR https://github.com/calcom/cal.com/pull/13292#pullrequestreview-1875797381 should fix the issue

PR #13122, opened 2024-01-09
fix: event reschedule workflow trigger

### LINKED ISSUE #12881: [CAL-2829] RESCHEDULE_EVENT workflow trigger not working for event type that require confirmation
### Issue Summary

When an attendee reschedules a booking that requires confirmation (won't be automatically confirmed), the 'when event is rescheduled' workflow trigger is not triggered. 

### Steps to Reproduce

1. Create workflow with trigger 'when event is rescheduled' 
2. Set active on an event type that requires confirmation 
3. Book event type 
4. Reschedule event type (not with organizer, do it in icognito)
5. This should now trigger workflow but it doesn't

<sub>[CAL-2829](https://linear.app/calcom/issue/CAL-2829/reschedule-event-workflow-trigger-not-working-for-event-type-that)</sub>

Amit91848:
## What does this PR do?

Fixes #12881 

## Type of change

## Mandatory Tasks


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2015,7 +2015,9 @@ async function handler(
         workflows: eventType.workflows,
         smsReminderNumber: smsReminderNumber || null,
         calendarEvent: { ...evt, ...{ metadata, eventType: { slug: eventType.slug } } },
-        isNotConfirmed: evt.requiresConfirmation || false,
+        isNotConfirmed: rescheduleUid
+          ? originalRescheduledBooking?.status !== BookingStatus.ACCEPTED
```
CarinaWolli, line 2019: I think we should make it `false` in case of a reschedule

Even when the booking is not yet confirmed we want the reschedule workflow to trigger, let me know if you think differently about that 🙏

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2792,7 +2794,9 @@ async function handler(
       workflows: eventType.workflows,
       smsReminderNumber: smsReminderNumber || null,
       calendarEvent: evtWithMetadata,
-      isNotConfirmed: !isConfirmedByDefault,
+      isNotConfirmed: rescheduleUid
+        ? originalRescheduledBooking?.status !== BookingStatus.ACCEPTED
```
CarinaWolli, line 2798: same as above

CarinaWolli (changes requested):

packages/features/bookings/lib/handleNewBooking/test/reschedule.test.ts
```diff
@@ -850,7 +850,7 @@ describe("handleNewBooking", () => {
               },
             });
 
-            // expectWorkflowToBeTriggered({emails, organizer});
+            expectWorkflowToBeTriggered({ emails, organizer });
```
CarinaWolli, line 853: I uncommented this test, as this passes now 👏🏻

packages/features/bookings/lib/handleNewBooking.ts
Amit91848: Yeah that works too I will make the changes.

CarinaWolli (approved):

PR #13123, opened 2024-01-09
fix: scheduling/canceling workflow emails in cron api call

CarinaWolli:
## What does this PR do?

Set SENDGRID_API_KEY for Sengrid client and sgMail in corn job to schedule and cancel email reminders.


Udit-takkar (approved):
code LGTM

PR #13127, opened 2024-01-09
feat: Passkey support

### LINKED ISSUE #13342: [CAL-2984] Add Passkey support
Implemented in [https://github.com/calcom/cal.com/pull/13127](https://github.com/calcom/cal.com/pull/13127)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2984](https://linear.app/calcom/issue/CAL-2984/add-passkey-support)</sub>

merlinaudio:
## What does this PR do?
Fixes https://github.com/calcom/cal.com/issues/13342
Fixes CAL-2984

This PR adds passkey support for logging in through Hanko Passkeys.

## Type of change

- This change requires a documentation update (probably)

## How should this be tested?

- Go to [cloud.hanko.io](https://cloud.hanko.io) and create a "Passkey API" project
  - Then create a new API key
    set it to `HANKO_PASSKEYS_API_KEY="<api key>"`
    (quotes important)
  - Copy the tenant ID to `NEXT_PUBLIC_HANKO_PASSKEYS_TENANT_ID=<tenant id>`
- Log in as normal (with password/google/...)
- Go to settings > passkeys > create a new passkey
- Log out
- Go to login > sign in with a passkey

## Mandatory Tasks

## next-auth-options.ts

**Please review this file, there are substantial changes that might be unsafe.** I am not sure if the file in it's current form is okay:

This PR changes the very important `authorize` function. I'm not sure if we want to keep it like this, but the authorize function itself does a lot of checks, only some relevant to `credentials`.

Instead of duplicating it, I added a second parameter, `unsafe`, intended to only be used from server code like in the `PasskeyProvider`. Using it is very explicit, you have to pass `undefined` as a first parameter, so it's clear that you shouldn't be using this functionality unless you're sure what you're doing:

```ts
// Example usage of authorize function
const sessionData = authorize(undefined, { unsafe: { userId: 123 }})
```

When `unsafe` is used, **all the checks dependent on `credentials` are not being run anymore.**

I am not sure how safe this is, or if a larger refactor is needed that properly separates checks dependent on `credentials` from all other checks. (That would certainly be cleaner, though)

## Checklist

### Adds passkey button to login page:
CleanShot 2024-01-09 at 17 36 45@2x

### Adds "Passkeys" settings section to sidebar:
CleanShot 2024-01-09 at 17 44 38@2x

### Adds passkeys settings page

**Empty:**
CleanShot 2024-01-09 at 17 45 45@2x
- Is the icon alright here? it looks somewhat more chunky than the "link" icon used e.g. for event types and api keys
- On the other hand, it's the same icon we're using for passkeys everywhere (including the login page "Sign in with a passkey" button

---

Clicking on the "add" button and "sign in with a passkey" buttons opens a browser dialog. This looks different for every browser, phone, etc.

**For example:**
CleanShot 2024-01-09 at 17 47 48@2x

**With passkeys added:**
CleanShot 2024-01-09 at 17 49 38@2x
- The name is set by the Hanko Passkeys API (our backend). We'll improve the name soon so it's not just a semi-random string and more user friendly
- If the passkey was used, the description says e.g. "Created 3 days ago • Last used 5 minutes ago" 

**Dropdown** stolen straight from API settings page:
CleanShot 2024-01-09 at 17 53 13@2x


PeerRich (commented):
hell yeah passkeys babyyyy

excited for this, will review

apps/web/pages/settings/security/passkeys.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const remove = useMutation({
+    mutationFn: () =>
+      fetch("/api/auth/passkeys/remove", {
+        method: "DELETE",
+        body: JSON.stringify({ credentialId: props.id }),
+        headers: { "Content-Type": "application/json" },
+      }).then((res) => res.json()),
+    onSuccess: props.onRemoved,
+    onError: (e: any) => {
+      showToast(e.message || t("something_went_wrong"), "error");
```
merlindru, line 58: Could also replace with

```ts
onError: (e: unknown) => {
    showToast(typeof e === "object" && e != null && "message" in e && typeof e.message === "string" ? e.message : t("something_went_wrong"), "error");
```

packages/features/settings/layouts/SettingsLayout.tsx
```diff
@@ -55,6 +55,7 @@ const tabs: VerticalTabItemProps[] = [
       { name: "password", href: "/settings/security/password" },
       { name: "impersonation", href: "/settings/security/impersonation" },
       { name: "2fa_auth", href: "/settings/security/two-factor-auth" },
+      { name: "passkeys", href: "/settings/security/passkeys" },
```
merlindru, line 58: This still makes it appear *above* SSO. Is this position fine?

If we want it to be the last entry in the list, will have to add it below

apps/web/components/auth/PasskeyIcon.tsx
merlindru, line 1: Could also move this to an .svg file if preferred

PeerRich:
with this PR, can people also make an entire new account with passkeys? or just existing users?

packages/features/settings/layouts/SettingsLayout.tsx
PeerRich: positioning is fine

apps/web/components/auth/PasskeyIcon.tsx
PeerRich: this is fine, however, ideally you move this into calcom/components/icon

PeerRich:
added HANKO_PASSKEYS_API_KEY and NEXT_PUBLIC_HANKO_PASSKEYS_TENANT_ID to vercel

keithwillcode:
Moving to 4.0 based on current priorities.

alishaz-polymath:
Thank you for the PR 🙏 
Hpwever, upon further investigation we will go with nextauth passkeys implementation for now: https://authjs.dev/reference/core/providers/passkey#resources

zomars:
We had a call between Felix and @merlindru from the Hank team and  @keithwillcode and myself. We're expecting a new PR with a much simpler approach in the upcoming weeks.

Issue #13128, opened 2024-01-09
[CAL-2887] make sure all settings <Meta> has borderInShellHeader props


PeerRich:
and p-4 and border subtle for the content:

## like this

before (old layout):

![CleanShot 2024-01-09 at 17 17 14@2x](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/8d3878d0-1150-45e2-8c30-d9ccf4beff8b/df3300f6-54e0-4e9e-a545-0d730154528f?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2U4NmJmOTU3LWQ4MmYtNDY1ZS1iMjA1LTEzNTU1OWY0YjYyMy84ZDM4NzhkMC0xMTUwLTQ1ZTItOGMzMC1kOWNjZjRiZWZmOGIvZGYzMzAwZjYtNTRlMC00ZTllLWE1NDUtMGQ3MzAxNTQ1MjhmIiwiaWF0IjoxNzA0ODIwNzE0LCJleHAiOjE3MDQ5MDcxMTR9.EmMC27kcTQlJgyFKpSQlh-I9Ugz7xl7vhcNTred3-M8)

after (new layout):

![CleanShot 2024-01-09 at 17 18 01@2x](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/213e0f90-1432-4e28-bd53-cf37e6a2374e/a33d55e4-4e54-4f61-8054-18a436be2ec8?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2U4NmJmOTU3LWQ4MmYtNDY1ZS1iMjA1LTEzNTU1OWY0YjYyMy8yMTNlMGY5MC0xNDMyLTRlMjgtYmQ1My1jZjM3ZTZhMjM3NGUvYTMzZDU1ZTQtNGU1NC00ZjYxLTgwNTQtMThhNDM2YmUyZWM4IiwiaWF0IjoxNzA0ODIwNzE0LCJleHAiOjE3MDQ5MDcxMTR9.PbvysK8JZfkxT98dej2QDnvvvtNBuEWJA9dwRrSAER0)

[...]

PR #13129, opened 2024-01-09
fix: added border to calendar view

Gmin2:
Fixes # [CAL-2887]
/claim #13128


PeerRich:
the task is to add this to every missing settings page

PeerRich (changes requested):
the task is to add this to every missing settings page

also, make sure that it looks good (check other pages)

i.e. no border at the top, test all EmptyScreens etc.

Min2who:
ok working on it

PR #13146, opened 2024-01-10
fix: Payment Page not showing

hariombalhara:
Fixes e2e and the issue of payment page not showing(or we can say showing for a very fleeting moment)


emrysal (approved):
LGTM, that's a good spot.

PR #13158, opened 2024-01-10
fix: disable save button in event-type page when no change in form

### LINKED ISSUE #13112: [bug] events can be saved even if it is as same as before

### Issue Summary

User can update already created events, but we can click 'save' button even if user hasn't updated any of fields.

### Steps to Reproduce

1. Create event if don't have any.
2. Navigate to the event
3. You will be able to save and it will really make a fetch request even if you haven't updated any field

### Actual Results

Right now, save button is working even if any field is not updated yet.

### Expected Results

Save button should be disabled if none of fields is updated yet.

### Evidence

https://github.com/calcom/cal.com/assets/118199354/e06c2959-9359-4f76-99c7-250a8e98bda1

moulibrota-das:
## What does this PR do?

Previously the save button in the event-types page was always not disabled and can be clicked even if no changes in the form are made.
This PR disable the save button until a change occurs in the form data.

Fixes #13112 

[Screencast from 10-01-24 09:56:06 PM IST.webm](https://github.com/calcom/cal.com/assets/88196189/61921897-f90e-4fe5-919b-141fa1b22db3)

## Type of change


apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -548,6 +548,8 @@ const EventTypePage = (props: EventTypeSetupProps) => {
       clearTimeout(timeout);
     };
   }, []);
+
+  const [disableSave, setDisableSave] = useState<boolean>(true);
```
Comment (Ryukemeister) [inline apps/web/pages/event-types/[type]/index.tsx:552]: @moulibrota-das instead of creating a state to track form changes how about we use the watch api https://react-hook-form.com/docs/useform/watch instead?

soham2k06 (approved):

soham2k06 (approved):

soham2k06 (changes requested):
Great job! but it seems that if you undo your changes, you will be able to click 'save' button.
eg. title saved already: 'test'
if you updated it with 'test 1' and undo it and set it back to 'test' then you will be able to save.

Udit-takkar (changes requested):
@moulibrota-das I can't add a location

https://github.com/calcom/cal.com/assets/53316345/4250ffed-4011-4f67-839e-6724a7c0350b

The button is still active and clickable even after event type has been saved successfully

https://github.com/calcom/cal.com/assets/53316345/06ed63d5-c511-4a38-8201-becf9db79043

Your approach doesn't disable the button after saving or when the i reverse all the changes.

moulibrota-das:
I am changing my previous approach and also looking into the useForm docs. Thinking about storing the initial and current form state and comparing them in an useEffect hook to update the save button.

keithwillcode:
Converting to draft while feedback and conflicts are addressed.

PR #13161, opened 2024-01-11
fix: Bundle analysis message

keithwillcode:
## What does this PR do?

Attempting to fix an error we are now seeing where the body can't be found to write into the PR comment.

## Type of change

## How should this be tested?

- After merge, ensure bundle analysis succeeds on PRs

## Mandatory Tasks


emrysal (approved):
That does look like an error.

PR #13167, opened 2024-01-11
fix: Use constants.ts for brand colours > db DEFAULT

sean-brydon:
## What does this PR do?
Removes the default and not null constraint in the users -> brandColor and darkBrandColour
This allows us to rely on the values defined in constants to use APP wide. This is useful for self hosted instances. 

## How should this be tested?
Start up the app and run the migrations
Change the value of the brand colours in constants.ts and notice the app wide colours change.

### DIFF
--- a/packages/lib/getBrandColours.tsx
+++ b/packages/lib/getBrandColours.tsx
@@ -1,7 +1,9 @@
 import { useBrandColors } from "@calcom/embed-core/embed-iframe";
 
-const BRAND_COLOR = "#292929";
-const DARK_BRAND_COLOR = "#fafafa";
+import { DEFAULT_DARK_BRAND_COLOR, DEFAULT_LIGHT_BRAND_COLOR } from "./constants";
+
+const BRAND_COLOR = DEFAULT_LIGHT_BRAND_COLOR;
+const DARK_BRAND_COLOR = DEFAULT_DARK_BRAND_COLOR;
 
 type Rgb = {
   r: number;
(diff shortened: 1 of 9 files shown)


packages/prisma/migrations/20240111075727_remove_brand_colours_default/migration.sql
```diff
@@ -0,0 +1,17 @@
...
+ALTER TABLE "users" ALTER COLUMN "brandColor" DROP NOT NULL,
+ALTER COLUMN "brandColor" DROP DEFAULT,
+ALTER COLUMN "darkBrandColor" DROP NOT NULL,
+ALTER COLUMN "darkBrandColor" DROP DEFAULT;
+
+UPDATE "Team" SET "brandColor" = NULL WHERE "brandColor" = '#292929';
+UPDATE "Team" SET "darkBrandColor" = NULL WHERE "darkBrandColor" = '#fafafa';
+
+UPDATE "users" SET "brandColor" = NULL WHERE "brandColor" = '#292929';
+UPDATE "users" SET "darkBrandColor" = NULL WHERE "darkBrandColor" = '#fafafa';
```
sean-brydon, line 16: Set our brand colours to NULL -> we will only now set this if a user overrides these in settings

packages/lib/getBrandColours.tsx
```diff
@@ -1,7 +1,9 @@
 import { useBrandColors } from "@calcom/embed-core/embed-iframe";
 
-const BRAND_COLOR = "#292929";
-const DARK_BRAND_COLOR = "#fafafa";
+import { DEFAULT_DARK_BRAND_COLOR, DEFAULT_LIGHT_BRAND_COLOR } from "./constants";
+
+const BRAND_COLOR = DEFAULT_LIGHT_BRAND_COLOR;
+const DARK_BRAND_COLOR = DEFAULT_DARK_BRAND_COLOR;
```
sean-brydon, line 6: Use the values from constants -> These colours get loaded here https://github.com/calcom/cal.com/pull/13167/files#diff-abbbafbce3959191cbdfce95ed71dc38466b75d1695dd78dc81a66e4117e1d76L171-L172 if there is no brand colour passed in

Udit-takkar (approved):
LGTM

PR #12513, opened 2023-11-23
refactor: Handling Invalid GCal Credentials

joeauyeung:
## What does this PR do?

Fixes # (issue)

## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is ab UI/UX design document, please, share it here.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

- Are there environment variables that should be set?
- What are the minimal test data to have?
- What is expected (happy path) to have (input and output)?
- Any other important info that could help to test that PR

## Mandatory Tasks

- [ ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist

- I haven't read the [contributing guide](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md)
- My code doesn't follow the style guidelines of this project
- I haven't commented my code, particularly in hard-to-understand areas
- I haven't checked if my PR needs changes to the documentation
- I haven't checked if my changes generate no new warnings
- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes

PR #13189, opened 2024-01-11
fix: When declining OAuth return to previous page

### LINKED ISSUE #12277: [CAL-2821] Request details: redirect_uri=https://cal.com/api/auth/callback/google

### Issue Summary

During sign up, I encountered an error.

Request details: redirect_uri=[https://cal.com/api/auth/callback/google](https://cal.com/api/auth/callback/google)

image

After trying again, I was able to sign in successfully.

"When I attempted to connect to Microsoft 365, I encountered an error message and have been unable to proceed since."

{"message":"No code returned"}

image

[https://app.cal.com/api/integrations/office365video/callback?error=invalid_scope&error_description=The provided value for the input parameter 'scope' is not valid. The scope 'OnlineMeetings.ReadWrite offline_access' does not exist.&state={"returnTo":"https://app.cal.com/apps/installed"}](https://app.cal.com/api/integrations/office365video/callback?error=invalid_scope&error_description=The%20provided%20value%20for%20the%20input%20parameter%20%27scope%27%20is%20not%20valid.%20The%20scope%20%27OnlineMeetings.ReadWrite%20offline_access%27%20does%20not%20exist.&state=%7b%22returnTo%22:%22https://app.cal.com/apps/installed%22%7d)

### Steps to Reproduce

1. [https://app.cal.com/getting-started/connected-video](https://app.cal.com/getting-started/connected-video)

### Actual Results

* What's happening right now that is different from what is expected

### Expected Results

* This is an ideal result that the system should get after the tests are performed

### Technical details

* Browser version, screen recording, console logs, network requests: You can make a recording with [Bird Eats Bug](https://birdeatsbug.com/).
* Node.js version
* Anything else that you think could be an issue.

### Evidence

* How was this tested? This is quite mandatory in terms of bugs. Providing evidence of your testing with screenshots or/and videos is an amazing way to prove the bug and a troubleshooting chance to find the solution.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2821](https://linear.app/calcom/issue/CAL-2821/request-details-redirect-uri=httpscalcomapiauthcallbackgoogle)</sub>

### LINKED ISSUE #12602: [CAL-2781] Cancelling Access to google account gives Json output

### Issue Summary
For the first time i Login with google, website brings to the step 1 of 5 , while completing this on step 2 connecting to google calendar if i  cancel the access " Cal.com wants access to your Google Account "  , It brings to an error " https://app.cal.com/api/integrations/googlecalendar/callback?error=access_denied&state=%7B%22returnTo%22:%22https://app.cal.com/apps/installed%22%7D " Showing {"message":"`code` must be a string"}

### Steps to Reproduce

1. Login
2. Connect to Calendar Pop ups Step 1/5
3. Choose google calendar and connect
4. Choose account
5. If cancel the access error shows up

https://github.com/calcom/cal.com/assets/98404323/22c35115-be86-4625-8b96-a9cdd0a15b61

### Actual Results

- What's happening right now that is different from what is expected

### Expected Results

- This is an ideal result that the system should get after the tests are performed

### Technical details

- Node.js version

### Evidence

- How was this tested? This is quite mandatory in terms of bugs. Providing evidence of your testing with screenshots or/and videos is an amazing way to prove the bug and a troubleshooting chance to find the solution.

<sub>[CAL-2781](https://linear.app/calcom/issue/CAL-2781/cancelling-access-to-google-account-gives-json-output)</sub>

joeauyeung:
## What does this PR do?

This PR also takes some code from #12513 

This PR aims to redirect users back to the page before they started the OAuth flow. Before it would just display a HTTP error.

Fixes #12602 

https://www.loom.com/share/63289e3fae6343e8bfabe86a5aaf963e

## Requirement/Documentation

## Type of change

- From the app store, app page, and onboarding
        - For GCal, Outlook, and Stripe
            - Install the app
            - Cancel part way through the flow
            - You should be redirected back to the page you came from 
- For GCal
    - Only grant one permission. The credential should not be created and you should return to the page you were at before

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -28,10 +29,21 @@ type UseAddAppMutationOptions = CustomUseMutationOptions & {
 
 function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMutationOptions) {
   const { returnTo, ...options } = allOptions || {};
+  const router = useRouter();
+  const onErrorReturnTo = `${WEBAPP_URL}${router.asPath}`;
```
joeauyeung, line 33: Get the URL that the app is being installed from inside of the hook. That way we don't have to add the URL param to every place the hook is used.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -57,6 +69,8 @@ function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMuta
             { variant: variables && variables.variant, slug: variables && variables.slug },
             location.search
           ),
+      onErrorReturnTo,
+      fromApp: true,
```
joeauyeung, line 73: Since this hook is called within the app then return there.

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -19,6 +20,13 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
   const state = decodeOAuthState(req);
 
   if (typeof code !== "string") {
+    if (state?.onErrorReturnTo) {
+      res.redirect(
+        getSafeRedirectUrl(state.onErrorReturnTo) ??
+          getSafeRedirectUrl(state?.returnTo) ??
+          `${CAL_URL}/apps/installed`
```
joeauyeung, line 27: Fallback to the original returnTo URL if the error one is not present.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -28,10 +29,21 @@ type UseAddAppMutationOptions = CustomUseMutationOptions & {
...
-    { type?: App["type"]; variant?: string; slug?: string; isOmniInstall?: boolean; teamId?: number } | ""
+    | {
+        type?: App["type"];
+        variant?: string;
+        slug?: string;
+        isOmniInstall?: boolean;
+        teamId?: number;
+        onErrorReturnTo?: string;
+      }
+    | ""
```
sean-brydon, line 46: Lets extract this to a type 
```suggestion
type UseAddAppMutationVariables =
  | {
      type?: App["type"];
      variant?: string;
      slug?: string;
      isOmniInstall?: boolean;
      teamId?: number;
      onErrorReturnTo?: string;
    }
  | undefined;
```

packages/app-store/stripepayment/api/callback.ts
```diff
@@ -19,8 +20,13 @@ function getReturnToValueFromQueryState(req: NextApiRequest) {
 
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
   const { code, error, error_description } = req.query;
+  const state = decodeOAuthState(req);
 
   if (error) {
+    // User cancels flow
+    if (error === "access_denied") {
+      state?.onErrorReturnTo ? res.redirect(state.onErrorReturnTo) : res.redirect("/apps/installed/payment");
```
sean-brydon, line 28: Do we need to getSafeRedirectUrl here like we have the others? Not too familiar with this code so let me know if im wrong (and why so i learn a thing or two :P )

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -19,6 +20,13 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
...
+      );
+    }
     throw new HttpError({ statusCode: 400, message: "`code` must be a string" });
```
sean-brydon, line 30: Even when we redirect the user back onError - we are still throwing this HttpError - I think this needs to be in the above `else` statement.

sean-brydon (commented):
We are still throwing the Http error even when we are redirecting back to the onErrorReturnTo - is this desired?  

If so this is needed then this works perfectly fine and the user is redirected back in all cases i tested.

packages/app-store/stripepayment/api/callback.ts
joeauyeung: I'd like to have this in place. There's no harm in the way we're handling things but this could be a good safe guard if a self-hoster wants to make changes.

packages/app-store/googlecalendar/api/callback.ts
joeauyeung: I was getting weird type errors if I tried putting it in an `else` block. How do you feel about adding return statements? cfce15582317a5e32f4dcb891d380122b835ccd6

sean-brydon: Fine by me :)

sean-brydon (dismissed):
Tested again locally - http errors are no longer thrown... Great job @joeauyeung

packages/app-store/googlecalendar/api/callback.ts
Udit-takkar: @joeauyeung I think we need to use WEBAPP_URL here?

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -36,12 +45,32 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
...
+        if (!state?.fromApp) {
+          throw new HttpError({
+            statusCode: 400,
+            message: "You must grant all permissions to use this integration",
+          });
+        } else {
+          res.redirect(
+            getSafeRedirectUrl(state.onErrorReturnTo) ??
+              getSafeRedirectUrl(state?.returnTo) ??
+              `${CAL_URL}/apps/installed`
```
Udit-takkar, line 67: https://github.com/calcom/cal.com/pull/13189/files#r1458786148 WEBAPP_URL here

Udit-takkar (commented):
@joeauyeung great work but i found some issues

1) If you try to install the google calendar with only one permission again. It successfully adds the calendar but it shouldn't

2) We should add a toast with some message like "All Permissions not granted". Right now nothing is displayed

https://github.com/calcom/cal.com/assets/53316345/8d1ff3b9-5503-48ba-9b11-d8744bb77d74

CarinaWolli:
production build is failing, also fails on local. @joeauyeung can you take a look at that?

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -1,5 +1,6 @@
 import type { UseMutationOptions } from "@tanstack/react-query";
 import { useMutation } from "@tanstack/react-query";
+import { useRouter } from "next/router";
```
zomars, line 3: next/router is deprecated. We should be using next/navigation instead. This is causing all future routes to break.

zomars (dismissed):

packages/app-store/_utils/useAddAppMutation.ts
zomars: @joeauyeung more info on this https://nextjs.org/docs/pages/building-your-application/upgrading/app-router-migration#step-5-migrating-routing-hooks

Udit-takkar (dismissed):
> If you try to install the google calendar with only one permission instead of two and repeat this step then it successfully adds the calendar at the second time.

@joeauyeung This is the only blocking issue in the PR.

keithwillcode (approved):

PR #13191, opened 2024-01-12
feat: upgrade react query 5

emrysal:
## What does this PR do?

* Allow toast errors through custom meta in the useQuery hook.
* Remove the onSuccess callback from the useQuery listSchedules.
* BUGFIX: When selecting the default option, it will now unset the schedule; so it will change upon changing the default. (Before this fix the schedule would not be changed after toggling the select options, as the schedule id would be set in the DB)


keithwillcode:
@emrysal should we close this in favor of #13335?

PR #13193, opened 2024-01-12
fix: correctly use eventTriggers parsed and use Array.from

sean-brydon:
Fixes: `Type 'Set<WebhookTriggerEvents>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.` 

Also uses the parsed version of eventTriggers instead of inferring from body. 

I believe this should fix this issue - @CarinaWolli  @alishaz-polymath  might be best to confirm


alishaz-polymath (approved):
LGTM

PR #13211, opened 2024-01-13
refactor: API POST /teams

### LINKED ISSUE #9127: [CAL-1794] Allow admin API Key CRUD operation on all resources
I've found several  API routes that will only allow operations on resources of the creator of the used API key, despite him having the `ADMIN` role. We expect an admins API key to do all CRUD operation on all API endpoints without restriction.

This is quite a problem for our current project and it seems to affect multiples routes and methods.
Here are some routes I've found that are affected by this issue:

- PATCH /v1/users/:id
- GET /v1/memberships
- GET /v1/schedules
- GET /v1/availabilities

This list is **by no means comprehensive**, I've only tried some APIs and this issue seems to be spread across them.

<sub>[CAL-1794](https://linear.app/calcom/issue/CAL-1794/allow-admin-api-key-crud-operation-on-all-resources)</sub>

supalarry:
Fixes https://github.com/calcom/cal.com/issues/9127

Loom: https://www.loom.com/share/c9e6c761be9b48d3847b176d458d81b1

This PR has 2 refactors:
1. Allow admins to create teams on the behalf of other users.
2. Create team only after receiving payment.

### Allow admins to create teams on the behalf of other users
File [apps/api/lib/validations/team.ts](https://github.com/calcom/cal.com/compare/lauris/cal-1794-allow-admin-api-key-crud-operation-on-all-resources?expand=1#diff-4b94b90bbe89b92ea0ca78e60afb9e2737b8f4b1df94310aa1837a4f3840a5b4) has been updated to allow passing "ownerId" in the request body to specify on behalf of who should the team be created. Notably, only admins can do this as seen by the "checkPermissions" in [api/teams/_post.ts](https://github.com/calcom/cal.com/compare/lauris/cal-1794-allow-admin-api-key-crud-operation-on-all-resources?expand=1#diff-0819d94fac5fcdb3b81e49514fb37c2cf09f7514e5c1bd84ed6b3572db87ddeb) - if user is admin and "ownerId" is in the request body, then we specify "const effectiveUserId = isAdmin && ownerId ? ownerId : userId;" which is used to determine owner of the team.

### Create team only after receiving payment

New Prisma model - "AwaitingPaymentTeam" has been added to [schema.prisma](https://github.com/calcom/cal.com/compare/lauris/cal-1794-allow-admin-api-key-crud-operation-on-all-resources?expand=1#diff-56fa4d384ba6fb94b130e9eb42c5cad00734b025ebd6c197f76873469725ae87) in order to store team information submitted to [api/teams/_post.ts](https://github.com/calcom/cal.com/compare/lauris/cal-1794-allow-admin-api-key-crud-operation-on-all-resources?expand=1#diff-0819d94fac5fcdb3b81e49514fb37c2cf09f7514e5c1bd84ed6b3572db87ddeb). We need it because we can't store keys with boolean values in Stripe checkout metadata while we have "hideBookATeamMember": true / fase and "isPrivate": true / false as required properties when creating a team.

1. After receiving request [api/teams/_post.ts](https://github.com/calcom/cal.com/compare/lauris/cal-1794-allow-admin-api-key-crud-operation-on-all-resources?expand=1#diff-0819d94fac5fcdb3b81e49514fb37c2cf09f7514e5c1bd84ed6b3572db87ddeb) will store team's data in the "AwaitingPaymentTeam" table, and create a new stripe checkout session where metadata contains id of the newly created "AwaitingPaymentTeam" and "ownerId". User will receive stripe link to pay:
Screenshot 2024-01-13 at 16 21 03

2. After user pays, he / she will be re-directed to the success_url that was specified by "generateTeamCheckoutSession" when creating session in step 1, which is the [api/teams/api/create.ts](https://github.com/calcom/cal.com/compare/lauris/cal-1794-allow-admin-api-key-crud-operation-on-all-resources?expand=1#diff-46f4a56b7d49728273270b82fbfe3a43f197ceb66866988711d52d232839fd72) page. It will check stripe metadata "awaitingPaymentTeamId", fetch it and using the data create an actual team that is returned to user and the awaiting payment team is deleted.


apps/api/lib/validations/team.ts
```diff
@@ -7,12 +7,20 @@ export const schemaTeamBaseBodyParams = Team.omit({ id: true, createdAt: true })
...
+const schemaTeamRequiredParams = z.object({
+  name: z.string().max(255),
+});
 
 export const schemaTeamBodyParams = schemaTeamBaseBodyParams.merge(schemaTeamRequiredParams).strict();
 
 export const schemaTeamUpdateBodyParams = schemaTeamBodyParams.partial();
 
+const schemaOwnerId = z.object({
+  ownerId: z.number().optional(),
```
supalarry, line 19: Self review: Allow specifying ownerId in the POST "/teams" body.

apps/api/pages/api/teams/_post.ts
```diff
@@ -87,24 +124,78 @@ async function postHandler(req: NextApiRequest) {
...
+    paymentLink: checkoutSession.url,
   };
 }
 
+async function checkPermissions(req: NextApiRequest) {
+  const { isAdmin } = req;
+  const body = schemaTeamCreateBodyParams.parse(req.body);
+
+  /* Non-admin users can only create teams for themselves */
+  if (!isAdmin && body.ownerId)
```
supalarry, line 151: Self review: only admins can specify "ownerId" in the request body.

apps/api/pages/api/teams/_post.ts
```diff
@@ -87,24 +124,78 @@ async function postHandler(req: NextApiRequest) {
     ...data,
     metadata: data.metadata === null ? {} : data.metadata || undefined,
   };
-  const team = await prisma.team.create({
+  const awaitingPaymentTeam = await prisma.awaitingPaymentTeam.create({
```
supalarry, line 127: Self review: create "AwaitingPaymentTeam" to persist team data submitted to API. "AwaitingPaymentTeam" id will be passed to metadata of stripe checkout session, and after successful payment the id will be used to fetch the corresponding "AwaitingPaymentTeam", create a new "Team" using it, delete "AwaitingPaymentTeam" entry and return the actual "Team" to user.

apps/api/pages/api/teams/_post.ts
```diff
@@ -87,24 +124,78 @@ async function postHandler(req: NextApiRequest) {
...
+}: {
+  awaitingPaymentTeamId: number;
+  ownerId: number;
+}) => {
+  const customer = await getStripeCustomerIdFromUserId(ownerId);
+  const session = await stripe.checkout.sessions.create({
+    customer,
+    mode: "subscription",
+    allow_promotion_codes: true,
+    success_url: `${WEBAPP_URL}/api/teams/api/create?session_id={CHECKOUT_SESSION_ID}`,
```
supalarry, line 170: Self review: This function generates a Stripe checkout session that on success will lead to the new API endpoint "/api/teams/api/create". I decided not to re-use handler or generateTeamCheckoutSession created by Joe in https://github.com/calcom/cal.com/pull/12501 because mixing two resulted in more complex logic that was hard to understand.

apps/web/pages/api/teams/api/create.ts
```diff
@@ -0,0 +1,132 @@
+import type { NextApiRequest, NextApiResponse } from "next";
```
supalarry, line 1: Self review: This endpoint is the success_url of Stripe checkout session, so after payment user is re-directed here, where "AwaitingPaymentTeam" is fetched using awaiting payment team id in the metadata and actual Team is created and returned to user.

supalarry (commented):
Self review finished

apps/api/pages/api/teams/_post.ts
supalarry: We can't use stripe checkout metadata to store team data, because checkout metadata does not accept keys with values that are booleans, but we allow submitting properties to "/teams", like "isPrivate" that are boolean.

packages/prisma/schema.prisma
```diff
@@ -320,6 +320,35 @@ model Team {
...
+  theme               String?
+  brandColor          String?
+  darkBrandColor      String?
+  parentId            Int?
+  timeFormat          Int?
+  timeZone            String   @default("Europe/London")
+  weekStart           String   @default("Sunday")
+
+  @@unique([slug, parentId])
+}
```
zomars, line 350: Can we achieve this without having to repeat this code? If we make changes to Team we need to double check here as well. I would rather have some kind of "pendingPayment" Boolean value and filter out on team queries. We could even use a prisma middleware to omit teams with "pendingPayment" by default.

zomars: We can follow the soft-delete approach for the Prisma Docs https://www.prisma.io/docs/orm/prisma-client/client-extensions/middleware/soft-delete-middleware#option-2-use-middleware-to-determine-the-behavior-of-readupdate-queries-for-deleted-records

supalarry: Great idea! Added "pendingPayment" to schema and refactored existing code to work with it in commit [refactor: replace AwaitingPaymentTeam table with Team.pendingPayment](https://github.com/calcom/cal.com/pull/13211/commits/a2f25c1c3073fdf8375e5a2508616bc8f7196689).

Then, added prisma filtering extension in commit [feat: filter out teams with pending payments](https://github.com/calcom/cal.com/pull/13211/commits/9f19a5bff43da659082936f456fddc250a562a1c).

In the handler that is invoked after successful stripe checkout "apps/web/pages/api/teams/api/create.ts" I tested "prisma.team.findFirst({where: {id: pendingPaymentTeamId}})" aka team we created in POST "/teams" but with "pendingPayment: true" and null was returned, so the filter works. Notably, the filter only works for find operations and not update, because we invoke update after successful checkout to set "pendingPayment: false".

apps/api/pages/api/teams/_post.ts
```diff
@@ -49,8 +76,20 @@ import { schemaTeamBodyParams, schemaTeamReadPublic } from "~/lib/validations/te
...
 async function postHandler(req: NextApiRequest) {
-  const { prisma, body, userId } = req;
-  const data = schemaTeamBodyParams.parse(body);
+  if (!IS_TEAM_BILLING_ENABLED) {
+    const PRECONDITION_FAILED_STATUS_CODE = 412;
+    throw new HttpError({
+      statusCode: PRECONDITION_FAILED_STATUS_CODE,
+      message: "Team creation is unavailable because team billing is disabled.",
+    });
+  }
```
zomars, line 85: Will this prevent team creation for self-hosters that don't use team billing? If so, we should just create the team normally if billing is not enabled.

apps/api/pages/api/teams/_post.ts
```diff
@@ -59,13 +98,11 @@ async function postHandler(req: NextApiRequest) {
...
-      data.metadata = {
-        requestedSlug: data.slug,
-      };
-      delete data.slug;
-    }
+    // Setting slug in metadata, so it can be published later
+    data.metadata = {
+      requestedSlug: data.slug,
+    };
+    delete data.slug;
```
zomars, line 105: This is deprecated now since we require payment before creation now.

```suggestion
```

apps/web/pages/api/teams/api/create.ts
```diff
@@ -0,0 +1,118 @@
...
+      },
+      metadata: {
+        ...(typeof activatedTeam.metadata === "object" ? activatedTeam.metadata : {}),
+        paymentId: checkoutSession.id,
+        subscriptionId: checkoutSessionSubscription.id || null,
+        subscriptionItemId: checkoutSessionSubscription.items.data[0].id || null,
+      },
+    },
+    include: { members: true },
+  });
```
zomars, line 57: I don't think a new team with payment pending will have any significant metadata. Also the requetedSlug is deprecated anyways 

```suggestion
  const finalizedTeam = await prisma.team.update({
    where: { id: checkoutSessionMetadata.pendingPaymentTeamId },
    data: {
      pendingPayment: false,
      members: {
        create: {
          userId: checkoutSessionMetadata.ownerId as number,
          role: MembershipRole.OWNER,
          accepted: true,
        },
      },
      metadata: {
        paymentId: checkoutSession.id,
        subscriptionId: checkoutSessionSubscription.id || null,
        subscriptionItemId: checkoutSessionSubscription.items.data[0].id || null,
      },
    },
    include: { members: true },
  });
```

zomars (commented):
Awesome work @supalarry ! We're almost there. Left some QA comments to address before merging 🙏🏽

apps/web/pages/api/teams/api/create.ts
supalarry: Agreed. Done in [Ignore pending team metadata](https://github.com/calcom/cal.com/pull/13211/commits/d7a66ccd4832b84af56b4abf6e7e686b55e89357)

apps/api/pages/api/teams/_post.ts
supalarry: Good point. Done in [Create teams as before if billing disabled](https://github.com/calcom/cal.com/pull/13211/commits/5f0381ce83e5058edb4b45f284b99b005b3cc0ce)

zomars (approved):
Ship it

PR #13218, opened 2024-01-14
fix: 13217-Now redirects to apps page

Quanta-Naut:
Now, when the google calendar auth is canceled, the page gets redirected to apps page of cal.com


sean-brydon:
Hi @Quanta-Naut thanks for this PR. 

I believe this is a duplicate of @joeauyeung's PR here - https://github.com/calcom/cal.com/pull/13189 That seems to tackle more apps and gives some more explicit errors and a more app wide approach to solving this issue other than just for google OAuth.

PR #13248, opened 2024-01-16
fix: disable input elements while updating the form

### LINKED ISSUE #13111: [CAL-2939] input elements should be disabled while loading (updating or creating)

Most of the inputs are still accepting input even while loading. But it's leading to bugs.

Inputs should be disabled while loading.

---
##### House rules

<sub>[CAL-2939](https://linear.app/calcom/issue/CAL-2939/input-elements-should-be-disabled-while-loading-updating-or-creating)</sub>

Krishcode264:
## What does this PR do?
Fixes #13111

  i have added prop disabled={isLoading} for name and email and username 
  used same logic to disable add avatar buttons while updating the form 

### what could have been better 
if i could do same with "About" input but 
```
  <ContentEditable
                  readOnly={!editable}
                  style={{ height: props.height }}
                  className="editor-input"
                />
```
even if i put readOnly={false } or {true } for dev purpose it does chnage anything its always editable 

other than that all input fields are tested which are getting disabled while updating form 

this is my first PR ever and i am getting some issues with yarn.lock which is not matching with orignal one , so please check it .


yarn.lock
```diff
@@ -5,46 +5,6 @@ __metadata:
   version: 6
   cacheKey: 8
 
-"@0no-co/graphql.web@npm:^1.0.1":
```
CarinaWolli, line 8: Never commit yarn.lock if there are no changes in package.json. 

You can use `git checkout origin/main -- yarn.lock` to remove the changes :)

CarinaWolli (changes requested):
I am not a fan of having this flicker of gray input fields. @ciaranha what do you think about that: https://www.loom.com/share/549c87a1dbed4a7b87a3e1d2bfd827c3

Also as mentioned in the issue this happens in other places as well (event type update), unsure if this is even a real issue we want to fix (cc @ciaranha )

yarn.lock
Krishcode264: i just resolved that yarn.lock issue , updated yarn.lock to original repo . thank you for your help :)

ciaranha:
> I am not a fan of having this flicker of gray input fields. @ciaranha what do you think about that: https://www.loom.com/share/549c87a1dbed4a7b87a3e1d2bfd827c3

Also not a fan of the flicker. It's drawing my attention to things I otherwise wouldn't look at.

PR #13253, opened 2024-01-16
chore: Revert way bundle analysis produces output

keithwillcode:
## What does this PR do?

Fixes the bundle analysis failures in CI.


keithwillcode:
Bundle analysis now working again. Getting rid of the set-output calls like I had tried to do in not super high priority so closing this for the moment.

PR #13270, opened 2024-01-17
Remove constraint slug,isOrganization and reset migration

hariombalhara:
This is @sean-brydon's PR with migration fixed by removing the unique constraint [slug, isOrganization] which was unnecessary

unique constraint [slug, parentId] is what we need
Requirements:
1. calcom can be an Organization slug -> slug='calcom', parentId=null
2. calcom can be a sub-team slug in org1 -> slug='calcom', parentId="org1"
3. calcom can be a sub-team slug in org2 -> slug='calcom', parentId="org2"
4. calcom can't be a non-sub-team slug -> slug='calcom'; parentId=null - Conflicts with 1
5. calcom can't be another organization's slug -> slug='calcom'; parentId=null - Conflicts with 1

### DIFF
--- a/packages/prisma/schema.prisma
+++ b/packages/prisma/schema.prisma
@@ -321,7 +321,6 @@ model Team {
   pendingPayment       Boolean                 @default(false)
 
   @@unique([slug, parentId])
-  @@unique([slug, isOrganization])
 }
 
 model OrganizationSettings {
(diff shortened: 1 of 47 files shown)


packages/prisma/schema.prisma
```diff
@@ -315,12 +315,23 @@ model Team {
   apiKeys              ApiKey[]
   credentials          Credential[]
   accessCodes          AccessCode[]
+  isOrganization       Boolean                 @default(false)
+  organizationSettings OrganizationSettings?
   instantMeetingTokens InstantMeetingToken[]
   pendingPayment       Boolean                 @default(false)
 
   @@unique([slug, parentId])
```
hariombalhara, line 323: Notice there is no unique constraint [slug, isOrganization] which was there in original PR

PR #13271, opened 2024-01-17
fix: Revert -> "Revert "chore: Chore team metadata table + isOrganization migration

hariombalhara:
This is reverting the revert for @sean-brydon's PR and also fixing the migration by removing the unique constraint [slug, isOrganization] which was unnecessary

Unique constraint _**[slug, parentId]**_ is what we need which is already there.

Requirements:
1. calcom can be an Organization slug -> slug='calcom', parentId=null
2. calcom can be a sub-team slug in org1 -> slug='calcom', parentId="org1"
3. calcom can be a sub-team slug in org2 -> slug='calcom', parentId="org2"
4. calcom **can't be** a non-sub-team slug -> slug='calcom'; parentId=null - Conflicts with 1
5. calcom **can't be** another organization's slug -> slug='calcom'; parentId=null - Conflicts with 1


packages/prisma/schema.prisma
```diff
@@ -315,12 +315,23 @@ model Team {
   apiKeys              ApiKey[]
   credentials          Credential[]
   accessCodes          AccessCode[]
+  isOrganization       Boolean                 @default(false)
+  organizationSettings OrganizationSettings?
   instantMeetingTokens InstantMeetingToken[]
   pendingPayment       Boolean                 @default(false)
 
   @@unique([slug, parentId])
```
hariombalhara, line 323: Notice there is no unique constraint [slug, isOrganization] which was there in original PR [here ](https://github.com/calcom/cal.com/pull/12828/files#diff-56fa4d384ba6fb94b130e9eb42c5cad00734b025ebd6c197f76873469725ae87R323)

packages/prisma/migrations/20240117062434_move_org_metadata_to_org_settings/migration.sql
```diff
@@ -0,0 +1,46 @@
...
+    "isOrganizationConfigured" BOOLEAN NOT NULL DEFAULT false,
+    "isOrganizationVerified" BOOLEAN NOT NULL DEFAULT false,
+    "orgAutoAcceptEmail" TEXT NOT NULL,
+
+    CONSTRAINT "OrganizationSettings_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "OrganizationSettings_organizationId_key" ON "OrganizationSettings"("organizationId");
+
```
hariombalhara, line 17: Notice the constraint is not there now https://github.com/calcom/cal.com/pull/12828/files#r1453879560

hariombalhara:
I have deleted the previous faulty migration and added a new one. I believe this is the best solution considering that the faulty migration must have been already rolled back. cc @zomars

zomars:
So are we okay with nested subteams if the same org having the same slug?

Say id=1,slug=calcom,parentId=null can have a subteam id=2,slug=calcom,parentId=1 which in turn can have a subteam id=3,slug=calcom,parentId=2 and so on and so on... so an org can have an infinite amount of calcom subteams as long are they're not siblings. Are we okay with that? What would be the permalink of each one?

zomars (changes requested):
So are we okay with nested subteams if the same org having the same slug?

Say id=1,slug=calcom,parentId=null can have a subteam id=2,slug=calcom,parentId=1 which in turn can have a subteam id=3,slug=calcom,parentId=2 and so on and so on... so an org can have an infinite amount of calcom subteams as long are they're not siblings. Are we okay with that? What would be the permalink of each one?

hariombalhara:
> So are we okay with nested subteams if the same org having the same slug?
> 
> Say id=1,slug=calcom,parentId=null can have a subteam id=2,slug=calcom,parentId=1 which in turn can have a subteam id=3,slug=calcom,parentId=2 and so on and so on... so an org can have an infinite amount of calcom subteams as long are they're not siblings. Are we okay with that? What would be the permalink of each one?

We don't support nested sub teams and I don't think we are going to support them in near future. There is going to be an organization and it's children teams. That's it.

Yeah normally someone won't create a subteam in an organization with same slug as org but practically it happens when a team requestes to be upgraded to become an organization.

zomars (dismissed):

PeerRich (approved):
reapproving on behalf of @zomars (his review went stale after fixing things)

keithwillcode:
@hariombalhara What's the status of this one? I feel like we need to get it moving along again or close it.

hariombalhara:
I will pick it this milestone

hariombalhara:
Closing in favour of #13700

PR #13282, opened 2024-01-17
feat: Limit the amount of days to load for a schedule

keithwillcode:
## What does this PR do?

One step towards fixing https://github.com/calcom/cal.com/issues/13358. Allows a self-hoster to override how much data is loaded on the booker. In the case where users don't care to load an entire month's worth of data, they can change this value to a smaller number of days to increase perf.

## Type of change

## How should this be tested?

- Set NEXT_PUBLIC_BOOKER_NUMBER_OF_DAYS_TO_LOAD to the number of days of a schedule you want to preload.

## Mandatory Tasks


packages/features/bookings/Booker/Booker.tsx
```diff
@@ -87,13 +87,15 @@ const BookerComponent = ({
   const date = dayjs(selectedDate).format("YYYY-MM-DD");
 
   const prefetchNextMonth =
-    (layout === BookerLayouts.WEEK_VIEW &&
+    !process.env.NEXT_PUBLIC_BOOKER_DISALLOW_PREFETCH_NEXT_MONTH &&
```
keithwillcode, line 90: Would rather call this ALLOW_PREFETCH_NEXT_MONTH so we don't have a double negative here but we need this to default to work as is. Without the double negative, it would be necessary for self-hosters upgrading to set this to `true` before upgrading, which is not ideal. 

As it stands now, the default will be false and will continue working as usual.

packages/features/bookings/Booker/components/DatePicker.tsx
```diff
@@ -35,7 +36,7 @@ export const DatePicker = ({
         setMonth(date.format("YYYY-MM"));
         setSelectedDate(date.format("YYYY-MM-DD"));
       }}
-      includedDates={nonEmptyScheduleDays}
+      includedDates={BOOKER_NUMBER_OF_DAYS_TO_LOAD > 0 ? null : nonEmptyScheduleDays}
```
keithwillcode, line 39: This makes it such the entire calendar is clickable and data is loaded in once clicked.

packages/features/bookings/Booker/components/HavingTroubleFindingTime.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function HavingTroubleFindingTime(props: Props) {
+  const { t } = useLocale();
+  const [internalClick, setInternalClick] = useState(false);
+
+  // Eaiest way to detect if its not enabled
+  if (
+    (process.env.NEXT_PUBLIC_BOOKER_NUMBER_OF_DAYS_TO_LOAD == "0" && BOOKER_NUMBER_OF_DAYS_TO_LOAD == 0) ||
+    !process.env.NEXT_PUBLIC_BOOKER_NUMBER_OF_DAYS_TO_LOAD
+  )
+    return null;
```
sean-brydon, line 21: If default of "0" is set OR if there is no next public var hide this component

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -86,6 +91,17 @@ const BookerComponent = ({
...
+
+  const [dayCount, setDayCount] = useState(
+    layout === BookerLayouts.WEEK_VIEW
+      ? 7
+      : BOOKER_NUMBER_OF_DAYS_TO_LOAD > 0
+      ? BOOKER_NUMBER_OF_DAYS_TO_LOAD
+      : undefined
+  );
+
+  console.log(dayCount);
```
sean-brydon, line 103:
```suggestion
```

exception (dismissed):
Looks ok to me.

packages/features/schedules/lib/use-schedule/useSchedule.ts
```diff
@@ -21,17 +23,39 @@ export const useSchedule = ({
...
+  const now = dayjs();
+  const monthDayjs = month ? dayjs(month) : now;
   const nextMonthDayjs = monthDayjs.add(monthCount ? monthCount : 1, "month");
   // Why the non-null assertions? All of these arguments are checked in the enabled condition,
   // and the query will not run if they are null. However, the check in `enabled` does
   // no satisfy typescript.
+  let startTime;
+  let endTime;
+
+  if (!!dayCount && dayCount > 0) {
```
keithwillcode, line 43: Will have a follow-up PR to add unit tests to this.

exception (approved):

exception (approved):

PR #13292, opened 2024-01-18
refactor: Only send dirty fields for event type update call

### LINKED ISSUE #13183: [CAL-2920] Managed Event: Individual settings are overwritten when one parameter is changed
### Issue Summary
Not sure if this a bug, but it poses problems: When the owner of a managed event changes a setting in the managed event (e.g. switch on "Hide notes in calendar" in the advanced settings) this also overwrites all other individual settings of the team members (e.g. their individual limit of future bookings). This means every member that is part of the managed event has to change back the settings every time the admin makes a change.

### Steps to Reproduce

1. Create a managed event and add a member to it.
2. Set up some individual settings (e.g. Limit future bookings to 30 weekdays) in the limits section of the member.
3. Go to the managed event and to the "Advanced" section.
4. Activate the "Hide notes in calendar" switch and save it.

### Actual Results
Now all settings in "Advanced" and "Limits" are overwritten equally.

### Expected Results
There should be an option to just change one parameter without overwriting all other settings. Or at least not in another section. Or only for settings that are locked for team members and cannot be changed by them.

<sub>[CAL-2920](https://linear.app/calcom/issue/CAL-2920/managed-event-individual-settings-are-overwritten-when-one-parameter)</sub>

### LINKED ISSUE #13291: [CAL-2956] Only pass dirty fields for update in event-types
We currently pass all fields for updating the event types even if there's only one dirty field. This isn't necessarily the worst, but ideally we should only send the updated fields. This is, however, a necessity for Managed Events V2.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2956](https://linear.app/calcom/issue/CAL-2956/only-pass-dirty-fields-for-update-in-event-types)</sub>

alishaz-polymath:
## What does this PR do?

Earlier we would pass all event type fields even if the change is in one field. This PR updates the logic to only pass in dirty fields instead of all fields. This is a pre-requisite to Managed Events V2, so this PR would need to be merged before we can continue sorting out the Managed Events V2.

Here's a list of all fields that currently identify as dirty accurately:
- [x] Setup Tab
  - [x] Title
  - [x] slug
  - [x] Description
  - [x] locations
  - [x] length
- [x] hidden
- [x] hosts
- [x] users
- [x] hashedLink
- [x] bookings
- [x] availability tab
  - [x] Availability
  - [x] Common Schedule
- [x] Teams Tab
  - [x] Hosts
  - [x] Priority  
- [x] Limits Tab
  - [x] beforeEventBuffer
  - [x] afterEventBuffer
  - [x] Minimum Notice
  - [x] Time-slot Intervals
  - [x] Booking Frequency
  - [x] Booking only first slot
  - [x] Total Booking Duration
  - [x] Future Bookings
  - [x] Offset start times
- [x] Advanced Tab
  - [x] destinationCalendar
  - [x] eventName
  - [x] Layout (in Metadata)
  - [x] bookingFields
    - [x] Name
    - [x] Email Address
    - [x] Location
    - [x] What is this meeting about
    - [x] Additional Notes
    - [x] Add guests
    - [x] Reason for reschedule
  - [x] hideCalendarNotes
  - [x] hashed Link
  - [x] requiresConfirmation
  - [x] requiresBookerEmailVerification
  - [x] successRedirectURL
  - [x] lockTimeZoneToggleOnBookingPage
  - [x] Seated Events
- [x] recurringEvents
- [x] workflows
- [x] webhooks

Fixes #13291 
Fixes #13183 

## Requirement/Documentation

## Type of change

## How should this be tested?

- Rigorously and carefully, touches core (event-types) and so it should be tested against all kinds of event-types, with the idea of saving changes to an event type. In managed events, it should only over-write the fields changed by Admin and not all the fields.

## Mandatory Tasks


apps/web/pages/event-types/[type]/index.tsx
```diff
@@ -563,6 +564,20 @@ const EventTypePage = (props: EventTypeSetupProps) => {
...
+              // Iterate over watched fields and compare with initial values
+              Object.keys(watchedFields).forEach((key) => {
+                const typedKey = key as keyof FormValues;
+                if (!isEqual(watchedFields[typedKey], defaultValues[typedKey])) {
+                  Object.assign(dirtyFields, { [typedKey]: watchedFields[typedKey] });
+                }
+              });
+              return dirtyFields;
+            };
+            const dirtyFields = getDirtyFields();
```
Comment (sean-brydon) [inline apps/web/pages/event-types/[type]/index.tsx:580]: Instead of doing this deep comparision we can actually leverage react hook form a bit more to get this 

-> Import `useFormState` from "react-hook-form"
-> We can grab the dirty fields as follows

  const { dirtyFields: dirtyFieldsFromHook } = useFormState({ control: formMethods.control });
  
  This will return a  `{ [name]:boolean }` object with all fields that have been touched. We should iterate over that and grab the values. 
  
 This is much cleaner. 
 
 I also noticed the method we have created here is not 100% correct and returns fields that i've not touched when just changing title.

packages/lib/isEqual.ts
sean-brydon, line 1: I think we should try and avoid this deep comparison as much as possible. I have left a suggestion in the above comment on a way to do it in rhf itself

sean-brydon (dismissed):
I think we can really optimise this and leverage RHF to do this for us - I also noticed this method was including fields that were never touched when saving. Please see other comments to see my explanation

apps/web/pages/event-types/[type]/index.tsx
Comment (alishaz-polymath) [reply apps/web/pages/event-types/[type]/index.tsx:580]: Yeah I agree. The dirtyFields wasn't working as expected so I tried a workaround. Needless to say, it's way better to amke dirtyFields work instead, and that's what @emrysal recommends as well. That's the move ahead as well. 🙏

packages/lib/isEqual.ts
alishaz-polymath: I agree

apps/web/pages/event-types/[type]/index.tsx
Comment (Ryukemeister) [reply apps/web/pages/event-types/[type]/index.tsx]: @alishaz-polymath @sean-brydon  so I saw the exact same issue of the dirtyFields not working as expected for event-types. [RHF docs](https://www.react-hook-form.com/api/useform/formstate/) state this `Important: Make sure to provide defaultValues at the useForm, so hook form can have a single source of truth to compare each field's dirtiness.` Is it possible that we've missed providing some default values in the useForm inputs for event types and that is causing this weird behaviour for the dirtyFields value?

Comment (alishaz-polymath) [reply apps/web/pages/event-types/[type]/index.tsx]: That was the first thing I checked as indeed that was the first suspect. As it turns it out the cause is more complicated than this.

Comment (Ryukemeister) [reply apps/web/pages/event-types/[type]/index.tsx]: @alishaz-polymath can you try to use the RHF `dirtyFields` again after PR #13440 gets merged? The default value for the destinationCalendar was not set previously which has been fixed in that PR so hopefully `dirtyFields` should work as expected now

Comment (alishaz-polymath) [reply apps/web/pages/event-types/[type]/index.tsx]: There are other fields I'm concerned about, not in particular destinationCalendar. Fields such as `bookingFieds` for example is populated in a different way and as such doesn't recognize changes made when adding a question, but when removing one. So, it's not directly related to defaultValues. This is one example, but there are other fields in a similar pot. 🙏

packages/features/ee/managed-event-types/lib/handleChildrenEventTypes.ts
```diff
@@ -130,14 +138,19 @@ export default async function handleChildrenEventTypes({
...
+  let newUserIds: number[] = [];
+  let oldUserIds = oldEventType.children?.flatMap((ch) => ch.userId ?? []);
+  if (childrenUpdated) {
+    // Calculate if there are new/existent/deleted children users for which the event type needs to be created/updated/deleted
+    const previousUserIds = oldEventType.children?.flatMap((ch) => ch.userId ?? []);
+    const currentUserIds = children?.map((ch) => ch.owner.id);
+    deletedUserIds = previousUserIds?.filter((id) => !currentUserIds?.includes(id));
+    newUserIds = currentUserIds?.filter((id) => !previousUserIds?.includes(id));
+    oldUserIds = currentUserIds?.filter((id) => previousUserIds?.includes(id));
+  }
```
alishaz-polymath, line 149: Only do this if children updated, otherwise the existing children are lost as `children=[]` here assumes no children/assignees.

alishaz-polymath (commented):
Self Review added

Udit-takkar (commented):
@alishaz-polymath Great work. only EventInstantTab is left. I can't enable instant booking when i tried.

To reproduce 
1) Login to org
2) Create a team event type.
3) You would see Instant Booking Tab

apps/web/components/eventtype/EventTypeSingleLayout.tsx
```diff
@@ -133,6 +134,60 @@ function getNavigation({
   ];
 }
 
+function DeleteDialog({
+  isManagedEvent,
```
Udit-takkar, line 138: on this page https://github.com/calcom/cal.com/blob/main/apps/web/components/eventtype/EventTypeSingleLayout.tsx#L311 and https://github.com/calcom/cal.com/blob/main/apps/web/components/eventtype/EventTypeSingleLayout.tsx#L427

alishaz-polymath: Missing shouldDirty or something else? 🤔

Udit-takkar: yeah

Udit-takkar (dismissed):
LGTM 🚀

CarinaWolli (dismissed):
enabling 'Choose a common schedule' in Availability tab doesn't work for me

Also, can we disable the save button when there are no dirty fields? It's weird to click on the save button and have nothing happen

emrysal (dismissed):
Hey @alishaz-polymath Almost everything works, well done!

On the Advanced Tab, if you make changes to the booking layout this works, and it updates. But if you continue to change other Advanced Tab before saving changes the booking layout change is not updated.

Checked against production and it does not happen there, so this is the one regression I could find.

emrysal:
I've checked each setting and no setting is blocked from saving so that's good enough for merge (as it's not the critical feature we're going for here)

But, the disabled button does behave very buggy, some settings and even tabs set the form to a dirty state (e.g. Availability immediately toggles the button to on, same for Advanced) and on toggling (mostly in Advanced) some settings revert to a non-dirty state and some continue to stay dirty, so depending on which one you toggle the behaviour may be different).

CarinaWolli (dismissed):

alishaz-polymath:
> Hey @alishaz-polymath Almost everything works, well done!
> 
> On the Advanced Tab, if you make changes to the booking layout this works, and it updates. But if you continue to change other Advanced Tab before saving changes the booking layout change is not updated.
> 
> Checked against production and it does not happen there, so this is the one regression I could find.

@emrysal Can you please check now if it's still a problem? I tested with the latest changes and it appears to be gone now.


@CarinaWolli Can you check the Availability common schedule and Hosts are now working as expected?

The save button is still something I'm thinking over, because RHF isDirty is unreliable, as we can see. 🤔

apps/web/components/eventtype/EventTeamTab.tsx
```diff
@@ -472,10 +472,12 @@ const Hosts = ({
       initialValue.current = { hosts: getValues("hosts"), schedulingType, submitCount };
       return;
     }
-    resetField("hosts", {
```
CarinaWolli, line 472: hosts weren't reset when changing the scheduling type as it wasn't set dirty

CarinaWolli:
> @CarinaWolli Can you check the Availability common schedule and Hosts are now working as expected?

I added one more fix to reset the hosts, now everything seems to work

apps/web/components/eventtype/EventTeamTab.tsx
alishaz-polymath: Thank you 🙌 🚀

alishaz-polymath:
Still not finding an efficient solution for the SAVE button's disabled state. I could potentially extract the dirtyFields check and then maybe set it up in a useEffect inside the

alishaz-polymath:
Closed thanks to unintended use of shortcut 😢 
Reopened 🙏

alishaz-polymath:
Still not finding an efficient solution for the SAVE button's disabled state. I could potentially extract the dirtyFields check and then maybe set it up in a useEffect inside the EventTypeSingleLayout.tsx file, but I'm wondering if it could be more efficient.
@emrysal Opinion?

emrysal (commented):
Approved 😎

apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -183,7 +183,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps,
               checked={useEventTypeDestinationCalendarEmail}
               onCheckedChange={(val) => {
                 setUseEventTypeDestinationCalendarEmail(val);
-                formMethods.setValue("useEventTypeDestinationCalendarEmail", val);
+                formMethods.setValue("useEventTypeDestinationCalendarEmail", val, { shouldDirty: true });
```
CarinaWolli, line 186: came in from the latest merge with main

CarinaWolli (dismissed):
Tested again and everything worked. Great work @alishaz-polymath  🙌🏻

emrysal (approved):
Let's get this merged, nice work!

PR #12290, opened 2023-11-09
fix: slug value in webhook payload

manpoffc:
## What does this PR do?
This PR fixes the slug value in the payload of webhook which was earlier returning the value of event title.

Fixes #12285  

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #13316, opened 2024-01-18
fix: event title in calendar event

CarinaWolli:
## What does this PR do?

Fixes that the calendar event description shows the event slug instead of the title under 'what'


packages/lib/CalEventParser.ts
```diff
@@ -14,7 +14,7 @@ const translator = short();
 export const getWhat = (calEvent: CalendarEvent, t: TFunction) => {
   return `
 ${t("what")}:
-${calEvent.type}
```
CarinaWolli, line 17: `type` used to be the event title but this got changed here: https://github.com/calcom/cal.com/pull/12290

PeerRich (approved):

zomars (approved):

PR #13319, opened 2024-01-18
fix: skip e2e tests that are memory crashing

zomars:
## What does this PR do?

Skips E2E tests that are currently failing on future app dir routes.


zomars:
@calcom/foundation will merge in order to unblock main pipeline 🙏🏽

PR #13323, opened 2024-01-19
fix: Make usernameList optional and make consistent return obj structure

### LINKED ISSUE #13322: [CAL-2971] API v1 Slots endpoint does not work for non-dynamic event types
There are 2 issues mainly:

1. The slots endpoint doesn't work for non-dynamic event types anymore due to the requirement of passing usernameList. 
2. The return object structure for when timezone is passed as a parameter is not consistent with when tiemzone isn't passed.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2971](https://linear.app/calcom/issue/CAL-2971/api-v1-slots-endpoint-does-not-work-for-non-dynamic-event-types)</sub>

alishaz-polymath:
## What does this PR do?

This PR sorts 2 things:
1. Makes usernameList optional so that it works like before for event-type slots (non dynamic), as is expected
2. When someone passes in a timezone, it now wraps the return in top-level `slots` key which is how we return it when there is no timezone passed.

Fixes #13322 

## Requirement/Documentation

NA
NA

## Type of change

## How should this be tested?

- Query the Slots endpoint with and without usernameList parameter passed in. It should work fine.
- Query the Slots endpoint with and without timezone parameter passed in, the response in both situations should be consistent in terms of object structure..

## Mandatory Tasks

### DIFF
--- a/apps/api/pages/api/slots/_get.ts
+++ b/apps/api/pages/api/slots/_get.ts
@@ -22,7 +22,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     const { usernameList, ...rest } = req.query;
     let slugs = usernameList;
     if (!Array.isArray(usernameList)) {
-      slugs = usernameList ? [usernameList] : [];
+      slugs = usernameList ? [usernameList] : undefined;
     }
     const input = getScheduleSchema.parse({ usernameList: slugs, ...rest });
     const timeZoneSupported = input.timeZone ? isSupportedTimeZone(input.timeZone) : false;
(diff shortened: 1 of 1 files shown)


apps/api/pages/api/slots/_get.ts
```diff
@@ -22,7 +22,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     const { usernameList, ...rest } = req.query;
     let slugs = usernameList;
     if (!Array.isArray(usernameList)) {
-      slugs = usernameList ? [usernameList] : [];
+      slugs = usernameList ? [usernameList] : undefined;
```
alishaz-polymath, line 25: Make slug optional so that users can query an event type alone and get slots for that event type.

apps/api/pages/api/slots/_get.ts
```diff
@@ -40,7 +40,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
         )
       : availableSlots;
 
-    return slotsInProvidedTimeZone;
+    return { slots: slotsInProvidedTimeZone };
```
alishaz-polymath, line 43: wraps the return in this case to the `slots` key to maintain consistent return object structure.

PeerRich (approved):

PR #13330, opened 2024-01-19
fix: fix-e2e-tests

dmytro068:
## What does this PR do?

Fixes `apps/web/playwright/ab-tests-redirect.e2e.ts`

## Type of change

## How should this be tested?

ab-tests-redirect.e2e.ts should pass

## Mandatory Tasks

### DIFF
--- a/apps/web/lib/buildLegacyCtx.tsx
+++ b/apps/web/lib/buildLegacyCtx.tsx
@@ -32,22 +32,18 @@ export const buildLegacyCtx = (
   params: Params,
   searchParams: SearchParams
 ) => {
-  return new Proxy<Record<string, any>>(
-    {
-      query: { ...searchParams, ...params },
-      params,
-      req: { headers: buildLegacyHeaders(headers), cookies: buildLegacyCookies(cookies) },
-    },
-    {
-      get(obj, prop) {
-        if (prop === "res") {
-          throw new Error(
-            "You are trying to access the 'res' property of the context, which is not supported in app dir"
-          );
-        }
-
-        return obj[prop as keyof typeof obj];
+  return {
+    query: { ...searchParams, ...params },
+    params,
+    req: { headers: buildLegacyHeaders(headers), cookies: buildLegacyCookies(cookies) },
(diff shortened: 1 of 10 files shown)


apps/web/lib/buildLegacyCtx.tsx
```diff
@@ -32,22 +32,18 @@ export const buildLegacyCtx = (
...
+  return {
+    query: { ...searchParams, ...params },
+    params,
+    req: { headers: buildLegacyHeaders(headers), cookies: buildLegacyCookies(cookies) },
+    res: new Proxy(Object.create(null), {
+      // const { req, res } = ctx - valid
+      // res.anything - throw
+      get() {
+        throw new Error(
+          "You are trying to access the 'res' property of the context, which is not supported in app dir"
```
grzpab, line 44: the error message is not adequate right now

apps/web/pages/auth/forgot-password/index.tsx
```diff
@@ -148,8 +148,9 @@ ForgotPassword.PageWrapper = PageWrapper;
 export const getServerSideProps = async (context: GetServerSidePropsContext) => {
   const { req, res } = context;
 
-  const session = await getServerSession({ req, res });
+  const session = await getServerSession({ req });
 
+  // @TODO res will not be available in future pages (app dir)
```
grzpab, line 153: will this throw in the future pages?

DmytroHryshyn: yes, it will throw when auth page is migrated. (it will throw in both legacy and future pages, because we pass legacyContext to both of them)

grzpab: what is the proper replacement? doing the following operation in the middleware?

DmytroHryshyn: yes, all code that needs to modify headers/cookies should be moved to middleware, because in pages headers/cookies are readonly and res is not available at all.

keithwillcode (approved):

PR #13336, opened 2024-01-19
chore: [app-router-migration 39] refactor ssr/ssg functions

dmytro068:
## What does this PR do?

Replaces ssgInit with `getTranslations` function that does not have dependency on trpc router. 
Refactors ssrInit, removes the import of trpc router, imports only specific procedures instead. 

The main reason of increased RAM usage in dev mode is trpc router that was imported from ssg/ssr init. The quick workaround for this issue is to import only procedures that are used on the server side. 

Fixes # (issue)
OOM issue in dev mode

## Type of change

## How should this be tested?

- measure RAM usage on  `/event-types` page on main branch (it will be ~7GB)
- switch to current branch, clear `.next` cache, measure RAM usage (should be ~2 GB)

## Mandatory Tasks

### DIFF
--- a/apps/web/app/not-found.tsx
+++ b/apps/web/app/not-found.tsx
@@ -2,13 +2,13 @@ import NotFoundPage from "@pages/404";
 import { WithLayout } from "app/layoutHOC";
 import type { GetStaticPropsContext } from "next";
 
-import { ssgInit } from "@server/lib/ssg";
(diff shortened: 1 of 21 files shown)


zomars (approved):
Awesome improvement @DmytroHryshyn 🙏🏽

PR #13345, opened 2024-01-20
fix: browser back button not working

### LINKED ISSUE #13324: [CAL-2972] Back button on the booking page is not working properly
[https://www.loom.com/share/69d30529b3e44b8dae7e9953ab5ba9d7?sid=010987a8-55d9-4832-ac1d-2db8dfbd5b14](https://www.loom.com/share/69d30529b3e44b8dae7e9953ab5ba9d7?sid=010987a8-55d9-4832-ac1d-2db8dfbd5b14)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2972](https://linear.app/calcom/issue/CAL-2972/back-button-on-the-booking-page-is-not-working-properly)</sub>

abhijeetsingh-22:
## What does this PR do?
Fixes a bug due to which the back button on the booking page was not working.
The issue was due to multiple calls to `setQueryParams` and `removeQueryParams` as it was pushing multiple URLs to the browser history.  

## My fix
Instead of pushing to the history I am now replacing the last entry in the history.

Fixes #13324 

https://github.com/calcom/cal.com/assets/55000107/79f9645e-2264-4f47-bb85-d00c81fec69d

## Type of change

## How should this be tested?
- Go to the booking page of a user with multiple events present. 
- Open one of the events and change the date selected.
- Press the browser back button (You should be taken back to the user page.)
- Try repeating this with the other event .

## Mandatory Tasks

- [ x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.


packages/features/calendars/DatePicker.tsx
```diff
@@ -192,10 +192,15 @@ const Days = ({
...
-
+    if (isSelectedDateAvailable) {
+      props.onChange(dayjs(selected));
+    }
     if (!firstAvailableDateOfTheMonth) {
       props.onChange(null);
     }
+    return () => {
+      props.onChange(null);
+    };
```
PeerRich, line 203: what do these do?

abhijeetsingh-22: - After we open an event from the user page and then click back button the date queryParam is removed but the selectedDate is still present in the store. So this code sets the Date queryParam on subsequent visit to event page.
- The cleanup function makes sure to remove the query parameter on unmounting ( going back to user page)

Udit-takkar (dismissed):
https://github.com/calcom/cal.com/assets/53316345/ca34a6c2-b752-45bb-90fa-58c45bad5dec

I should be back to the booking form page and not /[user]

emrysal:
@abhijeetsingh-22 looks like Udit is right; not an issue with your PR but it looks like the history is not set properly (perhaps due to next router) when coming from the booking page?

abhijeetsingh-22:
> @abhijeetsingh-22 looks like Udit is right; not an issue with your PR but it looks like the history is not set properly (perhaps due to next router) when coming from the booking page?

@emrysal The problem is not the next-router but the way we are handling the queryParams in the /[user]/[type] . I have made the changes have a look


https://github.com/calcom/cal.com/assets/55000107/ab09f63e-511c-41dd-ad59-277c7b3025fa

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -213,6 +213,10 @@ const BookerComponent = ({
     return setBookerState("booking");
   }, [event, selectedDate, selectedTimeslot, setBookerState]);
 
+  const slot = getQueryParam("slot");
+  useEffect(() => {
+    setSelectedTimeslot(slot || null);
```
abhijeetsingh-22, line 218: This makes sure that the `selectedTimeslot` updates in the store when it is changed in the query param. And it sets it to null when the `slot` query param is not present

packages/features/bookings/Booker/utils/query-param.ts
```diff
@@ -2,13 +2,19 @@ export const updateQueryParam = (param: string, value: string | number) => {
...
   if (value === "" || value === "null") {
-    url.searchParams.delete(param);
+    removeQueryParam(param);
+    return;
   } else {
     url.searchParams.set(param, `${value}`);
   }
-
-  window.history.pushState({}, "", url.href);
+  if (param == "slot") {
```
abhijeetsingh-22, line 13: For the desired behavior, we  need to smartly choose between `pushState` and `replaceState` so that the browser history is created the way we want.  This is  in-line with the change that @Udit-takkar requested earlier.

PeerRich (dismissed):

packages/features/calendars/DatePicker.tsx
PeerRich: ahh I see. @CarinaWolli @emrysal whats your take on this?

emrysal:
@abhijeetsingh-22 before approving just want to run something by you - do you think it makes sense to move the "is it a push/replace" to another parameter in the setQueryParam function? Just because the literal check for "slot" feels like too much of an unexpected side-effect to me.

abhijeetsingh-22:
> @abhijeetsingh-22 before approving just want to run something by you - do you think it makes sense to move the "is it a push/replace" to another parameter in the removeQueryParam function? Just because the literal check for "slot" feels like too much of an unexpected side-effect to me.

I think it makes sense to move `isReplace` to function parameter. It will make the function more reusable. Thanks for the suggestion I will make the changes and push shortly

PeerRich:
/bonus 20

hey @abhijeetsingh-22 any chance you can take a look at the merge conflict and the changes mentioned above?

abhijeetsingh-22:
Got a little busy. Will do it today.

abhijeetsingh-22:
> /bonus 20
> 
> 
> 
> hey @abhijeetsingh-22 any chance you can take a look at the merge conflict and the changes mentioned above? 

Got a little busy. Will do it today

abhijeetsingh-22:
@emrysal made the changes pl review again.

abhijeetsingh-22:
> Hello guys. I've stumbled across this issue and wanted to add something that may be potentially useful or maybe without any importance. In Next's Router (from `next/router`) object and in the `Link` component (`next/link`) there's an option to perform [shallow routing.](https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing) that I've always been using for such cases. It works exactly as we'd like it to, replacing url without altering the browser's history state.
> 
> It's just nicer to write:
> 
> ```js
> router.replace(Url.toString(), undefined, { shallow: isShallow })
> ```
> 
> Than to
> 
> ```js
> if (shouldReplace) {
>     window.history.replaceState({ ...window.history.state, as: url.href }, "", url.href);
> } else {
>     window.history.pushState({ ...window.history.state, as: url.href }, "", url.href);
> }
> ```
> 
> (that btw seems to be duplicated in this PR 😅 once in `updateQueryParam` and secondly in `removeQueryParam` instead of being pushed outside as an `util` function)
> 
> Anyway, I'm not sure if this brings any value to the issue. Can't wait for this major bug fix to happen 🙌

Thanks for the suggestion. But here are few things to consider
- The `router` comes from `useRouter` so if we want to use it we need to convert the `query-param` util to a hook.
- Even if we use `router` we have to have an additional param that will tell if we should use `router.replace` or `router.push` as for setting slot we should be using push because we need to have a browser history entry for it and for other query-params we should be using replace.

emrysal (dismissed):
Happy with the changes and justification of using window.history - useRouter is next specific and we don't want to depend on it here.

Thank you @abhijeetsingh-22

sean-brydon (dismissed):
Agreed - with the comments made about window.location - this is what we discussed when we first setup this atom. 

Thanks!

PeerRich:
checks are failing @abhijeetsingh-22

CarinaWolli (dismissed):
The changes cause booking-limits e2e tests to fail. 

I found one issue that happens in https://github.com/calcom/cal.com/blob/main/apps/web/playwright/booking-limits.e2e.ts#L306. For monthly duration limits, it doesn't redirect to the booking page but stays on the availability page instead

abhijeetsingh-22:
I will look into it

CarinaWolli (approved):

devklepacki:
Yes! Couldn't wait for this! I'm so thankful @abhijeetsingh-22! Now what's the timeline for this to be released?

PR #13361, opened 2024-01-22
feat: user timezone management in `CalProvider`

Ryukemeister:
## What does this PR do?

Check for changes in a users preferred timezone and the timezone that a user is currently in and update the timezone accordingly

### DIFF
--- a/packages/platform/atoms/hooks/useUpdateUserTimezone.ts
+++ b/packages/platform/atoms/hooks/useUpdateUserTimezone.ts
@@ -1,0 +1,24 @@
+import { useMutation } from "@tanstack/react-query";
+
+import type { ApiResponse } from "@calcom/platform-types";
+
+type updateTimezoneInput = {
+  timeZone: string;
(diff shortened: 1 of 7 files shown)


packages/platform/atoms/hooks/useUpdateUserTimezone.ts
```diff
@@ -0,0 +1,24 @@
...
+  timeZone: string;
+  key: string;
+};
+
+export const useUpdateUserTimezone = () => {
+  const mutation = useMutation<ApiResponse<undefined>, unknown, updateTimezoneInput>({
+    mutationFn: (data) => {
+      const { key, timeZone } = data;
+
+      return fetch(`/api/v2/me?apiKey=${key}`, {
```
sean-brydon, line 15: I think we should construct proper query params here to make it expandable if we need 

Check out `new URL(...)`

ThyMinimalDev (approved):

PR #13374, opened 2024-01-23
feat: Added new API route to get event-types using slug

### LINKED ISSUE #12411: [CAL-2745] API for getting an event type by slug

abhijeetsingh-22:
## What does this PR do?

Added a new API route to get the `event-types` using the `slug` . 

Closes #12411 
/claim #12411

### Rationale for Introducing a New API Route
There were two options: either to update the existing `/event-types/{id}` route to `/event-types/{idOrSlug}`  or to add a new route `/event-types/bySlug/{slug}`. 
I chose to add a new route as updating the existing one to `/event-types/{idOrSlug}` would make it ambiguous whether the event-type is fetched using slug or the id.

## Type of change
- This change requires a documentation update (need to update API docs [here](https://cal.com/docs/enterprise-features/api/api-reference/event-types) )

## How should this be tested?
- Try to get the event-types using API `/event-types/bySlug/{slug}`
- If a valid `API key ` is passed and `event-type` with the `slug` exists for the user, it should be returned; otherwise forbidden. 

## Mandatory Tasks

## Checklist
All done


apps/api/lib/validations/shared/querySlug.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const schemaQuerySlug = baseApiParams
+  .extend({
+    slug: z.string(),
+    ownerId: z
+      .string()
+      .optional()
+      .transform((value) => (value ? parseInt(value, 10) : null)),
+  })
+  .strict();
```
supalarry, line 13: I refactored this file in commit [refactor: schema query slug](https://github.com/calcom/cal.com/pull/13374/commits/9c6b51858d5252b2312438417a25671865f09ca2):
1. There is no need to include "String" in querySlugString.ts because slug is a string already, so i renamed file from "querySlugString.ts" to "querySlug.ts"
2. "export const withValidQuerySlugString" was not used anywhere in codebase so i deleted it
3. Admins should be able to access event types of other users, so optional "ownerId" query parameter has been added in commit [feat: allow admins to fetch event types on behalf of others](https://github.com/calcom/cal.com/pull/13374/commits/6765cc7b049607234b4440bbe436a1aa0931ac0d).

apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  const isAuthorized = isAdmin || eventType.users.some((user) => user.id === effectiveUserId);
+
+  if (!isAuthorized) {
+    throw new HttpError({ statusCode: 403, message: "Forbidden" });
+  }
+}
+
+export default authMiddleware;
```
Comment (supalarry) [inline apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts:33]: I refactored this file in commit [feat: allow admins to fetch event types on behalf of others](https://github.com/calcom/cal.com/pull/13374/commits/6765cc7b049607234b4440bbe436a1aa0931ac0d):
1. Before you could pass non existing slug name and the "authMiddleware" would pass without errors. Now it throws 404.
2. If user is not admin and his / her user Id is not part of event type, then throw 403 Forbidden. Also, using effectiveUserId allow admins to fetch event types on behalf of others. Done in commit [feat: allow admins to fetch event types on behalf of others](https://github.com/calcom/cal.com/pull/13374/commits/6765cc7b049607234b4440bbe436a1aa0931ac0d)

apps/api/pages/api/event-types/by-slug/[slug]/_get.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { schemaEventTypeReadPublic } from "~/lib/validations/event-type";
+import { schemaQuerySlug } from "~/lib/validations/shared/querySlug";
+
+import { checkPermissions } from "../../[id]/_get";
+import getCalLink from "../../_utils/getCalLink";
+
+/**
+ * @swagger
+ * /event-types/by-slug/{slug}:
```
Comment (supalarry) [inline apps/api/pages/api/event-types/by-slug/[slug]/_get.ts:13]: Renamed the route from "/event-types/bySlug" to "/event-types/by-slug" to follow same case in commit [refactor: event-types/bySlug to event-types/by-slug route](https://github.com/calcom/cal.com/pull/13374/commits/372e1bde94fad306f1ba66cb5a3f730ddb7c1bd0)

apps/api/pages/api/event-types/by-slug/[slug]/_get.ts
```diff
@@ -0,0 +1,84 @@
...
+ *       401:
+ *        description: Authorization information is missing or invalid.
+ *       404:
+ *         description: EventType was not found
+ */
+export async function getHandler(req: NextApiRequest) {
+  const { userId, prisma, isAdmin, query } = req;
+  const { slug, ownerId } = schemaQuerySlug.parse(query);
+
+  const effectiveUserId = isAdmin && ownerId ? ownerId : userId;
```
Comment (supalarry) [inline apps/api/pages/api/event-types/by-slug/[slug]/_get.ts:53]: Allow admins to fetch event types on behalf of other users

supalarry (dismissed):
Thank you for your contribution Abhijeet! 💯

I refactored code to make it cleaner and provide clearer error messages to the user and enabled admins to fetch event types on behalf of other users - admins being able to perform actions on behalf of other users is a requirement in our API.

abhijeetsingh-22:
@supalarry Thanks for the detailed review. But I have some concerns. 

- The `event-type` cannot be queried from the DB just by using slug as slugs are not globally unique. A `userId` is required in the prisma query to make sure we are getting the correct `event-type`

apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { schemaQuerySlug } from "~/lib/validations/shared/querySlug";
+
+async function authMiddleware(req: NextApiRequest) {
+  const { userId, isAdmin, prisma } = req;
+  const { slug, ownerId } = schemaQuerySlug.parse(req.query);
+
+  const effectiveUserId = isAdmin && ownerId ? ownerId : userId;
+
+  const eventType = await prisma.eventType.findFirst({
```
Comment (abhijeetsingh-22) [inline apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts:13]: This will always return the first `event-type` with the given `slug` from the DB. So this will always throw forbidden for all the users except admin and the owner of the first `event-type` with the given slug.

Comment (supalarry) [reply apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts:13]: Thank you for your fast review, appreciate and well spotted! Pushed update in commit [fix: auth middleware](https://github.com/calcom/cal.com/pull/13374/commits/3123ae4b6249995668c3911b87017274eb3b9751).

supalarry (dismissed):

apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts
Comment (supalarry) [reply apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts:13]: Alternatively, I was thinking we could `.findUnique` by `EventType.slug` and `EventType.userId` 🤔

Comment (abhijeetsingh-22) [reply apps/api/pages/api/event-types/by-slug/[slug]/_auth-middleware.ts:13]: No, the `.findUnique` expects at least one unique field. I think `.findFirst` is good here.

keithwillcode (changes requested):
Having an extra route with by-slug is not ideal. According to REST specs, when you add another URL segment, the resource you are getting should be nested within that parent resource. This is not the case here. Instead, we are looking up the event-type resource but just using a different param to search by. 

Also, the slug is not unique across all users/teams in the system so finding a slug in this manner doesn’t make sense for system admins. 

we need to consider a different way to get this data. potential options I see (will dig further to come up with more options on Monday)
- update the main GET for event-types to add a query param of slug to filter by that slug. For system admins, this would find all event types that have that slug. For normal users and teams, it would be limited to their event types with that slug. 
- make this searchable by team/user routes

abhijeetsingh-22:
@keithwillcode I agree with you. It makes sense to have the `slug` as a query params in the `/event-type` route. That way the admins can get all the `event-type` with that slug and inidvidual users and temas can get their `event-type`. I can make the changes once you confirm .

keithwillcode:
> @keithwillcode I agree with you. It makes sense to have the `slug` as a query params in the `/event-type` route. That way the admins can get all the `event-type` with that slug and inidvidual users and temas can get their `event-type`. I can make the changes once you confirm .

Yeah let’s add it [here](https://github.com/calcom/cal.com/blob/cc0ff17d975f49223f6e686550beaa30aa033ba0/apps/api/pages/api/event-types/_get.ts#L37)

abhijeetsingh-22:
@keithwillcode, thanks for confirmation. I am making the required changes. I was thinking should we also add other optional query params to this route, like `length`, `title`, `price` to filter the `event-type`

supalarry (approved):
Amazing work Abhijeet! 🚀

I tested the updated endpoints and it respects what Keith added and works as expected!

keithwillcode (approved):

aecorredor:
Is this change available in production already? I'm testing with valid slugs and get 404 every time.

PR #13386, opened 2024-01-24
fix: Namespaced react embeds not working with Floating and Element Click Popups

### LINKED ISSUE #12983: [CAL-3366] Avoid duplication of inline embed for same namespace

### LINKED ISSUE #13388: [CAL-3012] Fix embed behaviour with soft navigations when namespaces are used
1. If the element added by inline embed is removed, re-executing the  **inline** instruction is ignored.
2. If a namespace is created after embed.js has executed it doesn't work at all
3. Repeat execution of **init** instruction causes the namespace to stop executing further instructions

Also, we need to add Unit/E2E tests to verify that the API keeps working in all these cases.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3012](https://linear.app/calcom/issue/CAL-3012/fix-embed-behaviour-with-soft-navigations-when-namespaces-are-used)</sub>

### LINKED ISSUE #13458: [CAL-3052] Floating Button Popup and Element Click Popup don't work with namespaces
Initially reported on Intercom.
Private Link: [https://app.intercom.com/a/inbox/z9l4o9ru/inbox/conversation/5409?view=List](https://app.intercom.com/a/inbox/z9l4o9ru/inbox/conversation/5409?view=List)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3052](https://linear.app/calcom/issue/CAL-3052/floating-button-popup-and-element-click-popup-dont-work-with)</sub>

hariombalhara:
## What does this PR do?

Fixes #13388 
Fixes #13458
Fixes #12983

- `getCalApi` didn't have support for namespace. So, floating and element-click popup code for React which uses `getCalApi` wasn't working with namespaces. - Support has been added to `getCalApi` for namespace now in a backward compatible fashion.  Added e2e tests as well for floating and element-click popup in embed-react. The issue wasn't caught because we were just having automation tests for inline embed in embed-react.
- The case where a namespace was created after embed.js was loaded, wasn't working. It wasn't part of playground. So, added the flow in playground  and added an e2e test as well.
- The case where someone re-initializes a namespace and gives repeat instruction for 'inline' embed is now handled.  We won't add inline embed twice now. Added e2e as well to test this scenario for both default and non-default namespace.

## Type of change

## How should this be tested?
See the loom https://www.loom.com/share/f627dfba74b8477aa4e88ea85eebf748

## Mandatory Tasks


hariombalhara:
* `main`
  * **PR #13386** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/13386?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/13386?utm_source=stack-comment).

packages/embeds/embed-react/src/index.ts
```diff
@@ -1,14 +1,20 @@
...
+import type { GlobalCal, GlobalCalWithoutNs } from "@calcom/embed-core";
 import EmbedSnippet from "@calcom/embed-snippet";
 
 import Cal from "./Cal";
 
-export const getCalApi = (embedJsUrl?: string): Promise<GlobalCal> =>
-  new Promise(function tryReadingFromWindow(resolve) {
-    EmbedSnippet(embedJsUrl);
-    const api = window.Cal;
+export const getCalApi = (options?: {
```
hariombalhara, line 8: We are changing the API here in a breaking fashion. 

Users who are using older version of embed-react, would use the old API(but that won't support namespacing) and those moving to new versions would have to use the new API which supports namespacing.

packages/embeds/embed-react/src/index.ts
```diff
@@ -1,14 +1,20 @@
...
+  embedJsUrl?: string;
+  namespace?: string;
+}): Promise<GlobalCal | GlobalCalWithoutNs> => {
+  options = options || {};
```
hariombalhara, line 12: We might also support options as a string as well which is backward compatible. 🤔

packages/embeds/embed-core/src/embed.ts
```diff
@@ -415,8 +415,17 @@ class CalApi {
...
 
   init(namespaceOrConfig?: string | InitArgConfig, config = {} as InitArgConfig) {
+    let namespace = "";
     if (typeof namespaceOrConfig !== "string") {
       config = (namespaceOrConfig || {}) as Config;
+    } else {
+      namespace = namespaceOrConfig;
+    }
+
+    if (!globalCal.ns[namespace].instance && namespace !== this.cal.namespace) {
```
hariombalhara, line 425: If a namespace encounters the instruction for some other namespace, create an instance for that namespace and that would handle related work.

Ideally this should happen with default namespace encountering other namespace instruction. We can explicitly confirm that here I think.

packages/embeds/embed-core/index.html
```diff
@@ -51,6 +51,8 @@
               const namespace = ar[1];
               api.q = api.q || [];
               typeof namespace === "string" ? (cal.ns[namespace] = api) && p(api, ar) : p(cal, ar);
+              // Also send init to the default namespace queue so that the namespace instance can be created which will then take care of it's own queue
+              p(cal, arguments)
```
hariombalhara, line 55: Need to put this change in embed-snippet as well and other places where we show Cal Snippet.

packages/embeds/embed-core/src/embed.ts
```diff
@@ -458,6 +480,12 @@ class CalApi {
       },
     });
 
+    // If someone re-executes inline embed instruction, we want to ensure that duplicate inlineEl isn't added to the page per namespace
+    if (this.cal.inlineEl && document.body.contains(this.cal.inlineEl)) {
+      console.warn("Inline embed already exists. Ignoring this call");
```
hariombalhara, line 485: Adding it back what I reverted here https://github.com/calcom/cal.com/pull/13637/files#diff-e895c26df52266ecfd6334060b1dc6fa05cac4106eee7f713d0a07472cc08d07

Now I am checking for the presence of inlineEl in DOM instead of just checking the property which could be out-of-sync with DOM. So, it seems safe to add it back.

apps/web/playwright/fixtures/embeds.ts
```diff
@@ -43,14 +48,18 @@ export const createEmbedsFixture = (page: Page) => {
               // eslint-disable-next-line @typescript-eslint/ban-ts-comment
               //@ts-ignore
               api = window.Cal.ns[calNamespace];
+              console.log("Using api from namespace-", { calNamespace, api });
             }
-            console.log("PlaywrightTest:", `Adding listener for __iframeReady on namespace:${calNamespace}`);
             if (!api) {
-              throw new Error(`namespace "${calNamespace}" not found`);
+              console.log(`namespace "${calNamespace}" not found yet - Trying again`);
+              setTimeout(tryAddingListener, 500);
```
hariombalhara, line 55: Handle the case when namespace is created after a CTA is clicked and not on page load.

packages/embeds/embed-react/src/index.ts
```diff
@@ -1,14 +1,32 @@
...
 
-export const getCalApi = (embedJsUrl?: string): Promise<GlobalCal> =>
-  new Promise(function tryReadingFromWindow(resolve) {
-    EmbedSnippet(embedJsUrl);
-    const api = window.Cal;
+export function getCalApi(options?: {
+  embedJsUrl?: string;
+  namespace?: string;
+}): Promise<GlobalCal | GlobalCalWithoutNs>;
+export function getCalApi(embedJsUrl: string): Promise<GlobalCal | GlobalCalWithoutNs>;
```
hariombalhara, line 12: This is backward compatible signature.

packages/embeds/embed-core/src/embed.ts
```diff
@@ -211,7 +211,7 @@ export class Cal {
     return instruction;
   }
 
-  processQueue(queue: IArguments[]) {
+  processQueue(queue: Queue) {
```
hariombalhara, line 214: Using custom type here.

packages/embeds/embed-core/src/embed.ts
```diff
@@ -766,13 +795,16 @@ class CalApi {
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Queue = any[];
```
hariombalhara, line 799: A queue basically holds the arguments to all the instruction functions, so it could be anything.

packages/embeds/embed-core/src/preview.ts
```diff
@@ -7,41 +9,7 @@ if (!bookerUrl || !embedLibUrl) {
...
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        //@ts-ignore
-        typeof namespace === "string" ? (cal.ns[namespace] = api) && p(api, ar) : p(cal, ar);
-        return;
-      }
-      p(cal, ar);
-    };
-})(window, embedLibUrl, "init");
-
+EmbedSnippet();
```
hariombalhara, line 12: Reuse embed snippet

sean-brydon (approved):
Tested this - can confirm it fixes the issues mentioned. 

Thanks for the loom and tests were helpful <3

packages/embeds/embed-core/src/preview.ts
```diff
@@ -26,22 +29,25 @@ if (!bookerUrl || !embedLibUrl) {
...
+          // Make sure that even after re-execution of the snippet, the namespace is not overridden
+          cal.ns[namespace] = cal.ns[namespace] || api;
+          p(cal.ns[namespace], ar);
+          p(cal, ["initNamespace", namespace]);
+        } else p(cal, ar);
         return;
       }
       p(cal, ar);
     };
-})(window, embedLibUrl, "init");
```
hariombalhara, line 43: It caused the previews to break :(

PR #13396, opened 2024-01-24
chore: Proposal for new module based file structure

zomars:
## What does this PR do?

Extracts `bookings` related routes into it's own module in `apps/web`

## Why?

In a recent call with @DmytroHryshyn and @grzpab  from the Codemod team we found out some of reasons we're experiencing crashes is because of mixed imports and circular dependencies. This PR propose a file structure that will help mitigate this and get these benefits:

- Benefit from a feature driven structure inside `apps/web`
- We can re-export them in both `pages` and future `app` routes until migration is completed
- Prevent OOM issues due to mixed imports
- Prevent errors due to circular dependencies.

## Type of change

## How should this be tested?

- yarn build
- `app` and `pages` routes should be working as expected:
- /bookings
- /future/bookings
- /booking/[uid]
- /future/booking/[uid]

## Mandatory Tasks

## Checklist


apps/web/app/future/booking/[uid]/embed/page.tsx
```diff
@@ -1,9 +1,10 @@
-import { type PageProps, getServerSideProps } from "@pages/booking/[uid]";
-import OldPage from "@pages/booking/[uid]/embed";
 import { withAppDirSsr } from "app/WithAppDirSsr";
 import withEmbedSsrAppDir from "app/WithEmbedSSR";
 import { WithLayout } from "app/layoutHOC";
 
+import OldPage from "~/bookings/views/bookings-single-view";
+import { getServerSideProps, type PageProps } from "~/bookings/views/bookings-single-view.getServerSideProps";
```
Comment (zomars) [inline apps/web/app/future/booking/[uid]/embed/page.tsx:6]: This avoids circular dependencies between `pages` and `app`

apps/web/modules/bookings/lib/validStatuses.ts
zomars, line 1: Extracted reused constant

apps/web/modules/bookings/views/bookings-listing-view.tsx
zomars, line 1: This is just the bookings page (or view IMO) without the server side code

zomars (commented):
Ready for review

apps/web/pages/bookings/[status].tsx
Comment (zomars) [inline apps/web/pages/bookings/[status].tsx:1]: This now lives in `apps/web/modules/bookings/views/bookings-listing-view.tsx`

apps/web/modules/bookings/views/bookings-listing-view.tsx
```diff
@@ -0,0 +1,257 @@
...
+              </div>
+            )}
+          </div>
+        </main>
+      </div>
+    </ShellMain>
+  );
+}
+
+Bookings.PageWrapper = PageWrapper;
```
DmytroHryshyn, line 256: What do you think if we assign `PageWrapper` and `getLayout` in the page file, not in the view. In this case it will be more reusable by `app` and `pages`. If we keep `PageWrapper` in the view, then resulting client bundle will include both `PageWrapper` and `PageWrapperAppDir` which adds few kb to the size.

apps/web/pages/booking/[uid]/embed.tsx
```diff
@@ -1,8 +1,9 @@
 "use client";
```
Comment (DmytroHryshyn) [inline apps/web/pages/booking/[uid]/embed.tsx:1]: Legacy page does not need "use client" now, we import view directly to the future page.

keithwillcode (dismissed):

apps/web/modules/bookings/views/bookings-listing-view.tsx
zomars: Will do.

zomars:
addressed feedback. Please re-review @keithwillcode 🙏🏽

keithwillcode (approved):

PR #13407, opened 2024-01-25
feat: store event-type filters to localStorage

### LINKED ISSUE #13394: [CAL-3015] event-type filters dont persist
we should store the setting here in localStorage

CleanShot 2024-01-24 at 18 08 23@2x

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3015](https://linear.app/calcom/issue/CAL-3015/event-type-filters-dont-persist)</sub>

abhijeetsingh-22:
## What does this PR do?
This PR introduces a new feature that stores `event-type` filter values in localStorage, ensuring that the selected filter persists when users reload the page.

Fixes #13394 

https://github.com/calcom/cal.com/assets/55000107/228f2be2-7139-471a-ac2a-81b2440f1aea

## Type of change

## How should this be tested?
- Open `/event-types`.
- Change the event-type filter .
- Reload the page. 

### Desired Outcome:
 The event-type filter you selected previously should persist, and the query parameters should be appropriately populated.

## Mandatory Tasks

## Checklist
All done


packages/lib/hooks/useTypedQuery.ts
```diff
@@ -32,7 +33,11 @@ export const queryStringArray = z
   .preprocess((a) => z.string().parse(a).split(","), z.string().array())
   .or(z.string().array());
 
-export function useTypedQuery<T extends z.AnyZodObject>(schema: T) {
+export function useTypedQuery<T extends z.AnyZodObject>(
+  schema: T,
+  useLocalStorage = false,
+  localStorageKey = ""
```
supalarry, line 39: How about we remove `useLocalStorage` key and only rely on `localStorageKey` - if `localStorageKey` is not passed (can set it's default value to null) then `useLocalStorage` would be always false. If developer passes `localStorageKey` then it implicitly implies that developer wants to use local storage. So I think its better to have for simplicity:

```
export function useTypedQuery<T extends z.AnyZodObject>(
  schema: T,
  localStorageKey = null
) {
```

packages/features/filters/components/TeamsFilter.tsx
```diff
@@ -17,9 +17,10 @@ import { filterQuerySchema } from "../lib/getTeamsFiltersFromQuery";
 export type IEventTypesFilters = RouterOutputs["viewer"]["eventTypes"]["listWithTeam"];
 export type IEventTypeFilter = IEventTypesFilters[0];
 
+export const TEAMS_FILTER_KEY = "TEAMS_FILTER_KEY";
```
supalarry, line 20: To reduce possibility that "TEAMS_FILTER_KEY" is used elsewhere I would append "EVENT_TYPES_TEAMS_FILTER_KEY" in front to reduce possibility of collision but its your call.

supalarry (commented):
Great work as always Abhijeet! I tested PR locally and it works.

I have 1 suggestion on simplifying the usage of "useTypedQuery" by passing only 2 parameters. After that it is good.

packages/lib/hooks/useTypedQuery.ts
supalarry: What do you think?

abhijeetsingh-22: @supalarry thanks for the review. You are right we should remove the redundant parameter. I will make the changes and push

packages/features/filters/components/TeamsFilter.tsx
abhijeetsingh-22: Sure will also change this

packages/lib/hooks/useTypedQuery.ts
supalarry: Amazing!

packages/features/filters/components/TeamsFilter.tsx
supalarry: Yeah actually now that i think about it the current implementation is good - i mistakenly thought the file was in event-types related folder.

packages/lib/hooks/useTypedQuery.ts
PeerRich: awesome thank you! will approve after the change

abhijeetsingh-22:
Made the requested changes

supalarry:
> Made the requested changes

Amazing, looks good now! 🚀

supalarry (approved):

PR #13427, opened 2024-01-26
fix: prefill avatar when signing up

### LINKED ISSUE #13426: [CAL-3036] prefill avatar when signing up with Google
example: https://github.com/dubinc/dub/blob/main/apps/web/lib/auth/options.ts#L195

<sub>[CAL-3036](https://linear.app/calcom/issue/CAL-3036/prefill-avatar-when-signing-up-with-google)</sub>

neo773:
## What does this PR do?
prefills avatar when signing up if it exists

Fixes #13426

## Type of change

/claim #13426

### DIFF
--- a/packages/features/auth/lib/next-auth-options.ts
+++ b/packages/features/auth/lib/next-auth-options.ts
@@ -788,6 +788,7 @@ export const AUTH_OPTIONS: AuthOptions = {
             username: orgId ? slugify(orgUsername) : usernameSlug(user.name),
             emailVerified: new Date(Date.now()),
             name: user.name,
+            ...(user.image && { avatarUrl: user.image }),
             email: user.email,
             identityProvider: idP,
             identityProviderId: account.providerAccountId,


packages/features/auth/lib/next-auth-options.ts
```diff
@@ -788,6 +788,7 @@ export const AUTH_OPTIONS: AuthOptions = {
             username: orgId ? slugify(orgUsername) : usernameSlug(user.name),
             emailVerified: new Date(Date.now()),
             name: user.name,
+            ...(user.image && { avatarUrl: user.image }),
```
PeerRich, line 791: this would save the google URL right? I wonder if we should convert this to base64 and save on our server? 🤔 @emrysal

neo773: yeah, I think saving the URL should be fine as it's static and never expires

PeerRich (approved):

PR #13436, opened 2024-01-27
fix: allow all sign in methods when being invited

### LINKED ISSUE #13433: [CAL-3039] allow "Sign in with Google" when being invited
right now you can only use email + password after being invited to a team or to an Org.

lets enable all sign in methods 

<sub>[CAL-3039](https://linear.app/calcom/issue/CAL-3039/allow-sign-in-with-google-when-being-invited)</sub>

anikdhabal:
## What does this PR do?

Fixes #13433


PeerRich:
thank you for the PR! we will test it this week

PeerRich:
can you record a loom.com demo?

anikdhabal:
> can you record a loom.com demo?

Yeah, sure.

Udit-takkar (changes requested):
I created a team and invited an user by email and I opened the link from the email and tried using google sign in instead of email and password.

1) I shouldn't be seeing all the connected google emails but only the one which was used to invite me. If i select any other email then it just created a new account for me ignoring the team invitation process.

2) Also i got this error when i used google sign in on signup page when using the same email that was used to invite me

Screenshot 2024-02-14 at 2 40 37 PM

Udit-takkar:
This is the complete error message 

http://localhost:3000/api/auth/error?error=
Invalid `prisma.user.update()` invocation:

An operation failed because it depends on one or more records that were required but not found. No 'UserPassword' record was found for a nested delete on relation 'UserToUserPassword'.

keithwillcode:
We reverted out the UserPassword changes so pulling into main ideally resolves this issue you're seeing @Udit-takkar

joeauyeung (dismissed):
I agree with @Udit-takkar. Could we add the prompt as `force_login` so the user can't select a different account.

Also we should add some safeguard if the email that's returned from the SSO callback doesn't match the email on the invite.

keithwillcode:
This PR is being closed due to a lack of activity. Please reopen if work is to continue.

Udit-takkar (changes requested):
When i tried coping the invite link of a team. he callbackUrl and token was removed from url when i clicked on "Don't have an account" to sign up.

https://github.com/user-attachments/assets/008138eb-14c4-442e-ac2c-846b06026c61

anikdhabal:
> When i tried coping the invite link of a team. he callbackUrl and token was removed from url when i clicked on "Don't have an account" to sign up.
Hey Udit, I don't fully understand the issue. Could you please explain it a bit more?  Also, from what I can gather, this seems to be a different bug. My changes are limited to only OAuth. I'll address that as well

Udit-takkar (approved):
LGTM.

Udit-takkar:
@anikdhabal try inviting a user using "Copy invite link" option on the modal. Open that url in incognito and you would be directed to login page to login and then when you try to sign up for a new account the token/callbackurl from url parameters is missing.

anikdhabal:
> @anikdhabal try inviting a user using "Copy invite link" option on the modal. Open that url in incognito and you would be directed to login page to login and then when you try to sign up for a new account the token/callbackurl from url parameters is missing.

Hey Udit, the link we copied from there is for users who already have an account. For the signup case, we need a pre-populated email, which we get from the email invitation. This is misleading now, as users don’t have any idea about that

Udit-takkar:
> For the signup case, we need a pre-populated email, which we get from the email invitation.

It used to work fine before. you could also signup and accept invitation

anikdhabal:
> > For the signup case, we need a pre-populated email, which we get from the email invitation.
> 
> It used to work fine before. you could also signup and accept invitation
Ooo, it seems there's a bug. I will fix it in a follow-up PR

Issue #13417, opened 2024-01-26
[CAL-3042] Odd behavior on copy and on changing calendar


dpalic:
### Issue Summary

if you change to a specific google calendar
you save
and you reload the page
=> the specific calendar is gone and the default calendar of the account is selected

This behavior is occuring on Cloning a eventtype too and creating a mess!
Since you do not think about that if you clone some fields are changed to something else

the bad thing: this was working in the past!

### Steps to Reproduce

default case:
1. ensure to have a google calendar from another account on which you have got full permissions
2. on an event type, select this specific calender
3. save it
4.  go to a different event type
5. reload the prior event type
6. you see that your account calender is now selected and the specific is not anymore

this same behavior occurs if you do a clone of a event type

### Actual Results

- see above

### Expected Results

- the user selected calendar will never be changed automatically
- on clone the identical calendar is selected also for the new cloned

### Technical details

- Browser independent

### Evidence

- via UI

<sub>[CAL-3042](https://linear.app/calcom/issue/CAL-3042/odd-behavior-on-copy-and-on-changing-calendar)</sub>

milospuac:
reported here #13419

PR #13438, opened 2024-01-28
fix: Ensure event type duplication respects private link setting

### LINKED ISSUE #13389: [CAL-3001] Duplicating an event type doesn't duplicate advanced settings
not all fields are properly duplicated when selecting "Duplicate Event Type". Ones that I've seen that are not duplicated:

1. The calendar the event is tied to (always goes back to default).
2. Whether or not private links are enabled (always defaults to off).

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3001](https://linear.app/calcom/issue/CAL-3001/duplicating-an-event-type-doesnt-duplicate-advanced-settings)</sub>

ibex088:
## What does this PR do?

Fixes #13389

* This PR takes care of fixing private links duplication when duplicating event types. 
* the calendar issue doesn't seem to be connected to event type duplication. Even if you pick a different calendar and save it, when you come back later, it defaults back to the default calendar
A separate issue #13417 has alredy been created for that
The calendar the event is tied to (always goes back to default).
* [UPDATE]: the calendar issue (always going back to default), has been fixed in #13440


packages/trpc/server/routers/viewer/eventTypes/duplicate.handler.ts
```diff
@@ -15,6 +17,13 @@ type DuplicateOptions = {
   input: TDuplicateInputSchema;
 };
 
+export const generateHashedLink = (id: number) => {
```
supalarry, line 20: If we search `const generateHashedLink` it is duplicated across 3 files

Screenshot 2024-02-06 at 11 37 13

It would be good to re-use `generateHashedLink` - the copies exist across "apps/web", "packages/features" and "packages/trpc" and all of them import from "@calcom/lib" package ("packages/lib"), so I think it would be good to move `generateHashedLink` there and then re-use across these three packages (it is used in 3 files in total)

supalarry (changes requested):
Amazing work as always Somay! 🚀

I can confirm that your fix works - duplicated event types have the secret URL setup. 

However, small comment below about code re-usage. Let me know what you think.

packages/trpc/server/routers/viewer/eventTypes/duplicate.handler.ts
SomayChauhan: absolutely

SomayChauhan:
just fyi, 
if we try to duplicate the collective-team-event, with private links enabled it is giving error on duplicate

check out this [loom](https://www.loom.com/share/03a9247ccc814c929723abf0b07de66d?sid=2418bc37-88ac-48e6-9a84-97324a96739c)

supalarry (approved):
Tested and works! Great stuff! 🚀

PR #13440, opened 2024-01-28
fix: destination calendar selection resetting to default on page reload

### LINKED ISSUE #13389: [CAL-3001] Duplicating an event type doesn't duplicate advanced settings

### LINKED ISSUE #13417: [CAL-3042] Odd behavior on copy and on changing calendar

ibex088:
## What does this PR do?

Fixes #13417 

also Fixes #13389 1st point

## Evidence

https://www.loom.com/share/89afbe7aa0b8432fb6ceff10d35fc5a8


Udit-takkar (approved):
LGTM

PR #13444, opened 2024-01-29
chore: create platform libraries package

ThyMinimalDev:
## What does this PR do?

Enable platform to use functions from @calcom/lib

## Type of change


supalarry (approved):
Tested it and it is finally possible to import stuff from lib into v2 API and works during runtime too 🎉 Super important PR for the platform team to move forward, so kudos for resolving this!

PR #13449, opened 2024-01-29
feat: v2 api event type endpoints

supalarry:
Adds an endpoint to fetch user event type in preparation for the availability atom.
Also, create default event types when creating managed user.


apps/api/v2/src/ee/event-types/event-types.repository.ts
```diff
@@ -0,0 +1,34 @@
...
+    try {
+      const isUserOrganizationAdmin = user?.organizationId
+        ? await this.membershipsRepository.isUserOrganizationAdmin(user.id, user.organizationId)
+        : false;
+
+      const eventType = await getEventTypeById({
+        eventTypeId,
+        userId: user.id,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
supalarry, line 25: I tried refactoring "getEventTypeById" for it to accept prisma client without extensions (the PrismaClient it now accepts is one from @calcom/prisma") but extended clients do not have $on and $use methods, and extended prisma has accelerate which changes the type even more.

since the prisma operations in getEventTypebyId are simple i think its fine to ignore the type problem here

supalarry (commented):
Self review

apps/api/v2/src/ee/event-types/event-types.repository.ts
ThyMinimalDev: `
import type {PrismaClient} from '@calcom/prisma'


this.dbRead.prisma as unknown as PrismaCLient`

apps/api/v2/src/ee/event-types/event-types.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+
+@Injectable()
+export class EventTypesRepository {
+  constructor(private readonly dbRead: PrismaReadService, private readonly dbWrite: PrismaReadService) {}
+
+  async createUserEventType(userId: number, body: CreateEventTypeInput) {
+    return this.dbWrite.prisma.eventType.create({
+      data: {
+        ...body,
+        userId,
```
exception, line 15: is userId not set by the following connect call?

apps/api/v2/src/ee/event-types/services/event-types.service.ts
```diff
@@ -0,0 +1,46 @@
...
+    const defaultEventTypes = await Promise.all([
+      this.eventTypesRepository.createUserEventType(userId, thirtyMinutes),
+      this.eventTypesRepository.createUserEventType(userId, sixtyMinutes),
+    ]);
+
+    return defaultEventTypes;
+  }
+}
+
+export const DEFAULT_EVENT_TYPES = {
```
exception, line 43: I'd put this in a constants file or something that's more easily searchable.

exception (approved):
Nice work!

apps/api/v2/src/ee/event-types/event-types.repository.ts
supalarry: `EventType.userId` is used to reference `EventType.owner` (one to one) and connect is for `EventType.users` (one to many)

apps/api/v2/src/ee/event-types/services/event-types.service.ts
supalarry: Good call - will do so.

PR #13469, opened 2024-01-30
fix: initialized flags on event-types page

### LINKED ISSUE #13468: unable to create a managed event type

### Issue Summary

unable to create a managed event type even though I've enabled managed event types for my instance and I'm an admin of the teams.

### Technical details

- The issue I discovered was that the useFlagMap function was initially returning an empty object on the event-types page, which should not occur. Flags should be consistently available to all pages at all times.

### Evidence

https://www.loom.com/share/5d8eff2d5ac949dba3d00176a210b7aa?sid=64eb2ea9-bc3b-4247-84d3-c6e8152121d4

ibex088:
## What does this PR do?

Fixes #13468 

added it for event-types page only, fixes the isses in hand

Future consideration: Potentially enabling flags to be accessible on all pages consistently.


PeerRich:
can we also remove the feature flag for managed event types? its not needed anymore

SomayChauhan:
removed

packages/prisma/migrations/20240131021824_revert_20230404202721_add_feature_flag_managed_event_types/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- revert 20230404202721_add_feature_flag_managed_event_types 
+
```
keithwillcode, line 2: Why are we deleting this feature flag?

keithwillcode (dismissed):

packages/prisma/migrations/20240131021824_revert_20230404202721_add_feature_flag_managed_event_types/migration.sql
PeerRich: yes, i am ok removing this. this was used to roll it out slowly but now managed event types is a core part of our offering and doenst need a feature flag imo

keithwillcode (approved):

apps/web/lib/event-types/getServerSideProps.tsx
```diff
@@ -0,0 +1,23 @@
...
+        redirect: {
+          permanent: false,
+          destination: "/auth/login",
+        },
+      },
+    };
+  }
+
+  return { props: { trpcState: ssr.dehydrate() } };
+};
```
PeerRich, line 23: do we even need this if we remove the feature flag for managed event types?

SomayChauhan: i think we should still include this.

PeerRich: ok.

btw. end to end tests failing, not sure if its related to this PR @keithwillcode ?

keithwillcode: It’s not unfortunately. We have E2E test suite issues at the moment. Hoping to resolve by tomorrow and will get this merged.

PR #13484, opened 2024-01-31
fix: Invalid key length error

### LINKED ISSUE #13290: [CAL-3060] JSON Error during login  due to incorrect encoding on CALENDSO_ENCRYPTION_KEY
### Issue Summary

After a fresh install, I go to login and after about 20 seconds, i see this error pop up in the login screen:

image

Note related issues: #9527 #9690
Original issue: [https://github.com/calcom/docker/issues/321](https://github.com/calcom/docker/issues/321)

### Steps to Reproduce

1. Setup repository locally and generate template .env file as per the [instructions](!%5Bimage%5D(https://github.com/calcom/cal.com/assets/3743662/340d8c50-2485-4da1-996b-07e8c36649ec))
2. Update the .env file to use a secret generated by this command as described in the [comments](https://github.com/calcom/cal.com/blob/e99e8b169afe72e188a56f8ecb7c7e82335a1043/.env.example#L92) for the CALENDSO_ENCRYPTION_KEY env variable

```
openssl rand -base64 32
```

1. spin up docker containers via `docker compose up -d`
2. Login
3. Be greeted by this banner 'You are admin but you do not have a password length of at least 15 characters or no 2FA yet Change Password to gain admin access'
4. Attempt to setup twofactor auth and receive error messages shown in images below

### Actual Results

image

### Expected Results

Not have issues with the encryption key.

### Technical details

*Environment:*

Hosting service: GCP e2-small
OS: Debian Bullseye
Docker Images: see [docker compose](https://github.com/calcom/docker/blob/main/docker-compose.yaml) file

*Config:*

I kept the defaults from here, however ive updated the NEXT_PUBLIC_WEBAPP_URL to be my own domain.

### Evidence

This was tested using the docker containers outlined above, both locally and in the cloud to limit any additional sources of error.

Here are my findings in the order in which i found them:

**update 1**

I do see this error in the browser console:

```
Error setting up two-factor authentication SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data two-factor-auth-5a03d5ad2d4dbd78.js:1:5962
handleSetup NextJS
```

**update 2**

It looks like the fetch call to '...api/auth/two-factor/totp/setup' is returning an internal server error, which then sends back the JSON error when the response.json() function is called.

**update 3**

Default config is outdated. The newer config file makes sure to specify the length of the encryption key:

```
# Application Key for symmetric encryption and decryption
# must be 32 bytes for AES256 encryption algorithm
# You can use: `openssl rand -base64 32` to generate one
CALENDSO_ENCRYPTION_KEY=
```

**update 4**

So even changing the key to a 32 byte key using the algo mentioned above, the error still persists:

```
: RangeError: Invalid key length
@calcom/web:start:     at Cipheriv.createCipherBase (node:internal/crypto/cipher:122:19)
@calcom/web:start:     at Cipheriv.createCipherWithIV (node:internal/crypto/cipher:141:3)
@calcom/web:start:     at new Cipheriv (node:internal/crypto/cipher:249:3)
@calcom/web:start:     at Object.createCipheriv (node:crypto:141:10)
@calcom/web:start:     at symmetricEncrypt (/calcom/apps/web/.next/server/chunks/69559.js:1:278)
@calcom/web:start:     at handler (/calcom/apps/web/.next/server/pages/api/auth/two-factor/totp/setup.js:1:2125) {
@calcom/web:start:   code: 'ERR_CRYPTO_INVALID_KEYLEN'
```

**update 5**

Doing some local testing with the touched code, it looks like the base64 encoding of the string generated through openssl rand -base64 32 command, is failing when converting into a buffer which sets the encoding to latin1, resulting in a 44 length string, not 32. This is what causes the error.

**update 6**

Temporary solution would be to just create a 32 character long alphanumeric string without the openssl function, and updating the docs/comments to reflect the importance of this, seeing as the Buffer.from function expects the string to be binary/latin1 encoded

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3060](https://linear.app/calcom/issue/CAL-3060/json-error-during-login-due-to-incorrect-encoding-on-calendso)</sub>

abhijeetsingh-22:
## What does this PR do?

Fixes invalid key length error that occurred during symmetric encoding and decoding of the token in the `crypto`
Fixes #13290 

https://github.com/calcom/cal.com/assets/55000107/aad52d73-5a2a-4f3a-a5b6-3b3ab221fb1f

## Type of change

## How should this be tested?

- The symmetric encoding and decoding using the encryption key should work without any errors. 

## Mandatory Tasks

## Checklist
All done

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -296,8 +296,8 @@ CALCOM_WEBHOOK_SECRET=""
 CALCOM_WEBHOOK_HEADER_NAME="calcom-webhook-secret"
 CALCOM_CREDENTIAL_SYNC_ENDPOINT=""
 # Key should match on Cal.com and your application
-# must be 24 bytes for AES256 encryption algorithm
-# You can use: `openssl rand -base64 24` to generate one
+# must be 32 bytes for AES256 encryption algorithm
+# You can use: `openssl rand -base64 32` to generate one
 CALCOM_APP_CREDENTIAL_ENCRYPTION_KEY=""
 
 # - OIDC E2E TEST *******************************************************************************************
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,7 +3,7 @@ tasks:
       yarn &&
       cp .env.example .env &&
       next_auth_secret=$(openssl rand -base64 32) &&
-      calendso_encryption_key=$(openssl rand -base64 24) &&
+      calendso_encryption_key=$(openssl rand -base64 32) &&
       sed -i -e "s|^NEXTAUTH_SECRET=.*|NEXTAUTH_SECRET=$next_auth_secret|" \
              -e "s|^CALENDSO_ENCRYPTION_KEY=.*|CALENDSO_ENCRYPTION_KEY=$calendso_encryption_key|" \
              -e "s|http://localhost:3000|https://localhost:3000|" \
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,7 @@ To develop locally:
 
    - Duplicate `.env.example` to `.env`.
    - Use `openssl rand -base64 32` to generate a key and add it under `NEXTAUTH_SECRET` in the `.env` file.
-   - Use `openssl rand -base64 24` to generate a key and add it under `CALENDSO_ENCRYPTION_KEY` in the `.env` file.
+   - Use `openssl rand -base64 32` to generate a key and add it under `CALENDSO_ENCRYPTION_KEY` in the `.env` file.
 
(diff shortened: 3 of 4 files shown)


keithwillcode:
Related to and potentially fixed by #12698

abhijeetsingh-22:
@keithwillcode, thanks for pointing it out . You are right #12698 will fix this issue and it also has the migration logic which I was about to implement. 
So closing this in favour of #12698

PR #13485, opened 2024-01-31
feat: Create default team events on a new team creation

### LINKED ISSUE #13300: [CAL-2960] default event-types after creating a team
right now, a team starts without event-types:

CleanShot 2024-01-18 at 13 08 09@2x

we should make some generic event-types as a starter:

* Collective
* Round Robin
* 30min (managed event type?)

we could also ask a few questions during team creation and have templates

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2960](https://linear.app/calcom/issue/CAL-2960/default-event-types-after-creating-a-team)</sub>

kart1ka:
## What does this PR do?

Creates Collective, Round Robin and Managed Team Event on new team creation.

Fixes #13300 

## Type of change

## How should this be tested?

Create a new team and observe that default team event types (Collective, Round Robin, Managed Event) are also created as a template along with it.

## Mandatory Tasks

### DIFF
--- a/packages/lib/createEventType.ts
+++ b/packages/lib/createEventType.ts
@@ -1,0 +1,108 @@
+import type { Prisma } from "@prisma/client";
+import { PrismaClientKnownRequestError } from "@prisma/client/runtime/library";
+import type { z } from "zod";
+
+import getAppKeysFromSlug from "@calcom/app-store/_utils/getAppKeysFromSlug";
+import { DailyLocationType } from "@calcom/app-store/locations";
+import getApps from "@calcom/app-store/utils";
+import { getUsersCredentials } from "@calcom/lib/server/getUsersCredentials";
+import type { PrismaClient } from "@calcom/prisma";
+import { SchedulingType } from "@calcom/prisma/enums";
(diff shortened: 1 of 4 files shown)


packages/lib/createEventType.ts
joeauyeung, line 1: Not sure how I feel about abstracting this since it's only used in the create event type endpoint. If we do decide to keep this it should live under `packages/features/eventtypes/lib`

joeauyeung (changes requested):
Thanks for your contribution @kart1ka.

For this PR to be merged I think we should show a step in the team creation form asking the user which event types they would like to be created.

https://github.com/calcom/cal.com/issues/13300#issuecomment-1901790887

kart1ka:
Hi @joeauyeung 
Should it be the same kind of form that is mentioned here https://github.com/calcom/cal.com/issues/13300#issuecomment-1901790887?

If yes, then that would basically mean asking the user to create a team event of his/her choice, which is different from the description of the issue https://github.com/calcom/cal.com/issues/13300#issue-2088256252.
The description, as per my understanding, is about creating all 3 team events as a starter template on team creation.

PR #13486, opened 2024-01-31
feat: v2 api me PUT endpoint

supalarry:
1. Return timeZone for /me
2. Add PATCH for /me with in depth validation for timeFormat, weekStart and timeZone


apps/api/v2/src/modules/users/users.repository.ts
```diff
@@ -83,3 +97,17 @@ export class UsersRepository {
...
+  const capitalizedSegments = segments.map((segment) => {
+    return capitalize(segment);
+  });
+
+  return capitalizedSegments.join("/");
+}
+
+function capitalize(str: string) {
+  return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
+}
```
Ryukemeister, line 113: Can we maybe create a single function `capitalize` and this would take in two params `userInput` and `userInputType` (either timeZone or normal string) and then based on our userInputType we return the modified `userInput` ? Just throwing it out there although it's most probably an overkill

apps/api/v2/src/modules/users/inputs/create-user.input.ts
```diff
@@ -1,6 +1,28 @@
...
   email!: string;
+
+  @IsNumber()
+  @IsOptional()
+  @Validate(IsTimeFormat)
+  timeFormat?: number;
+
+  @IsNumber()
+  @IsOptional()
+  defaultScheduleId?: number;
```
Ryukemeister, line 17: We need a default schedule id at the time of creating a user in order for our default schedule endpoint to work so I though it's better if `defaultScheduleId` isn't optional for the create user input

apps/api/v2/src/modules/users/inputs/validators/is-week-start.ts
```diff
@@ -0,0 +1,23 @@
...
+    return (
+      lowerCaseWeekStart === "monday" ||
+      lowerCaseWeekStart === "tuesday" ||
+      lowerCaseWeekStart === "wednesday" ||
+      lowerCaseWeekStart === "thursday" ||
+      lowerCaseWeekStart === "friday" ||
+      lowerCaseWeekStart === "saturday" ||
+      lowerCaseWeekStart === "sunday"
+    );
+  }
```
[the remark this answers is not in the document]
exception: Array includes is slower than direct compare, at most you should use a `Set(...)`

exception (changes requested):

apps/api/v2/src/modules/oauth-clients/controllers/oauth-client-users/zod/response/response.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from "zod";
+
+export const userSchemaResponse = z.object({
+  id: z.number().int(),
+  email: z.string(),
+  timeFormat: z.number().int().default(12),
+  defaultScheduleId: z.number().int().nullable(),
+  weekStart: z.string(),
+  timeZone: z.string(),
```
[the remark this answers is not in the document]
supalarry: Good call - implemented in [refactor: defaultTimezone Europe/London in response schema](https://github.com/calcom/cal.com/pull/13486/commits/6a681b006e0a3f467237816922fbbad094e31270)

apps/api/v2/src/modules/users/inputs/validators/is-time-zone.ts
```diff
@@ -0,0 +1,18 @@
+import { ValidatorConstraint, ValidatorConstraintInterface } from "class-validator";
+
+import { TIMEZONES } from "@calcom/platform-constants";
+
+@ValidatorConstraint({ name: "isTimeZone", async: false })
+export class IsTimeZone implements ValidatorConstraintInterface {
+  validate(timeZone: string) {
+    if (!timeZone) return true;
```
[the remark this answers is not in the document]
supalarry: Good catch - implemented in [refactor: validators](https://github.com/calcom/cal.com/pull/13486/commits/5951866556ddc4aaba2788b17852d943d85799c4)

apps/api/v2/src/modules/users/users.repository.ts
supalarry: I think its fine as is

apps/api/v2/src/modules/users/inputs/create-user.input.ts
supalarry: The thing is that when creating a schedule `userId` is needed if `schedule.availabilities` are passed in the request body - see function calls within `createSchedule` of `apps/api/v2/src/ee/schedules/controllers/schedules.controller.ts`, so there might be a situation where we can't make a schedule without having a user created before.

So we can either:
1. Add new optional property `isDefault` that is "true" by default when creating a schedule, and without default value when updating it.
2. Create a new schedules endpoint POST /default that accepts request body { scheduleId: number } and sets "defaultScheduleId" of user to it.
3. Don't do anything and tell our users to create user, create schedule, and then make call to update user with the defaultScheduleId of the schedule.

I went with option 1 for now in [refactor: default schedule in schedules endpoints](https://github.com/calcom/cal.com/pull/13486/commits/eb0dfc4fa5bb222e2db70f311e843de5d4e46c56)

supalarry: So now after creating a user schedule it will be sure that the user has `defaultScheduleId` set.

apps/api/v2/src/ee/me/me.controller.ts
```diff
@@ -10,23 +16,35 @@ import { ApiResponse } from "@calcom/platform-types";
...
-        defaultScheduleId: user.defaultScheduleId,
-        weekStart: user.weekStart,
+        user: me,
       },
     };
   }
-}
 
-export type UserReturned = Pick<User, "id" | "email" | "timeFormat" | "defaultScheduleId" | "weekStart">;
+  @Put("/")
```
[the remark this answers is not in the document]
supalarry: Good catch - refactored me, schedules, oauth clients and oauth client users endpoints to use patch instead of put in [refactor: patch instead of put](https://github.com/calcom/cal.com/pull/13486/commits/ca582119974a21e27ed647396ed7b0b3e5e11144)

exception (approved):

Issue #13358, opened 2024-01-22
[CAL-2994] Fix round robin per issues


keithwillcode:
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2994](https://linear.app/calcom/issue/CAL-2994/fix-round-robin-per-issues)</sub>

Udit-takkar:
@keithwillcode and I keep seeing this warning message in terminal whenever I go to booking page.  


<img width="1253" alt="Screenshot 2024-01-22 at 5 22 38 PM" src="https://github.com/calcom/cal.com/assets/53316345/085a5118-3cc4-4285-a330-948cec2c296e">

PR #13495, opened 2024-02-01
perf: Don't process date overrides outside of requested boundaries

keithwillcode:
## What does this PR do?

Fixes #13358.

Date overrides on user's availability settings were being added to the list of slots and their calculations even when the overrides fell outside of the start/end window of the requested schedule.

## Type of change

## How should this be tested?

- Ensure test suite passes and date overrides outside of getSchedule boundaries are not included in the list of available slots

## Mandatory Tasks


sean-brydon (approved):
Tested locally - can confirm we now don't fetch these date overrides outside of the range

After:
CleanShot 2024-02-07 at 11 12 01@2x

PR #13502, opened 2024-02-02
test: booking limits flakiness

sean-brydon:
This PR removes the flakiness from booking limits tests - for some reason adding 1 month had no availability. 

If we want to test this with the current browser date we should look at ways to mock this in the future for example `sinon` provides a good way to mock browser dates in e2e tests instead of doing this logic inside of e2e tests which isnt ideal.

### DIFF
--- a/apps/web/playwright/booking-limits.e2e.ts
+++ b/apps/web/playwright/booking-limits.e2e.ts
@@ -37,15 +37,12 @@ const BOOKING_LIMITS_MULTIPLE = {
   PER_YEAR: 4,
 };
 
-// prevent tests from crossing year boundaries - if currently in Oct or later, start booking in Jan instead of Nov
-// (we increment months twice when checking multiple limits)
-const firstDayInBookingMonth =
-  dayjs().month() >= 9 ? dayjs().add(1, "year").month(0).date(1) : dayjs().add(1, "month").date(1);
+// If we wanted to use browser date we should look into using sinon or another browser date mocking lib.
+// Mock the current day - we dont need to do this based on the browsers date - this is how we get flaky tests
+const mockedDate = dayjs().add(2, "year").month(6);
 
 // avoid weekly edge cases
-const firstMondayInBookingMonth = firstDayInBookingMonth.day(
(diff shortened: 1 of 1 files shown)


apps/web/playwright/booking-limits.e2e.ts
```diff
@@ -37,15 +37,12 @@ const BOOKING_LIMITS_MULTIPLE = {
   PER_YEAR: 4,
 };
 
-// prevent tests from crossing year boundaries - if currently in Oct or later, start booking in Jan instead of Nov
-// (we increment months twice when checking multiple limits)
-const firstDayInBookingMonth =
-  dayjs().month() >= 9 ? dayjs().add(1, "year").month(0).date(1) : dayjs().add(1, "month").date(1);
+// If we wanted to use browser date we should look into using sinon or another browser date mocking lib.
+// Mock the current day - we dont need to do this based on the browsers date - this is how we get flaky tests
+const mockedDate = dayjs().add(2, "year").month(6);
```
sean-brydon, line 42: Random date in the future - we have no way of going "out of bounds"

apps/web/playwright/booking-limits.e2e.ts
```diff
@@ -40,7 +40,7 @@ const BOOKING_LIMITS_MULTIPLE = {
 // prevent tests from crossing year boundaries - if currently in Oct or later, start booking in Jan instead of Nov
 // (we increment months twice when checking multiple limits)
 const firstDayInBookingMonth =
-  dayjs().month() >= 9 ? dayjs().add(1, "year").month(0).date(1) : dayjs().add(1, "month").date(1);
+  dayjs().month() >= 8 ? dayjs().add(2, "year").month(0).date(1) : dayjs().add(2, "month").date(1);
```
sean-brydon, line 43: For some reason this was always returning 2023 for me 1 when I changed to +2 it was returning the correct date time we expected

sean-brydon:
Closing in favour of skipping these tests for now - ideally we re-write this suite to take into account feb having 29 days at times.

PR #13522, opened 2024-02-03
fix: 'Invalid event length' error for 1 hour dynamic events

### LINKED ISSUE #13507: “Invalid event length” when creating dynamic events with a 1 hour duration

### Issue Summary

### Steps to Reproduce

1. Go to a dynamic event URL (cal.com/username1+username2)
2. Pick “1 hour” as the duration
3. Try to create the event

### Actual Results

image
image

### Expected Results

The event is created.

### Technical details

Happens with any browser, for all our team members, independent whether they have 1 hour meeting types set up individually.

### Evidence

(See above)

Nis-Han:
## What does this PR do?
- Fixes #13507 

[Loom video](https://www.loom.com/share/6a90b2cf68e240458183bfdbaec2f3e7?sid=40cbd41f-c528-42b8-84a9-37600a7c33a2)

## Requirement/Documentation
- By default, dynamic events have a [set length of 30](https://github.com/calcom/cal.com/blob/4aa040d96edc2e24c48c86fcc70f6f842186df87/packages/lib/defaultEvents.ts#L107C3-L107C14), which is causing the given issue.
- The `handleNewBooking` method tries to [generate a list of `validEventLengths`  from `eventType.metadata` ](https://github.com/calcom/cal.com/blob/4aa040d96edc2e24c48c86fcc70f6f842186df87/packages/features/bookings/lib/handleNewBooking.ts#L1096) if it is present, else takes up the default value.
- Passing `multipleDuration` attribute while generating the metadata for dynamic events seemed to solve the issue at first. (But this was a breaking change, it was enforcing those default duration on every dynamic meeting on the UI as well).
- Changing the `dynamicEvent.length` to array type, with same allowed lengths as in [`Duration.tsx`](https://github.com/calcom/cal.com/blob/af1382c5ba07af91c168a271d985c6db67c036c4/packages/features/bookings/components/event-meta/Duration.tsx#L56) also seemed to solve the issue, but it failed E2E tests.
- So, finally I've tried injecting the same values as [`Duration.tsx`](https://github.com/calcom/cal.com/blob/af1382c5ba07af91c168a271d985c6db67c036c4/packages/features/bookings/components/event-meta/Duration.tsx#L56) in `eventType.metadata.multipleDuration` only when `eventType.slug ==='dynamic'` and `!eventType.metadata.multipleDuration.length` are followed

## Question for maintainers
- Hardcoding might be an issue here, I want to avoid breaking anything, that's why I've added the lengths separately. Would it help, if `Duration.tsx` and `defaultEvents.ts` pick valid event lengths from some constant file? 

## Type of change

## How should this be tested?

- Git pull and run as mentioned in the readme.
- Book a 60 dynamic event with 2 or more users.

## Mandatory Tasks

## Checklist

- I haven't commented my code, particularly in hard-to-understand areas (Its just a single line change)


Nis-Han:
@keithwillcode Multile E2E tests are failing, with errors that seems to be related to translation. Is this normal? Or are my changes breaking something?

PS: This is my first OS PR. I apologize for tagging you if this is something trivial that I am not able to understand.

Udit-takkar (commented):
Screenshot 2024-02-05 at 12 45 06 PM

I am getting this error. 
http://localhost:3000/pro+free/dynamic
'Pro' -> timezone is Asia/calcutta
'Free' ->  timezone is Europe/London

The slots are also different

https://github.com/calcom/cal.com/assets/53316345/5fa0c96e-fd9b-4e5b-b05a-bddde51b1134

Nis-Han:
> I am getting this error. http://localhost:3000/pro+free/dynamic 'Pro' -> timezone is Asia/calcutta 'Free' -> timezone is Europe/London
> 
> The slots are also different
Hi @Udit-takkar, this issue seems to be unrelated the changes in this PR and is present in prod as well. (Only if the hosts havn't applied some custom settings).
[Screencast from 2024-02-05 15-05-07.webm](https://github.com/calcom/cal.com/assets/56070571/5647e1e9-755e-42bb-be75-abb5a7447b99)

I also tried the same with main branch on my local:
[Screencast from 2024-02-05 15-06-36.webm](https://github.com/calcom/cal.com/assets/56070571/57de9c89-b568-4785-a103-a89958ada705)

I will try investigating and fixing this issue. Should this be commited along with this PR, or does this require a new issue creation?

Nis-Han:
Somehow, the time slots for 30 min meetings are rendered correctly, but not for other valid event lengths. 
That is why, even if the original error(that this PR aims to solve) doesn't happen, it will still cause problem if someone tries to book by clicking on a slot that shouldn't have been rendered in the first place.
This seems to be a problem with the API/component that GETs available slot, not the API that POSTs a request for booking.

CarinaWolli:
@Nis-Han What about setting the duration already in the default dynamic event-type? Here: https://github.com/calcom/cal.com/blob/af1382c5ba07af91c168a271d985c6db67c036c4/packages/lib/defaultEvents.ts#L101

Nis-Han:
> @Nis-Han What about setting the duration already in the default dynamic event-type? Here:

@CarinaWolli I did the same thing initially and yes that fixes the problem.
But since the object is named `common`, that means it is intended to be used for other events in future as well (I might be wrong). So, this felt to be a better solution.

If we want to avoid changing the `handleNewBooking` file, another solution might be overriding the `multipleDuration` property while constructing the `dynamicEvent` object itself

Something like this
Screenshot from 2024-02-05 22-31-59

CarinaWolli:
> If we want to avoid changing the `handleNewBooking` file, another solution might be overriding the `multipleDuration` property while constructing the `dynamicEvent` object itself

I think this would be the best solution. If we ever want to add a duration to dynamic events we only have to add it in one place

```
const dynamicEvent = {
  length: 30,
  slug: "dynamic",
  title: "Group Meeting",
  eventName: "Group Meeting",
  description: "Join us for a meeting with multiple people",
  descriptionAsSafeHTML: "",
  position: 0,
  ...commons,
  metadata: EventTypeMetaDataSchema.parse({[15, 30, 60, 90]}),
};
```

Nis-Han:
> I think this would be the best solution. If we ever want to add a duration to dynamic events we only have to add it in one place

Updated the code accordingly

CarinaWolli (approved):
Thank you 🙏

PR #13526, opened 2024-02-04
fix: Issue with "Limit total booking duration" and offering seats

### LINKED ISSUE #13509: Issue with "Limit total booking duration" and offering seats.
### Issue Summary

Remaining available seats do not display when "Limit total booking duration" condition is met.  Taking a seat in existing meeting does not increase durations so booking should remain avalaible in existing booking with remaing seats. 

### Issue description

The goal is to organize on demand fitness class. for 4 attendees A first person creates the initial booking and other people can then join in. I would like to offer maximum two classes of 15 minutes per day.  The schedule of the class varies depending on my availablilities and when the first person has made the booking for each of the 2 slots. 

In order to ensure that I dont do more than 2 classes per day. I decide to limit 'total booking duration' to 30 minutes. 

We would like the following behavior. 

- step1:  Person A creates an initial booking at slot1.  Result is slot 1 taken and 3 seats left. 
- step2: Person B creates joins the slot 1. Result : 2 seats left in slot 1
- step3: Person C creates a new booking at slot 2.  3 seats left in slot 2.
- step4: Now new people can join in slot 1 or slot 2, depending on remaining availabilities.  

Unfortunately step 4 is not possible because the displayed availabilities do not take into account remaining seats.  Once the "Limit Total booking duration" of the day is met, then it will not be possible to join existing slots with remaining seats.  The booking interface will not display any avalaible slot 1. 

This is currently possible: 

Scenario 1: 
- A creates slot 1
- B joins slot 1 
- C joins slot 1
- D joins slot 1
- E creates slot 2
- Nothing else possible. 
 
Scenario 2: 
- A creates slot 1
- B creates slot 2
- Nothing else possible

MehulZR:
## What does this PR do?

Fixes #13509 (issue)

https://github.com/calcom/cal.com/assets/86384652/0dd9920e-68ec-43cf-a942-dd474425a974

https://github.com/calcom/cal.com/assets/86384652/9a1cc3ef-eba7-4958-bbb7-be03a46f80ab

## Type of change

## Mandatory Tasks


MehulZR:
Approaches

This [commit](https://github.com/calcom/cal.com/pull/13526/commits/028d2ce55f3e36591b9247722490e6effce3c0f3) is according to 2nd approach. As I looked into limits more, I think it might be better to go with the 1st approach. 

What do you guys think? As for now, further commits will be in according to the 1st approach.

MehulZR:
@keithwillcode Sorry, I have been busy past couple days. This PR contains some bugs which I have resolved locally, just have to test them thoroughly.

packages/core/getBusyTimes.ts
```diff
@@ -144,7 +144,7 @@ export async function getBusyTimes(params: {
   const bookingSeatCountMap: { [x: string]: number } = {};
   const busyTimes = bookings.reduce(
     (aggregate: EventBusyDetails[], { id, startTime, endTime, eventType, title, ...rest }) => {
-      if (rest._count?.seatsReferences) {
+      if (eventType?.seatsPerTimeSlot) {
```
MehulZR, line 147: In unit tests the prisma mock doesn't return `seatsReferences` which ends up failing the tests.

MehulZR:
image

So for a long time, these particular tests were always failing for me (main branch with no changes). At that time I was not able to figure out why at a glance. But tests were passing on the github actions. I have checked this on multiple linux distros. I think it's probably due to timezone issues?

As I was working on the issue and fixing tests for *limits with seats*, I made the required changes.

https://github.com/calcom/cal.com/blob/01b5c1f11eba9ced597a2b5e33810bcd9f7b9b44/packages/core/getUserAvailability.ts#L334-L346

We were trying to subtract `formattedBusyTimes` (which are in GMT) from `dateRanges` (which can be in diff tz). Leading to some issue.

So after quick fix. I think the currently failing test have become more consistent.

MehulZR:
https://github.com/calcom/cal.com/assets/86384652/87e23780-d943-432d-b6a5-34fdf3a675ac

It looks like it is related to timezone. I will look into this bit more.

MehulZR:
Question

Can anyone confirm what should be the case when both before event buffer and after event buffer are applied on an event.

https://github.com/calcom/cal.com/blob/0a795f5363b82fd5a5b6ffa1a4fa31a035a3f44f/apps/web/test/lib/getSchedule.test.ts#L584-L596

The test suggests 

- 4-6 am is our google calendar busy time. 
- 6-8 am is before event buffer for the 8 am slot.
- 8-10 am available slot

Isn't this incorrect? I used to think that it would mean something like 

- 4-6 am google calendar busy time.
- 6-8 am after event buffer for 4 am slot.
- 8-10 am before event buffer for 10 am slot.
- 10-12 pm available slot.

keithwillcode:
Not sure if related but just found important time zone fix in #13738

MehulZR:
> Question
> 
> Can anyone confirm what should be the case when both before event buffer and after event buffer are applied on an event.
> 
> https://github.com/calcom/cal.com/blob/0a795f5363b82fd5a5b6ffa1a4fa31a035a3f44f/apps/web/test/lib/getSchedule.test.ts#L584-L596
> 
> The test suggests
> 
>     * 4-6 am is our google calendar busy time.
> 
>     * 6-8 am is before event buffer for the 8 am slot.
> 
>     * 8-10 am available slot
> 
> 
> Isn't this incorrect? I used to think that it would mean something like
> 
>     * 4-6 am google calendar busy time.
> 
>     * 6-8 am after event buffer for 4 am slot.
> 
>     * 8-10 am before event buffer for 10 am slot.
> 
>     * 10-12 pm available slot.
> 
> 
> After looking into it bit more I found that for Non-Cal busy times `Case B` is happening & for Cal busy times `Case A` is happening. (All this is being done in the unit tests. Unless the calendarManagerMock is faulty, It should be the same in the app)

Okay, So I think the bookings which are not made throught cal don't have event buffers applied. 
In the test it's just that the `8:00AM` slot needs before event buffer. 

Correct me, If I am wrong.

MehulZR:
> Not sure if related but just found important time zone fix in #13738

Probably not, It doesn't change anything below the getUserAvailability layer.

apps/web/test/lib/getSchedule.test.ts
```diff
@@ -993,6 +995,340 @@ describe("getSchedule", () => {
...
+
+      expect(availableSlotsInTz.filter((slot) => slot.format().startsWith(plus2DateString)).length).toBe(1); // 1 seated booking with remaining slots
+
+      availableSlotsInTz = [];
+      for (const date in thisUserAvailabilityBookingLimitTwo.slots) {
+        thisUserAvailabilityBookingLimitTwo.slots[date].forEach((timeObj) => {
+          availableSlotsInTz.push(dayjs(timeObj.time).tz(Timezones["+6:00"]));
+        });
+      }
+      expect(availableSlotsInTz.filter((slot) => slot.format().startsWith(plus2DateString)).length).toBe(23);
```
MehulZR, line 1163: 23 because there is a booking for `event id 1` on the same day.

apps/web/test/lib/getSchedule.test.ts
```diff
@@ -574,20 +576,20 @@ describe("getSchedule", () => {
...
           timeZone: Timezones["+5:30"],
           isTeamEvent: false,
         },
       });
 
       expect(scheduleForEventOnADayWithNonCalBooking).toHaveTimeSlots(
         [
-          // `04:00:00.000Z`, // - 4 AM is booked
-          // `06:00:00.000Z`, // - 6 AM is not available because 08:00AM slot has a `beforeEventBuffer`
-          `08:00:00.000Z`, // - 8 AM is available because of availability of 06:00 - 07:59
```
MehulZR, line 588: Earlier this was passing (in UTC) because we were subtracting the busyTimes (UTC) from dateRanges (any tz).

https://github.com/calcom/cal.com/blob/0a795f5363b82fd5a5b6ffa1a4fa31a035a3f44f/packages/core/getUserAvailability.ts#L334-L346

So, we were getting somethings like this.

Explanation

And this was being sent to `getSlots` fn. 

While generating the slots, the value of `minute` for slots 6 AM(UTC) & further were being read as `0` because the `rangeStart` was in UTC format. So, the value of `minute` was fitting perfectly with the slot interval.

packages/core/getUserAvailability.ts
```diff
@@ -339,10 +339,11 @@ const _getUserAvailability = async function getUsersWorkingHoursLifeTheUniverseA
   });
 
   const formattedBusyTimes = detailedBusyTimes.map((busy) => ({
-    start: dayjs(busy.start),
-    end: dayjs(busy.end),
+    start: dayjs(busy.start).tz(timeZone),
+    end: dayjs(busy.end).tz(timeZone),
   }));
 
+  //INFO  formattedBusyTimes need to be in the same tz as dateRanges to subtract correctly.
```
MehulZR, line 346: `Dayjs().isBefore()/isAfter()` is not affected by the tz. But while subtracting, it will push the `Dayjs` object (which can be in diff tz) in the returned value(array).

And the fn (ex. getSlots) who is dependant on the returned value can get affected by it & produce incorrect behaivour.

MehulZR:
These PRs are also closely related.
https://github.com/calcom/cal.com/pull/13707/
https://github.com/calcom/cal.com/pull/13651/

packages/core/getUserAvailability.ts
```diff
@@ -495,10 +496,44 @@ const _getBusyTimesFromLimits = async (
...
+            {
+              start: dayjs.utc(booking.end).toISOString(),
+              end: busyTime.end,
+            }
+          );
+          break;
+        }
+      }
+    }
+  }
```
joeauyeung, line 531: **Non-blocking**: Something about a loop inside of a loop didn't sit well with me. What if instead of iterating through `bookings` and then looping through that new array. What if we iterate through bookings, see if it fits the criteria for `bookingWithRemainingSeats`, then splice `busyTimes`. If it doesn't fit the criteria continue to the next booking.

joeauyeung (commented):
@calcom/foundation saw this PR was older and involved seats so I thought I would take a stab at it.

This PR is fixing the issue.

I left a non-blocking comment but I'll wait for the @calcom/foundation to give an approval.

packages/core/getUserAvailability.ts
joeauyeung: I've gone ahead and made this change in this commit b5a43cce44ffc4628b5d947761de182a9b57cc2b

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1627,6 +1628,24 @@ async function handler(
...
+        startAsDate,
+        eventType.id,
+        eventType.schedule?.timeZone
+      );
+    }
+    if (eventType.durationLimits) {
+      await checkDurationLimits(eventType.durationLimits as IntervalLimit, startAsDate, eventType.id);
+    }
+  }
+
```
CarinaWolli, line 1648: Why are we adding this code here again (looks same as https://github.com/calcom/cal.com/pull/13526/files#diff-a7c21727d9a0d4d1aaff9d24fca9293bc28468861bc8022dfa03116908b6fae7R1124)? And aren't we now calling `checkBookingLimit` and `checkDurationLimits` twice for non-seated events?

packages/core/getUserAvailability.ts
```diff
@@ -520,10 +521,43 @@ const _getBusyTimesFromLimits = async (
...
+                start: dayjs.utc(booking.end).toISOString(),
+                end: busyTime.end,
+              }
+            );
+            break;
+          }
+        }
+      }
+    }
+  }
```
CarinaWolli, line 555: can you add some self-review comments here, I can't follow what we are trying to do here

CarinaWolli (changes requested):
Limit total booking duration with seats doesn't work for me. 

I have a 30 min seated event with 60min limit per week. When I create two bookings on different days but same week, then the whole week is blocked even the two slots that should have still seats left

CarinaWolli (changes requested):
Limit total booking duration with seats doesn't work for me. 

I have a 30 min seated event with 60min limit per week. When I create two bookings on different days but same week, then the whole week is blocked even the two slots that should have still seats left

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1627,6 +1628,24 @@ async function handler(
...
+    Object.prototype.hasOwnProperty.call(eventType, "bookingLimits") ||
+    Object.prototype.hasOwnProperty.call(eventType, "durationLimits")
+  ) {
+    const startAsDate = dayjs(reqBody.start).toDate();
+    if (eventType.bookingLimits) {
+      await checkBookingLimits(
+        eventType.bookingLimits as IntervalLimit,
+        startAsDate,
+        eventType.id,
+        eventType.schedule?.timeZone
```
CarinaWolli, line 1641: type error here

CarinaWolli (commented):
Unit tests are failing and there is a type (see comment)

packages/features/bookings/lib/handleNewBooking.ts
MehulZR: Originally I placed the code here for the time being (performing the limits check on seated event was just blocking the logic) and It was and is open for discussion on how to structure this.

A quick fix will be place a condition such as `eventType.seatsPerTimeSlot` but i don't feel good about it. What do you suggest?

```
  if (
    eventType.seatsPerTimeSlot &&
    (Object.prototype.hasOwnProperty.call(eventType, "bookingLimits") ||
    Object.prototype.hasOwnProperty.call(eventType, "durationLimits"))
  ) { }
```

packages/core/getUserAvailability.ts
```diff
@@ -520,10 +521,43 @@ const _getBusyTimesFromLimits = async (
...
+  if (eventType.seatsPerTimeSlot) {
+    for (let i = 0; i < bookings.length; i++) {
+      const booking = bookings[i];
+
+      if (
+        booking.attendeesCount &&
+        booking.attendeesCount < eventType.seatsPerTimeSlot &&
+        dayjs(booking.start).isBetween(dateFrom, dateTo)
+      ) {
+        for (let j = i; j < busyTimes.length; j++) {
```
MehulZR, line 535: To fix https://github.com/calcom/cal.com/pull/13526#pullrequestreview-2016014929 
```suggestion
        for (let j = 0; j < busyTimes.length; j++) {
```
`let j = i` will only work if the bookings are sorted.

packages/core/getUserAvailability.ts
```diff
@@ -520,10 +521,43 @@ const _getBusyTimesFromLimits = async (
...
+
+  if (eventType.seatsPerTimeSlot) {
+    for (let i = 0; i < bookings.length; i++) {
+      const booking = bookings[i];
+
+      if (
+        booking.attendeesCount &&
+        booking.attendeesCount < eventType.seatsPerTimeSlot &&
+        dayjs(booking.start).isBetween(dateFrom, dateTo)
+      ) {
```
MehulZR, line 534: Checks for the booking with remaining seats.

packages/core/getUserAvailability.ts
```diff
@@ -520,10 +521,43 @@ const _getBusyTimesFromLimits = async (
...
+              1,
+              {
+                start: busyTime.start,
+                end: dayjs(booking.start).subtract(1, "ms").toISOString(),
+              },
+              {
+                start: dayjs.utc(booking.end).toISOString(),
+                end: busyTime.end,
+              }
+            );
```
MehulZR, line 549: Iterates over the busy times to check if the *booking with remaining seats* fits in it.
If it does then split the busy time into two.

packages/core/getUserAvailability.ts
MehulZR: Explanation

packages/features/bookings/lib/handleNewBooking.ts
MehulZR: A new fourth parameter got added for checkBookingLimits() last month on the main branch.
It is complaining about that.
```suggestion
        rescheduleUid,
        eventType.schedule?.timeZone
```

packages/core/getBusyTimes.ts
```diff
@@ -343,15 +343,17 @@ export async function getBusyTimesForLimitChecks(params: {
       },
       title: true,
       userId: true,
+      attendees: true,
```
emrysal, line 346: I tried _count: { select: { attendees: true } } and indeed that doesn't work.

I think this should be: attendees: { select: { id: true } } to ensure minimal perf hits; also; ideally we only run the attendees subQuery when it is a seatsPerTimeSlot > 1 situation.

What do you think?

Udit-takkar:
@MehulZR any updates on this? Type checks and unit tests are failing

MehulZR:
@Udit-takkar I will look into it soon; currently a bit busy.

I had already left some comments regarding the type errors. Unit test failure might be related to the issue brought up by Carina (for which I have also left a comment) .

packages/core/getBusyTimes.ts
MehulZR: > I think this should be: attendees: { select: { id: true } } to ensure minimal perf hits.

Thanks for the suggestion. I have implemented it.

As for the subQuery; yeah it is necessary only for the seated events.

I have started working on this PR after quite some time and would like to stabilize it a bit. 
After that I can look into the conditional subQuery.

MehulZR:
@Udit-takkar PR is ready for review.

CarinaWolli (changes requested):
Found a bug when testing. Changing the host's profile timezone effects the available slots: https://www.loom.com/share/5121dd24653c433f9e4093425c7253fd

MehulZR:
@CarinaWolli Can you provide me some more details.
I tried to replicate the issue with what I could observe from the video but didn't find anything unusual in my testing.

MehulZR:
Explaination

@CarinaWolli I hope the explanation clears things up.

CarinaWolli:
> @CarinaWolli I hope the explanation clears things up.

ah that makes a lot of sense, thank you for clarify that for me 🙏

keithwillcode:
@CarinaWolli Is this one good to go now with the clarification?

packages/core/getUserAvailability.ts
CarinaWolli: Was that a bug? if yes, how can I reproduce that bug?

CarinaWolli (changes requested):
Found an issue, could you take a look at that? 

- created seated event type (15 min), with:
  - Limit booking frequency: 1 per day
  - Limit total booking duration: 15 min per week 
- Book one seat 
- Now the whole week is blocked
- If disabling one of the limits it works, but together the slot with the already booked seat is not showing for me

MehulZR:
> Was that a bug? if yes, how can I reproduce that bug?

@CarinaWolli Yes. It can produce values in incorrect tz and feed those to dependent functions such as getSlots().
More info regarding this bug & how to reproduce it can be found in [this](https://github.com/calcom/cal.com/pull/13526#discussion_r1494147523) msg.

MehulZR:
> Found an issue, could you take a look at that?
> 
>     * created seated event type (15 min), with:
>       
>       * Limit booking frequency: 1 per day
>       * Limit total booking duration: 15 min per week
> 
>     * Book one seat
> 
>     * Now the whole week is blocked
> 
>     * If disabling one of the limits it works, but together the slot with the already booked seat is not showing for me
Explanation

It's because limit manager can have _duplicate_ entries based on the ordering of the addition of busy times. 
Check the image above for what I mean by _duplicate_ in the current scenario.


@CarinaWolli  Should I pull a separate PR for this? As the issue isn't directly related to the current PR.

emrysal:
Sorry this is taking so long, the reason is is that there are many separate issues being targeted in the PRs that make it really time intensive to review. - and also the PRs themselves are overlapping.

I'm currently working on investigating your issue regarding subtraction - which we should open a new PR for as well with  a date-ranges.test.ts to proof this is actually the case. I'd like to do so with your support but will spearhead tomorrow.

The claim is that in non-UTC tz's the subtract function does not work properly - as the `.tz` function in dayjs is very slow I'm trying to identify the underlying issue of why it is not working so we can fix that part - it is important to be as UTC compatible  where possible as Dayjs.utc is ~400x faster.

```json
{
  "dateRangesInWhichUserIsAvailable": [
    {
      "start": "2024-08-29T07:59:59.000Z",
      "end": "2024-08-29T12:30:00.000Z"
    }
  ],
  "dateRanges": [
    {
      "start": "2024-08-29T04:00:00.000Z", // 9:30 until (as per TestData.schedules.IstWorkHours)
      "end": "2024-08-29T12:30:00.000Z" // 18:00 - 5:30 offset
    }
  ],
  "formattedBusyTimes": [
    {
      "start": "2024-08-29T02:00:00.000Z",
      "end": "2024-08-29T07:59:59.000Z" // 2H time, then a 2H after event buffer = 4 hours, is 8:00:00-1s
    }
  ]
}
```

OK - so then the first slot returned is 08:00:00.000Z - @CarinaWolli to confirm but I believe this is then wrong as I think we interval on the hour; where as this slot is starting at 13:30 - it should start at 14:00 - indeed 08:30 (I think).

I can see adding the .tz does fix the issue; but we need to extract this to in another issue to review it in isolation and get to a potentially better fix for overall performance.

      CalendarManagerMock.getBusyCalendarTimes.mockResolvedValue([
        {
          start: `${plus3DateString}T04:00:00.000Z`, // Day 3 9:30 PM IST
          end: `${plus3DateString}T05:59:59.000Z`, // Day 3 11:30 IST
        },
      ]);

Failing test case with new code:

 - Expected
+ Received

  Array [
-   "2024-08-29T08:30:00.000Z",
-   "2024-08-29T10:30:00.000Z",
+   "2024-08-29T08:00:00.000Z",
+   "2024-08-29T10:00:00.000Z",
  ]

MehulZR:
@emrysal Are you targeting the issue in a separate PR? I was waiting for an activity. As you suggested, you will be taking the initial action. Is there something that you want me to start with? I am happy to help.

PeerRich:
this was solved by our inhouse engineer @CarinaWolli

packages/core/getUserAvailability.ts
```diff
@@ -567,10 +568,43 @@ const _getBusyTimesFromLimits = async (
...
+            break;
+          }
+        }
+      }
+    }
+  }
+
```
ThyMinimalDev, line 603: complexity of this code snippet is relatively high, hard to understand, improvements can be done to make this more readable and performant, I would avoid using splice as well on the array we are iterating on

ThyMinimalDev (changes requested):

packages/core/getBusyTimes.ts
ThyMinimalDev: can we trust that this will match with the previous if statement 100% of the time ?

kart1ka:
can you pls resolve merge conflicts?

MehulZR:
@kart1ka 
The PR is around 1.5 year old. It has diverged significantly from the `main`.
It will require a significant amount of time and effort, which I currently can't commit to.

I am happy to contribute to the project in the future if there are any new opportunities.

kart1ka:
> @kart1ka The PR is around 1.5 year old. It has diverged significantly from the `main`. It will require a significant amount of time and effort, which I currently can't commit to.
> 
> I am happy to contribute to the project in the future if there are any new opportunities.

Hey @MehulZR, No problem. Thanks for your contributions. We appreciate it.

emrysal:
Hello @MehulZR - fully agree! I've bookmarked this page for future reference and will pick up based on your changes once we get to it. Right now it's not something that's really affecting us - but it's something we'd like to fix going forward.

For now closing; but that's not the end of it.

PR #13530, opened 2024-02-05
fix: add emailVerified in seed script

Udit-takkar:
## What does this PR do?

Fixes # (issue) Add EmailVerified value in User table field while seeding org members 

## Type of change


packages/prisma/seed.ts
```diff
@@ -120,6 +120,7 @@ async function createOrganizationAndAddMembersAndTeams({
         ...(await prisma.user.create({
           data: {
             ...member.memberData,
+            emailVerified: new Date(),
```
Udit-takkar, line 123: I was getting 'email verify banner' when i logged in as owner1-acme@example.com

hariombalhara (approved):

keithwillcode (approved):

PR #13535, opened 2024-02-05
fix: organization user shouldn't be asked for payment for a premium username

### LINKED ISSUE #13009: [CAL-2858] Inviting an email to an organization that passes premium username rules, requires payment
There are no premium usernames within an organization and such a user should be able to signup normally.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2858](https://linear.app/calcom/issue/CAL-2858/inviting-an-email-to-an-organization-that-passes-premium-username)</sub>

hariombalhara:
## What does this PR do?

Fixes #13009

## Type of change

## How should this be tested?
- Login to Acme.
- Invite a user like jon@acme.com which would have username **jon** which is premium in global scope(outside organization)
- When user tries to signup through the email CTA, he shouldn't be asked for payment (or get an error if stripe isn't setup locally).

NOTE that to be able to replicate premium signup flow we need to temporarily set it to true
https://github.com/calcom/cal.com/blob/42de39f173ddc012525fb08e5a85b577ca67fec7/packages/lib/constants.ts#L121

## Mandatory Tasks

## Checklist


packages/lib/server/username.ts
```diff
@@ -181,12 +181,11 @@ const usernameCheckForSignup = async ({
   const user = await prisma.user.findFirst({
     where: {
       email,
-      // Simply remove it when we drop organizationId column
-      organizationId: null,
```
hariombalhara, line 185: Membership check is enough that is there below.

keithwillcode (commented):
Could we add some tests for this one?

joeauyeung (approved):
Tested inviting "p@o.com" to suggest a shorter username.

- When inviting to a team suggests "p" username, premium username triggers
- When inviting to an org suggests  "p-o" username, premium username does not trigger
- When inviting to a team under an org suggests "p-o" username, premium username does not trigger.

I'll approve for now since it's working but let's add some tests from @keithwillcode's suggestion

zomars (approved):
LGTM

PR #13539, opened 2024-02-05
perf: use forked dayjs

keithwillcode:
## What does this PR do?

Fixes #13358 

Using our own fork of dayjs that includes major performance improvements to the .tz() function.

## Type of change

## How should this be tested?

- Ensure entire test suite is passing and do a general sweep of the system to ensure dates are handled properly.

## Mandatory Tasks


packages/dayjs/index.ts
```diff
@@ -1,16 +1,17 @@
...
-import isBetween from "dayjs/plugin/isBetween";
-import isToday from "dayjs/plugin/isToday";
-import localizedFormat from "dayjs/plugin/localizedFormat";
-import minmax from "dayjs/plugin/minMax";
-import relativeTime from "dayjs/plugin/relativeTime";
-import timeZone from "dayjs/plugin/timezone";
-import toArray from "dayjs/plugin/toArray";
-import utc from "dayjs/plugin/utc";
+
+import dayjs from "@calcom/dayjs-fork";
```
keithwillcode, line 4: Named this `@calcom/dayjs-fork` because inside the monorepo we have another package already called `@calcom/dayjs` which imports the `dayjs` package. Want to avoid collisions/confusion around this and didn't think it warranted going through the entire app to change all of the `@calcom/dayjs` imports to something like `@calcom/dayjs{something}`.

Issue #13532, opened 2024-02-05
allow emails and invite people to a team event-type directly from "assignment" if not in team yet


PeerRich:
- [ ] allow emails in assignment 
- [ ] invite person if its not in the team yet (as Member)
- [ ] allow multiple emails comma separated: user1@example.com, user2@example.com

![CleanShot 2024-02-05 at 10 17 08@2x](https://github.com/calcom/cal.com/assets/8019099/af2bc1d2-5916-4c63-9545-8431bdf68536)

atulkrs:
/ attempt #13532

anglerfishlyy:
/ attempt https://github.com/calcom/cal.com/issues/13532

pra2107tham:
Is the bounty still valid?

FayadIdriss:
I can work on this ticket I can be assigned ? thank you

VanshChitransh:
"Hi @PeerRich, I'd love to take on this issue! I’ll ensure email invites work seamlessly from 'assignment' and handle multiple emails properly. Could you assign me? Thanks!"

andresbu93:
@PeerRich  the image is not working

![Image](https://github.com/user-attachments/assets/c62e345a-1f0b-49ac-8535-276a5c5c38bc)

jayantpranjal0:
@PeerRich attached image is not showing up

VanshikaSabharwal:
Are you working on this @batmnnn

batmnnn:
@VanshikaSabharwal go for it

VanshikaSabharwal:
Yes Thank you @batmnnn

Hettbhutak:
Attempting #13532

yuvrajangadsingh:
Working on this in PR #26957 - implementation allows typing emails in team event-type assignment dropdown, auto-invites non-team-members, and supports comma-separated emails.

desaikalpana:
I’d like to work on this issue and contribute a clean, production-ready solution.

shravankumarpatell:
Hi 👋  
Is this issue still open? If yes, I’d like to work on it. Please let me know if I can take it up.  
Thanks!

Hettbhutak:
yes, It's Still Open You can Contribute. maybe others are still Working on the same may check that.

jaisshiv:
/ attempt #13532

yuvrajangadsingh:
[...]

PR #13555, opened 2024-02-06
feat: invite emails from assignment

ibex088:
## What does this PR do?

Fixes #13532
/claim #13532

[loom](https://www.loom.com/share/e22d9acd73544a279ba59d1bc4d832b4?sid=2b64d6f3-fb06-485b-a5c7-460bfa3b8037)

ps: have disabled the pending users in dropdown, because pending users dont have their Availability set-up.


supalarry:
1. Is it hard to allow adding existing team members by e-mail? I think that if we can add new members by e-mail, we should allow searching existing ones by e-mail too.
2. I think under "Team" heading above dropdown to "Add attendees" should be an explanation - "Add existing team members by name or e-mail or invite new team members using e-mail" because otherwise this feature is hidden from users.

@PeerRich do you agree?

joeauyeung:
> 1. Is it hard to allow adding existing team members not only by name but also by e-mail? I think that if we can add new members by e-mail, we should allow searching existing ones by e-mail too.
> 2. I think under "Team" heading above dropdown to "Add attendees" should be an explanation - "Add existing team members by name or e-mail or invite new team members using e-mail" because otherwise this feature is hidden from users.
> 
> @PeerRich do you agree?

What if under the user dropdown link we add an option "Invite new team members" or something similar that opens the member invite dialog. That way we can reuse the logic or email or username look up and role assignment. I think the only thing that needs to be added would be to auto assign them to the team but not included in the availability query (might already exist?)

CarinaWolli (changes requested):
Can you please address my comments below, let me know if you have any questions 
🙏

Also, there are merge conflicts. Most of them are probably coming from this PR: https://github.com/calcom/cal.com/pull/13566

SomayChauhan:
@CarinaWolli 
are we sure we want to go forward with this approach
i mean @joeauyeung suggestion was pretty good too

CarinaWolli:
> What if under the user dropdown link we add an option "Invite new team members" or something similar that opens the member invite dialog. That way we can reuse the logic or email or username look up and role assignment. I think the only thing that needs to be added would be to auto assign them to the team but not included in the availability query (might already exist?)

Sorry, I totally missed that comment. I do think we should make it more clear, like @supalarry said, right now this feature would be hidden for users. @joeauyeung where would you put the option to 'Invite new team members'? I think we want to avoid to add anything below the list of hosts, as this list can get long

CarinaWolli:
Maybe @ciaranha can give some input here how to best do that ui/ux-wise

CarinaWolli:
@SomayChauhan Is this the same as https://github.com/calcom/cal.com/pull/14185?

SomayChauhan:
yes!!, closing this in favour of https://github.com/calcom/cal.com/pull/14185

packages/features/eventtypes/components/CheckedTeamSelect.tsx
```diff
@@ -33,6 +104,34 @@ export const CheckedTeamSelect = ({
...
         value={value}
+        onBlur={() => {
+          setError(null);
+        }}
+        onInputChange={(e) => {
+          const targetValues = e.split(/[\n,]/);
+          const emails =
+            targetValues.length === 1
+              ? targetValues[0].trim().toLocaleLowerCase()
+              : targetValues.map((email) => email.trim().toLocaleLowerCase());
```
hamzakhalidchouhdary, line 115: `const emails = targetValues.map((email) => email.trim().toLocaleLowerCase());`

PR #13566, opened 2024-02-06
feat: round-robin priority ranking

### LINKED ISSUE #11545: [CAL-2604] Priority Ranking for Members in Round Robin Events

(Running a remote sales team, I'm always monitoring who the high-performers are on the team so I can send more calls into their calendars & less into the newly onboarded salesreps. 

When everyone is on the same calendar, round-robin style - it's super important we're able to prioritize the best team members. So they get inbound calls first, and then when their calendar is full it moves to the next person on the priority list.)

(Simply add a Star or "Score" rating inside of the event, next to the attached members - so we can prioritize members by order of importance on each event. 

Ideally this is also accessible through the API, so we can automatically update member priority on our sales calendar based on indicators we're measuring in our back-end business systems.)

(Other scheduling tools like Calendly, but the issue is their prioritization is very limiting... it only has a rank of 1-3, and no API access so it's not useful with larger teams - especially when trying to automate it.)

<sub>[CAL-2604](https://linear.app/calcom/issue/CAL-2604/priority-ranking-for-members-in-round-robin-events)</sub>

CarinaWolli:
## What does this PR do?

Adds priority ranking to round-robin: Every round-robin host has a priority, by default 'medium'. Until now round-robin booked the least recently booked available users. Now it will book the available user with the highest priority, if several users have the same priority it will choose the least recently booked user. 

Also, adds some design changes to the assignment tab of event-type settings as it had several design issues. @ciaranha not sure if we have a final figma design, but I used this [Figma Design](https://www.figma.com/file/xk4HOxtSI82J0F7enMxeak/Live?type=design&node-id=6287-304807&mode=design&t=LQECV5Vo7qqjjjTN-0). I am open for feedback, here's how it looks:

#### Clickable priority label on round-robin host (sorted hosts):
Screenshot 2024-02-07 at 10 38 44 AM

#### Dialog to change priority:
Screenshot 2024-02-07 at 10 35 24 AM

#### Activate fixed hosts on round-robin event
Screenshot 2024-02-07 at 10 32 49 AM

#### Collective Event
Screenshot 2024-02-07 at 10 33 13 AM

#### Enable 'Assign all team members'
Screenshot 2024-02-07 at 10 34 39 AM

Fixes #11545

## Type of change


Udit-takkar (commented):
Screenshot 2024-02-08 at 10 39 19 PM

Can we set priority when this switch is turned on?

CarinaWolli:
> Can we set priority when this switch is turned on?

No that's currently not possible, all hosts will have medium priority. I can revisit that in a follow-up

emrysal (commented):
codewise looks good, will test through ui tomorrow morning

packages/prisma/schema.prisma
```diff
@@ -41,6 +41,7 @@ model Host {
   eventType   EventType @relation(fields: [eventTypeId], references: [id], onDelete: Cascade)
   eventTypeId Int
   isFixed     Boolean   @default(false)
+  priority    Int?
```
Udit-takkar, line 44: can we just add a default value of 2 here?

Udit-takkar (approved):
@CarinaWolli LGTM. Awesome work

I'll wait for Alex review before merging

PR #13571, opened 2024-02-06
fix: org user event types with same event slug

CarinaWolli:
## What does this PR do?

When org members had the same event slugs wrong event types were shown on org event type links (`acme.cal.local:3000/user/slug`)


exception (approved):

joeauyeung (approved):
Found the same fix.

This was used in a `findFirst` query so it was grabbing the first member of the org with the same slug.

zomars (approved):

PR #13581, opened 2024-02-07
feat: overlay calendar busy dates events

ThyMinimalDev:
## What does this PR do?

overlay calendar busy times events for booker atoms


apps/api/v2/src/ee/overlay-calendars/controllers/overlay-calendars.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+import { AccessTokenGuard } from "@/modules/auth/guards/access-token/access-token.guard";
+import { Controller, Get, Logger, UseGuards, Body } from "@nestjs/common";
+import { User } from "@prisma/client";
+
+import { SUCCESS_STATUS } from "@calcom/platform-constants";
+import { ApiResponse } from "@calcom/platform-types";
+import { EventBusyDate } from "@calcom/types/Calendar";
+
+@Controller({
+  path: "ee/overlay-calendars",
```
supalarry, line 13: Why the `ee` prefix for path? I think that some endpoints (non ee folder) not having ee in front and others having might be more confusing developer experience unless I am missing something.

apps/api/v2/src/ee/overlay-calendars/controllers/overlay-calendars.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+  ): Promise<ApiResponse<EventBusyDate[]>> {
+    const { loggedInUsersTz, dateFrom, dateTo, calendarsToLoad } = body;
+    if (!dateFrom || !dateTo) {
+      return {
+        status: SUCCESS_STATUS,
+        data: [],
+      };
+    }
+
+    const credentials = await this.overlayCalendarsService.getUniqCalendarCredentials(
```
supalarry, line 35: Should OverlayCalendarService have `getBusyTimes` function and controller instead calls it?

apps/api/v2/src/ee/overlay-calendars/inputs/get-overlay-calendars-busy-times.input.ts
```diff
@@ -0,0 +1,28 @@
...
+  externalId!: string;
+}
+
+export class GetOverlayCalendarsBusyTimesInput {
+  @IsString()
+  loggedInUsersTz!: string;
+
+  @IsString()
+  @IsOptional()
+  dateFrom!: string | null;
```
supalarry, line 18: For dateFrom and dateTo - what is the expected date shape? Is it `day-month-year` or `day/month/year` ? Maybe we can create a custom validator like `create-user.input.ts` has `IsTimeFormat`, but in case of dateTo and dateFrom it would check date passed and if its not good respond with the appropriate format.

apps/api/v2/src/ee/overlay-calendars/services/overlay-calendars.service.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    if (credentials.length !== uniqueCredentialIds.length) {
+      throw new UnauthorizedException("These credentials do not belong to you");
+    }
+
+    return credentials;
+  }
+
+  async getCalendarsWithCredentials(
+    credentials: ValueType<ReturnType<typeof this.credentialsRepository.getUserCredentialsByIds>>,
```
supalarry, line 26: Wherever `ValueType<ReturnType...` is used - shall that be created as a custom type in the `credentialsRepository` in this case and then have `credentials: Credentials` ?

apps/api/v2/src/ee/overlay-calendars/services/overlay-calendars.service.ts
```diff
@@ -0,0 +1,77 @@
...
+import { CredentialsRepository } from "@/modules/credentials/credentials.repository";
+import { Injectable, InternalServerErrorException, UnauthorizedException } from "@nestjs/common";
+import { User } from "@prisma/client";
+import { DateTime } from "luxon";
+
+import { getBusyCalendarTimes } from "@calcom/platform-libraries";
+import { ValueType } from "@calcom/platform-types";
+
+@Injectable()
+export class OverlayCalendarsService {
```
supalarry, line 11: Should the service be called `CalendarsService`? Functions like `getUniqCalendarCredentials` and `getCalendarsWithCredentials` sounds generic and applicable to all calendars (my guess).

Also, maybe then we should not have `ee/overlay-calendars` folder but just `ee/calendars` and also simply `/calendars` endpoint that returns data, and then is it used in overlay calendar or not depends on consumer if possible.

apps/api/v2/src/ee/overlay-calendars/controllers/overlay-calendars.controller.ts
ThyMinimalDev: I think all the endpoints in ee should have this prefix to avoid confusion

apps/api/v2/src/ee/overlay-calendars/inputs/get-overlay-calendars-busy-times.input.ts
ThyMinimalDev: this is the format 2024-02-07

PR #13583, opened 2024-02-07
fix: 2FA OTP flaky test

Udit-takkar:
## What does this PR do?

Fixes # (issue)

Reason why this test failed:-

We auto submit 2FA code when all the input elements are filled. In the E2E we first test wrong input and check for error message but when we fill in the correct six digit code every time we fill the code in input the form gets auto submitted because of 

```
  useEffect(() => {
    if (totpCode?.trim().length === 6) {
      form.handleSubmit(handleEnableRef.current)();
    }
  }, [form, handleEnableRef, totpCode]);
  
  ```
  
  so an edge case like in the screenshot while we are filling in the code the last digit is same of both the correct code and incorrect code the form will successfully get submitted and line page.fill
locator('input[name="2fa6"]') failed

Screenshot 2024-02-07 at 6 16 51 PM

## Requirement/Documentation

## Type of change

### CODE REVIEW

File: apps/web/playwright/login.2fa.e2e.ts [hunk 1]
```diff
@@ -162,6 +164,14 @@ test.describe("2FA Tests", async () => {
   });
 });
 
+async function removeOtpInput(page: Page) {
+  await page.locator('input[name="2fa6"]').waitFor({ state: "visible", timeout: 60_000 });
```
keithwillcode, line 168: This is a long timeout. Can we do 30 seconds instead?

Udit-takkar: Done

keithwillcode (dismissed):

zomars (approved):
LGTM

PR #13585, opened 2024-02-07
refactor: schedules for atom

supalarry:
This PR modifies [schedules.controller.ts](https://github.com/calcom/cal.com/pull/13585/files#diff-831f89f78a1bca4514e7a14a19c32fd673be76fdb8d1da9ec0fa5128a60627f4) in order to transform returned schedule if an atom requested it: 

1. As you see in the controller, we use "ForAtom()" decorator to check if query param "?for=atom" was attached and if schedule should be formatted.
2. Since formatting logic is more complex, I created a new [response.service.ts](https://github.com/calcom/cal.com/pull/13585/files#diff-b5b60c01c7bafa7fe78dcb203b8ce02041b6090f02af499518d428b1d24efb47) that is responsible for formatting schedules endpoint response. Its functions accept fetched schedule, user associated with access token and "forAtom" parameter. If it is for atom, then it will transform it using transformers from the [platform/libraries/index.ts](https://github.com/calcom/cal.com/pull/13585/files#diff-7beeedab74af0908112d5126751ede01561e06ce3a3ffa7e2dc6853c16fdc482) - these transformers are used in our web trpc handlers when returning schedule for current availability / schedule component (#13586). If it is not for atom, just parse via zod to only include fields we want in the response.


apps/api/v2/src/ee/schedules/controllers/schedules.controller.ts
```diff
@@ -27,81 +31,91 @@ import { CreateScheduleInput } from "../inputs/create-schedule.input";
...
-    @Body() bodySchedule: CreateScheduleInput
-  ): Promise<ApiResponse<{ schedule: ScheduleResponse }>> {
-    const schedule = await this.schedulesService.createUserSchedule(userId, bodySchedule);
-    const scheduleResponse = schemaScheduleResponse.parse(schedule);
+    @GetUser() user: User,
+    @Body() bodySchedule: CreateScheduleInput,
+    @ForAtom() forAtom: boolean
+  ): Promise<ApiResponse<{ schedule: ScheduleResponse | ScheduleWithAvailabilitiesForWeb }>> {
+    const schedule = await this.schedulesService.createUserSchedule(user.id, bodySchedule);
+    const scheduleFormatted = await this.schedulesResponseService.formatSchedule(forAtom, user, schedule);
```
Ryukemeister, line 46: NIT: wouldn't it be better if we name this as `formattedSchedules`?

apps/api/v2/src/ee/schedules/services/response/response.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    schedules: ScheduleWithAvailabilities[]
+  ): Promise<ScheduleResponse[] | ScheduleWithAvailabilitiesForWeb[]> {
+    if (forAtom) {
+      const usersSchedulesCount = await this.schedulesRepository.getUserSchedulesCount(user.id);
+      return Promise.all(
+        schedules.map((schedule) => this.transformScheduleForAtom(schedule, usersSchedulesCount, user))
+      );
+    }
+    return schedules.map((schedule) => schemaScheduleResponse.parse(schedule));
+  }
```
Ryukemeister, line 45: Why not keep the `forAtom` parameter here optional and instead give it a default value of false? That way only when we want to fetch schedule for any atom we'll pass that param.

Ryukemeister (approved):

apps/api/v2/src/ee/schedules/controllers/schedules.controller.ts
supalarry: both are fine - if formatted is pre-fixed like you suggest or suffix like i have both are good

apps/api/v2/src/ee/schedules/services/response/response.service.ts
supalarry: It is not optional because it will always be defined - the controller (endpoint) uses "@ForAtom" decorator that always returns true or false, which then is passed to formatter.

PR #13596, opened 2024-02-08
PLA-44 feat: Platform getPublicEvent

exception:
Implement `getPublicEvent` from `features`, used by the `useEvent` hook within the Booker Atom.


apps/api/v2/src/modules/events/controllers/events.controller.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Controller("events")
+export class EventsController {
+  constructor(private readonly prismaReadService: PrismaReadService) {}
+
+  @Get("/")
+  @Version(VERSION_NEUTRAL)
+  async getPublicEvent(@Body() input: GetPublicEventInput): Promise<ApiResponse> {
+    const event = await getPublicEvent(
+      input.username,
```
supalarry, line 17: We can spread "...input" when passing to getPublicEvent

apps/api/v2/src/modules/events/input/get-public-event-input.ts
```diff
@@ -0,0 +1,17 @@
...
+  username!: string;
+
+  @IsString()
+  eventSlug!: string;
+
+  @IsBoolean()
+  isTeamEvent?: boolean;
+
+  @IsString()
+  org!: string | null;
```
supalarry, line 16: If it must be defined, but string value can't be passed, then null must be supplied. I think it would be better dev experience if org IsOptional instead of explicitly requiring to pass null.

apps/api/v2/src/modules/events/input/get-public-event-input.ts
```diff
@@ -0,0 +1,17 @@
...
+export class GetPublicEventInput {
+  @IsString()
+  @Transform(({ value }: { value: string }) => value.toLowerCase())
+  username!: string;
+
+  @IsString()
+  eventSlug!: string;
+
+  @IsBoolean()
+  isTeamEvent?: boolean;
```
supalarry, line 13: I see ? in "isTeamEvent?" - it should have @IsOptional() then

PR #13588, opened 2024-02-07
fix: Revert event-type profile based querying

hariombalhara:
Revert event-type profile based querying as we need to test with huge number of team events first before doing profile querying.

[Acme profile still working after the changes](https://www.loom.com/share/7313876da0994c4291590579829cf27e)
[teampro also working](https://www.loom.com/share/d4c5313264ba43bba8dd0be4e1e899bd)

PR #13599, opened 2024-02-08
fix: Reduce payload even further.

hariombalhara:
Reverts https://github.com/calcom/cal.com/pull/13588 and then improves upon the following to further decrease the payload size
- Removes event-types from being available inside eventTypes[number].team
- Deduplicate all users in the payload. event-type just has the reference to userId.

**NOTE**: Because the first commit is just a revert commit of #13588. Use this link to review the changes https://github.com/calcom/cal.com/pull/13599/files/309e073303c60e3cbaff7a24401883131b62d538..8d9df5ff92e007b11ffa7dde75185dc9dd6e77d9#diff-5294731b960e76c72f504ce7e1381d51e5c054131976fe7f43bb38f485d76aa6 (it excludes that revert commit)

Profile Based Querying - Before
image

Profile Based Querying - After - THIS BRANCH -> **Though the time taken to serve it is significantly more than non-profile based querying, the payload is smaller. I am hoping this won't be a big issue in production.** We can work on improving that later. I expect mostly increase is due to every **user** needing a join with **profile** now.
image

Non Profile Based Querying 
image


packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -1,16 +1,20 @@
-import { Prisma } from "@prisma/client";
 // eslint-disable-next-line no-restricted-imports
```
hariombalhara, line 1: This file is the same as it was earlier before we reverted profile querying except for doing the normalization before sending the payload.

packages/trpc/server/routers/viewer/eventTypes/getByViewer.handler.ts
```diff
@@ -332,16 +289,38 @@ export const getByViewerHandler = async ({ ctx, input }: GetByViewerOptions) =>
...
+        }),
+      };
+    });
+
+    return {
+      ...payload,
+      allUsersAcrossAllEventTypes,
+      eventTypeGroups,
+    };
   }
```
hariombalhara, line 325: This is the normalization change

apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -858,15 +880,19 @@ const Main = ({
   }
 
   const isFilteredByOnlyOneItem =
-    (filters?.teamIds?.length === 1 || filters?.userIds?.length === 1) && data.eventTypeGroups.length === 1;
+    (filters?.teamIds?.length === 1 || filters?.userIds?.length === 1) &&
+    rawData.eventTypeGroups.length === 1;
+
+  const data = denormalizePayload(rawData);
```
hariombalhara, line 886: Because the sent payload is in a different format now, we need to denormalize it before others can keep consuming it without any change.

apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -858,15 +880,19 @@ const Main = ({
...
+
   return (
     <>
       {data.eventTypeGroups.length > 1 || isFilteredByOnlyOneItem ? (
         <>
           {isMobile ? (
             <MobileTeamsTab eventTypeGroups={data.eventTypeGroups} />
           ) : (
-            data.eventTypeGroups.map((group: EventTypeGroup, index: number) => (
+            data.eventTypeGroups.map((group, index: number) => (
```
hariombalhara, line 895: Wasn't needed.

apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -238,7 +260,7 @@ export const EventTypeList = ({
       await utils.viewer.eventTypes.getByViewer.cancel();
       const previousValue = utils.viewer.eventTypes.getByViewer.getData();
       if (previousValue) {
-        const newList = [...types];
+        const newList = [...types.map(normalizeEventType)];
```
hariombalhara, line 263: Optimistic updates need to do in memory conversion to make the format consistent with react query response for getByViewer

Udit-takkar (approved):

PR #13607, opened 2024-02-09
feat: PLA-47 Reserve Slots Handler

exception:
Implement `POST /slot` for reserving slots, used by the Booker Atom.


apps/api/v2/src/modules/slots/inputs/reserve-slot.input.ts
```diff
@@ -0,0 +1,16 @@
+import { IsDate, IsInt, IsOptional, IsString } from "class-validator";
+
+export class ReserveSlotInput {
+  @IsInt()
+  eventTypeId!: number;
+
+  @IsDate()
+  slotUtcStartDate!: string;
```
supalarry, line 8: We use `@IsDate()` decorator while `slotUtcStartDate` and `slotUtcEndDate` are of `: string` type.

We could use `@IsISO8601()` validator instead - Checks if the string is a valid ISO 8601 date. If given value is not a string, then it returns false. Use the option strict = true for additional checks for a valid date, e.g. invalidates dates like 2019-02-29.

supalarry (approved):
Looks good - only question is comment below for the input validation.

PR #13612, opened 2024-02-09
Feat booker bookings

ThyMinimalDev:
## What does this PR do?

Bookings endpoints for booker


apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      const booking = await handleNewBooking(req as unknown as NextApiRequest & { userId?: number });
+      return {
+        status: SUCCESS_STATUS,
+        data: booking,
+      };
+    } catch (err) {
+      handleBookingErrors(err);
+    }
+    throw new InternalServerErrorException("Could not create booking.");
```
ThyMinimalDev, line 68: typescript is not able to detect that we already return or throw already in the try catch, adding this to avoid type errors but this is unreachable code

PR #13625, opened 2024-02-09
fix: oncancel modal action week layout

alannnc:
## What does this PR do?

When in week layout after you've selected a slot and then click on the Back button to remove the modal the page crashes as it tries to resolve dayjs(null).

Fixes #
https://www.loom.com/share/96348c529d7944108144e6f7f7c6644e

## Type of change

## Mandatory Tasks


zomars (approved):
LGTM

PR #13627, opened 2024-02-09
feat: Add Google Profile Photo when connecting Google Calendar

### LINKED ISSUE #13492: [CAL-3065] Prefill Avatar with Google avatar after connecting a Google Calendar
i think this is possible

would be nice if people sign up with email + password but then add a Google Calendar

<sub>[CAL-3065](https://linear.app/calcom/issue/CAL-3065/get-someones-google-avatar-after-connecting-a-google-calendar)</sub>

zeeshanbhati:
## What does this PR do?

If a user links their Google Calendar and hasn't set an avatar, we'll retrieve their profile photo from their connected Google account and use it as their profile picture.

Fixes : #13492

## Requirement/Documentation
  - A new Scope `https://www.googleapis.com/auth/userinfo.profile` will be added to the calendar scope.
 
## Type of change

## How should this be tested?

-   Create an account using email and password.
-   Do not set a profile picture during profile creation or through the profile section.
-   Go to the Apps section and install Google Calendar. Ensure that you use an account that has a profile picture.
-   Verify if that profile picture has been set as the profile picture for cal.com.

## Checklist


keithwillcode:
@joeauyeung is it worth adding another scope just to get the profile image?

baileypumfleet:
The scope change could add a lot of friction to making this PR possible

zeeshanbhati:
Can we fetch the profile picture without using this scope ? 

This is also a non critical scope

Also, what kind of friction can it introduce?

Is it something related to compliance or something?

joeauyeung:
@zeeshanbhati the fear was if we added a new scope would it invalidate access tokens from other users. Testing this by authenticating GCal w/o the profile scope. Then on another account authenticating w/ the scope. When refreshing the token and creating a calendar event with the token w/o the profile scope, it worked as expected.

@baileypumfleet @keithwillcode the profile scope is non-sensitive.
CleanShot 2024-02-14 at 10 48 24

joeauyeung (approved):
Reviewed new scope change (see https://github.com/calcom/cal.com/pull/13627#issuecomment-1944104188)

Test cases:
- With no avatar set ✅ 
- Shouldn't overwrite preset avatar ✅ 
- Throwing an error in `setProfilePicture` doesn't block creation of calendar credential ✅

packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -91,6 +93,10 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
       throw new HttpError({ message: "Internal Error", statusCode: 500 });
     }
 
+    // Update the user's profile photo with google profile photo if it's null
+    // Since we don't want to block the user from using the app if this fails, we don't await this
```
zomars, line 97: This doesn't really work in serverless environments. If you don't await and the main endpoint responds first the call will be dropped.

zomars:
There's a race condition happening in this PR.

packages/app-store/googlecalendar/api/callback.ts
joeauyeung: TIL. Can we just wrap it in a try/catch block and if it fails just continue on with the rest of the code?

joeauyeung: @zomars there's already a try/catch in the function it's self. I tested throw an error in the function and it still creates the credential. I'll add the await and one more change in a follow up PR.

Issue #13549, opened 2024-02-06
[CAL-3082] eventTypes/getByViewer regression on v3.7.12


zomars:
rel #13002

It seems like we're having a increased payload on latest release. It is causing this endpoint to fail. This is what we have on the runtime logs:

```
[PERF]: getByViewer(109147) took 402.34358200000133ms
API response for /api/trpc/eventTypes/getByViewer?batch=1&input={"0":{"json":{"filters":{"teamIds":[REDACTED],"upIds":null,"userIds":null}},"meta":{"values":{"filters.upIds":["undefined"],"filters.userIds":["undefined"]}}}} exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit
[ERROR] [1707184498477] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 413.
```

There's also a potential leakage of password of your team members as well.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3082](https://linear.app/calcom/issue/CAL-3082/eventtypesgetbyviewer-regression-on-v3712)</sub>

zomars:
Testing locally seems like the payload is now 34MB 🤯 

![image.png](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/dbd2cb26-b03d-4154-be39-547b48635213/03b2ad59-a588-44f6-b2a4-200761c4d7f4)

zomars:
Reduced to 1MB in #13552 

![image](https://github.com/calcom/cal.com/assets/3504472/8c118339-af27-49fe-b8d2-69ee6c773d0b)

PR #13002, opened 2024-01-03
feat: New Organization Schema along with the ability to invite existing users to an organization

hariombalhara:
## What does this PR do?

Existing cal.com user can be invited to an organization now. 
Fixes #13013
Fixes #13002 
Fixes #13367
Fixes #13359

Demo - https://www.loom.com/share/a980bb3fc14140b99355aed36b639134

image

This is a big PR but it wasn't possible to break it down significantly further.

## Schema Changes
- Deprecate organizationId and organization from User table
   - Will be removed in near future.
- Instead have Profile table with organizationId and userId relation.

## Seeder
- Added organization seeder that adds Acme and Dunder Mifflin organizations with some event-types and teams.

## Repositories
- Some repositories created `Profile`, `User`, `Membership` which mostly abstract out the reusable queries together.
- A repository is a class at the moment just to be a collection of methods, we can do other OOP stuff later on.

## Brainstorming and Release Plan here 
https://docs.google.com/document/d/1ii3eyn0F0LaUCrX1c58YgdMt2V7xZWI2AGX1yKKM2Ik/edit

## New Organization Accept Invite View inside Teams itself
image

## TODO
- [x] Write migration to create Profile table and populate it with all org user entries
- [x]  ~~Open a separate PR that would go live before this. [Will do this once the PR is fully reviewed and is ready to be merged.~~ Chose against doing it. It is a bit of effort to separate the migration and I see no risk with going with the migration along with the changes as part of this PR.

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## How should this be tested?
There is a lot to test

## Mandatory Tasks

- [ ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist
- I haven't added tests that prove my fix is effective or that my feature works

PR #13628, opened 2024-02-09
refactor: migrates user password to it's own table

zomars:
## What does this PR do?

Follow up to #13002 #13549 

- Migrates user password to it's own table
- Makes it harder to leak user password by using include user: true
- Makes it explicit when you're trying to manipulate password
- Adds special migration so there no user action required

## Type of change

## How should this be tested?

- yarn e2e
- Test signup

## Mandatory Tasks

### DIFF
--- a/packages/prisma/migrations/20240209223121_adds_user_password/migration.sql
+++ b/packages/prisma/migrations/20240209223121_adds_user_password/migration.sql
@@ -1,0 +1,31 @@
+/*
+Warnings:
+
+- You are about to drop the column `password` on the `users` table. All the data in the column will be lost.
+
+ */
+-- CreateTable
+CREATE TABLE
+  "UserPassword" (
+    "hash" TEXT NOT NULL,
+    "userId" INTEGER NOT NULL
+  );
+
+-- CreateIndex
+CREATE UNIQUE INDEX "UserPassword_userId_key" ON "UserPassword" ("userId");
+
+-- AddForeignKey
+ALTER TABLE "UserPassword" ADD CONSTRAINT "UserPassword_userId_fkey" FOREIGN KEY ("userId") REFERENCES "users" ("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- Inserts current user password into new table
+INSERT INTO
+  "UserPassword" ("hash", "userId")
+SELECT
+  u."password",
+  u.id
+FROM
+  users u;
+
+-- AlterTable
+ALTER TABLE "users"
+DROP COLUMN "password";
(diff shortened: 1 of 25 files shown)


zomars:
* `main`
  * **PR #13628** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/13628?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/calcom/cal.com/13628?utm_source=stack-comment).

zomars (commented):
Self review done

packages/prisma/migrations/20240209223121_adds_user_password/migration.sql
```diff
@@ -0,0 +1,31 @@
...
+  "UserPassword" ("hash", "userId")
+SELECT
+  u."password",
+  u.id
+FROM
+  users u;
+
+-- AlterTable
+ALTER TABLE "users"
+DROP COLUMN "password";
```
sean-brydon, line 31: Should we drop this column in a future PR to ensure nothing goes wrong during this migration? Just a thought i had reading this

keithwillcode: Yeah agreed we should do this in a two-step process.

zomars: Valid concerns although there shouldn't be much problem. Already tested with production-like data. 

I'm going to rename the column to passwordDeprecated to avoid accidental use instead and drop it in a future release. 

That way we can have a recoverable state just in case.

sean-brydon: Perfect - i just know how much drama i caused last time with the org migration PR :') i want to make sure that doesnt happen again

zomars: @sean-brydon @keithwillcode Addressed feedback. Please re-review 🙏🏽

keithwillcode (approved):

packages/prisma/migrations/20240209223121_adds_user_password/migration.sql
```diff
@@ -0,0 +1,31 @@
...
+INSERT INTO
+  "UserPassword" ("hash", "userId")
+SELECT
+  u."password",
+  u.id
+FROM
+  users u;
+
+-- We rename instead of dropping the column to avoid possible data loss, will be dropped in the next migration.
+ALTER TABLE "users"
```
keithwillcode, line 30: @zomars i think this alter statement needs to come in a later migration because otherwise this will create downtime where users can’t log in

zomars: Will separate it then

PR #13646, opened 2024-02-12
fix: Add location to `ics` file

joeauyeung:
## What does this PR do?

This PR adds the `LOCATION` field to the ics file Cal.com generates.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a booking with a conferencing app. Check the ics file that was sent with the booking confirmation email. The `LOCATION` field should be included
- Do this for phone numbers, address, and a custom link
## Mandatory Tasks

### DIFF
--- a/packages/emails/lib/generateIcsString.ts
+++ b/packages/emails/lib/generateIcsString.ts
@@ -6,6 +6,7 @@ import { RRule } from "rrule";
 import dayjs from "@calcom/dayjs";
 import { getRichDescription } from "@calcom/lib/CalEventParser";
 import { getWhen } from "@calcom/lib/CalEventParser";
+import { getVideoCallUrlFromCalEvent } from "@calcom/lib/CalEventParser";
 import type { CalendarEvent, Person } from "@calcom/types/Calendar";
 
 export enum BookingAction {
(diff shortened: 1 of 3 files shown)


packages/lib/test/builder.ts
```diff
@@ -176,7 +179,7 @@ export const buildCalendarEvent = (event?: Partial<CalendarEvent>): CalendarEven
     customInputs: {},
     additionalNotes: faker.lorem.paragraph(),
     organizer: buildPerson(),
-    videoCallData: buildVideoCallData(),
+    ...(!omitVideoCallData && { videoCallData: buildVideoCallData() }),
```
zomars, line 182: Just curious. Why is this needed?

joeauyeung: When generating the ics string we look for video call data first and then the location. Since video call data was always included in the event builder the ics string would always fallback to the video call url.

PeerRich (dismissed):

packages/emails/lib/generateIcsString.ts
```diff
@@ -93,6 +96,7 @@ const generateIcsString = ({
           }))
         : []),
     ],
+    location,
```
Udit-takkar, line 99: Type checks are failing

Udit-takkar (approved):

PR #13649, opened 2024-02-13
feat: Disallowing use of Prisma includes using only "true" and not a field selector

### LINKED ISSUE #13577: [CAL-3087] Make linting rule disallowing use of Prisma includes using only "true" and not a field selector
To ensure we only retrieve necessary data from the db and to protect unwanted data from crossing application boundaries.

Example:

include: {

  user: true;

}

Should cause a linting error

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3087](https://linear.app/calcom/issue/CAL-3087/make-linting-rule-disallowing-use-of-prisma-includes-using-only-true)</sub>

suyash5053:
## What does this PR do?
Created a linting rule for Prisma that disallow's use of Prisma, including only true not a field selector
Fixes #13577 

## Type of change

## Mandatory Tasks

## Checklist


keithwillcode:
Also implemented in https://github.com/calcom/cal.com/pull/13652, which is recursive. Going to favor that implementation

PR #13651, opened 2024-02-13
fix: get slots & get slots test

MehulZR:
## What does this PR do?

Fixes: One of the test which fails when run in a different timezone & rare get slots issue with min. booking notice.

https://github.com/calcom/cal.com/pull/13526#issuecomment-1938638415

## Type of change

- [x] Tests (Unit/Integration/E2E or any other test)

## Mandatory Tasks


packages/lib/slots.ts
```diff
@@ -194,7 +194,7 @@ function buildSlotsWithDateRanges({
 
     slotStartTime = slotStartTime.add(offsetStart ?? 0, "minutes").tz(timeZone);
 
-    while (!slotStartTime.add(eventLength, "minutes").subtract(1, "second").utc().isAfter(rangeEnd)) {
```
MehulZR, line 197: We don't have to convert it into the UTC as it doesn't changes the absolute time, just changes from where we are observing the time.

packages/lib/slots.test.ts
```diff
@@ -210,7 +210,9 @@ describe("Tests the slot logic", () => {
       inviteeDate: dayjs().add(1, "day"),
       frequency: 20,
       minimumBookingNotice: 0,
-      dateRanges: [{ start: dayjs("2021-06-21T00:00:00.000Z"), end: dayjs("2021-06-21T23:45:00.000Z") }],
```
MehulZR, line 213: In different tz like IST, this will be converted into `2021-06-21T05:30:00+05:30`
We will try to find the first time slot keeping in mind that the availability interval is `20`.
The value of minutes is `30` which ends up eating a slot and failing the test in IST.

packages/lib/slots.ts
```diff
@@ -163,11 +163,11 @@ function buildSlotsWithDateRanges({
   const slots: { time: Dayjs; userIds?: number[] }[] = [];
 
   dateRanges.forEach((range) => {
-    const startTimeWithMinNotice = dayjs.utc().add(minimumBookingNotice, "minute");
```
MehulZR, line 166: Again this being in UTC might cause issue with availability interval due to change in value of minutes.

packages/lib/slots.test.ts
```diff
@@ -210,7 +210,9 @@ describe("Tests the slot logic", () => {
...
+      dateRanges: [
+        { start: dayjs("2021-06-21T00:00:00.000Z").utc(), end: dayjs("2021-06-21T23:45:00.000Z").utc() },
```
keithwillcode, line 214: @MehulZR In which format does the caller send these dates? Utc? I want to make sure that the test and the production code send the dates in the same format

packages/lib/slots.test.ts
emrysal: That is a great diagnosis, but don't we want to keep the test the same and support this feature in IST? This way seems more like we explicitly don't work in IST in order to keep the tests passing - but dateRanges isn't necessarily in UTC

emrysal (changes requested):
Just putting this in changes requested so we can get to the bottom of understanding this fix and prevent an auto merge.

packages/lib/slots.test.ts
MehulZR: Dates can be in any tz as Dayjs Object.

packages/lib/slots.test.ts
MehulZR: I am afraid I don't understand what you mean by supporting this feature in IST. The test result is correct for UTC and any offset such as `xx:20`/`xx:40`/`xx:00` (`xx:20`/`xx:40` probably doesn't exist).

In case for IST or any offset such as `xx:30`, the correct answer would be 70.

Explanation

In UTC the correct result would be 5. In Iran the correct result is 4.

This is because the slots interval takes attendee tz in mind.

MehulZR: The higher the interval, the more prominent the issue.
Currently we support these intervals `60` `30` `20` `15` `10`.
GCF of all these is `5` which will solve the issue but that would be a performance issue.

packages/lib/slots.test.ts
keithwillcode: Yeah I know but I would like the tests to call the functions like the production code calls them so we expect and see consistent results.

MehulZR: In production, the fn is called with different timezones. and the result (no. of slots) is dependent on the that timezone.
So, In test I have converted the tz into UTC because the result is correct for the UTC tz.

anikdhabal (commented):
hey @MehulZR pls fix the conflict

packages/lib/slots.test.ts
MehulZR: `getSlots` fn takes `dateRanges` in the host's tz.

MehulZR: `dayjs()` will return the dates in the local tz causing the test to fail in tz other than UTC.

keithwillcode:
Closing this as it's a year old. Thank you for the contribution @MehulZR but this has continued to be pushed due to the risk of the changes and we haven't been able to properly vet it in order to merge.

PR #13662, opened 2024-02-13
fix: amend user password migration

zomars:
## What does this PR do?

- Amending #13628 since it hasn't been released. 
- Will delegate altering the column to next release to prevent possible downtime.


keithwillcode (approved):

kart1ka:
HI @zomars The password column is still being dropped.
When running yarn workspace @calcom/prisma db-migrate a new migration is being created which drops the password column from users table.

PR #13665, opened 2024-02-13
fix: route always selects 'custom' after saving an event redirect

CarinaWolli:
## What does this PR do?

Fixes issue that an event redirect route always switches to custom after saving, even when an event type was picked in the select. Everything was saved correctly to the db, this is just a displaying issue.

### DIFF
--- a/packages/app-store/routing-forms/pages/route-builder/[...appPages].tsx
+++ b/packages/app-store/routing-forms/pages/route-builder/[...appPages].tsx
@@ -142,7 +142,8 @@ const Route = ({
 
   useEffect(() => {
     if (!isLoading) {
-      const isCustom = !eventOptions.find((eventOption) => eventOption.value === route.action.value);
+      const isCustom =
+        !isRouter(route) && !eventOptions.find((eventOption) => eventOption.value === route.action.value);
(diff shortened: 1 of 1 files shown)


packages/app-store/routing-forms/pages/route-builder/[...appPages].tsx
```diff
@@ -138,9 +138,14 @@ const Route = ({
...
 
-  const [customEventTypeSlug, setCustomEventTypeSlug] = useState<string>(
-    !isRouter(route) ? route.action.value.split("/").pop() ?? "" : ""
-  );
+  const [customEventTypeSlug, setCustomEventTypeSlug] = useState<string>("");
+
+  useEffect(() => {
+    if (!isLoading) {
+      const isCustom = !eventOptions.find((eventOption) => eventOption.value === route.action.value);
+      setCustomEventTypeSlug(isCustom ? route.action.value.split("/").pop() ?? "" : "");
```
Comment (CarinaWolli) [inline packages/app-store/routing-forms/pages/route-builder/[...appPages].tsx:146]: only set the `CustomEventTypeSlug`, when the event type doesn't exist in the `eventOptions`

PeerRich (approved):

PR #13675, opened 2024-02-14
fix: Organization Migration - Prevent the redirected username from being claimed by others.

### LINKED ISSUE #13589: [CAL-3092] User able to take the migrated username in global namespace
We should prevent it like we do in cal.com/signup by checking for tempOrgRedirect entries

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3092](https://linear.app/calcom/issue/CAL-3092/user-able-to-take-the-migrated-username-in-global-namespace)</sub>

hariombalhara:
## What does this PR do?

Fixes #13589

## Type of change

## How should this be tested?
image

Repeat the tests with calcomHandler(By setting `IS_PREMIUM_USERNAME_ENABLED=1` in constants.ts)

## Mandatory Tasks

## Checklist


packages/features/auth/signup/handlers/selfHostedHandler.ts
```diff
@@ -72,9 +74,26 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+          metadata: team.metadata,
+        },
+      });
+      const isInviteForATeamInOrganization = team.parent;
+      const isCheckingUsernameInGlobalNamespace = !isInviteForOrganization && !isInviteForATeamInOrganization;
+
       const teamMetadata = teamMetadataSchema.parse(team?.metadata);
 
+      if (isCheckingUsernameInGlobalNamespace) {
+        const isUsernameAvailable = !(await isUsernameReservedDueToMigration(correctedUsername));
```
hariombalhara, line 90: Redirect is only applicable for global namespace and not inside an organization namespace.

packages/lib/server/checkUsername.ts
```diff
@@ -3,4 +3,6 @@ import { IS_PREMIUM_USERNAME_ENABLED } from "@calcom/lib/constants";
 import { checkRegularUsername } from "./checkRegularUsername";
 import { usernameCheck as checkPremiumUsername } from "./username";
 
+// FIXME: Why don't we use checkPremiumUsername from "@calcom/features/ee/common/lib/checkPremiumUsername" here?
```
hariombalhara, line 6: Discussed with @sean-brydon. We should replace lib/checkPremiumUsername  with `usernameCheck` which completely removes dependency on website's `username` endpoint from the app.

packages/lib/server/username.ts
```diff
@@ -106,7 +109,9 @@ const usernameHandler =
     return handler(req, res);
   };
 
-const usernameCheck = async (usernameRaw: string) => {
+const usernameCheck = async (usernameRaw: string, currentOrgDomain?: string | null) => {
```
joeauyeung, line 112: If we're going to use this function instead of `checkPremiumUsername` can we add some unit tests?

joeauyeung (commented):
Overall this PR makes sense. I'm just wondering when does a username become available in the global space once it's been migrated over to an org?

hariombalhara:
> Overall this PR makes sense. I'm just wondering when does a username become available in the global space once it's been migrated over to an org?

It is right now applicable to hundreds of users that were migrated before the new org schema.

packages/lib/server/username.ts
hariombalhara: Haven't started using it as a replacement of checkPremiumUsername. That would be a followup thing. Would certainly need some unit tests when that happens

joeauyeung (commented):
I was going to approve but @hariombalhara do you know if we have a test where there isn't a clash between a username that is in the global space and in an org after the migration is complete. What about between orgs as well?

hariombalhara:
Didn't get your question fully? Could you share an example maybe of what you are asking

joeauyeung (dismissed):
@hariombalhara I see the test `"Can create a booking for user with same username in and outside organization"` in `organization/booking.e2e.ts` that answers my question. 

Great work, just has some merge conflicts. Retag me for review and I'll approve again after those are fixed.

keithwillcode:
@joeauyeung I fixed the merged conflicts.

packages/features/auth/signup/handlers/selfHostedHandler.ts
```diff
@@ -78,7 +80,26 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
     if (team) {
+      const isInviteForOrganization = isOrganization({
+        team: {
+          metadata: team.metadata,
+        },
+      });
+      const isInviteForATeamInOrganization = team.parent;
+      const isCheckingUsernameInGlobalNamespace = !isInviteForOrganization && !isInviteForATeamInOrganization;
+
+      const teamMetadata = teamMetadataSchema.parse(team?.metadata);
```
joeauyeung: Getting a linting warning this isn't being used so let's remove it.

```suggestion
```

packages/features/auth/signup/handlers/selfHostedHandler.ts
```diff
@@ -78,7 +80,26 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         organizationSettings: true,
       },
     });
+
     if (team) {
+      const isInviteForOrganization = isOrganization({
+        team: {
+          metadata: team.metadata,
+        },
+      });
```
joeauyeung, line 89: Since #13700 added the `isOrganization` property to the team record. Can we remove selecting `organizationSettings` from the team query and just select the `isOrganization` field under the parent?

joeauyeung (commented):
@keithwillcode thoughts?

packages/features/auth/signup/handlers/selfHostedHandler.ts
keithwillcode: Done

packages/features/auth/signup/handlers/selfHostedHandler.ts
keithwillcode: Inner functions that get called still need the settings but I've changed the logic to use `isOrganization` in a clean way.

joeauyeung (dismissed):
Code makes sense and tests are passing. Approving this PR.

These comments are not blocking. https://github.com/calcom/cal.com/pull/13675#pullrequestreview-1917294038

joeauyeung (approved):

PR #13680, opened 2024-02-14
revert: UserPassword

emrysal:
## What does this PR do?

reverts #13628 

Removes UserPassword, we probably have to address our migration history as well to account for an unknown migration in the history.


keithwillcode (approved):

Udit-takkar (approved):
LGTM. works fine now

PR #13685, opened 2024-02-14
revert: Revert UserPassword (#13680)

zomars:
This reverts commit 0a68c6d

## What does this PR do?

amends #13628 

There was an edge case where a user may not have a password set and use CAL identity provider. Causing to fail setting up a password. This PR amends the logic so we make sure to "upsert" every time instead of just "updating".

### DIFF
--- a/apps/web/pages/api/auth/changepw.ts
+++ b/apps/web/pages/api/auth/changepw.ts
@@ -38,7 +38,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   const oldPassword = req.body.oldPassword;
   const newPassword = req.body.newPassword;
 
-  const currentPassword = user.password;
+  const currentPassword = user.password?.hash;
   if (!currentPassword) {
     return res.status(400).json({ error: ErrorCode.UserMissingPassword });
   }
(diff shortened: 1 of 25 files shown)


apps/web/pages/api/auth/changepw.ts
```diff
@@ -53,12 +53,16 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   }
 
   const hashedPassword = await hashPassword(newPassword);
-  await prisma.user.update({
+  await prisma.userPassword.upsert({
     where: {
```
zomars, line 57: This is ht main difference from the original PR

packages/trpc/server/routers/viewer/auth/changePassword.handler.ts
```diff
@@ -54,12 +49,16 @@ export const changePasswordHandler = async ({ input, ctx }: ChangePasswordOption
...
+    create: {
+      hash: hashedPassword,
+      userId: user.id,
     },
-    data: {
-      password: hashedPassword,
+    update: {
+      hash: hashedPassword,
     },
   });
```
zomars, line 63: This is the main difference from the original PR

packages/trpc/server/routers/viewer/organizations/setPassword.handler.ts
```diff
@@ -44,12 +45,16 @@ export const setPasswordHandler = async ({ ctx, input }: UpdateOptions) => {
     });
 
   const hashedPassword = await hashPassword(newPassword);
-  await prisma.user.update({
+  await prisma.userPassword.upsert({
```
zomars, line 48: Same here

zomars (commented):
Self review done

emrysal (approved):
👍  - looks good to me now, trust your diagnosis also

PR #13686, opened 2024-02-14
fix: event type redirect for orgs in preview test (routing-forms)

### LINKED ISSUE #13669: [CAL-3126] Preview test from routing forms redirects to non-org url (event redirect)
event link doesn't redirect to the org URL and it shows 404

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3126](https://linear.app/calcom/issue/CAL-3126/preview-test-from-routing-forms-redirects-to-non-org-url-event)</sub>

CarinaWolli:
## What does this PR do?
Fixes that clicking on the event URL didn't redirect to the correct org URL. 
Screenshot 2024-02-14 at 11 57 14 AM

Fixes #13669


packages/app-store/routing-forms/components/getServerSidePropsSingleForm.ts
```diff
@@ -67,10 +89,32 @@ export const getServerSidePropsForSingleFormView = async function getServerSideP
     };
   }
 
+  const { user: u, ...formWithoutUser } = form;
+
+  const formWithoutProfilInfo = {
```
joeauyeung, line 94:
```suggestion
  const formWithoutProfileInfo = {
```

joeauyeung (changes requested):
When testing this it still points to the non org link for both the preview and the form it self.

I was taking a look at this but before I dive into some more are you also seeing the same thing?

CarinaWolli:
> When testing this it still points to the non org link for both the preview and the form it self.
> 
> I was taking a look at this but before I dive into some more are you also seeing the same thing?

Works for me. So if you create a routing form for an org user with event redirect it redirects you to a 404 page?

joeauyeung (dismissed):
Tested again and it's working now. Odd it wasn't before.

PeerRich (approved):
approved on behalf of @joeauyeung

PR #13690, opened 2024-02-14
chore: removed org link for avatar, fixed tooltip

### LINKED ISSUE #13689: [CAL-3142] remove org link alltogether
keep privacy for orgs

<sub>[CAL-3142](https://linear.app/calcom/issue/CAL-3142/remove-org-link-alltogether)</sub>

PeerRich:
to protect privacy, lets not link to the org page here


Udit-takkar (changes requested):
Screenshot 2024-02-15 at 1 47 01 PM

I can still see booking url of user

Udit-takkar (commented):
We also send the username of all users in viewer.public.event endpoint so it can be accessed from network tab

PR #13694, opened 2024-02-14
chore: adds configurable post migration scripts

zomars:
## What does this PR do?

There have been multiple releases where a new user table is being added to the DB. Due to our custom DB schema that we use across apps we need to run some custom SQL queries that are not relevant to self-hosts:

```sql
GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO DATABASE_SPECIFIC_USER;
GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO DATABASE_SPECIFIC_USER;
```

This PR allows to set scripts like these to be run post-migration to ensure db users have the proper permissions and prevent permission denied errors in production.

## Type of change

## How should this be tested?

- Since this is a production-specific issue we could test by running a migration by adding a new table in Canary and ensure users have permission to read/write to it.

## Mandatory Tasks


packages/prisma/scripts/post-migration.sql
zomars, line 1: This can be any SQL script

packages/prisma/package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "scripts": {
     "clean": "rm -rf .turbo && rm -rf node_modules",
-    "build": "yarn prisma migrate deploy || true",
+    "build": "yarn prisma migrate deploy && yarn post-build || true",
+    "post-build": "yarn prisma db execute --file scripts/post-migration.sql",
```
keithwillcode, line 8: Does this mean that if you set the env variable it will run the script every single time migrations are run? 

In our case, as you mentioned in the description, it’s only necessary sometimes. I think this should take that in to account.

zomars: I think that would make the logic unnecessarily more complex as we would need to check if the schema changed in a way that a new column was added.

Running the script multiple times doesn't yield side effect if there's no new tables are added

keithwillcode: It’s just heavy-handed to grant all new permissions on all tables in the database for no reason when there aren’t new tables added. I like the idea of automating this but I really feel this isn’t the correct approach.

zomars: I'm open to suggestions. We could orchestrate all of this from gh actions themselves. Although that would increase the chance of having a larger timeframe where the DB is de-synced from the code. Just trying to be a close to the metal as possible.

keithwillcode: We write a trigger that runs when create table occurs in the db and we run this script then and only then.

PR #13696, opened 2024-02-14
chore: Platform get schedule

exception:
### Description

Implements the `getAvailableSlots` handler exported by the `@calcom/platform-libraries` package.

### DIFF
--- a/apps/api/v2/src/modules/slots/inputs/get-available-slots.input.ts
+++ b/apps/api/v2/src/modules/slots/inputs/get-available-slots.input.ts
@@ -1,12 +1,4 @@
-import {
-  IS_ISO8601,
-  IsArray,
-  IsBoolean,
-  IsDateString,
-  IsNumber,
-  IsOptional,
-  IsString,
-} from "class-validator";
+import { IsArray, IsBoolean, IsDateString, IsNumber, IsOptional, IsString } from "class-validator";
 
 export class GetAvailableSlotsInput {
   @IsDateString()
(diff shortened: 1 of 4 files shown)


apps/api/v2/src/modules/slots/inputs/get-available-slots.input.ts
```diff
@@ -0,0 +1,36 @@
...
+  @IsBoolean()
+  @IsOptional()
+  debug?: boolean;
+
+  @IsNumber()
+  @IsOptional()
+  duration?: number;
+
+  @IsBoolean()
+  isTeamEvent!: boolean;
```
keithwillcode, line 35: Curious why the client would have to specify if it’s a team event or not?

keithwillcode: It feels weird that the person calling it needs to know that versus just passing the eventTypeId and we figure it out

exception: That makes sense, I'll change logic for that.

apps/api/v2/src/modules/slots/services/slots.service.ts
```diff
@@ -40,4 +40,13 @@ export class SlotsService {
 
     return uid;
   }
+
+  async checkIfIsTeamEvent(eventTypeId?: number) {
+    if (!eventTypeId) return false;
+
+    const event = await this.eventTypeRepo.getEventTypeById(eventTypeId);
+    if (event?.teamId) return true;
```
keithwillcode, line 48: Nit: can be 1 line of code since you are evaluating a Boolean to return a Boolean

ThyMinimalDev (approved):

PR #13706, opened 2024-02-15
fix: password reset when provider is changed

Udit-takkar:
## What does this PR do?

Fixes 

image

How to reproduce the bug?

1) Login in with google

2) Change your email which would send password reset email and change your provider from google to CAL

### DIFF
--- a/apps/web/pages/api/auth/reset-password.ts
+++ b/apps/web/pages/api/auth/reset-password.ts
@@ -51,7 +51,6 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
       },
     });
   } catch (e) {
-    console.error("Error updating password", e);
     return res.status(404).end();
   }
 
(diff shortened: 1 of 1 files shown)


apps/web/pages/api/auth/reset-password.ts
```diff
@@ -42,14 +42,16 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+          upsert: {
+            create: { hash: hashedPassword },
+            update: { hash: hashedPassword },
           },
         },
         emailVerified: new Date(),
       },
     });
   } catch (e) {
+    console.error("Error updating password", e);
```
ThyMinimalDev, line 54: the error is not necessarily related to updating the password, could also fail if the email is not found

emrysal, line 54: This has the potential to thoroughly spam our logs if someone bruteforces the password reset

Udit-takkar: or add checkRateLimitAndThrowError

ThyMinimalDev (approved):

PR #13734, opened 2024-02-16
chore: Refactor feature flags

exception:
## What does this PR do?

Refactors `getFeatureFlagMap` into a more concise and single-use `getFeatureFlag`, caching the whole feature flag map seems unnecessary as most of these will not be used all at once.


packages/features/flags/server/utils.ts
```diff
@@ -12,3 +12,14 @@ export async function getFeatureFlagMap(prisma: PrismaClient) {
     return acc;
   }, {} as Partial<AppFlags>);
 }
+
+export const getFeatureFlag = async (prisma: PrismaClient, slug: keyof AppFlags): Promise<boolean> => {
+  const flag = await prisma.feature.findUnique({
+    where: {
+      slug,
+    },
+    cacheStrategy: { swr: 300, ttl: 300 },
```
keithwillcode, line 21: We can actually remove this since we are not using accelerate

apps/api/v2/src/modules/slots/controllers/slots.controller.e2e-spec.ts
```diff
@@ -0,0 +1,28 @@
+import { bootstrap } from "@/app";
```
keithwillcode, line 1: This file supposed to be in another branch?

exception: Thats... interesting. Yea.

zomars:
Since feature flags are reused everywhere I'd rather cache them all than every single one. There are not that many records and IMO it's better to have this data from a single reusable query than multiple ones

keithwillcode:
> Since feature flags are reused everywhere I'd rather cache them all than every single one. There are not that many records and IMO it's better to have this data from a single reusable query than multiple ones

Fair point

keithwillcode:
@exception Why don’t we cache all but still provide the function to get 1. I liked the refactor you had done. Instead of exposing the full map to callers, they still just have a function that says yes or no to whether the flag they need is enabled.

exception:
> @exception Why don’t we cache all but still provide the function to get 1. I liked the refactor you had done. Instead of exposing the full map to callers, they still just have a function that says yes or no to whether the flag they need is enabled. 

Can do, will refactor util!

packages/features/flags/server/utils.ts
```diff
@@ -12,3 +12,59 @@ export async function getFeatureFlagMap(prisma: PrismaClient) {
...
+      featureFlagCache.set(flag.slug as keyof AppFlags, {
+        value: flag.enabled,
+        expiry: Date.now() + options.ttl,
+      });
+    });
+  }
+
+  const cacheEntry = featureFlagCache.get(slug);
+
+  // Check if the flag is in the cache and not expired
```
keithwillcode, line 50: Nit: This comment is not needed because the line of code it's describing is well-written and self-explanatory

packages/features/flags/server/utils.ts
```diff
@@ -12,3 +12,59 @@ export async function getFeatureFlagMap(prisma: PrismaClient) {
...
+  if (cacheEntry && !isExpired(cacheEntry)) {
+    return cacheEntry.value;
+  }
+
+  // Fetch from the database if not in cache or cache is expired
```
keithwillcode, line 55: Nit: This comment is not needed because the line of code it's describing is well-written and self-explanatory based on the context and code directly above

zomars (dismissed):
Discarded unrelated changes. And LGTM!

zomars (dismissed):
We're getting a unit test failure 

image

keithwillcode (approved):

PR #13735, opened 2024-02-16
fix: Update credentialId when null for a given selectedCalendar

emrysal:
## What does this PR do?

When a credential is created - when a selectedCalendar exists with credentialId null; instead of failing as before; we find the orphaned selectedCalendar and connect it to the new credential.


packages/app-store/googlecalendar/api/callback.ts
```diff
@@ -105,23 +106,32 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
...
         },
       });
     } catch (error) {
-      await prisma.credential.delete({ where: { id: credential.id } });
       let errorMessage = "something_went_wrong";
       if (error instanceof Prisma.PrismaClientKnownRequestError && error.code === "P2002") {
+        // it is possible a selectedCalendar was orphaned, in this situation-
+        // we want to recover by connecting the existing selectedCalendar to the new Credential.
+        if (await renewSelectedCalendarCredentialId(selectedCalendarWhereUnique, credential.id)) return;
+        // else
```
Udit-takkar, line 131: I tried removing replicating this by editing credentialId to null  in 'Selected Calendar' Table. 

Now when i tried connecting the same calendar again I am stuck on this screen
probably because of this return statement.

Screenshot 2024-02-20 at 2 53 35 PM


should it be 
```

 res.redirect(
    getSafeRedirectUrl(state?.returnTo) ??
      getInstalledAppPath({ variant: "calendar", slug: "google-calendar" })
  );

```

emrysal: Oh yeah because it doesn't return there's no return path!

Udit-takkar (dismissed):

Udit-takkar (approved):

Issue #13709, opened 2024-02-15
[CAL-3148] Booking an event in US timezone leads to "Could not book this meeting error"


milospuac:
[https://www.loom.com/share/3845fcb170924978be3222007d4f6713?sid=fdf90a5f-3617-4db9-9ca8-ede2a4fc51de](https://www.loom.com/share/3845fcb170924978be3222007d4f6713?sid=fdf90a5f-3617-4db9-9ca8-ede2a4fc51de)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3148](https://linear.app/calcom/issue/CAL-3148/booking-an-event-in-us-timezone-leads-to-could-not-book-this-meeting)</sub>

keithwillcode:
Wow. Strange. I’m focused on getting rid of the “no available users” error so will look at this.

PR #13738, opened 2024-02-16
fix: No available users found

### LINKED ISSUE #13311: Remove No users found error

Screenshot 2024-01-19 at 12 18 04 AM

To the end user this error doesn't really tell much about what's wrong and why user is not found.

https://threads.com/thread/34561253728/34561258175?s=cetaiy3DJ4ESFWeh9D7Vjq

 Investigate further when this error occurs and display better error message like "time slot conflicting with calendar event"

keithwillcode:
## What does this PR do?

Fixes #13311, #13747, #13709

## Type of change

## How should this be tested?

- ensure test suite passes
- ensure that you can book 1 individual slot on a future date that is only available because a date override has been set 

## Mandatory Tasks


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -339,15 +339,25 @@ export async function ensureAvailableUsers(
     });
   }
 
+  const getStartTime = (startTimeInput: string, timeZone?: string) => {
```
keithwillcode, line 342: This code was taken from `packages/trpc/server/routers/viewers/slots/util.ts` so that we mirror how the `getUserAvailability` function is called.

I noticed that the `.utc()` function was being call for the `endTime` but not the `startTime`. Have added that here and will look to see if `util.ts` needs it to be added as well.

MehulZR: Is the startTimeInput in Zulu format or something like `2024-10-1T05:00` ?

MehulZR: If it isn't in Zulu format or ISO format. It reads the time in local tz. So if you run `dayjs("2024-10-01T05:00")` in India.

1. It will be `2024-10-01T05:00:00+05:30`. 
2. Then we convert it into the specified timezone.
3. And then to UTC. 

Which I guess isn't what we want.

I think here we are thinking of reading `2024-10-01T05:00` in the specified timezone.
For that you need to do something like `dayjs.tz(2024-10-01T05:00, timeZone)`

MehulZR: And I guess github actions are run in UTC tz, so It won't show any errors. And the productions server will be fine to as long as it is in UTC too.

keithwillcode: Yeah this is exactly the problem. This function keeps the date in the time zone specified by the input but along the path of checking availability, it's never converted to UTC so time slots don't line up, resulting in "No available users" errors

keithwillcode: It's in ISO format with the timezone information included. So, for example, if on the booking page you select New York for the time zone, it will pass `2024-02-17T12:00:00-05:00`.

keithwillcode: The main problem is that we were calling `stringToDayjs` with this string and it was never been converted to UTC before doing all the appropriate checks and so the time slots weren't lining up properly, resulting in "No available users found" errors.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -331,8 +339,8 @@ export async function ensureAvailableUsers(
     busyTimesFromLimitsBookingsAllUsers = await getBusyTimesForLimitChecks({
       userIds: eventType.users.map((u) => u.id),
       eventTypeId: eventType.id,
-      startDate: input.dateFrom,
-      endDate: input.dateTo,
+      startDate: startTime.format(),
+      endDate: endTime.format(),
```
keithwillcode, line 343: Need to do more testing but this could be potentially why limits checks have been weird/buggy. This value was being sent in a local date format

MehulZR: `dayjs().format()` can also be local tz.

MehulZR: Unless we really need a specific tz format. It's generally better to use `dayjs.utc()`. It is explicit and anyone working on it can instantly get the tz the date is in.

keithwillcode: This is now using .utc()

packages/features/bookings/lib/handleNewBooking/test/date-overrides.test.ts
```diff
@@ -0,0 +1,207 @@
...
+export type CustomNextApiResponse = NextApiResponse & Response;
+// Local test runs sometime gets too slow
+const timeout = process.env.CI ? 5000 : 20000;
+
+describe("handleNewBooking", () => {
+  setupAndTeardown();
+
+  describe("Booking for slot only available by date override:", () => {
+    test(
+      `should be able to create a booking for the exact slot overridden`,
```
keithwillcode, line 47: First wrote this test against `main` and saw it fail. Then made it pass in this branch.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -315,6 +315,14 @@ export async function ensureAvailableUsers(
   loggerWithEventDetails: Logger<unknown>
 ) {
   const availableUsers: IsFixedAwareUser[] = [];
+  const getStartDateTimeUtc = (startTimeInput: string, timeZone?: string) => {
```
zomars, line 318: Shouldn't it be: 

```suggestion
  const getStartDateTimeUtc = (startDateTimeInput: string, timeZone?: string) => {
```

zomars (dismissed):
Very nice!

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -374,9 +384,8 @@ export async function ensureAvailableUsers(
     //check if event time is within the date range
     for (const dateRange of dateRanges) {
       if (
-        (dayjs.utc(input.dateFrom).isAfter(dateRange.start) ||
-          dayjs.utc(input.dateFrom).isSame(dateRange.start)) &&
-        (dayjs.utc(input.dateTo).isBefore(dateRange.end) || dayjs.utc(input.dateTo).isSame(dateRange.end))
+        (startDateTimeUtc.isAfter(dateRange.start) || startDateTimeUtc.isSame(dateRange.start)) &&
+        (endDateTimeUtc.isBefore(dateRange.end) || endDateTimeUtc.isSame(dateRange.end))
```
CarinaWolli, line 388: Found another `No available users found` error (not new, but sill exists in this branch)

When booking the slot before midnight it is throwing 'No available users found'
dateRange.end is 23:29 but endDateTimeUtc is 00:00 (so `endDateTimeUtc.isSame(dateRange.end) ` is false but should be true)

CarinaWolli (commented):
Fix looks good, did a lot of testing and everything worked for me. When testing I found one more case were this error is thrown, I left a comment

zomars (approved):

PR #13739, opened 2024-02-17
fix: zoho installaltion

### LINKED ISSUE #12405: [CAL-2733] Zoho calendar issues
Busy events not checked or not even able to connect it

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2733](https://linear.app/calcom/issue/CAL-2733/zoho-calendar-issues)</sub>

### LINKED ISSUE #12955: [CAL-2842] Zoho Calendar integration results in 500 server error
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2842](https://linear.app/calcom/issue/CAL-2842/zoho-calendar-integration-results-in-500-server-error)</sub>

Pritam-sd-dev:
## What does this PR do?

This PR fixes error that we get on zohocalender installation.

Fixes # (issue)
fixes #12955 
fixes #12405 

Before
https://www.loom.com/share/b82a551a2a2c4b11ad6aefdcc55104e9?sid=d2a638a4-ad7f-4670-8a77-0edab4f7c3ff
After
https://www.loom.com/share/06ad87f1c8ee4e30a44753daea034f67?sid=478e9091-9028-418e-8783-4592e94fc3d2

## Type of change

## How should this be tested?

Changes of this PR can be verified by installing zohocalender for different region's accounts like US, IN, EU etc


joeauyeung:
Thanks for your contribution @Pritam-sd-dev. I'm reviewing your PR and I'm trying to understand how this PR would solve those issues. 

From what I see the `serverDomain` is something we would set on our side. If the server is in the US and stays as `.com` how would that address the issue?

Pritam-sd-dev:
The first api that we hit returns the user's domain

Pritam-sd-dev:
For more details, you can refer to zoho calendar docs

Pritam-sd-dev:
@joeauyeung 
You can refer to https://www.zoho.com/crm/developer/docs/api/v6/multi-dc.html
When we make Auth request in response we get a value of server.

joeauyeung (changes requested):
@Pritam-sd-dev thanks for the clarification. 

If each user has a different server domain I think it would it be better to store that information under `credential.key` as that's unique to each user rather than `app` as that data is used for every user.

Pritam-sd-dev:
But what if user wants to have multiple calendars with different domain ?

joeauyeung:
> But what if user wants to have multiple calendars with different domain ?

That's a good point. If a user has a calendar with a different domain is that also a different Zoho account? Every time a Cal.com user goes through the Zoho OAuth flow we create a different credential so we can still store the domain data under the `key` field of each credential.

joeauyeung:
Closing in favor of #14905

PR #13742, opened 2024-02-18
fix: correct German translation

mokkin:
## What does this PR do?

correct German translation

## Type of change

- translation

## How should this be tested?

## Mandatory Tasks

### DIFF
--- a/apps/web/public/static/locales/de/common.json
+++ b/apps/web/public/static/locales/de/common.json
@@ -607,7 +607,7 @@
   "danger_zone": "Achtung",
   "account_deletion_cannot_be_undone": "Vorsicht. Löschen eines Kontos kann nicht rückgängig gemacht werden.",
   "back": "Zurück",
-  "cancel": "Absagen",
+  "cancel": "Abbrechen",
   "cancel_all_remaining": "Alle verbleibenden absagen",
   "apply": "Anwenden",
   "cancel_event": "Termin absagen",


apps/web/public/static/locales/de/common.json
```diff
@@ -607,7 +607,7 @@
   "danger_zone": "Achtung",
   "account_deletion_cannot_be_undone": "Vorsicht. Löschen eines Kontos kann nicht rückgängig gemacht werden.",
   "back": "Zurück",
-  "cancel": "Absagen",
+  "cancel": "Abbrechen",
```
Ryukemeister, line 610: @mokkin don't both of those words have the same meaning?

Ryukemeister: in that case @PeerRich you might have more context around here

PeerRich: here's the problem, we need to introduce two types of "cancel"

because we also use "Cancel" in a different context for "Cancel appointment" which where "Meeting absagen" is correct.

which means we need to introduce a new variable:


"cancel": "Abbrechen",
"cancel_action": "Abbrechen",

and then use the cancel_action in all places where cancel needs to be "Abbrechen"

PeerRich: but i think we can merge this, but i will need to watch out where we use "Abbrechen" correclty

PeerRich (approved):

apps/web/public/static/locales/de/common.json
mokkin: > here's the problem, we need to introduce two types of "cancel"
> 
> because we also use "Cancel" in a different context for "Cancel appointment" which where "Meeting absagen" is correct.
> 
> which means we need to introduce a new variable:
> 
> "cancel": "Abbrechen", "cancel_action": "Abbrechen",
> 
> and then use the cancel_action in all places where cancel needs to be "Abbrechen"

That's right! The "cancel" I was referring to is the "cancel_action" such as in every dialogue like here in LibreOffice for example at no. 2:
grafik

Issue #13748, opened 2024-02-19
scroll bar that follows the design pattern of website 


DipakKhade:
### Description
adding a scroll bar that follows the design pattern of website  i.e  black scrollbar-thumb and some sort of white scrollbar-track

PR #13749, opened 2024-02-19
feat: scroll bar that follows the design pattern of website

DipakKhade:
## What does this PR do?
Feature #13748 

this PR add few line of code to style the scroll bar  which follows the design and color pattern of website
i have tested it locally , you can confirm it

### DIFF
--- a/packages/atoms/globals.css
+++ b/packages/atoms/globals.css
@@ -3,8 +3,21 @@
  * When using this component in any Cal project, the globals are automatically imported
  * in that project.
  */
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+ @tailwind base;
+ @tailwind components;
+ @tailwind utilities;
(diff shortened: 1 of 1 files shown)


packages/atoms/globals.css
```diff
@@ -3,8 +3,21 @@
...
+ @import "../ui/styles/shared-globals.css";
 
-@import "../ui/styles/shared-globals.css";
+ ::-webkit-scrollbar{
+    width: 10px; 
+  }
+  
+  ::-webkit-scrollbar-track{
+    background-color: aliceblue;
+  }
```
Udit-takkar, line 18: This property is not compatible with all browsers. (Firefox, Safari for mobile)

https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

PR #12633, opened 2023-12-02
feat: Select darkmode in app.cal.com

swh00tw:
## What does this PR do?

Add a section in the appearance page to switch dark/light mode for the Main page(not the booking page).

Solved #11997

https://github.com/calcom/cal.com/assets/71842426/6c165617-283e-4aee-9845-8b3e660985d2

## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is a UI/UX design document, please, share it here.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How should this be tested?

1. Go to `localhost:3000`
2. Go to your setting page's appearance tab
3.  Switch the "App theme", and you will see the color change of the UI. (Cal.com's system default is dark mode)

## Mandatory Tasks

- [x] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Note

Currently, for the wording here, it's only in English for now (I've asked a person from Cal.com in the OSS hackathon, and he said using English for now is good). We need to create another PR to support all other languages.

Screenshot 2023-12-02 at 12 12 24 PM

PR #13779, opened 2024-02-20
feat: Settings for dashboard theme that is separate from system 

### LINKED ISSUE #11997: No separate option to enable dark mode
Yes
  

(Write your answer here.)
So almost every website or saas these days have a separate button within there UI to enable dark mode it's a very small but big thing that effects usability for the service because let's say I only want dark mode in notion or cal.com but not on twitter or Instagram. By default it would force dark mode if I changed my device theme to dark and all other websites where I don't want to see dark mode but without doing it cal.com wouldn't change to dark mode which is a problem and I think it should fixed it's a small thing but does effect usability. 
Add a separate option to change cal.com theme to dark mode in settings or main UI without changing my device theme. 

(Describe your proposed solution here.)
Written above
There's no alternative whatsoever I have to change my device theme to dark then only it works no other way. 

(Write your answer here.)

### Additional context

(Write your answer here.)

### Requirement/Document

(Share it here.)

sean-brydon:
Took over from : https://github.com/calcom/cal.com/pull/12633

Fixes: #11997 CAL-2761

Most of the credit goes to @swh00tw:  I couldnt push changes to your PR #12633  so i just thought it'd be easier to push my own. 

Loom: https://github.com/calcom/cal.com/assets/55134778/a3537270-654c-4370-a7f9-a9f9581e0381


Udit-takkar (approved):
LGTM. Great work @sean-brydon and @swh00tw

PR #13785, opened 2024-02-20
feat: platform use slots hooks

ThyMinimalDev:
## What does this PR do?

add platform hooks for slots


packages/platform/types/slots.ts
```diff
@@ -23,3 +32,24 @@ export class GetAvailableSlotsInput {
...
+  @IsOptional()
+  uid?: string;
+}
+
+export class ReserveSlotInput {
+  @IsInt()
+  eventTypeId!: number;
+
+  @IsDate()
+  slotUtcStartDate!: string;
```
keithwillcode, line 47: Suggesting we add `Time` onto this to follow our new naming standards 
```suggestion
  slotUtcStartDateTime!: string;
```

packages/platform/types/slots.ts
```diff
@@ -23,3 +32,24 @@ export class GetAvailableSlotsInput {
...
+
+  @IsDate()
+  slotUtcEndDate!: string;
```
keithwillcode, line 50:
```suggestion
  slotUtcEndDateTime!: string;
```

PR #13797, opened 2024-02-21
feat: platform use create bookings

ThyMinimalDev:
## What does this PR do?

platform booker create bookings hooks


apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -56,10 +53,9 @@ export class BookingsController {
   @Post("/reccuring")
   async createReccuringBooking(
     @Req() req: Request & { userId?: number },
-    @Body() _: CreateReccuringBookingInput[],
-    @GetUser() user: User
+    @Body() _: CreateReccuringBookingInput[]
```
supalarry, line 56: Can we remove @Body() if its not being used "_" here and in other places in controller?

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -30,17 +27,17 @@ import { ApiResponse } from "@calcom/platform-types";
   path: "ee/bookings",
   version: "2",
 })
-@UseGuards(AccessTokenGuard)
```
supalarry, line 33: Wonder if public endpoints should be in a separate folder.

supalarry (approved):
Looks good - just 2 small comments below.

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
ThyMinimalDev: still need to validate the params

PR #13807, opened 2024-02-21
fix: crash on signup page while inviting a new member to an org

ibex088:
## What does this PR do?

Fixes - 
crash on signup page while inviting a new member to an organization
[LOOM](https://www.loom.com/share/c8d6dd3942fa46e8b1cba8e55e4d4aa6?sid=91a3681e-2f71-4d8f-982c-6fd866c16586)


Udit-takkar (approved):
LGTM. Nice catch

PR #13811, opened 2024-02-21
feat: Adding a setting to change the default home view

### LINKED ISSUE #13778: [CAL-3176] Add a setting to change the default home view
User request:

> Friendly request to @calcom team. please make /bookings the default. I use /event-types once or twice. I visit Upcoming Bookings every time

As this is not the case for all users, instead of making it the default, we're going to make it an option that you can set:

[Update default home view - New Features (Figma)](https://www.figma.com/file/UajFu4M1APhn2AywAEJkJr/New-Features?node-id=18174%3A35817&mode=dev)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3176](https://linear.app/calcom/issue/CAL-3176/add-a-setting-to-change-the-default-home-view)</sub>

Shyam-Raghuwanshi:
## What does this PR do?

this PR Add a setting to change the default home view

Fixes #13778

Loom Video

https://www.loom.com/share/2f7b521aa35c4b2fb669bcc6523c6ea9?sid=4f2f7e8f-49b6-4698-9ef8-ed00b5b980f9


apps/web/pages/index.tsx
```diff
@@ -13,7 +13,7 @@ export async function getServerSideProps({ req, res }: GetServerSidePropsContext
     return { redirect: { permanent: false, destination: "/auth/login" } };
   }
 
-  return { redirect: { permanent: false, destination: "/event-types" } };
+  return { redirect: { permanent: false, destination: `/${session.user.homeView}` || "event-types" } };
```
ajitzero, line 16: Minor: even if empty, the first value would at least be `/`, so the fallback value won't be reached with this approach.

Maybe moves it inside the `${...}` ?

packages/prisma/schema.prisma
```diff
@@ -215,6 +215,7 @@ model User {
   avatarUrl           String?
   timeZone            String               @default("Europe/London")
   weekStart           String               @default("Sunday")
+  homeView            String               @default("event-types")
```
ajitzero, line 218: nit: This could be an enum for slightly better type-safety

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -220,6 +231,22 @@ const GeneralView = ({ localeProp, user }: GeneralViewProps) => {
               </>
             )}
           />
+          <Controller
+            name="homeView"
+            control={formMethods.control}
+            render={({ field: { value } }) => (
+              <>
+                <Label className="text-emphasis mt-6">Default home view</Label>
```
PeerRich, line 239:
```suggestion
                <Label className="text-emphasis mt-6">{t("default_home_view")}</Label>
```

PeerRich: TODO add: 

"default_home_view ": "Default home view"

to /en/common.json

packages/prisma/migrations/20240221100119_add_home_view_field/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "users" ADD COLUMN     "homeView" TEXT NOT NULL DEFAULT 'event-types';
```
PeerRich, line 2: instead of adding this to the database, what if we just save this setting on localStorage?

seems to do the job and does not pollute the database

PeerRich: its not worth to sync this setting across devices

apps/web/pages/index.tsx
Shyam-Raghuwanshi: yeah you are right

packages/prisma/schema.prisma
Udit-takkar, line 218: This isn't a very good way to store url. Enum would be better

packages/prisma/migrations/20240221100119_add_home_view_field/migration.sql
Shyam-Raghuwanshi: Ok let's do with localStorage

Shyam-Raghuwanshi: may be if you want tell me

PeerRich: yes please use localStorage (there is a hook for it already) and not change the database 🙏

Shyam-Raghuwanshi: Ok

Shyam-Raghuwanshi: can i use cookies for storing the values ?

Shyam-Raghuwanshi: because i can't able to get localStorage in server side component in index.tsx file that redirect the user
`import type { GetServerSidePropsContext } from "next";

import { getServerSession } from "@calcom/features/auth/lib/getServerSession";

function RedirectPage() {
  return;
}

export async function getServerSideProps({ req, res }: GetServerSidePropsContext) {
  const session = await getServerSession({ req, res });

  if (!session?.user?.id) {
    return { redirect: { permanent: false, destination: "/auth/login" } };
  }

  return { redirect: { permanent: false, destination: "/event-types" } };
}

export default RedirectPage;
`

keithwillcode:
Moving to draft because there are merge conflicts, type errors and build errors

keithwillcode:
Closing due to staleness. Feel free to reopen if work continues.

PR #13814, opened 2024-02-21
feat: Platform SDK

exception:
## What does this PR do?

Introduces the `@calcom/sdk` package, used to interact with the new API endpoints and (soon) the v1 API.


apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
```diff
@@ -15,11 +16,14 @@ export class OAuthClientRepository {
...
 
   async createOAuthClient(organizationId: number, data: CreateOAuthClientInput) {
+    const id = "cal_pla_" + randomBytes(28).toString("hex");
+
     return this.dbWrite.prisma.platformOAuthClient.create({
       data: {
         ...data,
         secret: this.jwtService.sign(data),
         organizationId,
+        id,
```
ThyMinimalDev, line 26: wondering the reason behind creating the id manually ?

packages/platform/sdk/src/endpoints/events/event-types/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.assertAccessToken("createEventType");
+
+    const { data } = await this.sdk.httpCaller.post<BasicPlatformResponse<BodyWithEventTypeResponse>>(
+      Endpoints.CREATE_EVENT_TYPE,
+      {
+        body: args,
+      }
+    );
+
+    return data.eventType;
```
ThyMinimalDev, line 26: I think now we return data directly in the eventtype controller

packages/platform/sdk/src/endpoints/events/event-types/types.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const GetEventTypeByIdSchema = z.object({
+  id: z.string(),
+  forAtom: z.boolean().optional(),
+});
+
+export type CreateEventTypeArgs = z.infer<typeof CreateEventTypesSchema>;
+export type GetEventTypeByIdArgs = z.infer<typeof GetEventTypeByIdSchema>;
+
+export type EventType = {
```
ThyMinimalDev, line 17: maybe if we don't reuse types from the monorepo we need to think about adding integration tests for the sdk so that we are informed in CI that the request failed

apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
exception: Was having issues for some reason with Prisma, will revert.

packages/platform/sdk/src/endpoints/events/event-types/types.ts
exception: Yes, integration testing is planned. Was talking to Keith about this yesterday.

PR #13790, opened 2024-02-20
fix: og images avatars

zomars:
## What does this PR do?

Fixes some edge cases for OG avatars not displaying properly

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- Verify generated images are displaying properly in public pages

## Mandatory Tasks

- [ ] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

Issue #13777, opened 2024-02-20
[CAL-3175] og-image broken for website and for app.cal.com


PeerRich:
[https://www.opengraph.xyz/url/https%3A%2F%2Fcal.com%2Fpeer](https://www.opengraph.xyz/url/https%3A%2F%2Fcal.com%2Fpeer)

![image](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/72fa69c4-efb6-42a3-b392-f61aa5695a37/0a833193-c581-43fe-9a6e-e9aa76e5c3e3?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2U4NmJmOTU3LWQ4MmYtNDY1ZS1iMjA1LTEzNTU1OWY0YjYyMy83MmZhNjljNC1lZmI2LTQyYTMtYjM5Mi1mNjFhYTU2OTVhMzcvMGE4MzMxOTMtYzU4MS00M2ZlLTlhNmUtZTlhYTc2ZTVjM2UzIiwiaWF0IjoxNzA4NDIxMDQzLCJleHAiOjE3MDg1MDc0NDN9.r-N_UmJzCUPucqUbqCIctvmgf_XXSLMEmj4WbRrT5qI)

![image](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/c3851f8f-e63d-4061-b9b8-3e043d025d4d/adb61c21-d3f6-4b8d-9483-eddbccabaf78?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2U4NmJmOTU3LWQ4MmYtNDY1ZS1iMjA1LTEzNTU1OWY0YjYyMy9jMzg1MWY4Zi1lNjNkLTQwNjEtYjliOC0zZTA0M2QwMjVkNGQvYWRiNjFjMjEtZDNmNi00YjhkLTk0ODMtZWRkYmNjYWJhZjc4IiwiaWF0IjoxNzA4NDIxMDQzLCJleHAiOjE3MDg1MDc0NDN9.NyCvkKwPdJ9RlT0hUGm9sCtwlbpj5d4yZqQ5t29MPZ0)

[...]

PR #13817, opened 2024-02-22
fix: more og avatar fixes

zomars:
## What does this PR do?

refs #13777 

follow up to #13790 

Hopefully this cover ups and fixes @PeerRich profile edge-case as well

## Type of change

## How should this be tested?

- On preview check OG image is rendered correctly.

## Mandatory Tasks


packages/lib/constants.ts
```diff
@@ -83,7 +83,7 @@ export const SEO_IMG_DEFAULT = `${CAL_URL}/og-image.png`;
 // parameters you pass to the /api/social/og/image endpoint, you wrap them in encodeURIComponent
 // as well, otherwise the URL won't be valid.
 export const SEO_IMG_OGIMG = `${CAL_URL}/_next/image?w=1200&q=100&url=${encodeURIComponent(
-  "/api/social/og/image"
+  `${CAL_URL}/api/social/og/image`
```
emrysal, line 86: So this was it ._.

zomars: part of it

emrysal (approved):
Looks good to me, absolute URL should fix most issues, getUserAvatar the rest

PR #13828, opened 2024-02-22
chore: remove CF-region

Udit-takkar:
## What does this PR do?

Fixes # (issue)

Too many error logs of "Skipping HTTP Request header"

NEXT STEPS:-

Disable this on cloudflare https://threads.com/thread/34538151691/34560437009?s=cutsX49CgqKKNEqfWRL16S (cc: @emrysal )

Screenshot 2024-02-22 at 6 57 14 PM

## Requirement/Documentation

## Type of change


joeauyeung (approved):
Approving since it seems like we're moving away from using Cloudflare for this. Would still need to coordinate disabling this on Cloudflare itself.

PR #13837, opened 2024-02-23
fix: OAuth client form improvements

Ryukemeister:
## What does this PR do?

- Adds multiple inputs and validation for redirect urls
- Clicking on labels toggles check boxes
- Adds button to select all permissions at once
- Add permission for profile read and write


apps/web/public/static/locales/en/common.json
```diff
@@ -2154,6 +2154,8 @@
   "access_bookings": "Read, edit, delete your bookings",
   "allow_client_to_do": "Allow {{clientName}} to do this?",
   "oauth_access_information": "By clicking allow, you allow this app to use your information in accordance with their terms of service and privacy policy. You can remove access in the {{appName}} App Store.",
+  "oauth_form_title":"OAuth client creation form",
+  "oauth_form_description":"This is the form to create a new OAuth client",
```
supalarry, line 2158: The description pretty much repeats the title so I don't see a reason to keep the description.

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
```diff
@@ -128,13 +162,24 @@ export const OAuthClientForm: FC = () => {
...
         </div>
         <div className="mt-6">
-          <h1 className="text-base font-semibold underline">Permissions</h1>
+          <TextField type="url" label="Redirect uri three" {...register("redirectUriThree")} />
+        </div>
+        <div className="mt-6">
+          <div className="flex justify-between">
+            <h1 className="text-base font-semibold underline">Permissions</h1>
+            <Button type="button" onClick={selectAllPermissions}>
+              Select all
```
supalarry, line 177: When all are selected the "Select all" should change to de-select all instead.

Screenshot 2024-02-27 at 10 21 04

I would also change "Submit" to "Create client" text as its more clear in my opinion what the action is.

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
```diff
@@ -128,13 +162,24 @@ export const OAuthClientForm: FC = () => {
...
+            </Button>
+          </div>
           <div>{permissionsCheckboxes}</div>
         </div>
-        <Button className="mt-6" type="submit" loading={isLoading}>
+        <Button className="mt-6" type="submit" loading={isPending}>
           Submit
         </Button>
       </form>
```
supalarry, line 185: Padding left and bottom of form is not equal

Screenshot 2024-02-27 at 10 20 35

supalarry: I would also add more margin to the "Submit" button because otherwise it is closer to permissions that the select all permissions button, so we need to space it away to make it clearer that its a separate action.

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
```diff
@@ -20,13 +23,31 @@ type FormValues = {
...
+  const { register, handleSubmit, setValue, getValues } = useForm<FormValues>({});
+  const [isSelectAllPermissionsChecked, setIsSelectAllPermissionsChecked] = useState(false);
 
-  const { mutateAsync, isLoading } = useCreateOAuthClient({
+  const selectAllPermissions = () => {
+    const allInputValues = getValues();
+
+    Object.keys(allInputValues).forEach((input) => {
+      if (!isSelectAllPermissionsChecked && (input.includes("Read") || input.includes("Write"))) {
+        setValue(input, true);
```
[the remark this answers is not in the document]
Ryukemeister: You're right this is not needed and there we can be a better logic for this, will fix this

apps/web/public/static/locales/en/common.json
Ryukemeister: This is just for future references. Ideally the description and title is same for now but we might have a better description for it in the future right so yeah just for v1

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
Ryukemeister: @supalarry actually the form title and description should let a user know the purpose of the form and that is why imho its better to keep it as submit no?

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
```diff
@@ -38,6 +59,12 @@ export const OAuthClientForm: FC = () => {
 
   const onSubmit = (data: FormValues) => {
     let userPermissions = 0;
+    const userRedirectUris = [data.redirectUriOne];
```
supalarry, line 62: alternative:

const userRedirectUris = [data.redirectUriOne, data.redirectUriTwo, data.redirectUriThree].filter(uri => !!uri);

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
```diff
@@ -25,27 +25,27 @@ type FormValues = {
...
   const { t } = useLocale();
   const router = useRouter();
-  const { register, handleSubmit, setValue, getValues } = useForm<FormValues>({});
+  const { register, handleSubmit, setValue } = useForm<FormValues>({});
   const [isSelectAllPermissionsChecked, setIsSelectAllPermissionsChecked] = useState(false);
 
-  const selectAllPermissions = () => {
-    const allInputValues = getValues();
+  const selectAllPermissions = useCallback(() => {
+    Object.keys(PERMISSIONS_GROUPED_MAP).forEach((key) => {
```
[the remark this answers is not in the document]
supalarry: Personally, i would still split the permissions out because now if some permission's "key" is changed in PERMISSIONS_GROUPED_MAP or in FormValues then it would break but your call.

apps/web/pages/settings/organizations/platform/oauth-clients/components/OAuthClientForm.tsx
```diff
@@ -1,32 +1,61 @@
...
   eventTypeWrite: boolean;
   bookingRead: boolean;
   bookingWrite: boolean;
   scheduleRead: boolean;
   scheduleWrite: boolean;
+  appsRead: boolean;
+  appsWrite: boolean;
+  profileRead: boolean;
+  profileWrite: boolean;
+  redirectUris: {
```
Ryukemeister, line 27: `useFieldArray` hook only accepts an array of objects that's why had to put in uris inside an object in the array

ThyMinimalDev (approved):

PR #13842, opened 2024-02-23
fix: video call url in event location for recurring events

CarinaWolli:
## What does this PR do?

Fixes the video call URL in the location of the calendar event for recurring events. The event description had the correct URL but the event location showed the URL of the last recurring event. 

Now each recurring event is showing a different video call URL in the location that matches the one in the event description. 

Screenshot 2024-02-23 at 4 52 03 PM


sean-brydon (approved):
LGTM: tested and the URL is now correct

PR #13851, opened 2024-02-25
fix: Limit future bookings - Default is 1 day even though on UI it is 30 days

### LINKED ISSUE #13848: Limit future bookings - Default is 1 day even though on UI it is 30 days
### Issue Summary

I enabled the limit future bookings option in my event. On opening the preview link only today's slots were shown as open. The default value is 30 (business) days into the future, so the expectation is that slots for the next 30 days should be open.

### Steps to Reproduce

1. Create a new event
2. Go to "Limits" section on the left bar
3. Enable the "Limit future bookings" option and save
4. Check the preview

### Actual Results

- Only today's slots are open.

### Expected Results

- Slots in the next 30(default option) days should be open

### Technical details

- Chrome, Version 121.0.6167.185 (Official Build) (64-bit)

Pritam-sd-dev:
## What does this PR do?

Fixes #13848 

## Type of change

## How should this be tested?
to test changes of this PR set limit on future books with value 30. and then check booking page whether bookings are available or not.

## Checklist

- I haven't checked if  existing unit tests pass locally with my changes

### DIFF
--- a/apps/web/components/eventtype/EventLimitsTab.tsx
+++ b/apps/web/components/eventtype/EventLimitsTab.tsx
@@ -360,10 +360,12 @@ export const EventLimitsTab = () => {
               title={t("limit_future_bookings")}
               description={t("limit_future_bookings_description")}
               checked={isChecked}
-              onCheckedChange={(bool) =>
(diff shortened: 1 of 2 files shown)


emrysal (approved):
Good improvement, thank you 🙏

PR #13853, opened 2024-02-25
fix: [CAL-3201] All attempts to self-host and run locally following self-host instructions return error and container fails to start up.

### LINKED ISSUE #13836: All attempts to self-host and run locally following self-host instructions return error and container fails to start up.
=

### Issue Summary
Whenever I try to run commands `yarn dx` or `yarn dev`, I get this error (see pictures) and the container fails to start up. I've followed the self-host instructions to the letter. <running on macOS>

### Steps to Reproduce

1. Navigate to cal.com folder.
2. Run `yarn`
3. Create 32-character ssl keys and put them in correct fields in .env duplicated from .env.example.
4. Run yarn dx/yarn dev

### Actual Results

Container fails to spin up and web application fails to run locally.

### Expected Results

Ability to run app locally.

### Technical details

- Node.js version => v18.16.1

### Evidence
Screenshot_2024-02-21_at_9 22 42_PM

Screenshot 2024-02-21 at 9 48 47 PM

khsinat:
## What does this PR do?

Fixed the error that appeared when running locally after doing ```yarn dx```  
it looks like earlier a new userPassword field was created for every case, so i changed the code so as to just update the field when needed 

Fixes #13836 

## Type of change

### DIFF
--- a/packages/prisma/seed-utils.ts
+++ b/packages/prisma/seed-utils.ts
@@ -40,11 +40,6 @@ export async function createUserAndEventType({
 }) {
   const userData = {
     ...user,
-    password: {
-      create: {
-        hash: await hashPassword(user.password),
-      },
-    },
     emailVerified: new Date(),
     completedOnboarding: user.completedOnboarding ?? true,
     locale: "en",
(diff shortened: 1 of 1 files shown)


Udit-takkar (commented):
@khsinat Thanks for your contribution but this has been fixed fixed here https://github.com/calcom/cal.com/pull/13868

PR #13865, opened 2024-02-26
fix: Prevent duplicate bookings with idempotency key

emrysal:
## What does this PR do?

Implements idempotency key

* Tests added to show it is not possible to re-book the same event with the same email(s). Even if it is an event as requiresConfirmation.
* The existing test "can cancel the recently created booking and rebook the same timeslot" proves that a cancellation removes the idempotency key and allows for future booking of the same event at the same time.

### DIFF
--- a/packages/features/bookings/lib/handleNewBooking.ts
+++ b/packages/features/bookings/lib/handleNewBooking.ts
@@ -1694,7 +1694,7 @@ async function handler(
       err.message
     );
     if (err.code === "P2002") {
-      throw new HttpError({ statusCode: 409, message: "booking.conflict" });
+      throw new HttpError({ statusCode: 409, message: "booking_conflict" });
     }
     throw err;
   }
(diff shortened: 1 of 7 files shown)


packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1627,6 +1627,19 @@ async function handler(
     })
   );
 
+  // update original rescheduled booking (no seats event)
+  if (!eventType.seatsPerTimeSlot && originalRescheduledBooking?.uid) {
+    await prisma.booking.update({
```
emrysal, line 1632: The update unsets the idempotencyKey - so it needs to happen before createBooking to prevent a conflict.

packages/prisma/extensions/booking-idempotency-key.ts
zomars, line 1: Love this approach 🙏🏽

zomars (approved):

PR #13874, opened 2024-02-26
chore: use transaction

Udit-takkar:
## What does this PR do?

Fixes: 

## Type of change


packages/trpc/server/routers/viewer/teams/delete.handler.ts
```diff
@@ -23,25 +23,27 @@ export const deleteHandler = async ({ ctx, input }: DeleteOptions) => {
...
+        id: input.teamId,
+      },
+    });
+
+    const deletedTeamMetadata = teamMetadataSchema.parse(deletedTeam.metadata);
+
+    if (deletedTeamMetadata?.isOrganization && deletedTeam.slug) deleteDomain(deletedTeam.slug);
+
+    // Sync Services: Close.cm
+    closeComDeleteTeam(deletedTeam);
```
keithwillcode, line 45: Third party services typically shouldn’t be wrapped in transactions like this. Was this intentional?

keithwillcode (approved):

PR #13876, opened 2024-02-26
fix: remove special characters from email display name

CarinaWolli:
## What does this PR do?

Fixes the following error when email is sent: SEND_BOOKING_CONFIRMATION_ERROR Error: Message failed: 550 The from address does not match a verified Sender Identity

### DIFF
--- a/packages/emails/lib/sanitizeDisplayName.ts
+++ b/packages/emails/lib/sanitizeDisplayName.ts
@@ -1,0 +1,16 @@
+export const sanitizeDisplayName = (nameAndEmail: string) => {
+  const match = nameAndEmail.match(/^(.*?)\s<(.*)>$/);
+
+  if (match) {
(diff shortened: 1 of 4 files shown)


keithwillcode (commented):
Instead of going to multiple places and calling the sanitize function, can we call it one layer deeper in the code right before sending the email so we call it once and never have to remember to call it for new email code written?

packages/emails/lib/sanitizeDisplayName.ts
```diff
@@ -0,0 +1,16 @@
+export const sanitizeDisplayName = (nameAndEmail: string) => {
+  const match = nameAndEmail.match(/^(.*?)\s<(.*)>$/);
```
CarinaWolli, line 2: format of nameAndEmail is `Team pro <teampro@example.com>`

keithwillcode (approved):
Beautiful!

PR #13888, opened 2024-02-27
fix: team event type error when when no one is assigned

### LINKED ISSUE #13922: [CAL-3244] Could not book the meeting "Cannot read properties of undefined (reading 'name')
Details in [threads](https://threads.com/thread/34565738923?s=34S5Wx9KDPRxtJvSnrGHz9).

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3244](https://linear.app/calcom/issue/CAL-3244/could-not-book-the-meeting-cannot-read-properties-of-undefined-reading)</sub>

Udit-takkar:
## What does this PR do?

Fixes # (issue) " Could not book the meeting "Cannot read properties of undefined (reading 'name')"

TODO: 

Need help:-

- [x] Figure out unit test failure 

https://threads.com/thread/34565738923/34565738922?s=34S5Wx9KDPRxtJvSnrGHz9

How to reproduce error? 

1) Create a team event type and remove all assigned members.
2) Go the booking page and create a booking

You would see error: Could not book the meeting "Cannot read properties of undefined (reading 'name')". 

## Type of change


packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
```diff
@@ -83,6 +83,22 @@ export const createHandler = async ({ ctx, input }: CreateOptions) => {
...
+            user: {
+              connect: {
+                id: userId,
+              },
+            },
+            isFixed: data.schedulingType === SchedulingType.COLLECTIVE,
+            priority: 2, //default
+          },
+        ],
+      };
```
Udit-takkar, line 100: Add owner of the event type to host when creating event type

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -320,6 +320,10 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
     usersWithCredentials = eventType.hosts.map(({ isFixed, user }) => ({ isFixed, ...user }));
   }
 
+  if (eventType.schedulingType && !eventType.hosts?.length) {
+    throw new TRPCError({ message: "No user is assigned.", code: "NOT_FOUND" });
+  }
+
```
Udit-takkar, line 326: Earlier no slots would be displayed when no one was assigned to the team event type.

CarinaWolli: There are slots displayed but when booking I am getting the error `Cannot read properties of undefined (reading 'name')`

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
CarinaWolli: not sure if we want that? What if we just don't show any availability when no one is assigned, that would make the most sense to me

CarinaWolli (dismissed):
I am still getting the error, left you some comments

packages/trpc/server/routers/viewer/slots/util.ts
Udit-takkar: Screenshot 2024-02-28 at 10 22 17 AM

@CarinaWolli if there are no hosts then you won't see any slots

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
Udit-takkar: i added this check in util.ts to throw error and not display any slots.

Udit-takkar: this is just to prevent a scenario when no host is assigned when a team event type is created and we already add the owner to users[] field in event type

packages/trpc/server/routers/viewer/slots/util.ts
Udit-takkar: how can i reproduce the bug?

CarinaWolli: > Earlier no slots would be displayed when no one was assigned to the team event type.

I meant before they were displayed because you said earlier there were no slots displayed. Now there are not

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
CarinaWolli: What I meant is, is that I would prefer to not have any host assigned after creating the event type

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -320,6 +320,10 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
     usersWithCredentials = eventType.hosts.map(({ isFixed, user }) => ({ isFixed, ...user }));
   }
 
+  if (eventType.schedulingType && !eventType.hosts?.length) {
```
CarinaWolli, line 323: I think we need to check for `eventType.users` too, I think older event types can have set users instead of hosts. We also take them into account in line 314

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
Udit-takkar: @CarinaWolli yeah but a lot of people get confused when they create team event type and go to booking page and get error Could not book the meeting "Cannot read properties of undefined (reading 'name')

Udit-takkar: and then they contact intercom

CarinaWolli: I think what we should do is, is not having any user assigned when creating by changing [line 60](https://github.com/calcom/cal.com/pull/13888/files#diff-6b49c95fdbabc50de1c8646d7e179f252070e18047a135d010052e7665ea5c16R60): 



`users: isManagedEventType || schedulingType ? undefined : { connect: { id: userId } },`

packages/trpc/server/routers/viewer/slots/util.ts
CarinaWolli: I just talked to @emrysal, we should still support `eventtype.users`. So we need to fix the error when booking `"Cannot read properties of undefined (reading 'name')".`

CarinaWolli (dismissed):
As already mentione [here](https://github.com/calcom/cal.com/pull/13888/files#r1507897643), we  should still support `eventtype.users`. So we need to fix the error when booking `"Cannot read properties of undefined (reading 'name')"` if there are no hosts but users

packages/trpc/server/routers/viewer/slots/util.ts
Udit-takkar: i am not able to find this error message. I don't see anything in console.log and booking is created in DB and "Cannot read properties of undefined (reading 'name') messages is displayed on booking page.

packages/features/bookings/Booker/components/hooks/useBookings.ts
```diff
@@ -124,7 +124,7 @@ export const useBookings = ({ event, hashedLink, bookingForm, metadata }: IUseBo
           date: responseData?.startTime?.toString() || "",
           duration: validDuration,
           organizer: {
-            name: users?.[0].name || "Nameless",
+            name: users?.[0]?.name || "Nameless",
```
Udit-takkar, line 127: There was no error in handlenewbooking

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -244,12 +244,17 @@ export const getPublicEvent = async (
...
 
   if (users === null) {
     throw new Error("Event has no owner");
   }
 
+  // For backward compatibility when team event type has users[] but not hosts[]
+  if (!users.length) {
+    users = await getOwnerFromUsersArray(prisma, event.id);
+  }
+
```
Udit-takkar, line 257: users was empty array when users[] was present but not hosts[] in a team event type

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -244,7 +244,12 @@ export const getPublicEvent = async (
     hosts: hosts,
   };
 
-  const users = getUsersFromEvent(eventWithUserProfiles) || (await getOwnerFromUsersArray(prisma, event.id));
+  let users = getUsersFromEvent(eventWithUserProfiles) || (await getOwnerFromUsersArray(prisma, event.id));
```
keithwillcode, line 247: @Udit-takkar We already call `getOwnerFromUsersArray(prisma, event.id)` here in the OR and then you added it again down below which seems odd. 

Why do we need this call twice?

Udit-takkar: getUsersFromEvent was returning an empty array in case of legacy team event types where hosts[] is empty but users[] existed because of 

```
  if (team) {
    return (hosts || []).filter((host) => host.user.username).map(mapHostsToUsers);
  }

```

and getOwnerFromUsersArray(prisma, event.id) wouldn't be executed if empty array is returned.

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -342,11 +344,15 @@ function getUsersFromEvent(
...
+  prisma: PrismaClient
 ) {
-  const { team, hosts, owner } = event;
+  const { team, hosts, owner, id } = event;
   if (team) {
-    return (hosts || []).filter((host) => host.user.username).map(mapHostsToUsers);
+    // getOwnerFromUsersArray is used here for backward compatibility when team event type has users[] but not hosts[]
+    return hosts.length
+      ? hosts.filter((host) => host.user.username).map(mapHostsToUsers)
+      : (await getOwnerFromUsersArray(prisma, id)) ?? [];
```
Udit-takkar, line 355: @keithwillcode now i have moved the logic here

keithwillcode (approved):

PR #13889, opened 2024-02-27
fix: impersonation detecting orgs correctly.

sean-brydon:
Fixes issues customer was having returning to their account when impersonating. 

After:
CleanShot 2024-02-27 at 12 23 27


Udit-takkar (dismissed):
LGTM

packages/features/ee/impersonation/lib/ImpersonationProvider.ts
```diff
@@ -213,7 +214,10 @@ async function isReturningToSelf({ session, creds }: { session: Session | null;
 
   if (returningUser) {
     // Skip for none org users
-    if (returningUser.role !== "ADMIN" && !returningUser.organizationId) return;
+    const inOrg =
```
keithwillcode, line 217: This doesn’t make sense. 

The variable is inOrg but the first thing it checks is the user has no organizationId set. Is this correct?

packages/features/ee/impersonation/lib/ImpersonationProvider.ts
```diff
@@ -213,7 +214,10 @@ async function isReturningToSelf({ session, creds }: { session: Session | null;
...
+      !returningUser.organizationId || // Keep for backwards compatability
```
sean-brydon, line 218:
```suggestion
      returningUser.organizationId || // Keep for backwards compatability
```

keithwillcode (approved):

PR #13892, opened 2024-02-27
fix: broken google account avatar fix

therealrinku:
## What does this PR do?
seems like previoulsy avatar url was broken when signing up and loging in with google... turns out CAL_URL was added to the avatar url that was coming from google account. so I made small changes to check if avatar url starts with slash, only then prepend CAL_URL to it.

before:
image

after:
image

## Type of change

## How should this be tested?

To confirm the issue, sign up with google then login and go to the profile page (https://app.cal.com/settings/my-account/profile)
and profile picture on the top of sidebar as well as profile picture on the upload avatar modal was broken, as shown in screenshot 1

## Mandatory Tasks


Udit-takkar (approved):
LGTM. pushed some changes

PR #13896, opened 2024-02-27
chore: Removed redundant E2E tests

keithwillcode:
## What does this PR do?

When the first pass at sweeping E2E tests was done, there were redundant tests added. We now have unit tests wrapped around field requirements, which means having these bulky E2E tests is no longer needed.

## Type of change

## Mandatory Tasks


apps/web/playwright/booking/longTextQuestion.e2e.ts
```diff
@@ -52,432 +52,4 @@ test.describe("Booking With Long Text Question and Each Other Question", () => {
     await bookingPage.cancelBooking(eventTypePage);
     await bookingPage.assertBookingCanceled(eventTypePage);
   });
-
-  test.describe("Booking With Long Text Question and Checkbox Group Question", () => {
```
keithwillcode, line 56: We now have unit tests wrapping logic around required fields so these bulky E2E tests are no longer needed

keithwillcode:
@emrysal took care of this in #13900

PR #13900, opened 2024-02-27
fix: Test flakiness due to unfilled name/email

emrysal:
## What does this PR do?

* `data-testId` instead of `data-testid` throws console errors
* A test with if statements is a bad idea, removed the if statement as this is preventing a flaky test to sometimes pass
* Also fixes the flaky test.

Following picture sums it up nicely

image


sean-brydon (dismissed):

keithwillcode:
@emrysal looks like the profile test is failing consistently

Udit-takkar (approved):

PR #13913, opened 2024-02-28
feat: implement turnstile on signup

sean-brydon:
## What does this PR do?
Implements turnstile captcha to verify signup requests are not coming from bots 

EDIT:

Appearance mode is now https://developers.cloudflare.com/turnstile/get-started/client-side-rendering/#appearance-modes set to interactive-only -> " If appearance is set to interaction-only, the widget will become only visible in cases where an interaction is required"

Fixes #13913

https://github.com/calcom/cal.com/assets/55134778/9b33def6-97b9-4dca-ade7-fe7d148d4b00

## Type of change

## How should this be tested?

> Cloudflare Turnstile
NEXT_PUBLIC_CLOUDFLARE_SITEKEY=
CLOUDFLARE_TURNSTILE_SECRET=
> Please setup turnstile by logging in to a cloudflare account - clicking turnstile on the left and side. Create a site and add `localhost` as the domain without the port. Set the env variables above.


.env.example
```diff
@@ -110,6 +110,8 @@ NEXT_PUBLIC_HELPSCOUT_KEY=
 NEXT_PUBLIC_FRESHCHAT_TOKEN=
 NEXT_PUBLIC_FRESHCHAT_HOST=
 
+NEXT_PUBLIC_CLOUDFLARE_SITEKEY=
+
```
sean-brydon, line 114:
```suggestion
```

apps/web/components/auth/Turnstile.tsx
```diff
@@ -0,0 +1,12 @@
+import type { TurnstileProps } from "react-turnstile";
+import Turnstile from "react-turnstile";
+
+import { CLOUDFLARE_SITE_ID } from "@calcom/lib/constants";
+
+type Props = Omit<TurnstileProps, "sitekey">;
+
+export default function TurnstileWidget(props: Props) {
+  if (!CLOUDFLARE_SITE_ID || process.env.NEXT_PUBLIC_IS_E2E) return null;
```
sean-brydon, line 9: Hide in E2E or if there is no cloudflare site id

apps/web/lib/signup/getServerSideProps.tsx
```diff
@@ -25,13 +25,12 @@ const querySchema = z.object({
...
   const prisma = await import("@calcom/prisma").then((mod) => mod.default);
   const flags = await getFeatureFlagMap(prisma);
-  const ssr = await ssrInit(ctx);
+  await ssrInit(ctx);
   const token = z.string().optional().parse(ctx.query.token);
 
   const props = {
     isGoogleLoginEnabled: IS_GOOGLE_LOGIN_ENABLED,
     isSAMLLoginEnabled,
-    trpcState: ssr.dehydrate(),
```
sean-brydon, line 34: This wasnt being used and was causing signup to be too large

apps/web/pages/signup.tsx
```diff
@@ -32,10 +40,15 @@ import { getServerSideProps } from "@lib/signup/getServerSideProps";
 
 import PageWrapper from "@components/PageWrapper";
 
+// import TurnstileCaptcha from "@components/auth/Turnstile";
+
```
sean-brydon, line 44:
```suggestion
```

apps/web/pages/signup.tsx
```diff
@@ -202,7 +215,8 @@ export default function Signup({
 
   const isOrgInviteByLink = orgSlug && !prepopulateFormValues?.username;
 
-  const signUp: SubmitHandler<FormValues> = async (data) => {
+  const signUp: SubmitHandler<FormValues> = async (_data) => {
+    const { cfToken, ...data } = _data;
```
sean-brydon, line 219: Remove the cfToken from the data payload so we can set it on a header

Udit-takkar (commented):
Nice. should we also use this on the booking page?

sean-brydon:
> Nice. should we also use this on the booking page?

I was actually just thinking about this 20 mins ago - I don't see any reason why we shouldn't

@calcom/foundation

keithwillcode:
> > Nice. should we also use this on the booking page?
> 
> I was actually just thinking about this 20 mins ago - I don't see any reason why we shouldn't
> 
> @calcom/foundation

Curious what @PeerRich and @ciaranha think but my current thinking is this should only be added to the booking page if non-intrusive and not visible.

ciaranha:
> > > Nice. should we also use this on the booking page?
> 
> > 
> 
> > I was actually just thinking about this 20 mins ago - I don't see any reason why we shouldn't
> 
> > 
> 
> > @calcom/foundation
> 
> 
> 
> Curious what @PeerRich and @ciaranha think but my current thinking is this should only be added to the booking page if non-intrusive and not visible. 

Fully agree on the not visible part. It's not pretty.

packages/lib/server/checkCfTurnstileToken.ts
```diff
@@ -0,0 +1,35 @@
+import { HttpError } from "../http-error";
+
+const TUNRSTILE_SECRET_ID = process.env.CLOUDFLARE_TURNSTILE_SECRET;
+
+export async function checkCfTurnstileToken({ token, remoteIp }: { token?: string; remoteIp: string }) {
+  // This means the instant doesnt have turnstile enabled - we skip the check and just return success.
+  // OR the instance is running in CI so we skip these checks also
```
adasarpan404, line 7:
```
  if (!remoteIp || !/^(\d{1,3}\.){3}\d{1,3}$/.test(remoteIp)) {
    throw new HttpError({ statusCode: 400, message: "Invalid remote IP address" });
  }
```
This ensures that remoteIp is always provided and is a valid IP address.

packages/lib/server/checkCfTurnstileToken.ts
```diff
@@ -0,0 +1,35 @@
+import { HttpError } from "../http-error";
+
+const TUNRSTILE_SECRET_ID = process.env.CLOUDFLARE_TURNSTILE_SECRET;
```
adasarpan404, line 3: There's a typo in the variable name `TUNRSTILE_SECRET_ID`, it should be `TURNSTILE_SECRET_ID`

packages/lib/server/checkCfTurnstileToken.ts
```diff
@@ -0,0 +1,35 @@
...
+    return {
+      success: true,
+    };
+  }
+
+  if (!token) {
+    throw new HttpError({ statusCode: 401, message: "Invalid cloudflare token" });
+  }
+
+  const form = new URLSearchParams();
```
adasarpan404, line 18:
```javascript
  try {
    const form = new URLSearchParams();
    form.append("secret", TURNSTILE_SECRET_ID);
    form.append("response", token);
    form.append("remoteip", remoteIp);

    const result = await fetch("https://challenges.cloudflare.com/turnstile/v0/siteverify", {
      method: "POST",
      body: form,
    });

    const data = await result.json();

    if (!data.success) {
      throw new HttpError({ statusCode: 401, message: "Invalid cloudflare token" });
    }

    return data;
  } catch (error) {
    throw new HttpError({ statusCode: 500, message: "Internal Server Error" });
  }
```
can't we put this code in try catch block

sean-brydon:
> The lack of alignment with the other items, the harsh borders, the oversized check icon, and the fact it's the only item with colour (when the button is disabled); is making this stick out a lot when it's the only thing on the page we don't need them to engage with in some way since it's automatic right?

We can set the mode on cloudflares side to be invisible -> 
CleanShot 2024-03-12 at 14 47 01

But i'd honestly recommend Managed mode as thats what they recommend themselves for the most part. 

We dont have too much control over what this looks like from what i can tell. I can try my best to style this to match our theme

sean-brydon:
> > The lack of alignment with the other items, the harsh borders, the oversized check icon, and the fact it's the only item with colour (when the button is disabled); is making this stick out a lot when it's the only thing on the page we don't need them to engage with in some way since it's automatic right?
> 
> We can set the mode on cloudflares side to be invisible -> CleanShot 2024-03-12 at 14 47 01
> 
> But i'd honestly recommend Managed mode as thats what they recommend themselves for the most part.
> 
> We dont have too much control over what this looks like from what i can tell. I can try my best to style this to match our theme

Note: Everything is loaded in a iframe and we have no way of styling this

apps/web/pages/signup.tsx
```diff
@@ -247,319 +260,336 @@ export default function Signup({
...
+                <PasswordField
+                  data-testid="signup-passwordfield"
+                  label={t("password")}
+                  {...register("password")}
+                  hintErrors={["caplow", "min", "num"]}
+                />
+                {/* Cloudflare Turnstile Captcha */}
+                {CLOUDFLARE_SITE_ID ? (
+                  <TurnstileCaptcha
+                    appearance="interaction-only"
```
sean-brydon, line 346: If appearance is set to interaction-only, the widget will become only visible in cases where an interaction is required
 
 https://developers.cloudflare.com/turnstile/get-started/client-side-rendering/#appearance-modes

apps/web/lib/signup/getServerSideProps.tsx
```diff
@@ -24,8 +24,9 @@ const querySchema = z.object({
 
 export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   const prisma = await import("@calcom/prisma").then((mod) => mod.default);
+  await ssrInit(ctx);
```
CarinaWolli, line 27: can this be removed too?

apps/web/pages/signup.tsx
CarinaWolli: Great find, that looks like what we want

apps/web/lib/signup/getServerSideProps.tsx
sean-brydon: No they seems to be the only way translations load on signup 🤔

CarinaWolli (commented):
Looks good to me, but let's have somebody from @calcom/foundation to approve here

joeauyeung (changes requested):
This is much needed. I tested with a different combination of test keys and have some  questions before I feel comfortable approving.

- When failing we don't show the Cloudflare turnstile component but the create account button is blocked. Is this what we want? It renders though when I force an interaction
Example with a failing sitekey
image
Forced interaction
image

joeauyeung (dismissed):
Talking to @sean-brydon about my last review and I feel like my question has been answered. Since @CarinaWolli has also reviewed this PR as well I feel comfortable approving this PR with two reviews.

zomars (dismissed):
LGTM

zomars:
Let's fix conflicts and ensure tests are passing before merging 🙏🏽

keithwillcode (approved):
All checks now passing cc @zomars

PR #13934, opened 2024-02-29
chore: use availability settings atom and fix useTimezone platform hook

Ryukemeister:
## What does this PR do?

- Replace availability settings component in webapp with atom component
- Add more custom styling for atoms
- Enable `useTime` hook to automatically update a user's timezone on any timezone change

### DIFF
--- a/apps/web/pages/availability/[schedule].tsx
+++ b/apps/web/pages/availability/[schedule].tsx
@@ -1,39 +1,16 @@
 import { useRouter } from "next/navigation";
-import { useState } from "react";
-import { Controller, useFieldArray, useForm } from "react-hook-form";
+import { useForm } from "react-hook-form";
 
-import dayjs from "@calcom/dayjs";
-import { DateOverrideInputDialog, DateOverrideList } from "@calcom/features/schedules";
-import Schedule from "@calcom/features/schedules/components/Schedule";
-import Shell from "@calcom/features/shell/Shell";
-import { classNames } from "@calcom/lib";
-import { availabilityAsString } from "@calcom/lib/availability";
 import { useCompatSearchParams } from "@calcom/lib/hooks/useCompatSearchParams";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { HttpError } from "@calcom/lib/http-error";
+import { AvailabilitySettings } from "@calcom/platform-atoms";
 import { trpc } from "@calcom/trpc/react";
 import useMeQuery from "@calcom/trpc/react/hooks/useMeQuery";
-import type { Schedule as ScheduleType, TimeRange, WorkingHours } from "@calcom/types/schedule";
-import {
-  Button,
-  ConfirmationDialogContent,
-  Dialog,
-  DialogTrigger,
-  Form,
-  Label,
-  showToast,
-  Skeleton,
-  SkeletonText,
-  Switch,
-  TimezoneSelect,
-  Tooltip,
-  VerticalDivider,
-  EditableHeading,
-} from "@calcom/ui";
-import { Info, MoreVertical, ArrowLeft, Plus, Trash } from "@calcom/ui/components/icon";
+import type { Schedule as ScheduleType, TimeRange } from "@calcom/types/schedule";
+import { showToast } from "@calcom/ui";
 
 import PageWrapper from "@components/PageWrapper";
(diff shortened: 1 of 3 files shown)


apps/web/pages/availability/[schedule].tsx
```diff
@@ -158,282 +86,40 @@ export default function Availability() {
...
-            </div>
-          </div>
-        </Form>
-      </div>
-    </Shell>
+      isDeleting={deleteMutation.isPending}
+      isLoading={isPending}
+      isSaving={updateMutation.isPending}
+      timeFormat={timeFormat}
+      weekStart="Sunday"
```
Comment (ThyMinimalDev) [inline apps/web/pages/availability/[schedule].tsx:109]: I think this need to use the me/session data, this is how it was written before

```weekStart={
                      ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"].indexOf(
                        me.data?.weekStart
                      ) as 0 | 1 | 2 | 3 | 4 | 5 | 6
                    }```

File: packages/platform/atoms/availability/AvailabilitySettings.tsx [hunk 2]
```diff
@@ -76,6 +76,13 @@ export type CustomClassNames = {
   formClassName?: string;
   timezoneSelectClassName?: string;
   subtitlesClassName?: string;
+  scheduleClassName?: {
```
ThyMinimalDev, line 79: I think we can call this scheduleClassnames with an S at this end to make it clear it contains multiple one and is not just a string like the rest

packages/platform/atoms/availability/AvailabilitySettings.tsx
```diff
@@ -389,6 +396,14 @@ export function AvailabilitySettings({
...
                       deleteTime: translations?.delete ?? t("delete"),
                     }}
+                    className={{
+                      schedule: customClassNames?.scheduleClassName?.scheduleContainerClassName,
+                      scheduleDay: customClassNames?.scheduleClassName?.scheduleDayClassName,
+                      dayRanges: customClassNames?.scheduleClassName?.dayRangesClassName,
+                      labelAndSwitchContainer:
+                        customClassNames?.scheduleClassName?.labelAndSwitchContainerClassName,
+                      timeRanges: customClassNames?.scheduleClassName?.timeRangesClassName,
+                    }}
```
ThyMinimalDev, line 406: I think we can use the spread operator here

```
className={customClassNames?.scheduleClassName ? {...customClassNames?.scheduleClassName} : {} }
```

Ryukemeister: Will have to rename some inner props inside scheduleClassNames prop but yeah no issues otherwise

apps/web/pages/availability/[schedule].tsx
Comment (Ryukemeister) [reply apps/web/pages/availability/[schedule].tsx:109]: @ThyMinimalDev this is done internally in our atom, all we need from the outside is the `me.data?.weekStart` which is a string so I guess the only thing that I need to pass in from the outside is the weekStart from a user

packages/platform/atoms/availability/AvailabilitySettings.tsx
```diff
@@ -76,12 +76,12 @@ export type CustomClassNames = {
   formClassName?: string;
   timezoneSelectClassName?: string;
   subtitlesClassName?: string;
-  scheduleClassName?: {
-    scheduleContainerClassName?: string;
-    scheduleDayClassName?: string;
-    dayRangesClassName?: string;
-    timeRangesClassName?: string;
-    labelAndSwitchContainerClassName?: string;
+  scheduleClassNames?: {
```
Ryukemeister, line 79: Had to rename these since there was naming mismatches happening at the time of spreading

apps/web/pages/availability/[schedule].tsx
```diff
@@ -106,7 +106,7 @@ export default function Availability() {
       isLoading={isPending}
       isSaving={updateMutation.isPending}
       timeFormat={timeFormat}
-      weekStart="Sunday"
+      weekStart={me.data?.weekStart || "Sunday"}
```
Comment (Ryukemeister) [inline apps/web/pages/availability/[schedule].tsx:109]: If there is no user weekStart we assume it to be Sunday and the rest of logic for weekStart is being handled inside of the atom so only thing we need from outside is user week start day

Ryukemeister:
@ThyMinimalDev I've implemented the feedback, everything should work as expected now

PR #13938, opened 2024-02-29
fix: workfow e2e tests

CarinaWolli:
## What does this PR do?

Fixes failing workflow e2e tests, they always failed at the end of the month.


apps/web/playwright/workflow.e2e.ts
```diff
@@ -47,6 +47,7 @@ test.describe("Workflow Tab - Event Type", () => {
 
         await createWorkflow({ name: "A New Workflow", trigger: WorkflowTriggerEvents.NEW_EVENT });
         await page.goto(`/${user.username}/${eventType.slug}`);
+        await page.click('[data-testid="incrementMonth"]');
```
CarinaWolli, line 50: Otherwise on the last day of the month the next day would be booked. But the reminder is set for 24 hours before, which is already in the past at the time of booking

zomars (approved):

PR #13944, opened 2024-03-01
refactor: v2 user timezones

supalarry:
Dayjs on frontend detect timezones in a different format than IsTimezone validator accepts, which initially was created to be used with schedules, so the validator is removed to enable updating managed users.


Ryukemeister (approved):
LGTM

PR #13951, opened 2024-03-01
feat: travel schedules to schedule timezone changes

### LINKED ISSUE #6790: [CAL-963] ability to "plan" your timezone in the future when traveling
As someone who travels frequently, I want the ability to schedule a timezone change for a specific date range in the future.

**User flow**

* I should be able to go into my General settings and where I edit my timezone, see an action to "Schedule a timezone change".
* This will open a modal with a date range selection and the timezone editor.
* I enter my dates, pick the timezone (of the location I will be travelling to) and click confirm.
* I can then see the timezone override below the timezone select. Format:
* 
  * (GMT+01:00) Central European Standard Time
* 
  * March 1, 2023 — March 31, 2023
* I can also delete it

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-963](https://linear.app/calcom/issue/CAL-963/ability-to-plan-your-timezone-in-the-future-when-traveling)</sub>

CarinaWolli:
## What does this PR do?

New feature to add travel schedules for scheduling a timezone change. 

Loom showing availability with travel schedules: https://www.loom.com/share/29d8500385fb4ff78e025021ad5710ea

Fixes #6790

## Type of change

## How should this be tested?

To run locally change your system timezone to utc time.

- Create travel schedules 
- test availability 
- test CRON API endpoint `/api/cron/changeTimeZone`
    - change system timezone to right after travel schedule started (midnight of user timezone)
    - check if timezone changed & check if timezone of default schedule changed


PeerRich:
ouhhhh exciting!!

packages/features/schedules/components/DateOverrideList.tsx
```diff
@@ -72,7 +78,17 @@ const DateOverrideList = ({
             ) : (
               item.ranges.map((range, i) => (
                 <p key={i} className="text-subtle text-xs">
-                  {timeSpan(range)}
+                  {`${timeSpan(range)} ${
```
CarinaWolli, line 81: Show timezone to date override if it is within a travel schedule (only on default schedule): 
Screenshot 2024-03-04 at 6 18 51 PM

packages/lib/date-ranges.ts
```diff
@@ -10,25 +10,45 @@ export type DateRange = {
...
+function getAdjustedTimezone(date: Dayjs, timeZone: string, travelSchedules: TravelSchedule[]) {
+  let adjustedTimezone = timeZone;
+
+  travelSchedules.forEach((travelSchedule) => {
+    if (!date.isBefore(travelSchedule.startDate)) {
+      if (!travelSchedule.endDate || !date.isAfter(travelSchedule.endDate)) {
+        adjustedTimezone = travelSchedule.timeZone;
+      }
+    }
+  });
```
Udit-takkar, line 24: NIT

```suggestion
  travelSchedules.forEach((travelSchedule) => {
    if (!date.isBefore(travelSchedule.startDate) && (!travelSchedule.endDate || !date.isAfter(travelSchedule.endDate))  ) {
       
        adjustedTimezone = travelSchedule.timeZone;
      
    }
  });
```

apps/web/pages/settings/my-account/general.tsx
```diff
@@ -181,6 +222,64 @@ const GeneralView = ({ localeProp, user }: GeneralViewProps) => {
...
             )}
           />
+          {!watchedTzSchedules.length ? (
+            <Button color="minimal" className="mt-2" onClick={() => setIsTZScheduleOpen(true)}>
+              {t("schedule_timezone_change")}
+            </Button>
+          ) : (
+            <div className="bg-muted border-subtle mt-2 rounded-md border p-4">
+              <Label>{t("travel_schedule")}</Label>
+              <div className="border-subtle mt-4 rounded-md border bg-white text-sm">
```
Udit-takkar, line 232: Screenshot 2024-03-06 at 6 23 07 PM

Dark mode

packages/features/schedules/components/DateOverrideList.tsx
```diff
@@ -53,6 +58,7 @@ const DateOverrideList = ({
       new Date(end.toISOString().slice(0, -1))
     )}`;
   };
+  console.log(`items ${JSON.stringify(items)}`);
```
Udit-takkar, line 61: NIT: remove this

packages/lib/date-ranges.test.ts
```diff
@@ -207,10 +207,85 @@ describe("processWorkingHours", () => {
...
+    ).toBeTruthy();
+
+    // 8AM in America/New_York is 1PM in UTC, 5PM in America/New_York is 10PM in UTC
+    expect(
+      resultsWithNewYorkTz.every((result) => {
+        return result.start.utc().hour() === 13 && result.end.utc().hour() === 22;
+      })
+    ).toBeTruthy();
+
+    // 8AM in America/Kolkata is 2:30AM in UTC, 5PM in America/Kolkata is 11:30AM in UTC
```
Udit-takkar, line 280:
```suggestion
    // 8AM in America/New_York is 2:30AM in UTC, 5PM in America/New_York is 11:30AM in UTC
```

Udit-takkar, line 280: Asia/Kolkata or America/New_York?

apps/web/components/settings/TravelScheduleModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+                }}
+              />
+            </>
+          )}
+          <div className="text-error mt-1 text-sm">{errorMessage}</div>
+
+          <div className="mt-3">
+            <SettingsToggle
+              labelClassName="mt-1 font-normal"
+              title="Schedule timezone without end date"
```
Udit-takkar, line 121: NIT: use i18n

packages/lib/date-ranges.test.ts
CarinaWolli: Oops, should be Asia/Kolkata. Fixed now

packages/features/schedules/components/DateOverrideList.tsx
```diff
@@ -61,7 +66,17 @@ const DateOverrideList = ({
...
-                  {timeSpan(range)}
+                  {`${timeSpan(range)} ${
+                    isDefaultSchedule
+                      ? travelSchedules.find(
+                          (travelSchedule) =>
+                            !dayjs(item.ranges[0].start).isBefore(travelSchedule.startDate) &&
+                            (!dayjs(item.ranges[0].end).isAfter(travelSchedule.endDate) ||
+                              !travelSchedule.endDate)
+                        )?.timeZone || ""
+                      : ""
```
joeauyeung, line 77: Very nice considering this case.

packages/lib/date-ranges.ts
```diff
@@ -10,25 +10,46 @@ export type DateRange = {
 export type DateOverride = Pick<Availability, "date" | "startTime" | "endTime">;
 export type WorkingHours = Pick<Availability, "days" | "startTime" | "endTime">;
 
+type TravelSchedule = { startDate: Dayjs; endDate?: Dayjs; timeZone: string };
+
+function getAdjustedTimezone(date: Dayjs, timeZone: string, travelSchedules: TravelSchedule[]) {
+  let adjustedTimezone = timeZone;
+
+  travelSchedules.forEach((travelSchedule) => {
```
joeauyeung, line 18: Instead of looping through all the `travelSchedules` can we use a `.find()` or a `for ... of` loop and break once we find the schedule we're looking for

packages/trpc/server/middlewares/sessionMiddleware.ts
```diff
@@ -71,6 +71,7 @@ export async function getUserFromSession(ctx: TRPCContextInner, session: Maybe<S
       allowDynamicBooking: true,
       allowSEOIndexing: true,
       receiveMonthlyDigestEmail: true,
+      travelSchedules: true,
```
joeauyeung, line 74: Why are we adding `travelSchedules` to the user session object? I would want to keep this as learn as possible.

apps/web/pages/api/cron/changeTimeZone.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    await prisma.schedule.updateMany({
+      where: {
+        id: defaultScheduleId,
+      },
+      data: {
+        timeZone: timeZone,
+      },
+    });
```
joeauyeung, line 67: I don't see any messaging about updating the default schedule timezone. I have two types of availabilities that I use for different event types. What if a user wants all their schedules updated? Could be a @ciaranha @itsberkaya question.

apps/web/pages/api/cron/changeTimeZone.ts
joeauyeung, line 1: Love seeing Dayjs being used like this here.

joeauyeung (changes requested):
Great work so far @CarinaWolli. I tried testing hitting the cron endpoint manually but I got the timezone to change but I think at a certain point I missed the time window. 

## Blocking
- On the profile it looks like the timezones aren't the translated strings (Hong_kong vs Hong Kong)
image

- ### Date Selector
	- When selecting a travel schedule with no end on dark mode the date selector is rendering weird
	image
	- Also with the date selector, the week starts on Monday but my settings are set to Sunday. This might cause some confusion for people.
	- I'm able to choose dates in the past
	- When adding multiple travel schedules, I think the dialog state should reset. (start and end dates, timezone)

- I set the new timezone to be today. When I created a booking it shows the booking time as the original timezone. I think if a user sets the current day for the new travel schedule it should update the bookings page. I'm wondering as well does the bookings page show the new updated times?
image


## Non-blocking
- Would be cool to also schedule a timezone change here. Similar to how OOO is advertised
image
- Should travel schedules be automatically added rather than hitting "Update" on the settings page? It just feels more intuitive to me. 
- It would be nice to edit individual travel schedules

apps/web/pages/api/cron/changeTimeZone.ts
CarinaWolli: @ciaranha and me talked about this a while ago. When changing the profile timezone only the default schedule should be updated. I think it actually would make sense to update the default schedule + every schedule that has the same timezone as the old profile timezone

CarinaWolli:
> * On the profile it looks like the timezones aren't the translated strings (Hong_kong vs Hong Kong)

 fixed, not translated but same as in the timezone select we just remove the _

> * ### Date Selector
>   
>   * When selecting a travel schedule with no end on dark mode the date selector is rendering weird

fixed
Screenshot 2024-03-19 at 5 45 28 PM

>   * Also with the date selector, the week starts on Monday but my settings are set to Sunday. This might cause some confusion for people.

this is an issue with DatePicker/DateRangePicker component let’s handle that in a separate PR

>   * I'm able to choose dates in the past

adding minDate which would solve that causes the weird dark mode design, that's also an issue with the DatePicker/DateRangePicker component, let’s handle that in a separate PR as well

>   * When adding multiple travel schedules, I think the dialog state should reset. (start and end dates, timezone)

fixed

> * I set the new timezone to be today. When I created a booking it shows the booking time as the original timezone. I think if a user sets the current day for the new travel schedule it should update the bookings page. I'm wondering as well does the bookings page show the new updated times?

When the cron job gets triggered your profile timezone would then get update and then everything including the bookings would show in the new timezone. You think we should check if the start date is today and if yes change the timezone right away?

packages/trpc/server/middlewares/sessionMiddleware.ts
CarinaWolli: removed it and added a separate getTravelSchedules handler instead

CarinaWolli:
@joeauyeung I think I addressed all your blocking comments, let me know if I missed something. We can address the non-blocking comments in follow up

joeauyeung:
> > If I change the timezone today to a new one, I would want it to change now rather than wait until midnight. Could be another @ciaranha @itsberkaya question.
> 
> You wouldn't need to wait until midnight, we run the corn job on every full hour so it would trigger the next full hour. I just didn't see this as a use case of this feature to be honest, I would assume that the user would just change the timezone in the profile if they need it to be changed right away.

I was wondering if the person was already in the timezone and needed to change it on their profile immediately. It would be nice to give them this option to change it via the travel schedule since there's an end date. This is a non-blocking issue though.

apps/web/pages/availability/[schedule].tsx
```diff
@@ -100,6 +105,7 @@ export default function Availability() {
             }
           : undefined
       }
+      travelSchedules={isDefaultSchedule ? travelSchedules || [] : []}
```
Comment (CarinaWolli) [inline apps/web/pages/availability/[schedule].tsx:108]: @calcom/platform I assume I also need to add the travelSchedules to PlatformAvailabilitySettingsWrapper now?

apps/web/pages/api/cron/changeTimeZone.ts
```diff
@@ -0,0 +1,170 @@
...
+      }
+      await prisma.travelSchedule.delete({
+        where: {
+          id: travelSchedule.id,
+        },
+      });
+    }
+  }
+  res.status(200).json({ timeZonesChanged });
+}
```
ThyMinimalDev, line 170: this cron is doing lots of individual calls to the database and might create heavy load on it, I suggest checking if we could use transactions and making sure the right indexes are defined in the schema

apps/web/pages/api/cron/changeTimeZone.ts
```diff
@@ -0,0 +1,170 @@
...
+        // travel schedule ended, change back to original timezone
+        await setNewTimeZone(travelSchedule.prevTimeZone, travelSchedule.user);
+      }
+      await prisma.travelSchedule.delete({
+        where: {
+          id: travelSchedule.id,
+        },
+      });
+    }
+  }
```
ThyMinimalDev, line 168: we should avoid calling prisma inside of a for loop, can we here use the loop to aggregate the data to delete/update in vars and then do a deleteMany/updateMany after the for loop ?

apps/web/pages/api/cron/changeTimeZone.ts
ThyMinimalDev: now that I think about it, since it's a cron and it will most likely only execute once in a while, transaction is not important, only the indexes

joeauyeung:
@CarinaWolli looking through the commits since I last reviewed not much have changed so I'm ok to approve this PR once @ThyMinimalDev's comments have been addressed.

apps/web/pages/api/cron/changeTimeZone.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (!dayjs.utc().isBefore(endDateUTC)) {
+      if (travelSchedule.prevTimeZone) {
+        // travel schedule ended, change back to original timezone
+        await setNewTimeZone(travelSchedule.prevTimeZone, travelSchedule.user);
+      }
+      travelScheduleIdsToDelete.push(travelSchedule.id);
+    }
+  }
+
+  await prisma.travelSchedule.deleteMany({
```
CarinaWolli, line 164: @ThyMinimalDev using delteMany now instead of deleting side for loop

apps/web/pages/api/cron/changeTimeZone.ts
CarinaWolli: added indexes

apps/web/pages/api/cron/changeTimeZone.ts
CarinaWolli: @ThyMinimalDev  added `deleteMany`, for the updates I think I would need to use transactions to optimize it. Should I do that or are we good like that?

CarinaWolli:
@ThyMinimalDev & @joeauyeung can you take a look again?

joeauyeung (approved):
Looking at [`fd70b67` (#13951)](https://github.com/calcom/cal.com/pull/13951/commits/fd70b67e69d3d835ec2a5bf0b845596ea78aa118) and [`6f09c4c` (#13951)](https://github.com/calcom/cal.com/pull/13951/commits/6f09c4c1fefd1d42de162504bb92773321d397bf) seems like @ThyMinimalDev's comments have been addressed with no new changes to the functionality to be concerned about. Approving this PR.

Great job @CarinaWolli, I can't wait to use this feature myself!

joeauyeung (approved):
Looking at [`fd70b67` (#13951)](https://github.com/calcom/cal.com/pull/13951/commits/fd70b67e69d3d835ec2a5bf0b845596ea78aa118) and [`6f09c4c` (#13951)](https://github.com/calcom/cal.com/pull/13951/commits/6f09c4c1fefd1d42de162504bb92773321d397bf) seems like @ThyMinimalDev's comments have been addressed with no new changes to the functionality to be concerned about. Approving this PR.

Great job @CarinaWolli, I can't wait to use this feature myself!

PR #13963, opened 2024-03-04
refactor: Moving credential syncing to API

### LINKED ISSUE #13904: [CAL-3237] Credential syncing - add functionality to remove credentials
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3237](https://linear.app/calcom/issue/CAL-3237/credential-syncing-add-functionality-to-remove-credentials)</sub>

joeauyeung:
## What does this PR do?

This PR moves credential syncing from the web app to the V1 API. Also adds the ability to list and delete user credentials

Fixes #13904

## Requirement/Documentation

## Type of change

## How should this be tested?

Note: when making the API keys a header with `CALCOM_CREDENTIAL_SYNC_HEADER_NAME` should be set with the value of `CALCOM_CREDENTIAL_SYNC_SECRET`

- Set `CALCOM_CREDENTIAL_SYNC_SECRET`, `CALCOM_CREDENTIAL_SYNC_HEADER_NAME`, and `CALCOM_APP_CREDENTIAL_ENCRYPTION_KEY` in your .env
- With a test user install an app normally
- Call GET `/credential-sync` with the `userId` in the params
- Call POST `/credential-sync` with the `userId` in the params
	-  With a AES256 encrypted credential key that contains the properties `access_token` and `expiry_date`. The credential should be created
	- Remove `expiry_date` from the encrypted key. It should fail
	- Add random properties to the encrypted key. It should also write those properties to the DB
- Call PATCH `credential-sync` with the `userid` and `credentialId` in the params
	- Repeat the same steps with the POST call
- Call DELETE `credential-sync` with the `userid` and `credentialId` in the params

## Mandatory Tasks

## Checklist


apps/api/lib/helpers/verifyCredentialSyncEnabled.ts
```diff
@@ -0,0 +1,24 @@
+import type { NextMiddleware } from "next-api-middleware";
+
+import { APP_CREDENTIAL_SHARING_ENABLED } from "@calcom/lib/constants";
+
+export const verifyCredentialSyncEnabled: NextMiddleware = async (req, res, next) => {
+  const { isAdmin } = req;
+
+  if (!isAdmin) {
+    return res.status(401).json({ error: "Only admin API keys can access credential syncing endpoints" });
```
keithwillcode, line 9: Should this actually be a 403? If it reaches here does that mean we've authenticated the request based on the key supplied but they just don't have access? If so, should be a 403.

apps/api/pages/api/credential-sync/_delete.ts
```diff
@@ -0,0 +1,64 @@
...
+ *       505:
+ *        description: Credential syncing not enabled
+ */
+async function handler(req: NextApiRequest) {
+  const { prisma } = req;
+
+  const { userId: reqUserId, credentialId: reqCredentialId } = schemaCredentialDeleteParams.parse(req.query);
+
+  const userId = parseInt(reqUserId);
+  const credentialId = parseInt(reqCredentialId);
```
exception, line 48: These parses should be handled by zod, I think changing the schema makes more sense than this.

apps/api/pages/api/credential-sync/_get.ts
```diff
@@ -0,0 +1,65 @@
...
+ *        description: Authorization information is missing or invalid.
+ *       505:
+ *        description: Credential syncing not enabled
+ */
+async function handler(req: NextApiRequest) {
+  const { prisma } = req;
+
+  const { appSlug, userId: reqUserId } = schemaCredentialGetParams.parse(req.query);
+
+  const userId = parseInt(reqUserId);
```
exception, line 41: Same as last.

apps/api/pages/api/credential-sync/_patch.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+async function handler(req: NextApiRequest) {
+  const { prisma } = req;
+
+  const { userId: reqUserId, credentialId: reqCredentialId } = schemaCredentialPatchParams.parse(req.query);
+
+  const userId = parseInt(reqUserId);
+  const credentialId = parseInt(reqCredentialId);
+
+  const { encryptedKey } = schemaCredentialPatchBody.parse(req.body);
```
exception, line 62: I think parsing any of these zod defined schemas may error out and not be very descriptive, could be wrong though.

Would you mind checking and if not doing a `safeParse`?

joeauyeung (commented):
@exception added custom error messages when parsing the query params. 

When parsing the request body this is what the error message looks like. I think this should be descriptive enough.

`"message": "invalid_type in 'appSlug': Expected string, received number; invalid_type in 'encryptedKey': Required"`

apps/api/lib/validations/credential-sync.ts
```diff
@@ -0,0 +1,64 @@
...
+  const userIdInt = parseInt(val);
+
+  if (isNaN(userIdInt)) {
+    throw new HttpError({ message: "userId is not a valid number", statusCode: 400 });
+  }
+
+  return userIdInt;
+});
+const appSlug = z.string();
+const credentialId = z.string().transform((val) => {
```
emrysal, line 15: You can do this with coerce also, not a blocker though.

emrysal (approved):
Going to approve this; it seems to me that the guards are correct as well as the logic. Tested and looks good.

zomars:
@emrysal @joeauyeung seems like this got merged after #14069 which makes some of this code obsolete and causing failures. Pushing a fix soon.

PR #13964, opened 2024-03-04
feat: update event types for default availability change

### LINKED ISSUE #13381: [CAL-3010] Changing default availability doesn't update all event types and is unclear what it does update
**Problem from user:**

> When you change your availability default, it doesn't update that availability to other event-types, I don't think there's a straight forward solution here so I'm sharing this loom and it's very confusing as is and can lead to a lot of frustration if you don't understand properly how our product works to debug properly

[https://www.loom.com/share/4756392c798f46b4a3df94581b16b1e9?source=embed_watch_on_loom_cta](https://www.loom.com/share/4756392c798f46b4a3df94581b16b1e9?source=embed_watch_on_loom_cta)

**Proposed solution from user**

> When someone changes an availability to "default" a modal pop ups saying here is where you have your default availability already pre-programmed, do you want this to auto-update the availability of these event-types? And then you can check and uncheck boxes

**Design notes**

Follow similar functionality from how we handle this when changing default video app. Note, we should also better communicate which event types do have the default and which don't after you finish with this modal (which is something we don't currently do for videos either)

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3010](https://linear.app/calcom/issue/CAL-3010/changing-default-availability-doesnt-update-all-event-types-and-is)</sub>

Amit91848:
## What does this PR do?

Fixes #13381 

https://github.com/calcom/cal.com/assets/74371312/1797ecbf-9754-4ebb-91f9-f9840f197b40

## Type of change

## Mandatory Tasks


sean-brydon (approved):
LGTM - pushed a change to show a toast once we successfully update the event types. 

Will wait for CI to pass before merging. 

Thanks!

PR #13965, opened 2024-03-04
feat: v2 endpoints to fetch bookings and hooks

Ryukemeister:
## What does this PR do?

- Adds endpoints to fetch all bookings of a user and bookings based on uid
- Adds hooks for all user bookings and bookings based on uid


apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -30,7 +42,51 @@ import { ApiResponse } from "@calcom/platform-types";
...
+
+    return {
+      status: SUCCESS_STATUS,
+      data: bookings,
+    };
+  }
+
+  // note(Rajiv): currently this endpoint is atoms only
+  @Get("/:bookingUid")
+  @UseGuards(AccessTokenGuard)
```
ThyMinimalDev, line 77: I think this endpoint should not be behind access token guard, for example on app.cal.com, you can access booking success page of any booking without being auth.

for example open this link in an incognito tab: https://app.cal.com/booking/2JeoZEPp6GB7uCo1xTLbLE?allRemainingBookings=false&email=morgan%40cal.com

apps/api/v2/src/ee/bookings/controllers/bookings.controller.e2e-spec.ts
```diff
@@ -0,0 +1,110 @@
...
+import { getAllUserBookings, handleNewBooking, getBookingInfo } from "@calcom/platform-libraries";
+import { ApiSuccessResponse } from "@calcom/platform-types";
+
+describe("Bookings Endpoints", () => {
+  describe("User Authentication", () => {
+    let app: INestApplication;
+
+    let userRepositoryFixture: UserRepositoryFixture;
+
+    const userEmail = "schedules-controller-e2e@api.com";
```
ThyMinimalDev, line 26: "bookings-controller-e2e@api.com"

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
Ryukemeister: Right, makes sense. After a booking has been made there shouldn't be a need to use access token to fetch that booking details, will fix this.

apps/api/v2/src/ee/bookings/controllers/bookings.controller.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  getAllUserBookings,
+  handleNewBooking,
+  getBookingInfo,
+  handleNewRecurringBooking,
+  handleInstantMeeting,
+} from "@calcom/platform-libraries";
+import { ApiSuccessResponse, ApiResponse } from "@calcom/platform-types";
+
+describe("Bookings Endpoints", () => {
+  describe("User Authentication", () => {
```
supalarry, line 28: Having "User Authenticated" instead of "User Authentication" is clearer because we are not testing authentication but cases when a user is authenticated.

apps/api/v2/src/ee/bookings/controllers/bookings.controller.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      return request(app.getHttpServer())
+        .post("/api/v2/ee/bookings")
+        .send(body)
+        .expect(201)
+        .then(async (response) => {
+          const responseBody: ApiSuccessResponse<Awaited<ReturnType<typeof handleNewBooking>>> =
+            response.body;
+          expect(responseBody.status).toEqual(SUCCESS_STATUS);
+          expect(responseBody.data).toBeDefined();
+          expect(responseBody.data.user.email).toBeDefined();
```
supalarry, line 92: Is this the e-mail of "bookings-controller-e2e@api.com" ? if so, we can make test more specific by checking that user.email equals "bookings-controller-e2e@api.com"

apps/api/v2/src/ee/bookings/controllers/bookings.controller.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        });
+    });
+
+    it("should get bookings", async () => {
+      return request(app.getHttpServer())
+        .get("/api/v2/ee/bookings")
+        .then((response) => {
+          const responseBody: ApiSuccessResponse<Awaited<ReturnType<typeof getAllUserBookings>>> =
+            response.body;
+          const fetchedBooking = responseBody.data.bookings[0];
```
supalarry, line 110: I think a good one to add would be that responseBody.data.bookigns length should be just 1 because we made only 1 booking above.

apps/api/v2/src/ee/bookings/controllers/bookings.controller.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        .send(body)
+        .expect(201)
+        .then((response) => {
+          const responseBody: ApiResponse<Awaited<ReturnType<typeof handleInstantMeeting>>> = response.body;
+
+          expect(responseBody.status).toEqual("instant");
+        });
+    });
+  });
+});
```
supalarry, line 205: We also need to delete all the bookings that been created - you need to create "BookingsRepositoryFixture" just like "UserRepositoryFixture" and can add function that deletes all bookings made to a user aka in this case of "bookings-controller-e2e@api.com" and then add "afterAll" that will call "BookingsRepositoryFixture.deleteAllUserBookings("bookings-controller-e2e@api.com") to do the cleanup

supalarry (commented):
Great tests, just some comments!

packages/platform/types/bookings.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class GetBookingsInput {
+  @ValidateNested({ each: true })
+  filters!: Filters;
+
+  @IsNumber()
+  @Min(1)
+  @Max(100)
+  @IsOptional()
+  limit?: number | null;
```
supalarry, line 37: I think we don't need " | null" because either its a number or not passed at all and undefined.

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -34,7 +45,50 @@ import { ApiResponse } from "@calcom/platform-types";
...
+  async getBookings(
+    @GetUser() user: User,
+    @Query() queryParams: GetBookingsInput
+  ): Promise<ApiResponse<unknown>> {
+    const { filters, cursor, limit } = queryParams;
+    const bookings = await getAllUserBookings({
+      bookingListingByStatus: filters.status,
+      skip: cursor ?? 0,
+      take: limit ?? 10,
+      filters: filters,
```
supalarry, line 65: FYI: if key and variable are the same "filters: filters" you can just write:

...
take,
filters,
...

packages/platform/libraries/index.ts
```diff
@@ -2,8 +2,11 @@ import * as newBookingMethods from "@calcom/features/bookings/lib/handleNewBooki
 import { getPublicEvent } from "@calcom/features/eventtypes/lib/getPublicEvent";
 import * as instantMeetingMethods from "@calcom/features/instant-meeting/handleInstantMeeting";
 import { updateHandler as updateScheduleHandler } from "@calcom/trpc/server/routers/viewer/availability/schedule/update.handler";
+import getAllUserBookings from "@calcom/trpc/server/routers/viewer/bookings/getAllUserBookings";
 import { getAvailableSlots } from "@calcom/trpc/server/routers/viewer/slots/util";
 
+import getBookingInfo from "../../../apps/web/modules/bookings/lib/getBookingInfo";
```
supalarry, line 8: cal we use absolute "@calcom/apps/..." import instead of relative "../../../" ?

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -24,7 +32,10 @@ import {
   handleNewRecurringBooking,
   handleInstantMeeting,
 } from "@calcom/platform-libraries";
+import { getAllUserBookings, getBookingInfo } from "@calcom/platform-libraries";
 import { ApiResponse } from "@calcom/platform-types";
+import { GetBookingsInput } from "@calcom/platform-types/bookings";
```
supalarry, line 37: NIT: We import else where simply from the index file "@calcom/platform-types"

packages/platform/types/bookings.ts
```diff
@@ -0,0 +1,42 @@
+import { Type } from "class-transformer";
+import { ValidateNested, IsNumber, Min, Max, IsOptional, IsArray, IsEnum } from "class-validator";
+
+enum Status {
```
supalarry, line 4: In Prisma schema we have enum used for Booking.status:

```
enum BookingStatus {
  CANCELLED     @map("cancelled")
  ACCEPTED      @map("accepted")
  REJECTED      @map("rejected")
  PENDING       @map("pending")
  AWAITING_HOST @map("awaiting_host")
}
```

Can we re-use it? Also, it does not have "upcoming" and "recurring" like in enum here, so I am wondering how does the mis-match work?

supalarry (commented):
Great work Rajiv! Looks good, but only question I have is in [platform/types/bookings.ts](https://github.com/calcom/cal.com/pull/13965/files#diff-b81cad017cb98f6202119bb497682d405cdc809144a155084b6f21930f94e35b) about `enum Status`

packages/platform/libraries/index.ts
Ryukemeister: Yeah I tried this at first but for some reason it just doesn't seem to work :(

packages/platform/types/bookings.ts
Ryukemeister: @supalarry honestly idk why the booking status varies in our prisma schema and the filters object, I'm guessing we transform the data in a certain way in the webapp right and that's why we can'e reuse it other places

supalarry (dismissed):

Ryukemeister:
In the process of replacing this PR with #14164 since this has some insane merge conflicts and `yarn.lock` changes

Ryukemeister:
Closing this in favour of #14164

PR #13968, opened 2024-03-04
fix: add org settings into seed

sean-brydon:
Adds org settings into seed so orgs works as expected


emrysal (approved):
👍 Thanks for fixing

PR #13970, opened 2024-03-04
chore: Increase API rate limit

keithwillcode:
## What does this PR do?

Increase the API rate limit to 30 / minute (1800 / hour). Our limit before was knowingly quite restrictive. Until we have configurable rate limiting per user/partner, increasing the default so we are not hindering people's ability to use the API for their needs.

## Type of change

## Mandatory Tasks


emrysal (approved):
👍

Issue #13716, opened 2024-02-16
[CAL-3152] Give ORG Owners the Ability to lock Event-Types


shirazdole:
Some org customers are offering scheduling for hundreds of users and want users to only use their org account [acme.cal.com/user](http://acme.cal.com/user) for booking meetings related to their business.

Therefore, they don't want them to create other event-types for other purposes. They just want to use managed-event types that users can use to book meetings.

Proposal — give admins/owners the ability to lock personal event-type creation for all users in an organization.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3152](https://linear.app/calcom/issue/CAL-3152/give-org-owners-the-ability-to-lock-event-types)</sub>

PeerRich:
i like this. essentially what you mean: only allow managed event-types, disable existing personal links?

PeerRich:
where should this setting be? what happens to old links? block them? hide them? delete them?

shirazdole:
@peer correct, I'm not sure what happens to old links, maybe we can give it as an option to owers/admins to decide?

ciaranha:
> where should this setting be? what happens to old links? block them? hide them?

I think this should probably live under **General** under the **Org** in settings. We can provide an option to "lock" or "turn off" personal event-type creation. When they toggle it, we could present a modal that offers them a choice of what to do with existing links and essentially whether to keep or delete them.

shirazdole:
@keithwillcode I don't think these are duplicative, one is for availability, the other is for event-types

keithwillcode:
Oh my fault!

sean-brydon:
Ill take this and [CAL-3086](https://linear.app/calcom/issue/CAL-3086/give-org-owners-the-ability-to-lock-availability) next week

alishaz-polymath:
Maybe just redirect old links to the user's public booking page where they should have the managed event types 🤔

PeerRich:
do we really need to force this? can we not tell org admins to make managed-event types and tell their team to not use other links

PeerRich:
since we have a very short connection between org owners usually and we can help during onboarding

shirazdole:
@PeerRich sure no problem 👌🏽 that sounds good and makes sense

PeerRich:
i just learned that the PR is almost ready so we can just have @sean-brydon ship this

PeerRich:
i would never delete links as an admin tho. we can hide private events instead

ciaranha:
[...]

PR #14000, opened 2024-03-07
feat: lock event types for org users 

sean-brydon:
fixes: CAL-3152 #13716 

Adds the ability for org customers to stop users creating events on their own `acme.cal.com/[user]` profile and only rely on managed event types

Loom: https://www.loom.com/share/9dd0f8704d7e42a0891f50deffb9bf17


apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -884,41 +885,50 @@ const Main = ({
...
-                  profile={group.profile}
-                  membershipCount={group.metadata.membershipCount}
-                  teamId={group.teamId}
-                  bookerUrl={group.bookerUrl}
-                />
+            data.eventTypeGroups.map((group, index: number) => {
+              const userHasManagedEventType = group.eventTypes.find(
+                (event) => event.metadata?.managedEventConfig
+              );
+              if (group.profile.eventTypesLockedByOrg && !userHasManagedEventType) return null;
```
sean-brydon, line 899: If locked and no managed event types - we dont need to show to show this "group"

apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -884,41 +885,50 @@ const Main = ({
...
+                      group={group}
+                      bookerUrl={group.bookerUrl}
+                      groupIndex={index}
+                      readOnly={group.metadata.readOnly}
+                    />
+                  ) : group.teamId && !group.metadata.readOnly ? (
+                    <EmptyEventTypeList group={group} />
+                  ) : !group.metadata.readOnly ? (
+                    <CreateFirstEventTypeView slug={data.profiles[0].slug ?? ""} />
+                  ) : null}
```
sean-brydon, line 928: This last null fixes the bug where currently MEMBERS see the option to create events on teams that are empty even tho they dont have permission

packages/trpc/server/routers/viewer/eventTypes/create.handler.ts
```diff
@@ -85,6 +84,27 @@ export const createHandler = async ({ ctx, input }: CreateOptions) => {
     data.schedulingType = schedulingType;
   }
 
+  // If we are in an organization & they are not admin & they are not creating an event on a teamID
+  // Check if evenTypes are locked.
+  if (ctx.user.organizationId && !!!ctx.user?.organization?.isOrgAdmin && !teamId) {
```
CarinaWolli, line 89:
```suggestion
  if (ctx.user.organizationId && !ctx.user?.organization?.isOrgAdmin && !teamId) {
```

apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -884,41 +885,50 @@ const Main = ({
...
+                  {group.eventTypes.length ? (
+                    <EventTypeList
+                      types={group.eventTypes}
+                      group={group}
+                      bookerUrl={group.bookerUrl}
+                      groupIndex={index}
+                      readOnly={group.metadata.readOnly}
+                    />
+                  ) : group.teamId && !group.metadata.readOnly ? (
+                    <EmptyEventTypeList group={group} />
```
CarinaWolli, line 925: Do we not want to show the team event types anymore in read-only?

apps/web/modules/event-types/views/event-types-listing-view.tsx
CarinaWolli: what about the hidden event types? shouldn't we still show them

CarinaWolli: we should also lock the toggle of the hidden event types

CarinaWolli (changes requested):
Nice work 👏🏻 found some things that should be addressed before it can be approved 🙏

apps/web/modules/event-types/views/event-types-listing-view.tsx
sean-brydon: This is only the empty list - it has a cta to create events (when it’s read only the user doesn’t have permission to do so) and the user can start the flow and hit an error 

so this is just hiding it from their view - the event list itself is still rendered

apps/web/modules/event-types/views/event-types-listing-view.tsx
sean-brydon: Good spot! I always had a managed event in there when testing :P

apps/web/modules/event-types/views/event-types-listing-view.tsx
CarinaWolli: Oh yeah sorry, somehow it also didn't show the read-only event types for me. But works now 🙏

sean-brydon:
@CarinaWolli feedback has been addressed please re-review <3

apps/web/modules/event-types/views/event-types-listing-view.tsx
CarinaWolli, line 899: > - what about the hidden event types? shouldn't we still show them
> - we should also lock the toggle of the hidden event types

I don't see any changes here, these two things are still an issue for me. 

Example: Admin locks event types for members. Member has one personal event type (no managed event type). Then we want to mark the event type as hidden and it should still show in the event type list. Right now the hidden event types are not shown if the user doesn't have a managed event type. Also the toggle to make the event type public again should be disabled for the member

sean-brydon: I think I forgot to push some code 😅

sean-brydon: Commit history looks weird as the feedback was timestamped before this message :P 

Latest changes should be there

apps/web/modules/event-types/views/event-types-listing-view.tsx
```diff
@@ -810,6 +813,7 @@ const CTA = ({ data }: { data: GetByViewerResponse }) => {
 
   const profileOptions = data.profiles
```
CarinaWolli, line 814: We need to disable the `CreateButton` if we don't have any profileOptions. Now I am getting an error when clicking on the 'New' button `Uncaught TypeError: Cannot read properties of undefined (reading 'slug')`

CarinaWolli (changes requested):
Found two more small things that are blocking 🙏

We need to disable the toggle inside event type settings too
 Screenshot 2024-03-11 at 11 22 51 AM

apps/web/modules/event-types/views/event-types-listing-view.tsx
sean-brydon: Never even thought of this - nice spot

CarinaWolli (dismissed):
Looking good now, nice work 👏🏻

CarinaWolli:
@sean-brydon can you fix the merge conflicts?

CarinaWolli (approved):
re-approve after fixing merge conflicts

PR #14008, opened 2024-03-07
fix: add future team members toggle

ibex088:
## What does this PR do?
Fixes # (issue)

image_6487327


packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -202,6 +203,23 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
     data.avatarUrl = null;
     data.avatar = null;
   }
+  if (input.completedOnboarding) {
+    const userTeams = await prisma.user.findFirst({
+      where: {
+        id: user.id,
+      },
+      select: {
+        teams: true,
```
Udit-takkar, line 212: how about just getting the team id as we don't need all the columns?

teams : {
 id: true
}

SomayChauhan: Absolutely

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -202,6 +203,23 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
     data.avatarUrl = null;
     data.avatar = null;
   }
+  if (input.completedOnboarding) {
+    const userTeams = await prisma.user.findFirst({
```
keithwillcode, line 207: Could we just make `team` the primary table and have a where for `userId`?

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -202,6 +203,27 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
...
+        },
+      },
+    });
+    if (userTeams && userTeams.teams.length > 0) {
+      await Promise.all(
+        userTeams.teams.map(async (team) => {
+          await updateNewTeamMemberEventTypes(user.id, team.id);
+        })
+      );
+    }
```
Udit-takkar, line 225: updateProfile mutation is used in a lot of pages. why do we want to repeat this  updateNewTeamMemberEventTypes step everytime

packages/trpc/server/routers/viewer/teams/inviteMember/inviteMember.handler.ts
```diff
@@ -223,6 +223,11 @@ async function handleExistingUsersInvites({
         parentId: team.parentId,
         accepted: false,
       });
+      await Promise.all(
+        regularUsers.map(async (user) => {
+          await updateNewTeamMemberEventTypes(user.id, team.id);
+        })
+      );
```
Udit-takkar, line 230: Here regularUsers haven't accepted team invite so why are we assigning them to event types?

packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
SomayChauhan: only doing it when if `input.completedOnboarding` is true
which is called only once i think.
this will only run once the user has completed onboarding
couldn't think of a better place to do it

Udit-takkar: Oh okay. btw we already use updateNewTeamMemberEventTypes in acceptOrLeaveHandler.ts so whenever user accept invite. was that not working properly?

Udit-takkar (approved):

joeauyeung (commented):
I don't want to block this PR but was `updateNewTeamMemberEventTypes` in `packages\lib\server\queries\teams\index.ts` not working properly? It seems like it was only being called in `inviteMember.handler` were there other pathways for inviting team/org members we were missing?

joeauyeung (commented):
I don't want to block this PR but was `updateNewTeamMemberEventTypes` in `packages\lib\server\queries\teams\index.ts` not working properly? It seems like it was only being called in `inviteMember.handler` were there other pathways for inviting team/org members we were missing?

SomayChauhan:
> I don't want to block this PR but was `updateNewTeamMemberEventTypes` in `packages\lib\server\queries\teams\index.ts` not working properly? It seems like it was only being called in `inviteMember.handler` were there other pathways for inviting team/org members we were missing?


This PR addresses the scenario in which a new team member is added directly to a sub-team (user doesn't already exist within the organization). In this case, the function `handleExistingUsersInvites` where   `updateNewTeamMemberEventTypes` is called does not execute.

PR #14027, opened 2024-03-08
fix: Dsync bugs

joeauyeung:
## What does this PR do?

This PR aims to fix some dsync bugs that were missed in the reviews

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

## Mandatory Tasks

## Checklist


packages/trpc/server/routers/viewer/dsync/delete.handler.ts
```diff
@@ -27,7 +27,7 @@ export const deleteHandler = async ({ ctx, input }: Options) => {
     });
   }
 
-  await prisma.dSyncData.delete({
+  await prisma.dSyncData.findMany({
```
Udit-takkar, line 30: what is the user of findMany here ?

CarinaWolli: Also not sure what `findMany` does here, then we don't delete anything? Would `deleteMany` solve it? that shouldn't throw an error if it doesn't exist

joeauyeung: The dsync data still exists in the SAML DB which is handled with the Jackson package.

joeauyeung: I was blanking so hard on this. It was supposed to be deleteMany

CarinaWolli (approved):

PR #14034, opened 2024-03-08
fix: username inconsistency

### LINKED ISSUE #14037: [CAL-3164] Profile update causes user.username to be changed to profile.username
1. Invite an eixsting user teampro to Acme. The org username is teampro-example now. But user.username is still teampro
2. Go to profile, upload an avatar and save.
3. Now, user.username also changes to teampro-example.
4. This isn't a bug at the moment but is unexpected.
5. Now, remove the user from the organization. At this moment, the old link(app.cal.local/teampro) goes 404 for user because the username has now changed to teampro-example. This is a bug.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3164](https://linear.app/calcom/issue/CAL-3164/profile-update-causes-userusername-to-be-changed-to-profileusername)</sub>

ibex088:
## What does this PR do?

Fixes #14034

### DIFF
--- a/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
@@ -102,6 +102,27 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
         throw new TRPCError({ code: "BAD_REQUEST", message: t("username_already_taken") });
       }
     }
+  } else if (input.username && user.organizationId && user.movedToProfileId) {
+    const username = slugify(input.username);
+    const movedToProfile = await prisma.profile.findUnique({
+      where: {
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/loggedInViewer/updateProfile.handler.ts
```diff
@@ -102,6 +102,28 @@ export const updateProfileHandler = async ({ ctx, input }: UpdateProfileOptions)
         throw new TRPCError({ code: "BAD_REQUEST", message: t("username_already_taken") });
       }
     }
+  } else if (input.username && user.organizationId && user.movedToProfileId) {
```
hariombalhara, line 105: I think we can simplify it by simply removing `username` from the data as username isn't allowed to be changed for an organization profile.

emrysal (approved):
👍

PR #14039, opened 2024-03-08
fix: event length in webhook payload

CarinaWolli:
## What does this PR do?

Fixes event length in webhooks and zapier payload if event type has multiple durations. It used to show the default location instead

Issue was reported in Zapier.


emrysal (approved):
🚀

PR #14041, opened 2024-03-08
fix: org banner

Udit-takkar:
## What does this PR do?

Fixes # (issue)
Team logo wasn't using Avatar before when i first implemented org banner. so now need to add one more column to make the avatar unique

## Type of change


emrysal (dismissed):
In the future if we need to add another type of image, we should consider making this an enum - I don't have a significant issue with this implementation however.

emrysal (dismissed):

emrysal (approved):

PR #14042, opened 2024-03-09
fix: email displayed as organizer event type settings adjustments

### LINKED ISSUE #14001: [CAL-3268] Event type: Changing the email displayed as the Organizer copy adjustments
There's two things we need to update here.

In the implementation, 

1. there's an extra label on the input that implies we're just sending details to this email address. We are not just sending the email, we are using this person as the organiser. That means most importantly, that we will show that email as the organizer to your booker - thats' the main reason people want this feature. Simply removing that label makes it clearer
2. I know why we added this help text, but we need to make the tooltip actually helpful for the feature as per designs and find another solution for this text. My proposal is we show a toast the first time someone toggles this and never again.
3. We're missing additional help text below the input when this is toggled on

**Current**

**Designs**
CleanShot 2024-03-07 at 09 16 44@2x
CleanShot 2024-03-07 at 09 16 52@2x

Please also pay specific attention to the spacing of the input that appears below when toggled and match it to designs.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3268](https://linear.app/calcom/issue/CAL-3268/event-type-changing-the-email-displayed-as-the-organizer-copy)</sub>

Amit91848:
## What does this PR do?

Fixes #14001 

Screenshot 2024-03-09 at 1 56 09 PM
Screenshot 2024-03-09 at 1 56 19 PM

## Type of change

## Mandatory Tasks


Amit91848:
Recreated here #14102

PR #14049, opened 2024-03-11
feat: enable user control over timezone changes and helpers for query params

Ryukemeister:
## What does this PR do?

- Adds util functions to get and set query parameters of any page without using `next/link` (mostly meant to be used for atoms) 
- Adds props and logic to handle timezone updation preference of a user and handle timezone change based on that


Ryukemeister:
Just a heads up this needs to be reviewed only after #13934 is merged :)

packages/platform/atoms/cal-provider/CalProvider.tsx
```diff
@@ -10,12 +10,26 @@ export type CalProviderProps = {
   clientId: string;
   accessToken: string;
   options: { refreshUrl?: string; apiUrl: string };
+  timezoneUpdatePreference?: "auto" | "none";
```
supalarry, line 13: I think instead of having "timezoneUpdatePreference" is clearer to have "updateTimezoneAutomatically: boolean" with default set to true - because "timezoneUpdatePreference" right now is optional, so if it would not be passed, is it the same as passing "none" and having optional argument with "none" option that disables its functionality IMO is confusing.

packages/platform/atoms/cal-provider/BaseCalProvider.tsx
```diff
@@ -23,7 +30,13 @@ export function BaseCalProvider({ clientId, accessToken, options, children }: Ca
     [mutateAsync]
   );
 
-  useTimezone(handleTimezoneChange);
+  useTimezone(
+    onTimezoneChange
+      ? onTimezoneChange
+      : timezoneUpdatePreference === "auto"
+      ? handleTimezoneChange
+      : undefined
```
supalarry, line 38: Having multiple ternary operators is confusing. I think it's better to have a separate function that gets the handler and then pass it to useTimezone:

```
  const getTimezoneChangeHandler = useCallback(() => {
    if (onTimezoneChange) return onTimezoneChange;
    if (timezoneUpdatePreference === "auto") return handleTimezoneChange;
    return undefined;
  }, [onTimezoneChange, timezoneUpdatePreference, handleTimezoneChange]);

  useTimezone(getTimezoneChangeHandler());
  ```

File: packages/platform/atoms/lib/getQueryParam.ts [hunk 3]
```diff
@@ -0,0 +1,14 @@
...
+  if (typeof window !== "undefined") {
+    const currentUrl = new URL(window.location.href);
+    const searchParams = currentUrl.searchParams;
+
+    const paramater = searchParams.get(paramName);
+
+    return paramater;
+  }
+
+  return defaultParam;
```
supalarry, line 11: If we are not running this on browser (window is undefined) and don't pass defaultParam, why do we return "setup" - like what does "setup" mean?

packages/platform/atoms/cal-provider/CalProvider.tsx
Ryukemeister: True it's better to have updateTimezoneAutomatically prop with a boolean, makes it less confusing

packages/platform/atoms/lib/getQueryParam.ts
Ryukemeister: `setup` is the name of the default tab that we display in the event-type settings

supalarry: Then we should somehow denote that the getQueryParam is for event-type settings, because what if it is used somewhere else? Then having "setup" as a default query param seems random.

Ryukemeister: What if we make the defaultParam prop as not optional? that way it won't be limited to just event-type settings right?

supalarry: Idk i would just remove defaultParam and if whatever param is requested return undefined and then its up to the user to decide what to do

supalarry (approved):

PR #14062, opened 2024-03-12
fix: Can reschedule booking to different event type

keithwillcode:
## What does this PR do?

Blocks users from trying to reschedule bookings and manually setting a different eventTypeId

## Type of change

## How should this be tested?

- Ensure normal bookings can be made
- Ensure bookings can be rescheduled
- Ensure dynamic bookings can be rescheduled

## Mandatory Tasks


emrysal (dismissed):
That's simpler than I expected.

keithwillcode:
@emrysal Yeah thankfully https://github.com/calcom/cal.com/pull/14062/files#diff-a7c21727d9a0d4d1aaff9d24fca9293bc28468861bc8022dfa03116908b6fae7R1152 throws an error if the booking isn't found so if an incorrect eventTypeId is passed in that doesn't match the original booking, it will throw an error.

PR #14067, opened 2024-03-12
fix: Get `dsyncData` based on database

joeauyeung:
## What does this PR do?

This PR sets a unique tenant when creating `dsyncData` for an organization. It also relies on the directory ID present in the database rather than the tenant.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a new dsync directory in an org
- Delete the record in the DB
- It should not show the dsync data for that org

## Mandatory Tasks

## Checklist


packages/features/ee/dsync/components/ConfigureDirectorySync.tsx
```diff
@@ -39,7 +40,7 @@ const ConfigureDirectorySync = ({ organizationId }: { organizationId: number | n
     return <SkeletonLoader />;
   }
 
-  const directory = data ? data[0] : null;
+  const directory = data ? data : null;
```
joeauyeung, line 42: The `get` method returns a single object rather than an array

packages/trpc/server/routers/viewer/dsync/create.handler.ts
```diff
@@ -30,7 +30,9 @@ export const createHandler = async ({ ctx, input }: Options) => {
     });
   }
 
-  const tenant = organizationId ? `${tenantPrefix}${organizationId}` : (samlTenantID as string);
+  const organization = ctx.user.organization;
+
+  const tenant = input.organizationId ? `${organization.slug}-${organization.id}` : (samlTenantID as string);
```
joeauyeung, line 35: Set the tenant to the unique org slug and the id

packages/trpc/server/routers/viewer/dsync/get.handler.ts
```diff
@@ -26,9 +27,23 @@ export const getHandler = async ({ ctx, input }: Options) => {
...
+      code: "BAD_REQUEST",
+      message: "Organization ID is required",
+    });
+  }
+
+  const dsyncData = await prisma.dSyncData.findUnique({
+    where: {
+      organizationId: input.organizationId,
+    },
+  });
```
joeauyeung, line 42: Only render the dsync data if it exists in our DB. I was getting hard blocked because it would get the correct dsync data but it wasn't associated with an org.

packages/features/ee/dsync/components/ConfigureDirectorySync.tsx
```diff
@@ -39,7 +39,7 @@ const ConfigureDirectorySync = ({ organizationId }: { organizationId: number | n
...
```
zomars, line 42:
```suggestion
  const directory = data ?? null;
```

packages/trpc/server/routers/viewer/dsync/create.handler.ts
```diff
@@ -30,7 +30,9 @@ export const createHandler = async ({ ctx, input }: Options) => {
     });
   }
 
-  const tenant = organizationId ? `${tenantPrefix}${organizationId}` : (samlTenantID as string);
+  const organization = ctx.user.organization;
```
zomars, line 33:
```suggestion
  const { organization } = ctx.user;
```

emrysal (approved):
Looks cleaner to me

PR #14071, opened 2024-03-13
chore: update to see webhook creation error handling to see error in prod

sean-brydon:
Pr adds a log so we can see the error that happens here in prod. 

response doesnt throw if response.ok is false -> This error was never being hit.


packages/app-store/paypal/lib/Paypal.ts
```diff
@@ -191,17 +191,16 @@ class Paypal {
       ],
     };
 
-    try {
```
CarinaWolli, line 194: can we keep the try/catch here?

packages/app-store/paypal/lib/Paypal.ts
```diff
@@ -197,12 +197,14 @@ class Paypal {
...
+      if (!response.ok) {
+        throw new Error(await response.json());
       }
-    } catch (error) {
-      console.error(error);
+
+      const result = await response.json();
+      return result.id as string;
+    } catch (e) {
+      console.error(e);
```
Udit-takkar, line 207: can't we use logger from /lib/logger?

sean-brydon: Good shout

Udit-takkar (dismissed):

packages/app-store/paypal/lib/Paypal.ts
```diff
@@ -197,12 +198,14 @@ class Paypal {
         body: JSON.stringify(body),
       });
 
-      if (response.ok) {
-        const result = await response.json();
-        return result.id as string;
+      if (!response.ok) {
+        throw new Error(await response.json());
```
CarinaWolli, line 202: Should we add a message string in here, something like: 
```
const message = `${response.statusText}: ${JSON.stringify(await response.json())}`;
throw new Error(message);
```

Udit-takkar: Done

CarinaWolli (approved):

PR #14072, opened 2024-03-13
fix: Avoid CSP related warnings in console on booking pages.

### LINKED ISSUE #13844: [ERROR ]Refused to execute inline script because it violates the following Content Security Policy
Hi guys.
I have these errors:

[Report Only] Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'nonce-LeYI+W5S6XECpdjnp7IWAg==' 'strict-dynamic' 'self' 'unsafe-inline' https:". Note that 'unsafe-inline' is ignored if either a hash or nonce value is present in the source list.

[Report Only] Refused to load the image 'https://cal.com/api/avatar/cc9f080a-90ea-4d86-8c12-e56bc6d4798e.png' because it violates the following Content Security Policy directive: "img-src 'self' https://app.cal.com https://img.youtube.com https://eu.ui-avatars.com/api/ data:".

image

---

I realize that they have no effect on the functionality of the website and widget, and everything works. 

BUT! This is very bad for website optimization, and Google PageSpeed Insights reminds about these errors as well. How to solve them?

Since I see and know that these errors are also present in your direct link.

**Is there anything I can do to prevent these errors?**

hariombalhara:
## What does this PR do?

Fixes #13844

We are disabling Report only CSP mode now. It was added so that we can observe the errors that would come if we enable CSP and work towards fixing them. As, we are not actively working towards fixing them, it doesn't make sense to keep the mode enabled.

These are some of the reasons we get CSP warnings on booking pages as of now.
1. Analytics scripts
2. Embeds use app.cal.com which access cal.com which isn't whitelisted.

## Type of change

## How should this be tested?
 Set CSP_POLICY='non-strict' in .env
- After that visit app.cal.com/teampro and notice that it is serving CONTENT-SECURITY-POLICY-REPORT-ONLY header in main
- In this branch, the header shouldn't be served.

## Mandatory Tasks

## Checklist


.env.example
```diff
@@ -232,6 +232,7 @@ NEXT_PUBLIC_COMPANY_NAME="Cal.com, Inc."
 # NEXT_PUBLIC_DISABLE_SIGNUP=true
 NEXT_PUBLIC_DISABLE_SIGNUP=
 
+# Set this to 'non-strict' to enable CSP for support pages. 'strict' isn't supported yet. Also, check the README for details.
```
hariombalhara, line 235: Missing documentation

apps/web/lib/csp.ts
```diff
@@ -84,7 +84,14 @@ export function csp(req: IncomingMessage | NextRequest | null, res: OutgoingMess
     const enforced =
       "cache" in req ? req.headers.get("x-csp-enforce") === "true" : req.headers["x-csp-enforce"] === "true";
 
-    const name = enforced ? "Content-Security-Policy" : "Content-Security-Policy-Report-Only";
+    // No need to enable REPORT ONLY mode for CSP unless we start actively working on it. See https://github.com/calcom/cal.com/issues/13844
+    const name = enforced ? "Content-Security-Policy" : /*"Content-Security-Policy-Report-Only"*/ null;
+
+    if (!name) {
+      return {
```
hariombalhara, line 91: This is the same return value we do in similar cases above.

emrysal (approved):
👍

PR #14076, opened 2024-03-14
fix: url query params dynamic url server side app router

alannnc:
## What does this PR do?

While developing/testing a feature for orgs I found out that dynamic pages that were using getServerSideProps() where receiving context->query->user->**favicon.ico**

This looks like an error on nextjs app side but can be easily fixed adding favicon.ico to the root of app/

before:
image
after:
image

Related links:
[Issue](https://github.com/vercel/next.js/discussions/13301#discussioncomment-6892785)
[Official docs](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons#favicon)

Fixes # (issue)


zomars (approved):
🤯

PR #14077, opened 2024-03-14
feat: New app install flow 2

ibex088:
## What does this PR do?

Take over from @ThyMinimalDev existing work and complete the following tasks

* When clicking on install new app button, only redirect to the new app install flow if it is an oAuth app or an app that extends the EventType feature in it's config

* adapt oauth apps api callback / add to redirect to new app install flow


packages/app-store/_components/OmniInstallAppButton.tsx
```diff
@@ -21,11 +24,13 @@ export default function OmniInstallAppButton({
   className,
   returnTo,
   teamId,
+  eventType,
 }: {
   appId: string;
   className: string;
   returnTo?: string;
   teamId?: number;
+  eventType: EventTypeAppSettingsComponentProps["eventType"];
```
joeauyeung, line 33: I notice you're passing the whole event type object but we're only using the event type id. I think we should just pass the id down.

packages/app-store/_components/AppCard.tsx
```diff
@@ -34,6 +35,7 @@ export default function AppCard({
   disableSwitch?: boolean;
   switchTooltip?: string;
   hideSettingsIcon?: boolean;
+  eventType: EventTypeAppSettingsComponentProps["eventType"];
```
joeauyeung, line 38: See my comment on `OmniInstallAppButton` we should only pass down the event type id.

packages/lib/apps/getAppOnboardingRedirectUrl.ts
```diff
@@ -0,0 +1,13 @@
+import { AppOnboardingSteps } from "@calcom/lib/apps/appOnboardingSteps";
+
+import { getAppOnboardingUrl } from "./getAppOnboardingUrl";
+
+export const getAppOnboardingRedirectUrl = (slug: string, teamId?: number, eventTypeId?: number) => {
```
joeauyeung, line 5: I notice we're using this function across every callback function. Could we instead generate this redirect URL and add it to the OAuth state to be read on at the callback endpoint? 

The reason why is when we're developing new apps it's extra mental overhead to remember to add the `getAppOnboardingRedirectUrl` function to every callback endpoint. Whenever we can reduce mental overhead, reduces the possibilities of bugs.

joeauyeung (changes requested):
Great work so far @SomayChauhan. I left a few comments that need to be addressed first. @ThyMinimalDev @sean-brydon can you read through my review comments and let me know if you object to anything. 

## Blocking
- [x]  On app install page it's not respecting dark theme
- [x] On configure step, inputs aren't showing errors (ex. Non empty values for text inputs)
- [x] I don't think calendars should be a part of the flow. They're still only avaialble on a individual scope and not a team scope
- [x] For OAuth flow I don't think we need the OAuth step page. I think once you choose the account to install the app on it should immedietly take you to the app provider's OAuth page
- [x] When coming back from the OAuth provider's page, it redirects me back to the App Store. It should redirect back to choosing which event type to enable Stripe on and the config
    - [x] I just tested this again for a team and it brought me to the event type selection page. Although when I hit next it didn't bring me to the config step but back to the app store

## Non-blocking
- [x] Should be able to choose multiple event types
- [x] Not a huge fan that the new flow is named "onboarding". It might cause some confusion with our actual onboarding wizard. Can we maybe call the pathname "/apps/installation/"

image

apps/web/pages/apps/onboarding/[[...step]].tsx
```diff
@@ -255,14 +255,15 @@ const OnboardingPage = ({
...
       data-testid="onboarding"
-      style={
-        {
-          "--cal-brand": "#111827",
-          "--cal-brand-emphasis": "#101010",
-          "--cal-brand-text": "white",
-          "--cal-brand-subtle": "#9CA3AF",
-        } as CSSProperties
-      }>
+      // style={
```
Comment (SomayChauhan) [inline apps/web/pages/apps/onboarding/[[...step]].tsx:258]: not sure why it was added, but this is causing some dark mode ui issues

apps/web/pages/apps/onboarding/[[...step]].tsx
```diff
@@ -512,15 +512,20 @@ export const getServerSideProps = async (context: GetServerSidePropsContext) =>
         if (!hasEventTypes) {
           throw new Error(ERROR_MESSAGES.appNotExtendsEventType);
         }
-        eventTypes = await getEventTypes(user.id, parsedTeamIdParam);
+        const _eventTypes = await getEventTypes(user.id, parsedTeamIdParam);
+
+        if (!_eventTypes) {
+          return { redirect: { permanent: false, destination: `/apps/installed/${appMetadata.category}` } };
```
Comment (SomayChauhan) [inline apps/web/pages/apps/onboarding/[[...step]].tsx:518]: if user/team does not have event-type then skip configure step and goto apps/installed/[category]

SomayChauhan:
- [ ] edge case: multiple payment apps for a event type

joeauyeung (changes requested):
Great work so far @SomayChauhan. There are still some points that need to be addressed.

## Blocking
- [x] Not seeing an error message on the installation flow. For example, I'm installing GTM and when I leave "TrackingId" blank I see an error on the backend but not on the frontend.
- [x] I still don't think we need this setup page for OAuth apps. After selecting which account to install it on it should just bring you to the 3rd party OAuth page
image
- [x] On the install flow, it's starting at step 0 instead of step 1
image
- [ ] For stripe, after OAuth install and choosing an event type, when saving the data I'm getting an error on the backend that the `credentialId` isn't being passed. For payment apps and any event type app that requires a `credentialId` we need to ensure that this data is being passed


## Non-blocking
- [x] @ThyMinimalDev @sean-brydon I'm not a huge fan of how when installing an app via the Apps tab on the event type page it jumps you to the installation wizard. Seems pretty jarring being jumped around screens like that. For OAuth though I can understand.
https://www.loom.com/share/1d82fdda4f4b4f9bb1b13c3c2195e7ab

apps/web/components/apps/installation/ConfigureStepCard.tsx
```diff
@@ -28,18 +28,22 @@ export const ConfigureStepCard: FC<ConfigureStepCardProps> = ({
...
   credentialId,
   eventType,
-  onSave,
   loading,
 }) => {
   const { t } = useLocale();
   const { getAppDataGetter, getAppDataSetter } = useAppsData();
-  const { shouldLockDisableProps } = useLockedFieldsManager(
+  const formMethods = useForm();
+  const { shouldLockDisableProps } = useLockedFieldsManager({
```
joeauyeung, line 36: @ThyMinimalDev @sean-brydon, do we need the useLockedFieldsManager in the app installation flow? I see that apps are locked as a whole for managed event types and not on the app level. This is testing with GTM. I think if we pass the right data through here we don't need to use this function and worry about form methods. The formMethds taht useLockedFieldsManager expects is for a whole event type.

apps/web/components/apps/installation/ConfigureStepCard.tsx
```diff
@@ -1,61 +1,158 @@
...
+              />
+            )
+          );
+        })}
+      </div>
+      <Button
+        className="text-md mt-6 w-full justify-center"
+        onClick={() => {
+          // submitRefs.current[0].current?.click();
+          submitRefs.current.map((ref) => ref.current?.click());
```
SomayChauhan, line 151: @joeauyeung 
is there a better way to handle this?
its getting kinda messy?


the flow so far - 
[LOOM](https://www.loom.com/share/d6e517832a5c47cf8fdf313ba165e0c1?sid=b3873b0d-d50a-4ac9-a814-e50be0bf6602)

joeauyeung: I think this is awesome. Allowing users to configure multiple event types is going to help a ton. This gets my seal of approval. Thoughts @ThyMinimalDev @sean-brydon @ciaranha @itsberkaya?

SomayChauhan:
https://www.loom.com/share/85fa766b64e043fe864c29551305b365?sid=e38f657d-e08f-43b5-8a02-4105860f0160

apps/web/components/apps/installation/ConfigureStepCard.tsx
```diff
@@ -0,0 +1,211 @@
...
+        <Button
+          className="text-md mt-6 w-full justify-center"
+          type="button"
+          onClick={(e) => {
+            submitRefs.current.reverse().map((ref) => ref.current?.click());
+            setSubmit(true);
+          }}
+          loading={loading}>
+          Save
+        </Button>
```
joeauyeung, line 206: Can you explain what these two buttons are for? Why is one of them hidden?

apps/web/components/apps/installation/StepFooter.tsx
joeauyeung, line 1: Do we need this component in `[[...step]].tsx`? Commenting this out I didn't see a difference in UI
image

If we're not adding a "Go Back" button we should also remove the commented code.

apps/web/pages/apps/installation/[[...step]].tsx
```diff
@@ -0,0 +1,509 @@
...
+    onError: (err) => {
+      let message = "";
+      if (err instanceof HttpError) {
+        const message = `${err.statusCode}: ${err.message}`;
+        showToast(message, "error");
+      }
+
+      if (err.data?.code === "UNAUTHORIZED") {
+        message = `${err.data.code}: ${t("error_event_type_unauthorized_update")}`;
+        message = `${err.data.code}: ${t(err.message)}`;
```
Comment (joeauyeung) [inline apps/web/pages/apps/installation/[[...step]].tsx:146]: Message is being assigned two different values back to back. Which one should we use?

apps/web/pages/apps/installation/[[...step]].tsx
```diff
@@ -0,0 +1,509 @@
...
+      className="dark:bg-brand dark:text-brand-contrast text-emphasis min-h-screen px-4"
+      data-testid="onboarding"
+      // style={
+      //   {
+      //     "--cal-brand": "#111827",
+      //     "--cal-brand-emphasis": "#101010",
+      //     "--cal-brand-text": "white",
+      //     "--cal-brand-subtle": "#9CA3AF",
+      //   } as CSSProperties
+      // }
```
Comment (joeauyeung) [inline apps/web/pages/apps/installation/[[...step]].tsx:214]:
```suggestion

```

apps/web/pages/apps/installation/[[...step]].tsx
```diff
@@ -0,0 +1,509 @@
...
+      router.push("/apps");
+    },
+  });
+
+  const handleSelectAccount = async ({ id: teamId }: onSelectParams) => {
+    try {
+      setIsSelectingAccount(true);
+      if (appMetadata.isOAuth) {
+        const state = JSON.stringify({
+          appOnbaordingRedirectUrl: getAppOnboardingRedirectUrl(appMetadata.slug, teamId),
```
Comment (joeauyeung) [inline apps/web/pages/apps/installation/[[...step]].tsx:163]:
```suggestion
          appOnboardingRedirectUrl: getAppOnboardingRedirectUrl(appMetadata.slug, teamId),
```
When fixing the typo here. Ensure that other parts of the app aren't expecting the typo'd value

apps/web/pages/apps/installation/[[...step]].tsx
```diff
@@ -0,0 +1,509 @@
...
+      //   {
+      //     "--cal-brand": "#111827",
+      //     "--cal-brand-emphasis": "#101010",
+      //     "--cal-brand-text": "white",
+      //     "--cal-brand-subtle": "#9CA3AF",
+      //   } as CSSProperties
+      // }
+    >
+      <Head>
+        <title>Install {appMetadata?.name ?? ""}</title>
```
Comment (joeauyeung) [inline apps/web/pages/apps/installation/[[...step]].tsx:217]: Let's add a translation here

packages/app-store/_utils/oauth/decodeOAuthState.ts
```diff
@@ -7,6 +7,9 @@ export function decodeOAuthState(req: NextApiRequest) {
     return undefined;
   }
   const state: IntegrationOAuthCallbackState = JSON.parse(req.query.state);
+  if (state.appOnbaordingRedirectUrl) {
+    state.appOnbaordingRedirectUrl = decodeURIComponent(state.appOnbaordingRedirectUrl);
+  }
```
joeauyeung, line 12: Let's fix the typos here as mentioned earlier

joeauyeung (changes requested):
Great job so far @SomayChauhan feels like we're getting closer to getting this merged.

## Blocking
- I left some comments about a type in `state.appOnbaordingRedirectUrl` which should be `state.appOnboardingRedirectUrl`. Let's clean that up across different files.
- When selecting to install an app on a team
	- It highlights both the team and the user on the accounts step
	- On the event types step, even when a team is selected, the user's personal event types appear and no the team's
	- https://www.loom.com/share/9241b5dc2aad43b5b3fbc84f120eca70

apps/web/components/apps/installation/ConfigureStepCard.tsx
SomayChauhan: here in this scenario we have nested forms, and we want to submit the outer form only when all the inner forms have been filled out.
in this case these are the 2 individual inner forms
Screenshot 2024-04-03 at 6 17 09 AM

`submitRefs.current.reverse().map((ref) => ref.current?.click());` is iterating over the inner forms and triggering their submission. However, we want to ensure that the outer form is only submitted when all the inner forms contain values. To achieve this, you need to keep track of whether all the inner forms are filled out. this is what `updatedEventTypesStatus` state is for.

the hidden button is the actual submit button for the outer form which we trigger when all the inner form values have been filled out and saved.

apps/web/components/apps/installation/ConfigureStepCard.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  useEffect(() => {
+    const appDataSetter = getAppDataSetter(slug as EventTypeAppsList, categories, credentialId);
+    appDataSetter("enabled", true);
+  }, []);
+
+  return (
+    <EventTypeAppSettings
+      slug={slug}
+      disabled={shouldLockDisableProps("apps").disabled}
```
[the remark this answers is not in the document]
SomayChauhan: yes, removed 💯

joeauyeung (approved):
Went through everything and it looks good to me! Great job @SomayChauhan!

Let's chat with @ThyMinimalDev and @sean-brydon to see what's the best way to get this merged.

PR #13591, opened 2024-02-08
fix: view team bookings

ibex088:
## What does this PR do?

Fixes #13582

PR #14079, opened 2024-03-14
fix: view team bookings

### LINKED ISSUE #13582: [CAL-3089] View Bookings of Team

### Issue Summary

Reported on discord
 [https://discord.com/channels/1125390093386010654/1204544773697044591/1204544773697044591](https://discord.com/channels/1125390093386010654/1204544773697044591/1204544773697044591)

I created a team and added a few members. Then I added a few event types to those members. When customers started to book events I was expecting to see "Upcoming" events listed. However, I cannot see the bookings of my team members, only my own...I know they exist because my team member can see upcoming bookings but I cannot!

image

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3089](https://linear.app/calcom/issue/CAL-3089/view-bookings-of-team)</sub>

ibex088:
## What does this PR do?

re-doing #13591 - that was reverted due to DB load issues introduced.

Fixes #13582


packages/trpc/server/routers/viewer/bookings/get.handler.ts
joeauyeung, line 1: From this comment, I'm not seeing any `select` fields in the queries. https://github.com/calcom/cal.com/pull/13591#pullrequestreview-1931896242

Let's figure out the minimum amount of data we need from the DB for this endpoint.

joeauyeung (changes requested):

packages/trpc/server/routers/viewer/bookings/get.handler.ts
SomayChauhan: how about now?

joeauyeung (dismissed):
@SomayChauhan I like this commit: [`0aa47bd` (#14079)](https://github.com/calcom/cal.com/pull/14079/commits/0aa47bde45f6defebfbc3a132f48ef85b45963b4)

I'll approve for now, but let's see what @keithwillcode and @zomars have to say. 

As a follow up I found two things that we can improve on:
- `getUniqueBookings` filters duplicate bookings based on `uid` but when getting the nested fields from the bookings we filter on `id`. We should change `getUniqueBookings` to filter on just `id`s to reduce the amount of data we need to initially query the database for.
- Optimize `getUniqueBookings`
	- Right now it loops through the array of bookings theoretically two times but I think we can cut it down to once
```
const getUniqueBookings = <T extends { uid: string }>(arr: T[]) => {
	const  unique = new Set();
    arr.forEach((booking) => {
    set.add(booking.uid);
  	});
  return [ ...unique.values() ];
};
```

zomars:
For some reason e2e tests are broken. Will revisit after release.

emrysal (approved):
Tests passing, looks ready to go! :)

PR #14093, opened 2024-03-14
fix: don't show 'send event details to' for team events

CarinaWolli:
## What does this PR do?

Don't show 'send event details to' for team events. Also, only show if user has a secondary email to choose from

### DIFF
--- a/apps/web/components/eventtype/EventAdvancedTab.tsx
+++ b/apps/web/components/eventtype/EventAdvancedTab.tsx
@@ -216,7 +216,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps, 
             />
           </div>
         )}
-        {!useEventTypeDestinationCalendarEmail && verifiedSecondaryEmails.length > 0 && (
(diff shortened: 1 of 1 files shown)


apps/web/components/eventtype/EventAdvancedTab.tsx
```diff
@@ -216,7 +216,7 @@ export const EventAdvancedTab = ({ eventType, team }: Pick<EventTypeSetupProps,
             />
           </div>
         )}
-        {!useEventTypeDestinationCalendarEmail && verifiedSecondaryEmails.length > 0 && (
+        {!useEventTypeDestinationCalendarEmail && verifiedSecondaryEmails.length > 1 && !team && (
```
CarinaWolli, line 219: primary email of user is also part of `verifiedSecondaryEmails` see line https://github.com/calcom/cal.com/pull/14093/files#diff-e6807e04030a7e08ae8814936a8b27eba0efbb12277eba305e2bda7e25a42396R132

We don't need to show this if there is only the primary email to choose from

alishaz-polymath (approved):
LGTM

Issue #14103, opened 2024-03-15
[CAL-3278] Delete/Create WebhookScheduledTriggers for existing bookings


CarinaWolli:
* When a new webhook with 'Meeting Ended' trigger is created (or trigger is added to an existing webhook) we need to create WebhookScheduledTriggers for all existing bookings.
* When a webhook with 'Meeting Ended' trigger is deleted (or trigger is removed from an existing webhook) we need to delete all WebhookScheduledTriggers that belong to this webhook.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3278](https://linear.app/calcom/issue/CAL-3278/deletecreate-webhookscheduledtriggers-for-existing-bookings)</sub>

PR #14104, opened 2024-03-15
fix: failing webhookTriggers cron job

CarinaWolli:
## What does this PR do?

webhookTriggers cron job is failing right now with the following error: 
`NotFoundError [PrismaClientKnownRequestError]: No Webhook found`

This can happen because right now we don’t delete WebhookScheduledTriggers when the webhook itself is deleted. Created another issue for that https://github.com/calcom/cal.com/issues/14103


emrysal (approved):
👍

PR #14105, opened 2024-03-15
fix: Adding 'hosts' instead of 'users'

emrysal:
## What does this PR do?

The listing was previously based on users, which usually just contains the createdBy user - this amends the behaviour to how it used to be.


joeauyeung (approved):
LGTM

Tested by creating a team event and assigning other members. Only the assigned members and not the creator of the event appeared under the event type. 
Created by Team Pro user and assigned to Team Pro Example 2
image

PR #14107, opened 2024-03-15
fix: improve defined slot start times

CarinaWolli:
## What does this PR do?

If event interval is not dividable by any of the defined starting times (60, 30, 20, 15, 10, 5) the first slots of a date range starts at any arbitary minute. We used to always have the first slot start on a 15 minute slot. 

For 5 minute interval events we would also always start the first slot of a date range on a 15 minutes slot. But now we start at any 5 minute slot, to avoid having missing slots. 

More details can be found in my comments


packages/lib/slots.ts
```diff
@@ -162,24 +162,23 @@ function buildSlotsWithDateRanges({
   offsetStart = offsetStart ? minimumOfOne(offsetStart) : 0;
   const slots: { time: Dayjs; userIds?: number[] }[] = [];
 
+  let interval = Number(process.env.NEXT_PUBLIC_AVAILABILITY_SCHEDULE_INTERVAL) || 1;
```
CarinaWolli, line 165: Let a slot start at any minute if the frequency (timeslot interval) is not dividable by any of the defined starting times (60, 30, 20, 15, 10, 5). 

The main thing that will be changed is that the first slot of a date range from a 5 min event now doesn't need to start at 15 minute slot no more. This caused 'missing' slots, here is an example:
- 1:00, 1:05, 1:10, 1:15 is available 
- 1:00 gets booked
- With the old logic 1:15 would be the first available slot now
- With the new logic 1:05 is the first available slot

What also changes: 
- Slots with a timeslot interval (duration) of, for example, 7 minutes will now start at any arbitrary minute and not at a 15 minute slot.

joeauyeung (changes requested):
When I added a 5 minute buffer before or after, the next slot jumps to the :15 minute slot.

Ex. Booking at 4:05 and adding a 5 minute buffer either before or after an event removes the 4:00 and 4:10 slots.
image

I think this is a blocking comment but willing to defer to @emrysal's judgement.

CarinaWolli:
> When I added a 5 minute buffer before or after, the next slot jumps to the :15 minute slot.
> 
> Ex. Booking at 4:05 and adding a 5 minute buffer either before or after an event removes the 4:00 and 4:10 slots. image
> 
> I think this is a blocking comment but willing to defer to @emrysal's judgement.

That's intended. 

3:55 ->  Possible Slot
4:00-4:05 ->  Before buffer for booking
4:05-4:10 ->  **Booking**
4:10-4:15 ->  Before buffer for next possible slot 
4:15 ->  Next possible slot

joeauyeung (approved):
@CarinaWolli thanks for your clarifying comment. This PR is working as expected.

PR #14109, opened 2024-03-15
fix: rescheduling in round robin

### LINKED ISSUE #14108: Incorrect location in round robin rescheduling
Reported on slack

Screenshot 2024-03-16 at 1 28 38 AM

https://calcominc.slack.com/archives/C04CWEU35NV/p1710501163449549

How to reproduce?
1) Create a round robin event type and assign multiple people
2) Set daily video as location
3) Book a meeting and then reschedule the booking. (Check the location url starts meetco instead of https://app.cal.com/video/[uid])

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/14108

## Requirement/Documentation

## Type of change


packages/core/EventManager.ts
```diff
@@ -388,8 +389,7 @@ export default class EventManager {
 
         log.debug("RescheduleOrganizerChanged: Creating Event and Meeting for for new booking");
 
-        const newEvent = { ...evt, destinationCalendar: newDestinationCalendar };
-        const createdEvent = await this.create(newEvent);
```
Udit-takkar, line 392: why do we need to pass newDestinationCalendar to reschedule function?

packages/core/EventManager.ts
```diff
@@ -388,8 +389,7 @@ export default class EventManager {
...
+        const createdEvent = await this.create(originalEvt);
```
Udit-takkar, line 392: `create` function updates the videoCallData in evt(Calendar event object) like 
evt.videoCallData = { ... }

but in reschedule function we passed copy of event object to create in this line which updated the copy of event but not the original event

packages/core/EventManager.ts
CarinaWolli: Isn't this needed for when a reschedule is assigned to a different rr host?

Udit-takkar: then we can update the evt in handleNewBooking instead of doing creating a new evt here

CarinaWolli: I think in line https://github.com/calcom/cal.com/pull/14109/files/01d2f68cac33a3a94395ec5ae7ab826db7d44aa0#diff-f028774ebd744c3141fb60f0fb171c59a2ae4773b285b70f834114e7a3d7afa2L387 we need the destinationCalendar of the old host and in this line (392) we need the destinationCalendar of the new host, no?

CarinaWolli (approved):
fix works as expected 👏🏻

PR #14110, opened 2024-03-15
fix: dynamic event length change on reschedule

### LINKED ISSUE #14083: Dynamic Event length changes to 30min upon reschedule even if original is not 30min

So when I send a [cal.com](http://cal.com/) dynamic link for 60 minutes they are able to schedule a call for 60 minutes. However, when they want to reschedule, if they click reschedule link and go into reschedule process it will bring it back to 30 minutes.
I know they can choose 60 minutes from the menu, but I think they are not expecting that when they reschedule 60 minutes session, it will drop down to 30 minutes

### Steps to Reproduce

1. Schedule an event with dynamic link, 60 mins
2. Reschedule the event
3. Notice the event length changes back to 30 minutes

### Actual Results

- Event length changes to 30 mins upon rescheduling (can be changed to 60 but requires manual intervention)

### Expected Results

- Event length during reschedule should stay the same as the original event

Shobana7:
## What does this PR do?

 This PR fixes the dynamic-event length change upon rescheduling a booking. Following the reschedule link from email or simple URL reload fetched the default eventType length.

Fixes #14083

Video

## Type of change

## Steps to Reproduce

- Create a dynamic event type
- Schedule a booking with duration different from default event duration
- Click on reschedule from the booking confirmation page
- Refresh the URL by providing only the rescheduleUid and removing other query params

## Mandatory Tasks

### DIFF
--- a/packages/features/bookings/Booker/store.ts
+++ b/packages/features/bookings/Booker/store.ts
@@ -282,7 +282,15 @@ export const useBookerStore = create<BookerStore>((set, get) => ({
     // if the user reschedules a booking right after the confirmation page.
     // In that case the time would still be store in the store, this way we
     // force clear this.
-    if (rescheduleUid && bookingData) set({ selectedTimeslot: null });
+    if (rescheduleUid && bookingData) {
+      set({ selectedTimeslot: null });
+      const originalBookingLength = dayjs(bookingData?.endTime).diff(
+        dayjs(bookingData?.startTime),
+        "minutes"
+      );
+      set({ selectedDuration: originalBookingLength });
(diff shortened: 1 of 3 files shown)


Ryukemeister:
@Shobana7 looks like you've made changes to the `yarn.lock` file. Ideally we only change that file while adding or removing any new dependencies, can you revert those changes by referring to [this](https://github.com/calcom/cal.com/blob/main/CONTRIBUTING.md#guidelines-for-committing-yarn-lockfile)

Shobana7:
Reverted changes to yarn.lock

Udit-takkar (approved):
LGTM

PR #14119, opened 2024-03-16
chore: hide org banner i hideEventTypeDetails is true

### LINKED ISSUE #14118: hide org banner when hideEventTypeDetails is true
CleanShot 2024-03-16 at 20 40 25@2x

PeerRich:


Udit-takkar (dismissed):
LGTM

Udit-takkar (approved):

supalarry:
I have a customer asking to hide sidebar
image

and I find logic of this PR confusing - the variable is named `hideEventTypeDetails` but it hides only org banner. `hideEventTypeDetails` should hide all event type details (from the sound of it the whole sidebar) and if it's org banner specific we should have a specific prop for this.

supalarry:
Also, logic says `!hideEventTypeDetails` and PR description is hide org banner if `hideEventTypeDetails` is true, but we are hiding it right now if `hideEventTypeDetails` is false.

PR #14127, opened 2024-03-18
fix: timezone display on booking page to reflect event availability timezone

### LINKED ISSUE #13521: [CAL-3129] Lock timezone on booking page based on availability timezone

The toggle to 'Lock timezone on booking page' for in-person events on the advanced events setup page is a very useful feature. But it doesn't appear to be working as expected for our needs. This is because the timezone that the booking page is locked to isn't the timezone we have set in the 'availability' settings for that event, but rather it is based on the timezone of the user accessing the booking page. This could cause serious misunderstandings, for example when the event is a trade show in third country, which is neither in the Organizer's or Attendee's current timezone.     

We would like to be able to force the display of the timezone on the booking page for in-person events to a specific timezone. Ideally this would be the same timezone for which the 'availability' settings for that event have been defined, and not based on the current timezone of the user accessing the booking page.

The only solution for us at the moment seems to be to remind whoever we share the booking link with, to use the dropdown to set the appropriate timezone manually to the timezone of the trade show on the booking page. It might be possible to add a timezone parameter to the url string for an event booking page, but I haven't be successful when trying to do this. 

<sub>[CAL-3129](https://linear.app/calcom/issue/CAL-3129/lock-timezone-on-booking-page-based-on-availability-timezone)</sub>

Shaik-Sirajuddin:
## What does this PR do?

Currently when the "Lock timezone on booking page" feature is enabled for the event ,
the timezone on the booking page locked is based on the timezone of the user accessing the booking page

this pr updates the display of the timezone on the booking page to the same timezone for which the 'availability' settings for that event have been defined.

Fixes #13521 

[Fix Video](https://www.loom.com/share/5aaaf522deaa412a8a662facee8044ba?sid=332878f5-7655-479d-868e-afdbe507ef29)

## Type of change

## Mandatory Tasks


packages/features/bookings/Booker/components/EventMeta.tsx
```diff
@@ -44,6 +46,20 @@ export const EventMeta = ({
...
+
+  useEffect(() => {
+    if (event && event?.lockTimeZoneToggleOnBookingPage) {
+      if (event?.schedule?.timeZone) {
+        setTimezone(event.schedule?.timeZone);
+      } else if (eventOwnerDefaultShedule) {
+        setTimezone(eventOwnerDefaultShedule.timeZone);
+      }
+    }
+  }, [event, setTimezone, eventOwnerDefaultShedule]);
```
joeauyeung, line 61: **Blocking**: Let's move the logic of getting the owner's timezone to server side

packages/features/bookings/Booker/components/EventMeta.tsx
```diff
@@ -44,6 +46,20 @@ export const EventMeta = ({
   const isEmbed = useIsEmbed();
   const hideEventTypeDetails = isEmbed ? embedUiConfig.hideEventTypeDetails : false;
 
+  const { data: eventOwnerDefaultShedule } = trpc.viewer.availability.schedule.get.useQuery({
```
joeauyeung, line 49: Let's fix this typo here as well

```suggestion
  const { data: eventOwnerDefaultSchedule } = trpc.viewer.availability.schedule.get.useQuery({
```

joeauyeung (changes requested):
Thanks for your contribution @Shaik-Sirajuddin.

I only have one blocking comment to address.

Shaik-Sirajuddin:
Migrated fetching owners default schedule to server side :+1: 
PTAL @joeauyeung

joeauyeung (approved):
Working as expected. Thanks again @Shaik-Sirajuddin!

PR #14135, opened 2024-03-18
fix: booking limits for team event types

CarinaWolli:
## What does this PR do?

This fixes a bug for team event types with booking limits enabled. Slots were shown as available even though the limit was already reached. When trying to book the slot it was throwing "Booking Limit for this event type has been reached"

## Type of change

## How should this be tested?

- create round robin event type and assign hosts 
- add a limit of 1 booking per day 
- book one slot 
- go back to availabilities and see that the day is now blocked 
  - before the day was still open, but when trying to book a slot it was throwing the error

### DIFF
--- a/packages/trpc/server/routers/viewer/slots/util.ts
+++ b/packages/trpc/server/routers/viewer/slots/util.ts
@@ -472,9 +472,7 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
               const { attendees: _attendees, ...bookingWithoutAttendees } = bookings;
               return bookingWithoutAttendees;
             }),
-          busyTimesFromLimitsBookings: busyTimesFromLimitsBookingsAllUsers.filter(
-            (b) => b.userId === currentUser.id
-          ),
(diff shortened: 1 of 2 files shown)


packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -472,9 +472,7 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions) {
               const { attendees: _attendees, ...bookingWithoutAttendees } = bookings;
               return bookingWithoutAttendees;
             }),
-          busyTimesFromLimitsBookings: busyTimesFromLimitsBookingsAllUsers.filter(
-            (b) => b.userId === currentUser.id
-          ),
+          busyTimesFromLimitsBookings: busyTimesFromLimitsBookingsAllUsers,
```
CarinaWolli, line 475: The booking limit is event type wide and it doesn't matter which user was booked.

Udit-takkar (approved):
LGTM. Tested and worked fine

PR #14146, opened 2024-03-19
chore: adds Intercom admin link

zomars:
## What does this PR do?

Adds admin link for Intercom so we can have a shortcut to the user admin view

## Type of change

## How should this be tested?

## Mandatory Tasks


sean-brydon (approved):
LGTM - impersonation from here would be sick too

PR #14159, opened 2024-03-20
fix: bug in zapier when user has no bookings

CarinaWolli:
## What does this PR do?

Before: 
Screenshot 2024-03-20 at 2 44 59 PM

After:
Screenshot 2024-03-20 at 2 44 46 PM


Udit-takkar (approved):
LGTM

PR #14164, opened 2024-03-21
feat: platform endpoints to fetch/cancel/reschedule bookings and hooks

Ryukemeister:
## What does this PR do?

- Adds endpoint to fetch all bookings of a user and bookings based on uid
- Adds hooks for handling fetching of user bookings and bookings based on uid
- Adds endpoint to reschedule and cancel a booking
- Adds hooks for handling booking cancellation and rescheduling
- Adds e2e tests for booking controllers


apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -34,7 +44,50 @@ import { ApiResponse } from "@calcom/platform-types";
...
+    return {
+      status: SUCCESS_STATUS,
+      data: bookings,
+    };
+  }
+
+  // note(Rajiv): currently this endpoint is atoms only
+  @Get("/:bookingUid")
+  async getBooking(@Param("bookingUid") bookingUid: string): Promise<ApiResponse<unknown>> {
+    const { bookingInfo } = await getBookingInfo(bookingUid);
```
supalarry, line 80: I see that "getBookingInfo" consists of "prisma.booking.findFirst" call from "apps/web/modules/bookings/views/bookings-single-view.getServerSideProps.tsx" - don't we also need to perform the rest of the operations e.g. eventType that is returned to frontend is transformed into final form in the get server side props:
```
const eventType = {
    ...eventTypeRaw,
    periodStartDate: eventTypeRaw.periodStartDate?.toString() ?? null,
    periodEndDate: eventTypeRaw.periodEndDate?.toString() ?? null,
    metadata: EventTypeMetaDataSchema.parse(eventTypeRaw.metadata),
    recurringEvent: parseRecurringEvent(eventTypeRaw.recurringEvent),
    customInputs: customInputSchema.array().parse(eventTypeRaw.customInputs),
  };
```

apps/web/modules/bookings/lib/getBookingInfo.ts
```diff
@@ -0,0 +1,67 @@
+import { getBookingWithResponses } from "@calcom/features/bookings/lib/get-booking";
+import prisma from "@calcom/prisma";
+
+const getBookingInfo = async (uid: string) => {
```
supalarry, line 4: Instead of copying and pasting the prisma booking call, we should have the prisma.booking.findFirst as a function that is re-used here and where server side props are generated for web app in "apps/web/modules/bookings/views/bookings-single-view.getServerSideProps.tsx". If we don't we risk having disrepancies.

Ryukemeister: @supalarry created a separate function that handles the prisma call so should be good now :)

packages/trpc/server/routers/viewer/bookings/getAllUserBookings.ts
```diff
@@ -0,0 +1,105 @@
...
+  skip: number;
+  filters: {
+    status: InputByStatus;
+    teamIds?: number[] | undefined;
+    userIds?: number[] | undefined;
+    eventTypeIds?: number[] | undefined;
+  };
+};
+
+const getAllUserBookings = async ({ ctx, filters, bookingListingByStatus, take, skip }: GetOptions) => {
```
ThyMinimalDev, line 24: shouldn't this be called in the get.handler.ts somehwere ?

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
ThyMinimalDev: I think this is done because next.js getServerSideProps can't serialize dates and other data structures

ThyMinimalDev (approved):

PR #14167, opened 2024-03-21
refactor: availabilitySettings atoms web wrapper and translations

ThyMinimalDev:
## What does this PR do?

Refactor atom into web wrapper, and remove translations now that they come from useLocale


supalarry (approved):
Tested web app and atoms example app and availability works!

PR #12658, opened 2023-12-03
feat: add opt-in ready-to-deploy message queue (QStash+Next.js functions)

pablodecm:
## What does this PR do?

This PR introduces **a robust message queuing system that is tightly integrated in the codebase and very extensible**. It is a fully opt-in feature for production-like cal.com deployments that want to enhance the scalability, reliability, and performance of their application. It has already been integrated for all emails and webhooks.

image

## Benefits:

- Extensibility: Fully opt-in, designed to cater to diverse deployment scenarios.
- Efficiency: Simplifies email dispatch processes with unified parameter treatment, less code bloat.
- Minimal code and infrastructure overhead (less than 500 lines with emails and webhook integrations) and uses mostly existing infrastructure
- Enhanced Security: Robust token-based message sending and signature checks for task runners ensure high security.

## Short Technical Video and Demo

[Short Technical Video and Demo](https://youtu.be/6NhpcW8Rqok?si=0hC_2kLH_IN3EJeQ)

## More info

[Longer Video Description](https://youtu.be/0C8qExHmmg8?si=pMI9mAl-5U9pUCBT)

This PR also showcases the integration for:
- [DONE] All emails send in the application by means of substituting the sendEmail calls with a dispatcher (we unified the parameters in a single object to simplify the treatment)
- [DONE] All Webhook handling
- [FUTURE] Integration with 3rd Party Calendar Event APIs could also be done

Even if done as part of OSShack to address https://osshack.com/bounty/25 (aka https://github.com/calcom/cal.com/issues/12556 or  [CAL-2764](https://linear.app/calcom/issue/CAL-2764/implement-message-queuing-system)), the architecture and  technological choices were purposely done to minimise the amount of changes required and ensure minimal bloating and required changes. This changes will also improve the reliability and the performance of the cal.com app in general.

We only add a single lightweight external dependency ([Upstash QStash](https://upstash.com/docs/qstash/overall/getstarted)) as a message queue from a provider (Upstash) that is already used for another another feature in the code base (rate limiting). Furthermore, any similar queue system that has the capability to call URLs could also be integrated quite easily as an alternative.

As we use application functions as task runners, the infrastructure overhead is minimal and everything is integrated in the same code base. Because in production settings Next.js functions are serveless the system has great scalability. Similarly, existing tools for logging and performance monitoring (i.e. Sentry, Vercel logging and others) will work readily. In addition, there is an additional logging service for the message queue that is easily accesible from the QStash dashboard.

[...]

PR #12655, opened 2023-12-03
feat:  Implement A Message Queuing System

pmbanugo:
## What does this PR do?

Fixes #12556

This is an implementation of a message/task queue system implemented using [Temporal](temporal.io) (which is also an OSS project). I have added a new project (`@calcom/queue-worker`), Workflows for sending emails, and made scheduling an event to trigger the Workflow as a proof of concept.

There are some places where the value/data used could be configured to come from an environment variable but I left them hardcoded because I was running out of time to finish this hack on time 😅.

> For the email Workflows, the payload would need to be adjusted so that they're serialisable over the network. Currently, it contains a property of type `TFunction` which I think is used for translation. We would need a way to handle doing translations within the worker processes.

Loom Video: https://www.loom.com/share/b3f7d677b3cc4ec9a7f04f5eea255041?sid=39089f28-2838-48d0-a6a9-7efdcee7f1f0

## Loom Video

https://www.loom.com/share/b3f7d677b3cc4ec9a7f04f5eea255041?sid=39089f28-2838-48d0-a6a9-7efdcee7f1f0

## Requirement/Documentation

- There's an [architecture/proposal document](https://github.com/calcom/cal.com/blob/4641901720b45d6eca081e557b2d6d209bf5dfd0/apps/queue-worker/Architecture.md) which describes the solution on a high level. It also covers some info regarding the technical requirements.
- There's a [README](https://github.com/calcom/cal.com/blob/4641901720b45d6eca081e557b2d6d209bf5dfd0/apps/queue-worker/README.md) for how to run the new `queue-worker` project.

## Type of change

- [x] New feature (non-breaking change which adds functionality). Existing functionality would work as expected, only that the payload must be serialisable.

## How should this be tested?

1. Install the Temporal CLI using the command `brew install temporal`.
2. Start the Temporal dev server using `temporal server start-dev`.
3. Check out the branch and run `yarn install`.
4. Add a `.env` file to the **apps/queue-worker** project. You can copy the content of `.env.example` and make sure the `DATABASE_URL` matches the one also in the root `.env` file.
5. While in the queue-worker directory, you can start the application by running `yarn run start`. This will start the Worker process and will poll the Temporal server for new tasks.
6. Run the Cal.com web project. Go to the root directory and run `yarn run dev` to start the web application. Note: ensure you have the environment variables properly configured.
7. Open http://localhost:3000/event-types in your browser follow one of the links to schedule a meeting. This should trigger the `sendScheduledEmail` Workflow and process the event.

[...]

PR #12641, opened 2023-12-02
feat: Message Queuing System using Trigger.dev

wladpaiva:
## What does this PR do?

Add Trigger.dev as a queue system.

### Why trigger.dev?

First of all, they are Open Source!! 🎉🎉🎉

Second, they are a great fit for our needs. We need a queue system that can handle a lot of jobs and that can scale easily. We also need a queue system that can handle webhooks and that can be integrated with other services. Trigger.dev can do all of that and more.

Video:
[Trigger.dev](https://www.youtube.com/watch?v=cT1pfsaZtdA)

### What else needs to be done:
- [ ] Configure adicional integrations like Twilio, calendars and so on.
- [ ] Design triggers accordantly to each job to allow individual logging / debugging / retry / testing. For example, emails are currently being build when the `sendEmail` is called so pretty much all emails would be going through the same job "send email" and making it hard to identify stuff.
- [ ] Change current flows to appropriately add itens to queue (trigger) when Trigger.dev configurations are set.

Fixes # (issue)
<https://github.com/calcom/cal.com/issues/12556>

## Requirement/Documentation

<https://github.com/calcom/cal.com/issues/12556>

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How should this be tested?

Start the local version of Trigger.dev or [create an account](https://trigger.dev) on the cloud version.:

```bash
cd /infra/docker/queue
docker-compose up -d
```

The following environment variables are required to run this can be found in the `Environments & API Keys` section of the dashboard:

- `TRIGGER_API_KEY`: key of the environment/project in Trigger.dev.
- `NEXT_PUBLIC_TRIGGER_PUBLIC_API_KEY`: Public key of the environment/project in Trigger.dev (used in the frontend).
- `TRIGGER_API_URL`: URL of the Trigger.dev API. Defaults to `https://localhost:3030` if you are using the local version or you can use `https://cloud.trigger.dev`.
[...]

PR #14180, opened 2024-03-22
feat: Tasker

### LINKED ISSUE #12556: [CAL-2764]  Implement Message Queuing System
Our application currently lacks a robust message queuing system, leading to potential bottlenecks and inefficiencies, especially when handling tasks such as triggering emails, SMS messages and webhooks, and interfacing with 3rd party calendar event APIs. To enhance the scalability, reliability, and performance of our application, we need to integrate a message queuing system.

### Desired Features:

* **Task Queuing**: Implement a message queuing system to handle various tasks asynchronously, ensuring that critical processes don't block the main application flow.
* **Email Triggering**: Enable queuing for email sending tasks to improve responsiveness and prevent delays in user interactions.
* **Webhook Handling**: Queue tasks related to webhook processing, ensuring that they are processed efficiently and reliably.
* **Integration with 3rd Party Calendar Event APIs**: Facilitate the queuing of tasks involving interactions with external calendar event APIs, ensuring timely and reliable communication. e.g. If adding an event to a Google Calendar fails, it should be able to be retried to ensure the event is added successfully after a booking is created.

### Requirements:

* **Scalability**: The message queuing system should be scalable to handle increased loads seamlessly.
* **Reliability**: Ensure that queued tasks are reliably processed, with mechanisms for retrying failed tasks.
* **Monitoring and Logging**: Implement logging and monitoring features to facilitate easy debugging and performance analysis. Axiom and Sentry are two services we currently have support for.
* **Documentation**: Provide comprehensive documentation for integrating and maintaining the message queuing system within our application.

### Additional Considerations:

* **Technology Stack**: Propose suitable technologies for implementing the message queuing system in the context of our app.
* **Testing**: Develop a robust testing strategy to validate the functionality and performance of the message queuing system.
* **Security**: Implement security best practices to safeguard queued tasks and prevent unauthorized access or manipulation.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2764](https://linear.app/calcom/issue/CAL-2764/implement-message-queuing-system)</sub>

zomars:
## What does this PR do?

closes #12556

# Tasker

Tasker: "One who performs a task, as a day-laborer."

Task: "A function to be performed; an objective."

## What is it?

Introduces a new pattern called Tasker which may be switched out in the future for other third party services.

Also introduces a base `InternalTasker` which doesn't require third party dependencies and should work out of the box (by configuring a proper cron).

## Why is this needed?

The Tasker pattern is needed to streamline the execution of non-critical tasks in an application, providing a structured approach to task scheduling, execution, retrying, and cancellation. Here's why it's necessary:

1. **Offloading non-critical tasks**: There are tasks that don't need to be executed immediately on the main thread, such as sending emails, generating reports, or performing periodic maintenance tasks. Offloading these tasks to a separate queue or thread improves the responsiveness and efficiency of the main application.

2. **Retry mechanism**: Not all tasks succeed on the first attempt due to errors or external dependencies. This pattern incorporates a retry mechanism, which allows failed tasks to be retried automatically for a specified number of attempts. This improves the robustness of the system by handling temporary failures gracefully.

3. **Scheduled task execution**: Some tasks need to be executed at a specific time or after a certain delay. The Tasker pattern facilitates scheduling tasks for future execution, ensuring they are performed at the designated time without manual intervention.

4. **Task cancellation**: Occasionally, it's necessary to cancel a scheduled task due to changing requirements or user actions. The Tasker pattern supports task cancellation, enabling previously scheduled tasks to be revoked or removed from the queue before execution.

5. **Flexible implementation**: The Tasker pattern allows for flexibility in implementation by providing a base structure (`InternalTasker`) that can be extended or replaced with third-party services (`TriggerDevTasker`, `AwsSqsTasker`, etc.). This modularity ensures that the task execution mechanism can be adapted to suit different application requirements or environments.

Overall, the Tasker pattern enhances the reliability, performance, and maintainability by managing non-critical tasks in a systematic and efficient manner. It abstracts away the complexities of task execution, allowing developers to focus on core application logic while ensuring timely and reliable execution of background tasks.

## How does it work?

Since the Tasker is a pattern on itself, it will depend on the actual implementation. For example, a `TriggerDevTasker` will work very differently from an `AwsSqsTasker`.

For simplicity sake will explain how the `InternalTasker` works:

- Instead of running a non-critical task you schedule using the tasker:

  ```diff
  const examplePayload = { example: "payload" };
  - await sendWebhook(examplePayload);
  + await tasker.create("sendWebhook", JSON.stringify(examplePayload));
  ```

- This will create a new task to be run on the next processing of the task queue.
- Then on the next cron run it will be picked up and executed:

  ```ts
  // /app/api/tasks/cron/route.ts
  import tasker from "@calcom/features/tasker";

  export async function GET() {
    // authenticate the call...
    await tasker.processQueue();
    return Response.json({ success: true });
  }
  ```

- By default, the cron will run each minute and will pick the next 100 tasks to be executed.
- If the tasks succeeds, it will be marked as `suceededAt: new Date()`. If if fails, the `attempts` prop will increase by 1 and will be retried on the next cron run.
- If `attempts` reaches `maxAttemps`, it will be considered a failed and won't be retried again.
- By default, tasks will be attempted up to 3 times. This can be overridden when creating a task.
- From here we can either keep a record of executed tasks, or we can setup another cron to cleanup all successful and failed tasks:

  ```ts
  // /app/api/tasks/cleanup/route.ts
  import tasker from "@calcom/features/tasker";

  export async function GET() {
    // authenticate the call...
    await tasker.cleanup();
    return Response.json({ success: true });
  }
  ```

- This will delete all failed and successful tasks.
- A task is just a simple function receives a payload:

  ```ts
  type TaskHandler = (payload: string) => Promise<void>;
  ```

## How to contribute?

You can contribute by either expanding the `InternalTasker` or creating new Taskers. To see how to add new Taskers, see the `tasker-factory.ts` file.

You can also take some inspiration by looking into previous attempts to add various Message Queue pull requests:

- [feat: Messaging Bus Implementation using AWS SQS OSSHack Challenge](https://github.com/calcom/cal.com/pull/12663)
- [feat: add opt-in ready-to-deploy message queue (QStash+Next.js functions)](https://github.com/calcom/cal.com/pull/12658)
- [feat: Implement A Message Queuing System](https://github.com/calcom/cal.com/pull/12655)
- [Message Queuing System](https://github.com/calcom/cal.com/pull/12654)
- [feat: Message Queuing System using Trigger.dev](https://github.com/calcom/cal.com/pull/12641)


packages/features/tasker/repository.ts
zomars, line 1: This should be the only entry point to interact with the Task table in the DB

packages/features/tasker/README.md
zomars, line 1: Added README for future reference.

zomars (commented):
Self review done

packages/features/tasker/repository.ts
```diff
@@ -0,0 +1,148 @@
...
+      orderBy: {
+        scheduledAt: "asc",
+      },
+      take: 100,
+    });
+    return tasks;
+  }
+  static async getAll() {
+    const tasks = await db.task.findMany();
+    return tasks;
```
exception, line 67:
```suggestion
    return db.task.findMany();
```

packages/features/tasker/repository.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+  static async getFailed() {
+    const tasks = await db.task.findMany({
+      where: whereMaxAttempsReached,
+    });
+    return tasks;
```
exception, line 73: Same as above, and for the rest.

packages/features/tasker/repository.ts
```diff
@@ -0,0 +1,148 @@
...
+    const newTask = await db.task.create({
+      data: {
+        payload,
+        type,
+        scheduledAt,
+        maxAttempts,
+      },
+    });
+    return newTask;
+  }
```
exception, line 54: How come no spacing between methods?

packages/features/tasker/tasker.ts
```diff
@@ -0,0 +1,9 @@
+export type TaskerTypes = "internal" | "redis";
+export type TaskTypes = "sendEmail" | "sendWebhook" | "sendSms";
+export type TaskHandler = (payload: string) => Promise<void>;
+export interface Tasker {
+  /** Create a new task with the given type and payload. */
+  create(type: string, payload: string): Promise<void>;
```
exception, line 6: imo create should return the id of the task, could be useful.

exception, line 6: should `type` be of type `TaskTypes`?

packages/features/tasker/internal-tasker.ts
```diff
@@ -0,0 +1,38 @@
...
+  async create(type: TaskTypes, payload: string): Promise<void> {
+    await Task.create(type, payload);
+  }
+  async processQueue(): Promise<void> {
+    const tasks = await Task.getNextBatch();
+    const tasksPromises = tasks.map(async (task) => {
+      const taskHandlerGetter = tasksMap[task.type as keyof typeof tasksMap];
+      if (!taskHandlerGetter) throw new Error(`Task handler not found for type ${task.type}`);
+      const taskHandler = await taskHandlerGetter();
+      return taskHandler(task.payload)
```
exception, line 21: Alright, here's my more in-depth thoughts (and can be done in the future). A Task should be represented by it's own entity, so rather than `Task` being a class of static types, it should represent only itself (with it's id), so all methods would mutate it's own state, eg:

```ts
return taskHandler(payload).then((task) => task.succeed);
```

This way, tasks can be retried, mutated, canceled, etc without static methods.
It's only a `nit` by itself but could lead to a nicer structure.

packages/features/tasker/internal-tasker.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Then, you can use the TaskerFactory to select the new Tasker.
+ */
+export class InternalTasker implements Tasker {
+  async create(type: TaskTypes, payload: string): Promise<void> {
+    await Task.create(type, payload);
+  }
+  async processQueue(): Promise<void> {
+    const tasks = await Task.getNextBatch();
+    const tasksPromises = tasks.map(async (task) => {
+      const taskHandlerGetter = tasksMap[task.type as keyof typeof tasksMap];
```
exception, line 18: If `task.type` is inferred from the TaskTypes, this cast should not be required, no?

packages/features/tasker/redis-tasker.ts
```diff
@@ -0,0 +1,17 @@
+import { type Tasker, type TaskTypes } from "./tasker";
+
+/**
+ * RedisTasker is a tasker that uses Redis as a backend.
+ * WIP: This is a work in progress and is not fully implemented yet.
+ **/
+export class RedisTasker implements Tasker {
+  async create(type: TaskTypes, payload: string): Promise<void> {
+    throw new Error("Method not implemented.");
+  }
```
exception, line 10: Spacing

packages/features/tasker/repository.ts
```diff
@@ -0,0 +1,148 @@
...
+  static async cancel(taskId: number) {
+    const task = await db.task.delete({
+      where: {
+        id: taskId,
+      },
+    });
+    return task;
+  }
+  static async cleanup() {
```
exception, line 135: I think we need to probably follow the repository pattern, if we're already abstracting this to be DB agnostic on the actual handling, could do the same as Orgs v2 does.

exception (dismissed):
First round of reviews done!

keithwillcode:
@zomars curious why you went with triggering these via cron and not in more real-time. Must say I don’t love the idea that the default setup creates a default max lag of 1 minute to see tasks processed when in reality, our system should be able to handle any task immediately, which is how most message queuing systems work and how downstream systems/users would expect things to work.

packages/prisma/schema.prisma
```diff
@@ -1267,3 +1267,23 @@ model SecondaryEmail {
...
+model Task {
+  id          Int       @id @default(autoincrement())
+  createdAt   DateTime  @default(now())
+  updatedAt   DateTime  @updatedAt
+  // The time at which the task should be executed
+  scheduledAt DateTime  @default(now())
+  // The time at which the task was successfully executed
+  succeededAt DateTime?
+  // The task type to be executed. Left it as a freeform string to avoid more migrations for now. Will be enforced at type level.
+  type        String
```
keithwillcode, line 1281: Will this also handle versioning of the types? 

Many times when building these types of systems, we want to version the tasks so that when they are processed, they are processed by a handler that knows the version. 

Reason is when the system has immense amount of traffic and there can be no down time, migrating to a v2 handler, let’s say, needs to work seamlessly and all v1 tasks need to be processed correctly as well.

zomars: I don't see why not. We can follow up a naming convention like sendEmailV2, sendEmailV3 and so on and keep the handlers around for long as needed.

zomars:
> curious why you went with triggering these via cron and not in more real-time. 

The main principle for the internal tasker for me was "use what we already have" as much as possible. A cron have this limitation of running each minute. Even so, we could in theory setup a simple service who it's sole purpose is to ping this endpoint each X seconds and get quicker batching. But IMO that's out of scope for this PR.

packages/features/tasker/repository.ts
zomars: I didn't want mix these patterns up. As the Task model is intended to be separate from the other models. Hence the lack of relationships.

packages/features/tasker/internal-tasker.ts
zomars: I don't think I'm following your example here. I wanted to keep the class simple by providing a simple interface to call predefined methods. Static seemed like the natural way to go about it. Of course this can evolve if requirements change.

packages/features/tasker/tasks/sendEmail.ts
```diff
@@ -0,0 +1,25 @@
...
+  to: z.string(),
+  /** The email template to send */
+  template: z.string(),
+  payload: z.string(),
+});
+
+export async function sendEmail(payload: string): Promise<void> {
+  try {
+    const parsedPayload = sendEmailPayloadSchema.parse(JSON.parse(payload));
+    console.log(parsedPayload);
```
joeauyeung, line 14: Do we want to add this to our logs?

packages/features/tasker/internal-tasker.ts
```diff
@@ -0,0 +1,38 @@
+import { Task } from "./repository";
+import { type Tasker, type TaskTypes } from "./tasker";
+import tasksMap from "./tasks";
+
+/**
+ * This is the default internal Tasker that uses the Task repository to create tasks.
+ * It doens't have any external dependencies and is suitable for most use cases.
```
joeauyeung, line 7:
```suggestion
 * It doesn't have any external dependencies and is suitable for most use cases.
```

packages/features/webhooks/lib/schedulePayload.ts
```diff
@@ -0,0 +1,16 @@
...
+type SchedulePayload = typeof sendPayload;
+
+const schedulePayload: SchedulePayload = async (secretKey, triggerEvent, createdAt, webhook, data) => {
+  await tasker.create("sendWebhook", JSON.stringify({ secretKey, triggerEvent, createdAt, webhook, data }));
+  return {
+    ok: true,
+    status: 200,
+    message: "Webhook scheduled successfully",
+  };
+};
```
joeauyeung, line 14: What if this fails? Should we fallback to immedietly sending the webhook?

packages/features/tasker/tasks/sendEmail.ts
```diff
@@ -0,0 +1,25 @@
...
+    const email = emails[parsedPayload.template as keyof typeof emails];
+    if (!email) throw new Error("Invalid email template");
+    // @ts-expect-error - TODO bring back email type safety
+    await email(parsedPayload.to);
+  } catch (error) {
+    // ... handle error
+    console.error(error);
+    throw error;
+  }
+}
```
joeauyeung, line 25: I noticed the payload schema only has a single `to` property but you're importing templates from the `email-manager`? Those methods send emails to organizers, team members, and attendees. I would say this isn't atomic. What if sending emails to the organizer works but sending to attendees fail? On retries the organizer would get multiple emails.

joeauyeung (changes requested):
This is going to be sick @Zomars.

I was thinking we have to be careful about how our code is executed. For example, in `handleNewBooking` the calendar event is created but if an error occurs before the email, it's never sent. I would hate for 1/2 of a task failing but that 1/2 gets executed on retries. In the future, we need to make sure our tasks are as atomic as possible.

For example, this comment: https://github.com/calcom/cal.com/pull/14180#discussion_r1541451440

packages/features/tasker/internal-tasker.ts
zomars: task type is of type of string since it comes directly from db. Didn't want to make it a enum to avoid adding more migrations when adding more tasks.

packages/features/tasker/tasks/sendEmail.ts
zomars: RN there's no created sendEmail tasks yet. We're starting with only webhooks for now.

packages/features/tasker/tasks/sendEmail.ts
zomars: That's way upper in the logic. task handler should not concern about these things.

packages/features/webhooks/lib/schedulePayload.ts
zomars: TBD

joeauyeung:
> @zomars great work!
> 
> curious why you went with triggering these via cron and not in more real-time. Not sure I love the idea that the default setup creates a max lag of 1 minute to see tasks processed when in reality, our system should be able to handle any task immediately, which is how most message queuing systems work and how downstream systems/users would expect things to work.

@keithwillcode I talked with @zomars about what his immediate goals were for this PR. He mentioned that scheduling tasks for later is the benefit and I pictured the Tasker to minimize the time between when a booker presses the book button and when the success screen appears.

For V1 of the Tasker with the 1 min delay, we agreed that webhooks and workflows would be good candidates since those can be scheduled for later and only affect the organizer. From the booker POV they would expect emails immediately so it shouldn't be included in V1. 

The `EmailManager` relies on data from the `EventManager` so these two should be out of scope for V1.

zomars:
Yeah. The focus for v1 is the ability to schedule and delegate. We can implement more realtime stuff as we're needing it.

keithwillcode:
Great. Sounds like a plan @zomars @joeauyeung.

joeauyeung (changes requested):
Only thing blocking an approval for me is getting this comment resolved.
@zomars @exception 
https://github.com/calcom/cal.com/pull/14180/files#r1540233954

zomars:
> Only thing blocking an approval for me is getting this comment resolved. @zomars @exception [#14180 (files)](https://github.com/calcom/cal.com/pull/14180/files#r1540233954)

Out of scope for this PR IMO.

joeauyeung (approved):
Let's do it! (At least for webhooks for now)

emrysal (approved):
Tagging @CarinaWolli for visibility, looks good to me.

PR #14181, opened 2024-03-22
fix: Adding missing multiple_duration_timeUnit on both zh-tw and zh-cn locales.

anscg:
## What does this PR do?

Fixes `multiple_duration_timeUnit` missing on both `zh-tw` and `zh-cn` locales.

Fixes # (issue)

image

## Type of change

- Bug fix

## How should this be tested?

Check if multiple_duration_timeUnit displayed correctly on both locales.

## Mandatory Tasks


Udit-takkar (approved):
@antcar0929 Thanks. I have pushed some changes. could you please sign the CLA https://github.com/calcom/cal.com/pull/14181#issuecomment-2015132278 ?

dryden:
As shown in the screenshot, I noticed that although a fix was applied in March, the issue still persists with both Simplified and Traditional Chinese. Since users cannot choose their display language, they have no way of seeing the correct time durations, which prevents them from effectively using this valuable tool.
SCR-20240805-bfri

PR #14187, opened 2024-03-22
fix: Reinstall App Link Confusing

### LINKED ISSUE #14184: [CAL-3301] Reinstall App Link Confusing
When you press "reinstall app" you get taken to the App's page, this button should maybe uninstall the app for you so you can reinstall it. 

Now, you have to know to navigate to settings, conferencing, then delete the app then head back and reinstall it

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3301](https://linear.app/calcom/issue/CAL-3301/reinstall-app-link-confusing)</sub>

Amit91848:
## What does this PR do?

Fixes #14184 
Fixes CAL-3301

https://github.com/calcom/cal.com/assets/74371312/7d61b5c6-44bb-4416-91c8-6ab665996d21

## Type of change

## Mandatory Tasks


packages/features/users/components/InvalidAppCredentialsBanner.tsx
```diff
@@ -16,22 +16,37 @@ export function InvalidAppCredentialBanners({ data }: InvalidAppCredentialBanner
...
+      showToast(t("app_removed_successfully"), "success");
+    },
+    onError: () => {
+      showToast(t("error_removing_app"), "error");
+    },
+    async onSettled() {
+      await utils.viewer.connectedCalendars.invalidate();
+      await utils.viewer.integrations.invalidate();
+    },
+  });
```
Udit-takkar, line 46: I am not sure if we should delete credentials when user clicks on Reinstall app because it could be possible that google calendar integration has been expired and user has multiple google calendar accounts connected.

Udit-takkar: and not user doesn't remember which account was expired.

Udit-takkar: Or it could be possible that team's stripe app is invalid but not the personal stripe app.

Udit-takkar (dismissed):
got an error removing invalid stripe app (installed for team)

Screenshot 2024-03-22 at 6 57 26 PM

packages/features/users/components/InvalidAppCredentialsBanner.tsx
Amit91848: > I am not sure if we should delete credentials when user clicks on Reinstall app because it could be possible that google calendar integration has been expired and user has multiple google calendar accounts connected.

> and not user doesn't remember which account was expired.

Hmm this is tricky. 
We can show the email in the banner eg. "... revoked for Outlook Calendar(some@example.com)", and also show the email in toast message?

Any other suggestions?

Amit91848:
> got an error removing invalid stripe app (installed for team)

Works now.

https://github.com/calcom/cal.com/assets/74371312/e5ae5d07-6209-4142-a495-3a1822b60efb

joeauyeung (changes requested):
I'm going to have to agree with @Udit-takkar. We shouldn't be deleting the credential when the user clicks the banner. If a user decides to abandon the reinstall process half way through, then they're left without that credential
 
Since this mostly happens with OAuth apps, I would like to see the user being redirected to the OAuth install flow and then updating the credentials on the callback.

PR #14188, opened 2024-03-22
fix: improve api/cron/webhookTriggers endpoint

CarinaWolli:
## What does this PR do?

The endpoint was temporarily failing because it timed out. This PR introduces some performance improvements

### DIFF
--- a/packages/features/webhooks/lib/cron.ts
+++ b/packages/features/webhooks/lib/cron.ts
@@ -31,20 +31,35 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
         lte: dayjs().toISOString(),
       },
     },
+    select: {
+      id: true,
+      jobName: true,
+      payload: true,
+      subscriberUrl: true,
+      webhook: {
+        select: {
+          secret: true,
+        },
+      },
+    },
   });
+
+  const fetchPromises: Promise<any>[] = [];
 
   // run jobs
   for (const job of jobsToRun) {
     // Fetch the webhook configuration so that we can get the secret.
     const [appId, subscriberId] = job.jobName.split("_");
+    let webhook = job.webhook;
 
-    let webhook;
-    try {
-      webhook = await prisma.webhook.findUniqueOrThrow({
-        where: { id: subscriberId, appId: appId !== "null" ? appId : null },
-      });
-    } catch {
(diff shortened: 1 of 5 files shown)


packages/features/webhooks/lib/cron.ts
```diff
@@ -31,20 +31,35 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
-    let webhook;
-    try {
-      webhook = await prisma.webhook.findUniqueOrThrow({
-        where: { id: subscriberId, appId: appId !== "null" ? appId : null },
-      });
-    } catch {
-      logger.error(`Error finding webhook for subscriberId: ${subscriberId}, appId: ${appId}`);
+    if (!webhook) {
+      try {
+        webhook = await prisma.webhook.findUniqueOrThrow({
```
CarinaWolli, line 57: We now save the webhook to `webhookScheduledTriggers` so we don't have to do that anymore for each job

packages/features/webhooks/lib/cron.ts
```diff
@@ -56,33 +71,15 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+      fetch(job.subscriberUrl, {
         method: "POST",
         body: job.payload,
         headers,
-      });
-    } catch (error) {
-      console.log(`Error running webhook trigger (retry count: ${job.retryCount}): ${error}`);
-
-      // if job fails, retry again for 5 times.
-      if (job.retryCount < 5) {
```
CarinaWolli, line 69: retry was not implemented correctly and it returned `ok: false` on the first error. This also caused that all the other jobs were not triggered

I don't think we need to retry if the request fails, now we are just login the error and continue with the other jobs

packages/features/webhooks/lib/cron.ts
```diff
@@ -122,6 +119,8 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     });
   }
 
+  Promise.allSettled(fetchPromises);
```
CarinaWolli, line 122: process fetch operations asynchronously

joeauyeung (approved):
PR LGTM.

I wonder if 5 retries was too generous and that was causing the timeouts.

zomars (approved):
LGTM!

PR #14200, opened 2024-03-25
feat: prevent OAuth client creation for non platform orgs

supalarry:
## What does this PR do?

We need to distinguish between platform and non-platform orgs, because only platform orgs should be allowed to have OAuth clients and also they do not need domains created when the platform org is created. To accomplish that:

1. Add "isPlatform" flag to the "Team" model with default value of false in [feat: isPlatform flag for teams](https://github.com/calcom/cal.com/commit/e52cfd774d6d71023c3fa7c8eb955bb36505535a)
2. To distinguish between platform org an non-platform org, in [feat: create organizations service to check if org is platform](https://github.com/calcom/cal.com/commit/4ac7f148ab88595432fd86dc5c80fc8596ac00a4) create organizations service that based on organizationId informs if organization is platform or not.
3. OAuth clients are created in the "oauth-clients.controller.ts" which uses the "OrganizationRolesGuard". - extended the guard to throw Forbidden exception if the organization is not platform in [feat: check if organization is platform in the OrganizationRolesGuard](https://github.com/calcom/cal.com/commit/0357f9435f8ac01584815daf7c254c5d52d929fb) by using service from step 2.
4. To prevent domain being created for platform orgs we must modify "organizations/create.schema.ts" and  "organizations/create.handler.ts". First, i refactored handler in [refactor: reduce nesting in create handler](https://github.com/calcom/cal.com/commit/60206a79f96f96db5890376293b22a8ed60653c0) to reduce nesting and abstract check in a function. Then, I updated schema with "isPlatform" flag that is checked by the handler to only create domain if "isPlatform" is false in [feat: dont create domain for platform orgs](https://github.com/calcom/cal.com/commit/f226a3f10beee8b1ec9a8353a5f5dfc19e7f7435).
5. Added test in [oauth-clients.controller.e2e-spec.ts](https://github.com/calcom/cal.com/compare/lauris/pla-67-isplatform-field-in-database-for-orgs?expand=1#diff-bc94ed181a1f1945fcc93c16eec9f602e8dd942b25d124f0a3a0dcb219ab85ed) that checks if Forbidden is returned if the org is not platform to make sure that the updated guard in step 3 works as expected.

## Creating platform orgs
Given that orgs are created via the updated "organizations/create.handler.ts", to create platform org need to pass "isPlatform: true" to the payload. As an example, can check “CreateANewOrganizationForm.tsx” component that invokes the handler to create a non platform org.

### DIFF
--- a/packages/trpc/server/routers/viewer/organizations/create.handler.ts
+++ b/packages/trpc/server/routers/viewer/organizations/create.handler.ts
@@ -71,119 +71,119 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
   const availability = getAvailabilityFromSchedule(DEFAULT_SCHEDULE);
   let isOrganizationConfigured = false;
 
-  if (check === false) {
-    isOrganizationConfigured = await createDomain(slug);
+  if (check) {
+    await checkOrg(adminEmail, input.language);
+    return { checked: true };
+  }
 
-    if (!isOrganizationConfigured) {
-      // Otherwise, we proceed to send an administrative email to admins regarding
-      // the need to configure DNS registry to support the newly created org
-      const instanceAdmins = await prisma.user.findMany({
-        where: { role: UserPermissionRole.ADMIN },
-        select: { email: true },
+  isOrganizationConfigured = await createDomain(slug);
+
+  if (!isOrganizationConfigured) {
+    // Otherwise, we proceed to send an administrative email to admins regarding
+    // the need to configure DNS registry to support the newly created org
+    const instanceAdmins = await prisma.user.findMany({
+      where: { role: UserPermissionRole.ADMIN },
+      select: { email: true },
+    });
+    if (instanceAdmins.length) {
+      await sendAdminOrganizationNotification({
+        instanceAdmins,
+        orgSlug: slug,
+        ownerEmail: adminEmail,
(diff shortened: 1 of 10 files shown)


packages/trpc/server/routers/viewer/organizations/create.handler.ts
```diff
@@ -69,121 +69,120 @@ export const createHandler = async ({ input, ctx }: CreateOptions) => {
...
-  let isOrganizationConfigured = false;
 
-  if (check === false) {
-    isOrganizationConfigured = await createDomain(slug);
+  if (check) {
+    await checkOrg(adminEmail, input.language);
+    return { checked: true };
+  }
+
+  const isOrganizationConfigured = isPlatform ? true : await createDomain(slug);
```
ThyMinimalDev, line 78: this should work too 
` const isOrganizationConfigured = isPlatform || (await createDomain(slug));`

ThyMinimalDev (approved):

packages/trpc/server/routers/viewer/organizations/create.handler.ts
supalarry: Even though proposed code is shorter and checks out its for my brain easier to comperhend one with ternary operators so will leave it.

PR #14262, opened 2024-03-29
fix: bad UX at end of month

### LINKED ISSUE #14260: [CAL-3330] bad UX at end of month
image

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3330](https://linear.app/calcom/issue/CAL-3330/bad-ux-at-end-of-month)</sub>

ASQUAREEE:
Fixed #14260

I just fixed the issue. If it's the last day of the month, then simply move to the next month


sean-brydon:
Hi @ASQUAREEE thank for the PR - we're discussing internally if this is the solution we want. It's not ideal to navigate the user and show no error if they manually navigate to a month via the url for example.

@itsberkaya @ciaranha i think we should maybe mockup the ideal flow for this

CarinaWolli (changes requested):
If it's the last day of the month and there are available slots on that day, we should still show the current month

We probably want to fix that differently, let's discuss that in the issue: https://github.com/calcom/cal.com/issues/14260#issuecomment-2046033674

CarinaWolli:
Closing this PR because we want to solve that differently: https://github.com/calcom/cal.com/issues/14260#issuecomment-2047537743

ASQUAREEE:
> Closing this PR because we want to solve that differently: [#14260 (comment)](https://github.com/calcom/cal.com/issues/14260#issuecomment-2047537743)

Understood, thanks for the update

PR #14263, opened 2024-03-29
refactor: prisma platform tables use pascal case

supalarry:
Platform tables and their properties were mapped to snake_case in database, however, the standard is PascalCase.

This migration removes @@map from tables and @map from properties.


emrysal (approved):
👍

PR #14264, opened 2024-03-29
refactor: OAuth creation form hide apps permissions

supalarry:
We do not support apps in atoms / v2, so we can hide "Apps" permissions when creation OAuth client. I just commented out code so we can uncomment it later on.

Before:
Screenshot 2024-03-29 at 10 36 53

After:
Screenshot 2024-03-29 at 10 40 22


Ryukemeister (approved):
LGTM

Issue #10233, opened 2023-07-18
[CAL-2200] Close.com app not working as expected


leog:
Testing API key button fails

Meeting not created inside Close.com

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2200](https://linear.app/calcom/issue/CAL-2200/closecom-app-not-working-as-expected)</sub>

PR #14272, opened 2024-03-30
refactor: Introduce `crmManager`

### LINKED ISSUE #14452: [CAL-3344] Introduce `CRMManager` class
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3344](https://linear.app/calcom/issue/CAL-3344/introduce-crmmanager-class)</sub>

joeauyeung:
## What does this PR do?

This PR separates the handling of CRM apps from the calendar manager into it's own `crmManager`. 

- Introduces `CrmManager`
- Moves CRM apps `CalandarService` into `CrmService`
- Refactor logic that was in individual apps into the `CrmManager`

Fixes #14452 

## Requirement/Documentation

## Type of change

## How should this be tested?

- For each CRM app
	- Install the CRM app
	- Create a booking
		-	A new contact should be created under the CRM app 
	- Create a second booking with the same attendee info
	- Update the booking. Should be reflected in the CRM app
	- Delete the booking. Should be reflected in the CRM app

- Should be tested on
	- Close.com
	- Hubspot
	- Pipedrive
	- Salesforce
	- Zoho CRM
	- Zoho Bigin

## Mandatory Tasks


packages/app-store/closecom/lib/CrmService.ts
```diff
@@ -110,27 +105,76 @@ export default class CloseComCalendarService implements Calendar {
...
+        id: contact.id,
+        email: contact.emails[0].email,
+        name: contact.name,
+      };
+    });
   }
 
-  async listCalendars(_event?: CalendarEvent): Promise<IntegrationCalendar[]> {
-    return Promise.resolve([]);
+  async createContacts(contactsToCreate: ContactCreateInput[]): Promise<Contact[]> {
```
joeauyeung, line 138: In Close.com contacts are connected to a lead. So we need to create the lead first before we can create contacts. If there is more than one attendee in the booking, then create all contacts under the same lead.

packages/app-store/closecom/lib/CrmService.ts
```diff
@@ -110,27 +105,76 @@ export default class CloseComCalendarService implements Calendar {
...
+    // In Close.com contacts need to be attached to a lead
+    // Assume all attendees in an event belong under a lead
+
+    const contacts = [];
+
+    // Create main lead
+    const lead = await this.closeCom.lead.create({
```
joeauyeung, line 145: Creating a lead already creates one contact connected to it.

packages/app-store/closecom/lib/CrmService.ts
joeauyeung, line 1: I think Close.com will need some more work that is out of scope of this PR. We should consider reopening this issue https://github.com/calcom/cal.com/issues/10233

packages/app-store/hubspot/lib/CrmService.ts
joeauyeung, line 1: In this file, I was moving the logic that was living in private methods to public ones that conform to the `CrmService` type.

packages/app-store/hubspot/lib/CrmService.ts
```diff
@@ -234,49 +178,11 @@ export default class HubspotCalendarService implements Calendar {
     return Promise.reject("Something went wrong when creating a meeting in HubSpot");
   }
 
-  async createEvent(event: CalendarEvent): Promise<NewCalendarEventType> {
```
joeauyeung, line 237: In every CRM app when creating events, there was logic to determine if contacts had to be created. I moved this into the `CrmManager`.

packages/app-store/hubspot/lib/CrmService.ts
```diff
@@ -234,49 +178,11 @@ export default class HubspotCalendarService implements Calendar {
...
-      const createContacts = await this.hubspotContactCreate(event.attendees);
-      this.log.debug("contact:created", { createContacts });
-      if (createContacts.length) {
-        this.log.debug("contact:creation:ok");
-        return await this.handleMeetingCreation(event, createContacts as SimplePublicObjectInput[]);
-      }
-    }
-    return Promise.reject("Something went wrong when searching/creating the attendees in HubSpot");
+
+    return await this.handleMeetingCreation(event, contacts);
```
joeauyeung, line 185: I want to keep the methods as atomic as possible. The `createEvent` method should only handle creating events.

packages/app-store/salesforce/lib/CrmService.ts
```diff
@@ -312,11 +276,38 @@ export default class SalesforceCalendarService implements Calendar {
...
+  async createContacts(contactsToCreate: { email: string; name: string }[]) {
+    const conn = await this.conn;
+    const createdContacts = await Promise.all(
+      contactsToCreate.map(async (attendee) => {
+        const [FirstName, LastName] = attendee.name ? attendee.name.split(" ") : [attendee.email, ""];
+        return await conn
+          .sobject("Contact")
+          .create({
+            FirstName,
+            LastName: LastName || "-",
```
joeauyeung, line 301: Last name is required when creating new contacts.

packages/core/EventManager.ts
```diff
@@ -97,7 +99,7 @@ export default class EventManager {
     // (type closecom_other_calendar)
     this.calendarCredentials = appCredentials.filter(
       // Backwards compatibility until CRM manager is implemented
-      (cred) => cred.type.endsWith("_calendar") || cred.type.endsWith("_crm")
+      (cred) => cred.type.endsWith("_calendar") && !cred.type.includes("other_calendar")
```
joeauyeung, line 102: Older crm credentials end in `other_calendar` so let's not lump them in with the calendar credentials

packages/core/EventManager.ts
```diff
@@ -908,4 +916,68 @@ export default class EventManager {
...
+          type: credential.type,
+          url: "",
+          credentialId: credential.id,
+          password: "",
+        },
+        id: createdEvent?.id || "",
+        originalEvent: event,
+        credentialId: credential.id,
+      });
+    }
```
joeauyeung, line 946: Not a huge fan of this but this is what the event manager is expecting

packages/core/crmManager/crmManager.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const log = logger.getSubLogger({ prefix: ["CrmManager"] });
+export default class CrmManager {
+  crmService: CRM | null | undefined = null;
+  credential: CredentialPayload;
+  constructor(credential: CredentialPayload) {
+    this.credential = credential;
+  }
+
+  private async getCrmService(credential: CredentialPayload) {
```
joeauyeung, line 15: Every public method will start by calling this. If it's the first time a method is called, get the appropriate `CrmService`

sean-brydon: Would it make sense to just call this in the CTOR of CRM manager to prevent us needing to call 

await this.getCrmService(this.credential);

In every method and future method.  I think adding this.getCrmService() into the ctor would work if we change the implementation to use `this.credential` instead of as an arg (as they are same)

See suggestion for an example

sean-brydon (commented):
Code looks good - one little NIT i spotted. 

I will test this on all CRM apps later today

packages/core/crmManager/crmManager.ts
joeauyeung: I was thinking about this as well but after some reading I learned that constructors should only run synchronous code. I was considering the factory pattern but I think it would have been overkill for this and I want to keep this class similar to what we have now. This is like a pseudo-decorator pattern.

sean-brydon (commented):
As said in DMs:
ZoHo CRM seems to be throwing an unauthorized error whenever i try to create a booking.

Tested all other CRM apps and they are working fine. Could you confirm this works on your end? Could be something with my workspace

packages/app-store/zohocrm/lib/CrmService.ts
```diff
@@ -59,6 +59,8 @@ export default class ZohoCrmCrmService implements CRM {
   }
 
   async createContacts(contactsToCreate: ContactCreateInput[]) {
+    const auth = await this.auth;
+    await auth.getToken();
```
joeauyeung, line 63: These auth functions used to live in their parent methods. Since we're making these public need to include these.

packages/app-store/zohocrm/lib/CrmService.ts
```diff
@@ -80,23 +82,41 @@ export default class ZohoCrmCrmService implements CRM {
     });
 
     const { data } = response.data;
-    return data;
+    return data.data.map((contact) => {
+      return {
+        id: contact.id,
+        email: contact.email,
+      };
+    });
```
joeauyeung, line 90: From the response we need to return the contacts as expected in our app

packages/app-store/zohocrm/lib/CrmService.ts
```diff
@@ -174,7 +194,7 @@ export default class ZohoCrmCrmService implements CRM {
       throw new HttpError({ statusCode: 400, message: "Zoho CRM client_secret missing." });
     const credentialKey = credential.key as unknown as ZohoToken;
     const isTokenValid = (token: ZohoToken) => {
-      const isValid = token && token.access_token && token.expiryDate && token.expiryDate < Date.now();
+      const isValid = token && token.access_token && token.expiryDate && token.expiryDate > Date.now();
```
joeauyeung, line 197: The token should be valid if the `expiryDate` is greater than the current date

sean-brydon (dismissed):
Can confirm the PR now includes fixes for Zoho. 

Nice job @joeauyeung much needed refactor

zomars (commented):
Overall LGTM. Do we have CRM Service specific tests? Can we come up with some?

zomars:
Also quick Q. What will happen to all the current CRM installs in prod?

joeauyeung:
@zomars no tests right now but I can create some along with some `EventManager` tests.

In terms of current CRM installs they'll still be enabled for all event types by default. This is our current behavior. 

This PR changes how CRM apps are triggered: https://github.com/calcom/cal.com/pull/14450

CarinaWolli:
@joeauyeung there are type errors

packages/app-store/_components/OmniInstallAppButton.tsx
```diff
@@ -61,7 +61,6 @@ export default function OmniInstallAppButton({
                 type: app.type,
                 variant: app.variant,
                 slug: app.slug,
-                isOmniInstall: true,
```
hariombalhara, line 64: Why is it removed? 

I see usage of `isOmniInstall` still in the mutation code https://github.com/calcom/cal.com/blob/45fe2a727eaf774b6e972bbc432602faae50fba6/packages/app-store/_utils/useAddAppMutation.ts#L57

packages/app-store/_utils/oauth/createOAuthAppCredential.ts
```diff
@@ -44,7 +44,7 @@ const createOAuthAppCredential = async (
 
     if (!team) throw new Error("User does not belong to the team");
 
-    await prisma.credential.create({
+    const credential = await prisma.credential.create({
```
hariombalhara, line 47:
```suggestion
   return await prisma.credential.create({
```

packages/app-store/_utils/oauth/createOAuthAppCredential.ts
```diff
@@ -53,12 +53,12 @@ const createOAuthAppCredential = async (
...
     });
 
-    return;
+    return credential;
   }
 
   await throwIfNotHaveAdminAccessToTeam({ teamId: state?.teamId ?? null, userId });
 
-  await prisma.credential.create({
+  const credential = await prisma.credential.create({
```
hariombalhara, line 61:
```suggestion
  return await prisma.credential.create({
```

packages/app-store/_utils/writeAppDataToEventType.ts
```diff
@@ -0,0 +1,75 @@
...
+  credentialId,
+}: {
+  userId?: number;
+  teamId?: number;
+  appSlug: string;
+  appCategories: AppCategories[];
+  credentialId: number;
+}) => {
+  //   Search for event types belonging to the user / team
+  const eventTypes = await prisma.eventType.findMany({
```
hariombalhara, line 21: It could be a good candidate for a method in `EventTypeRepository`. The name could be `findByTeamOrUserOwner`.

packages/app-store/_utils/writeAppDataToEventType.ts
```diff
@@ -0,0 +1,75 @@
...
+          id: eventType.id,
+        },
+        data: {
+          metadata,
+        },
+      })
+    );
+  }
+
+  await Promise.all(updateEventTypeMetadataPromises);
```
hariombalhara, line 72: nit: Could be simplified as 
image

packages/core/EventManager.ts
```diff
@@ -566,9 +581,7 @@ export default class EventManager {
        *  fallback to the first connected calendar - Shouldn't be a CRM calendar
        */
       // Backwards compatibility until CRM manager is created
-      const [credential] = this.calendarCredentials.filter(
-        (cred) => !cred.type.endsWith("other_calendar") || !cred.type.endsWith("crm")
```
hariombalhara, line 570: Though this code is being modified now but I think it should have been `!cred.type.endsWith("other_calendar") && !cred.type.endsWith("crm")`

packages/core/EventManager.ts
```diff
@@ -566,9 +581,7 @@ export default class EventManager {
...
-      );
+      const [credential] = this.calendarCredentials.filter((cred) => !cred.type.endsWith("other_calendar"));
```
hariombalhara, line 584: Don't we still need to skip all CRM calendar credentials(including those that end in `_crm`) as we are in `createAllCalendarEvents` method

packages/core/EventManager.ts
```diff
@@ -673,7 +686,7 @@ export default class EventManager {
       await Promise.all(
         this.calendarCredentials
           // Backwards compatibility until CRM manager is created
```
hariombalhara, line 688: This and other similar comments might be outdated

packages/core/EventManager.ts
```diff
@@ -673,7 +686,7 @@ export default class EventManager {
       await Promise.all(
```
hariombalhara, line 686: Shouldn't se just delete this code otherwise calendar events could be created twice.

One due to this and the other due to createAllCRMEvents(as that includes both old and new CRM credentials)

packages/core/EventManager.ts
```diff
@@ -907,4 +920,68 @@ export default class EventManager {
...
   }
+
+  private async createAllCRMEvents(event: CalendarEvent) {
+    const createdEvents = [];
+    for (const credential of this.crmCredentials) {
+      const crm = new CrmManager(credential);
+
+      let success = true;
+      const createdEvent = await crm.createEvent(event).catch((error) => {
+        success = false;
```
hariombalhara, line 931: We should log the error

packages/core/EventManager.ts
```diff
@@ -907,4 +920,68 @@ export default class EventManager {
...
+      });
+      createdEvents.push({
+        type: credential.type,
+        appName: credential.appId || "",
+        uid: createdEvent?.id || "",
```
hariombalhara, line 936: We are using `getUid` method  to get the UID in CalendarManager::createEvent, shouldn't we reuse that?

packages/core/EventManager.ts
hariombalhara: I think we should add all props that are there and explicitly add the values as undefined/null to make it clear that certain props are intentionally left undefined and why,

packages/core/crmManager/crmManager.ts
hariombalhara: I think we can change the approach to something like this atleast. It keeps the code almost same but still making it more intuitive.

image

packages/core/crmManager/crmManager.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (contacts.length == event.attendees.length) {
+      return await this.crmService?.createEvent(event, contacts);
+    } else {
+      // Figure out which contacts to create
+      const contactsToCreate = event.attendees.filter(
+        (attendee) => !contacts.some((contact) => contact.email === attendee.email)
+      );
+      const createdContacts = await this.createContacts(contactsToCreate);
+      contacts = contacts.concat(createdContacts);
+      return await this.crmService?.createEvent(event, contacts);
```
hariombalhara, line 40: Could be simplified by removing `else` 
Screenshot 2024-05-03 at 3 43 59 PM

packages/core/crmManager/crmManager.ts
```diff
@@ -0,0 +1,65 @@
...
+    await this.getCrmService(this.credential);
+    return await this.crmService?.updateEvent(uid, event);
+  }
+
+  public async deleteEvent(uid: string) {
+    await this.getCrmService(this.credential);
+    return await this.crmService?.deleteEvent(uid);
+  }
+
+  public async getContacts(email: string | string[]) {
```
hariombalhara, line 54: nit
```suggestion
  public async getContacts(emailOrEmails: string | string[]) {
```

packages/app-store/zohocrm/lib/CrmService.ts
```diff
@@ -82,23 +81,44 @@ export default class ZohoCrmCalendarService implements Calendar {
       },
       data: JSON.stringify({ data: contacts }),
     });
-  };
 
-  private contactSearch = async (event: CalendarEvent) => {
-    const searchCriteria = `(${event.attendees
-      .map((attendee) => `(Email:equals:${encodeURI(attendee.email)})`)
-      .join("or")})`;
+    const { data } = response.data;
```
hariombalhara, line 85: It should be 
`const {data} = response`

hariombalhara: Right now it is throwing error.

hariombalhara (changes requested):
Great changes @joeauyeung!! Really makes CRM app a first class citizen.

I have left a few comments. Some might be blocking. 


Also, I wanted to mention that this change deserves integration tests for CRM flow. So, we could maybe write a few `handleNewBooking` tests for this.

packages/app-store/_components/OmniInstallAppButton.tsx
joeauyeung: Heads up that https://github.com/calcom/cal.com/pull/14450 was merged early into this branch so a lot of context was lost. 

I thought this prop wasn't being used but good catch.

packages/core/EventManager.ts
joeauyeung: We sort the CRM credentials into `crmCredentials` array so they won't be included in the `calendarCredentials` array that's being filtered on.

packages/app-store/_utils/writeAppDataToEventType.ts
```diff
@@ -0,0 +1,57 @@
...
+  teamId?: number;
+  appSlug: string;
+  appCategories: AppCategories[];
+  credentialId: number;
+}) => {
+  //   Search for event types belonging to the user / team
+  const eventTypes = teamId
+    ? await EventTypeRepository.findAllByTeamIdIncludeManagedEventTypes({ teamId })
+    : userId
+    ? await EventTypeRepository.findAllByUserId({ userId })
```
joeauyeung, line 25: @hariombalhara I've created two methods to the `EventTypeRepository` it feels like once we start adding "or" to the method name we should split it into two separate methods.

packages/embeds/embed-react/vite.config.js
```diff
@@ -27,6 +27,12 @@ export default defineConfig({
           "react-dom": "ReactDOM",
         },
       },
+      onLog(level, log, handler) {
+        if (log.cause && log.cause.message === `Can't resolve original location of error.`) {
+          return;
+        }
+        handler(level, log);
+      },
```
hariombalhara, line 35: How did the need for this change arise?

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -151,8 +180,7 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
               metadata: {
                 ...metadata,
                 apps: {
-                  ...metadata?.apps,
-                  [appSlug]: undefined,
```
hariombalhara, line 155: I am unable to catch the bug we are fixing here by using `removeAppFromEventTypeMetadata`

hariombalhara: Also, if we still want to use a fn for some reason, I think we should make a generic one, something like `removePropFromObject`

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -136,11 +137,39 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
...
     }
 
+    if (credential.app?.categories.includes(AppCategories.crm)) {
+      const metadata = EventTypeMetaDataSchema.parse(eventType.metadata);
+      const appSlugToDelete = credential.app?.slug;
+
+      if (appSlugToDelete) {
+        const appMetadata = removeAppFromEventTypeMetadata(appSlugToDelete, metadata);
+
+        await prisma.$transaction(async () => {
```
hariombalhara, line 147: We don't need a transaction here it seems

hariombalhara, line 147: This code seems same as the following code for payment app handling. We should merge these two.

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
```diff
@@ -72,5 +75,44 @@ export const getAllCredentials = async (
...
+  }
+
+  allCredentials = allCredentials.filter((credential) => {
+    if (!credential.type.includes("_crm") && !credential.type.includes("_other_calendar")) {
+      return credential;
+    }
+
+    // Backwards compatibility: All CRM apps are triggered for every event type. Unless disabled on the event type
+    // Check if the CRM app exists on the event type
+    if (eventTypeCrmCredentials[credential.id]) {
```
hariombalhara, line 100: Following code could be abstracted out in something like `filterOutDisabledCrmCredentials`. Also that makes me realize, why do we check for CRM only. Shouldn't we check for any app here.

We could assume that if `enabled` is undefined it is actually true otherwise whatever it says.

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
```diff
@@ -72,5 +75,44 @@ export const getAllCredentials = async (
...
+      if (eventTypeCrmCredentials[credential.id].enabled) {
+        return credential;
+      }
+    } else {
+      // If the CRM app doesn't exist on the event type metadata, check that the credential belongs to the user/team/org
+      if (
+        credential.userId === eventType?.userId ||
+        credential.teamId === eventType?.team?.id ||
+        credential.teamId === eventType?.parentId
```
hariombalhara, line 109: I am confused, so needs confirmation. Aren't we comparing teamId with eventTypeId here as eventType.parentId is parent event's ID.

packages/lib/server/repository/eventType.ts
```diff
@@ -185,4 +185,27 @@ export class EventTypeRepository {
...
+
+  static async findAllByTeamIdIncludeManagedEventTypes({ teamId }: { teamId?: number }) {
+    return await prisma.eventType.findMany({
+      where: {
+        OR: [
+          {
+            teamId,
+          },
+          {
+            parentId: teamId,
```
hariombalhara, line 205: We are mistakenly querying for eventType using `parentId=teamId.` parentId is supposed to be parent event's ID which has its teamId that we should compare with.

packages/app-store/_utils/writeAppDataToEventType.ts
```diff
@@ -0,0 +1,57 @@
+import { EventTypeRepository } from "@calcom/lib/server/repository/eventType";
+import prisma from "@calcom/prisma";
+import type { AppCategories } from "@calcom/prisma/enums";
+import { EventTypeMetaDataSchema } from "@calcom/prisma/zod-utils";
+
+import type { appDataSchemas } from "../apps.schemas.generated";
+
+const writeAppDataToEventType = async ({
```
hariombalhara, line 8: Just realized that this could cause a lot of eventType updates in parallel possibly impacting DB.

On a CRM app's install, we are updating all existing event-types to set `enabled:false` to get the new behaviour of no-automatic CRM connection from event-type. But if we create a new event-type, after installing the new CRM app, it would still have the CRM app automatically enabled for that event. This is because appMetadata isn't set for that event-type and we end up thinking that the app should be considered enabled.	


What if we identify that the installation is a new one and then decide the behaviour from that. **In this way, we don't need to update all event-types and the bug I mentioned above is also fixed automatically.**

One approach could be to check if `credential.type.endsWith('_crm').` I know that the change that switched to `_crm` has been live for some time but it could be very few people having installed the crm apps after that. We could run a migration for such users maybe to move them back to `_other_calendar`.

But the point is that once I know that the credential is a new installation, we can then simply decide based on that if we need to consider the default value of `enabled`(when it is undefined) to be true or false.


This should remove the requirement of calling this function.

packages/app-store/zoho-bigin/lib/CrmService.ts
```diff
@@ -141,16 +140,18 @@ export default class BiginCalendarService implements Calendar {
       },
       data: JSON.stringify({ data: contacts }),
     });
+
+    return response.data;
```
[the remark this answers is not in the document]
hariombalhara: Also, I am thinking that we should replace all usages of axiom in AppStore to our own wrapper of Axiom that would make the `data` as `unknown` instead of `any`. This would fail TS check which is passing right now.

packages/app-store/zoho-bigin/lib/CrmService.ts
```diff
@@ -141,16 +140,18 @@ export default class BiginCalendarService implements Calendar {
...
+  async getContacts(emails: string | string[]) {
     const token = await this.auth.getToken();
-    const searchCriteria = `(${event.attendees
-      .map((attendee) => `(Email:equals:${encodeURI(attendee.email)})`)
-      .join("or")})`;
+    const emailsArray = Array.isArray(emails) ? emails : [emails];
+
+    const searchCriteria = `(${emailsArray.map((email) => `(Email:equals:${encodeURI(email)})`).join("or")})`;
 
     return await axios({
```
hariombalhara, line 156: Same issue is here I believe.

hariombalhara (dismissed):
I have identified a few blockers(marked in comment) that i think we should fix at the least.

Apart from that I think we should avoid `writeAppDataToEventType` approach as mentioned as comment on the file.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
joeauyeung: On the event types page we parse the app metadata. If one of the apps is undefined it throws a zod error and stops the loading of the event types page completely.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
joeauyeung: This is interesting. The idea is that we should be updating all the event types as one unit but this isn't the case. This logic is similar for other event type apps. Should we follow up with a refactor as I think this is happening in other parts of the handler.

packages/app-store/zoho-bigin/lib/CrmService.ts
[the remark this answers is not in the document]
joeauyeung: I'm not a huge fan of using axiom in our app store. What if we just followed up with replacing instances of it with `fetch`?

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
joeauyeung: Payment apps are handled in a different function. Ideally that logic should live here right?

hariombalhara: Yeah. If it is similar logic it should be together only.

packages/app-store/zoho-bigin/lib/CrmService.ts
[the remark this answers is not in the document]
hariombalhara: Yeah we could do that as an enhancement. Easy to pick by a contributor.

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
hariombalhara: Yeah we can fix it in a followup

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
hariombalhara: That is weird. This change is really old. Maybe something change elsewhere recently. 

I am okay with this change for now.

packages/app-store/_components/AppCard.tsx
zomars, line 1: This file will need a refactor soon. It has way too many props and ternaries.

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -113,3 +113,17 @@ function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMuta
...
+  returnTo,
+}: {
+  stateStr: string;
+  teamId?: number;
+  returnTo?: string;
+}) => {
+  const teamIdParam = teamId ? `&teamId=${teamId}` : "";
+  const returnToParam = returnTo ? `&returnTo=${returnTo}` : "";
+  return `?state=${stateStr}${teamIdParam}${returnToParam}`;
+};
```
zomars, line 129: Not a blocker but let's rely on constructing URLs using the the URL class or our own extended one.

https://nodejs.org/api/url.html
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

packages/app-store/closecom/components/EventTypeAppCardInterface.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({ app, eventType }) {
+  const pathname = usePathname();
+
+  const { enabled, updateEnabled } = useIsAppEnabled(app);
+
+  // CRM backwards compatibility
+  if (enabled === undefined) {
+    updateEnabled(true);
+  }
```
zomars, line 16: This will trigger 2 renders on component mount. Is there a way tell the `useIsAppEnabled` hook to default to `true` if enabled is undefined?

sean-brydon: I think i have resolved this issue @zomars - made changes to the useIsAppEnabled hook. Please take a look at the latest commit

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.test.ts
```diff
@@ -0,0 +1,699 @@
...
+
+  describe("Handle CRM credentials", () => {
+    describe("If CRM is enabled on the event type", () => {
+      describe("With _crm credentials", () => {
+        test("For users", async () => {
+          vi.spyOn(UserRepository, "enrichUserWithItsProfile").mockReturnValue({
+            profile: null,
+          });
+
+          const getAllCredentials = (await import("./getAllCredentials")).getAllCredentials;
```
sean-brydon, line 67: This should be mocked i think?

sean-brydon: Or maybe not we just have instances of prisma being called in here when teamId is present

keithwillcode:
@zomars Can you please review this one more time.

keithwillcode:
Running all checks against this branch here: https://github.com/calcom/cal.com/actions/runs/9276744136

emrysal (dismissed):
Well justified changes @joeauyeung - late to the party in terms of review but it looks good to me.

keithwillcode:
@joeauyeung @CarinaWolli @sean-brydon Quite a few tests failing here. https://github.com/calcom/cal.com/actions/runs/9276788573

ibex088 (commented):
https://www.loom.com/share/f1b52b7c4d044a9fbcc3e8d05c164688?sid=f67e15ec-74ea-475d-8be5-879165effdbf

is this intentional - 
1. install any analytics app
2. click "set up later"
3. go to event-type 
4. the app is enabled, where as it should be disabled

ibex088 (commented):
Just wanted to share that with the new app flow installed and CRM apps now extending eventTypes, the process for installing flow for CRM apps has changed. Here's a demonstration of the updated installation process:

[Watch the video](https://www.loom.com/share/ffde7397cb4e4025ab7a7e207d54d55d?sid=6688fdba-6dd8-4bb1-8a8c-165aeecfe97d)

Note that I only installed Zoho for 2 events, but it is enabled for all personal account events.

sean-brydon:
> Just wanted to share that with the new app flow installed and CRM apps now extending eventTypes, the process for installing flow for CRM apps has changed. Here's a demonstration of the updated installation process:
> 
> [Watch the video](https://www.loom.com/share/ffde7397cb4e4025ab7a7e207d54d55d?sid=6688fdba-6dd8-4bb1-8a8c-165aeecfe97d)
> 
> Note that I only installed Zoho for 2 events, but it is enabled for all personal account events.

@SomayChauhan are you able to take this up to look into why this is happening? I feel like you may be the most comfortable with this part of app store code alongside @hariombalhara and @joeauyeung

SomayChauhan:
>@SomayChauhan are you able to take this up to look into why this is happening? I feel like you may be the most comfortable with this part of app store code alongside @hariombalhara and @joeauyeung

 I'm working on it!

keithwillcode:
@zomars @SomayChauhan was able to get the tests passing! We are ready to ship this separately today. https://github.com/calcom/cal.com/actions/runs/9302242199

keithwillcode (approved):

PR #14278, opened 2024-04-01
fix: `Booker` atom modal not rendering for mobile devices

Ryukemeister:
## What does this PR do?

The `Booker` atom displays a modal for small screen devices like the one we have below.

Screenshot 2024-04-01 at 3 59 25 PM

Previously we were only displaying this modal based on the logic that if the component is not in platform i.e. `!isPlatform` only then the modal is displayed. Because of this the booker atom never displays the modal meant for mobile devices and so a user using a small screen device wouldnt be able to make a booking through our atom. This PR aims to display the booking modal for platform as well as the web app. I modified the wrapper being used in the modal such that only if we're in the web app then it makes trpc calls otherwise we don't use trpc if we're in platform

### DIFF
--- a/packages/features/bookings/Booker/Booker.tsx
+++ b/packages/features/bookings/Booker/Booker.tsx
@@ -447,15 +447,11 @@ const BookerComponent = ({
         )}
       </div>
 
-      {!isPlatform ? (
-        <BookFormAsModal
-          onCancel={() => setSelectedTimeslot(null)}
-          visible={bookerState === "booking" && shouldShowFormInDialog}>
-          {EventBooker}
-        </BookFormAsModal>
-      ) : (
-        <></>
-      )}
+      <BookFormAsModal
+        onCancel={() => setSelectedTimeslot(null)}
+        visible={bookerState === "booking" && shouldShowFormInDialog}>
+        {EventBooker}
+      </BookFormAsModal>
     </>
   );
 };
--- a/packages/features/bookings/Booker/components/BookEventForm/BookFormAsModal.tsx
+++ b/packages/features/bookings/Booker/components/BookEventForm/BookFormAsModal.tsx
@@ -1,6 +1,7 @@
 import { Calendar, Clock } from "lucide-react";
 import type { ReactNode } from "react";
 
+import { useIsPlatform } from "@calcom/atoms/monorepo";
 import dayjs from "@calcom/dayjs";
 import { useLocale } from "@calcom/lib/hooks/useLocale";
 import { Badge, Dialog, DialogContent } from "@calcom/ui";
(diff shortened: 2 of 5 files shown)


packages/features/bookings/Booker/components/BookEventForm/BookFormAsModal.tsx
```diff
@@ -9,11 +10,33 @@ import { useTimePreferences } from "../../../lib";
...
+};
+
+export const BookEventFormWrapperComponent = ({
+  child,
+  eventLength,
+}: {
+  onCancel: () => void;
+  child: ReactNode;
+  eventLength?: number;
+}) => {
```
ThyMinimalDev, line 36: event could be passed as a props here tbh, we already do it multiple times in the rest of the booker

packages/features/bookings/Booker/Booker.tsx
```diff
@@ -447,15 +447,11 @@ const BookerComponent = ({
...
 
-      {!isPlatform ? (
-        <BookFormAsModal
-          onCancel={() => setSelectedTimeslot(null)}
-          visible={bookerState === "booking" && shouldShowFormInDialog}>
-          {EventBooker}
-        </BookFormAsModal>
-      ) : (
-        <></>
-      )}
```
ThyMinimalDev, line 458: not sure to understand how this was preventing the modal to appear for web ?

packages/platform/atoms/hooks/event-types/useGetEventTypeById.ts
```diff
@@ -0,0 +1,27 @@
+import { useQuery } from "@tanstack/react-query";
+
+import { BASE_URL, API_VERSION, V2_ENDPOINTS, SUCCESS_STATUS } from "@calcom/platform-constants";
+import type { EventType as AtomEventType } from "@calcom/platform-libraries";
+import type { ApiResponse, ApiSuccessResponse } from "@calcom/platform-types";
+
+import http from "../../lib/http";
+
+export const QUERY_KEY = "get-event-by-id";
+export const useGetEventTypeById = (id: number | null) => {
```
ThyMinimalDev, line 10: this does not seem to be used in this pr ?

packages/features/bookings/Booker/components/BookEventForm/BookFormAsModal.tsx
ThyMinimalDev: this would remove the need to split into platform / web wrapper

Ryukemeister: We only need the length of the event here. The trpc handler that we have inside of the `useEvent` hook returns the event length property, however we don't return event length property in platform endpoint for event types. As ai result had to split this into platform / web wrapper

ThyMinimalDev (dismissed):

packages/features/bookings/Booker/Booker.tsx
Ryukemeister: The `BookFormAsModal` is the actual component which renders the modal. Based on the above logic only if we're inside web we display the modal. As a result when we're in platform (atoms) we don't display anything, so a user won't be able to proceed further with his booking if we don't display the modal.

packages/platform/atoms/hooks/event-types/useGetEventTypeById.ts
Ryukemeister: Yeah this was mainly for future reference in case we might need a hook like this.

packages/features/bookings/Booker/Booker.tsx
ThyMinimalDev: gotcha

ThyMinimalDev (approved):

PR #14280, opened 2024-04-01
fix: correct language is used when booking with a modal + location icon visible in booker when using light mode

Jojo14234:
## What does this PR do?

This PR fixes the following:
- The "location" icon is now visible in light move when using the "in person (Organizer Address)" and "Show On Booking Screen" option
- The language in which the date and length of an event is shown in modals now corresponds to the language of the creator of the calendar (before this is was always in english) and also shows the unit in which the event length is indicated (before this, only a number was shown)

- If the text is too wide for the modal (which before this never happened because it was always in English), they will adapt into a column

- Refer to the screenshots to get a better idea of the impact of this PR.

Before screenshots:
Screenshot 2024-04-01 at 13 13 38
Screenshot 2024-04-01 at 13 14 24

After screenshots:
Screenshot 2024-04-01 at 13 14 41
Screenshot 2024-04-01 at 13 14 48
Screenshot 2024-04-01 at 13 15 36

PS: please note that the word "minute" is the same in English and French so it kind of looks like it isn't using the correct language in the screenshots but it is :). Probably better to test with another language than French

## Requirement/Documentation

N/A

## Type of change

## How should this be tested?

Book an appointment with the column view in any other language than English while using light mode. Be sure to include an "in person (Organizer Address" address so you can confirm that the "location" icon is visible.


packages/features/bookings/components/event-meta/Duration.tsx
```diff
@@ -9,7 +9,10 @@ import { Badge } from "@calcom/ui";
 import type { PublicEvent } from "../../types";
 
 /** Render X mins as X hours or X hours Y mins instead of in minutes once >= 60 minutes */
-const getDurationFormatted = (mins: number, t: TFunction) => {
+export const getDurationFormatted = (mins: number | null | undefined, t: TFunction) => {
```
Udit-takkar, line 12: This won't work if mins is undefined

Jojo14234: Hi, would you mind being more specific about why it won't work? Sorry if this is relatively basic, I'm still a student :).

Udit-takkar (dismissed):
LGTM

packages/features/bookings/components/event-meta/Duration.tsx
Udit-takkar: you were only returning null when mins == null and not checking undefined

Jojo14234: Hi, I was basing my code on this answer (https://stackoverflow.com/questions/28975896/is-there-a-way-to-check-for-both-null-and-undefined) which much not be correct... Just wanted to bring to your attention that the way you corrected this will not work if the length of an event is 0 minutes (don't know if that's a possibility but that's why I had coded it the way I did instead of the way you did). Cheers

Udit-takkar:
@Jojo14234 can you please fix the conflicts? you can tag me if you face any problem

Udit-takkar (approved):

Jojo14234:
@Udit-takkar thx for taking the time to fix the conflicts... i had not had time to look into it... any idea of when this could enter the live version of cal.com? I myself am a user and this issue is very annoying since it makes the mobile expierience for non-English speakers pretty poor.

PR #13139, opened 2024-01-10
feat: new org onboarding wizard

PeerRich:
previously we required org members to create a new account. We are now making it work to upgrade an existing team to an organisation ([subdomain.cal.com](http://subdomain.cal.com)) without breaking previous links and redirecting back.

---
Sean Looms:
* Admin Creating custom pricing plan: https://www.loom.com/share/119fa5db22c44138a645ce434fccd2ef

* User creating org + paying + migrating teams (long one): https://www.loom.com/share/a5d3bc6276e049f492b8b0046a4f6234
---

Hariom Looms
Demo Loom-1  https://www.loom.com/share/25f674ac80e14b7da869e4868bb57ef8
Demo Loom-2  https://www.loom.com/share/a5f83e6d206648e59cfd07b34b70361a
[Demo that shows that in self serve Owner can pay for the org during onboarding itself.
](https://www.loom.com/share/87a3339ed5a24be0b9501712ed2c1116)
this needs a migration path triggered by the team owner:

--- 

- [x] team owner sees a banner with more than 2 teams: "You have a lot of teams. Consider making an organisation to get a subdomain, unify links, billing and analytics"
- [x] if OrgPlan: `<Button>Set up your Organization</Button>`
- [x] if NotOrgPlan: `<Button>Contact Sales</Button>`
- [x] after clicking the Get Started button, a team owner can select all teams (check boxes) in step 1 
- [x] IS_ADMIN or IS_IMPERSONATING: be able to create an Org on behalf of someone
   - [x] be able to change amount of seats and price per seat

CleanShot 2024-01-18 at 20 21 42

- [x] and choose a new name for the team in step 2:

| old. | new |
| -- | -- |
| cal.com/team/acme-sales | acme.cal.com/sales |

- [x] we can pre-fill these inputs with a simple regex to remove "acme-" or "-acme" if subdomain = acme
- [x] we should mention that even tho links change, they will continue to work with a redirect
- [x] after clicking: "Done", every member gets an email:

> Hey Peer.
>
> Anna invited you to the **Acme** organisation. Your link has been changed from [cal.com/peer-richelsen-acme](http://cal.com/peer-richelsen-acme) to [acme.cal.com/peer](http://acme.cal.com/peer) but dont worry, all previous links still work and redirect appropriately.
>
> Dan has upgraded the previous [cal.com](http://cal.com/) account to unify usernames under one subdomain. Every team member link is now according to their email at **Acme**:
>
[...]

PR #14284, opened 2024-04-02
feat: isPlatform toggle during org creation

supalarry:
Platform orgs are setup manually, so add a toggle that sets "Team.isPlatform" when creating the org. By default it is false:

Screenshot 2024-04-02 at 10 44 40

Platform org in database has "isPlatform" true:
Screenshot 2024-04-02 at 10 45 04

After this "settings/organizations/platform/oauth-clients" can be browsed to create client and because org has "isPlatform" client creation is successful. Non platform orgs can't create OAuth clients.

### DIFF
--- a/packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
+++ b/packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
@@ -10,7 +10,7 @@ import { useLocale } from "@calcom/lib/hooks/useLocale";
 import slugify from "@calcom/lib/slugify";
 import { telemetryEventTypes, useTelemetry } from "@calcom/lib/telemetry";
 import { trpc } from "@calcom/trpc/react";
-import { Button, Form, TextField, Alert } from "@calcom/ui";
+import { Button, Form, TextField, Alert, SettingsToggle } from "@calcom/ui";
 import { ArrowRight } from "@calcom/ui/components/icon";
 
 function extractDomainFromEmail(email: string) {
@@ -34,9 +34,11 @@ export const CreateANewOrganizationForm = ({ slug }: { slug?: string }) => {
     slug: string;
     adminEmail: string;
     adminUsername: string;
+    isPlatform: boolean;
   }>({
     defaultValues: {
       slug: `${slug ?? ""}`,
+      isPlatform: false,
     },
   });
   const watchAdminEmail = newOrganizationFormMethods.watch("adminEmail");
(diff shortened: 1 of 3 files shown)


ThyMinimalDev (approved):

sean-brydon:
Do you think this could wait till https://github.com/calcom/cal.com/pull/13139 is merged? 

We can handle this only in the admin view in this PR instead of this PR being anyone can create a org with this as its a public URL

cc: @hariombalhara

supalarry:
> Do you think this could wait till #13139 is merged?
> 
> We can handle this only in the admin view in this PR instead of this PR being anyone can create a org with this as its a public URL
> 
> cc: @hariombalhara

@sean-brydon do you know estimated time to merge #13139 ? Also, I am not sure i followed "We can handle this only in the admin view in this PR instead of this PR being anyone can create a org with this as its a public URL" - so is there a way to accomplish what this PR does before yours gets merged? I didn't understand the connection to the public URL since only admins can create orgs, and non-admins can visit this page but can't actually create org.

hariombalhara:
Soon, a regular user will be able to create an org.

packages/features/ee/organizations/components/CreateANewOrganizationForm.tsx
```diff
@@ -196,6 +198,25 @@ export const CreateANewOrganizationForm = ({ slug }: { slug?: string }) => {
...
+                  title={t("organization_is_platform")}
+                  checked={value}
+                  onCheckedChange={(checked) => {
+                    newOrganizationFormMethods.setValue("isPlatform", checked);
+                  }}
+                />
+              </div>
+            )}
+          />
+        </div>
```
PeerRich, line 218: is this for admins only? if yes, it should be wrapped in isAdmin

if this is for self-serve too, we should instead build a self-serve checkout for platform customerss

PeerRich: https://app.cal.com/settings/platform/new can reuse some of the same UI but is more tailored towards platform inputs

PeerRich (changes requested):

sean-brydon:
> > Do you think this could wait till #13139 is merged?
> > We can handle this only in the admin view in this PR instead of this PR being anyone can create a org with this as its a public URL
> > cc: @hariombalhara
> 
> @sean-brydon do you know estimated time to merge #13139 ? Also, I am not sure i followed "We can handle this only in the admin view in this PR instead of this PR being anyone can create a org with this as its a public URL" - so is there a way to accomplish what this PR does before yours gets merged? I didn't understand the connection to the public URL since only admins can create orgs, and non-admins can visit this page but can't actually create org.

Sorry i made no sense lol.

Hopefully we can get the migration PR merged pretty soon. Just concerned about this as in that PR this page isnt just for admins only and we'd have to fix it while merging with this branch.

sean-brydon:
@supalarry we have merged self serve 🙏

PR #14285, opened 2024-04-02
chore: rewrite api.cal.com/v2 to api v2

ThyMinimalDev:
## What does this PR do?

Rewrite api v1 requests to new api v2


apps/api/v1/next.config.js
```diff
@@ -47,6 +47,22 @@ const nextConfig = {
...
+        {
+          source: "/api/v2",
+          destination: `${process.env.NEXT_PUBLIC_API_V2_ROOT_URL}/health`,
+        },
+        {
+          source: "/api/v2/health",
+          destination: `${process.env.NEXT_PUBLIC_API_V2_ROOT_URL}/health`,
+        },
+        {
+          source: "/api/v2/docs/:path*",
```
supalarry, line 59: Idk if we need ":path*" for "docs", because either it is "/docs" or "/docs-yaml" returning docs in yaml format, so i would have "api/v2/docs" source and another object with "api/v2/docs-yaml" source.

supalarry (dismissed):
Looks good with 1 comment below on docs:

apps/api/v1/next.config.js
ThyMinimalDev: we need the path because of the scripts being downloaded for swagger ui

emrysal (approved):
Looks good to me.

apps/api/v1/next.config.js
```diff
@@ -47,6 +47,22 @@ const nextConfig = {
           source: "/v:version/:rest*",
           destination: "/api/v:version/:rest*",
         },
+        {
+          source: "/api/v2",
+          destination: `${process.env.NEXT_PUBLIC_API_V2_ROOT_URL}/health`,
```
keithwillcode, line 52: Root intentionally goes to /health?

keithwillcode (approved):

apps/api/v1/next.config.js
emrysal: Common practise yeah. Either /health or a /docs EP

PR #14304, opened 2024-04-03
refactor: delete OAuth managed users on OAuth client delete

supalarry:
1. When deleting OAuth client also delete users connected to it in [refactor: delete OAuth managed users on OAuth client delete](https://github.com/calcom/cal.com/pull/14304/commits/d4e5d6c9602208dfc756390dc93ddb243faba0db)
2. When deleted return only deleted OAuth client id and name, do not return the secret - even if the OAuth client is deleted just in case something failed do not expose the secret in
[refactor: hide deleted oauth client secret](https://github.com/calcom/cal.com/pull/14304/commits/7d36403ce9accd4264ee86bc3abf6ccd7cf59c67)

### DIFF
--- a/apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
+++ b/apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
@@ -95,8 +95,21 @@ export class OAuthClientRepository {
   }
 
   async deleteOAuthClient(clientId: string): Promise<PlatformOAuthClient> {
-    return this.dbWrite.prisma.platformOAuthClient.delete({
-      where: { id: clientId },
+    return await this.dbWrite.prisma.$transaction(async (prisma) => {
+      await prisma.user.deleteMany({
+        where: {
+          platformOAuthClients: {
+            some: {
+              id: clientId,
+            },
+          },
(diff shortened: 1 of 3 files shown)


apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
```diff
@@ -95,8 +95,21 @@ export class OAuthClientRepository {
   }
 
   async deleteOAuthClient(clientId: string): Promise<PlatformOAuthClient> {
-    return this.dbWrite.prisma.platformOAuthClient.delete({
-      where: { id: clientId },
+    return await this.dbWrite.prisma.$transaction(async (prisma) => {
```
keithwillcode, line 98: This is a scary query. Can we please add automated tests to ensure it only deletes the users we want it to delete?

keithwillcode (changes requested):

apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
keithwillcode: Why do we want to delete the users?

PR #14308, opened 2024-04-03
feat: Round robin reassignment via round robin algorithm [CAL-3138]

joeauyeung:
## What does this PR do?

This PR allows teams to reassign round robin hosts via the round robin algorithm.

Fixes # (issue)

https://www.loom.com/share/3baf73e0a4754719871c6def104dabad

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a round robin event with no fixed hosts
- Create a booking
- Go to the bookings page and under edit choose the reassign action
	- The booking should now have a new round robin host
	- The calendar events should reflect this
- Do this again with fixed hosts
	- The calendar event should still belong to the fixed host but the attendees would have changed

## Mandatory Tasks


packages/features/ee/round-robin/roundRobinReassignment.ts
joeauyeung, line 1: Abstracted so it can be used via the API as well @alishaz-polymath

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  if (!booking?.user) {
+    console.error(`No user associated with booking ${bookingId}`);
+    throw new Error("Booking not found");
+  }
+
+  const originalOrganizer = booking.user;
+
+  const attendeeEmailsSet = new Set(booking.attendees.map((attendee) => attendee.email));
```
joeauyeung, line 84: This set contains emails of both team members and attendees. We create a set because we're comparing team members to attendees in a few places. Rather than iterating through both arrays with a complexity of O(n * m) with a set we reduce it to O(n + m).

packages/features/ee/teams/components/TeamAssignList.tsx
joeauyeung, line 1: This branch is based on frontend work that @CarinaWolli did in preparation for the round robin handoff feature.

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,365 @@
...
+        },
+        sendTo: reassignedRRHost.email,
+        template: step.template,
+      });
+    }
+  }
+
+  // TODO create relationship between reminder and userId in order to determine which one to delete
+  // deleteScheduledEmailReminder(reminder.id, reminder.referenceId);
+};
```
joeauyeung, line 363: @CarinaWolli I was thinking about this. Is this something we have to do or am I over thinking it?

packages/trpc/server/routers/viewer/teams/roundRobin/roundRobinReassign.handler.ts
```diff
@@ -0,0 +1,19 @@
...
+    user: NonNullable<TrpcSessionUser>;
+  };
+  input: TRoundRobinReassignInputSchema;
+};
+
+export const roundRobinReassignHandler = async ({ input }: RoundRobinReassignOptions) => {
+  const { bookingId } = input;
+
+  await roundRobinReassignment({ bookingId });
+};
```
Udit-takkar, line 17: should we also add a check here to verify if logged in user can access this booking id or not?

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,365 @@
...
+    attendeePromises.push({
+      email: attendee.email,
+      name: attendee.name,
+      timeZone: attendee.timeZone,
+      language: { translate: tAttendee, locale: attendee.locale ?? "en" },
+    });
+  }
+
+  const attendeeList = await Promise.all(attendeePromises);
+
```
Udit-takkar, line 209: I don't think this code is utilizing Promise.all() correctly. we are pushing objects and not promises in the array.

Udit-takkar:
```
const attendeePromises = [];
for (const attendee of booking.attendees) {
  if (
    attendee.email === reassignedRRHost.email ||
    attendee.email === previousRRHost.email ||
    teamMembers.some((member) => member.email === attendee.email)
  ) {
    continue;
  }
  // Push the promise into the array
  const attendeePromise = getTranslation(attendee.locale ?? "en", "common").then(tAttendee => ({
    email: attendee.email,
    name: attendee.name,
    timeZone: attendee.timeZone,
    language: { translate: tAttendee, locale: attendee.locale ?? "en" },
  }));

  attendeePromises.push(attendeePromise);
}

// Await all promises
const attendeeList = await Promise.all(attendeePromises);
```

packages/features/ee/teams/components/TeamAssignList.tsx
```diff
@@ -0,0 +1,98 @@
...
+            <li key={host.user.id}>
+              <div className="flex justify-between">
+                <div className="flex w-full flex-col justify-between sm:flex-row">
+                  <label className="flex w-full p-0 py-2 first-letter:cursor-pointer hover:bg-gray-50">
+                    <div className="w-3" />
+                    <div className="flex w-full items-center">
+                      <div className="relative flex-none">
+                        <Avatar
+                          size="mdLg"
+                          imageSrc={`${WEBAPP_URL}/${host.user.username}/avatar.png`}
```
Udit-takkar, line 68: use avatarUrl

packages/lib/server/repository/booking.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export class BookingRepository {
+  static async getBookingAttendees(bookingId: number) {
+    return await prisma.attendee.findMany({
+      where: {
+        bookingId,
+      },
+    });
+  }
+}
```
Udit-takkar, line 11: do we really need this class? you have not defined any select field here

apps/web/components/dialog/ReassignDialog.tsx
```diff
@@ -0,0 +1,156 @@
...
+            assignedHosts={assignedHosts}
+          />
+        ) : (
+          <></>
+        )} */}
+        <DialogFooter>
+          <DialogClose />
+          <Button
+            data-testid="rejection-confirm"
+            onClick={() => {
```
Udit-takkar, line 142: NIT: display loader when mutation is loading

apps/web/components/booking/BookingListItem.tsx
```diff
@@ -323,6 +338,17 @@ function BookingListItem(booking: BookingItemProps) {
...
       />
+      <ReassignDialog
+        isOpenDialog={isOpenReassignDialog}
+        setIsOpenDialog={setIsOpenReassignDialog}
+        bookingId={booking.id}
+        teamId={booking.eventType?.team?.id || 0}
+        assignedHosts={
+          booking.user
+            ? [booking.user.email || ""].concat(booking.attendees.map((attendee) => attendee.email))
+            : []
```
Udit-takkar, line 349: I think instead of `booking.user.email` we should use `booking.userPrimaryEmail`

packages/lib/server/repository/booking.ts
joeauyeung: While it would be nice to have a full flushed out `BookingRepository` we can add methods to this one by one and slowly transition over. When getting resources we don't need to add select fields here as it's a small record.

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,364 @@
...
+  const reassignedRRHost = await getLuckyUser("MAXIMIZE_AVAILABILITY", {
+    availableUsers,
+    eventTypeId: eventTypeId,
+  });
+  const hasOrganizerChanged = booking.userId === previousRRHost.id;
+  const organizer = hasOrganizerChanged ? reassignedRRHost : booking.user;
+  const organizerT = await getTranslation(organizer?.locale || "en", "common");
+
+  if (hasOrganizerChanged) {
```
CarinaWolli, line 127: when the organizer changes we also need to fix the booking title

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,364 @@
...
+    {
+      dateFrom: dayjs(booking.startTime).format(),
+      dateTo: dayjs(booking.endTime).format(),
+      timeZone: eventType.timeZone || originalOrganizer.timeZone,
+    },
+    roundRobinReassignLogger
+  );
+
+  const reassignedRRHost = await getLuckyUser("MAXIMIZE_AVAILABILITY", {
+    availableUsers,
```
CarinaWolli, line 120: priority is missing in availableUsers

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,364 @@
...
+  const eventType = await getEventTypesFromDB(eventTypeId);
+
+  if (!eventType) {
+    console.error(`Event type ${eventTypeId} not found`);
+    throw new Error("Event type not found");
+  }
+
+  eventType.users = eventType.hosts.map((host) => ({ ...host.user, isFixed: host.isFixed }));
+
+  const roundRobinHosts = eventType.hosts.filter((host) => !host.isFixed);
```
CarinaWolli, line 76: We should probably also still support `eventType.users` as we do in `handleNewBooking` (Also in all places below where eventType.hosts ist used)

CarinaWolli (changes requested):
Great work so far 👏🏻 I found a few things that should be addressed before it can be merged

packages/features/ee/teams/components/TeamAssignList.tsx
```diff
@@ -0,0 +1,98 @@
...
+  user: {
+    id: number;
+    name: string | null;
+    email: string;
+    username: string | null;
+  };
+  isFixed: boolean;
+};
+
+export const TeamAssignList = ({
```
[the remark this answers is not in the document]
joeauyeung: I have a feeling it's going to over complicate things by keeping this code in. Let me remove everything that isn't being used.

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,371 @@
...
+      where: {
+        bookingUid: booking.uid,
+        method: WorkflowMethods.EMAIL,
+        workflowStep: {
+          action: WorkflowActions.EMAIL_HOST,
+          workflow: {
+            trigger: {
+              in: [
+                WorkflowTriggerEvents.BEFORE_EVENT,
+                WorkflowTriggerEvents.NEW_EVENT,
```
CarinaWolli, line 329: NEW_EVENT triggers shouldn't have reminders saved as they are sent immediately, but I think it would make sense to also send out workflow emails to the new host for the NEW_EVENT trigger

CarinaWolli (changes requested):
- [ ] The host details are mixed up for me. I have an event type with 2 round robin hosts: 
Screenshot 2024-07-11 at 3 41 16 PM

- [ ] Booking confirmation email for new host still has the old host the booking title
- [ ] Cancelled email has the new host in 'who', it should have the old one that receives the email 
- [ ] new calendar event has old host in booking title

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,517 @@
...
+                WorkflowTriggerEvents.AFTER_EVENT,
+              ],
+            },
+          },
+        },
+      },
+      include: {
+        workflowStep: {
+          include: {
+            workflow: true,
```
CarinaWolli, line 431: would `workflowSelect` from `getAllWorkflows.ts` work here instead of using `include`?

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,517 @@
...
+          },
+          sendTo: reassignedRRHost.email,
+          template: workflowStep.template,
+        });
+      }
+
+      await deleteScheduledEmailReminder(workflowReminder.id, workflowReminder.referenceId);
+    }
+    // Send new event workflows to new organizer
+    const newEventWorkflows = await prisma.workflow.findMany({
```
CarinaWolli, line 461: Can we maybe use `getAllWorkflowsFromEventType` or `getAllWorkflows` here? 

One thing I see missing in this query is the workflows that are 'active on all event types or teams' enabled.

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,517 @@
...
+          },
+          select: {
+            template: true,
+          },
+        },
+      },
+    });
+    for (const workflow of newEventWorkflows) {
+      for (let i = 0; i < workflow.steps.length; i++) {
+        await scheduleEmailReminder({
```
CarinaWolli, line 498: Can we use `scheduleWorkflowReminders` for this? `scheduleWorkflowReminders` also takes care of other cases for example to not send workflow notification for not yet confirmed events.

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,517 @@
...
+            eventType,
+          },
+          action: WorkflowActions.EMAIL_HOST,
+          triggerEvent: workflow.trigger,
+          timeSpan: {
+            time: workflow.time,
+            timeUnit: workflow.timeUnit,
+          },
+          sendTo: reassignedRRHost.email,
+          template: workflow.steps[i].template,
```
CarinaWolli, line 510: Comparing it with the when we call `scheduleEmailReminder` in `reminderScheduler.ts` it looks like we are missing some data here: 
- emailSubject
- emailBody
- sender
- workflowStepId
- hideBranding
- seatReferenceUid
- includeCalendarEvent

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,517 @@
...
+  user: {
+    include: {
+      destinationCalendar: true,
+    },
+  },
+  attendees: true,
+  references: true,
+};
+
+export const roundRobinReassignment = async ({ bookingId }: { bookingId: number }) => {
```
CarinaWolli, line 50: This function is really big and does a lot of things, I think we can improve that to make it better readable and easier to understand. This can be done later on though, I think we should focus on getting this merged now 🙏

packages/features/ee/round-robin/roundRobinReassignment.test.ts
```diff
@@ -0,0 +1,295 @@
...
+                timeZone: "Asia/Kolkata",
+              }),
+            ],
+          },
+        ],
+        organizer: originalHost,
+        usersApartFromOrganizer: users.slice(1),
+      })
+    );
+    const workflowReminders = await addWorkflowReminders([
```
CarinaWolli, line 150: what's that for?

packages/features/ee/round-robin/roundRobinReassignment.ts
joeauyeung: The `workflowSelect` nesting is `workflow -> steps`. This query is `step -> workflow`. I just selected the fields that would be relevant.

packages/features/ee/round-robin/roundRobinReassignment.ts
joeauyeung: What are your thoughts here. I only want workflows where the trigger is `NEW_EVENT` and only select the steps with the action `EMAIL_HOST`. I'm thinking if I got all the workflows I'd have to iterate through the workflows array and inside of them iterate over the steps array for O(m * n). Instead I would rather the database do that filtering.

I will add in the `OR` `{isActiveOnAll: true}`

packages/features/ee/round-robin/roundRobinReassignment.ts
joeauyeung: Resolving since I'm now using `scheduleWorkflowReminders`

packages/features/ee/round-robin/roundRobinReassignment.ts
CarinaWolli: Sorry missed that. Can we define all fields we need in a select instead of an include? [Never use Includes - use select](https://handbook.cal.com/engineering/best-practices/only-select-data-you-need)

packages/features/ee/round-robin/roundRobinReassignment.ts
```diff
@@ -0,0 +1,517 @@
...
+
+      await deleteScheduledEmailReminder(workflowReminder.id, workflowReminder.referenceId);
+    }
+    // Send new event workflows to new organizer
+    const newEventWorkflows = await prisma.workflow.findMany({
+      where: {
+        trigger: WorkflowTriggerEvents.NEW_EVENT,
+        OR: [
+          {
+            isActiveOnAll: true,
```
CarinaWolli, line 475: I think we need to add something like this: 

- For team workflows: 

```
OR: {
	isActiveOnAll: true,
	teamId: eventType.teamId
}
```

- For org workflows: 

```
OR: {
	isActiveOnAll: true,
	teamId: eventType.team.parent.id
}
```

packages/features/ee/round-robin/roundRobinReassignment.ts
CarinaWolli: Yes, you are right, we should only query the necessary workflows 👍

packages/features/ee/round-robin/roundRobinReassignment.ts
joeauyeung: Changed this one to `select`. The other `include` statements are needed.

CarinaWolli (approved):
Nice work! I tested everything including calendar events and emails and looks good now 🙏

PR #14326, opened 2024-04-04
fix: change identity provider on password reset

Udit-takkar:
## What does this PR do?

Fixes: When resetting password provider is not updated to CAL even though Password has been saved  resulting in this error

Screenshot 2024-04-04 at 1 45 46 PM

Expected Behaviour:- When user password reset is successfull then user should be able to login in with that password and identity provider should be changed to CAL (if it was CAL)

## Type of change


sean-brydon (dismissed):

apps/web/pages/api/auth/reset-password.ts
```diff
@@ -50,6 +51,20 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         emailVerified: new Date(),
       },
     });
+
+    // Change Identity Provider to CAL if it was previously Google
+    if (updatedUser.identityProvider === IdentityProvider.GOOGLE) {
```
hariombalhara, line 56: Why do we check for previous provider to be Google. Technically it can be anything right?

hariombalhara (changes requested):
Have a confusion that needs to be cleared

apps/web/pages/api/auth/reset-password.ts
Udit-takkar: We only have CAL, GOOGLE and SAML as Identity Provider

PeerRich: @sean-brydon whats your take on this?

hariombalhara: We could have more providers later. Because we want the identityProvider to be CAL if password reset happens, regardless of what it was earlier, I don't see a reason for this check.

hariombalhara: Also, I think that just requesting a password reset will change the identity provider to CAL. Shouldn't we do it only through the confirm password reset flow?

In the current approach, anyone can get someone else's identity provider to be changed.

Udit-takkar: > Also, I think that just requesting a password reset will change the identity provider to CAL. Shouldn't we do it only through the 'confirm password reset' flow?

> In the current approach, anyone can get someone else's identity provider to be changed.

no , the provider and password is only updated after entering new password from reset password link (in email). clicking on forgot password just send the email and doesn't update anything

hariombalhara: Cool

hariombalhara (approved):

PR #14328, opened 2024-04-04
chore: downgrade react query atoms

supalarry:
One of platform customers uses React 16 while react-query v5 requires at least React 18, so react-query is downgraded to v4.

Only difference is that queries do not return "isPending" but "isLoading", so refactored that in code.

Tested with examples app and it runs.


sean-brydon (commented):
I understand this works and im happy with this downgrade. Lets just make sure for context wise that we are aware isLoading !== isPending

https://github.com/TanStack/query/discussions/6297#discussioncomment-7467010

sean-brydon (changes requested):
Also just spotted type check failing in embed still 

Looks like there is a use of isPending

supalarry:
> Also just spotted type check failing in embed still
> 
> Looks like there is a use of isPending

Thanks for checking the PR! Type check now passes!

sean-brydon (approved):
LGTM! Thanks for making them changes - Lets keep an eye out to see if we see anything un expected with using isLoading > isPending

PR #14332, opened 2024-04-04
revert: chore downgrade react query atoms

supalarry:
Reverts calcom/cal.com#14328

The user tried atoms with downgraded react-query and issue still persisted so we can keep stuff how it was.


sean-brydon (approved):
LGTM

Issue #14318, opened 2024-04-03
[CAL-3355] purchaseTeamSubscription rename is causing API v1 builds to fail


zomars:
```
          Heads up the this function rename is causing API v1 builds to fail:
```

![image](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/d75080ec-57b9-4bde-a217-47ce27316b42/39ae0b4f-88d0-4246-b0cd-1fa3efb9736b?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2U4NmJmOTU3LWQ4MmYtNDY1ZS1iMjA1LTEzNTU1OWY0YjYyMy9kNzUwODBlYy01N2I5LTRiZGUtYTIxNy00N2NlMjczMTZiNDIvMzlhZTBiNGYtODhkMC00MjQ2LWIwY2QtMWZhM2VmYjk3MzZiIiwiaWF0IjoxNzEyMTgzMTIwLCJleHAiOjE3MTIyNjk1MjB9.0y_6-K3mimlMgCYKRFwRTmIk2KixAhBPu1fc_LdS7NQ)

![image](https://uploads.linear.app/e86bf957-d82f-465e-b205-135559f4b623/98972436-f763-4f80-bf67-dc436f385a7e/6921fca6-a05f-4e9e-ab03-a2cc7bb13d20?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2U4NmJmOTU3LWQ4MmYtNDY1ZS1iMjA1LTEzNTU1OWY0YjYyMy85ODk3MjQzNi1mNzYzLTRmODAtYmY2Ny1kYzQzNmYzODVhN2UvNjkyMWZjYTYtYTA1Zi00ZTllLWFiMDMtYTJjYzdiYjEzZDIwIiwiaWF0IjoxNzEyMTgzMzQ5LCJleHAiOjE3MTIyNjk3NDl9.nItHbBFdTgQKCnV3uzN8Fn3PT498vVV3Ln3Wxvv20w4)

\_Originally posted by @zomars in [https://github.com/calcom/cal.com/issues/13139#issuecomment-2035711704\_](https://github.com/calcom/cal.com/issues/13139#issuecomment-2035711704\_)

[...]

PR #14345, opened 2024-04-04
ci: adds api v1 builds to required checks

zomars:
## What does this PR do?

#14318 #14319

There were some refactors that broke API v1 builds. This aims to prevent that.


keithwillcode (approved):
Glorious. Thank you!

PR #14423, opened 2024-04-05
feat: always show intercom on desktop by default (hidden on mobile)

### LINKED ISSUE #14043: [CAL-3361] always show intercom on desktop by default (hidden on mobile)
we wanna start showing the intercom launcher on every page by default.

lets make it possible to toggle it on and off:

CleanShot 2024-04-04 at 11 49 50@2x

CleanShot 2024-04-04 at 11 48 42@2x

CleanShot 2024-03-09 at 10 13 51@2x

we only wanna hide it on mobile, to open it on mobile, you can still go through help:

CleanShot 2024-03-09 at 10 14 31@2x

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3361](https://linear.app/calcom/issue/CAL-3361/always-show-intercom-on-desktop-by-default-hidden-on-mobile)</sub>

ibex088:
## What does this PR do?

Fixes #14043, 
Fixes CAL-3361

https://www.loom.com/share/27a2fad8d12e43ca92d2b95293d93f62?sid=d4e3b466-409a-4aa7-aa0f-a06acd3a20ab


packages/prisma/schema.prisma
```diff
@@ -302,6 +302,7 @@ model User {
   movedToProfile             Profile?                     @relation("moved_to_profile", fields: [movedToProfileId], references: [id], onDelete: SetNull)
   secondaryEmails            SecondaryEmail[]
   isPlatformManaged          Boolean                      @default(false)
+  showSupport                Boolean                      @default(true)
```
PeerRich, line 305: we dont need to save this on a db level. we should do this either localStorage or as a cookie

PeerRich, line 305: we dont need to save this on a db level. we should do this either localStorage or as a cookie

PeerRich (dismissed):

packages/prisma/schema.prisma
SomayChauhan: saved in localStorage 👍

PeerRich (approved):
looks great! haven't tested

zomars:
Will this affect hosters that don't have intercom enabled?

PeerRich:
> Will this affect hosters that don't have intercom enabled?

uhm. it depends how it was implemented. i think if its wrapped in env && then no, @SomayChauhan any ideas?

the previous implementation worked without selfhosters having intercom

zomars (approved):
Just tested and useIntercom already handles if enabled for self hosters. My only concern is the sudden change of behavior if a self hoster has intercom enabled.

PR #14429, opened 2024-04-05
fix: filter and listMember handler

### LINKED ISSUE #14427: [CAL-3363] User filter on bookings page shows other users in same org even when org has been made private
### Issue Summary

If you create an org and set it to be private, your users in that org should not see other users in the filter

### Steps to Reproduce

1. Create an org
2. Add team members
3. Set the org to private
4. Log in as one of the members and go to filter upcoming/unconfirmed/etc bookings

### Actual Results

- You see all org members

### Expected Results

- You should only see yourself

<sub>[CAL-3363](https://linear.app/calcom/issue/CAL-3363/user-filter-on-bookings-page-shows-other-users-in-same-org-even-when)</sub>

sean-brydon:
fixes: #14427 

Hide people filter 
CleanShot 2024-04-05 at 14 05 14

Also remove members being returned form the org listMember page if its private and youre not an admin. 
This fixes the other place this could leak into the UI in the app OOO page. 
CleanShot 2024-04-05 at 14 06 29

### DIFF
--- a/packages/features/bookings/components/PeopleFilter.tsx
+++ b/packages/features/bookings/components/PeopleFilter.tsx
@@ -17,7 +17,7 @@ export const PeopleFilter = () => {
 
   const { data: currentOrg } = trpc.viewer.organizations.listCurrent.useQuery();
   const isAdmin = currentOrg?.user.role === "ADMIN" || currentOrg?.user.role === "OWNER";
-  const hasPermToView = currentOrg?.isPrivate ? isAdmin : true;
+  const hasPermToView = !currentOrg?.isPrivate || isAdmin;
 
   const { data: query, pushItemToKey, removeItemByKeyAndValue, removeAllQueryParams } = useFilterQuery();
   const [searchText, setSearchText] = useState("");
--- a/packages/features/bookings/components/PeopleFilter.tsx
+++ b/packages/features/bookings/components/PeopleFilter.tsx
@@ -15,6 +15,10 @@ export const PeopleFilter = () => {
   const { t } = useLocale();
   const orgBranding = useOrgBranding();
 
+  const { data: currentOrg } = trpc.viewer.organizations.listCurrent.useQuery();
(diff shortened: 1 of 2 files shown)


packages/trpc/server/routers/viewer/teams/listMembers.handler.ts
```diff
@@ -14,6 +14,13 @@ type ListMembersOptions = {
 
 export const listMembersHandler = async ({ ctx, input }: ListMembersOptions) => {
   const { prisma } = ctx;
+  const { isOrgAdmin } = ctx.user.organization;
+  const hasPermsToView = ctx.user.organization.isPrivate ? isOrgAdmin : true;
```
keithwillcode, line 18: Nit: Do we need the ternary?

```suggestion
  const hasPermsToView = !ctx.user.organization.isPrivate || isOrgAdmin;
```

keithwillcode (dismissed):

packages/trpc/server/routers/viewer/teams/listMembers.handler.ts
sean-brydon: Good spot

packages/features/bookings/components/PeopleFilter.tsx
```diff
@@ -15,6 +15,10 @@ export const PeopleFilter = () => {
   const { t } = useLocale();
   const orgBranding = useOrgBranding();
 
+  const { data: currentOrg } = trpc.viewer.organizations.listCurrent.useQuery();
+  const isAdmin = currentOrg?.user.role === "ADMIN" || currentOrg?.user.role === "OWNER";
+  const hasPermToView = currentOrg?.isPrivate ? isAdmin : true;
```
keithwillcode, line 20:
```suggestion
  const hasPermToView = !currentOrg?.isPrivate || isAdmin;
```

joeauyeung (approved):
Tested PR and people filter no longer renders for members in a private org. The filter still renders for admins or when the privacy setting is turned off.

image

PR #14450, opened 2024-04-07
feat: Enable CRM apps on a per event type basis

### LINKED ISSUE #14449: [CAL-3343] Transition CRM apps into event type apps
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3343](https://linear.app/calcom/issue/CAL-3343/transition-crm-apps-into-event-type-apps)</sub>

joeauyeung:
# This PR is stacked on #14272 

## What does this PR do?

This PR transitions CRM apps from being applied to every event type to now being able to choose which CRM apps are applied. This allows us to build upon our CRM apps in the future.

### **BREAKING** : This means users will have to go in and choose which event types they want their CRM apps to be enabled on. 

Fixes #14449 

## Requirement/Documentation

## Type of change

## How should this be tested?

- Do this for teams and individual credentials
- Install a CRM app from the app store
- Go to an event type
- Under the "Apps" tab you should see the CRM app. Enable it
- Create a booking
    - The CRM app action should trigger
- Disable the CRM app. 
	- The CRM app action should not trigger

## Mandatory Tasks

## Checklist


joeauyeung (commented):
Converting an app into an event type app required these steps:

- In the config/metadata add "extendsFeature: "EventType"
- Create an `/components/EventTypeAppCardInterface.tsx` under the app package
- Ensure that `appDataSchema` existed in `zod.ts` under the app package
	- `appDataSchema` should be extended from `eventTypeAppCardZod`
- The event type app card lives under `EventTypeAddonMap` in `packages/app-store/apps.browser.generated.tsx`

joeauyeung (commented):
Converting an app into an event type app required these steps:

- In the config/metadata add "extendsFeature: "EventType"
- Create an `/components/EventTypeAppCardInterface.tsx` under the app package
- Ensure that `appDataSchema` existed in `zod.ts` under the app package
	- `appDataSchema` should be extended from `eventTypeAppCardZod`
- The event type app card lives under `EventTypeAddonMap` in `packages/app-store/apps.browser.generated.tsx`

packages/app-store/_utils/useAddAppMutation.ts
```diff
@@ -113,3 +114,17 @@ function useAddAppMutation(_type: App["type"] | null, allOptions?: UseAddAppMuta
...
+  returnTo,
+}: {
+  stateStr: string;
+  teamId?: number;
+  returnTo?: string;
+}) => {
+  const teamIdParam = teamId ? `&teamId=${teamId}` : "";
+  const returnToParam = returnTo ? `&returnTo=${returnTo}` : "";
+  return `?state=${stateStr}${teamIdParam}${returnToParam}`;
+};
```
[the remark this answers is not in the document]
sean-brydon: NIT: If we're improving this lets set the search params the "proper" way with URL

const url = new URL('https://example.com');
url.searchParams.set('param1', 'value1');
url.searchParams.set('param2', 'value2');

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
```diff
@@ -73,5 +73,27 @@ export const getAllCredentials = async (
...
+
+  allCredentials = allCredentials.filter((credential) => {
+    if (!credential.type.includes("_crm") && !credential.type.includes("_other_calendar")) {
+      return credential;
+    }
+
+    if (crmCredentialIds.some((id) => id === credential.id)) {
+      return credential;
+    }
+  });
```
sean-brydon, line 96:
```suggestion
NIT: Havent tested but i think we can improve this on the number of iterations we do.  

allCredentials = allCredentials.filter(credential =>
  !credential.type.includes("_crm") &&
  !credential.type.includes("_other_calendar") ||
  crmCredentialIds.includes(credential.id)
);

```

sean-brydon (commented):
Will test it works soon but code looks good. 

I wonder how we want to address this breaking change as i'd hate for it to stop working for users that currently use CRM apps (Not sure the numbers here but probably a fair few).

packages/trpc/server/routers/loggedInViewer/deleteCredential.handler.ts
```diff
@@ -150,8 +154,7 @@ export const deleteCredentialHandler = async ({ ctx, input }: DeleteCredentialOp
               metadata: {
                 ...metadata,
                 apps: {
-                  ...metadata?.apps,
-                  [appSlug]: null,
```
joeauyeung: When uninstalling an event type app, by setting the app object to `null` it causes a zod error which means the event type page won't load. Instead we should be removing the app from the event type.

packages/app-store/_utils/writeAppDataToEventType.ts
joeauyeung, line 1: For backwards compatibility, we should write the CRM app metadata to the event types and set `enabled: false`.

packages/app-store/closecom/components/EventTypeAppCardInterface.tsx
```diff
@@ -10,6 +10,11 @@ const EventTypeAppCard: EventTypeAppCardComponent = function EventTypeAppCard({
 
   const { enabled, updateEnabled } = useIsAppEnabled(app);
 
+  // CRM backwards compatibility
+  if (enabled === undefined) {
+    updateEnabled(true);
+  }
```
joeauyeung, line 16: For CRM apps not in the event type metadata, we assume that they're credentials created before this PR hits production. By default they are enabled automatically. If a user goes to the event type page we should indicate that to the user. Also by taking this approach it causes a dirty field so if the user does hit save it'll create the app data in the metadata.

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
```diff
@@ -72,5 +67,45 @@ export const getAllCredentials = async (
     }
   }
 
+  // Only return CRM credentials that are enabled on the event type
+  const eventTypeAppMetadata = eventType?.metadata?.apps;
+
+  // Will be [credentialId]: { enabled: boolean }]
+  const eventTypeCrmCredentials = {};
```
joeauyeung, line 74: @sean-brydon regarding your comment regarding performance. Switching to an object here instead of an array.

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
```diff
@@ -72,5 +67,45 @@ export const getAllCredentials = async (
...
+      return credential;
+    }
+
+    // Backwards compatibility: All CRM apps are triggered for every event type. Unless disabled on the event type
+    // Check if the CRM app exists on the event type
+    if (eventTypeCrmCredentials[credential.id]) {
+      if (eventTypeCrmCredentials[credential.id].enabled) {
+        return credential;
+      }
+    } else {
```
joeauyeung, line 96: If the CRM credential exists in the event type metadata and is enabled, then return the credential

packages/features/bookings/lib/getAllCredentialsForUsersOnEvent/getAllCredentials.ts
```diff
@@ -72,5 +67,45 @@ export const getAllCredentials = async (
...
+      // If the CRM app doesn't exist on the event type metadata, check that the credential belongs to the user/team/org
+      if (
+        credential.userId === eventType.userId ||
+        credential.teamId === eventType.team?.id ||
+        credential.teamId === eventType.parentId
+      ) {
+        // If the CRM app doesn't exist on the event type metadata, assume it's an older CRM credential
+        return credential;
+      }
```
joeauyeung, line 105: If it doesn't exist in the metadata, and the credential fits the right scope for the event type then return the credential and it'll be triggered by default. This will happen for all credentials before this PR.

joeauyeung (commented):
For backwards compatibility, we should still trigger all CRM apps on every event type. This means no actions will be required by the user. 

All CRM credentials before this PR will not exist in the event type metadata. This means we should:
- Trigger the CRM if it does not exist in the metadata
- Add new CRM credentials to the metadata
- When users visit the event type apps page, trigger having the CRM app added to the event type metadata

CarinaWolli (approved):
Tested it with salesforce and everything worked. Comments are non-blocking, nice work 👏🏻

packages/app-store/closecom/pages/setup/index.tsx
```diff
@@ -63,13 +65,17 @@ export default function CloseComSetup() {
...
-                    method: "POST",
-                    body: JSON.stringify(values),
-                    headers: {
-                      "Content-Type": "application/json",
-                    },
-                  });
+                  const { returnTo } = query;
+                  const res = await fetch(
+                    `/api/integrations/closecom/add${returnTo ? `?returnTo=${returnTo}` : ""}`,
+                    {
```
zomars, line 71: NIT:

```suggestion
                  const url = new URL('/api/integrations/closecom/add');
                  if (returnTo) url.searchParams.append('returnTo', returnTo);
                  const res = await fetch(url.href, {
```

PeerRich: opened a new issue for the NIT: https://github.com/calcom/cal.com/issues/14804

PR #14475, opened 2024-04-09
fix: Remove exports in barrel file which are causing prisma import on client side

hariombalhara:
There are `className` imports on client side which is imported from @calcom/lib/index file which was importing `getEventTypesByViewer` file which imports repository which then imports prisma.

className -> @calcom/lib/index -> getEventTypesByViewer -> repository/event-type -> prisma


JeroenReumkens:
Had this issue in website repo, and can confirm this fix fixes that issue. 

CleanShot 2024-04-09 at 16 58 13@2x

keithwillcode (approved):

PR #14476, opened 2024-04-09
fix: always load troubleshooter without  SSR

### LINKED ISSUE #14477: [CAL-3376] Prevent hydration errors in troubleshooter
Troubleshooter should not be loaded with ssr and should be a client only component

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3376](https://linear.app/calcom/issue/CAL-3376/prevent-hydration-errors-in-troubleshooter)</sub>

sean-brydon:
This should be a client only component 

Lets prevent hydration errors in dev 

Fixes: #14477


joeauyeung (approved):
Code LGTM.

@sean-brydon is there a specific way we should test this? I'm not getting dehydration errors when running locally. I'll approve for now since it's working as expected and doesn't introduce any new bugs.

sean-brydon:
> Code LGTM.
> 
> @sean-brydon is there a specific way we should test this? I'm not getting dehydration errors when running locally. I'll approve for now since it's working as expected and doesn't introduce any new bugs.

Youll get hydration errors locally when(you/server are) in a different timezone to what the schedule is loaded in. Easier to make it client only and go from there

PR #14478, opened 2024-04-09
fix: Ensure the session is invalidated during booking-seats test

emrysal:
## What does this PR do?

Fixes test fail on localhost due to the session user not being immediately invalidated, therefore still seeing isHost = true and seeing all attendee information.


keithwillcode (approved):

sean-brydon (approved):
Nice find! I'd have thought we redirect from auth/logout but i guess it makes sense

PR #13621, opened 2024-02-09
feat: ooo-v2

alannnc:
## What does this PR do?

- Removes old feature that triggered a profile forwarding.
- New feature allows to view current user calendar with out of office days distinction and a Click to Action if forwarded.
- Revamps Out of office from settings
- Adds reasons and a emoji to out of office entries
- New reasons table with seeded for default options

Fixes #13558

 Loom Video: https://www.loom.com/share/2c906d2c73ba4e4aabee4cc77b6b3d23?sid=5327dc00-c8c2-40ae-98e6-ac96f8e4c9fc

## Requirement/Documentation

- If there is a requirement document, please, share it here.
- If there is ab UI/UX design document, please, share it here.

## Type of change

- Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- Are there environment variables that should be set? None
- What are the minimal test data to have? 2 users in a team with events
- What is expected (happy path) to have (input and output)? One user can mark days OOO with/without forward to team member.
- Any other important info that could help to test that PR

## Mandatory Tasks

- [X] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

PR #14486, opened 2024-04-09
chore: Adds deprecated-imports-next-router

zomars:
## What does this PR do?

Follow up for #13621 

Since we're aiming to migrate to app router. It's not very explicit that we should avoid using `next-router`. This eslint rule aims to make it pretty clear.

image

## Type of change

## How should this be tested?

- yarn lint

## Mandatory Tasks


zomars (commented):
Self review done

PeerRich (approved):
pretty solid rule

PR #14495, opened 2024-04-10
feat: SDK managed users + PLA fixes

exception:
## What does this PR do?

- Uses proper guard for all managed user endpoints, rather than access tokens as this was replaced by the `/me` endpoint. 
- Adds (partial) support for managed user endpoints within the SDK


keithwillcode (approved):
Code looks good. Have not tested

PR #14497, opened 2024-04-10
fix: Broken SSR for event booking pages

hariombalhara:
## What does this PR do?

Fixes broken OG images.

Happened because React Query is unable to reuse already sent `/event` request's response and because SSR doesn’t wait for re-renders, it gets `event` as `undefined` causing the issue of OG image generation(which is completely SSR)

It happened because of newly introduced param `orgRedirection` which is by default assumed `false` but for React Query `orgRedirection=undefined `and `orgRedirection=false` are two different queries.

Wrote a failing test now that’s passing in the PR

## Type of change

## How should this be tested?
- Simply see the source code for pro/30min and notice that title is wrong in it.
- It has og:image URL also wrong

## Mandatory Tasks

### DIFF
--- a/apps/web/playwright/booking-pages.e2e.ts
+++ b/apps/web/playwright/booking-pages.e2e.ts
@@ -1,4 +1,5 @@
 import { expect } from "@playwright/test";
+import { JSDOM } from "jsdom";
 
 import { randomString } from "@calcom/lib/random";
 import { SchedulingType } from "@calcom/prisma/client";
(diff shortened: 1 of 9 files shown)


hariombalhara:
* **#14497** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/14497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/web/playwright/booking-pages.e2e.ts
```diff
@@ -22,6 +23,33 @@ test.afterEach(async ({ users }) => {
...
+  const [response] = await Promise.all([
+    // This promise resolves to the main resource response
+    page.waitForResponse(
+      (response) => response.url().includes(`/${user.username}/30-min`) && response.status() === 200
+    ),
+
+    // Trigger the page navigation
+    page.goto(`/${user.username}/30-min`),
+  ]);
+  const ssrResponse = await response.text();
```
hariombalhara, line 40: Important to verify on raw text as on client it get's fixed as soon as react hydration kicks in.

packages/features/eventtypes/lib/getPublicEvent.ts
```diff
@@ -291,6 +292,7 @@ export const getPublicEvent = async (
     profile: getProfileFromEvent(eventWithUserProfiles),
     users,
     entity: {
+      fromRedirectOfNonOrgLink,
```
hariombalhara, line 295: Send it back so that react query can sent it again and cached response can be served by react query.

zomars (dismissed):
Awesome find! Nice work @hariombalhara 🙏

packages/trpc/server/routers/viewer/organizations/createTeams.handler.ts
```diff
@@ -234,7 +239,7 @@ async function moveTeam({
   await addTeamRedirect({
     oldTeamSlug: team.slug,
     teamSlug: newSlug,
-    orgSlug: org.slug,
+    orgSlug: org.slug || (orgMetadata?.requestedSlug ?? null),
```
hariombalhara, line 242: Found this while doing test that in case payment is required for the organization(which is true for cal.com), org.slug won't be set and thus we need to fallback to requestedSlug

emrysal (approved):
LGTM

PR #14509, opened 2024-04-10
fix: platform booking endpoints should be public

ThyMinimalDev:
## What does this PR do?

create / cancel / reschedule / get booking info by UID should be public on platform for customer to build their own public pages


apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -118,12 +116,11 @@ export class BookingsController {
   }
 
   @Post("/")
-  @Permissions([BOOKING_WRITE])
   async createBooking(
     @Req() req: Request & { userId?: number },
     @Body() _: CreateBookingInput
   ): Promise<ApiResponse<unknown>> {
-    req.userId = await this.getOwnerId(req);
+    req.userId = (await this.getOwnerId(req)) ?? -1;
```
ThyMinimalDev, line 123: adding -1 to mimic current behaviour of app.cal.com

supalarry (approved):

PR #14525, opened 2024-04-11
feat: atoms team booking

supalarry:
Use dynamic bookings logic to simulate a team event using atoms booker.

For PlatformBookerWrapper to act as a team booking we need:
1. username array containing usernames to book the booking with
2. entity containing orgSlug
3. duration of the event
```
        <Booker
            eventSlug={eventTypeSlug}
            username={[
              "npwruvxgpl-clujvibu30001khltnifvcca8-example",
              "dpgobfhtli-Clujvibu30001khltnifvcca8-example",
            ]}
            onCreateBookingSuccess={(data) => {
              setBookingTitle(data.data.title);
              router.push(`/${data.data.uid}`);
            }}
            duration={eventTypeDuration}
            entity={{
              orgSlug: "ecorp",
              considerUnpublished: false,
            }}
          />
```

## ❗ Forced slug - important to enable feature
When booking atom makes a booking, it calls v2 bookings.controller.ts which calls handleNewBooking.ts which calls loadUsers, which calls orgDomainConfig which returns orgSlug and if it's valid. The orgSlug is based on organization subdomain e.g. acme.cal.com, and is created so that someone booking acme organization does not book another organization.

Platform teams do not have subdomains, so we can enforce orgSlug that is necessary for handleNewBooking.ts to work by:
1. Providing `x-cal-client-id` request header to note that request is coming from platform.
2. Providing `x-cal-force-slug` request header as a way to pass the enforced orgSlug, which is passed to Booker atom via "entity.orgSlug prop".

Implemented in [fix: force platform orgSlug in handleNewBooking.ts](https://github.com/calcom/cal.com/pull/14525/commits/c1e5bda0b8c23ef07c1f4b7356b8fc119226f367) and confirmed with Hariom that approach of client ID + enforced slug as headers is fine.


packages/platform/atoms/hooks/usePublicEvent.tsx
```diff
@@ -30,6 +37,12 @@ export const usePublicEvent = (props: { username: string; eventSlug: string }) =
         })
         .then((res) => {
           if (res.data.status === SUCCESS_STATUS) {
+            if (props.duration && res.data.data) {
+              // note(Lauris): In case of "dynamic" event type default event duration returned by the API is 30,
```
sean-brydon, line 41: This caught my eye - dynamic event also changes a few things like the location (defaults to the first username in the lists "default conferencing" type)

Not sure if we need to take that into account here also

supalarry: Good note, but for now it looks like it should be fine as is.

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
```diff
@@ -149,6 +149,9 @@ export class BookingsController {
   ): Promise<ApiResponse<unknown>> {
     const oAuthClientId = clientId?.toString();
     const locationUrl = body.locationUrl;
+
+    const { orgSlug } = req.body;
```
ThyMinimalDev, line 153:
```
const {orgSlug, locationUrl } = body
```

packages/features/bookings/components/event-meta/Duration.tsx
```diff
@@ -52,7 +54,7 @@ export const EventDuration = ({ event }: { event: PublicEvent }) => {
       setSelectedDuration(event.length);
   }, [selectedDuration, setSelectedDuration, event.metadata?.multipleDuration, event.length, isDynamicEvent]);
 
-  if (!event?.metadata?.multipleDuration && !isDynamicEvent)
+  if ((!event?.metadata?.multipleDuration && !isDynamicEvent) || isPlatform)
     return <>{getDurationFormatted(event.length, t)}</>;
```
ThyMinimalDev, line 58: this makes it so that platform booker can never display multiple durations right?

supalarry: yes

apps/api/v2/src/ee/bookings/controllers/bookings.controller.ts
supalarry: Implemented in [refactor: destructure properies from body instead of req.body](https://github.com/calcom/cal.com/pull/14525/commits/485c5218825af831da82bf76a035e22c443692f3)

packages/platform/examples/base/src/pages/booking.tsx
```diff
@@ -54,6 +59,11 @@ export default function Bookings(props: { calUsername: string; calEmail: string
               setBookingTitle(data.data.title);
               router.push(`/${data.data.uid}`);
             }}
+            duration={eventTypeDuration}
+            entity={{
+              orgSlug: "ecorp",
+              considerUnpublished: false,
+            }}
```
ThyMinimalDev, line 66: I am wondering what are the potential risk of enabling users to control their orgSlug

ThyMinimalDev (approved):

PR #14528, opened 2024-04-11
fix: Invalid 403 forbidden for event-types GET

keithwillcode:
## What does this PR do?

403 Forbidden was being thrown for GET requests even when a user in a team with the event type was making the request. The reason is because we had an auth middleware running before the GET endpoints. The auth middleware was too strict.

I removed the auth middleware since the GET, DELETE and PATCH endpoints all handle their own permissions checks. The extra middleware was 1) creating a bug and 2) making an extra call to the database for every single request made to these endpoints.

## Type of change

## How should this be tested?

- Ensure you can call the GET endpoint for an event-type using an API key of a user within the team that has the event type.


keithwillcode:
This one is urgent to fix so automated tests might need to come in a follow-up PR but I will start work on them now.

emrysal (dismissed):
Nice

apps/api/v1/pages/api/event-types/_utils/checkTeamEventEditPermission.ts
```diff
@@ -26,7 +26,7 @@ export default async function checkTeamEventEditPermission(
 
     if (!membership?.role || !["ADMIN", "OWNER"].includes(membership.role)) {
       throw new HttpError({
-        statusCode: 401,
```
keithwillcode, line 29: 401 was the wrong HTTP status code to use here because the user is authenticated at this point given the API key passed in but they are forbidden from performing this action.

keithwillcode:
Running into blockers writing unit tests for this endpoint because our test suite currently doesn't support resolving the ~/lib/{...} imports we use all over the API. Will be investigating this and putting tests in a follow up PR.

keithwillcode:
WIP for unit tests here: https://github.com/calcom/cal.com/pull/14536

zomars (dismissed):
Fair enough. LGTM

exception (approved):

PR #14536, opened 2024-04-11
chore: Add unit tests event types api

keithwillcode:
Adds unit tests to the GET handler for event-types/[id] in v1 of the API.

This is a WIP because we need to figure out resolving ~/ in vitest. - EDIT @zomars Fixed this

### DIFF
--- a/vitest.workspace.ts
+++ b/vitest.workspace.ts
@@ -41,10 +41,31 @@ const workspaces = packagedEmbedTestsOnly
             "packages/embeds/**/*",
             "packages/lib/hooks/**/*",
             "packages/platform/**/*",
+            "apps/api/v1/**/*",
             "apps/api/v2/**/*",
           ],
           name: "@calcom/core",
           setupFiles: ["setupVitest.ts"],
+        },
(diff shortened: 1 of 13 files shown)


vitest.workspace.ts
```diff
@@ -41,12 +41,33 @@ const workspaces = packagedEmbedTestsOnly
...
+            "packages/lib/hooks/**/*",
+            "packages/platform/**/*",
+            "apps/api/v2/**/*",
+          ],
+          name: "@calcom/api",
+          setupFiles: ["setupVitest.ts"],
+        },
+        resolve: {
+          alias: {
+            "~": new URL("./apps/api/v1", import.meta.url).pathname,
```
zomars, line 67: This might be it @keithwillcode

zomars (approved):

vitest.workspace.ts
keithwillcode: Beautiful. Thank you, @zomars!

zomars:
Merged into #14528

PR #14542, opened 2024-04-12
feat: update platform oauth client form

ThyMinimalDev:
## What does this PR do?

Enable edit of oauth client for platform


Ryukemeister (approved):

emrysal:
Also looks good to me, but wondering if the i18n of this file is OK - looks like it misses a lot of translations

PR #14549, opened 2024-04-12
fix: Remove hardcoded stripe..

### LINKED ISSUE #14482: [CAL-3377] Issue with stripe meta data corrupts many different endpoints
\[ { "code": "invalid_type", "expected": "string", "received": "undefined", "path": \[ "apps", "stripe", "currency" \], "message": "Required" } \]

Go to Sentry to see all the places it's affecting

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3377](https://linear.app/calcom/issue/CAL-3377/issue-with-stripe-meta-data-corrupts-many-different-endpoints)</sub>

emrysal:
## What does this PR do?

Fixes whenever any payment app credential was deleted, if there was an installed payment app of any type on an event type, it would add stripe incorrectly to the event type and brick the user.

Fixes: #14482
Fixes CAL-3377

This created invalid data in our database, like:
image

Instead, the right behaviour should be to remove the deletedCredential from the eventType(s) in it's entirety, no reason to keep historical data. For now we keep the `hidden` behaviour; however this should be reevaluated @PeerRich  

Needs (but it's quite urgent):

* Tests for payment app disconnect / reconnecting / deletion
* Migrations to rectify broken apps (perhaps do it without migrations, instead fixing our data).


sean-brydon (approved):
Well... That was fun <3

PR #14555, opened 2024-04-12
chore: adds maintenance mode for API v1

zomars:
## What does this PR do?

As per @keithwillcode request for upcoming upgrade.

Adds a simple maintenance mode for preventing unwanted traffic during migrations/upgrades.


apps/api/v1/middleware.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+};
+
+export const config = { matcher: "/:path*" };
+
+export async function middleware() {
+  const isInMaintenanceMode = await safeGet<boolean>("isInMaintenanceMode");
+  if (!isInMaintenanceMode) return NextResponse.next();
+  // If is in maintenance mode, return a 503 status code
+  return NextResponse.json({ message: "API is currently in maintenance. Please try again at a later time." });
```
keithwillcode, line 18: We need to set the status code here, no?

zomars: Correct. Pushing change.

keithwillcode (approved):

PR #14560, opened 2024-04-12
feat: Platform Rate limiting + access token caching

exception:
## What does this PR do?

- Introduces rate limiting via the `@Throttle` decorator, can be customized on a handler-by-handler manner.
- Adds a caching layer to access token strategy, leveraging Redis `HMSET` / `EXPIREAT`.

To test you need to set the `REDIS_URL` in ENV.

**BEFORE MERGING WE NEED TO PROVISION A REDIS INSTANCE IN FLIGHTCONTROL / AWS.**


apps/api/v2/src/ee/schedules/controllers/schedules.controller.ts
```diff
@@ -70,6 +71,7 @@ export class SchedulesController {
 
   @Get("/:scheduleId")
   @Permissions([SCHEDULE_READ])
+  @Throttle({ default: { limit: 3, ttl: 60000 } }) // allow 3 requests per minute
```
keithwillcode, line 74: 3 per minute seems really low. Why this value?

apps/api/v2/src/modules/auth/strategies/access-token/access-token.strategy.ts
```diff
@@ -27,7 +27,10 @@ export class AccessTokenStrategy extends PassportStrategy(BaseStrategy, "access-
         throw new UnauthorizedException(INVALID_ACCESS_TOKEN);
       }
 
-      await this.oauthFlowService.validateAccessToken(accessToken);
+      const accessTokenValid = await this.oauthFlowService.validateAccessToken(accessToken);
+      if (!accessTokenValid) {
+        throw new UnauthorizedException("Access Token may no longer be valid.");
```
keithwillcode, line 32: Should this be “is no longer valid”?

ThyMinimalDev (approved):

PR #14572, opened 2024-04-13
perf: Improve event-types query

keithwillcode:
This query runs on average in 376.81ms very frequently, consuming a lot of the DB resources. Rewriting it this way with the union brings it down to 190ms on average.


keithwillcode:
Going to close this as this query now runs on average in 45ms after our db upgrades.

PR #14583, opened 2024-04-15
fix: removing stripe results in the event-types page loading indefintely

ibex088:
fixes [LOOM](https://www.loom.com/share/60aba6ae04ca4cc8a70173efaec32afb?sid=423cd98a-3b65-40c5-9432-19df02eda02b)


emrysal (approved):
Nice one 👍 - surprised undefined works (not entirely unsurprised null doesn't)

PR #14594, opened 2024-04-15
fix: email not sending with : in from name

CarinaWolli:
## What does this PR do?

When Full Name has several : sending the the booking confirmation (and any other bookings emails) to the attendee failed. It was throwing: `The from address does not match a verified Sender Identity`

Fixes https://linear.app/calcom/issue/CAL-3339/booker-doesnt-get-confirmation-email


emrysal (approved):
Nice find! 👍

PR #14600, opened 2024-04-16
fix: custom attributes not being synced in intercom

ibex088:
Fixes: 

a lot of users have undefined data in their custom data in intercom
they basically contact us, but our custom data is not being shared so there is no way to identify users

CleanShot 2024-04-15 at 20 59 50@2x

### DIFF
--- a/apps/web/server/lib/ssr.ts
+++ b/apps/web/server/lib/ssr.ts
@@ -81,6 +81,7 @@ export async function ssrInit(context: GetServerSidePropsContext, options?: { no
     // Provides a better UX to the users who have already upgraded.
     ssr.viewer.teams.hasTeamPlan.prefetch(),
     ssr.viewer.public.session.prefetch(),
+    ssr.viewer.loggedInRouter.me.prefetch(),
   ]);
 
   return ssr;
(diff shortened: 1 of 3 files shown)


apps/web/server/lib/ssr.ts
```diff
@@ -81,6 +81,7 @@ export async function ssrInit(context: GetServerSidePropsContext, options?: { no
     // Provides a better UX to the users who have already upgraded.
     ssr.viewer.teams.hasTeamPlan.prefetch(),
     ssr.viewer.public.session.prefetch(),
+    ssr.viewer.loggedInRouter.me.prefetch(),
```
SomayChauhan, line 84: not sure if adding this is a good idea or not!!

packages/features/shell/Shell.tsx
```diff
@@ -215,15 +215,17 @@ const useBanners = () => {
 const Layout = (props: LayoutProps) => {
   const banners = useBanners();
 
-  const showIntercom = localStorage.getItem("showIntercom");
+  const { data: user } = trpc.viewer.me.useQuery();
```
Udit-takkar, line 218: can you not user useSession here?

SomayChauhan: the main reason to use this here is because in useIntercom this api is called and the data is not available initially when boot() is called [LINK](https://github.com/SomayChauhan/cal.com/blob/33b9e8f48c8477a274d946123cf661e30143c57b/packages/features/ee/support/lib/intercom/useIntercom.ts#L26)

PeerRich (approved):

PR #14602, opened 2024-04-16
refactor: remove /ee/ url segment from atoms hooks

supalarry:
TODO: add redirects


keithwillcode (approved):
@supalarry nice. @ThyMinimalDev already added the rewrites so we should be good

PR #14604, opened 2024-04-16
fix: api v2 - platform users can disable emails and display their links  

ThyMinimalDev:
## What does this PR do?

fix an issue preventing platform users from disabling emails
 we now only display links if they are set in the oauth client of user
 we now display booking success page link instead of cancel / reschedule in the emails in case these are not set


packages/emails/src/components/ManageLink.tsx
```diff
@@ -1,13 +1,19 @@
...
+  const rescheduleLink = getRescheduleLink(props.calEvent);
+  const bookingLink = getBookingUrl(props.calEvent);
   if (
-    props.attendee.email === props.calEvent.attendees[0]?.email ||
-    props.calEvent.organizer.email === props.attendee.email
+    (props.attendee.email === props.calEvent.attendees[0]?.email ||
+      props.calEvent.organizer.email === props.attendee.email) &&
+    (Boolean(cancelLink) ||
+      (!props.calEvent.recurringEvent && Boolean(rescheduleLink)) ||
+      Boolean(bookingLink))
```
supalarry, line 16: We have `!props.calEvent.recurringEvent && Boolean(rescheduleLink))` and then the same on line 37. We could have boolean "showReschedule" and "showCancel" and be used to check if both are false return null and then return rest of actualy content without nesting.

Also, i wonder - i see that we ask `!props.calEvent.recurringEvent && Boolean(rescheduleLink)) ||
      Boolean(bookingLink)` - if rescheduleLink is not given, but bookingLink is, then we still would not show reschedule because on line 37 it does not include `||
      Boolean(bookingLink)`

apps/web/pages/settings/organizations/platform/oauth-clients/create.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { useCompatSearchParams } from "@calcom/lib/hooks/useCompatSearchParams";
 
 import PageWrapper from "@components/PageWrapper";
 import { OAuthClientForm } from "@components/settings/organizations/platform/oauth-clients/OAuthClientForm";
 
 export const CreateOAuthClient = () => {
-  const router = useRouter();
-  const clientId = router.query.clientId as string | undefined;
+  const searchParams = useCompatSearchParams();
+  const clientId = searchParams?.get("clientId") || "";
```
supalarry, line 11: Question: when is the "clientId" set? I see below is OAuthClientForm and its create.ts file used to create oauth client, so when do we need clientId if we are creating a new client?

packages/emails/src/components/ManageLink.tsx
supalarry: And also abstract `props.calEvent.organizer.email === props.attendee.email` into `const isOrganizer` and same for `const isBooker` - i assume attendees[0] represents person who booked.

packages/lib/CalEventParser.ts
```diff
@@ -154,6 +154,17 @@ const getSeatReferenceId = (calEvent: CalendarEvent): string => {
   return calEvent.attendeeSeatId ? calEvent.attendeeSeatId : "";
 };
 
+export const getBookingUrl = (calEvent: CalendarEvent) => {
+  if (calEvent.platformClientId) {
+    if (!calEvent.platformBookingUrl) return "";
```
supalarry, line 159: Question: don't know in details, but when would line 159 be true?

supalarry (dismissed):
Looks good just got some questions and little refactor suggestion

packages/emails/src/components/ManageLink.tsx
ThyMinimalDev: true, I will rewrite this in a cleaner way

apps/web/pages/settings/organizations/platform/oauth-clients/create.tsx
ThyMinimalDev: clientId is set when clicking on the oauuth client card edit button, we currently reuse the same create form to update

supalarry (dismissed):

supalarry (approved):

PR #14607, opened 2024-04-16
fix: On updating event type with private url, private url stops working

### LINKED ISSUE #14573: [CAL-3469] On updating event type with private url, private url stops working.
### Issue Summary

Private url stops working if event is updated after its creation.

### Steps to Reproduce

1. Create event
2. Enable private url and save.
3. Go to the private url and it works.
4. Make some other change and save.
5. Now the private url stops working, also shows hashedLink as undefined.

- ### Evidence

https://github.com/calcom/cal.com/assets/74371312/f0c9e570-4116-4fd4-8125-4c1a40f180ec

<sub>[CAL-3469](https://linear.app/calcom/issue/CAL-3469/on-updating-event-type-with-private-url-private-url-stops-working)</sub>

Amit91848:
## What does this PR do?

Fixes #14573 
Fixes CAL-3369

Since we are only sending dirty fields to update, on every request the hashedLink is undefined and the connection gets deleted.

https://github.com/calcom/cal.com/blob/e6056bc4a4db8abdf9c52441e75d905a4fff5edd/packages/trpc/server/routers/viewer/eventTypes/update.handler.ts#L305-L343

https://github.com/calcom/cal.com/assets/74371312/179951ab-e2f1-4e49-9c12-f91c4ce21685

## Type of change

## Mandatory Tasks


Udit-takkar (changes requested):
@Amit91848 Thanks for your contribution but someone from the team is already working on the fix with tests

PR #14610, opened 2024-04-16
fix: N Calendar Invites For Collective Events

### LINKED ISSUE #14580: [CAL-3412] a collective event of N people sends N invites
apparently the person booking receives N emails and N calendar invites

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3412](https://linear.app/calcom/issue/CAL-3412/a-collective-event-of-n-people-sends-n-invites)</sub>

joeauyeung:
## What does this PR do?

This PR aims to prevent attendees receiving multiple calendar invites by only creating one calendar event.

Fixes #14580 Fixes CAL-3412 Fixes CAL-3468

https://www.loom.com/share/89fecb968c9a4511869bfa807c273ca4

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a collective team event with each team member having Outlook as a destination calendar
- Create a booking under the collective event
- The attendee should receive one email from Outlook and all team members should be invited to the event

## Mandatory Tasks

- [] Make sure you have self-reviewed the code. A decent size PR without self-review might be rejected.

## Checklist


packages/app-store/_utils/calendars/processExternalId.ts
joeauyeung, line 1: For Outlook since the calendarId is for internal purposes we can only invite the primary email to calendar events. https://www.loom.com/share/60ac4ad245414714979dc2e11d8d347a?sid=6b089642-a97d-40cd-a015-37fc97a3e8d9

packages/core/EventManager.ts
```diff
@@ -535,78 +535,104 @@ export default class EventManager {
...
+              log.warn(
+                "No credentialId found for destination calendar, falling back to first found calendar of same type as destination calendar",
+                safeStringify({
+                  destination: getPiiFreeDestinationCalendar(destinationCalendar),
+                  firstConnectedCalendar: getPiiFreeCredential(firstCalendarCredential),
+                })
+              );
+              createdEvents.push(await createEvent(firstCalendarCredential, event));
+              eventCreated = true;
+              break;
```
joeauyeung, line 595: Not a huge fan of this approach but keeping this in here to maintain what we currently have.

keithwillcode:
@joeauyeung I just tested created an OAuthClient on the preview development and I can't click the Submit button at all

keithwillcode (changes requested):
@joeauyeung I just tested creating an OAuthClient on the preview development and I can't click the Submit button at all

keithwillcode:
> @joeauyeung I just tested creating an OAuthClient on the preview development and I can't click the Submit button at all

Ok. Tested again. Not sure what was going on but now I can submit successfully.

packages/app-store/_utils/calendars/processExternalId.ts
```diff
@@ -0,0 +1,18 @@
+import type { DestinationCalendar } from "@prisma/client";
+
+import { metadata as OutlookMetadata } from "../../office365calendar";
+
+/**
+ * When inviting attendees to a calendar event, sometimes the external ID is only used for internal purposes
+ * Need to process the correct external ID for the calendar service
+ */
+const processExternalId = (destinationCalendar: DestinationCalendar) => {
```
keithwillcode, line 9: Do you think this function has the potential to grow in the number of "if" statements moving forward? i.e. We'll have to have different logic for a large number of different calendar apps. If so, we should move it off to its own file per calendar integration.

Since this is urgent to get fixed, that could be a follow-up refactor.

keithwillcode (commented):
It's concerning to me that all this logic is changing and no automated tests are updated/added. We need to follow up ASAP with a PR adding unit tests to verify all of this logic.

joeauyeung:
>It's concerning to me that all this logic is changing and no automated tests are updated/added. We need to follow up ASAP with a PR adding unit tests to verify all of this logic.

@keithwillcode there are 0 tests at all for the `EventManager` class so I'm writing tests out but I'll follow up in another PR so we're not waiting for this one. 

Regarding `processExternalId` AFAIK it's just Outlook that treats calendarIds like this. I see what you're getting at though. If there are other calendars that are added that have their own custom logic we should refactor processing the externalIds under their respective app package.

packages/core/EventManager.ts
```diff
@@ -535,78 +535,105 @@ export default class EventManager {
     };
 
     if (event.destinationCalendar && event.destinationCalendar.length > 0) {
-      // Since GCal pushes events to multiple calendars we only want to create one event per booking
-      let gCalAdded = false;
-      const destinationCalendars: DestinationCalendar[] = event.destinationCalendar.reduce(
```
zomars, line 540: Is the reduce refactor necessary to fix the bug? Or was it done for getting two birds with one stone?

If it's the latter then I'm afraid it is making the PR harder to review since it's not clear "where" is the bug that we're fixing.

Also as @keithwillcode mentioned we don't have EventManager tests to ship this PR with 100% confidence. RN if I approve this I'm not 100% confident this will not break other event type configurations.

My suggestion would be:

Make a PR with "only" the changes needed to fix the current bug. Then other PR for adding tests. Then other PR to do chores and refactoring.

zomars (changes requested):

packages/core/EventManager.ts
joeauyeung: The refactor is necessary since ideally only one calendar event is created for the booking. When inviting team members it should add events to their default destination calendars by inviting those calendars to the event.

Since this is affecting collective events, a PR with "only" the fix would be to remove pushing multiple destination calendars when the event is a collective one. That would be the safest bet. Looking through the `CalendarService` files of GCal and Outlook we already add team members as the attendees. Although for Outlook we still need to process the `externalId` properly so all team members are added to the event.

keithwillcode: Yeah I think we should focus the rest of day on getting tests wired up here and we ship this tomorrow. 

This is the type of change that is high risk of making things worse than the original bug we are trying to fix.

packages/app-store/_utils/calendars/processExternalId.ts
keithwillcode: The tests should written against main and not against this code first to verify all existing logic stays the same. Once verified, we then add tests to verify this change in logic

packages/core/EventManager.ts
keithwillcode: The tests should written against main and not against this code first to verify all existing logic stays the same. Once verified, we then add tests to verify this change in logic

zomars:
Closing in favor of https://github.com/calcom/cal.com/pull/14617

PR #14612, opened 2024-04-16
fix: booker atoms using next/router

ThyMinimalDev:
## What does this PR do?

For now not displaying OOOSlots for platform, we need to refactor to not use next/router in this component when we want to enable it for platform atoms


packages/platform/examples/base/src/pages/[bookingUid].tsx
```diff
@@ -2,7 +2,8 @@ import { Navbar } from "@/components/Navbar";
 import { CheckCircle2Icon } from "lucide-react";
 import { X } from "lucide-react";
 import { Inter } from "next/font/google";
-import { useRouter } from "next/navigation";
+// eslint-disable-next-line @calcom/eslint/deprecated-imports-next-router
+import { useRouter } from "next/router";
```
Comment (ThyMinimalDev) [inline packages/platform/examples/base/src/pages/[bookingUid].tsx:6]: puting back next/router here since we don't user app router at all

Comment (ThyMinimalDev) [reply packages/platform/examples/base/src/pages/[bookingUid].tsx:6]: example app is its own next.js

Ryukemeister (dismissed):

supalarry (dismissed):

ThyMinimalDev:
fixed ci

zomars (approved):

PR #14617, opened 2024-04-16
fix: Minimum fix for N Calendar Invites

### LINKED ISSUE #14580: [CAL-3412] a collective event of N people sends N invites

joeauyeung:
## What does this PR do?

Right now attendees are being spammed by emails from the calendar provider because we were creating calendar events on EACH team member's calendars. So if we have 4 members in a collective event, we were creating 4 different calendar events and the attendee was receiving an email for each event.

This PR is a minimum fix for this bug by only creating one calendar event on the first destination calendar that appears. 

I tested this with a collective event and all team members are still on the single calendar event and is pointing to the  member's destination calendar.

We should follow up with the changes found in this PR https://github.com/calcom/cal.com/pull/14610 once tests are created for the `EventManager` class.

Fixes https://github.com/calcom/cal.com/issues/14580 Fixes CAL-3412 Fixes CAL-3468

## Requirement/Documentation

## Type of change

## How should this be tested?

- Create a collective team event with each team member having Outlook as a destination calendar
- Create a booking under the collective event
- The attendee should receive one email from Outlook and all team members should be invited to the event

## Mandatory Tasks

## Checklist

### DIFF
--- a/packages/core/EventManager.ts
+++ b/packages/core/EventManager.ts
@@ -535,6 +535,7 @@ export default class EventManager {
     };
 
     if (event.destinationCalendar && event.destinationCalendar.length > 0) {
+      let eventCreated = false;
       // Since GCal pushes events to multiple calendars we only want to create one event per booking
       let gCalAdded = false;
       const destinationCalendars: DestinationCalendar[] = event.destinationCalendar.reduce(
@@ -554,6 +555,7 @@ export default class EventManager {
         [] as DestinationCalendar[]
       );
       for (const destination of destinationCalendars) {
+        if (eventCreated) break;
         log.silly("Creating Calendar event", JSON.stringify({ destination }));
         if (destination.credentialId) {
           let credential = this.calendarCredentials.find((c) => c.id === destination.credentialId);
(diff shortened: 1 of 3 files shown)


packages/core/EventManager.ts
```diff
@@ -554,6 +555,7 @@ export default class EventManager {
         [] as DestinationCalendar[]
       );
       for (const destination of destinationCalendars) {
```
joeauyeung, line 557: This `for ... of ` loop was the cause of creating multiple calendar events.

packages/core/EventManager.ts
```diff
@@ -554,6 +555,7 @@ export default class EventManager {
...
+        if (eventCreated) break;
```
joeauyeung, line 558: If a calendar event was created then set `eventCreated` to `true`. If the event was created on the first destination calendar then break out of the loop.

zomars (approved):
Way way clearer. Thanks a ton @joeauyeung 

Ship it

PR #14620, opened 2024-04-16
fix: API v2 recurring spelling

keithwillcode:
Fixes spelling of "recurring". It is currently "reccuring"

### DIFF
--- a/apps/api/v2/src/app.rewrites.middleware.ts
+++ b/apps/api/v2/src/app.rewrites.middleware.ts
@@ -10,6 +10,9 @@ export class RewriterMiddleware implements NestMiddleware {
     if (req.url.startsWith("/v2/ee")) {
       req.url = req.url.replace("/v2/ee", "/v2");
     }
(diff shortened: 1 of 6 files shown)


zomars (approved):

apps/api/v2/src/app.rewrites.middleware.ts
```diff
@@ -10,6 +10,9 @@ export class RewriterMiddleware implements NestMiddleware {
     if (req.url.startsWith("/v2/ee")) {
       req.url = req.url.replace("/v2/ee", "/v2");
     }
+    if (req.url.contains("reccuring")) {
```
exception, line 13: Are people already using `reccuring`?

exception (approved):

apps/api/v2/src/app.rewrites.middleware.ts
ThyMinimalDev: it's possible but probably nobody is since we didnt expose the hook to create recurring booking

PR #14624, opened 2024-04-17
fix: use 'success' value for better accuracy

chronark:
## What does this PR do?

Use the more accurate `success` response parameter to handle ratelimit decisions.
As discussed with @CarinaWolli over slack, this is the correct way to do it.

`remaining` can be inaccurate in `async` mode.

`reset` is safe to use though, as that is independent of the mode.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

Just ran it locally and it still does the same

## Mandatory Tasks

## Checklist


sean-brydon (dismissed):

packages/lib/checkRateLimitAndThrowError.test.ts
```diff
@@ -35,6 +36,7 @@ describe("checkRateLimitAndThrowError", () => {
         limit: 10,
         remaining: 5,
         reset: Date.now() + 10000,
+        success: true,
```
CarinaWolli, line 39: adding success here should fix the test

CarinaWolli (approved):

PR #14627, opened 2024-04-17
fix: Hide org tab when you are team admin

sean-brydon:
fixes: CAL-3473 CAL-3475

See [linear](https://linear.app/calcom/issue/CAL-3475/members-can-see-other-invited-team-members-to-a-team#comment-19635f2a) for loom

CleanShot 2024-04-17 at 09 54 49@2x

### DIFF
--- a/packages/trpc/server/routers/viewer/organizations/getMembers.handler.ts
+++ b/packages/trpc/server/routers/viewer/organizations/getMembers.handler.ts
@@ -15,6 +15,11 @@ export const getMembersHandler = async ({ input, ctx }: CreateOptions) => {
 
   if (!ctx.user.organizationId) return [];
 
+  const isOrgPrivate = ctx.user.organization.isPrivate;
+  const isOrgAdmin = ctx.user.organization.isOrgAdmin;
+
+  if (isOrgPrivate && !isOrgAdmin) return [];
+
   const teamQuery = await prisma.team.findUnique({
     where: {
       id: ctx.user.organizationId,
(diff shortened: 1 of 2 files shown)


Udit-takkar (approved):

Udit-takkar (commented):
Screenshot 2024-04-17 at 2 49 53 PM

As a team admin(and member of org) i don't see members of team like org admin

Screenshot 2024-04-17 at 2 49 57 PM

packages/trpc/server/routers/viewer/organizations/getMembers.handler.ts
```diff
@@ -15,6 +15,11 @@ export const getMembersHandler = async ({ input, ctx }: CreateOptions) => {
 
   if (!ctx.user.organizationId) return [];
 
+  const isOrgPrivate = ctx.user.organization.isPrivate;
+  const isOrgAdmin = ctx.user.organization.isOrgAdmin;
+
+  if (isOrgPrivate && !isOrgAdmin) return [];
+
```
Udit-takkar, line 22: This is returning empty array even when team is not private but org is private. so event team admin cannot view all the team members

sean-brydon: Oh great shout

sean-brydon: I just checked that this is nothing to do with this PR (still happens on main) 

I will merge this and do a follow up for history sake

keithwillcode:
@sean-brydon @Udit-takkar automated tests need to be added to verify this functionality

PR #14630, opened 2024-04-17
chore: settings design match

### LINKED ISSUE #12866: [CAL-2824] Settings sidebar needs a UI update
The settings sidebar doesn't match the designs. The reason is that the component in the sidebar is re-used in the event types sidebar and any changes are reflected in event types, resulting in the icon not being aligned properly. 

It was fine in the sidebar until you have a lot more items (like in Orgs) and then things start to get pretty messy. The sidebar in orgs is way too long and all the extra padding and spacing makes it really hard to scan. Particularly in the teams area.

Summary of issues

* The menu items are 4px higher than in design. Doesn't seem like much but when all items are added up it's a big difference.
* Spacing is not accurately reflected (all spec'd in Figma)
* We have the wrong colour on titles making them hard to differentiate from menu items
* We show the Organisation name instead of "Organisation" in design. Another user pointed this out recently and I agree that it's harder to scan this way.
* He also made the point about changing the top one from the username to just "My account". Given it's the first in the list I actually think it's easy to identify as is and doesn't need to change and shows context but open to trying it.
* We should alphabetize teams

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-2824](https://linear.app/calcom/issue/CAL-2824/settings-sidebar-needs-a-ui-update)</sub>

sean-brydon:
A design match for settings from a 1-1 with @ciaranha 

Fixed: [CAL-2824](https://linear.app/calcom/issue/CAL-2824/settings-sidebar-needs-a-ui-update) 

Fixed: #12866

### DIFF
--- a/apps/web/components/eventtype/EventTypeSingleLayout.tsx
+++ b/apps/web/components/eventtype/EventTypeSingleLayout.tsx
@@ -481,7 +481,6 @@ function EventTypeSingleLayout({
               sticky
               linkShallow
               itemClassname="items-start"
-              iconClassName="md:mt-px"
             />
           </div>
           <div className="p-2 md:mx-0 md:p-0 xl:hidden">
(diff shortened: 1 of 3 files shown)


Udit-takkar (approved):
Screenshot 2024-04-17 at 4 59 04 PM

LGTM

PR #14636, opened 2024-04-17
fix: 500 on forbidden showing up in logs

emrysal:
## What does this PR do?

When accessing an event type that doesn't exist or doesn't belong to you, redirect back to the /event-types page instead of throwing a 500 error.

Future improvement, show toast or error, but practically this should happen only when intentional.


keithwillcode (dismissed):
Logic looks good. Some type errors to fix though

CarinaWolli (approved):

PR #14643, opened 2024-04-17
feat: handle v2 access tokens and api keys using BearerGuard

supalarry:
## What does this PR do?

Combine AccessTokenStrategy and ApiKeyStrategy into BearerStrategy that can be applied to be used by v1 and v2 calls.


apps/api/v2/src/modules/auth/strategies/bearer/bearer.strategy.ts
```diff
@@ -0,0 +1,100 @@
...
+    super();
+  }
+
+  async authenticate(request: Request) {
+    try {
+      const bearerToken = this.getBearerToken(request);
+      let bearerTokenUser: UserWithProfile;
+
+      if (!bearerToken) {
+        throw new UnauthorizedException(INVALID_ACCESS_TOKEN);
```
ThyMinimalDev, line 29: maybe rename this error to invalid authentication credentials

PR #14652, opened 2024-04-18
feat: include org role in org session data

sean-brydon:
After discussing with hariom we have decided to include org role in org session data to make it eaiser when checking permissions on an org. 

Previosuly we were calling listCurrent every time we needed access to this value


CarinaWolli (dismissed):

Udit-takkar (dismissed):
LGTM. works fine

Udit-takkar (dismissed):

emrysal (approved):
Nice 👍  - no comments, this is a nice fixup.

PR #14672, opened 2024-04-19
fix: remove away view

CarinaWolli:
## What does this PR do?

With the new OOO feature we got rid of the away feature. Remove the following view form the profile page: 

Screenshot 2024-04-19 at 11 45 53 AM

Also, always show green dot here: 
Screenshot 2024-04-19 at 11 49 59 AM

### DIFF
--- a/apps/web/modules/users/views/users-public-view.tsx
+++ b/apps/web/modules/users/views/users-public-view.tsx
@@ -98,13 +98,13 @@ export function UserPage(props: InferGetServerSidePropsType<typeof getServerSide
               {!isOrg && user.verified && (
                 <Icon
                   name="badge-check"
-                  className=" mx-1 -mt-1 inline h-6 w-6 fill-blue-500 text-white dark:text-black"
+                  className="mx-1 -mt-1 inline h-6 w-6 fill-blue-500 text-white dark:text-black"
                 />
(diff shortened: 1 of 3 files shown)


Udit-takkar (commented):
@CarinaWolli do we also want to remove Away  from packages/features/bookings/Booker/Booker.tsx?

Screenshot 2024-04-19 at 9 42 18 PM

CarinaWolli:
> @CarinaWolli do we also want to remove Away from packages/features/bookings/Booker/Booker.tsx?
> 
> Screenshot 2024-04-19 at 9 42 18 PM

Removed, we can remove all the logic around it in a follow-up if we decide to fully get rid of the away feature

sean-brydon (approved):
LGTM

zomars (approved):
LGTM. Can we remove the away column from the users table in a followup? Also we could at least mark it as deprecated in the prisma schema

CarinaWolli:
> LGTM. Can we remove the away column from the users table in a followup? Also we could at least mark it as deprecated in the prisma schema

I will do that in a follow up, I am not sure if we made that decision to get rid of away or if it just happened with the introduction of OOO. Will clarify that first with @ciaranha

PR #14675, opened 2024-04-20
fix: Avatar API isn't ran at signup handler

emrysal:
## What does this PR do?

Start using Avatar API in the calcomHandler


keithwillcode (approved):
Code looks fine. Haven’t tested

PR #14699, opened 2024-04-22
feat: new-app-install-flow follow up

ibex088:
fixes CAL-3604

Screenshot 2024-04-22 at 6 38 18 PM

Screenshot 2024-05-11 at 1 31 02 PM


joeauyeung (changes requested):
To specify, on my comment. We should remove any instances of `userAdminTeams` now that we have the new install flow. 

This is an expensive query since we're looking for all teams & orgs the user is an admin / owner of. This should live in the new app install flow.

SomayChauhan:
Got it!!, i'm looking into it

packages/features/ee/teams/lib/getUserAdminTeams.ts
```diff
@@ -1,96 +1,29 @@
...
-  getUserInfo,
-  getParentInfo,
-  includeCredentials = false,
-}: {
-  userId: number;
-  getUserInfo?: boolean;
-  getParentInfo?: boolean;
-  includeCredentials?: boolean;
-}): Promise<UserAdminTeams> => {
-  const teams = await prisma.team.findMany({
```
SomayChauhan, line 35: instead of querying all teams and then looking weather this team has current user as admin/owner (which is slow because we have to look at all the teams in db)
queried the user using userId got all his teams and then filter all the teams that have the current user asadmin/owner
(this should be faster becaseu user_id is the primary key for the table so getting all the teams of user should be fast)

packages/features/ee/teams/lib/getUserAdminTeams.ts
```diff
@@ -1,96 +1,29 @@
...
+export const getUserAdminTeams = async (userId: number): Promise<number[]> => {
+  const user = await prisma.user.findFirst({
```
SomayChauhan, line 7: queried the user using userId got all his teams and then filter all the teams that have the current user asadmin/owner
(this should be faster becaseu user_id is the primary key for the table so getting all the teams of user should be fast)

packages/features/ee/teams/lib/getUserAdminTeams.ts
```diff
@@ -1,96 +1,29 @@
...
     where: {
-      members: {
-        some: {
-          userId: userId,
+      id: userId,
+    },
+    select: {
+      teams: {
+        where: {
```
SomayChauhan, line 13: filter all the teams that have the current user as admin/owner

packages/features/ee/teams/lib/getUserAdminTeams.ts
```diff
@@ -1,96 +1,29 @@
...
           accepted: true,
           role: { in: [MembershipRole.ADMIN, MembershipRole.OWNER] },
         },
+        select: { id: true },
```
SomayChauhan, line 17: only select the required fields

apps/web/pages/apps/installation/[[...step]].tsx
```diff
@@ -155,47 +159,46 @@ const OnboardingPage = ({
...
   const handleSelectAccount = async (teamId?: number) => {
-    try {
-      setIsSelectingAccount(true);
-      if (appMetadata.isOAuth) {
-        const state = JSON.stringify({
-          appOnboardingRedirectUrl: getAppOnboardingRedirectUrl(appMetadata.slug, teamId),
-          teamId,
-        });
-
-        const res = await fetch(
```
Comment (SomayChauhan) [inline apps/web/pages/apps/installation/[[...step]].tsx:166]: decided to remove this api call and use `useAddAppMutation` instead

apps/web/pages/apps/installation/[[...step]].tsx
```diff
@@ -155,47 +159,46 @@ const OnboardingPage = ({
...
-            appMetadata.slug == "stripe" ? "stripepayment" : appMetadata.slug
-          }/add?state=${state}`,
-          {
-            method: "GET",
-            headers: {
-              "Content-Type": "application/json",
-            },
+    mutation.mutate(
+      {
+        isOmniInstall: true,
```
Comment (SomayChauhan) [inline apps/web/pages/apps/installation/[[...step]].tsx:164]: setting `isOmniInstall` to true so that we can be redirected back to /apps/installation/[step]?slug
if not set it redirects to `apps/installed/[category]` and the redirection can't be controlled

packages/app-store/_utils/oauth/decodeOAuthState.ts
```diff
@@ -7,9 +7,6 @@ export function decodeOAuthState(req: NextApiRequest) {
     return undefined;
   }
   const state: IntegrationOAuthCallbackState = JSON.parse(req.query.state);
-  if (state.appOnboardingRedirectUrl) {
```
SomayChauhan, line 10: no longer needed as we now use returnTo only

apps/web/components/apps/AppPage.tsx
```diff
@@ -225,7 +225,6 @@ export const AppPage = ({
                         multiInstall
                         concurrentMeetings={concurrentMeetings}
                         paid={paid}
-                        dirName={dirName}
```
joeauyeung, line 228: If we're removing the use of the `dirName`, let's remove the prop from the component as well.

apps/web/components/apps/InstallAppButtonChild.tsx
```diff
@@ -108,78 +66,19 @@ export const InstallAppButtonChild = ({
...
-        onClick={_onClick}
-        color="primary"
-        size="base"
-        {...props}>
-        {multiInstall ? t("install_another") : t("install_app")}
-      </Button>
-    );
-  }
   return (
-    <Dropdown>
```
joeauyeung, line 125: I love to see this gone!

joeauyeung, line 125: I love to see this gone.

apps/web/components/apps/InstallAppButtonChild.tsx
joeauyeung, line 1: Let's try and completely remove the need for `userAdminTeams` prop here. Ideally we should be using the new app install flow.

packages/features/ee/teams/lib/getUserAdminTeams.ts
joeauyeung, line 1: I would argue the aim of this PR should be to get rid of this function. The new app install flow handles this in better way. I see that this function is only used for apps so the rest of the web app doesn't rely on it.

Anything that relied on this should be handled in the new app install flow.

joeauyeung (changes requested):

packages/features/ee/teams/lib/getUserAdminTeams.ts
SomayChauhan: Not sure if we can remove it completly
we need this query for -
* fetching the credentials for an app for the current logged in user and all his teams [LINK](https://github.com/SomayChauhan/cal.com/blob/7cb178ba2f4e60e2a3b1eea70f78e8171a783309/packages/app-store/_appRegistry.ts#L67)

* this credentials are used to check weather this app has been added or not (the green badge that says 2 installed) 
* weather to show the install button or not here for instance i have installed fathom on both my personal acc and the only team i have so we dont show the install button
Screenshot 2024-05-02 at 1 30 39 PM

so we do need the list of teams the logged in user is admin of
plus now that we are querying it using the primary ke it shouldnt be that bad

SomayChauhan:
closing in favour of #15193

PR #14700, opened 2024-04-22
fix: invalid Cancelled / Rescheduled downstream calendar block

emrysal:
## What does this PR do?

Until we've got bi-directional communication; our bookings table should not generate busyTimes when external calendars are connected.

### DIFF
--- a/packages/core/getBusyTimes.ts
+++ b/packages/core/getBusyTimes.ts
@@ -144,43 +144,50 @@ export async function getBusyTimes(params: {
       });
 
   const bookingSeatCountMap: { [x: string]: number } = {};
-  const busyTimes = bookings.reduce(
-    (aggregate: EventBusyDetails[], { id, startTime, endTime, eventType, title, ...rest }) => {
-      if (rest._count?.seatsReferences) {
-        const bookedAt = `${dayjs(startTime).utc().format()}<>${dayjs(endTime).utc().format()}`;
-        bookingSeatCountMap[bookedAt] = bookingSeatCountMap[bookedAt] || 0;
-        bookingSeatCountMap[bookedAt]++;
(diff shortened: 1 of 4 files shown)


CarinaWolli (approved):
Code looks good 👍

packages/core/getBusyTimes.ts
```diff
@@ -144,43 +144,50 @@ export async function getBusyTimes(params: {
...
-      return aggregate;
-    },
-    []
-  );
+  // Only take bookings as source of truth when there are no connected calendars
+  // this is because cancellations downstream aren't reflected in our Booking database.
+  // TODO: Reflect downstream cancellations in our Booking database and use Booking as
+  //       source of truth.
+  const busyTimes =
+    credentials?.length > 0
```
CarinaWolli, line 152: other non-calendar apps are also in credentials. So if there is no calendar but for example true we wouldn't get any busyTimes

packages/core/getBusyTimes.ts
```diff
@@ -144,43 +144,50 @@ export async function getBusyTimes(params: {
...
+              // if it does get blocked at this point; we remove the bookingSeatCountMap entry
+              // doing this allows using the map later to remove the ranges from calendar busy times.
+              delete bookingSeatCountMap[bookedAt];
+            }
+            if (rest.uid === rescheduleUid) {
+              return aggregate;
+            }
+            aggregate.push({
+              start: dayjs(startTime)
+                .subtract((eventType?.beforeEventBuffer || 0) + (afterEventBuffer || 0), "minute")
```
CarinaWolli, line 179: Aren't we now breaking our buffers with that?

CarinaWolli (commented):
@emrysal I looked at this one again and have some concerns

PR #14705, opened 2024-04-22
chore: Remove deprecated logo/avatar/away fields

emrysal:
## What does this PR do?

* Drops away, avatar fields from `users`
* Drops logo field from `Team`


CarinaWolli (dismissed):

zomars (dismissed):
Ship it

CarinaWolli:
Why do all e2e tests fail here? I already tried re-running them 🤔

keithwillcode:
@emrysal Looks like e2e 6/7 are failing consistently. Might need to look into it a bit more

keithwillcode (changes requested):
Requesting changes for this reason: https://github.com/calcom/cal.com/pull/14705#issuecomment-2096994345

keithwillcode (dismissed):
@emrysal Do we need to make 2 separate PRs here?
1) Remove the fields from the schema
2) Drop the columns

Will dropping the columns before the deployment hits production start throwing errors if queries are still using the schema will these 3 fields?

PeerRich:
what do we do here @emrysal @keithwillcode

emrysal:
> @emrysal Do we need to make 2 separate PRs here?
> 
> 1. Remove the fields from the schema
> 2. Drop the columns
> 
> Will dropping the columns before the deployment hits production start throwing errors if queries are still using the schema will these 3 fields?

So I don't think this is needed if there are no usages, but if there are there must be two PRs - unfortunately since the PR removing all logo/avatar uses this has sneaked back in. I've reverted this in this PR however this does mean two PRs are required after all. 

No problem, I'll remove the migration files, then do another PR.

zomars (approved):

Udit-takkar:
@emrysal there is a missing migration in the main branch after this was merged

PR #14730, opened 2024-04-24
fix: reschedule with google meet

### LINKED ISSUE #14086: [CAL-3462] Google Meet link missing after rescheduling team meeting
### Issue Summary

If someone schedules a team meeting and organizers have Google Meet connected as a default conferencing app, Google Meet link is on the event (both on the meeting scheduling page an in the email). If organizer **confirms it and then reschedules the event**, Google Meet link is missing from the rescheduled event (only in the email - on the meeting page it's still there).

### Prerequisites

* "Requires confirmation" option should be enabled for the event type.

### Steps to Reproduce

1. Schedule new team meeting (Google Meet should be connected) as a guest user (for example using incognito mode in the browser).
2. Organizer should **confirm it first**.
3. Then organizer should reschedule it.
4. Check the email.

### Actual Results

* After rescheduling, the email does not have a link to Google Meet.

### Expected Results

* After rescheduling, the email have a link to Google Meet.

### Technical details

* [Cal.com](http://Cal.com) v3.8.9 (docker)

### Evidence

* First email (Google Meet link is in place):

image

image

* Rescheduling email (Google Meet link is absent):

image

image

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3462](https://linear.app/calcom/issue/CAL-3462/google-meet-link-missing-after-rescheduling-team-meeting)</sub>

Udit-takkar:
## What does this PR do?

Fixes https://github.com/calcom/cal.com/issues/14086
Fixes CAL-3462

Before:-
image

After:-
Screenshot 2024-04-24 at 6 25 32 PM

## Type of change


emrysal (dismissed):
Thanks @Udit-takkar! Looks like a good fix to me for updatedEvents - nice one.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1866,7 +1866,13 @@ async function handler(
             });
           }
 
-          if (googleCalResult?.createdEvent?.hangoutLink) {
+          const googleHangoutLink =
+            googleCalResult?.createdEvent?.hangoutLink ??
+            (Array.isArray(googleCalResult?.updatedEvent)
+              ? googleCalResult.updatedEvent[0]?.hangoutLink
+              : googleCalResult?.updatedEvent?.hangoutLink);
```
hariombalhara, line 1873: AFAIU, the issue occurred because the `originalEvent` didn't have videoCallData(because it was confirmed later and then the videoCallData isn't updated I guess). 

The thing is that we should have used the latest link returned by the updateMeeting/updateEvent call(which is available in `updatedEvent`). We can fallback to `createdEvent` just in case. It ensure that the link expiry is up-to-date(and maybe other details).

Though I tested, and it makes sense that, during reschedule `createdEvent` won't be there. So, technically we should just use `updatedEvent`.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -1875,31 +1881,33 @@ async function handler(
...
             });
           }
         }
-
-        metadata.hangoutLink = results[0].createdEvent?.hangoutLink;
-        metadata.conferenceData = results[0].createdEvent?.conferenceData;
-        metadata.entryPoints = results[0].createdEvent?.entryPoints;
+        const createdOrUpdatedEvent =
+          results[0]?.createdEvent ??
+          (Array.isArray(results[0]?.updatedEvent) ? results[0]?.updatedEvent[0] : results[0]?.updatedEvent);
```
hariombalhara, line 1903: Again because we are in reschedule case we should use `updatedEvent` and fallback to `createdEvent`(which I think won't exist in reschedule case)

packages/features/bookings/lib/handleNewBooking/test/reschedule.test.ts
```diff
@@ -889,6 +891,203 @@ describe("handleNewBooking", () => {
           timeout
         );
 
+        test(
```
hariombalhara, line 894: This is inside 'User Event Type' describe. There is a 'Team Event Type' as well. We should move it below to that describe

Also, I believe the issue is with a regular User event as well if requires-confirmation is enabled, so we can have a test for User Event as well at this place.
Screenshot 2024-05-02 at 1 53 24 PM

packages/features/bookings/lib/handleNewBooking/test/reschedule.test.ts
```diff
@@ -889,6 +891,203 @@ describe("handleNewBooking", () => {
...
+              location: OrganizerDefaultConferencingAppType,
+              subscriberUrl,
+              eventType: scenarioData.eventTypes[0],
+            });
+
+            expectSuccessfulCalendarEventDeletionInCalendar(calendarMock, {
+              externalCalendarId: "MOCK_EXTERNAL_CALENDAR_ID",
+              calEvent: {},
+              uid: "MOCK_ID",
+            });
```
hariombalhara, line 1086: These assertions seem to be passing in main as well for me. Could you recheck that if we ever have this issue again this test would catch it.

hariombalhara (changes requested):
Left a few comments

packages/features/bookings/lib/handleNewBooking.ts
Udit-takkar: > Though I tested, and it makes sense that, during reschedule createdEvent won't be there. So, technically we should just use updatedEvent.

I also thought about only using updatedEvent but when you reschedule and add a new guest in booking form then you would get createdEvent.

Screenshot 2024-05-04 at 7 20 55 PM

Udit-takkar:
Screenshot 2024-05-06 at 4 57 16 PM

@hariombalhara Added appStatus assertion and verified that it fails on main branch

packages/features/bookings/lib/handleNewBooking/test/reschedule.test.ts
```diff
@@ -1138,6 +1139,250 @@ describe("handleNewBooking", () => {
           },
           timeout
         );
+        test(
+          `[GOOGLE MEET AS LOCATIOn]should rechedule a booking, that requires confirmation, without confirmation - When booker is the organizer of the existing booking as well as the event-type
```
Udit-takkar, line 1143: Added test on individual event type

packages/features/bookings/lib/handleNewBooking.ts
hariombalhara: Ah cool that's something i missed. But we should still prioritize `updatedEvent` over `createdEvent`. Not sure if that change is done already.

apps/web/test/utils/bookingScenario/bookingScenario.ts
```diff
@@ -1164,18 +1190,36 @@ export function mockCalendar(
...
+                  iCalUID: normalizedCalendarData.update?.iCalUID,
+
+                  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                  id: normalizedCalendarData.update?.uid || "FALLBACK_MOCK_ID",
+                  // Password and URL seems useless for CalendarService, plan to remove them if that's the case
+                  password: "MOCK_PASSWORD",
+                  url: "https://UNUSED_URL",
+                  location: "https://UNUSED_URL",
+                  hangoutLink: "https://UNUSED_URL",
+                  conferenceData: event.conferenceData,
```
hariombalhara, line 1207: nit: We could just add these props conditionally.  Allows us to easily understand the difference

hariombalhara (approved):

PR #14736, opened 2024-04-24
fix: Check permissions for team OOO entries

zomars:
## What does this PR do?

Fixes CAL-3527

## Requirement/Documentation

## Type of change

## How should this be tested?

Please refer to CAL-3527 ticket in Linear

## Mandatory Tasks

### DIFF
--- a/packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
@@ -50,7 +50,7 @@ export const outOfOfficeCreate = async ({ ctx, input }: TBookingRedirect) => {
     const user = await prisma.user.findUnique({
       where: {
         id: input.toTeamUserId,
-        /** You can only create OOO for members of teams you're either ADMIN or OWNER */
+        /** You can only create OOO for members of teams you belong to */
         teams: {
           some: {
             userId: ctx.user.id,
--- a/packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
@@ -55,7 +55,6 @@ export const outOfOfficeCreate = async ({ ctx, input }: TBookingRedirect) => {
           some: {
             userId: ctx.user.id,
             accepted: true,
-            role: { in: ["OWNER", "ADMIN"] },
           },
         },
       },
--- a/packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
+++ b/packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
@@ -50,6 +50,14 @@ export const outOfOfficeCreate = async ({ ctx, input }: TBookingRedirect) => {
     const user = await prisma.user.findUnique({
       where: {
         id: input.toTeamUserId,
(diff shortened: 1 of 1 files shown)


packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
```diff
@@ -50,6 +50,14 @@ export const outOfOfficeCreate = async ({ ctx, input }: TBookingRedirect) => {
...
       where: {
         id: input.toTeamUserId,
+        /** You can only create OOO for members of teams you're either ADMIN or OWNER */
+        teams: {
+          some: {
+            userId: ctx.user.id,
+            accepted: true,
+            role: { in: ["OWNER", "ADMIN"] },
+          },
+        },
```
zomars, line 60: You will receive an obfuscated not found error if you try to create entries for other users.

packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
```diff
@@ -194,28 +202,18 @@ export const outOfOfficeEntryDelete = async ({ ctx, input }: TBookingRedirectDel
     throw new TRPCError({ code: "BAD_REQUEST", message: "out_of_office_id_required" });
   }
 
-  // Validate outOfOfficeEntry belongs to the user deleting it
-  const outOfOfficeEntry = await prisma.outOfOfficeEntry.findFirst({
-    select: {
-      uuid: true,
-      userId: true,
-    },
+  const deletedOutOfOfficeEntry = await prisma.outOfOfficeEntry.delete({
```
zomars, line 205: Refactored so this runs in a single query

zomars (commented):
Self review done

emrysal (dismissed):

packages/trpc/server/routers/loggedInViewer/outOfOffice.handler.ts
```diff
@@ -50,6 +50,14 @@ export const outOfOfficeCreate = async ({ ctx, input }: TBookingRedirect) => {
     const user = await prisma.user.findUnique({
       where: {
         id: input.toTeamUserId,
+        /** You can only create OOO for members of teams you're either ADMIN or OWNER */
+        teams: {
```
CarinaWolli, line 54: shouldn't we rather check here if `ctx.user.id` and `toTeamUserId` are in the same teams
?

zomars: RN we're checking if `toTeamUserId` belongs to a team where the `ctx.user.id` is either an ADMIN or OWNER. Isn't that what we want?

zomars: I see. Then we can either add MEMBER to the role check or skip the role check altogether.

zomars: I see. Then we can either add MEMBER to the role check or skip the role check altogether.

zomars: Addressed here [b9ac480](https://github.com/calcom/cal.com/pull/14736/commits/b9ac480c42c0664a6f6388621ace4a0f01fe171f)

CarinaWolli (approved):

PR #14738, opened 2024-04-24
fix: Impersonation flicker on timezone change

emrysal:
## What does this PR do?

* Removes dependency on `IS_VISUAL_REGRESSION_TESTING` - instead we set a global cookie that disables the timezone change popup. It is possible to remove this cookie to confirm the timezone change works properly (but maybe best tested with vitest and not e2e).
* Extract DialogContent for improved performance of Shell
* Extract open/setOpen to custom hook for clarity & also performance


zomars (approved):
LGTM

PR #14741, opened 2024-04-25
feat: Hanko passkeys support

### LINKED ISSUE #13342: [CAL-2984] Add Passkey support

merlinaudio:
Note: This previously was #13127 but has since been reworked and simplified

## What does this PR do?
Fixes https://github.com/calcom/cal.com/issues/13342
Fixes CAL-2984

This PR adds passkey support for logging in through Hanko Passkeys.

## Type of change

- This change requires a documentation update (probably)

## How should this be tested?

- Go to [cloud.hanko.io](https://cloud.hanko.io) and create a "Passkey API" project
  - Then create a new API key
    set it to `HANKO_PASSKEYS_API_KEY="<api key>"`
    (quotes important)
  - Copy the tenant ID to `NEXT_PUBLIC_HANKO_PASSKEYS_TENANT_ID=<tenant id>`
- Log in as normal (with password/google/...)
- Go to settings > passkeys > create a new passkey
- Log out
- Go to login > sign in with a passkey

## Mandatory Tasks

## next-auth-options.ts

This file now includes a PasskeyProvider.

For its `authorize` function, I've collected all of the important checks I could find in the email+password provider (a.k.a. CredentialsProvider), such as rate-limiting, user.locked, belongsToActiveTeams, ...) and included them in the PasskeyProvider.

For some of these, the fields/relations returned by `UserRepository.findById` wasn't enough. Unfortunately `.findTeamsByUserId` doesn't return enough info either

I didn't want to touch any of the existing queries, so I've added `UserRepository.findByIdAndIncludeProfilesAndPassword`

...which is the exact same as the email version of this function (`findByEmailAndInclude...`). It just looks the user up by their ID instead.

## Checklist

- I haven't checked if my PR needs changes to the documentation (unlikely though)
- I haven't added tests that prove my fix is effective or that my feature works (this is possible with Playwright)

#### Note about images

These are the same images from the old PR. There's one minor change since then: since icon components aren't supported for `Button`, `EmptyPage`, ... anymore, I used Lucide's `key-round` icon

This is what **key-round** looks like:

Versus the passkey icon:

---

### Adds passkey button to login page:
CleanShot 2024-01-09 at 17 36 45@2x

### Adds "Passkeys" settings section to sidebar:
CleanShot 2024-01-09 at 17 44 38@2x

### Adds passkeys settings page

**Empty:**
CleanShot 2024-01-09 at 17 45 45@2x

---

Clicking on the "add" button and "sign in with a passkey" buttons opens a browser dialog. This looks different for every browser, phone, etc.

**For example:**
CleanShot 2024-01-09 at 17 47 48@2x

**With passkeys added:**
CleanShot 2024-01-09 at 17 49 38@2x
- The name is set by the Hanko Passkeys API (our backend). We'll improve the name soon so it's not just a semi-random string and more user friendly
- If the passkey was used, the description says e.g. "Created 3 days ago • Last used 5 minutes ago" 

**Dropdown** stolen straight from API settings page:
CleanShot 2024-01-09 at 17 53 13@2x


.env.example
```diff
@@ -116,15 +116,31 @@ NEXT_PUBLIC_FRESHCHAT_HOST=
 # @see https://support.google.com/cloud/answer/6158849#public-and-internal&zippy=%2Cpublic-and-internal-applications
 GOOGLE_LOGIN_ENABLED=false
 
-#   - GOOGLE CALENDAR/MEET/LOGIN
+# Hanko Passkeys Config
```
PeerRich, line 119: lets mention this lower in the file

PeerRich (dismissed):

.env.example
merlindru: yeah was unsure about that. wanted to place it below google auth stuff since i did that everywhere else, but agreed, for the env file its a little too high up

merlindru:
We'd really like to use the passkey icon (not part of lucide, it's an SVG)

however the components that made use of it (Button, EmptyPage, ...) have all been changed to only accept a lucide icon name (string)

Is there any way we still could use a custom component as an icon for these?

merlindru:
While PASSKEY_LOGIN_ENABLED hides all passkey related UI on the login page, the settings page as well as the API routes to register and remove passkeys stay *enabled*

These aren't strictly related to logging in with Passkeys; do you feel like the env var should disable everything related to passkeys or just the ability to login?

apps/web/pages/api/auth/passkeys/register/finalize.ts
```diff
@@ -0,0 +1,41 @@
...
+import prisma from "@calcom/prisma";
+
+const tenantId = process.env.NEXT_PUBLIC_HANKO_PASSKEYS_TENANT_ID ?? "";
+const apiKey = process.env.HANKO_PASSKEYS_API_KEY ?? "";
+
+const passkeyApi = tenant({ tenantId, apiKey });
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  if (!tenantId || !apiKey) {
+    return res.status(503).json({ message: "Passkey API not configured" });
```
emrysal, line 15: Is 503 the best status code to use here?

apps/web/pages/auth/login.tsx
```diff
@@ -254,6 +258,24 @@ inferSSRProps<typeof getServerSideProps> & WithNonceProps<{}>) {
...
+                {isPasskeyLoginEnabled && (
+                  <Button
+                    color="secondary"
+                    className="w-full justify-center"
+                    disabled={formState.isSubmitting}
+                    StartIcon="key-round"
+                    onClick={async (e) => {
+                      e.preventDefault();
+                      await signInWithPasskey({
+                        tenantId: process.env.NEXT_PUBLIC_HANKO_PASSKEYS_TENANT_ID!,
```
emrysal, line 270: Forbidden non-null assertion needs an eslint comment to exclude or check to prevent

apps/web/pages/settings/security/passkeys.tsx
```diff
@@ -0,0 +1,197 @@
...
+};
+
+interface Passkey {
+  id: string;
+  name?: string;
+  createdAt: string;
+  lastUsedAt?: string;
+}
+
+const PasskeyListItem = (props: Passkey & { onRemoved?: () => any }) => {
```
emrysal, line 44: Try to avoid using any

emrysal (changes requested):
> While PASSKEY_LOGIN_ENABLED hides all passkey related UI on the login page, the settings page as well as the API routes to register and remove passkeys stay enabled
> 
> These aren't strictly related to logging in with Passkeys; do you feel like the env var should disable everything related to passkeys or just the ability to login?

In line with current practises it's not needed to fully "unregister" routes 

> We'd really like to use the passkey icon (not part of lucide, it's an SVG)

Feel free to modify the button component to allow StartIcon's that aren't strings, ideally in a way it doesn't accept lucide icons (strings should be used instead).

apps/web/package.json
```diff
@@ -62,6 +63,7 @@
     "@stripe/react-stripe-js": "^1.10.0",
     "@stripe/stripe-js": "^1.35.0",
     "@tanstack/react-query": "^5.17.15",
+    "@teamhanko/passkeys-next-auth-provider": "^0.2.7",
```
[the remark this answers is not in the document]
merlindru: I'm not sure since when npm has this, but there is **optionalDependencies**

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies

Was supported in yarn v1 so I assume v2+ have it too

apps/web/pages/auth/login.tsx
merlindru: I'll check for it explicitly

apps/web/pages/api/auth/passkeys/register/finalize.ts
merlindru: Other routes use 403: https://github.com/calcom/cal.com/blob/main/apps/web/pages/api/auth/signup.ts#L23

However, [MDN says](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)

> [...] 503 Service Unavailable server error response code indicates that the server is not ready to handle the request.
>
> Common causes are a server that is down for maintenance or that is overloaded. This response should be used for temporary conditions [...]

While in my mind 403 has always been more for insufficient permissions etc

What do you think?

keithwillcode: I don't think a 403 is correct here. 501 is used for not implemented.

emrysal: I'd prefer 403 Forbidden, 503 implies a server issue but most of the time this'll be a client issue (don't access the route)

keithwillcode: Don't think a 403 is correct here though because that leads the client to believe they are lacking permissions when in reality the functionality isn't supported at all. Would consider 501 or 405 instead.

keithwillcode: See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

merlindru: If everyone is fine with it, will go with 501 - seems like the most appropriate IMO

apps/web/package.json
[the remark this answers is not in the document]
merlindru: Opted not to do this for now because optionalDependencies isn't in use for any other packages atm. Also I'm unsure if a build without installing optional deps would be successful, especially because of `import` behavior, haven't tested it

Do you want me to test this? What other not-strictly-needed dependencies are there that could be moved to `optionalDependencies` (at least in theory)?

packages/features/auth/lib/next-auth-options.ts
```diff
@@ -34,11 +39,19 @@ import CalComAdapter from "./next-auth-custom-adapter";
...
   JSON.parse(GOOGLE_API_CREDENTIALS)?.web || {};
 const GOOGLE_LOGIN_ENABLED = process.env.GOOGLE_LOGIN_ENABLED === "true";
 const IS_GOOGLE_LOGIN_ENABLED = !!(GOOGLE_CLIENT_ID && GOOGLE_CLIENT_SECRET && GOOGLE_LOGIN_ENABLED);
+
+const IS_PASSKEY_LOGIN_ENABLED = !!(
+  process.env.PASSKEY_LOGIN_ENABLED === "true" &&
+  process.env.NEXT_PUBLIC_HANKO_PASSKEYS_TENANT_ID &&
+  process.env.HANKO_PASSKEYS_API_KEY
+);
+
```
sean-brydon, line 54: I think we already have this in constants - lets put it from there

merlindru: We do, but the `GOOGLE` variables are also duplicated in next-auth-options, so I did the same to stay consistent. Do you know if there's any reason for this?

Otherwise we could import both from constants.ts.
(I don't think the google constants differ from the ones in next-aut-options.ts in any way)

emrysal (dismissed):

PeerRich (dismissed):
approved on behalf of @emrysal -- i just updated yarn.lock

merlindru:
Thank you @PeerRich and reviewers! If this is it, then I'll make one commit to change the `: any` to `void` that Sean mentioned above. I don't think there's anything else to change.

emrysal (dismissed):
There was a missing "use client"; which I added, should be good to merge now 👏

merlindru:
@emrysal Sorry, maybe I'm being dumb, but which e2e test is failing/how can I run it?

With `yarn e2e`, I don't see anything passkeys related on my end, but there are hundreds of failing tests there, so maybe my setup is borked

Also thanks for staying on top of this :) Much appreciated!

Amit91848:
All tests seems to be passing @emrysal , can you have a look again ?

PeerRich:
and what are the pros and cons vs in-house without dependency?

merlindru:
@PeerRich one of the major benefits is that you don't need to make any changes to your DB

Also, Hanko maintains the passkey related stuff, so you seldom (if ever) will need to make changes to Cal.com to keep up with new specs/browser changes

PR #14756, opened 2024-04-25
feat: Billing Service

exception:
## What does this PR do?

Introduces a new Billing Service to Platform API. For now, it is used to automate billing and usage for platform-enabled teams. 

Relies upon Stripe for handling subscriptions.

### How to Test

- You need to have a valid platform-enabled team locally
- You need to configure `config/stripe.config.json` for the plans you need to test
- You need Stripe API key and you need to install stripe client
- Run `stripe listen --events customer.subscription.created,customer.subscription.updated --forward-to localhost:5555/api/v2/billing/webhook`
- Do a `POST` to `localhost:5555/api/v2/billing/{{your team id}}/subscribe`. The body needs to have the desired plan, eg `"plan": "STARTER"`. You will get a redirect URI, which you need to visit` and run through the subscribe flow. This will trigger a webhook call which will update your team with the required billing data on Prisma.


keithwillcode:
@exception csn you please add a description

apps/api/v2/src/modules/billing/billing.service.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  constructor(
+    private readonly teamsRepository: OrganizationsRepository,
+    public readonly stripeService: StripeService,
+    private readonly billingRepository: BillingRepository,
+    private readonly configService: ConfigService<AppConfig>
+  ) {
+    this.webAppUrl = configService.get("app.baseUrl", { infer: true }) ?? "https://app.cal.com";
+    this.plansToPriceId = new Map<PlatformPlan, string>();
+    // for (const plan in Object.keys(PlatformPlan)) {
```
keithwillcode, line 25: Delete this?

apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
```diff
@@ -59,7 +59,7 @@ export class OAuthClientRepository {
   }
 
   async getOAuthClientWithRefreshSecret(clientId: string, clientSecret: string, refreshToken: string) {
-    return await this.dbRead.prisma.platformOAuthClient.findFirst({
+    return this.dbRead.prisma.platformOAuthClient.findFirst({
```
keithwillcode, line 62: Why no await here?

apps/api/v2/src/modules/stripe/stripe.service.ts
```diff
@@ -0,0 +1,15 @@
...
+import { ConfigService } from "@nestjs/config";
+import Stripe from "stripe";
+
+@Injectable()
+export class StripeService {
+  public stripe: Stripe;
+
+  constructor(configService: ConfigService<AppConfig>) {
+    this.stripe = new Stripe(configService.get("stripe.apiKey", { infer: true }) ?? "", {
+      apiVersion: "2020-08-27",
```
keithwillcode, line 12: Why this version?

keithwillcode (commented):
Overall looks good. 

Had some questions and we should add some tests.

apps/api/v2/src/modules/stripe/stripe.service.ts
exception: No reason, just the one it auto filled, will set to latest.

apps/api/v2/src/modules/oauth-clients/oauth-client.repository.ts
exception: It's already a Promise, it's being awaited in the actual usage.

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+  @Get("/:teamId/check")
+  @UseGuards(NextAuthGuard, OrganizationRolesGuard)
+  @Roles(["OWNER", "ADMIN"])
+  async checkTeamBilling(@Param("teamId") teamId: number): Promise<CheckPlatformBillingResponseDto> {
+    const teamBilling = await this.billingService.getBillingData(teamId);
+
+    return {
+      status: "success",
+      valid: teamBilling.status === "valid",
+    };
```
ThyMinimalDev, line 52: should we return the active plan of the user here ?

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+    private readonly billingService: BillingService,
+    private readonly configService: ConfigService<AppConfig>
+  ) {
+    this.stripeWhSecret = configService.get("stripe.webhookSecret", { infer: true }) ?? "";
+  }
+
+  @Get("/:teamId/check")
+  @UseGuards(NextAuthGuard, OrganizationRolesGuard)
+  @Roles(["OWNER", "ADMIN"])
+  async checkTeamBilling(@Param("teamId") teamId: number): Promise<CheckPlatformBillingResponseDto> {
```
ThyMinimalDev, line 46: should we use ApiResponse type here

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+    };
+  }
+
+  @Post("/:teamId/subscribe")
+  @UseGuards(NextAuthGuard, OrganizationRolesGuard)
+  @Roles(["OWNER", "ADMIN"])
+  async subscribeTeamToStripe(
+    @Param("teamId") teamId: number,
+    @Body() input: SubscribeToPlanInput
+  ): Promise<SubscribeTeamToBillingResponseDto> {
```
ThyMinimalDev, line 61: should we use ApiResponse type here

ThyMinimalDev: Promise<ApiResponse<SubscribeTeamToBillingResponseDto>>

PeerRich:
can you record a loom.com demo0

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
exception: What's the benefit against the Response Dto?

apps/api/v2/src/app.module.ts
```diff
@@ -53,7 +55,17 @@ import { AppController } from "./app.controller";
 })
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer): void {
-    consumer.apply(AppLoggerMiddleware).forRoutes("*");
-    consumer.apply(RewriterMiddleware).forRoutes("/");
+    consumer
+      .apply(RawBodyMiddleware)
```
exception, line 59: Stripe requires a Buffer/String for the Webhooks to be built. Since NestJS by default parses everything as JSON, this was not letting the event through. That's why this change was required.

apps/api/v2/src/modules/billing/billing.service.ts
```diff
@@ -63,17 +64,20 @@ export class BillingService {
         line_items: [
           {
             price: this.plansToPriceId.get(plan),
+            quantity: 1,
```
exception, line 67: This is having some sort of weird behaviour due to metered usage.

apps/api/v2/src/modules/stripe/stripe.service.ts
exception: @keithwillcode this is actually hardcoded by Stripe to this version. Doesn't look like we can differ from it, all other Stripe usages in Web App use this version too, and it's set to default in our Stripe dashboard too.

apps/api/v2/src/middleware/app.logger.middleware.ts
```diff
@@ -1,7 +1,7 @@
 import { Injectable, NestMiddleware, Logger } from "@nestjs/common";
 import { Request, NextFunction } from "express";
 
-import { Response } from "@calcom/platform-types";
+import { Response } from "../../../../../packages/platform/types/dist";
```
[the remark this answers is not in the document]
exception: Huh, I am actually unsure why this changed...

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
exception: Doesn't look like we're using ApiResponse much anymore, so I don't see a point on using it when we're already shaping responses as outputs as classes.

keithwillcode:
Please don’t enable auto-merge.

apps/api/v2/src/modules/billing/billing.repository.ts
```diff
@@ -0,0 +1,51 @@
...
+          some: {
+            id: clientId,
+          },
+        },
+      },
+      select: {
+        id: true,
+      },
+    });
+  }
```
ThyMinimalDev, line 30: this should probably be in the organization.repository ?

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
ThyMinimalDev: the main goal is to force a standard on the shape of api responses
 and so that that the DTO only targets the actual data sent back

Promise<ApiResponse < DTO> >

apps/api/v2/src/modules/oauth-clients/controllers/oauth-clients/oauth-clients.controller.ts
```diff
@@ -60,11 +68,17 @@ export class OAuthClientsController {
   async createOAuthClient(
     @GetUser() user: UserWithProfile,
     @Body() body: CreateOAuthClientInput
-  ): Promise<CreateOAuthClientResponseDto> {
+  ): Promise<CreateOAuthClientResponseDto | CreateOauthClientRedirect> {
```
ThyMinimalDev, line 71: I think the Redirect part is not needed anymore ?

apps/web/lib/hooks/settings/organizations/platform/oauth-clients/usePersistOAuthClient.ts
```diff
@@ -34,6 +39,8 @@ export const useCreateOAuthClient = (
     onSuccess: (data) => {
       if (data.status === SUCCESS_STATUS) {
         onSuccess?.();
+      } else if (data.status === REDIRECT_STATUS) {
+        window.location.href = data.url;
```
ThyMinimalDev, line 43: seems like the api endpoint never sends redirect status ?

apps/api/v2/src/modules/billing/controllers/billing.controller.ts
ThyMinimalDev: apiResponse could be used directly in the dto as well I think, as you see fit

keithwillcode (approved):

PR #14788, opened 2024-04-29
fix: With Credential Sync enabled, fix expired google calendar token handling - Refresh wasn't working

hariombalhara:
## What does this PR do?

Fixes issue with refreshed access_token not being used when credential sync is enabled.
Note that the issue isn't replicable if the refresh_token is set. You have to delete the `refresh_token` to be able to replicate the issue(with credential sync enabled)

## Type of change

## How should this be tested?
1.  Make sure that credential sync is enabled
2. Connect Google Calendar
3. Delete keys.refresh_token from DB
4. `expiry_date` Use a past time here so that the token expires
5. Simply refresh `apps/installed/calendar` here. On main, auto-refresh won't occur but in this branch it will occur.

## Mandatory Tasks

## Tests
Working for tests in a follow-up PR. Hitting some issues with mocking. Need to figure that out.

## Checklist


hariombalhara:
* **#14788** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/14788?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -85,13 +85,15 @@ export default class GoogleCalendarService implements Calendar {
     this.credential = credential;
     this.auth = this.initGoogleAuth(credential);
     this.log = log.getSubLogger({ prefix: [`[[lib] ${this.integrationName}`] });
-    this.credential = credential;
   }
 
   private async getMyGoogleAuthSingleton() {
-    const googleCredentials = OAuth2UniversalSchema.parse(this.credential.key);
+    if (this.myGoogleAuth) {
```
hariombalhara, line 91: Simplified the code.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -85,13 +85,15 @@ export default class GoogleCalendarService implements Calendar {
...
+      return this.myGoogleAuth;
+    }
     const { client_id, client_secret, redirect_uris } = await getGoogleAppKeys();
-    this.myGoogleAuth = this.myGoogleAuth || new MyGoogleAuth(client_id, client_secret, redirect_uris[0]);
+    const googleCredentials = OAuth2UniversalSchema.parse(this.credential.key);
+    this.myGoogleAuth = new MyGoogleAuth(client_id, client_secret, redirect_uris[0]);
     this.myGoogleAuth.setCredentials(googleCredentials);
```
hariombalhara, line 97: Now `setCredentials` is also called only the first time. As `getMyGoogleAuthSingleton` can be called multiple times, we don't need to set it again and again.

packages/app-store/googlecalendar/lib/CalendarService.ts
```diff
@@ -151,14 +153,18 @@ export default class GoogleCalendarService implements Calendar {
...
           where: {
             id: credential.id,
           },
           data: {
             key: token,
           },
         });
+
+        // Update cached credential as well
+        this.credential.key = key;
```
hariombalhara, line 167: This is important line which fixes the issue.

We update the token in DB. But in case of google-calendar there is a cached value here as well. We should keep it upto-date in case someone use this value.

joeauyeung (approved):
Tested credential syncing GCal w/o `refresh_token` it's grabbing the new token and working as expected.

PR #14796, opened 2024-04-29
feat: orgs trigger alert when loading a calendar and no availability is found

### LINKED ISSUE #14609: Trigger alert when loading a calendar and no availability is loaded at all

For auditing reasons, having alerts set up that inform org owners/admins that certain team members have their availability completely turned off or blocked will help the owners/admins work with team members to get their availability set correctly.

Upon loading a booking page for a user, if the calendar is completely unavailable, send an alert to the configured person.
Configurable to be sent to the owner, admin or user in the org.

sean-brydon:
## What does this PR do?

Fixes CAL-3480
Fixes #14609 

Loom: https://share.cleanshot.com/B5yBz18n

## Type of change

## How should this be tested?

- Run migrations
- Have upstash setup 
- Enable this setting for orgs 
- Visit a booking page with 0 availability 2 times 
- Check mails
- Wait for expiry (in dev its 60 seconds)
- try again 
- See email again

## Mandatory Tasks


packages/trpc/server/routers/viewer/organizations/update.handler.ts
```diff
@@ -32,17 +32,24 @@ const updateOrganizationSettings = async ({
...
-    return;
+  const data: Prisma.OrganizationSettingsUpdateInput = {};
+
+  if (input.hasOwnProperty("lockEventTypeCreation")) {
+    data.lockEventTypeCreationForUsers = input.lockEventTypeCreation;
+  }
+
+  if (input.hasOwnProperty("adminGetsNoSlotsNotification")) {
+    data.adminGetsNoSlotsNotification = input.adminGetsNoSlotsNotification;
   }
```
sean-brydon, line 43: We have to construct data property so we only update what has changed - allows us to pass in only the orgSetting we want to update and not have the previous state of for data before

packages/trpc/server/routers/viewer/slots/handleNotificationWhenNoSlots.ts
```diff
@@ -0,0 +1,123 @@
...
+  eventSlug: string;
+  startTime: Dayjs;
+  visitorTimezone?: string;
+  visitorUid?: string;
+};
+
+// 7 days or 60s in dev
+const NO_SLOTS_NOTIFICATION_FREQUENCY = IS_PRODUCTION ? 604_800 : 60;
+
+const NO_SLOTS_COUNT_FOR_NOTIFICATION = 2;
```
sean-brydon, line 20: We will make these options configureable on a perOrg basis if required. 

Will allign with customer to find out what they want initally

packages/trpc/server/routers/viewer/slots/handleNotificationWhenNoSlots.ts
```diff
@@ -0,0 +1,123 @@
...
+  if (!orgSettings?.organizationSettings?.adminGetsNoSlotsNotification) return;
+
+  const redis = Redis.fromEnv();
+
+  const usersUniqueKey = constructRedisKey(eventDetails, orgDetails.currentOrgDomain);
+  // Get only the required amount of data so the request is as small as possible
+  // We may need to get more data and check the startDate occurrence of this
+  // Not trigger email if the start months are the same
+  const usersExistingNoSlots = await redis.lrange(usersUniqueKey, 0, NO_SLOTS_COUNT_FOR_NOTIFICATION - 1);
+  await redis.lpush(usersUniqueKey, constructDataHash(eventDetails));
```
sean-brydon, line 72: lpush creates and pushes to key if no key is found

packages/trpc/server/routers/viewer/slots/handleNotificationWhenNoSlots.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  if (!usersExistingNoSlots.length) {
+    await redis.expire(usersUniqueKey, NO_SLOTS_NOTIFICATION_FREQUENCY);
+  }
```
sean-brydon, line 76: If this is the first push to the key - we push a expiry date (to the key not each item in the array)

packages/features/ee/organizations/pages/components/AdminNoSlotsNotificationSwitch.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  if (!isAdminOrOwner) return null;
+
+  return (
+    <>
+      <SettingsToggle
+        toggleSwitchAtTheEnd={true}
+        title={t("organization_no_slots_notification_switch_title")}
+        disabled={mutation?.isPending || !isAdminOrOwner}
+        description={t("organization_no_slots_notification_switch_description")}
```
Udit-takkar, line 40: organization_no_slots_notification_switch_title and organization_no_slots_notification_switch_description doesn't exist in common.json

packages/features/ee/organizations/pages/components/AdminNoSlotsNotificationSwitch.tsx
```diff
@@ -0,0 +1,52 @@
...
+        checked={notifcationActive}
```
Udit-takkar, line 41:
```suggestion
        checked={!!notifcationActive}
```

To fix type error

packages/features/ee/organizations/pages/components/AdminNoSlotsNotificationSwitch.tsx
sean-brydon: Ahh yeah this should be draft still!

packages/features/ee/organizations/pages/settings/general.tsx
```diff
@@ -87,6 +88,7 @@ const OrgGeneralView = () => {
       />
 
       <LockEventTypeSwitch currentOrg={currentOrg} isAdminOrOwner={!!isAdminOrOwner} />
+      <AdminNoSlotsNotificationSwitch currentOrg={currentOrg} isAdminOrOwner={!!isAdminOrOwner} />
```
hariombalhara, line 91: nit: I think we should rename the component to `NoSlotNotificationSwitch` as it takes `isAdminOrOwner`  that helps it to decide what should be shown.

Another way could be that we keep the name same(but no isAdminOrOwner prop. We can then instead conditionally render the component here based on isAdminOrOwner

packages/features/ee/organizations/pages/components/AdminNoSlotsNotificationSwitch.tsx
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  if (!isAdminOrOwner) return null;
+
+  return (
+    <>
+      <SettingsToggle
+        toggleSwitchAtTheEnd={true}
+        title={t("organization_no_slots_notification_switch_title")}
+        disabled={mutation?.isPending || !isAdminOrOwner}
```
hariombalhara, line 39: This component renders null if `isAdminOrOwner` is falsy. So, it is unused here.

packages/features/redis/IRedisService.ts
hariombalhara, line 1: Shouldn't we make it a d.ts?

sean-brydon: Sure - we should agree on a standard across the app cause we don't always do this

packages/features/redis/IRedisService.d.ts
```diff
@@ -0,0 +1,15 @@
+export interface IRedisService {
```
hariombalhara, line 1: Why do we even need this file? I just tried having RedisService.ts without `implements` and type-check passes.

hariombalhara: Makes sense !

packages/trpc/server/routers/viewer/slots/handleNotificationWhenNoSlots.ts
```diff
@@ -0,0 +1,128 @@
...
+  startTime: Dayjs;
+  visitorTimezone?: string;
+  visitorUid?: string;
+};
+
+// Incase any updates are made - lets version the key so we can invalidate
+const REDIS_KEY_VERSION = "V1";
+
+// 7 days or 60s in dev
+const NO_SLOTS_NOTIFICATION_FREQUENCY = IS_PRODUCTION ? 604_800 : 60;
```
hariombalhara, line 20:
```suggestion
const NO_SLOTS_NOTIFICATION_FREQUENCY = IS_PRODUCTION ? 7*24*3600 : 60;
```

packages/trpc/server/routers/viewer/slots/handleNotificationWhenNoSlots.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  return JSON.stringify(obj);
+};
+
+export const handleNotificationWhenNoSlots = async ({
+  eventDetails,
+  orgDetails,
+}: {
+  eventDetails: EventDetails;
+  orgDetails: { currentOrgDomain: string | null; isValidOrgDomain: boolean };
```
hariombalhara, line 45:
```suggestion
  orgDetails: { currentOrgDomain: string | null;};
```

`isValidOrgDomain` doesn't seem to be used

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -680,6 +681,18 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions): Pro
   );
   loggerWithEventDetails.debug(`Available slots: ${JSON.stringify(computedAvailableSlots)}`);
 
+  // We only want to run this on single targeted events and not dynamic
+  if (!computedAvailableSlots.slots && input.usernameList?.length === 1) {
+    await handleNotificationWhenNoSlots({
```
hariombalhara, line 686: This could potentially delay the response for slots. Not sure by how significant that would be in production.

Worst case, if upstash goes down our availabilities go down.

sean-brydon: Good shout - we can wrap this in a try catch I think that’s possible the best way to solve this

sean-brydon: Ideally this is the type of stuff we put in a background job but we don’t have a way to do this rn

sean-brydon: I don’t think the removing the await is reliable especially in server less right? 

I’ll deffo implement the try catch

packages/trpc/server/routers/viewer/slots/handleNotificationWhenNoSlots.ts
```diff
@@ -0,0 +1,128 @@
...
+        user: eventDetails.username,
+        slug: eventDetails.eventSlug,
+        startTime: eventDetails.startTime.format("YYYY-MM"),
+        // For now navigate here - when impersonation via parameter has been pushed we will impersonate and then navigate to availability
+        editLink: `${WEBAPP_URL}/availability?type=team`,
+      };
+
+      emailsToSend.push(sendOrganizationAdminNoSlotsNotification(payload));
+    }
+    Promise.all(emailsToSend);
```
hariombalhara, line 126: I am not sure if intentional to not await it. But we should add a comment here if that is the case here.

sean-brydon: Youre right - its not intentional - this is still here from playing around trying to get tasker to work

hariombalhara (commented):
PR looks good to me. 

What we need to see though is if the latency added to getSchedule due to this is significant or not.

hariombalhara:
Awesome work @sean-brydon 🙏

joeauyeung (approved):
Giving another reapproval for @hariombalhara

Udit-takkar (approved):
LGTM.

keithwillcode:
@sean-brydon Great PR. One suggestion for a follow-up. 
- Add a test at the util.ts level so we can verify that the logic on https://github.com/calcom/cal.com/pull/14796/files#diff-79280ba0694fc7e34a8239fa701618d6b9b2a91846913f41f0745018cc22ce6bR685 is correct and triggers the function call when desired.

sean-brydon:
> @sean-brydon Great PR. One suggestion for a follow-up.
> 
> * Add a test at the util.ts level so we can verify that the logic on https://github.com/calcom/cal.com/pull/14796/files#diff-79280ba0694fc7e34a8239fa701618d6b9b2a91846913f41f0745018cc22ce6bR685 is correct and triggers the function call when desired.

Great idea - i cant find any tests currently for that file so we'd need to write them before implementing this

shirazdole:
@sean-brydon what a banger, just to double check, if a user within an org is added to an event-type — individual or collective and they don't have a calendar connected, that triggers a notification email to all org owners or admins? Is that programmable in terms of can that be turned on or off? Is there a nice pane to see this within the app outside of the members view which has the long list of apps per user — kind of like a filterable table

PR #14803, opened 2024-04-29
fix: Unbork Team POST API

alishaz-polymath:
## What does this PR do?

Add newly added mandatory fields to partial to allow Team creation using API V1

## Type of change

## How should this be tested?

- Run API V1 locally and create a Team using :https://cal.com/docs/enterprise-features/api/api-reference/teams#create-a-new-team as guide

## Mandatory Tasks

### DIFF
--- a/apps/api/v1/lib/validations/team.ts
+++ b/apps/api/v1/lib/validations/team.ts
@@ -6,6 +6,9 @@ export const schemaTeamBaseBodyParams = Team.omit({ id: true, createdAt: true })
   hideBranding: true,
   metadata: true,
   pendingPayment: true,
+  isOrganization: true,
+  isPlatform: true,
+  smsLockState : true,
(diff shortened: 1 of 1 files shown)


apps/api/v1/lib/validations/team.ts
```diff
@@ -6,6 +6,9 @@ export const schemaTeamBaseBodyParams = Team.omit({ id: true, createdAt: true })
   hideBranding: true,
   metadata: true,
   pendingPayment: true,
+  isOrganization: true,
+  isPlatform: true,
+  smsLockState: true,
```
alishaz-polymath, line 11: Marks these newly added mandatory columns as partial to allow team creation without them as input (they're not relevant for API V1)

emrysal (approved):

PR #14806, opened 2024-04-29
chore: Allow disabling attendee emails if `SMS_ATTENDEE` is a part of the workflow

joeauyeung:
… SMS

## What does this PR do?

Some users want to disable the attendee email if they're already sending an SMS to the attendee. We should show the option to disable attendee default emails if their workflow contains this step.

Fixes # (issue)

## Requirement/Documentation

## Type of change

## How should this be tested?

- Add a workflow step to send an SMS to an attendee
- In the event type the workflow is enabled, the option to disable default emails to attendee should appear

## Mandatory Tasks

## Checklist


CarinaWolli (commented):
Screenshot 2024-04-30 at 9 54 07 AM

Like @ciaranha said, we need to update the text here to: 
At least one workflow is active on this event type that sends a booking confirmation to the attendees.

CarinaWolli (approved):

PR #14809, opened 2024-04-30
fix: Speed up booking time

### LINKED ISSUE #14810: [CAL-3592] Booking timeouts with Outlook Calendar when Team has many members
<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3592](https://linear.app/calcom/issue/CAL-3592/booking-timeouts-with-outlook-calendar-when-team-has-many-members)</sub>

hariombalhara:
## What does this PR do?

Fixes #14810

[Loom](https://www.loom.com/share/fe8c707f6f7741c68c15b5ae442a753f) Take almost same amount of time to book as to load the availability
## Type of changef

## How should this be tested?
- Create a team event with many members (say 3) with all of them connected to Outlook Calendar(or any other Calendar)
- Add a fake delay to `getAvailability` inside `CalendarService`. Something like this would work 
```Javascript
 await new Promise((resolve) => {
      setTimeout(resolve, 1000);
    });
    return [];
```
- Try booking the event. It would take more than 3*10 seconds in main. 
- In this branch it should take 10 seconds. 

## Mandatory Tasks

## Checklist


hariombalhara:
* **#14809** <a href="https://app.graphite.dev/github/pr/calcom/cal.com/14809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/core/getUserAvailability.ts
```diff
@@ -794,3 +794,43 @@ const _getOutOfOfficeDays = async ({
...
+type GetUserAvailabilityQuery = Parameters<typeof getUserAvailability>[0];
+type GetUserAvailabilityInitialData = NonNullable<Parameters<typeof getUserAvailability>[1]>;
+
+const _getUsersAvailability = async ({
+  users,
+  query,
+  initialData,
+}: {
+  users: (NonNullable<GetUserAvailabilityInitialData["user"]> & {
+    currentBookings?: GetUserAvailabilityInitialData["currentBookings"];
```
hariombalhara, line 807: currentBookings are per user so expecting it in user only.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -362,24 +362,25 @@ export async function ensureAvailableUsers(
     });
   }
 
-  for (const user of eventType.users) {
-    const { oooExcludedDateRanges: dateRanges, busy: bufferedBusyTimes } = await getUserAvailability(
-      {
+  (
+    await getUsersAvailability({
```
hariombalhara, line 366: Use function that can operate on `users`

packages/trpc/server/routers/viewer/slots/util.ts
```diff
@@ -456,44 +456,44 @@ export async function getAvailableSlots({ input, ctx }: GetScheduleOptions): Pro
     });
   }
 
+  const users = usersWithCredentials.map((currentUser) => {
+    return {
+      ...currentUser,
+      currentBookings: currentBookingsAllUsers
```
hariombalhara, line 462: Extract currentBookings out here as that is specific to user.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -362,24 +362,25 @@ export async function ensureAvailableUsers(
...
+      initialData: {
         eventType,
         rescheduleUid: input.originalRescheduledBooking?.uid ?? null,
         busyTimesFromLimitsBookings: busyTimesFromLimitsBookingsAllUsers,
-      }
-    );
+      },
+    })
+  ).forEach(({ oooExcludedDateRanges: dateRanges, busy: bufferedBusyTimes }, index) => {
+    const user = eventType.users[index];
```
hariombalhara, line 383: As userAvailabilities are returned in order of users, we could use the index to retrieve the current user from `users`.

packages/trpc/server/routers/viewer/slots/util.ts
sean-brydon, line 462: I wonder if we can use a reduce here to make this even faster - we are filtering and then mapping which is 2 itterations - i think we can make it just one with something like 
(Note this will type error as we'd need to type the array in the reduce function and i didnt spend enough time to figure it out)
```
currentBookingsAllUsers.reduce((accumulator, booking) => {
  if (booking.userId === currentUser.id || booking.attendees?.some((a) => a.email === currentUser.email)) {
    const { attendees: _attendees, ...bookingWithoutAttendees } = booking;
    accumulator.push(bookingWithoutAttendees);
  }
  return accumulator;
}, []);
```

sean-brydon (commented):
I have tested this in a fair few times and havent once ran into any issues - tests are passing. I think this is an amazing thing to ship well done bud!

Will wait for another approval since this seems like it could be high risk @emrysal @CarinaWolli

packages/trpc/server/routers/viewer/slots/util.ts
hariombalhara: I didn't do that because even with 500 users, multiple iterations won't make a noticeable difference.

Plus reduce is less readable than map.

keithwillcode: Yeah great call, @hariombalhara. Scalars are removed in Big-O notation. i.e. O(2n) = O(n)

keithwillcode (approved):
Great change, @hariombalhara.

emrysal (approved):

PR #14835, opened 2024-05-01
chore: Use better buildjet hardware for e2e

keithwillcode:
Swaps out our 4 vCPU server on buildjet for one with 8 vCPU and 32 GB RAM. Want to see how our main e2e test suite performs with better hardware.

This also backs down some hardware on jobs that don't really need it.


.github/workflows/check-types.yml
joeauyeung, line 1: Do we want to slow down our type checking? It feels like this is a really useful step, especially if locally type checking isn't working.

joeauyeung (approved):
Discussed this during consumer weekly call. Feel good approving.

PR #14836, opened 2024-05-01
chore: Removed duplicate production build

keithwillcode:
Removed duplicate production build. Originally, the "Production Build (without database)" build was added to ensure that we could build the web app without the dependency on the database being there. Now that this is the case, the extra build with the database doesn't seem to be required as part of the PR checks. This reduces our time to build, services created, cost, etc.

I also repositioned the environment variables used by the e2e jobs so that we avoid seeing the "Please set NEXTAUTH_SECRET" and "Please set DATABASE_URL" errors when jobs are run out of sync or individually.


.github/actions/cache-build/action.yml
```diff
@@ -14,16 +14,14 @@ runs:
         cache-name: prod-build
         key-1: ${{ inputs.node_version }}-${{ hashFiles('yarn.lock') }}
         key-2: ${{ hashFiles('apps/**/**.[jt]s', 'apps/**/**.[jt]sx', 'packages/**/**.[jt]s', 'packages/**/**.[jt]sx', '!**/node_modules') }}
-        key-3: ${{ github.event.pull_request.number || github.ref }}
```
keithwillcode, line 17: @zomars @emrysal Did this absolutely need to be here? Will revert if so. Was just thinking that since we already use the `sha`, this might not be necessary.

.github/workflows/api-v1-production-build.yml
```diff
@@ -1,4 +1,4 @@
-name: Production Build
+name: Production Builds
```
keithwillcode, line 1: Changed this name so all of the production builds we run get lumped into the same group visually.

.github/workflows/e2e-app-store.yml
```diff
@@ -3,6 +3,41 @@ on:
   workflow_call:
 env:
   NODE_OPTIONS: --max-old-space-size=4096
+  ALLOWED_HOSTNAMES: ${{ vars.CI_ALLOWED_HOSTNAMES }}
```
keithwillcode, line 6: Putting these vars at the top of the file helps with making sure child jobs always have them set

.github/workflows/production-build-without-database.yml
```diff
@@ -37,8 +36,8 @@ env:
 
 jobs:
   build:
-    name: Build
-    runs-on: ubuntu-latest
+    name: Web App
+    runs-on: buildjet-4vcpu-ubuntu-2204
```
keithwillcode, line 40: This was causing cache corruption problems downstream

zomars (approved):
Always love a good cleanup! Nice find!

Ship it

.github/actions/cache-build/action.yml
keithwillcode: Turns out that github.sha here is the current commit on `main` and not the branch that's attempting to be merged in, causing bad caching issues and CI failures when in main.

keithwillcode: Fix is here. https://github.com/calcom/cal.com/pull/14932

PR #14846, opened 2024-05-01
fix: Sender ID info not shown on hover

### LINKED ISSUE #14842: [CAL-3602] Sender ID info not shown on hover
Hovering over the info icon doesn't show sender ID explanation. Also, should be in line with the text before it.

It should show the translation for sender_id_info:

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3602](https://linear.app/calcom/issue/CAL-3602/sender-id-info-not-shown-on-hover)</sub>

Amit91848:
## What does this PR do?

Icon directly renders svg which doesn't work too well with Tooltip, wrapped it around a span instead.

Fixes #14842 

Screenshot 2024-05-01 at 9 01 06 PM

## Type of change

## Mandatory Tasks


CarinaWolli (approved):
Thank you for fixing this 🙏

shirazdole:
@sean-brydon what a banger, just to double check, if a user within an org is added to an event-type — individual or collective and they don't have a calendar connected, that triggers a notification email to all org owners or admins? Is that programmable in terms of can that be turned on or off? Is there a nice pane to see this within the app outside of the members view which has the long list of apps per user — kind of like a filterable table

PR #14852, opened 2024-05-02
build: Use "standalone" output to build Next.js as a standalone app

jckw:
## What does this PR do?

This PR changes the Next.js output type of the @calcom/web app to "standalone", which is needed if https://github.com/calcom/docker/pull/355  is to merge.

The ["output" config property](https://nextjs.org/docs/pages/api-reference/next-config-js/output) should not affect builds being run on Vercel, but will affect those built with Docker.

Please see the linked PR for a full explanation of why this is useful. The summary is:

- "standalone" results in a 40% reduction in image size
- "standalone" is the standard config property used in [the Next.js example Dockerfile](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile#L65-L66), so allows the calcom/docker repo to use a Dockerfile similar to that, thereby requiring less maintenance

## Requirement/Documentation

- Most of the details are here: https://github.com/calcom/docker/pull/355

## Type of change

- This isn't really a breaking change unless someone is running "next build" outside of Vercel and their PaaS cannot handle "standalone" output or they are managing the build output themselves. This seems unlikely though.
  - The exception is Docker. This PR will break docker builds unless the linked PR is merged. But.. Docker builds have been broken for 2+ months anyway, so nbd.

## How should this be tested?

- What is expected (happy path) to have (input and output)? **Everything works exactly as it does now**

Test out this branch on whatever build process is used by Cal (presumably Vercel), and verify that it works and deploys. Unless there is any special config used by Cal, everything should work as normal on Vercel.

## Mandatory Tasks

## Checklist

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -158,6 +158,7 @@ const matcherConfigUserTypeEmbedRoute = {
 
 /** @type {import("next").NextConfig} */
 const nextConfig = {
+  output: "standalone",
   experimental: {
     // externalize server-side node_modules with size > 1mb, to improve dev mode performance/RAM usage
     serverComponentsExternalPackages: ["next-i18next"],


emrysal (approved):
Happy to give this a shot, if this does create issues (which will be caught by our checks) we should consider an environment variable like so:

output: process.env.BUILD_STANDALONE === "true" ? "standalone" : undefined,

emrysal:
Hi @jckw I'm unable to update this branch (critical to do so as the checks will not run properly otherwise) - will close and re-open a mirror.

PR #14854, opened 2024-05-02
fix: date override modal

sean-brydon:
CleanShot 2024-05-02 at 11 20 57@2x
CleanShot 2024-05-02 at 11 20 50@2x
CleanShot 2024-05-02 at 11 20 46@2x


joeauyeung (approved):
Tested on different screen sizes and compared to what's on `main`. LGTM. Thanks for fixing this so quickly @sean-brydon!

PR #14856, opened 2024-05-02
chore: don't try to increase usage for legacy plans

exception:
## What does this PR do?

Some platform plans are not using metered usage for now, so we need to not attempt to increase or Stripe will error out.


apps/api/v2/src/modules/billing/services/billing.service.ts
```diff
@@ -99,22 +99,36 @@ export class BillingService {
...
-      quantity: 1,
-      timestamp: "now",
-    });
+      const stripeSubscription = await this.stripeService.stripe.subscriptions.retrieve(
+        billingSubscription.subscriptionId
+      );
+      const item = stripeSubscription.items.data[0];
+      // legacy plans are licensed, we cannot create usage records against them
+      if (item.price?.recurring?.usage_type === "licensed") {
+        return void 0;
```
supalarry, line 117: I think it's more Javascript way to write empty `return;` than `return void 0`

supalarry (approved):

PR #14860, opened 2024-05-02
feat: Add Organization Admin scope to API V1

### LINKED ISSUE #14974: [CAL-3677] Add Organization Admin Scope for API V1
Add the ability for the organization owners or admins to access their member resources via API V1

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3677](https://linear.app/calcom/issue/CAL-3677/add-organization-admin-scope-for-api-v1)</sub>

alishaz-polymath:
## What does this PR do?

- Updates the admin guard `isAdmin` → `isSystemWideAdmin` for added clarity by clarifying scope of access to be system wide
- Updates the admin guard to consider Organization owners or admins with `isOrganizationOwnerOrAdmin` and the scope of access is the members of the organization
- Implements `isOrganizationOwnerOrAdmin` on bookings endpoint as an example:
  - /bookings GET
  - /bookings POST
  - /bookings/[id] PATCH
  - /bookings/[id]/cancel DELETE 
  - /bookings/[id] GET

This PR also adds Pagination to the bookings endpoint, accessible by passing `take` & `page` parameters, where take represents the number of items in the resulting array and page represents the page number when taking 'take' number of items per page.

For example, if there are 20 results, we can use take=5 and page=3 to get results 11-15.

- Fixes #14974 
- Fixes CAL-3677

- [x] I have added a Docs issue [here](https://github.com/calcom/docs/issues/63) if this PR makes changes that would require a [documentation change](https://docs.cal.com)


apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,123 @@
...
+  const userMemberships = await getAllMemberships(memberUserIds);
+  const adminMemberships = await getAllAdminMemberships(adminUserId);
+
+  const accessibleUserIds = new Set<number>();
+
+  const adminOrganizationId = adminMemberships.filter((m) => m.team.isOrganization).map((m) => m.team.id)[0];
+  if (adminOrganizationId) {
+    userMemberships
+      .filter((um) => um.team.id === adminOrganizationId)
+      .forEach((um) => accessibleUserIds.add(um.userId));
```
joeauyeung, line 80: NIT: I don't like chaining array methods together. If we use reduce here we can bring this from `O(2n)` down to `O(n)`

apps/api/v1/pages/api/bookings/[id]/_patch.ts
```diff
@@ -109,14 +110,27 @@ export async function patchHandler(req: NextApiRequest) {
...
     throw new HttpError({
       statusCode: 403,
       message: "Only admin can change the organizer of a booking",
     });
   }
+
+  if (body.userId && isOrganizationOwnerOrAdmin) {
+    const accessibleUsersIds = await getAccessibleUsers({
+      adminUserId: userId,
+      memberUserIds: [body.userId],
```
Comment (joeauyeung) [inline apps/api/v1/pages/api/bookings/[id]/_patch.ts:125]: Are we calling this in the endpoint and in the `auth-middleware`? This is starting to make me concerned about the performance for bigger resources.

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+export const retrieveOrgScopedAccessibleUsers = async ({ adminId }: { adminId: number }) => {
+  const adminMemberships = await getAllAdminMemberships(adminId);
+  const adminOrganizationId = adminMemberships.filter((m) => m.team.isOrganization).map((m) => m.team.id)[0];
+  if (adminOrganizationId) {
+    const teamsInOrganization = await getTeamsInOrganization(adminOrganizationId);
+    const teamIds = teamsInOrganization.map((team) => team.id);
+
+    const allMembershipsInteams = await getAllUsersInTeams(teamIds);
```
joeauyeung, line 94:
```suggestion
    const allMembershipsInTeams = await getAllUsersInTeams(teamIds);
```

joeauyeung (commented):
This is awesome @alishaz-polymath.

Left a few comments that I don't think are blocking. Personally I'm not a huge fan of querying ids and then making an additional query to get the records of those ids. Although I see why you did it.

apps/api/v1/pages/api/api-keys/[id]/_auth-middleware.ts
```diff
@@ -6,10 +6,10 @@ import prisma from "@calcom/prisma";
 import { schemaQueryIdAsString } from "~/lib/validations/shared/queryIdString";
 
 export async function authMiddleware(req: NextApiRequest) {
-  const { userId, isAdmin } = req;
+  const { userId, isSystemWideAdmin } = req;
   const { id } = schemaQueryIdAsString.parse(req.query);
   // Admin can check any api key
-  if (isAdmin) return;
+  if (isSystemWideAdmin) return;
```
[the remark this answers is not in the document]
Comment (alishaz-polymath) [reply apps/api/v1/pages/api/api-keys/[id]/_auth-middleware.ts:12]: isSystemWideAdmin (previously isAdmin) figured out as follows:
1. we extract userId associated with the apiKey
2. we make a call to check the role of the user with the userId
`if (userRole === UserPermissionRole.ADMIN) return { isAdmin: true, scope: ScopeOfAdmin.SystemWide };`

ref: `\apps\api\v1\lib\utils\isAdmin.ts`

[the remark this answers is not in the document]
Comment (alishaz-polymath) [reply apps/api/v1/pages/api/api-keys/[id]/_auth-middleware.ts:12]: So technically, the apiKey is used to find the user, and if the user is Admin, the scope of the apiKey elevates.

Just FYI, this logic hasn't changed much, just the name around it 😄 It needed more clarity given that we were going to have organization level admin as well, so we needed to differentiate between system wide and organization admins

apps/api/v1/pages/api/bookings/[id]/_patch.ts
Comment (alishaz-polymath) [reply apps/api/v1/pages/api/bookings/[id]/_patch.ts:125]: It is generally not done in middleware because not each endpoint would require this, so if we put this in the middleware, it would make queries even when it doesn't need to. 
Two queries are made, one to get member's memberships (for all users provided, one if only one) and admin memberships (always one). I could think of getting both in the same query but I wonder if the performance difference would be this big, as it might be at the expense of clarity and scalability of the code itself. 

Can you elaborate, what is your concern exactly in terms of performance for bigger resources here?

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
alishaz-polymath: Nice catch!

apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts
```diff
@@ -3,15 +3,34 @@ import type { NextApiRequest } from "next";
...
+      select: { userId: true },
+    });
+    if (booking) {
+      const bookingUserId = booking.userId;
+      if (bookingUserId) {
+        const accessibleUsersIds = await getAccessibleUsers({
+          adminUserId: userId,
+          memberUserIds: [bookingUserId],
+        });
+        if (accessibleUsersIds.length > 0) return;
```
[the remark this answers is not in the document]
Comment (alishaz-polymath) [reply apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts:29]: Why not though? Do you have alternative recommendation?

apps/api/v1/lib/utils/isAdmin.ts
```diff
@@ -1,10 +1,44 @@
...
+          isOrganization: true,
+        },
+      },
+    },
+  });
+  if (!ownerOrAdminMemberships.length) return { isAdmin: false, scope: null };
+
+  const organizations = ownerOrAdminMemberships.filter(
+    (membership) => membership.team.isOrganization === true
+  );
```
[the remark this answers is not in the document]
alishaz-polymath: Good call! 🙌

apps/api/v1/pages/api/bookings/[id]/_patch.ts
Comment (alishaz-polymath) [reply apps/api/v1/pages/api/bookings/[id]/_patch.ts:125]: What I think I should do here though, is perhaps reduce the number of queries made. I think if I merge the two DB queries and then only perform meaningful data extraction for member and admin memberships, that'll reduce the overhead. 👀

Comment (alishaz-polymath) [reply apps/api/v1/pages/api/bookings/[id]/_patch.ts:125]: > Are we calling this in the endpoint and in the auth-middleware?

We only call this when necessary and it isn't called more than once per request, be it `/bookings POST`, `/bookings GET`. Then, since for the specific booking ID paths, admin checks happen in middleware, it is done only in the `auth-middleware`, and **not** the subsequent endpoints(`/bookings/[id] — PATCH/DELETE/GET`). I hope that clears it a bit

apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts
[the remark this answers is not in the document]
Comment (joeauyeung) [reply apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts:29]: I was reading this commit and this is what I was thinking. You're already one step ahead.

[`dd46291` (#14860)](https://github.com/calcom/cal.com/pull/14860/commits/dd46291c99fc2073707bc9d38c2cda5eaec0ceb6)

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
alishaz-polymath: Nice find. Resolved!

apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts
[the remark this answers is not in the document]
Comment (alishaz-polymath) [reply apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts:29]: Oh awesome! Yeah, thanks for the shout, I thought over it again on how it could be improved ❤️

apps/api/v1/lib/utils/isAdmin.ts
```diff
@@ -1,10 +1,39 @@
 import type { NextApiRequest } from "next";
 
 import prisma from "@calcom/prisma";
-import { UserPermissionRole } from "@calcom/prisma/enums";
+import { UserPermissionRole, MembershipRole } from "@calcom/prisma/enums";
+
+export const ScopeOfAdmin = {
+  SystemWide: "SystemWide",
+  OrgOwnerOrAdmin: "OrgOwnerOrAdmin",
+} as const;
```
alishaz-polymath, line 9: Currently we have decided to go with System-Wide admins and Organization owner or admin, leaving out team admins due to a security concern.

apps/api/v1/lib/utils/isAdmin.ts
```diff
@@ -1,10 +1,39 @@
...
+    },
+    select: {
+      team: {
+        select: {
+          id: true,
+          isOrganization: true,
+        },
+      },
+    },
+  });
```
alishaz-polymath, line 35: Only fetching from memberships where `isOrganization` within teams is true.

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,90 @@
...
+  return accessibleUserIds;
+};
+
+export const retrieveOrgScopedAccessibleUsers = async ({ adminId }: { adminId: number }) => {
+  const adminMemberships = await getAllAdminMemberships(adminId);
+  const organizationId = adminMemberships.find((membership) => membership.team.isOrganization)?.team.id;
+  if (organizationId) {
+    const allMemberships = await getAllOrganizationMembers(organizationId);
+    return allMemberships.map((membership) => membership.userId);
+  }
```
alishaz-polymath, line 88: Since a user MUST be a part of the org, before they are part of any teams within the org; so once we have orgId, we can just check if the user has membership with that orgId

apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts
```diff
@@ -3,15 +3,33 @@ import type { NextApiRequest } from "next";
...
+import { getAccessibleUsers } from "~/lib/utils/retrieveScopedAccessibleUsers";
 import { schemaQueryIdParseInt } from "~/lib/validations/shared/queryIdTransformParseInt";
 
 async function authMiddleware(req: NextApiRequest) {
-  const { userId, isAdmin, query } = req;
-  if (isAdmin) {
+  const { userId, isSystemWideAdmin, isOrganizationOwnerOrAdmin, query } = req;
+  if (isSystemWideAdmin) {
     return;
   }
```
Comment (alishaz-polymath) [inline apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts:13]: No need to have further checks here, since if is SystemWideAdmin, they have access to all users.

apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts
```diff
@@ -3,15 +3,33 @@ import type { NextApiRequest } from "next";
...
+      const bookingUserId = booking.userId;
+      if (bookingUserId) {
+        const accessibleUsersIds = await getAccessibleUsers({
+          adminUserId: userId,
+          memberUserIds: [bookingUserId],
+        });
+        if (accessibleUsersIds.length > 0) return;
+      }
+    }
+  }
```
Comment (alishaz-polymath) [inline apps/api/v1/pages/api/bookings/[id]/_auth-middleware.ts:31]: Here we ensure that the user to whom the booking id belongs is an accessible member of the organization, if they are, no further checks are needed

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,90 @@
...
+      role: true,
+      teamId: true,
+    },
+  });
+
+  const orgId = memberships.find((membership) => membership.userId === adminUserId)?.teamId;
+  if (!orgId) return [];
+
+  const allAccessibleMemberUserIds = await getAllMemberships(memberships, orgId);
+  const accessibleUserIds = allAccessibleMemberUserIds.filter((userId) => userId !== adminUserId);
```
alishaz-polymath, line 78: We don't want adminUserId to piggyback here, so we remove it.

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,90 @@
...
+        { userId: { in: memberUserIds } },
+        { userId: adminUserId, role: { in: [MembershipRole.OWNER, MembershipRole.ADMIN] } },
+      ],
+    },
+    select: {
+      userId: true,
+      role: true,
+      teamId: true,
+    },
+  });
```
alishaz-polymath, line 72: I initially had separate calls for user memberships & admin memberships, but this way we could do it in one DB call with the only downside to have to remove the admin userID from the memberships return further down.

alishaz-polymath (commented):
Self review added

apps/api/v1/lib/utils/isAdmin.ts
```diff
@@ -1,10 +1,39 @@
...
+  SystemWide: "SystemWide",
+  OrgOwnerOrAdmin: "OrgOwnerOrAdmin",
+} as const;
 
 export const isAdminGuard = async (req: NextApiRequest) => {
   const { userId } = req;
-  const user = await prisma.user.findUnique({ where: { id: userId } });
-  return user?.role === UserPermissionRole.ADMIN;
+  const user = await prisma.user.findUnique({ where: { id: userId }, select: { role: true } });
+  if (!user) return { isAdmin: false, scope: null };
```
Udit-takkar, line 14: should we throw error here? user not found

alishaz-polymath: It's here just to satisfy Typescript really, it will pretty much never happen as `userId` is fetched from the API key in the `verifyApiKey` and it returns not found error if APIKey is invalid. 
image

isAdmin is called after that check, so it would pretty much never satisfy !user here, ever.

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
Udit-takkar, line 88: NIT: we can early return 

if(!organizationId) return [];

const allMemberships = await getAllOrganizationMembers(organizationId);
return allMemberships.map((membership) => membership.userId);

alishaz-polymath: Yeah nice find!

alishaz-polymath: Fresh pair of eyes are always great ❤️

apps/api/v1/pages/api/teams/[teamId]/_auth-middleware.ts
```diff
@@ -22,13 +22,13 @@ export async function checkPermissions(
...
   const { teamId } = schemaQueryTeamId.parse({
     teamId: req.query.teamId,
     version: req.query.version,
     apiKey: req.query.apiKey,
   });
-  return canUserAccessTeamWithRole(userId, isAdmin, teamId, role);
+  return canUserAccessTeamWithRole(userId, isSystemWideAdmin, teamId, role);
 }
 
 export async function canUserAccessTeamWithRole(
```
Comment (Udit-takkar) [inline apps/api/v1/pages/api/teams/[teamId]/_auth-middleware.ts:34]: should we change second param inside canUserAccessTeamWithRole from `isAdmin` to `isSystemWideAdmin`

Comment (alishaz-polymath) [reply apps/api/v1/pages/api/teams/[teamId]/_auth-middleware.ts:34]: I don't think it is breaking, but we probably should for clarity and consistency!

apps/api/v1/pages/api/event-types/_post.ts
```diff
@@ -311,18 +311,18 @@ async function postHandler(req: NextApiRequest) {
 }
 
 async function checkPermissions(req: NextApiRequest) {
-  const { isAdmin } = req;
+  const { isSystemWideAdmin } = req;
   const body = schemaEventTypeCreateBodyParams.parse(req.body);
   /* Non-admin users can only create event types for themselves */
-  if (!isAdmin && body.userId)
+  if (!isSystemWideAdmin && body.userId)
```
Udit-takkar, line 317: I tried creating an event type for org user and it is not showing up in /event-types

Screenshot 2024-05-10 at 2 28 36 PM
Screenshot 2024-05-10 at 2 28 52 PM

Udit-takkar: The newly created event Type does not contain `Profile` column in event type

alishaz-polymath: Org Admin is not yet implemented for event-types. It is only implemented for /bookings endpoints as a starting point

apps/api/v1/pages/api/event-types/[id]/_patch.ts
```diff
@@ -237,9 +237,9 @@ export async function patchHandler(req: NextApiRequest) {
 }
 
 async function checkPermissions(req: NextApiRequest, body: z.infer<typeof schemaEventTypeBaseBodyParams>) {
-  const { userId, isAdmin } = req;
+  const { userId, isSystemWideAdmin } = req;
   const { id } = schemaQueryIdParseInt.parse(req.query);
-  if (isAdmin) return;
+  if (isSystemWideAdmin) return;
```
Comment (Udit-takkar) [inline apps/api/v1/pages/api/event-types/[id]/_patch.ts:242]: the logic inside `checkTeamEventEditPermission` only checks if it's a team event then user must be admin or owner of the team so org admin or owner not part of the team cannot edit event type.

Comment (alishaz-polymath) [reply apps/api/v1/pages/api/event-types/[id]/_patch.ts:242]: Event type logic hasn't been changed yet. It is the same as it was, and only considers system-wide admin.

REF:  image

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,90 @@
+import prisma from "@calcom/prisma";
+import { MembershipRole } from "@calcom/prisma/enums";
+
+type AccessibleUsersType = {
+  memberUserIds: number[];
+  adminUserId: number;
+};
+
+const getAllMemberships = async (
```
Udit-takkar, line 9: NIT: probably a better name like `getAllOrgMemberships`

alishaz-polymath:
@Udit-takkar all suggestions sorted 🙏

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,89 @@
+import prisma from "@calcom/prisma";
```
keithwillcode, line 1: Do we not have these functions already written somewhere in the core libraries of the code base that we could reuse?

keithwillcode (commented):
Code changes look really good, @alishaz-polymath. Nice work!

We just need to do some heavy testing before shipping this out.

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
alishaz-polymath: Let me look into it. These functions evolved in the past few commits so it is possible this they're already in the repo in the current state. 🙏

joeauyeung:
> Code changes look really good, @alishaz-polymath. Nice work!
> 
> We just need to do some heavy testing before shipping this out.

I see that you updated some tests. It would be great to see some new tests around determining the scope of admin permissions and access around it.

alishaz-polymath:
> > Code changes look really good, @alishaz-polymath. Nice work!
> > We just need to do some heavy testing before shipping this out.
> 
> I see that you updated some tests. It would be great to see some new tests around determining the scope of admin permissions and access around it.

Yeah I did discuss this with Keith but the thing is, we plan to sunset V1 in favour of V2 fairly soon. I wonder if adding a test suite is worth it.  I can add test for sure, not certain though if that's worth it in the grand scheme of things on this occasion

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,89 @@
...
+    },
+  });
+};
+
+export const getAccessibleUsers = async ({
+  memberUserIds,
+  adminUserId,
+}: AccessibleUsersType): Promise<number[]> => {
+  const memberships = await prisma.membership.findMany({
+    where: {
```
CarinaWolli, line 58: we also need to check if membership is accepted, we need to do that in all other places as well where we query memberships

CarinaWolli (dismissed):
We should also check if the memberships are accepted. Everything else looked good 🙏

zomars (changes requested):
I saw that we're planning to sunset this. Although I'm not sure if by sunsetting we mean removing it completely or just stop doing updates. Either way. We're touching a lot of files and adding no new tests. I'm not confortable shipping in the current state. Tests are also to prove that something works right now and it won't break unexpectedly. Even if we're sunsetting. Tests are transferable. I would try to add tests anyways.

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
alishaz-polymath: Done!

apps/api/v1/lib/utils/retrieveScopedAccessibleUsers.ts
```diff
@@ -0,0 +1,92 @@
...
+      team: {
+        isOrganization: true,
+        accepted: true,
+      },
```
CarinaWolli, line 64:
```suggestion
      },
      accepted: true,
```

.github/workflows/integration.yml
zomars, line 1: Can we PR this first in a separate one?

keithwillcode: @zomars Yeah sure thing. Was just testing here as part of writing these new tests

keithwillcode: Might be trickier to actually test without having tests though

sean-brydon:
@keithwillcode tests look good - is there any reason this is still in draft? Do we need to cover a few more edge cases some where

apps/api/v1/test/lib/bookings/_auth-middleware.integration-test.ts
```diff
@@ -0,0 +1,93 @@
...
+import prisma from "@calcom/prisma";
+
+import authMiddleware from "../../../pages/api/bookings/[id]/_auth-middleware";
+
+type CustomNextApiRequest = NextApiRequest & Request;
+type CustomNextApiResponse = NextApiResponse & Response;
+
+describe("Bookings auth middleware", () => {
+  it("Returns 403 when user has no permission to the booking", async () => {
+    const trialUser = await prisma.user.findFirstOrThrow({ where: { email: "trial@example.com" } });
```
keithwillcode, line 16: Note: ideally we don’t directly use prisma here. Will check to see if a reusable repository is available. 

Note: instead of hardcoding the emails of users, would like to build out utils that defined these built-in test users.

apps/api/v1/test/lib/bookings/_patch.integration-test.ts
```diff
@@ -0,0 +1,88 @@
...
+  it("Returns 403 when user has no permission to the booking", async () => {
+    const memberUser = await prisma.user.findFirstOrThrow({ where: { email: "member2-acme@example.com" } });
+    const proUser = await prisma.user.findFirstOrThrow({ where: { email: "pro@example.com" } });
+    const booking = await prisma.booking.findFirstOrThrow({ where: { userId: proUser.id } });
+
+    const { req, res } = createMocks<CustomNextApiRequest, CustomNextApiResponse>({
+      method: "PATCH",
+      body: {
+        title: booking.title,
+        startTime: booking.startTime.toISOString(),
```
keithwillcode, line 23: Patching with the same data for now because we are testing permissions. 

Will add test to test the patching functionality itself.

keithwillcode:
@alishaz-polymath @zomars Update here: I've added a bunch of unit and integration tests here. All pass locally except 1 integration test that I think is due to a change in functionality that I'm not totally sure is correct. @alishaz-polymath is going to follow up on it.

What I'm going to focus on now is making sure our integration test suite can properly run on CI. Right now, when I run it, the tests pass but then almost all downstream E2E jobs fail so seems like the integration test suite is corrupting something.

keithwillcode:
Ok - CI has been updated and is now correctly running all tests. See https://github.com/calcom/cal.com/actions/runs/9132154000. The 1 failing integration as mentioned can be seen there.

.github/workflows/pr.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     secrets: inherit
 
   test:
-    name: Unit tests
+    name: Tests
```
keithwillcode, line 39: Screenshot 2024-05-17 at 3 02 56 PM

Naming all of these `Tests` results in the following grouping, which is nicer.

alishaz-polymath:
> @alishaz-polymath @zomars Update here: I've added a bunch of unit and integration tests here. All pass locally except 1 integration test that I think is due to a change in functionality that I'm not totally sure is correct. @alishaz-polymath is going to follow up on it.
> 
> What I'm going to focus on now is making sure our integration test suite can properly run on CI. Right now, when I run it, the tests pass but then almost all downstream E2E jobs fail so seems like the integration test suite is corrupting something.

So I had a look and basically the system-wide admin functionality hasn't been changed. As you can see [here](https://github.com/calcom/cal.com/blob/main/apps/api/v1/pages/api/bookings/_get.ts#L185-L199) the system-admin check is only supposed to function when userId is passed along with it. Now I'm not entirely sure about the reasoning behind this, but I think with pagination, and continuing to allow this as well, we could very well do the same for systemwide admins as we do with org admins here as well. 
What are your thoughts? @keithwillcode

apps/api/v1/test/lib/bookings/_get.integration-test.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  it("Returns bookings for all users when accessed by system-wide admin", async () => {
+    const adminUser = await prisma.user.findFirstOrThrow({ where: { email: "owner1-acme@example.com" } });
+    const { req } = createMocks<CustomNextApiRequest, CustomNextApiResponse>({
+      method: "GET",
+      pagination: {
+        take: 100,
+        skip: 0,
+      },
```
alishaz-polymath, line 81: We need to take more bookings here because the first many bookings that are returned belong to the system-wide admin, which is why it was failing when we had pagination at 10. Once we increase it to a bigger size like 100 in this case, it passes as there are bookings for other users as well, and so the logic is working as expected.

keithwillcode: oh wow. nice.

keithwillcode:
@alishaz-polymath @zomars Peeling out the integration tests skeleton to [here](https://github.com/calcom/cal.com/pull/15148) so it's separated. We've already verified that all new integration tests added in this PR pass.

vitest.workspace.ts
```diff
@@ -36,7 +36,24 @@ const workspaces = packagedEmbedTestsOnly
       },
     ]
   : // It doesn't seem to be possible to fake timezone per test, so we rerun the entire suite with different TZ. See https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286
-  timeZoneDependentTestsOnly
+  integrationTestsOnly
```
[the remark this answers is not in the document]
zomars: Will rebase

zomars (approved):

PR #14871, opened 2024-05-03
fix: buffer times in handleNewBooking

### LINKED ISSUE #14868: [CAL-3631] Buffer times not respected when URL param is changed
When changing the param with the slot time in the URL, the buffer is not taken into account. This round-robin event has a 1 hour before event buffer, as you can see that’s not the case here (see screenshot). The slots didn’t show as available, but I was still able to book them

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3631](https://linear.app/calcom/issue/CAL-3631/buffer-times-not-respected-when-url-param-is-changed)</sub>

CarinaWolli:
## What does this PR do?

Fixes the issue that bookings could be made without respecting before or after buffers of other existing bookings. Slots are shown correctly, but when you changed the slot URL param to a time that is within a buffer `handleNewBooking` didn't throw an error and instead, the booking was successful.

Fixes #14868
Fixes https://linear.app/calcom/issue/CAL-3631/buffer-times-not-respected-when-url-param-is-changed


Udit-takkar (commented):
I added a before event buffer of 60 minute in this event type and booked 3:00 PM - 3:15 PM  Slot on 10th 

Screenshot 2024-05-07 at 1 27 04 PM

~but the slots 3:15. 3:30, 3:45, ... are missing. is this expected behaviour?~

Screenshot 2024-05-07 at 1 30 22 PM

Edit: Ignore this comment.

sean-brydon:
> I added a before event buffer of 60 minute in this event type and booked 3:00 PM - 3:15 PM Slot on 10th
> 
> Screenshot 2024-05-07 at 1 27 04 PM
> but the slots 3:15. 3:30, 3:45, ... are missing. is this expected behaviour?
> 
> Screenshot 2024-05-07 at 1 30 22 PM

I think this is right? You need 60 minutes BEFORE your next event so thus 3:00-3:15 should block till 4:15 which is what i think this is doing.

packages/core/getBusyTimes.ts
```diff
@@ -85,10 +87,15 @@ export async function getBusyTimes(params: {
   const endTimeDate =
     rescheduleUid && duration ? dayjs(endTime).add(duration, "minute").toDate() : new Date(endTime);
 
+  // to also get bookings that are outside of start and end time, but the buffer falls within the start and end time
+  const maxBuffer = getDefinedBufferTimes()[getDefinedBufferTimes().length - 1];
```
sean-brydon, line 91: NIT - saves calling this twice (Very small nit its lit a function returning a very small array its not expensive)
```suggestion
const definedBufferTimes = getDefinedBufferTimes()
  const maxBuffer = definedBufferTimes[definedBufferTimes.length - 1];
```

sean-brydon (commented):
Will wait for @Udit-takkar's response but i think this is working as expected

CarinaWolli:
> > I added a before event buffer of 60 minute in this event type and booked 3:00 PM - 3:15 PM Slot on 10th
> > Screenshot 2024-05-07 at 1 27 04 PM
> > but the slots 3:15. 3:30, 3:45, ... are missing. is this expected behaviour?
> > Screenshot 2024-05-07 at 1 30 22 PM
> 
> I think this is right? You need 60 minutes BEFORE your next event so thus 3:00-3:15 should block till 4:15 which is what i think this is doing.

Exactly!

Udit-takkar (approved):
LGTM

PR #14900, opened 2024-05-06
chore: Use 4vCPU BuildJet hardware for type checks

keithwillcode:
In #14835, I backed down the vCPUs for type-checks since that job runs in parallel to everything. This is apparently not compatible with our type checks. Putting back for now to fix.


alishaz-polymath (approved):

Udit-takkar:
@keithwillcode type checks were failing. i have re triggered failed jobs

keithwillcode:
Force merging here because the type checks are broken in main, which will be fixed by this PR.

PR #14905, opened 2024-05-06
fix: [CAL-3578] [CAL-2733] Zoho calendar issues 

### LINKED ISSUE #12405: [CAL-2733] Zoho calendar issues

vachmara:
## What does this PR do?

This PR addresses a bug identified in issue #12405 and builds upon the initial work done in PR #13739 by @Pritam-sd-dev.
fixes #12405 
/claim #12405 

The existing implementation using the [free/busy](https://www.zoho.com/calendar/help/api/get-user-freebusy-details.html) endpoint was found insufficient for accurate availability checks. I've enhanced the `getAvailability` function to include an event check within the specified time range to ensure more reliable results.

## Type of change

## How should this be tested?

The fix was manually tested by adding multiple events to my Zoho calendar and verifying whether these were correctly reflected in the calendar event schedule. Future updates should include automated tests to cover these scenarios.

## Mandatory Tasks

## Checklist


PeerRich (dismissed):

packages/app-store/_utils/updateAppServerLocation.ts
```diff
@@ -0,0 +1,15 @@
+import prisma from "@calcom/prisma";
+
+import { appKeysSchema as zohoKeysSchema } from "../zohocalendar/zod";
+
+async function updateAppServerLocation(slug: string, serverUrl: string) {
+  const app = await prisma.app.findUnique({ where: { slug } });
+  const { client_id, client_secret } = zohoKeysSchema.parse(app?.keys) || {};
+  const updatedKeys = { client_id, client_secret, server_location: serverUrl };
+  await prisma.app.update({
```
joeauyeung, line 9: We shouldn't write the server location to the app data as this will affect all users. We should be writing the location as a part of the user's Zoho credentials

packages/app-store/zohocalendar/api/callback.ts
```diff
@@ -46,14 +54,14 @@ async function getHandler(req: NextApiRequest, res: NextApiResponse) {
 
   const query = stringify(params);
 
-  const response = await fetch(`${OAUTH_BASE_URL}/token?${query}`, {
+  const response = await fetch(`${getOAuthBaseUrl(server_location || "com")}/token?${query}`, {
```
joeauyeung, line 57: We need to handle when "US" is returned as this will add "us" to the domain when it should be "com"

packages/app-store/zohocalendar/lib/CalendarService.ts
```diff
@@ -263,6 +265,37 @@ export default class ZohoCalendarService implements Calendar {
     );
   }
 
+  private async getUnavailability(
```
joeauyeung, line 268: Where did this method come from?

joeauyeung (changes requested):
Thanks for taking this up @vachmara. This is much needed. This still needs some work but we're on the right track.

vachmara:
@joeauyeung on it!

packages/app-store/zohocalendar/lib/CalendarService.ts
vachmara: As mention on the PR the existing implementation using the [free/busy](https://www.zoho.com/calendar/help/api/get-user-freebusy-details.html) endpoint was found insufficient for accurate availability checks:
 
- To use this endpoint, we need to enable the `free/busy` functionnality [on settings](Steps_to_enable_or_disable_free_busy_consideration)
- To correctly fetch busy events, you also need to [activate this functionality](https://help.zoho.com/portal/en/kb/calendar/create-and-manage-events/articles/free-busy-sharing#Steps_to_include_a_calendar_to_your_free_busy_schedule) for each event created. This is a complex mechanism.

Therefore, I've enhanced the `getAvailability` function to include checks for all events without differentiating between free and busy statuses.

joeauyeung (changes requested):
Tested this with a VPN to EU servers and this works as expected. I think we should include private events when checking for availabilities and then we should be good to merge this. 

As a follow up I think we have to do this for other Zoho apps right?

packages/app-store/zohocalendar/lib/CalendarService.ts
```diff
@@ -263,6 +262,37 @@ export default class ZohoCalendarService implements Calendar {
...
+      // List all events within the range
+      const response = await this.fetcher(`/calendars/${calendarId}/events?${query}`);
+      const data = await this.handleData(response, this.log);
+
+      // Check for no data scenario
+      if (!data.events || data.events.length === 0) return [];
+
+      return (
+        data.events
+          .filter((event: any) => event.isprivate === false)
```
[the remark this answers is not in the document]
vachmara: Exactly @joeauyeung!

vachmara:
> Tested this with a VPN to EU servers and this works as expected. I think we should include private events when checking for availabilities and then we should be good to merge this.
> 
> As a follow up I think we have to do this for other Zoho apps right?

The OAuth flow indeed needs to be implemented in the two other apps. However, the `getAvailability` function does not seem to be useful.

 Would you like me to open separate PRs to integrate the OAuth flow into the two other apps, @joeauyeung?

joeauyeung:
> > Tested this with a VPN to EU servers and this works as expected. I think we should include private events when checking for availabilities and then we should be good to merge this.
> > As a follow up I think we have to do this for other Zoho apps right?
> 
> The OAuth flow indeed needs to be implemented in the two other apps. However, the `getAvailability` function does not seem to be useful.
> 
> Would you like me to open separate PRs to integrate the OAuth flow into the two other apps, @joeauyeung?

I'm working on a PR that removes unused methods from those apps so don't worry about that part. I would like to see a separate PR changing the OAuth flows of those apps as well. Thanks!

joeauyeung (approved):
Changes look good to me. Thanks for again for the contribution @vachmara!

PR #14907, opened 2024-05-06
fix: remove global twilio variable

### LINKED ISSUE #14908: [CAL-3651] SMS sending not working with API
When creating a booking through the API SMS sending is throwing the follow error: 

`ERROR[smsReminderManager] Error sending SMS with error Error: The requested resource /2010-04-01/Accounts/.../Messages.json was not found`

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3651](https://linear.app/calcom/issue/CAL-3651/sms-sending-not-working-with-api)</sub>

CarinaWolli:
## What does this PR do?

Attempt to fix SMS sending when booking is created with the API.

Couldn't reproduce this locally but in production the following error is thrown when trying to send a SMS for a booking that was created with the API: 
`ERROR[smsReminderManager] Error sending SMS with error Error: The requested resource /2010-04-01/Accounts/.../Messages.json was not found`

Fixes #14908
Fixes https://linear.app/calcom/issue/CAL-3651/sms-sending-not-working-with-api

### DIFF
--- a/packages/features/ee/workflows/lib/reminders/providers/twilioProvider.ts
+++ b/packages/features/ee/workflows/lib/reminders/providers/twilioProvider.ts
@@ -6,25 +6,13 @@ import { setTestSMS } from "@calcom/lib/testSMS";
 import prisma from "@calcom/prisma";
 import { SMSLockState } from "@calcom/prisma/enums";
 
-declare global {
-  // eslint-disable-next-line no-var
-  var twilio: TwilioClient.Twilio | undefined;
-}
-
 const log = logger.getSubLogger({ prefix: ["[twilioProvider]"] });
 
-export const twilio =
-  globalThis.twilio ||
-  (process.env.TWILIO_SID && process.env.TWILIO_TOKEN && process.env.TWILIO_MESSAGING_SID)
-    ? TwilioClient(process.env.TWILIO_SID, process.env.TWILIO_TOKEN)
-    : undefined;
-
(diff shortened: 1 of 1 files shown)


packages/features/ee/workflows/lib/reminders/providers/twilioProvider.ts
```diff
@@ -6,25 +6,13 @@ import { setTestSMS } from "@calcom/lib/testSMS";
 import prisma from "@calcom/prisma";
 import { SMSLockState } from "@calcom/prisma/enums";
 
-declare global {
```
CarinaWolli, line 9: removing the global variable and creating TwilioClient for every function separately. Not sure if that will fix it, but that's the only thing I found that I thought could be the issue

packages/features/ee/workflows/lib/reminders/providers/twilioProvider.ts
sean-brydon, line 1: I have nothing against this approach - do you think it'd make sense to move these to some sort of class in the future and we can just reference this.client instead of creating the client every time we need it?

sean-brydon (approved):

packages/features/ee/workflows/lib/reminders/providers/twilioProvider.ts
CarinaWolli: I will revisit that, but sounds like a good idea👍

PR #14910, opened 2024-05-07
fix: save workflow button not visible on dev env

### LINKED ISSUE #14909: The "Save Workflow" button is not visible on development environment
### Issue Summary

On the development environment, the "Save" is not visible on the development environment.

### Steps to Reproduce

1. Click on "Workflows" in the sidebar
2. Click "+ Create a workflow" or edit a workflow.
3. On the top right corner, the "Save" button is not present which is supposed to be present.

This prevents the developers to test the create / edit workflow flow using the client app.

### Actual Results

Save workflow button is absent

### Expected Results

Save workflow button should be present.

### Evidence

Screenshot 2024-05-07 at 11 15 14 AM

mhetreayush:
## What does this PR do?

This PR renders the "Save" button in edit/create-workflow flow on the development environment.

Fixes #14909 

## Type of change

## How should this be tested?

1. Click on "Workflows" in the sidebar
2. Click "+ Create a workflow" or edit a workflow.
3. On the top right corner, the "Save" button should be visible.

image

## Mandatory Tasks

## Checklist


packages/features/ee/workflows/pages/workflow.tsx
```diff
@@ -277,7 +279,7 @@ function WorkflowPage() {
         }
         hideHeadingOnMobile
         heading={
-          session.data?.hasValidLicense &&
+          (IS_DEV || session.data?.hasValidLicense) &&
```
CarinaWolli, line 282: We don't need the IS_DEV check. We an always show the save button, all we need is the banner

CarinaWolli (changes requested):

packages/features/ee/workflows/pages/workflow.tsx
mhetreayush: This was added because:
1. On the local environment, session.data?.hasValidLicense evaluates to `false`, hence making the `heading` prop null.
2. The shell component only renders the CTA if the heading is not null.
3. Hence, in order to pass in the heading, this check was added.

reference: https://github.com/mhetreayush/cal.com/blob/22ccb69c6e5ba5d6ea24b24bd2535b2677abe8d9/packages/features/shell/Shell.tsx#L1030-L1066

mhetreayush: @CarinaWolli Let me know what you think about this.

CarinaWolli: we can remove the complete `IS_DEV || session.data?.hasValidLicense` check. I'd love to make it look the same as for routing forms:
Screenshot 2024-05-08 at 10 57 16 AM

mhetreayush: Sure thing

mhetreayush:
@CarinaWolli  Fixed with [f0551c8](https://github.com/calcom/cal.com/pull/14910/commits/f0551c813cc83ed96e99d1bb443bba6a5fab7dee)

image

Also fixed it on the /workflows page:

- Initial:
image

- With fix:
image

mhetreayush:
Let me know if the changes look good @CarinaWolli

CarinaWolli (approved):
Changes look good now, thank you 🙏

PR #14920, opened 2024-05-07
fix: pass query params to dynamic meetings

### LINKED ISSUE #14918: Query parameters are stripped on redirect from dynamic URLs
### Issue Summary

When adding query parameters to a dynamic URL they get stripped as it gets redirected to `…/dynamic`

 (e.g. `https://cal.com/foo+bar?duration=60`)

### Steps to Reproduce

1. Open an URL like `https://cal.com/foo+bar?duration=60`

### Actual Results

It will redirect to `https://cal.com/foo+bar/dynamic`, stripping the query parameters and select the default duration.

### Expected Results

It should redirect to `https://cal.com/foo+bar/dynamic?duration=60`, retaining the query parameters so the duration is set correctly.

### Technical details

The bug is on this line: https://github.com/calcom/cal.com/blob/a2620a0e50fb6133172de4fc53a24c773f677d46/apps/web/modules/users/views/users-public-view.getServerSideProps.tsx#L106

### Evidence

Trivial to test.

sean-brydon:
Fixes: #14918

Correctly pass query param to dynamic bookings

Ensure we select the right duration based on the value passed in


emrysal (approved):
Nice! Approved

apps/web/modules/users/views/users-public-view.getServerSideProps.tsx
```diff
@@ -104,11 +104,13 @@ export const getServerSideProps: GetServerSideProps<UserPageProps> = async (cont
 
   if (isDynamicGroup) {
     const destinationUrl = `/${usernameList.join("+")}/dynamic`;
+    const originalQueryString = new URLSearchParams(context.query as Record<string, string>).toString();
```
eelco, line 107: I would expect that all `user` values would be removed from this query string, because that is already extracted earlier and used as `usernameList`.

PR #14930, opened 2024-05-07
fix: Collective event types with seats - hosts count towards the seat count

### LINKED ISSUE #14926: [CAL-3658] Collective event types with seats - hosts count towards the seat count
We should only be counting the number of attendees towards the number of seats occupied. Not the team members associated with the collective event type.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3658](https://linear.app/calcom/issue/CAL-3658/collective-event-types-with-seats-hosts-count-towards-the-seat-count)</sub>

Amit91848:
## What does this PR do?

Fixes #14926

Before :

https://github.com/calcom/cal.com/assets/74371312/4dc9ddcd-0665-4bb9-b276-ab6aba69a1ab

After:

https://github.com/calcom/cal.com/assets/74371312/8260eead-658d-4413-9963-ad19d19ad07c

## Type of change

## Mandatory Tasks


packages/core/getUserAvailability.ts
```diff
@@ -123,8 +133,14 @@ export const getCurrentSeats = async (
...
 };
 
-const _getCurrentSeats = async (eventTypeId: number, dateFrom: Dayjs, dateTo: Dayjs) => {
-  return await prisma.booking.findMany({
+const _getCurrentSeats = async (
+  eventTypeId: number,
+  dateFrom: Dayjs,
+  dateTo: Dayjs,
+  isTeamEvent?: boolean,
+  hosts?: string[]
```
supalarry, line 141: `_getCurrentSeats` now takes in 5 parameters, which is a lot for a function. I propose combining `eventTypeId`, `isTeamEvent` and `hosts` into 1 `eventType` parameter:
```
eventType: Pick<EventType, "id" | "schedulingType" | "hosts">,
dateFrom: Dayjs,
dateTo: Dayjs
```

and then within `_getCurrentSeats` we access event type properties to check:
```
const isTeamEvent = eventType?.schedulingType !== null;
const hostEmails = eventType?.hosts.map((host) => host.user.email);
```

And to invoke function:
```
currentSeats = await getCurrentSeats(eventType, dateFrom, dateTo);
```

packages/core/getUserAvailability.ts
```diff
@@ -215,11 +258,13 @@ const _getUserAvailability = async function getUsersWorkingHoursLifeTheUniverseA
   let eventType: EventType | null = initialData?.eventType || null;
   if (!eventType && eventTypeId) eventType = await getEventType(eventTypeId);
 
+  const isTeamEvent = eventType?.schedulingType !== null;
```
supalarry, line 261: In comment about "_getCurrentSeats now takes in 5 parameters" I suggest "const isTeamEvent" logic should be moved inside "_getCurrentSeats", but then I am also thinking:

I think it's better to be explicit about when something is a team event:
```
const isTeamEvent = eventType?.schedulingType === SchedulingType.COLLECTIVE  || eventType?.schedulingType === SchedulingType.ROUND_ROBIN || eventType?.schedulingType === SchedulingType.COLLECTIVE
```

My logic for this is that currently `SchedulingType` indeed shows only scheduling types of teams, but what if someone adds a `SchedulingType` that is non-team? Then this logic would potentially break.

supalarry (commented):
Great work and thank your for the "Before" and "After" Looms - they are useful!

I have some refactors in mind:

packages/core/getUserAvailability.ts
Amit91848: We can also directly filter on hosts as only team events have hosts but that might change in the future so using hosts and schedulingType

supalarry (approved):
Great work Amit! It's nice to have this fixed.

Tested with collective and round robin and it works! : )

PR #14932, opened 2024-05-07
fix: Caching key for CI database

keithwillcode:
Fixes issues in `main` where e2e pipeline is completely broken due to caching issues. See comment added to this yml file for more details.

### DIFF
--- a/.github/actions/cache-build/action.yml
+++ b/.github/actions/cache-build/action.yml
@@ -11,17 +11,21 @@ runs:
       uses: buildjet/cache@v3
       id: cache-build
       env:
+        # WARN: Don't touch this cache key. Currently github.sha refers to the latest commit in main and not
+        # the branch that's attempting to merge to main, which causes CI errors when merged to main.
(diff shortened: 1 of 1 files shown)


emrysal (dismissed):

emrysal (approved):
Approved the comment too

zomars (approved):

keithwillcode:
We need to force merge this to main since the pipeline is broken without this change.

keithwillcode:
> We need to force merge this to main since the pipeline is broken without this change. I have manually kicked off the full suite targeted at this branch and we will only merge when that succeeds.

I take this back. Now seeing that GitHub uses the last run, manual or automatic, to determine if the checks passed. The current manual run I kicked off is currently running for this PR's checks.

keithwillcode:
> We need to force merge this to main since the pipeline is broken without this change. I have manually kicked off the full suite targeted at this branch and we will only merge when that succeeds.

Never mind, GitHub is using the latest manual run as the principal checks. Just succeed and auto-merge went through.

PR #14941, opened 2024-05-08
chore: Update PR template to include more mandatory tasks

keithwillcode:
## What does this PR do?

- Updates our PR template so that mandatory tasks we expect to be completed for PRs are properly noted.
- Removed extra parts that are redundant, like the type of change. Since we require the PR title to have a prefix which is the type of change, this section is obsolete.
- Requirement/Documentation section was rarely ever used and often we have these details in the GitHub issue instead.


alishaz-polymath (approved):

emrysal (approved):
Looks good to me.

PR #14943, opened 2024-05-08
fix: add retryCount to workflowReminder

CarinaWolli:
## What does this PR do?

Right now when SMS fails to send because of whatever reason (e.g. invalid phone number), we are trying to keep scheduling it in our corn job until the date is reached. This PR introduces a retry count, after scheduling failed twice we delete the workflow reminder.


packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -28,10 +28,19 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+      OR: [
+        {
+          method: WorkflowMethods.SMS,
+          scheduledDate: {
+            lte: dayjs().toISOString(),
+          },
+        },
+        {
+          retryCount: {
+            gt: 1,
```
CarinaWolli, line 40: After scheduling SMS failed twice we delete the workflow reminder

Udit-takkar (approved):

mfeuerstein (commented):
## PR Review — approved

Reviewed 3 files. 0 high-severity issues found. Verdict: approved.

### `packages/prisma/migrations/20240508134359_add_retry_count_to_workflow_reminder/migration.sql` (low)

- Reviewed packages/prisma/migrations/20240508134359_add_retry_count_to_workflow_reminder/migration.sql — looks good

### `packages/features/ee/workflows/api/scheduleSMSReminders.ts` (low)

- Reviewed packages/features/ee/workflows/api/scheduleSMSReminders.ts — looks good

### `packages/prisma/schema.prisma` (low)

- Reviewed packages/prisma/schema.prisma — looks good

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -28,10 +28,19 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
-      },
+      OR: [
+        {
+          method: WorkflowMethods.SMS,
+          scheduledDate: {
+            lte: dayjs().toISOString(),
+          },
+        },
+        {
+          retryCount: {
```
ron-x5labs, line 39: 🟡 **Non-blocking — deleteMany retry branch is not scoped to SMS.** The second OR clause `{ retryCount: { gt: 1 } }` has no `method: WorkflowMethods.SMS` filter, but `retryCount` is a column on every `WorkflowReminder` row (SMS, EMAIL, WHATSAPP) per `schema.prisma:1000`. Today only this SMS handler increments `retryCount`, so no non-SMS reminder is affected yet — but any future code that sets `retryCount` on an EMAIL or WHATSAPP reminder would have it silently hard-deleted by this SMS-only cron. Note the sibling `scheduleWhatsappReminders.ts` scopes its `deleteMany` with `method: WorkflowMethods.WHATSAPP` (line 25); this branch should be symmetric. Add `method: WorkflowMethods.SMS` (and optionally `scheduled: false`) to the retryCount clause.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -163,9 +175,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
               referenceId: scheduledSMS.sid,
             },
           });
+        } else {
+          await prisma.workflowReminder.update({
+            where: {
+              id: reminder.id,
+            },
+            data: {
+              retryCount: reminder.retryCount + 1,
```
ron-x5labs, line 184: 🟡 **Non-blocking — empty-message / no-recipient reminders bypass the retry counter and still retry forever.** This increment (and the success branch above) only runs inside the `if (message?.length && message?.length > 0 && sendTo)` guard at line 158. When `message` resolves to empty (null `reminderBody` and template ≠ REMINDER) or `sendTo` is falsy and no exception is thrown, neither the success nor the retry branch runs, so `retryCount` stays 0 and the `retryCount > 1` delete never fires. Such reminders are re-fetched and re-skipped every cron tick until `scheduledDate` passes — the same indefinite-retry behavior the PR set out to eliminate. Consider incrementing `retryCount` (or deleting) in this skipped path too.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -163,9 +175,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
               referenceId: scheduledSMS.sid,
             },
           });
+        } else {
+          await prisma.workflowReminder.update({
```
ron-x5labs, line 179: 🟡 **Non-blocking — silent falsy result; no log for why scheduling returned falsy.** `twilio.scheduleSMS` returns `undefined` when the recipient is SMS-locked (`twilioProvider.ts:42,94`, `smsLockState === LOCKED`) — a possibly-permanent condition. This new `else` branch silently bumps `retryCount`, so a locked recipient is retried twice then deleted with zero diagnostic trail. Add a log here (and ideally distinguish "locked" from other falsy returns) so operators can debug missing SMS without querying the DB.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -163,9 +175,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+              id: reminder.id,
+            },
+            data: {
+              retryCount: reminder.retryCount + 1,
+            },
+          });
         }
       }
     } catch (error) {
+      await prisma.workflowReminder.update({
```
ron-x5labs, line 190: 🟡 **Non-blocking — the failure-path DB write can itself throw and abort the loop.** The `prisma.workflowReminder.update` added inside `catch` (and the one at line 179) can throw — e.g. P2025 if a concurrent cron run already deleted the row, or a transient DB error. If it throws, the error escapes the catch, the `for` loop aborts, and every subsequent reminder in the batch is skipped this cron tick. The retry mechanism's own persistence step is unguarded. Wrap these updates in try/catch (or use a non-throwing write) so a failure to record a retry does not derail processing of unrelated reminders.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -28,10 +28,19 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
   //delete all scheduled sms reminders where scheduled date is past current date
```
ron-x5labs, line 28: 💡 **Suggestion — stale comment.** `// delete all scheduled sms reminders where scheduled date is past current date` no longer describes the `deleteMany`, which now ALSO deletes any reminder whose `retryCount > 1` (new OR branch, lines 38-42). Update the comment so the silent-deletion policy is documented at the source.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -44,8 +53,11 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
         lte: dayjs().add(7, "day").toISOString(),
       },
     },
-    select,
-  })) as PartialWorkflowReminder[];
+    select: {
+      ...select,
+      retryCount: true,
+    },
+  })) as (PartialWorkflowReminder & { retryCount: number })[];
```
ron-x5labs, line 60: 💡 **Suggestion — redundant hand-cast.** `as (PartialWorkflowReminder & { retryCount: number })[]` is redundant: with `select: { ...select, retryCount: true }` Prisma already infers `retryCount: number` on the payload. The manual cast can silently mask type drift if the select is later changed (e.g. removing `retryCount` from the select would still compile). Prefer extending `PartialWorkflowReminder` to include `retryCount`, or rely on the inferred Prisma type. (Not `as any`, so no AGENTS.md rule violation — minor type-safety nit.)

packages/prisma/schema.prisma
```diff
@@ -997,6 +997,7 @@ model WorkflowReminder {
   cancelled           Boolean?
   seatReferenceId     String?
   isMandatoryReminder Boolean?        @default(false)
+  retryCount          Int             @default(0)
```
ron-x5labs, line 1000: 💡 **Suggestion — no index on `retryCount`.** The `deleteMany` OR branch `{ retryCount: { gt: 1 } }` (`scheduleSMSReminders.ts:39`) cannot use the existing `@@index([method, scheduled, scheduledDate])`, so it falls back to a sequential scan over the growing `WorkflowReminder` table on every cron run. Consider `@@index([retryCount])` or a partial index for `retryCount > 1`.

ron-x5labs (changes requested):
## Code Review: fix: add retry count to workflow reminder

### Problem
When an SMS workflow reminder fails to send (e.g. invalid phone number), the cron job keeps re-scheduling it until the scheduled date passes. This PR adds a `retryCount` column and, after two failed attempts, deletes the reminder.

### Solution Reviewed
Adds `retryCount Int @default(0)` to `WorkflowReminder` (+ migration). The top-of-run `deleteMany` now also deletes rows where `retryCount > 1`. The scheduling loop increments `retryCount` on a falsy Twilio result (new `else` branch) and on a thrown error (new update inside the existing `catch`).

### Summary
The core mechanism is correct — `retryCount` 0→1→2 then deleted at the next run yields exactly two attempts, matching the stated intent, and the schema/migration are consistent. However the `deleteMany` retry branch is not scoped to SMS (a cross-method deletion hazard), the empty-message/no-recipient path still retries forever (a gap vs the PR's goal), and the new DB writes on the failure path are themselves unguarded. Recommend changes before merge.

### Files Reviewed
- `packages/features/ee/workflows/api/scheduleSMSReminders.ts` — deeply reviewed
- `packages/prisma/schema.prisma` — deeply reviewed
- `packages/prisma/migrations/20240508134359_add_retry_count_to_workflow_reminder/migration.sql` — deeply reviewed

### Verification
- Schema/migration consistency — passed: `retryCount INTEGER NOT NULL DEFAULT 0` matches `retryCount Int @default(0)` (name, type, default, NOT NULL all agree).
- Sibling convention check — passed: `scheduleEmailReminders.ts` uses `logger` (lines 8, 301, 357); `twilioProvider.scheduleSMS` returns `undefined` when the recipient is SMS-locked (`twilioProvider.ts:42,94`), so the falsy-result path silently drops locked-user reminders.
- Type-check / test suite — skipped: full `yarn type-check:ci --force` is too expensive for a review pass; the diff is small and manual verification covered the touched contracts.
- No tests exist for the new retry/delete behavior (only e2e asserts reminder count == 1 on booking; `reminderScheduler` is mocked in tests).

### Issues Found
See inline comments. Additional body-only notes:
- **Non-blocking — no tests:** the new retry-increment-on-failure and silent-deletion-after-retryCount>1 contracts are untested. A unit test exercising the falsy-result and throw paths (and the delete threshold) would guard this behavior.
- **Suggestion — pre-existing `console.log` (line 198):** `console.log(\`Error scheduling SMS with error ${error}\`)` coerces the Error to a string, dropping the stack trace, and Twilio REST errors embed the recipient phone number (attendee PII). The sibling `scheduleEmailReminders.ts` uses the structured `@calcom/lib/logger`; consider switching here (not introduced by this PR, but touched by it).

### Verdict
Recommend changes before merge — the retry mechanism is sound, but scope the delete to SMS, handle the no-message/no-recipient path, and guard the failure-path DB writes.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
ron-x5labs, line 40: 🟡 **Non-blocking — `deleteMany` retry branch is not scoped to SMS.** The second OR clause `{ retryCount: { gt: 1 } }` has no `method: WorkflowMethods.SMS` filter, but `retryCount` is a column on every `WorkflowReminder` row (SMS, EMAIL, WHATSAPP — `schema.prisma:1000`). Today only this SMS handler increments `retryCount`, so no non-SMS reminder is affected yet, but any future code that sets `retryCount > 1` on an email/whatsapp reminder would have it silently deleted here. Add the method filter to make the deletion intentional:
```diff
         {
+          method: WorkflowMethods.SMS,
           retryCount: {
             gt: 1,
           },
         },
```

packages/features/ee/workflows/api/scheduleSMSReminders.ts
ron-x5labs, line 184: 🟡 **Non-blocking — transient failures are treated as permanent; and the no-message/no-recipient path bypasses this increment entirely.** `twilio.scheduleSMS` returns `undefined` when the user/team is SMS-locked (`twilioProvider.ts:90`, `isLockedForSMSSending`) — a possibly-transient admin state — and throws via `checkSMSRateLimit` on rate-limit excess (`twilioProvider.ts:101`). Both now increment `retryCount` and are deleted after two attempts, which contradicts the PR's stated intent of giving up on *permanent* failures like an invalid phone number. Consider distinguishing transient vs permanent failure (e.g. only increment on a Twilio error indicating an unrecoverable recipient problem), or at minimum log why the result was falsy so the silent deletion leaves a diagnostic trail.

Separately, this increment (and the success branch above) only runs inside the `if (message?.length && message?.length > 0 && sendTo)` guard. When `message` resolves to empty (null `reminderBody` and template ≠ `REMINDER`) or `sendTo` is falsy and no exception is thrown, neither the success nor the retry branch runs — so `retryCount` is never incremented and the reminder is never deleted by the `retryCount > 1` clause. It silently retries every cron tick until `scheduledDate` passes, which is the exact forever-retry behavior this PR aims to eliminate.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -163,9 +175,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+          });
         }
       }
     } catch (error) {
+      await prisma.workflowReminder.update({
+        where: {
+          id: reminder.id,
+        },
+        data: {
+          retryCount: reminder.retryCount + 1,
```
ron-x5labs, line 195: 🟡 **Non-blocking — the failure-path DB write can itself throw and abort the loop.** The `prisma.workflowReminder.update` added inside `catch` (and the one at line 179) can throw — e.g. P2025 if a concurrent cron run already deleted the row, or a transient DB error. If it throws, the error escapes the catch, the `for...of` loop aborts, and every subsequent reminder in the batch is skipped this cron tick (the email sibling uses `Promise.allSettled` for batch isolation; this handler uses a serial `for...of`). Wrap the failure-path write so it cannot escape:
```ts
} catch (error) {
  try {
    await prisma.workflowReminder.update({
      where: { id: reminder.id },
      data: { retryCount: reminder.retryCount + 1 },
    });
  } catch {
    // ignore — best-effort retry counter increment
  }
  console.log(`Error scheduling SMS with error ${error}`);
}
```

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -163,9 +175,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+        },
+      });
       console.log(`Error scheduling SMS with error ${error}`);
```
ron-x5labs, line 198: 💡 **Suggestion — log PII/stack loss.** `console.log(`Error scheduling SMS with error ${error}`)` string-coerces the Error (drops the stack trace) and Twilio REST errors embed the recipient `To` phone number (attendee PII). The sibling `scheduleEmailReminders.ts` and the `twilioProvider` both use the structured `@calcom/lib/logger` (email handler lines 8, 301, 357). Switch to `logger.error("Error scheduling SMS", error)` to preserve the stack and avoid stringifying PII. (Pre-existing, but touched by this PR.)

ron-x5labs (changes requested):
## Code Review: fix: add retry count to workflow reminder

### Problem
When an SMS workflow reminder fails to send (e.g. invalid phone number), the cron job re-schedules it until the scheduled date passes. This PR adds a `retryCount` column and, after two failed attempts, deletes the reminder.

### Solution Reviewed
Adds `retryCount Int @default(0)` to `WorkflowReminder` (+ migration). The top-of-run `deleteMany` now also deletes rows where `retryCount > 1`. The scheduling loop increments `retryCount` on a falsy Twilio result (new `else` branch) and on a thrown error (new update inside the existing `catch`).

### Summary
The retry math is correct — `retryCount` 0→1→2, then `deleteMany` removes the row at the start of the next run, yielding exactly two send attempts. The schema/migration are consistent and need no backfill. However the deletion is not scoped to SMS, transient failures are treated as permanent, the no-message/no-recipient path still retries forever, and the failure-path DB write can itself throw and abort the batch. Recommend changes before merge.

### Files Reviewed
- `packages/features/ee/workflows/api/scheduleSMSReminders.ts` — deeply reviewed
- `packages/prisma/schema.prisma` — deeply reviewed
- `packages/prisma/migrations/20240508134359_add_retry_count_to_workflow_reminder/migration.sql` — deeply reviewed

### Verification
- Schema/migration consistency — passed: `retryCount INTEGER NOT NULL DEFAULT 0` matches `retryCount Int @default(0)` (name, type, default, NOT NULL all agree); no backfill needed.
- Retry math — passed: 0→1→2 over two cron runs, deleted on the third run's `deleteMany`.
- Type-check / test suite — skipped: full `yarn type-check:ci --force` is too expensive for a review pass; the diff is small and the touched contracts were verified manually.
- Test coverage gap — no test in the repo covers the new retry-increment or silent-delete behavior (`packages/features/ee/workflows` has zero test files; matches existing convention for cron handlers, but the new retry/delete contracts are load-bearing and silently executing). A unit test exercising the falsy-result, thrown-error, and delete-threshold paths would guard this behavior.

### Issues Found
See inline comments.

### Verdict
Recommend changes before merge — scope the delete to SMS, decide whether transient Twilio failures (rate-limit, SMS-locked) should really be deleted after two attempts, close the no-message/no-recipient retry-forever gap, and guard the failure-path DB write.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -28,10 +28,19 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
-        lte: dayjs().toISOString(),
-      },
+      OR: [
+        {
+          method: WorkflowMethods.SMS,
+          scheduledDate: {
+            lte: dayjs().toISOString(),
+          },
+        },
+        {
```
ron-x5labs, line 38: **Non-blocking — `deleteMany` retry branch is not scoped to SMS.** The `OR: [{ retryCount: { gt: 1 } }]` clause has no `method: WorkflowMethods.SMS` filter, so it deletes reminders of *any* method whose `retryCount > 1`. `WorkflowReminder` is shared with `EMAIL` reminders (see `getAllRemindersToDelete` filtering `method: EMAIL`). Today only this SMS handler increments `retryCount`, so the blast radius is limited, but if an email retry path ever reuses this column the SMS cron will silently delete email reminders. Suggest scoping it:
```ts
OR: [
  { method: WorkflowMethods.SMS, scheduledDate: { lte: dayjs().toISOString() } },
  { method: WorkflowMethods.SMS, retryCount: { gt: 1 } },
]
```

packages/features/ee/workflows/api/scheduleSMSReminders.ts
```diff
@@ -163,9 +175,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
               referenceId: scheduledSMS.sid,
             },
           });
+        } else {
```
ron-x5labs, line 178: **Non-blocking — transient failures are counted as permanent.** `twilio.scheduleSMS` returns `undefined` when the user/team is locked for SMS sending (`isSMSSendingLocked`), and `checkSMSRateLimit` *throws* when the per-user monthly SMS limit is exceeded. Both this `else` branch and the `catch` below increment `retryCount`, so after two cron runs under a transient lock or rate-limit the reminder is deleted by the top `deleteMany` without ever being sent. This contradicts the PR's stated intent (handle permanent failures like invalid phone numbers). Consider not incrementing on known-transient conditions, e.g.:
```ts
} else if (!isSMSSendingLocked) {
  // only count non-transient failures toward retry
  await prisma.workflowReminder.update({ where: { id: reminder.id }, data: { retryCount: reminder.retryCount + 1 } });
}
```
…and similarly narrowing the `catch` (e.g. skip `SMSRateLimitError`) so a Twilio outage or rate burst doesn't permanently drop scheduled reminders.

ron-x5labs (changes requested):
## Code Review: fix: add retry count to workflow reminder

### Problem
When an SMS workflow reminder fails to schedule (e.g. invalid phone number), the cron job kept re-attempting it until `scheduledDate` passed. This PR adds a `retryCount` column and deletes the reminder after two failed attempts.

### Solution Reviewed
Adds `retryCount Int @default(0)` to `WorkflowReminder` (+ migration with `DEFAULT 0`, safe backfill). The top-of-run `deleteMany` now also removes rows with `retryCount > 1`. The scheduling loop increments `retryCount` on a falsy Twilio result (new `else`) and on a thrown error (new update in the existing `catch`). The retry math is correct: 0→1→2 then deleted at the start of the next run yields exactly two attempts.

### Summary
The core mechanism is sound and the migration is clean, but the retry counter conflates **transient** failures (SMS-send lock, rate-limit) with **permanent** ones (invalid number), so temporary Twilio/account states can silently drop reminders. The cleanup query also lacks a `method` filter. Worth addressing before merge.

### Files Reviewed
- `packages/features/ee/workflows/api/scheduleSMSReminders.ts` — deeply reviewed
- `packages/prisma/migrations/20240508134359_add_retry_count_to_workflow_reminder/migration.sql` — lightly reviewed (DDL matches schema, safe default)
- `packages/prisma/schema.prisma` — lightly reviewed (field + default consistent with migration)

### Verification
- `yarn type-check:ci` / `yarn biome check` — skipped: the worktree has no installed toolchain (`yarn`/`node_modules` absent). Findings are grounded in source reading of `scheduleSMSReminders.ts`, `twilioProvider.ts`, and `getWorkflowReminders.ts`.

### Verdict
Recommend changes before merge — the retry logic is correct but treats transient (lock/rate-limit) and permanent failures identically, risking silent reminder loss.

packages/features/ee/workflows/api/scheduleSMSReminders.ts
ron-x5labs, line 190: **Non-blocking — `catch`-block update can abort the whole batch.** The new `await prisma.workflowReminder.update(...)` inside `catch` can itself throw (concurrent delete by the top `deleteMany`, DB blip, connection drop). Previously `catch` only did `console.log` and the loop continued to the next reminder; now a throw from this update propagates out of `catch`, aborts the `for` loop at line 67, and every remaining reminder in `unscheduledReminders` goes unprocessed for that cron run. Wrap the update in its own try/catch so one failure can't halt the batch:
```ts
} catch (error) {
  try {
    await prisma.workflowReminder.update({
      where: { id: reminder.id },
      data: { retryCount: reminder.retryCount + 1 },
    });
  } catch {
    // best-effort retry-count increment; don't let it abort the loop
  }
  console.log(`Error scheduling SMS with error ${error}`);
}
```

packages/features/ee/workflows/api/scheduleSMSReminders.ts
ron-x5labs, line 178: **Non-blocking — retry counter doesn't cover the empty-message / missing-recipient path.** This `else` (and the success branch above it) lives inside `if (message?.length && message?.length > 0 && sendTo)` (line 158, outside this hunk). When `message` is empty (no `reminderBody` and the template branch yields an empty string) or `sendTo` is missing, the entire `if` block is skipped — no success update, no `retryCount` increment, and no throw, so the `try` completes normally. `retryCount` stays 0, the `retryCount > 1` cleanup never fires, and the reminder is re-found and re-attempted every cron run until `scheduledDate` passes (then the first `deleteMany` branch finally removes it). That is the unbounded-retry behavior this PR sets out to eliminate, left uncovered for this path. Consider incrementing (or explicitly deleting) when the guard fails, e.g. adding an `else` on the outer guard:
```ts
if (message?.length && message?.length > 0 && sendTo) {
  // …existing schedule + update / else-increment…
} else {
  await prisma.workflowReminder.update({
    where: { id: reminder.id },
    data: { retryCount: reminder.retryCount + 1 },
  });
}
```

ron-x5labs (changes requested):
## Code Review (follow-up): fix: add retry count to workflow reminder

Follow-up to the prior review — two findings missed on the first pass, both confirmed against the current diff.

PR #14945, opened 2024-05-08
fix: sync book event form values with bookingData

ThyMinimalDev:
## What does this PR do?

Fix: BookEventForm values were not updated when the bookinData finished fetching

### DIFF
--- a/packages/platform/atoms/booker/BookerPlatformWrapper.tsx
+++ b/packages/platform/atoms/booker/BookerPlatformWrapper.tsx
@@ -93,6 +93,7 @@ export const BookerPlatformWrapper = (props: BookerPlatformWrapperAtomProps) => 
       setSelectedMonth(null);
       setSelectedDuration(null);
       if (props.rescheduleUid) {
+        // clean booking data from cache
         queryClient.removeQueries({
           queryKey: [BOOKING_RESCHEDULE_KEY, props.rescheduleUid],
           exact: true,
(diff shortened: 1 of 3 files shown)


packages/platform/atoms/booker/BookerPlatformWrapper.tsx
```diff
@@ -88,6 +92,14 @@ export const BookerPlatformWrapper = (props: BookerPlatformWrapperAtomProps) =>
       setOrg(null);
       setSelectedMonth(null);
       setSelectedDuration(null);
+      if (props.rescheduleUid) {
+        // clean booking data from cache
+        queryClient.removeQueries({
+          queryKey: [BOOKING_RESCHEDULE_KEY, props.rescheduleUid],
```
supalarry, line 98: Question: how do we know the `queryKey` ? Should it by any chance reflect `const key` in [useInitialFormValues](https://github.com/calcom/cal.com/pull/14945/files#diff-3c09f208072b8ff9454fb42d439f1d9b8c9d90288a9abe1ca9f5a933d7ad787eR143) ?

ThyMinimalDev: this is not directly related to the useInitialFormValues, this is to clean the data from react-query so that the booking data is fresh when the component is remounted

packages/platform/atoms/booker/BookerPlatformWrapper.tsx
```diff
@@ -88,6 +92,14 @@ export const BookerPlatformWrapper = (props: BookerPlatformWrapperAtomProps) =>
...
+          exact: true,
+        });
+        setBookingData(null);
```
ThyMinimalDev, line 101: we reset the booking data, previously the platform booker was remounted with the data from the previous booking while the fetch was in progress

supalarry (approved):

PR #14952, opened 2024-05-09
feat: Add zod schemas for app metadata living in config.json

### LINKED ISSUE #14808: [CAL-3584] Add zod schemas for app metadata living in `config.json`
When an app's metadata lives in a `config.json` file we lose typing. Instead we should have a zod schema similar to our `AppMeta` type that can parse the json file and give us proper typing.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3584](https://linear.app/calcom/issue/CAL-3584/add-zod-schemas-for-app-metadata-living-in-configjson)</sub>

smk1993:
## What does this PR do?

provides type check using "AppMeta" type  on existing config.json files .
Fixed imports on dependent files 

- Fixes #14808 (GitHub issue number)
- Fixes CAL-3584 (Linear issue number - should be visible at the bottom of the GitHub issue description)

- [x] I have added a Docs issue [here](https://github.com/calcom/docs/issues/new) if this PR makes changes that would require a [documentation change](https://docs.cal.com)

## How should this be tested?


hariombalhara:
Hey @smk1993, thanks for the huge PR 🙏

But I think we should keep the config.json files as they are auto-generated through AppStore CLI and instead just parse config.json through a zod schema.

It is easier to generate JSON files as compared to a .ts/.js file and automate it's modification through AppStore CLI.

smk1993:
> Hey @smk1993, thanks for the huge PR 🙏
> 
> But I think we should keep the config.json files as they are auto-generated through AppStore CLI and instead just parse config.json through a zod schema. That way config is still typed when we import it
> 
> It is easier to generate JSON files as compared to a .ts/.js file and automate it's modification through AppStore CLI.
> 
> Also, as zod parsing happens at run time and there are many many apps which could be a possible slow down. I was thinking that maybe we parse the config.json at compile time somehow. We want to ensure that people don't enter invalid, non-existent props in config.json.

smk1993:
Thanks @hariombalhara  for your feedback , I could try to wrap the config.json imports with a Zod schema , I will raise a sample PR for this , I will do it for all config files once its fine . 
zod takes most of the time in the parse method , however there is a huge improvement in parsing from version 3.5 . In case of slowdown zod can be upgraded to version 3.5.
https://github.com/colinhacks/zod/issues/205
Let me know your thoughts on this. 
Thanks

PR #14955, opened 2024-05-09
fix: managed user timeZone, timeFormat and weekStart during creation

supalarry:
## What does this PR do?

When creating managed user `timeZone`, `timeFormat` and `weekStart` could be provided to the input, but they were never actually passed to the function that saved managed user in the database aka they were ignored and Prisma schema defaults were used.

## Before

timeZone, timeFormat and weekStart are ignored during manged user creation:
Screenshot 2024-05-09 at 10 47 18

## After

Now, if we pass timeZone, timeFormat and weekStart we get it in the response:

Screenshot 2024-05-09 at 10 22 06

If we don't, then defaults are used:
Screenshot 2024-05-09 at 10 37 16

## Implementation

Commit [357ce66](https://github.com/calcom/cal.com/pull/14955/commits/357ce6655c12bf6fef1e8f5e99b7048ee964ca64) updates create and update inputs to rely on types instead of custom validators to enable type safety.

Commit [db09537](https://github.com/calcom/cal.com/pull/14955/commits/db09537c33ce578347fe556f3c7e776fde3b9bf0) and [6afceeb](https://github.com/calcom/cal.com/pull/14955/commits/6afceeb20c809f05ca987faa5aceb663f57aaa96) takes data from create input and passes it to function responsible for user creation.

Commit [4cf4025](https://github.com/calcom/cal.com/pull/14955/commits/4cf40255e8e3c45a1940434ccc28f277fa111d2c) updates e2e test to confirm that passed values are in response.


apps/api/v2/src/modules/users/users.repository.ts
```diff
@@ -157,7 +157,7 @@ export class UsersRepository {
 
   formatInput(userInput: CreateManagedUserInput | UpdateManagedUserInput) {
     if (userInput.weekStart) {
-      userInput.weekStart = capitalize(userInput.weekStart);
+      userInput.weekStart = userInput.weekStart;
```
ThyMinimalDev, line 160: this line can be removed I guess

apps/api/v2/src/modules/users/inputs/update-managed-user.input.ts
```diff
@@ -11,19 +11,21 @@ export class UpdateManagedUserInput {
...
   defaultScheduleId?: number;
 
-  @IsString()
   @IsOptional()
-  @Validate(IsWeekStart)
-  weekStart?: string;
+  @IsString()
+  @ApiProperty({
+    example: "Monday",
+    enum: ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"],
```
ThyMinimalDev, line 26: can we put this array in a Const in platform-constants package ?

ThyMinimalDev: or is it gonna break documentation

apps/api/v2/src/modules/users/inputs/create-managed-user.input.ts
```diff
@@ -1,8 +1,8 @@
-import { IsTimeFormat } from "@/modules/users/inputs/validators/is-time-format";
-import { IsWeekStart } from "@/modules/users/inputs/validators/is-week-start";
 import { ApiProperty } from "@nestjs/swagger";
-import { IsNumber, IsOptional, IsTimeZone, IsString, Validate } from "class-validator";
+import { IsOptional, IsTimeZone, IsString } from "class-validator";
 
+export type WeekDay = "Monday" | "Tuesday" | "Wednesday" | "Thursday" | "Friday" | "Saturday" | "Sunday";
```
ThyMinimalDev, line 4: if can we put the list of days in a Const, we can use 

type weekDay = (type WEEKDAYS)[number]

ThyMinimalDev (approved):

PR #14962, opened 2024-05-09
fix: gcal save incomplete app credentials

ThyMinimalDev:
## What does this PR do?

Fixing api v2 google credentials, token.res?.data did not contain all the informations required


apps/api/v2/src/ee/gcal/gcal.controller.ts
```diff
@@ -106,10 +107,11 @@ export class GcalController {
 
     const oAuth2Client = await this.gcalService.getOAuthClient(this.redirectUri);
     const token = await oAuth2Client.getToken(parsedCode);
-    const key = token.res?.data;
+    // Google oAuth Credentials are stored in token.tokens
+    const key = token.tokens;
```
ThyMinimalDev, line 111: token.tokens contains more information than token.res.data

keithwillcode (approved):

PR #14968, opened 2024-05-10
chore: dashboard fixes

Ryukemeister:
## What does this PR do?

This PR fixes the flickering loading state for platform dashboard. It also adds some checks for oauth client and edit pages so that we don't display those for everyone and removes displaying any kind of banner in the platform dashboard. There is also a rewrite that points the path `/settings/organizations/platform/`  to `/settings/platform/` since org and platform are two different entities and we now have the dashboard ready which was previously not present for platform user.

### DIFF
--- a/apps/web/pages/settings/platform/index.tsx
+++ b/apps/web/pages/settings/platform/index.tsx
@@ -8,15 +8,12 @@ import {
   useOAuthClients,
   useGetOAuthClientManagedUsers,
 } from "@lib/hooks/settings/platform/oauth-clients/useOAuthClients";
-import {
-  useDeleteOAuthClient,
-  useCheckTeamBilling,
-} from "@lib/hooks/settings/platform/oauth-clients/usePersistOAuthClient";
-import useMeQuery from "@lib/hooks/useMeQuery";
+import { useDeleteOAuthClient } from "@lib/hooks/settings/platform/oauth-clients/usePersistOAuthClient";
 
 import PageWrapper from "@components/PageWrapper";
 import { ManagedUserList } from "@components/settings/platform/dashboard/managed-user-list";
 import { OAuthClientsList } from "@components/settings/platform/dashboard/oauth-clients-list";
+import { useGetUserAttributes } from "@components/settings/platform/hooks/useGetUserAttributes";
 import { PlatformPricing } from "@components/settings/platform/pricing/platform-pricing";
 
 const queryClient = new QueryClient();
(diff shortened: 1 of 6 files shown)


apps/web/pages/settings/platform/index.tsx
```diff
@@ -90,7 +96,7 @@ export default function Platform() {
 
   return (
     <div>
-      <Shell hideHeadingOnMobile withoutMain={false} SidebarContainer={<></>}>
+      <Shell isPlatformUser={true} hideHeadingOnMobile withoutMain={false} SidebarContainer={<></>}>
```
ThyMinimalDev, line 99: I think this could use a comment to give some context

apps/web/pages/settings/platform/oauth-clients/[clientId]/edit/index.tsx
```diff
@@ -29,6 +31,14 @@ export default function EditOAuthClient() {
   const router = useRouter();
   const params = useParams<{ clientId: string }>();
   const clientId = params?.clientId || "";
+
+  const { data: user, isLoading } = useMeQuery();
+  const { data: userBillingData, isPending: isUserBillingDataLoading } = useCheckTeamBilling(
```
Comment (ThyMinimalDev) [inline apps/web/pages/settings/platform/oauth-clients/[clientId]/edit/index.tsx:36]: use isFetching

apps/web/pages/settings/platform/oauth-clients/create.tsx
```diff
@@ -16,6 +18,11 @@ export default function CreateOAuthClient() {
   const router = useRouter();
   const clientId = searchParams?.get("clientId") || "";
 
+  const { data: user, isLoading } = useMeQuery();
+  const { data: userBillingData } = useCheckTeamBilling(user?.organizationId);
+  const isPlatformUser = user?.organization.isPlatform;
+  const isPaidUser = userBillingData?.valid;
```
ThyMinimalDev, line 24: we should put this logic in a hook that we can reuse in both create/edit and in future pages

apps/web/pages/settings/platform/oauth-clients/[clientId]/edit/index.tsx
Comment (Ryukemeister) [reply apps/web/pages/settings/platform/oauth-clients/[clientId]/edit/index.tsx:36]: > use isFetching

The `isPending` above is actually not being used so will completely remove that.

apps/web/pages/settings/platform/index.tsx
```diff
@@ -32,7 +32,9 @@ export default function Platform() {
     data: managedUserData,
     refetch: refetchManagedUsers,
   } = useGetOAuthClientManagedUsers(initialClientId);
-  const { data: userBillingData } = useCheckTeamBilling(user?.organizationId);
+  const { data: userBillingData, isFetching: isUserBillingDataLoading } = useCheckTeamBilling(
```
ThyMinimalDev, line 35: I think we can use useGetUserAttributes here as well, if we remove the previous useMe and return isFetching in useGetUserAttributes

Ryukemeister: Indeed

ThyMinimalDev (approved):

PR #14982, opened 2024-05-10
chore: Introduce a new class WebAppURL that extends URL

### LINKED ISSUE #14981: [CAL-3685] Introduce class WebAppURL that extends URL one
Relying on string concatenation to for URLs is prone to a lot of typos and confusion. We can use URL class for this.

Also we can extend it so it has the WEBAPP_URL as base by default.

Not a blocker but let's rely on constructing URLs using the URL class or our own extended one.

https://nodejs.org/api/url.html
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

<sub>[CAL-3685](https://linear.app/calcom/issue/CAL-3685/introduce-class-webappurl-that-extends-url-one)</sub>

anikdhabal:
## What does this PR do?

- Fixes #14981 

- [x] N/A-I have added a Docs issue [here](https://github.com/calcom/docs/issues/new) if this PR makes changes that would require a [documentation change](https://docs.cal.com)
- [x] N/A-I have added or modified automated tests that prove my fix is effective or that my feature works (PRs might be rejected if logical changes are not properly tested)

## How should this be tested?

## Checklist


utkershrajvenshi:
Great work! Should other references of WEBAPP_URL also be refactored in this PR itself or a separate one?

anikdhabal:
> Great work! Should other references of WEBAPP_URL also be refactored in this PR itself or a separate one?

No, will handle it in this Pr.

supalarry (approved):
Looks good! There 500+ `WEBAPP_URL` occurrances in 200+ files in our codebase, so I think it would take too much time to go through each of them and see how and if it can be replaced with the new `WebAppURL` class, which IMO does not contribute much.

However, I think that after merging this a new entry into https://handbook.cal.com/engineering/best-practices could be added describing this, and also a post made on Engineering channel to inform people about the new standard.

supalarry:
> Looks good! There 500+ `WEBAPP_URL` occurrances in 200+ files in our codebase, so I think it would take too much time to go through each of them and see how and if it can be replaced with the new `WebAppURL` class, which IMO does not contribute much.
> 
> However, I think that after merging this a new entry into https://handbook.cal.com/engineering/best-practices could be added describing this, and also a post made on Engineering channel to inform people about the new standard.

Was thinking though, that a helper function `getWebAppURL(path: string)` could be useful - I see in codebase when we concatenate strings we do `${WEB_APP_URL}{path}` or WEB_APP_URL + `${path}` or something, which means that for this use case when `new URL` was not used, developer would have to instantiate `new WebAppUrl(path)` and then access its `.href`, but the `getWebAppURL(path: string)` could instantiate `new WebAppURL(path)` and then return its `.href`.

anikdhabal:
> Looks good! There 500+ `WEBAPP_URL` occurrances in 200+ files in our codebase, so I think it would take too much time to go through each of them and see how and if it can be replaced with the new `WebAppURL` class, which IMO does not contribute much.
> 
> However, I think that after merging this a new entry into https://handbook.cal.com/engineering/best-practices could be added describing this, and also a post made on Engineering channel to inform people about the new standard.

Thanks, Lauris. Yeah, we should inform people about the new standard, and also great explanation as always.

PR #14985, opened 2024-05-11
feat: audit log

### LINKED ISSUE #1461: Audit log

TaduJR:
## What does this PR do?
This change tries to resolve #1461 by adding audit logger on v1 and v2 api and on nextjs backend.
These are the list of events handled on Teams and Organizations
```bash
Event-Types
	- Modified
		Event Setup
			Title
			Duration
			Location
		Assignment
		        Scheduling Type
	- Created
	- Delete

Bookings
	- Modified
		Reschedule
		Edit Location
	- Created
	- Cancelled
```

- Fixes #1461 (GitHub issue number)
- Fixes CAL-1461 (Linear issue number - should be visible at the bottom of the GitHub issue description)

Handling EventType Audit Logs
https://www.loom.com/share/96a655c5a5134744bd5ea900a3542420

## How should this be tested?

- What are the minimal test data to have? Team/Organization and User Entity Records
- Any other important info that could help to test that PR?

## Checklist

/claim #1461


packages/prisma/schema.prisma
```diff
@@ -402,6 +402,7 @@ model Team {
   platformOAuthClient   PlatformOAuthClient[]
   smsLockState          SMSLockState            @default(UNLOCKED)
   platformBilling       PlatformBilling?
+  auditLog              AuditLog[]
```
zomars, line 405: Is the relationship needed at all? We can ditch the relationship and still query by teamId don't we?

packages/prisma/schema.prisma
```diff
@@ -1379,3 +1380,15 @@ model PlatformBilling {
 
   team Team @relation(fields: [id], references: [id], onDelete: Cascade)
 }
+
+model AuditLog {
+  id         Int      @id @default(autoincrement())
+  action     String
+  actionDesc String?
+  actor      Json
```
zomars, line 1388: Instead of Json can we turn this into `actorId` so we can query/join the user on query? We're trying to avoid Json fields as much as possible are these are harder to index and query.

packages/prisma/schema.prisma
```diff
@@ -1379,3 +1380,15 @@ model PlatformBilling {
...
+  target     Json?
```
zomars, line 1389: Same here on target. We can have a `targetId: 1` and `targetType: "team"` instead of a Json field.

packages/prisma/schema.prisma
```diff
@@ -1379,3 +1380,15 @@ model PlatformBilling {
...
+  crud       String
```
zomars, line 1390: This could/should be an enum:

```

enum Crud {
  CREATE
  READ
  UPDATE
  DELETE
}
```

zomars (changes requested):
This PR has a lot of potential. Thank you for your contribution 🙏 I have some observations in order to ensure this is a top quality feature.

## Delegate auditing logs to a background task

Since we're doing extra writing to the database for each action being taken. We should find a way to delegate all this logging off the main thread. It is expected that this will make all logged actions slower. But we should try to make this the least intrusive as possible. Some possible approaches that come to mind is [using a prisma middleware](https://www.prisma.io/docs/orm/prisma-client/client-extensions/middleware/logging-middleware) to avoid polluting the code itself.

## Make the feature opt-in by default

Since this codebase is shared with all self posters. We should try to make all these features opt-in by default. That way we don't suddenly introduce more unexpected DB load or side effect after an update. A possible approach would be using our current feature flags to toggle logging on or off and it is turned off by default. That way we can also switch it off without having to redeploy in case something wrong happens.
 
## Use dependency injection pattern

This is so we can easily switch to a third party service if needed. We can disable it locally. We can unit test it. We can switch it with another one. I would try to come up with a AuditLogger abstraction and create an InternalAuditLogger as our first option. Here's a [very nice video explaining the pattern if you want to dive into that](https://www.youtube.com/watch?v=J1f5b4vcxCQ).

Feel free to reach out with questions and more guidance.

​May you be well 🙏

TaduJR:
Hello @zomars. I will look into and fix every changes you requested and thank you so much for the tip.

packages/prisma/schema.prisma
TaduJR: Done.

packages/prisma/schema.prisma
TaduJR: Done.

packages/prisma/schema.prisma
TaduJR: Unfortunately, We can't turn this in to primitive data type because there are different triggers. For Example Creation of event type only require targetEvent but Updating eventtype require changedAttribute for report. So We need to wrap it into target JSON.

packages/prisma/schema.prisma
TaduJR: Yes. It is needed because We onDelete Cascade expects opposite relation from auditlog table. If we omit onDelete cascade then If a team gets deleted the audit logs of that team stays. So inorder to also delete the auditlogs belonging to the deleted team must be also deleted.

zomars: I think we want to keep the logs even if the team has been deleted. That's the point of having an audit log doesn't it?

packages/prisma/schema.prisma
```diff
@@ -139,6 +139,8 @@ model EventType {
 
   secondaryEmailId Int?
   secondaryEmail   SecondaryEmail? @relation(fields: [secondaryEmailId], references: [id], onDelete: Cascade)
+  //This is used to pass session dependent actorUserId for AuditLogger Client Extension. It is always null.
```
zomars, line 142: What do you mean by "It is always null"?

packages/prisma/schema.prisma
```diff
@@ -1392,3 +1396,31 @@ model PlatformBilling {
...
+  BookingCreate
+  BookingUpdate
+  BookingDelete
+}
+
+model AuditLog {
+  id                      Int        @id @default(autoincrement())
+  actionType              ActionType
+  actorUserId             Int?
+  legacyActorUserFullName String?
```
zomars, line 1420: Why is this legacy?

packages/prisma/migrations/20240607090533_edit_audit_log_table/migration.sql
zomars, line 1: We should only commit one migration for all schema changes. All of this is not relevant for the code in `main` since this have never been deployed before.

packages/prisma/extensions/audit-log-custom-method-client-extension.ts
```diff
@@ -0,0 +1,183 @@
...
+import { EventTypeAuditLogOption } from "@calcom/features/audit-log/types/EventTypeAuditLogTypes";
+import { prisma } from "@calcom/prisma";
+
+export function auditLogCustomMethodClientExtension() {
+  return Prisma.defineExtension({
+    query: {
+      user: {
+        async delete({ args, query }) {
+          const deletedUser = await query(args); //
+          const auditLogFeature = await prisma.feature.findUnique({ where: { slug: "audit-log" } });
```
zomars, line 14: We should have a migration that adds the record as false by default so admins can toggle it without querying the DB directly. Here's an example: https://github.com/calcom/cal.com/blob/main/packages/prisma/migrations/20230329000000_add_insights_feature_flag/migration.sql

packages/features/audit-log/saveAuditLogWorker.ts
```diff
@@ -0,0 +1,159 @@
...
+  EventTypeUpdateAuditLogger,
+  EventTypeDeleteAuditLogger,
+} from "@calcom/features/audit-log/eventTypeAuditLogger";
+import type { BookingWithAttendees } from "@calcom/features/audit-log/types/BookingAuditLogTypes";
+import { BookingAuditLogOption } from "@calcom/features/audit-log/types/BookingAuditLogTypes";
+import { EventTypeAuditLogOption } from "@calcom/features/audit-log/types/EventTypeAuditLogTypes";
+import { prisma } from "@calcom/prisma";
+import type { EventType, Prisma } from "@calcom/prisma/client";
+
+async function saveAuditLogWorker(input: any) {
```
zomars, line 16: I feel that we should avoid using `any` here'. This is very prone to critical errors.

apps/web/public/static/locales/en/common.json
```diff
@@ -2474,5 +2474,17 @@
   "USER_PENDING_MEMBER_OF_THE_ORG": "User is a pending member of the organization",
   "USER_ALREADY_INVITED_OR_MEMBER": "User is already invited or a member",
   "USER_MEMBER_OF_OTHER_ORGANIZATION": "User is member of an organization that this team is not a part of.",
-  "ADD_NEW_STRINGS_ABOVE_THIS_LINE_TO_PREVENT_MERGE_CONFLICTS": "↑↑↑↑↑↑↑↑↑↑↑↑↑ Add your new strings above here ↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑"
-}
+  "ADD_NEW_STRINGS_ABOVE_THIS_LINE_TO_PREVENT_MERGE_CONFLICTS": "↑↑↑↑↑↑↑↑↑↑↑↑↑ Add your new strings above here ↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑↑",
```
zomars, line 2477: This line should be at the very bottom.

packages/prisma/extensions/audit-log-custom-method-client-extension.ts
```diff
@@ -0,0 +1,183 @@
...
+              createdEventType,
+            });
+          }
+          return createdEventType; //
+        },
+        async update({ args, query }) {
+          const actorUserId = args.data?.actorUserId;
+          args.data.actorUserId = null;
+          let updatedEventType;
+          const auditLogFeature = await prisma.feature.findUnique({ where: { slug: "audit-log" } });
```
zomars, line 38: This could be abstracted to a function: `isAuditLogEnabled`

packages/features/audit-log/eventTypeAuditLogger.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  private eventTypeUpdateDataMaker(
+    actorUserId: number,
+    targetEventId: number,
+    targetTeamId: number,
+    prevEventType: EventType,
+    targetEventType: EventType
+  ): IEventTypeUpdateLog {
+    const changedAttributes = deepDifference(prevEventType, targetEventType);
+    const changedAttributesList = [];
```
zomars, line 99: One single record per update event instead of one record per changed field as discussed in our previous review.

zomars (changes requested):
Thanks again for the effort so far. Left some comments to address to ensure this is top quality. 🙏

packages/features/audit-log/saveAuditLogWorker.ts
TaduJR: Done.

packages/prisma/schema.prisma
TaduJR: Great. Done.

packages/prisma/extensions/audit-log-custom-method-client-extension.ts
TaduJR: Done.

packages/lib/server/queries/teams/index.ts
```diff
@@ -359,3 +362,111 @@ export async function updateNewTeamMemberEventTypes(userId: number, teamId: numb
       })
     ));
 }
+
+export async function auditLogParser(teamId: number) {
```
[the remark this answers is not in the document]
TaduJR: As you saw before, this function was used to fetch the audit logs for the frontend belonging to a certain team which was rendered on team setting page and parses common.json for the belonging locale.

packages/prisma/schema.prisma
TaduJR: It is always holds null it doesn't hold any value the reason I used it is for this feature is to hold the session dependent actorUserId. In other words, it holds who invoked the query. But after getting the actorUserId on the client extension and registering it on the auditlog it will have to value so it is better to nullify it.

packages/prisma/migrations/20240607090533_edit_audit_log_table/migration.sql
TaduJR: Ok 👍

packages/prisma/extensions/audit-log-custom-method-client-extension.ts
TaduJR: As you have read on the above the feature flag it converted into process.env.AUDITLOGENABLE

TaduJR: Done.

packages/prisma/migrations/20240607090533_edit_audit_log_table/migration.sql
TaduJR: Done.

TaduJR:
> Delegate auditing logs to a background task

Done.

> Make the feature opt-in by default

Done.

> Use dependency injection pattern

Done. But not exactly DI. I used prima client extension which changing to another auditlog service will make it very easy.

packages/prisma/migrations/20240607090533_edit_audit_log_table/migration.sql
TaduJR: Done.

packages/prisma/extensions/audit-log-custom-method-client-extension.ts
TaduJR: Done. But using .env flag

TaduJR:
Summary for all requested changes

- All requested changes by @zomars are addressed and fixed the only one thing left is rendering the saved audit logs for the browsing teams on the frontend. 

For Example: {{actorUserFullName}} created an event type named {{targetEventTypeName}} at {{createdAt}}

But Before getting to this implementation of the frontend @zomars needs to give description of how and where it should be rendered. It's one day work after @zomars decision. This is all for now.

Thank you so much @zomars for the code review.

TaduJR:
Closing this PR and moving it to https://github.com/calcom/cal.com/pull/15948

PR #14988, opened 2024-05-12
fix: Improve payload typings for Tasker handlers

### LINKED ISSUE #14979: [CAL-3683] Improve payload typings for Tasker handlers
Right now all payload just receive a string is used to stringify and parse JSON. We already use zod to parse on runtime. We should use z.infer to get that benefit on typeland.

<sub>From [SyncLinear.com](https://synclinear.com) | [CAL-3683](https://linear.app/calcom/issue/CAL-3683/improve-payload-typings-for-tasker-handlers)</sub>

anikdhabal:
## What does this PR do?

- Fixes #14979 

- [ ] I have self-reviewed the code (A decent size PR without self-review might be rejected)

## How should this be tested?

## Checklist

### DIFF
--- a/packages/features/tasker/tasks/sendEmail.ts
+++ b/packages/features/tasker/tasks/sendEmail.ts
@@ -12,8 +12,7 @@ type SendEmailPayload = z.infer<typeof sendEmailPayloadSchema>;
 
 export async function sendEmail(payload: string): Promise<void> {
   try {
-    const parsedPayloadObj: SendEmailPayload = JSON.parse(payload);
-    const parsedPayload = sendEmailPayloadSchema.parse(parsedPayloadObj);
+    const parsedPayload: SendEmailPayload = sendEmailPayloadSchema.parse(JSON.parse(payload));
     console.log(parsedPayload);
     const emails = await import("@calcom/emails");
     const email = emails[parsedPayload.template as keyof typeof emails];
--- a/packages/features/tasker/tasks/sendEmail.ts
+++ b/packages/features/tasker/tasks/sendEmail.ts
@@ -8,9 +8,12 @@ const sendEmailPayloadSchema = z.object({
   payload: z.string(),
 });
 
+type SendEmailPayload = z.infer<typeof sendEmailPayloadSchema>;
+
 export async function sendEmail(payload: string): Promise<void> {
   try {
-    const parsedPayload = sendEmailPayloadSchema.parse(JSON.parse(payload));
+    const parsedPayloadObj: SendEmailPayload = JSON.parse(payload);
+    const parsedPayload = sendEmailPayloadSchema.parse(parsedPayloadObj);
     console.log(parsedPayload);
     const emails = await import("@calcom/emails");
     const email = emails[parsedPayload.template as keyof typeof emails];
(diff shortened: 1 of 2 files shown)


packages/features/tasker/tasks/sendEmail.ts
```diff
@@ -8,9 +8,12 @@ const sendEmailPayloadSchema = z.object({
   payload: z.string(),
 });
 
+type SendEmailPayload = z.infer<typeof sendEmailPayloadSchema>;
+
 export async function sendEmail(payload: string): Promise<void> {
   try {
-    const parsedPayload = sendEmailPayloadSchema.parse(JSON.parse(payload));
+    const parsedPayloadObj: SendEmailPayload = JSON.parse(payload);
```
utkershrajvenshi, line 15: So now if the string payload we pass to this function doesn't match the `SendEmailPayload` type that we have inferred from the zod schema defined earlier, we would get a `TypeError` exception, correct?

@anikdhabal Please help me out in understanding this aspect of the changes

packages/features/tasker/tasks/sendEmail.ts
```diff
@@ -8,9 +8,12 @@ const sendEmailPayloadSchema = z.object({
...
+    const parsedPayload = sendEmailPayloadSchema.parse(parsedPayloadObj);
     console.log(parsedPayload);
```
utkershrajvenshi, line 17: Also, is this console.log intentional or can be removed safely?

packages/features/tasker/tasks/sendEmail.ts
zomars: Payload is saved in DB as string always. So no. This won't prevent getting an error on compile time. We also need to keep the runtime protection as well.

zomars (changes requested):
Thank you for your contribution 🙏 Left some comments about the approach of this PR.

The idea overall is close to what we want but in the current state is not giving us the compile time protection while removing the runtime one. Which is not an acceptable trade off.

I would recommend coming up with an approach that checks both compile time and runtime.

anikdhabal:
> Thank you for your contribution 🙏 Left some comments about the approach of this PR.
> 
> The idea overall is close to what we want but in the current state is not giving us the compile time protection while removing the runtime one. Which is not an acceptable trade off.
> 
> I would recommend coming up with an approach that checks both compile time and runtime.

Thanks @zomars for your review. I will fix that shortly.

packages/features/tasker/tasks/sendEmail.ts
utkershrajvenshi: Got it @zomars. Will have a further look

packages/features/tasker/tasks/sendEmail.ts
```diff
@@ -8,9 +8,11 @@ const sendEmailPayloadSchema = z.object({
   payload: z.string(),
 });
 
+type SendEmailPayload = z.infer<typeof sendEmailPayloadSchema>;
+
 export async function sendEmail(payload: string): Promise<void> {
   try {
-    const parsedPayload = sendEmailPayloadSchema.parse(JSON.parse(payload));
+    const parsedPayload: SendEmailPayload = sendEmailPayloadSchema.parse(JSON.parse(payload));
```
zomars, line 15: Hey @anikdhabal as far as I know, parsedPayload already has the correct type inference. These typing should be added to the packages/features/tasker/tasker.ts typings in order to keep the task creation type safe. Not in here. Will close the PR for now. 🙏

zomars (changes requested):

PR #14997, opened 2024-05-13
feat: Add zod schemas for config.json -sample PR

### LINKED ISSUE #14808: [CAL-3584] Add zod schemas for app metadata living in `config.json`

smk1993:
## What does this PR do?

- Fixes #14808 (GitHub issue number)
- Fixes CAL-3584(Linear issue number - should be visible at the bottom of the GitHub issue description)

- [ ] I have self-reviewed the code (A decent size PR without self-review might be rejected)

## How should this be tested?

PR is not complete , just added sample for config json.

### DIFF
--- a/packages/app-store/zod/configSchema.ts
+++ b/packages/app-store/zod/configSchema.ts
@@ -1,0 +1,18 @@
+import { z } from "zod";
+
+export const appConfigSchema = z.object({
+  name: z.string(),
+  slug: z.string(),
+  type: z.string(),
+  logo: z.string(),
+  url: z.string().url(),
+  variant: z.string(),
+  categories: z.array(z.string()),
+  extendsFeature: z.literal("EventType"),
+  publisher: z.string(),
+  email: z.string().email(),
+  description: z.string(),
+  __createdUsingCli: z.boolean(),
+  _template: z.string(),
+  dirName: z.string(),
+});


smk1993:
@hariombalhara  , should be done this way

hariombalhara:
Thanks @smk1993. Simply having a zod file is unnecessary cost at run time as I shared in my comment in your other PR.

But I don't see a way to solve it at compile time using zod easily. 

Going back to the drawing board, I am thinking what if we generate `metadata.generated.ts` file from config.json? 

This way the validation is compile time only. We would commit metadata.generated.ts file similar to other *.generated.ts files but it would regenerate from config.json always. [This file](https://github.com/calcom/cal.com/blob/d4f3726875aa6f9b74b73b2fd16b9019f990451f/packages/app-store-cli/src/build.ts#L1) does the generation.

hariombalhara:
Closing this PR for now as we need to solve it in a better way.

PR #15001, opened 2024-05-13
chore: open disableGuests for event-types apiv2

ThyMinimalDev:
## What does this PR do?

Api v2 event-types can disable guests

### DIFF
--- a/apps/api/v2/src/ee/event-types/controllers/event-types.controller.e2e-spec.ts
+++ b/apps/api/v2/src/ee/event-types/controllers/event-types.controller.e2e-spec.ts
@@ -131,6 +131,7 @@ describe("Event types Endpoints", () => {
         description: "A description of the test event type.",
         length: 60,
         hidden: false,
+        disableGuests: true,
         locations: [
           {
             type: "Online",
@@ -148,6 +149,7 @@ describe("Event types Endpoints", () => {
           const responseBody: ApiSuccessResponse<EventType> = response.body;
           expect(responseBody.data).toHaveProperty("id");
           expect(responseBody.data.title).toEqual(body.title);
+          expect(responseBody.data.disableGuests).toEqual(body.disableGuests);
           eventType = responseBody.data;
         });
     });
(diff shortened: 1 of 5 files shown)


apps/api/v2/src/ee/event-types/controllers/event-types.controller.e2e-spec.ts
```diff
@@ -131,6 +131,7 @@ describe("Event types Endpoints", () => {
         description: "A description of the test event type.",
         length: 60,
         hidden: false,
+        disableGuests: true,
```
supalarry, line 134: Maybe we can have a separate "it" for `disablingGuests` event type `"it should create an event type with guests disabled"` and then using fixture delete it? It's just so that we don't put all event type test cases into 1 test, and if there are other event type cases create a separate test for each.

apps/api/v2/src/ee/event-types/inputs/create-event-type.input.ts
```diff
@@ -41,6 +41,10 @@ export class CreateEventTypeInput {
   @IsArray()
   locations?: EventTypeLocation[];
 
+  @IsBoolean()
+  @IsOptional()
+  disableGuests?: boolean;
```
supalarry, line 46: I see that `EventTypesService.createUserEventType` invoked by schedule create endpoint calls `createEventType` from platform libraries that is `packages/trpc/server/routers/viewer/eventTypes/create.handler.ts`.

I checked it's schema `createEventTypeInput` and there is no `disableGuests` - it works because in the handler we have `...rest` that ends up in create body.

So we should also update schema `packages/prisma/zod/custom/eventtype.ts` by adding `disableGuests` boolean and same for update event type i guess.

apps/api/v2/src/ee/event-types/controllers/event-types.controller.e2e-spec.ts
ThyMinimalDev: I think we need to create tests cases only to test different logics, for something so trival as updating root fields I think less is more

supalarry (approved):

Issue #14966, opened 2024-05-09
[CAL-3674] I would expect that all `user` values would be removed from this query string, because that is already extracted earlier and used as `usernameList`.


sean-brydon:
I would expect that all `user` values would be removed from this query string, because that is already extracted earlier and used as `usernameList`.

_Originally posted by @eelco in https://github.com/calcom/cal.com/pull/14920#discussion_r1592845303_
            

<sub>[CAL-3674](https://linear.app/calcom/issue/CAL-3674/i-would-expect-that-all-user-values-would-be-removed-from-this-query)</sub>

siddharthji07:
``

PR #15016, opened 2024-05-14
fix a issue

siddharthji07:
this change would fix the issue #14966


Udit-takkar (changes requested):
@siddharthji07 did you forgot to push the code?

siddharthji07:
@Udit-takkar yes i forget i guess

PR #15017, opened 2024-05-14
Update handleCancelBooking.ts

### LINKED ISSUE #14992: Client Late Cancellation Fee
If a client cancels a booking within 12-24 hours, they should be charged

siddharthji07:
tried to fix issue #14992

- Fixes #14992  (GitHub issue number)

### DIFF
--- a/packages/features/bookings/lib/handleCancelBooking.ts
+++ b/packages/features/bookings/lib/handleCancelBooking.ts
@@ -155,7 +155,23 @@ async function handler(req: CustomRequest) {
   if (!bookingToDelete.userId) {
     throw new HttpError({ statusCode: 400, message: "User not found" });
   }
+  const cancellationTime = dayjs();
+  const bookingStartTime = dayjs(bookingToDelete.startTime);
+  const hoursDifference = bookingStartTime.diff(cancellationTime, 'hours');
+  
+  if (hoursDifference >= 12 && hoursDifference <= 24) {
+    // Perform your cancellation charging logic here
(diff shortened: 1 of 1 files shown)


packages/features/bookings/lib/handleCancelBooking.ts
```diff
@@ -155,7 +155,23 @@ async function handler(req: CustomRequest) {
...
+  if (hoursDifference >= 12 && hoursDifference <= 24) {
+    // Perform your cancellation charging logic here
+  }
+
+  // If the booking is a seated event and there is no seatReferenceUid we should validate that logged in user is host
+  if (bookingToDelete.eventType?.seatsPerTimeSlot && !seatReferenceUid) {
+    // Your existing validation logic here...
+  }
+
+  // Your existing code for webhooks, attendees, etc...
```
Udit-takkar, line 171: @siddharthji07 I think we should add this duration in settings because everyone wouldn't want to charge if a user cancels withing 12-24 hours.

Udit-takkar (changes requested):
@siddharthji07 Thanks for taking up this issue but try to get any missing info by commenting on the issue first. Right now this PR is incomplete.

packages/features/bookings/lib/handleCancelBooking.ts
siddharthji07: i didn't understand its a kind of charge or penality for everyone who cancelled booking in that duration

PR #15031, opened 2024-05-14
chore: more fields event-types api v2

ThyMinimalDev:
## What does this PR do?

Chore: api v2 allows settings more fields when creating / updating event-types

- slotInterval
- afterEventBuffer
- beforeEventBugger
- minimumBookingNotice

### DIFF
--- a/apps/api/v2/src/ee/event-types/controllers/event-types.controller.e2e-spec.ts
+++ b/apps/api/v2/src/ee/event-types/controllers/event-types.controller.e2e-spec.ts
@@ -132,6 +132,10 @@ describe("Event types Endpoints", () => {
         length: 60,
         hidden: false,
         disableGuests: true,
+        slotInterval: 15,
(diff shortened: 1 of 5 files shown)


emrysal (approved):
Approved, but would this mean it's theoretically possible to negative slotInterval/after/beforeEventBuffer? Because if a user tries that it may work.. but the results of that I can't predict; won't be nice.

ThyMinimalDev:
> Approved, but would this mean it's theoretically possible to negative slotInterval/after/beforeEventBuffer? Because if a user tries that it may work.. but the results of that I can't predict; won't be nice.

hmm good point, the zod schema we use on web seems to not take this case into account either

here: packages/prisma/zod/eventtype.ts

ill make a follow up pr to change that

emrysal:
Exactly, that's primarily why I approved it anyway but it was good to voice that concern imo.

PR #15036, opened 2024-05-14
fix: body payload of MEETING_STARTED & MEETING_ENDED webhook

### LINKED ISSUE #15033: Improvement in webhooks documentation
### Issue Summary

1. The documentation has mentioned general payload structure for common events like BOOKING_CREATED, BOOKING_RESHEDULED. This isn't the case for all events, for example the payload for RECORDING_READY is as show below 

```
export interface RECORDING_READY {
  type: string
  title: string
  startTime: string
  endTime: string
  organizer: Organizer
  attendees?: AttendeesEntity[] | null
  uid: string
  downloadLink: string
}
```

2. When does the MEETING_ENDED webhook trigger, does it trigger at the endDate of booking or when the meeting end call button is pressed.?  Upon testing it looks like it is triggered upon the booking end date, but I got payload as `undefined` for this particular event, which again isn't documented. Is there a webhook that can be triggerred upon the end call button press ?

3. Is there a guide on how to test each of the webhook locally ? Right now because of the 1st 2 issues mentioned above I can't rely upon the payload structure and handle each event without testing. 

### Steps to Reproduce

1. Have webhook setup done  
2. Book a meeting & wait for few hours so that MEETING_ENDED webhook gets triggered (Make sure to add logs to view the payload)
3. Payload will be logged as `undefined`

hussamkhatib:
## What does this PR do?

The PR now provides the proper body in the webhook payload for scheduled Triggers (MEETING_STARTED & MEETING_ENDED)
- Fixes #15033 
- Fixed #15078

## How should this be tested?

- Follow steps to setup webhooks in your applciation from the [docs](https://cal.com/docs/core-features/webhooks)
- Adding logs to print (`req.body.payload, req.body`)
- Trigger the MEETING_START or MEETING_ENDED event
- You should find `req.body.payload` will be `undefined` and `req.body`  having the contents of payload which isn't the format mentioned in the docs, neither do other events have.

Booking a event creates the necessary records in the database, for ease of testing set buffer time to book to 0 and booking interval to 1 min, so that MEETING_STARTED & MEETING_ENDED will be trigerred faster.

## Checklist

### DIFF
--- a/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
+++ b/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
@@ -52,12 +52,14 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
     }
 
     const headers: Record<string, string> = {
-      "Content-Type": "application/json",
+      "Content-Type":
+        !job.payload || jsonParse(job.payload) ? "application/json" : "application/x-www-form-urlencoded",
     };
 
     const parsedPayload = jsonParse(job.payload);
 
     const body = JSON.stringify({
+      ...(job.payload && parsedPayload),
       triggerEvent: parsedPayload.triggerEvent,
       payload: parsedPayload,
       createdAt: new Date().toISOString(),
--- a/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
+++ b/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
@@ -22,6 +22,7 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
     select: {
       id: true,
       jobName: true,
+      legacyPayload: true,
       payload: true,
       subscriberUrl: true,
       webhook: {
@@ -56,11 +57,11 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
         !job.payload || jsonParse(job.payload) ? "application/json" : "application/x-www-form-urlencoded",
     };
 
+    const parsedLegacyPayload = jsonParse(job.legacyPayload);
     const parsedPayload = jsonParse(job.payload);
 
     const body = JSON.stringify({
-      ...(job.payload && parsedPayload),
-      triggerEvent: parsedPayload.triggerEvent,
+      ...parsedLegacyPayload, // for ensures backward compatibility. Note: It also populate the triggerEvent which is part of the actual payload
       payload: parsedPayload,
       createdAt: new Date().toISOString(),
     });
@@ -78,12 +79,6 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
         console.error(`Webhook trigger for subscriber url ${job.subscriberUrl} failed with error: ${error}`);
       })
     );
-
-    const parsedJobPayload = JSON.parse(job.payload) as {
-      id: number; // booking id
-      endTime: string;
-      triggerEvent: string;
-    };
 
     // clean finished job
     await prisma.webhookScheduledTriggers.delete({
--- a/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
+++ b/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
@@ -52,17 +52,25 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
     }
 
     const headers: Record<string, string> = {
-      "Content-Type":
-        !job.payload || jsonParse(job.payload) ? "application/json" : "application/x-www-form-urlencoded",
+      "Content-Type": "application/json",
     };
 
+    const parsedPayload = jsonParse(job.payload);
+
+    const body = JSON.stringify({
+      triggerEvent: parsedPayload.triggerEvent,
+      payload: parsedPayload,
+      createdAt: new Date().toISOString(),
+    });
+
     if (webhook) {
-      headers["X-Cal-Signature-256"] = createWebhookSignature({ secret: webhook.secret, body: job.payload });
+      headers["X-Cal-Signature-256"] = createWebhookSignature({ secret: webhook.secret, body });
     }
+
     fetchPromises.push(
       fetch(job.subscriberUrl, {
         method: "POST",
-        body: job.payload,
+        body,
         headers,
       }).catch((error) => {
         console.error(`Webhook trigger for subscriber url ${job.subscriberUrl} failed with error: ${error}`);
--- a/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
+++ b/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
@@ -61,9 +61,11 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
     const parsedPayload = jsonParse(job.payload);
 
     const body = JSON.stringify({
-      ...parsedLegacyPayload, // for ensures backward compatibility. Note: It also populate the triggerEvent which is part of the actual payload
-      payload: parsedPayload,
-      createdAt: new Date().toISOString(),
+      ...parsedLegacyPayload, // for ensures backward compatibility. Note: It also populate the triggerEvent which is part of the both actual and legacy payload
+      ...(job.payload && {
+        payload: parsedPayload,
+        createdAt: new Date().toISOString(),
+      }),
     });
 
     if (webhook) {
--- a/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
+++ b/packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
@@ -61,7 +61,7 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
     const parsedPayload = jsonParse(job.payload);
 
     const body = JSON.stringify({
-      ...parsedLegacyPayload, // for ensures backward compatibility. Note: It also populate the triggerEvent which is part of the both actual and legacy payload
+      ...parsedLegacyPayload, // ensures backward compatibility. Note: It also populates the triggerEvent which is part of the both the actual and legacy payload
       ...(job.payload && {
         payload: parsedPayload,
         createdAt: new Date().toISOString(),
--- a/packages/features/webhooks/lib/scheduleTrigger.ts
+++ b/packages/features/webhooks/lib/scheduleTrigger.ts
@@ -3,7 +3,10 @@ import { v4 } from "uuid";
 
 import { getHumanReadableLocationValue } from "@calcom/core/location";
 import { getCalEventResponses } from "@calcom/features/bookings/lib/getCalEventResponses";
+import { getWebhookPayloadForBooking } from "@calcom/features/bookings/lib/getWebhookPayloadForBooking";
+import { type WebhookDataType } from "@calcom/features/webhooks/lib/sendPayload";
 import logger from "@calcom/lib/logger";
+import { getBooking } from "@calcom/lib/payment/getBooking";
 import { safeStringify } from "@calcom/lib/safeStringify";
 import { getTranslation } from "@calcom/lib/server";
 import prisma from "@calcom/prisma";
@@ -72,8 +75,16 @@ export async function addSubscription({
       });
 
       for (const booking of bookings) {
+        const { booking: bookingWithWebhookPayload, evt } = await getBooking(booking.id);
+
+        const webhookData = getWebhookPayloadForBooking({
+          booking: bookingWithWebhookPayload,
+          evt,
+        });
+
         scheduleTrigger({
           booking,
+          webhookData,
           subscriberUrl: createSubscription.subscriberUrl,
           subscriber: {
             id: createSubscription.id,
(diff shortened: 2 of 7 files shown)


Udit-takkar:
The unit tests are failling

packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
```diff
@@ -52,17 +52,25 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
     }
 
     const headers: Record<string, string> = {
-      "Content-Type":
-        !job.payload || jsonParse(job.payload) ? "application/json" : "application/x-www-form-urlencoded",
```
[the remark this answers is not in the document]
hussamkhatib: Because the body is always a JSON regardless of payload being undefined or not.

apps/web/test/utils/bookingScenario/expects.ts
```diff
@@ -295,8 +295,6 @@ export function expectWebhookToHaveBeenCalledWith(
       }
       if (data.payload.responses !== undefined)
         expect(parsedBody.payload.responses).toEqual(expect.objectContaining(data.payload.responses));
-      const { responses: _1, metadata: _2, ...remainingPayload } = data.payload;
-      expect(parsedBody.payload).toEqual(expect.objectContaining(remainingPayload));
```
hussamkhatib, line 299: Compare the 2 snippets at the end of the [Issue](https://github.com/calcom/cal.com/issues/15033) for why this is not needed.

CarinaWolli, line 299: let's fix the test instead of deleting it, this function is also used for other tests were it passes

packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
```diff
@@ -52,17 +52,25 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
...
+      "Content-Type": "application/json",
     };
 
+    const parsedPayload = jsonParse(job.payload);
+
+    const body = JSON.stringify({
```
CarinaWolli, line 60: I think we should be able to use `sendPayload` that would handle everything including headers, body, secret and the fetch request

CarinaWolli (changes requested):
I’ve added some comments. I’m also concerned that these changes will break existing workflows for users, as fields won't be accessible anymore as they used to

CarinaWolli:
> I’m also concerned that these changes will break existing workflows for users, as fields won't be accessible anymore as they used to

To make sure we don't break existing workflows for users, we decided to keep all fields as they are right now but additionally add the missing fields `payload` and `createdAt`

packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts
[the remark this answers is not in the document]
l0g1x: @zomars can you explain the thumbs down?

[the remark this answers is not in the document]
l0g1x: @zomars can you explain the thumbs down? just curious what scenario the content type would not be json

hussamkhatib:
> > I’m also concerned that these changes will break existing workflows for users, as fields won't be accessible anymore as they used to
> 
> To make sure we don't break existing workflows for users, we decided to keep all fields as they are right now but additionally add the missing fields `payload` and `createdAt`

Sounds good @CarinaWolli, So could I go ahead without using `sendPayload` as using that won't allow to handle existing workflow for users

hussamkhatib:
Can anyone help with the unit tests ?

CarinaWolli:
> Can anyone help with the unit tests ?

Wasn't able to make changes to your branch. This should fix the test: 

  ```
test(`should trigger if current date is after startAfter`, async () => {
    const now = dayjs();
    const payload = { triggerEvent: "MEETING_ENDED" };
    await prismock.webhookScheduledTriggers.createMany({
      data: [
        {
          id: 1,
          subscriberUrl: "https://example.com",
          startAfter: now.add(5, "minute").toDate(),
          payload: JSON.stringify(payload),
        },
        {
          id: 2,
          subscriberUrl: "https://example.com/test",
          startAfter: now.subtract(5, "minute").toDate(),
          payload: JSON.stringify(payload),
        },
      ],
    });
    await handleWebhookScheduledTriggers(prismock);

    expectWebhookToHaveBeenCalledWith("https://example.com/test", { triggerEvent: "MEETING_ENDED", payload });
    expect(() =>
      expectWebhookToHaveBeenCalledWith("https://example.com", { triggerEvent: "MEETING_ENDED", payload })
    ).toThrow("Webhook not sent to https://example.com for MEETING_ENDED. All webhooks: []");
  });
```

**Comment by hussamkhatib** [comment]
Great Thannks.

File: packages/features/webhooks/lib/handleWebhookScheduledTriggers.ts [hunk 4]
```diff
@@ -56,13 +56,23 @@ export async function handleWebhookScheduledTriggers(prisma: PrismaClient) {
         !job.payload || jsonParse(job.payload) ? "application/json" : "application/x-www-form-urlencoded",
     };
 
+    const parsedPayload = jsonParse(job.payload);
+
+    const body = JSON.stringify({
+      ...(job.payload && parsedPayload),
+      triggerEvent: parsedPayload.triggerEvent,
+      payload: parsedPayload,
```
CarinaWolli, line 64: payload still has different fields than for the `Booking created` webhook and as it's described in the docs 

We need to make sure to already save the correct payload on creation of `WebhookScheduledTriggers`. We are sending different data than for the other triggers

CarinaWolli (dismissed):

CarinaWolli:
Hey @hussamkhatib, are you still trying to work on that? Let us know if you need any help 🙏

hussamkhatib:
Yes please, I tried looking into `BOOKING_CREATED` [here](https://github.com/calcom/cal.com/blob/main/packages/features/bookings/lib/handleConfirmation.ts#L364). It needs `evt: CalendarEvent` and `eventTypeInfo : EventTypeInfo`
I was able to pass them in scheduledTriggers in one of the places where it was invoked but wasn't able to do it from other places where It was invoked.

hussamkhatib:
Since both the payload varied significantly, I made a migration so that legacyPayload and payload is stored separately.

Alternatives i considered 
- Combine and store in the same payload: The payload usually has more data in it than what is mentioned in the types defined in the source code, moreover it would be very hard to pick the keys for each of the payload.
- Store in the same Object but separate out with keys: (`{legacy: ..., payload: ...}`) This is similar to the approach I took, I din't proceed this as I thought this would be harder to differentiate old and new records and generally wouldn't be clean.

PeerRich:
/bonus 20 

@hussamkhatib would you be down to look into the merge conflict?

zomars:
> Since both the payload varied significantly, I made a migration so that legacyPayload and payload is stored separately.
> 
> Alternatives i considered
> 
> * Combine and store in the same payload: The payload usually has more data in it than what is mentioned in the types defined in the source code, moreover it would be very hard to pick the keys for each of the payload.
> * Store in the same Object but separate out with keys: (`{legacy: ..., payload: ...}`) This is similar to the approach I took, I din't proceed this as I thought this would be harder to differentiate old and new records and generally wouldn't be clean.
> 
> **the migration** `payload` is copied over to `legacyPayload` and payload is set to '' to make it a falsy value. All new records will support both actual and legacy payload. The drawback is that It might confuse users If they encounter a old and new webhook, the old webhook will not have payload in the payload key, whereas the new webhooks will have payload in the payload key as well as the old payload directly in `req.body`
> 
> Also, do you have idea how can I approach the unit tests failing ?

@hussamkhatib care to give more details on why a migration is need in the first place? What are the actual differences between payload and legacyPayload? Both system and DB wise.

packages/features/webhooks/lib/scheduleTrigger.ts
```diff
@@ -438,8 +453,20 @@ export async function updateTriggerForExistingBookings(
...
   if (addedEventTriggers.length > 0) {
     const promise = bookings.map((booking) => {
-      return addedEventTriggers.map((triggerEvent) => {
-        scheduleTrigger({ booking, subscriberUrl: webhook.subscriberUrl, subscriber: webhook, triggerEvent });
+      return addedEventTriggers.map(async (triggerEvent) => {
+        const { booking: bookingWithWebhookPayload, evt } = await getBooking(booking.id);
+        const webhookData = getWebhookPayloadForBooking({
+          booking: bookingWithWebhookPayload,
+          evt,
+        });
```
CarinaWolli, line 461: Can we move the fetching of the booking and creation of the `webhookData` outside `addedEventTriggers.map`?

hussamkhatib: Sure, thanks for catching that.

hussamkhatib:
> > Since both the payload varied significantly, I made a migration so that legacyPayload and payload is stored separately.
> > Alternatives i considered
> > 
> > * Combine and store in the same payload: The payload usually has more data in it than what is mentioned in the types defined in the source code, moreover it would be very hard to pick the keys for each of the payload.
> > * Store in the same Object but separate out with keys: (`{legacy: ..., payload: ...}`) This is similar to the approach I took, I din't proceed this as I thought this would be harder to differentiate old and new records and generally wouldn't be clean.
> > 
> > **the migration** `payload` is copied over to `legacyPayload` and payload is set to '' to make it a falsy value. All new records will support both actual and legacy payload. The drawback is that It might confuse users If they encounter a old and new webhook, the old webhook will not have payload in the payload key, whereas the new webhooks will have payload in the payload key as well as the old payload directly in `req.body`
> > Also, do you have idea how can I approach the unit tests failing ?
> 
> @hussamkhatib care to give more details on why a migration is need in the first place? What are the actual differences between payload and legacyPayload? Both system and DB wise.


You can find example of legacy and payload below.
Reason for creating a migration was solely for storing the 2 payloads separately.
The legacy payload does not matches with what's shown in the docs. (Compare the responses Object below, the legacy payload has a flatter Object compared to what's shown in the docs)

After this [change](https://github.com/calcom/cal.com/pull/15036/files#diff-995e9336658fb1aa47e3bde5b24483bef7f9dc42c94739010725cd08ecc08d93R63) user should receive payload like 

```
createdAt: Date
triggerEvent: TriggerEvent
payload: { new payload } 
...legacyPayload  
```

https://github.com/calcom/cal.com/pull/15036#issuecomment-2149895317


### Example of legacy payload 
```
{
  "triggerEvent": "MEETING_ENDED",
  "id": 40,
  "uid": "cZQKhWVNzyX6EkuFPZwwRU",
  "idempotencyKey": "97d7cb03-5d7f-5713-ae8f-1ed776321bb2",
  "userId": 6,
  "userPrimaryEmail": "free@example.com",
  "eventTypeId": 1131,
  "title": "1min between Free Example and Hussam",
  "description": "",
  "customInputs": {},
  "responses": { "name": "Hussam", "email": "hussam@shippercrm.com", "guests": [] },
  "startTime": "2024-08-11T20:54:14.000Z",
  "endTime": "2024-08-11T20:55:14.000Z",
  "location": "integrations:daily",
  "createdAt": "2024-08-11T20:51:41.424Z",
  "updatedAt": null,
  "status": "ACCEPTED",
  "paid": false,
  "destinationCalendarId": null,
  "cancellationReason": null,
  "rejectionReason": null,
  "dynamicEventSlugRef": null,
  "dynamicGroupSlugRef": null,
  "rescheduled": null,
  "fromReschedule": null,
  "recurringEventId": null,
  "smsReminderNumber": null,
  "scheduledJobs": [],
  "metadata": {},
  "isRecorded": false,
  "iCalUID": "cZQKhWVNzyX6EkuFPZwwRU@Cal.com",
  "iCalSequence": 0,
  "rating": null,
  "ratingFeedback": null,
  "noShowHost": false,
  "user": {
    "email": "free@example.com",
    "name": "Free Example",
    "timeZone": "Asia/Dubai",
    "username": "free"
  },
  "attendees": [
    {
      "id": 40,
      "email": "hussam@shippercrm.com",
      "name": "Hussam",
      "timeZone": "Asia/Dubai",
      "locale": "en",
      "bookingId": 40,
      "noShow": false
    }
  ],
  "payment": [],
  "references": []
}


```
### Example of payload (i just realized triggerEvent is missing from here, but it will still be populated from legacyPayload)
```
{
  "bookerUrl": "http://localhost:3001",
  "type": "1min",
  "title": "1min between Free Example and Hussam",
  "description": "",
  "additionalNotes": "",
  "customInputs": {},
  "startTime": "2024-08-11T20:54:14Z",
  "endTime": "2024-08-11T20:55:14Z",
  "organizer": {
    "id": 6,
    "name": "Free Example",
    "email": "free@example.com",
    "username": "free",
    "timeZone": "Asia/Dubai",
    "language": { "locale": "en" },
    "timeFormat": "h:mma"
  },
  "responses": {
    "name": { "label": "your_name", "value": "Hussam", "isHidden": false },
    "email": { "label": "email_address", "value": "hussam@shippercrm.com", "isHidden": false },
    "location": {
      "label": "location",
      "value": { "value": "integrations:daily", "optionValue": "" },
      "isHidden": false
    },
    "title": { "label": "what_is_this_meeting_about", "isHidden": true },
    "notes": { "label": "additional_notes", "isHidden": false },
    "guests": { "label": "additional_guests", "value": [], "isHidden": false },
    "rescheduleReason": { "label": "reason_for_reschedule", "isHidden": false }
  },
  "userFieldsResponses": {},
  "attendees": [
    {
      "email": "hussam@shippercrm.com",
      "name": "Hussam",
      "firstName": "",
      "lastName": "",
      "timeZone": "Asia/Dubai",
      "language": { "locale": "en" }
    }
  ],
  "location": "integrations:daily",
  "destinationCalendar": null,
  "hideCalendarNotes": false,
  "requiresConfirmation": false,
  "eventTypeId": 1131,
  "seatsShowAttendees": true,
  "seatsPerTimeSlot": null,
  "seatsShowAvailabilityCount": true,
  "schedulingType": null,
  "iCalUID": "cZQKhWVNzyX6EkuFPZwwRU@Cal.com",
  "iCalSequence": 0,
  "uid": "cZQKhWVNzyX6EkuFPZwwRU",
  "conferenceData": { "createRequest": { "requestId": "2db644eb-37a5-581a-99fa-ebe6ce513834" } },
  "eventTitle": "1min",
  "eventDescription": "",
  "price": 0,
  "currency": "usd",
  "length": 1,
  "bookingId": 40,
  "metadata": {},
  "status": "ACCEPTED"
}

```

anikdhabal:
@hussamkhatib could you pls fix the conflict

anikdhabal:
Closing due to staleness.

PR #15053, opened 2024-05-15
fix: platform onboarding improvements

Ryukemeister:
## What does this PR do?

This PR adds the following things to the platform onboarding experience:
- Skips the getting started and confirm email pages for platform users at the time of account creation
- Add discount code option to stripe at the time of platform billing checkout 
- Platform billing cards responsiveness (4 column, then 2 column then 1 column

### DIFF
--- a/apps/web/pages/signup.tsx
+++ b/apps/web/pages/signup.tsx
@@ -234,6 +234,7 @@ export default function Signup({
   };
 
   const isOrgInviteByLink = orgSlug && !prepopulateFormValues?.username;
+  const isPlatformUser = redirectUrl?.includes("platform") && redirectUrl?.includes("new");
 
   const signUp: SubmitHandler<FormValues> = async (_data) => {
     const { cfToken, ...data } = _data;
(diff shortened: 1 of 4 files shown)


apps/web/pages/signup.tsx
```diff
@@ -255,14 +256,35 @@ export default function Signup({
...
+
+        const constructCallBackIfUrlPresent = () => {
+          if (isOrgInviteByLink) {
+            return `${WEBAPP_URL}/${searchParams.get("callbackUrl")}`;
+          }
+
+          return addOrUpdateQueryParam(`${WEBAPP_URL}/${searchParams.get("callbackUrl")}`, "from", "signup");
+        };
+
+        const constructCallBackIfUrlNotPresent = () => {
```
ThyMinimalDev, line 271: not sure to understand the naming of this function, what does ifUrlNotPresent means

ThyMinimalDev: I understand it's the callback url from searchParams but maybe we can find a better name or refactor to make stuff more clear

ThyMinimalDev: `getUrlIfCallbackNotProvided`

apps/web/pages/signup.tsx
```diff
@@ -255,14 +256,35 @@ export default function Signup({
...
-        const callBackUrl = `${
-          searchParams?.get("callbackUrl")
-            ? isOrgInviteByLink
-              ? `${WEBAPP_URL}/${searchParams.get("callbackUrl")}`
-              : addOrUpdateQueryParam(`${WEBAPP_URL}/${searchParams.get("callbackUrl")}`, "from", "signup")
-            : `${WEBAPP_URL}/${verifyOrGettingStarted}?from=signup`
-        }`;
+        const gettingStartedWithPlatform = "settings/platform/new";
+
+        const constructCallBackIfUrlPresent = () => {
```
ThyMinimalDev, line 263: we never need to check for platform in this case ?

apps/web/pages/signup.tsx
Ryukemeister: @ThyMinimalDev should we rename the functions to `constructCallBackUrlIfSearchParamsPresent` and `constructCallBackUrlIfSearchParamsNotPresent`?

apps/web/pages/signup.tsx
Ryukemeister: We do check for platform on line 272

ThyMinimalDev (approved):

PR #15077, opened 2024-05-16
feat: outlook 365 calendar endpoints

Ryukemeister:
## What does this PR do?

This PR adds the endpoints for the microsoft outlook calendar. The frontend can be done in a follow PR once the endpoints are merged. Also added some e2e tests for case where access token or origin might be invalid and also when all the credentials are good


apps/api/v2/src/ee/calendars/controllers/calendars.controller.ts
```diff
@@ -55,4 +80,102 @@ export class CalendarsController {
       data: calendars,
     };
   }
+
+  @UseGuards(AccessTokenGuard)
+  @Get("/office365/connect")
```
supalarry, line 85: As I understand, gcal and microsoft outlook have the same enpoints - getting redirect url, save and check.

First, why do we call `@Get("/office365/connect")` here but in `gcal.controller.ts` its `@Get("/oauth/auth-url")` ?

Next, since we will have 1 calendars endpoint and ultimately get rid of gcal controller (as i remember our discussion), we want to re-use route logic for google calendar, microsoft outlook and other calendars that follow this structure.

Secondly, that means that we need a dynamic route `@Get("/:calendar/connect")` that then based on `:calendar` be it google or microsoft outlook invokes their corresponding services be it `GCalService` or `MicrosoftOutlookService`.

Also, we can define an interface `CalendarApp` foe example that looks like (can probably have different function parameters but this is just example):
```
interface CalendarApp {
  redirect(authorization: string, req: Request): Promise<{
    status: string;
    data: {
      authUrl: string;
    };
  }>;

  save(state: string, code: string): Promise<{
    url: string;
  }>;

  check(userId: number): Promise<{
    status: string;
  }>;
}
```

Then, the MicrosoftOutlookService and GCalService will implement it.

supalarry (dismissed):
I think we need to remove microsoft outlook from "calendars.service.ts" into it's own `MicrosoftOutlookService` and then don't define in "calendars.controller.ts" microsoft outlook explicitly in routes but make it dynamic.

packages/platform/constants/apps.ts
```diff
@@ -1,6 +1,13 @@
 export const GOOGLE_CALENDAR_TYPE = "google_calendar";
 export const GOOGLE_CALENDAR_ID = "google-calendar";
+export const OFFICE_365_CALENDAR_TYPE = "office365_calendar";
+export const OFFICE_365_CALENDAR_ID = "office365-calendar";
+export const GOOGLE_CALENDAR = "google";
+export const MICROSOFT_OUTLOOK_CALENDAR = "google";
```
ThyMinimalDev, line 6: office365

packages/platform/constants/apps.ts
```diff
@@ -1,6 +1,13 @@
 export const GOOGLE_CALENDAR_TYPE = "google_calendar";
 export const GOOGLE_CALENDAR_ID = "google-calendar";
+export const OFFICE_365_CALENDAR_TYPE = "office365_calendar";
+export const OFFICE_365_CALENDAR_ID = "office365-calendar";
+export const GOOGLE_CALENDAR = "google";
```
ThyMinimalDev, line 5: gcal

apps/api/v2/src/ee/calendars/calendars.interface.ts
```diff
@@ -0,0 +1,10 @@
+import { Request } from "express";
+
+export interface CalendarApp {
+  save(state: string, code: string, origin: string): Promise<{ url: string }>;
+  check(userId: number): Promise<{ status: string }>;
```
ThyMinimalDev, line 5: `check(userId: number): Promise<ApiResponse>;`

apps/api/v2/src/ee/calendars/calendars.interface.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+export interface OAuthCalendarApp extends CalendarApp {
+  connect(authorization: string, req: Request): Promise<{ status: string; data: { authUrl: string } }>;
```
ThyMinimalDev, line 9: `  connect(authorization: string, req: Request): Promise<ApiResponse<{ authUrl: string }>>

apps/api/v2/src/ee/calendars/services/gcal.service.ts
```diff
@@ -0,0 +1,151 @@
...
+    private readonly tokensRepository: TokensRepository,
+    private readonly selectedCalendarsRepository: SelectedCalendarsRepository
+  ) {}
+
+  async connect(authorization: string, req: Request): Promise<{ status: string; data: { authUrl: string } }> {
+    const accessToken = authorization.replace("Bearer ", "");
+    const origin = req.get("origin") ?? req.get("host");
+    const redirectUrl = await await this.getCalendarRedirectUrl(accessToken, origin ?? "");
+
+    return { status: "SUCCESS", data: { authUrl: redirectUrl } };
```
ThyMinimalDev, line 41: `{ status: SUCCESS_STATUS, data: { authUrl: redirectUrl } };`

ThyMinimalDev (approved):

ThyMinimalDev (approved):

PR #15082, opened 2024-05-16
fix: Add organization logo to org dynamic events

emrysal:
## What does this PR do?

* Previous fix removed the duplicate avatar, but didn't bring in the org logo. This adjusts the dynamic logic to return the correct profile.


apps/web/playwright/dynamic-booking-pages.e2e.ts
```diff
@@ -91,6 +91,16 @@ test("dynamic booking info prefilled by query params", async ({ page, users }) =
   activeState = await badgeLocator.getAttribute("data-active");
   expect(activeState).toEqual("false");
 });
+// eslint-disable-next-line playwright/no-skipped-test
+test.skip("it contains the right event details", async ({ page }) => {
+  const response = await page.goto(`http://acme.cal.local:3000/owner1+member1`);
+  expect(response?.status()).toBe(200);
```
sean-brydon, line 97: NIT: this is a redundant assertion 

the next expects will always be false if this page is not a 200

sean-brydon (approved):
CleanShot 2024-05-21 at 11 26 26
Tested and works well nice job

PR #15083, opened 2024-05-16
feat: org-wide workflows

### LINKED ISSUE #14638: [CAL-3489] org-wide workflows
that can be added and removed to an entire org

<sub>[CAL-3489](https://linear.app/calcom/issue/CAL-3489/org-wide-workflows)</sub>

CarinaWolli:
## What does this PR do?

- Org-workflow can be set active on teams and will trigger for all the team event types + all the user event types of the team members. 
  - if a user is member of several active teams, the workflow only triggers once 
- Add 'apply to all, including future teams/event-types'
- `workflows/update.handler` is refactored into smaller functions that can be reused and tested. 
- Also fixes some small existing bugs that I found when refactoring the code

Todo:
- [x] When member is removed from team and workflow has 'active on all, including future' set then reminders aren't deleted (reminders should be deleted if user isn't member of any other team)
- [x] should org workflow be triggered if user is not part of any team?
- [x] write unit test for `deleteWorkfowRemindersOfRemovedMember`

Follow ups: 
- Create Workflow repository to handle data operations
- Improve types, for example: https://github.com/calcom/cal.com/pull/15083/files#diff-40b6e69d629325f5d2e47bb383fce1847ed9f69f763d0b92ae8067477fad2d24R30 
- https://github.com/calcom/cal.com/pull/15083#discussion_r1648373166

Fixes #14638
Fixes CAL-3489

- [x] I have added a Docs issue [here](https://github.com/calcom/docs/issues/new) if this PR makes changes that would require a [documentation change](https://docs.cal.com) --> Created an issue for that: https://github.com/calcom/docs/issues/93

## How should this be tested?

Basic test: 
- Create an org workflow
- Set active on a team (or all teams)
- Book a team event type --> see if workflow triggers 
- Book a user event type of a team member --> see if workflow triggers

Test other cases, for example:
- test with several teams when some team members are part of several teams --> a workflow should always just trigger once
- Test updating a 'before event' workflow and see if `WorkflowReminders` are correctly adjusted 
- Test removing team workflow and see if reminders are deleted (if user is still part of another team that is active, the reminder should stay)
- delete/add reminders for deleted/added steps
- Add reminders for new active teams
- Test if 'active on all teams/even-types, including future ones' works with new event types/teams


hariombalhara:
Workflow count is wrong on the left
image

hariombalhara:
I think to clearly communicate the meaning, we should change the message to `Active on all teams and their members`

image

hariombalhara (commented):
Getting my head around workflows right now. So, submitting partial review

hariombalhara (commented):
Getting my head around workflows right now. So, submitting partial review

hariombalhara:
Have we handled the cleanup of workflows when a user is removed from a team. In my testing I found that the reminder remains even when there is no team.

No Team is there. The user was a member of a team earlier
image

But workflow that came from Org is still there
image

packages/trpc/server/routers/viewer/workflows/util.ts
```diff
@@ -114,3 +245,533 @@ export async function removeSmsReminderFieldForBooking({
...
+
+  removedTeams.forEach((teamId) => {
+    const reminderToDelete = prisma.workflowReminder.findMany({
+      where: {
+        OR: [
+          {
+            //team event types + children managed event types
+            booking: {
+              eventType: {
+                OR: [{ teamId: teamId }, { parentId: teamId, teamId: null }],
```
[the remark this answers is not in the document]
CarinaWolli: yes, that's a bug. 

I think what I want to do here is: 

`
eventType: {
	OR: [
		{ teamId: teamId }, 
		{ 
			teamId: null,
			parent: {
				team: {
					id: teamId
				}
			}
		}
	]
}
`

CarinaWolli:
> I think to clearly communicate the meaning, we should change the message to `Active on all teams and their members`
> 
> Initially I thought it could be just the team events but it is those team's members' events as well.
> 
> image

What do you think about adding that information here?
Screenshot 2024-06-12 at 12 21 36 PM

hariombalhara:
> > Have we handled the cleanup of workflows when a user is removed from a team. In my testing I found that the reminder remains even when there is no team.
> 
> Should be handled here: https://github.com/calcom/cal.com/pull/15083/files#diff-00296740ffe644796cf7e681b96fe07cd4fa0283d1cef7ef345635f789a34cc9L1
> 
> do you have 'active on all including future teams' enabled? I think in that case it currently doesn't remove the reminder, which it should tho (https://github.com/calcom/cal.com/pull/15083/files#diff-00296740ffe644796cf7e681b96fe07cd4fa0283d1cef7ef345635f789a34cc9R160)

Yeah that was the case exactly

hariombalhara:
> > I think to clearly communicate the meaning, we should change the message to `Active on all teams and their members`
> > 
> > Initially I thought it could be just the team events but it is those team's members' events as well.
> > 
> > image
> 
> What do you think about adding that information here?
> Screenshot 2024-06-12 at 12 21 36 PM
> 

Yeah sounds good to me.

CarinaWolli:
> Workflow count is wrong on the left. 

I fixed the workflow count: https://github.com/calcom/cal.com/pull/15083/commits/303798e7239cafd278a3d4167289807dbf7efc42

apps/web/modules/event-types/views/event-types-single-view.tsx
```diff
@@ -692,7 +693,7 @@ const EventTypePage = (props: EventTypeSetupProps) => {
       <EventTypeSingleLayout
         enabledAppsNumber={numberOfActiveApps}
         installedAppsNumber={eventTypeApps?.items.length || 0}
-        enabledWorkflowsNumber={eventType.workflows.length}
+        enabledWorkflowsNumber={props.allActiveWorkflows.length || 0}
```
hariombalhara, line 696:
```suggestion
        enabledWorkflowsNumber={props.allActiveWorkflows.length}
```

hariombalhara:
Toggle status bug
https://www.loom.com/share/f05e5d66df15410c8c40454cdaaefb25

apps/web/modules/event-types/views/event-types-single-view.tsx
```diff
@@ -841,7 +842,24 @@ const EventTypePageWrapper: React.FC<PageProps> & {
...
+  const eventType = data.eventType;
+  const eventTypeWorkflows = eventType.workflows.map((workflowRel) => workflowRel.workflow);
+
+  const { data: workflows, isPending } = trpc.viewer.workflows.getAllActiveWorkflows.useQuery({
+    eventTypeWorkflows: eventTypeWorkflows,
+    teamId: eventType.team?.id,
+    userId: eventType.owner?.id,
+  });
+
+  if (isPending) return null;
```
hariombalhara, line 855: Can be improved in a follow-up but I think we shouldn't block the rendering of entire event-type page just for the workflows count.

apps/web/public/static/locales/en/common.json
```diff
@@ -1236,6 +1236,8 @@
   "may_require_confirmation": "May require confirmation",
   "nr_event_type_one": "{{count}} event type",
   "nr_event_type_other": "{{count}} event types",
+  "nr_team_one": "{{count}} team",
+  "nr_team_other": "{{count}} teams",
```
hariombalhara, line 1240: I think we could use a more descriptive name, I really don't know what `nr` means here.

packages/features/bookings/lib/handleConfirmation.ts
```diff
@@ -66,6 +68,38 @@ export async function handleConfirmation(args: {
...
+    ? !eventTypeMetadata?.managedEventConfig?.unlockedFields?.workflows
+    : false;
+
+  const workflows = await getAllWorkflows(
+    eventTypeWorkflows,
+    evt.organizer.id,
+    workflowsLockedForUser ? eventType?.parent?.teamId : teamId,
+    orgId
+  );
+
```
hariombalhara, line 102: This is some really important stuff. We should abstract it out and reuse here and in `handleCanceBooking` and `handleNewBooking` and maybe more places.

I think `getAllWorkflowsFromEventType` would be a good name. You give it an eventType and it does the rest.

packages/features/bookings/lib/handleNewBooking.ts
```diff
@@ -2203,9 +2212,7 @@ async function handler(
         let isHostConfirmationEmailsDisabled = false;
         let isAttendeeConfirmationEmailDisabled = false;
 
-        const workflows = eventType.workflows.map((workflow) => workflow.workflow);
-
-        if (eventType.workflows) {
+        if (workflows) {
```
hariombalhara, line 2215: `workflows` is an array, so it isn't needed.

packages/features/bookings/lib/handleSeats/cancel/cancelAttendeeSeat.ts
```diff
@@ -140,23 +138,11 @@ async function cancelAttendeeSeat(
...
-      } else if (reminder.method === WorkflowMethods.SMS) {
-        return deleteScheduledSMSReminder(reminder.id, reminder.referenceId);
-      } else if (reminder.method === WorkflowMethods.WHATSAPP) {
-        return deleteScheduledWhatsappReminder(reminder.id, reminder.referenceId);
-      }
-    });
-
-    await Promise.allSettled(deletionPromises);
-  }
+  await deleteAllWorkflowReminders(workflowRemindersForAttendee);
```
hariombalhara, line 145: Earlier we were using `Promise.allSettled` which would allow all reminders to be deleted even if one of them errors during deletion.


But `deleteAllWorkflowReminders` is using `Promise.all` now.

packages/features/bookings/lib/handleSeats/create/createNewSeat.ts
```diff
@@ -117,9 +118,7 @@ const createNewSeat = async (
     let isHostConfirmationEmailsDisabled = false;
     let isAttendeeConfirmationEmailDisabled = false;
 
-    const workflows = eventType.workflows.map((workflow) => workflow.workflow);
-
-    if (eventType.workflows) {
+    if (workflows) {
```
hariombalhara, line 121:
```suggestion
```

packages/features/ee/teams/lib/removeMember.ts
```diff
@@ -101,16 +113,91 @@ const removeMember = async ({
...
+        organizationId: team.id,
       }),
     ]);
+
+    // delete all workflowReminders of the removed team member that come from org workflows
+    const workflowRemindersToDelete = await prisma.workflowReminder.findMany({
+      where: {
+        workflowStep: {
+          workflow: {
+            teamId: team.id,
```
hariombalhara, line 125: nit: 

```const orgId = team.id;```
```suggestion
            teamId: orgId,
```

packages/features/ee/teams/lib/removeMember.ts
```diff
@@ -101,16 +113,91 @@ const removeMember = async ({
...
+        },
+      },
+      select: {
+        id: true,
+        referenceId: true,
+        method: true,
+      },
+    });
+
+    deleteAllWorkflowReminders(workflowRemindersToDelete);
```
hariombalhara, line 141: We should abstract these lines of code out in smaller functions with good names to convey the right meaning and make them unit testable as they can make or break things.

packages/features/ee/teams/lib/removeMember.ts
```diff
@@ -101,16 +113,91 @@ const removeMember = async ({
...
+                some: {
+                  userId: memberId,
+                },
+              },
+            },
+          },
+        },
+        ...(isUserActiveOnOtherTeam && {
+          isActiveOnAll: false,
+        }),
```
hariombalhara, line 174: I don't get this logic

packages/features/ee/teams/lib/removeMember.ts
hariombalhara: Also, as a followup maybe we should move these queries to workflow repository

hariombalhara:
I had created 3 workflows that directly or indirectly targeted a team(T1)
WF-1: Org Workflow -> With all future events selected(So it targets team T1 events as well)
WF-2: Org Workflow -> T1 Team.
WF-3: Team Workflow for T1

Steps:
1. Book a team T1 event say t1/30min
2. Observe that there are three reminders created
3. Disband the team
4. Observe that 2 reminders still exists corresponding to WF-1 and WF-2

apps/web/modules/event-types/views/event-types-single-view.tsx
CarinaWolli: Looks like this wasn't needed and actually introduced the [toggle status bug ](https://github.com/calcom/cal.com/pull/15083#issuecomment-2165019863)

packages/features/ee/teams/lib/removeMember.ts
```diff
@@ -101,16 +114,103 @@ const removeMember = async ({
...
+                },
+              },
+            },
+            // if user is still a member of other teams in the org, we also need to make sure that the found workflow is not active on all teams
+            ...(isUserMemberOfOtherTeams && {
+              isActiveOnAll: false,
+            }),
+          },
+          {
+            // workflows of the org that are set active on all teams (and the user is not member of any other team)
```
CarinaWolli, line 181: this case wasn't covered before: 

- if activeOnAll is true and user is not part of any other org team

CarinaWolli, line 181: this case wasn't covered before: 

- if activeOnAll is true and user is not part of any other org team

packages/features/ee/teams/lib/removeMember.ts
CarinaWolli: I added comments to the code to explain, also found another case that wasn't covered: https://github.com/calcom/cal.com/pull/15083/files#r1638549350

packages/features/bookings/lib/handleConfirmation.ts
CarinaWolli: good call! done: https://github.com/calcom/cal.com/pull/15083/files#diff-5d5a2dd3d089d283853e724977b9ae82d83612101436022a8e6c6fe83d4d604fR72

packages/features/ee/workflows/lib/getAllWorkflows.ts
```diff
@@ -0,0 +1,115 @@
...
+          },
+          team: true,
+        },
+      });
+
+      const orgUserWorkflows = orgUserWorkflowsRel.map((workflowRel) => workflowRel.workflow) ?? [];
+      allWorkflows.push(...orgUserWorkflows);
+    }
+
+    const activeOnAllOrgWorkflows = await prisma.workflow.findMany({
```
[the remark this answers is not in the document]
hariombalhara: Isn't it the case of all org members regardless of the team?

hariombalhara:
I want to confirm this as I am confused.

An org workflow which has ''include future events' flag enabled is applicable to the org members that aren't part of any team right?

packages/features/bookings/lib/handleSeats/cancel/cancelAttendeeSeat.ts
CarinaWolli: changed it to `promise.allSettled` in `deleteAllWorkflowReminders` and improved the error message

CarinaWolli:
> An org workflow which has ''include future events' flag enabled is applicable to the org members that aren't part of any team right?

That's something I was uncertain as well. But as it says 'Apply to all, including future teams', I think members that aren't part of a team shouldn't be included. That's why I implemented it like that in `removeMember` handler now.  

However, I still would need to fix `getAllWorkflows`, it also returns the org workflow if the user is not part of any team. 

Let me know if you think different about that

packages/features/ee/teams/lib/removeMember.ts
CarinaWolli: > Also, as a followup maybe we should move these queries to workflow repository

I agree with that, I will create a follow up issue 🙏

packages/trpc/server/routers/viewer/teams/delete.handler.ts
```diff
@@ -44,4 +51,60 @@ export const deleteHandler = async ({ ctx, input }: DeleteOptions) => {
...
+            },
+          },
+        });
+        remainingActiveOnIds = allRemainingOrgTeams.map((team) => team.id);
+      } else {
+        remainingActiveOnIds = workflow.activeOnTeams
+          .filter((activeOn) => activeOn.teamId !== team.id)
+          .map((activeOn) => activeOn.teamId);
+      }
+      deleteRemindersOfActiveOnIds([team.id], workflowSteps, true, remainingActiveOnIds);
```
CarinaWolli, line 105: Deletes all reminders associated with the workflow steps that are active on the deleted team and that originate from bookings made by the team or its team members. It also ensures that reminders are not deleted if the user associated with the booking is still part of another active team.

CarinaWolli:
> I had created 3 workflows that directly or indirectly targeted a team(T1) WF-1: Org Workflow -> With all future events selected(So it targets team T1 events as well) WF-2: Org Workflow -> T1 Team. WF-3: Team Workflow for T1
> 
> Steps:
> 
> 1. Book a team T1 event say t1/30min
> 2. Observe that there are three reminders created
> 3. Disband the team
> 4. Observe that 2 reminders still exists corresponding to WF-1 and WF-2

Nice find, deleting a team wasn't handled yet. This should fix the issue you described: https://github.com/calcom/cal.com/pull/15083/files#diff-2e09b54c1171b0e62fad257d21e021061f76bf6ad2a72c13d948d96be6127239R27

CarinaWolli:
> However, I still would need to fix getAllWorkflows, it also returns the org workflow if the user is not part of any team.

Fixed that. Now, if a user in't part of any team then org workflows with 'active on all, including future teams' will not trigger

hariombalhara:
> > However, I still would need to fix getAllWorkflows, it also returns the org workflow if the user is not part of any team.
> 
> Fixed that. Now, if a user in't part of any team then org workflows with 'active on all, including future teams' will not trigger
> 

I think we have taken different decisions for this at different places. This is getting confusing. Maybe we can confirm from @ciaranha @PeerRich 

e.g. An app installed at an org level is available for org members regardless of being a subteam member AFAIK
Not sure how webhooks work.

CarinaWolli:
@ciaranha's answer to "should org workflow be triggered if user is not part of any team?":

> I think the answer to this has to be yes. They're still in the org.

I fixed that now and also changed the text for the checkbox. I think that's more accurate: 
Screenshot 2024-06-17 at 4 02 26 PM

apps/web/modules/event-types/views/event-types-single-view.tsx
```diff
@@ -840,7 +841,22 @@ const EventTypePageWrapper: React.FC<PageProps> & {
   const { data } = trpc.viewer.eventTypes.get.useQuery({ id: props.type });
 
   if (!data) return null;
-  return <EventTypePage {...(data as EventTypeSetupProps)} />;
+
+  const eventType = data.eventType;
+  const eventTypeWorkflows = eventType.workflows.map((workflowRel) => workflowRel.workflow);
+
+  const { data: workflows, isPending } = trpc.viewer.workflows.getAllActiveWorkflows.useQuery({
```
hariombalhara, line 848:
```suggestion
  const { data: workflows } = trpc.viewer.workflows.getAllActiveWorkflows.useQuery({
```

sean-brydon (commented):
WOW!  This deserves a round of applause

packages/trpc/server/routers/viewer/workflows/util.ts
```diff
@@ -114,3 +250,587 @@ export async function removeSmsReminderFieldForBooking({
...
+  const eventTypeMetadata = EventTypeMetaDataSchema.parse(eventType?.metadata || {});
+
+  const workflowsLockedForUser = isManagedEventType
+    ? !eventTypeMetadata?.managedEventConfig?.unlockedFields?.workflows
+    : false;
+
+  const allWorkflows = await getAllWorkflows(
+    eventTypeWorkflows,
+    userId,
+    workflowsLockedForUser ? eventType?.parent?.teamId : teamId,
```
hariombalhara, line 831: Can we simply not pass `teamId` which is aware of getting the teamId from parent for a managed event?

CarinaWolli (commented):
some more fixes for managed event types and enabling/disabling workflows if the workflow has `isActiveOnAll` disabled

apps/web/modules/event-types/views/event-types-single-view.tsx
```diff
@@ -840,7 +840,26 @@ const EventTypePageWrapper: React.FC<PageProps> & {
   const { data } = trpc.viewer.eventTypes.get.useQuery({ id: props.type });
 
   if (!data) return null;
-  return <EventTypePage {...(data as EventTypeSetupProps)} />;
+
+  const eventType = data.eventType;
+
+  const { data: workflows } = trpc.viewer.workflows.getAllActiveWorkflows.useQuery({
+    eventType: {
+      workflows: eventType.workflows,
```
hariombalhara, line 848: This query can quickly become big because we are sending in all the workflows from client.

See the payload for just 1 workflow.
As a followup maybe we can just let the workflows be computed at backend only.
image

apps/web/modules/event-types/views/event-types-single-view.tsx
```diff
@@ -840,7 +840,26 @@ const EventTypePageWrapper: React.FC<PageProps> & {
...
+      teamId: eventType.teamId,
+      userId: eventType.userId,
+      parentId: eventType.parentId,
+      parent: eventType.parent,
```
hariombalhara, line 852: nit: We could have just parent which can have `parent.id ` same as parentId.

packages/trpc/server/routers/viewer/workflows/getAllActiveWorkflows.handler.ts
```diff
@@ -0,0 +1,19 @@
...
+
+type GetAllActiveWorkflowsOptions = {
+  ctx: {
+    user: NonNullable<TrpcSessionUser>;
+  };
+  input: TGetAllActiveWorkflowsInputSchema;
+};
+
+export const getAllActiveWorkflowsHandler = async ({ input }: GetAllActiveWorkflowsOptions) => {
+  const { eventType } = input;
```
hariombalhara, line 14: image
This doesn't seem to have any authorization check and I am able to retrieve different org's workflows

packages/trpc/server/routers/viewer/bookings/requestReschedule.handler.ts
```diff
@@ -149,24 +147,16 @@ export const requestRescheduleHandler = async ({ ctx, input }: RequestReschedule
...
-  }
-
-  await Promise.all(webhookWorkflowPromises).catch((error) => {
-    log.error("Error while scheduling or canceling webhook triggers", JSON.stringify({ error }));
+  await Promise.all(webhookPromises).catch((error) => {
+    log.error("Error while deleting scheduled webhook triggers", JSON.stringify({ error }));
   });
 
+  //cancel workflow reminders of previous booking
+  await deleteAllWorkflowReminders(bookingToReschedule.workflowReminders);
```
hariombalhara, line 158: If we are fine not passing `prisma` as the argument here, then maybe we could simply stop accepting that argument in `deleteAllWorkflowReminders`.

packages/trpc/server/routers/viewer/teams/delete.handler.ts
hariombalhara, line 105: nit: It would make sense to turn this function to accept a single argument with all these arguments as props.

It usually is a good idea when there are more than 2 arguments.

apps/web/modules/event-types/views/event-types-single-view.tsx
CarinaWolli: Yes agree, I was planning to do that. I create an issue with follow-ups: https://linear.app/calcom/issue/CAL-3964/follow-up-org-wide-workflows

packages/trpc/server/routers/viewer/workflows/getAllActiveWorkflows.handler.ts
CarinaWolli: Thank you for catching that 🙏 added an authorization check now

hariombalhara (dismissed):
Amazing work @CarinaWolli 🎉 It is a complex addition to workflows affecting a lot of places.

It was great reviewing this with your speed of iteration as well !!


I am not sure if we should merge this now or wait till after the retreat as it is a big PR.

hariombalhara:
> Amazing work @CarinaWolli 🎉 It is a complex addition to workflows affecting a lot of places.
> 
> It was great reviewing this with your speed of iteration as well !!
> 
> I am not sure if we should merge this now or wait till after the retreat as it is a big PR.

Also, someone else who has more experience with workflows should also review this to ensure the entire functionality is working

Udit-takkar (commented):
@CarinaWolli  can i disable global org workflow on any event type?

I see the message "Locked by team admin" even on my personal event types
Screenshot 2024-06-25 at 3 13 42 PM

CarinaWolli:
> @CarinaWolli can i disable global org workflow on any event type?
> 
> I see the message "Locked by team admin" even on my personal event types

org workflows can't be disabled on any event type, that's by design

Udit-takkar (dismissed):
LGTM. Great work. 

tested org, team and user workflows.
The only thing we can improve is to divide the large functions ( like scheduleBookingReminders) into smaller, more manageable functions.

keithwillcode:
@emrysal @zomars Just spoke with @CarinaWolli and we are going to wait until post-retreat to merge this. Moving back to draft for that reason.

CarinaWolli:
Fixed merge conflicts caused by https://github.com/calcom/cal.com/pull/15590/files (getEventTypesFromDB)

hariombalhara (approved):

